PR #23, opened 2019-01-07
Add contact details to provider

dankmitchell:
### Context
Providers endpoint

Add contact details i.e. `address1`, `address2`, `address3`, `address4`, `postcode`

### Guidance to review
`/api/v1/providers`

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -5,4 +5,39 @@ class Provider < ApplicationRecord
 
   has_many :sites
   has_many :enrichments, foreign_key: :provider_code, primary_key: :provider_code, class_name: "ProviderEnrichment"
+
+  def address
+    if enrichments.present?
+      attrs = enrichments.last.json_data
+      {
+        "address1" => attrs["Address1"],
+        "address2" => attrs["Address2"],
+        "address3" => attrs["Address3"],
+        "address4" => attrs["Address4"],
+        "postcode" => attrs["Postcode"]
+      }
+    else
+      attributes
+    end
+  end
+
+  def address1
+    address["address1"]
+  end
+
+  def address2
+    address["address2"]
+  end
+
+  def address3
+    address["address3"]
+  end
+
+  def address4
+    address["address4"]
+  end
+
+  def postcode
+    address["postcode"]
+  end
 end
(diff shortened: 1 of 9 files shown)


app/serializers/provider_serializer.rb
```diff
@@ -19,4 +19,44 @@ def institution_type
   def accrediting_provider
     # TODO: pull this thru from UCAS import
   end
+
+  def address1
+    if object.enrichments.present?
+      object.enrichments.last.json_data["Address1"]
+    else
+      object.address1
+    end
```
misaka: Feels to me like the model should be doing this, not the serializer. i.e. it should decide how to return the correct address info within it's class.

misaka: Also, @benilovj quite rightly mentioned and I agree that it might make more sense to actually return the address as a hash via a single attribute, `address` or something similar, instead of having each line as a separate attribute at the top of the data structure.

misaka:
Also, I noticed this isn't being used: https://github.com/devmynd/jsonb_accessor

Might be useful here, although still need to see if it'd handle the existing data (e.g. I don't think it supports nested data, do we have any?)

app/models/provider.rb
```diff
@@ -4,4 +4,40 @@ class Provider < ApplicationRecord
...
+      attrs = enrichments.last.json_data
+      {
+        "address1" => attrs["Address1"],
+        "address2" => attrs["Address2"],
+        "address3" => attrs["Address3"],
+        "address4" => attrs["Address4"],
+        "postcode" => attrs["Postcode"]
+      }
+    else
+      attributes
```
benilovj: This returns more than just the address, no?

misaka: It does. It's not the tidiest, true, but the whole ordeal is a bit messy anyway. I'd be up for tidying it up in the future, I reckon we'll be revisiting this bit of code anyway.

dankmitchell:
It didn’t help that the provider attributes are a different “case” to the enrichment attributes I.e. address1 vs Address1. The enrichments definitely needs refactoring.

defong:
[[M] Add contact details form](https://trello.com/c/M7Jhkztn/23-m-add-contact-details-form)

PR #25, opened 2019-01-10
Dockerise the app, build the container on CI, and push it to Azure

tvararu:
### Context

This builds a docker image of the app when we push to CI, runs the tests inside, and then pushes it to the Azure Container Registry under the `master|staging|production` tags respectively.

Also includes instructions on running the app locally using `docker-compose` to better mimic the live environment, and simplify the setup by not requiring specific versions of Ruby of Postgres installed locally.

Paired with @defong.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.bundle
+log
+tmp
+.byebug_history
+coverage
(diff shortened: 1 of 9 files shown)


tvararu:
Closing this PR to see if the travis-ci.com builds disappear.

PR #26, opened 2019-01-10
Dockerise the app, build the container on CI, and push it to Azure

tvararu:
### Context

This builds a docker image of the app when we push to CI, runs the tests inside, and then pushes it to the Azure Container Registry under the `master|staging|production` tags respectively.

Also includes instructions on running the app locally using `docker-compose` to better mimic the live environment, and simplify the setup by not requiring specific versions of Ruby or Postgres installed locally.

Paired with @defong.

### DIFF
--- a/docker-login.sh
+++ b/docker-login.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -e
+
+if [ -z "$AZURE_CR_PASSWORD" ];
+then
+  echo "AZURE_CR_PASSWORD environment variable is required"
+  exit 1
+fi
+
+echo "Logging in to docker host '$DOCKER_REGISTRY_HOST' ..."
+echo "$AZURE_CR_PASSWORD" | docker login "$DOCKER_REGISTRY_HOST" -u="batdevcontainerregistry" --password-stdin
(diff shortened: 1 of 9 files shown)


docker-login.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+if [ -z "$AZURE_CR_PASSWORD" ];
```
misaka: Where does this get defined? Travis?

tvararu: Yes, this Base64 looking thing encodes the variable name too: https://github.com/DFE-Digital/manage-courses-backend/pull/26/files#diff-354f30a63fb0907d4ad57269548329e3R26

misaka: Ah, to answer my own question, this is likely the `secure` value in `travis.yml`. Is it work documenting how to set this up/change it?

tvararu: @misaka Your call if you like me to add it, but the other repos don't bother, and I think it's not something that would trip us up if we don't ¯\_(ツ)_/¯

tvararu: @misaka documented! :)

defong:
[Use Docker to deploy manage-courses-backend to Azure on dev environment](https://trello.com/c/I2b9J8ey/830-use-docker-to-deploy-manage-courses-backend-to-azure-on-dev-environment)

PR #29, opened 2019-01-11
Fix provider contacts

misaka:
### Context

Previous attempt to display contact details for providers had a bug in it, this is a fix.

Re-did how we figure out the address info for a provider. Also added `jsonb_accessor` to try to make JSON attribute access easier.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,9 @@ gem 'kaminari'
 # Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin AJAX possible
 # gem 'rack-cors'
 
+# Access jsonb attributes like normal ActiveRecord model attributes
+gem 'jsonb_accessor'
+
 group :development, :test do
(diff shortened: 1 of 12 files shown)


Gemfile
```diff
@@ -34,6 +37,13 @@ group :development, :test do
 
   # Test framework
   gem 'rspec-rails'
+
+  # Prettyprint in console
+  gem 'awesome_print'
+
+  # A little extra console goodness
+  gem 'pry'
+  gem 'pry-byebug'
```
tvararu: <3

app/controllers/api/v1/providers_controller.rb
```diff
@@ -3,9 +3,12 @@ module V1
     class ProvidersController < ApplicationController
       def index
         @providers = Provider.all
-
         paginate json: @providers
       end
+
+      def show
```
dankmitchell: I think this can go. I only added it for testing

PR #30, opened 2019-01-11
Add wait-for-it.sh, wait for the database to spin up in CI

tvararu:
The sleep 5 was a race condition that ended up flaking some of our test runs.


tvararu:
screen shot 2019-01-11 at 18 03 11

PR #32, opened 2019-01-14
Added site region code

defong:
### Context
site `region_code`

Added `region_code` to site
region_code can be a nullable int

### DIFF
--- a/app/serializers/site_serializer.rb
+++ b/app/serializers/site_serializer.rb
@@ -1,5 +1,5 @@
 class SiteSerializer < ActiveModel::Serializer
-  attributes :campus_code, :name, :recruitment_cycle
+  attributes :campus_code, :name, :region_code, :recruitment_cycle
 
   def campus_code
     object.code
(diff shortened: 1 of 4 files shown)


defong:
@dankmitchell for `gttr_campus` the listed values are 0 to 10 and empty

defong:
[Add region code to sites.](https://trello.com/c/06ZwlDji/847-add-region-code-to-sites)

app/serializers/site_serializer.rb
```diff
@@ -9,6 +9,10 @@ def name
     object.location_name
   end
 
+  def region_code
```
misaka: Won't `ActiveModel::Serializer` automatically refer to `object.region_code` for us for this attribute?

dankmitchell: Ah, yes it will. I should have spotted that.

misaka: Let's take this out, then.

benilovj:
The `region_code` type in the implementation is integer, whereas in the documentation it's text: https://github.com/DFE-Digital/manage-courses-external-api-docs/pull/7. We need to align these.

PR #33, opened 2019-01-14
Present multiple (truncated) provider name versions in API

benilovj:
### Context
This is required by downstream consumers.

Provide multiple provider name versions of various lengths


defong:
Is there a reason for truncating values for them, other then they cannot be asked?

benilovj:
Closing for now.

PR #34, opened 2019-01-14
Added region code for provider

defong:
### Context
`region_code` for provider

Added `region_code` for provider

### Guidance to review
Similar for the site/`gttr_campus`'s `region_code`, its nullable number, through data suggests that there is no unset 

With regards to `region_code` who is the single source of truth and where does one get such a value for it from?

As this stems directly from the provider enrichment, where by the provider can amend their address therefore the question of 

what is a `region_code`?
what is a `region_code` use for?
Asserting validity of a `region_code`, as this don't seems to be related to geo location data, so is it arbitrary?

We also need to revisit provider enrichment as it now seems to suggest that it can be amended.

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -38,7 +38,11 @@ class Provider < ApplicationRecord
   has_many :sites
   has_many :enrichments, foreign_key: :provider_code, primary_key: :provider_code, class_name: "ProviderEnrichment"
   def address_info
-    (enrichments.with_address_info.last || self)
-      .slice('address1', 'address2', 'address3', 'address4', 'postcode', 'region_code')
+    @last_enrichment = enrichments.with_address_info.last
+    @address_info = (@last_enrichment || self)
+      .slice('address1', 'address2', 'address3', 'address4', 'postcode')
+    @region_code = @last_enrichment.present? ? @last_enrichment[:region_code] : self.region_code_before_type_cast
+    @address_info[:region_code] = @region_code
+    return @address_info
   end
 end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -37,6 +37,6 @@ class Provider < ApplicationRecord
   has_many :enrichments, foreign_key: :provider_code, primary_key: :provider_code, class_name: "ProviderEnrichment"
   def address_info
     (enrichments.with_address_info.last || self)
-      .slice('address1', 'address2', 'address3', 'address4', 'postcode')
+      .slice('address1', 'address2', 'address3', 'address4', 'postcode', 'region_code')
   end
 end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -38,11 +38,11 @@ class Provider < ApplicationRecord
   has_many :sites
   has_many :enrichments, foreign_key: :provider_code, primary_key: :provider_code, class_name: "ProviderEnrichment"
   def address_info
-    @last_enrichment = enrichments.with_address_info.last
-    @address_info = (@last_enrichment || self)
+    last_enrichment = enrichments.with_address_info.last
+    address_info = (last_enrichment || self)
       .slice('address1', 'address2', 'address3', 'address4', 'postcode')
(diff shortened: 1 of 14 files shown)


defong:
The data from the import is integer. Can we change the documentation instead?

defong:
[Add region code to providers.](https://trello.com/c/8sHoAlPi/848-add-region-code-to-providers)

spec/serializers/site_serializer_spec.rb
```diff
@@ -2,7 +2,27 @@
...
+
   subject { serialize(site) }
 
-  it { is_expected.to include(name: site.location_name, campus_code: site.code, region_code: site.region_code) }
+  it { is_expected.to include(name: site.location_name, campus_code: site.code) }
+  it { is_expected.to include(region_code: site.region_code) }
+end
+
+
+RSpec.describe SiteSerializer do
```
dankmitchell: @misaka can you help here? I don't think we want 2 describe blocks

misaka: Nest this describe into the above one ... 

```
RSpec.describe SiteSerializer do
  ...

  describe 'SiteSerializer#region_code' do
    ...
  end
end
```

spec/serializers/provider_serializer_spec.rb
```diff
@@ -14,3 +14,24 @@
   it { should include(institution_type: provider.provider_type) }
   it { should include(accrediting_provider: nil) }
 end
+
+RSpec.describe ProviderSerializer do
```
misaka: As discussed, move this up into the previous `describe` block.

app/models/provider.rb
```diff
@@ -36,7 +38,11 @@ class Provider < ApplicationRecord
   has_many :sites
   has_many :enrichments, foreign_key: :provider_code, primary_key: :provider_code, class_name: "ProviderEnrichment"
   def address_info
-    (enrichments.with_address_info.last || self)
+    @last_enrichment = enrichments.with_address_info.last
+    @address_info = (@last_enrichment || self)
       .slice('address1', 'address2', 'address3', 'address4', 'postcode')
+    @region_code = @last_enrichment.present? ? @last_enrichment[:region_code] : self.region_code_before_type_cast
```
misaka: I'd suggest you actually include the `region_code` enum into `ProviderEnrichment` instead of doing this juggling around here. You may be able to then do:

```
(enrichments.with_address_info.last || self)
  .attributes_before_type_cast
```

spec/serializers/provider_serializer_spec.rb
```diff
@@ -35,4 +35,33 @@
...
+
+    describe "provider region code 'London' can be overriden by enrichment region code 'Scotland'" do
+      let(:enrichment) do
+        build(:provider_enrichment, region_code: "Scotland")
+      end
+
+      let(:provider) { create :provider, region_code: "London", enrichments: [enrichment] }
+      it { is_expected.not_to eql(format("%02d", 1)) }
+      it { is_expected.to eql(format("%02d", 11)) }
+      it { expect(subject.length).to eql(2) }
```
misaka: Just some minor niggles here:

- `format("%02d", 11)` can be `'%02d' % 11`, which in turn can simply just be `'11'`.
- likewise for `'01'`
- testing the length is already done with `is_expected.to eql '11'`

spec/serializers/provider_serializer_spec.rb
```diff
@@ -35,4 +35,33 @@
...
+    end
+
+    region_code = 1
+    describe "provider region code 00 is overriden with #{region_code} " do
+      let(:enrichment) do
+        build(:provider_enrichment, region_code: region_code)
+      end
```
misaka: Same as above here, but also I would not use `region_code` here as a variable, the way that's usually done here is with a let:

```
describe 'overriding region_code' do
  let(:enrichment) do
    build(:provider_enrichment, region_code: region_code)
  end
  let(:region_code) { 1 }

  ...
end
```

misaka:
@defong will you squash these before merging?

PR #35, opened 2019-01-14
Site region_code enum to increase readability

benilovj:
### Context
Similar to #21, #22 and #28.

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -1,6 +1,8 @@
 class Site < ApplicationRecord
   self.table_name = "site"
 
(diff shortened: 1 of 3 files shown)


app/models/site.rb
```diff
@@ -1,5 +1,21 @@
 class Site < ApplicationRecord
   self.table_name = "site"
 
+  # These correspond to the first-level NUTS regions for the UK (minus Northern Ireland)
+  # https://en.wikipedia.org/wiki/First-level_NUTS_of_the_European_Union#United_Kingdom
+  enum region_code: {
```
defong: Can we have one more for nil, as that was exhibited in import

PR #36, opened 2019-01-15
Change region_code to be a 2-digit string as specified in API docs

benilovj:
### Context
Before this change, the `region_code` datatype didn't match the docs.


defong:
Looks like it will fit in with https://github.com/DFE-Digital/manage-courses-backend/pull/34

PR #42, opened 2019-01-17
Add build and detach to readme

timabell:
Second run didn't work for me due to stale container, so add build flag.

Add detach to explicitly background it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 Run this in a shell and leave it running:
 
 ```
-$ docker-compose up --build --detach
+docker-compose up --build --detach
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,7 @@
 Run this in a shell and leave it running:
 
 ```
-$ docker-compose up
+$ docker-compose up --build --detach
```
tvararu: I don't think we want to detach by default. The Rails server has good output.

timabell: you can get the output with 

    docker-compose logs --follow

would that do?

PR #43, opened 2019-01-17
Swap basic auth for token

dankmitchell:
### Context
HTTP Authentication

Swap basic auth for token based auth

### Guidance to review
```
curl "api_endpoint_here"  -H "Authorization: Bearer bat"
```


app/controllers/application_controller.rb
```diff
@@ -0,0 +1,11 @@
+class ApplicationController < ActionController::API
+  include ActionController::HttpAuthentication::Token::ControllerMethods
+
+  before_action :authenticate
+
+  def authenticate
+    authenticate_or_request_with_http_token do |token|
+      AUTHENTICATION.include?(token)
```
misaka: In the docs for [ActionController::HttpAuthentication::Token](https://api.rubyonrails.org/classes/ActionController/HttpAuthentication/Token.html) they use `secure_compare` which sounds like worth doing to me.

misaka: Does it make sense to use `AUTHENTICATION` as a hash here? Also, you're checking with the keys, which are the usernames.

I think what we want here is a single secret, similar to what they have in the docs linked-to above.

dankmitchell: Ah yes, great find. Thanks @misaka

dankmitchell:
[Token authorisation for the UCAS API](https://trello.com/c/3YyIOWlF/866-token-authorisation-for-the-ucas-api)

PR #45, opened 2019-01-18
Add accrediting_provider to provider

dankmitchell:
### Context
Is a provider and accrediting_provider?

Add accrediting_provider data to provider endpoint

### Guidance to review
`api/v1/providers`

### DIFF
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -39,7 +39,7 @@ class ProviderSerializer < ActiveModel::Serializer
   end
 
   def accrediting_provider
-    # TODO: pull this thru from UCAS import
+    object.scitt || 'N'
   end
 
   def address1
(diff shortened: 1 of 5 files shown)


app/serializers/provider_serializer.rb
```diff
@@ -39,7 +39,11 @@ def institution_type
   end
 
   def accrediting_provider
-    # TODO: pull this thru from UCAS import
+    if object.scitt.present?
```
benilovj: I think you can replace these 5 lines with `object.scitt || 'N'`

spec/factories/providers.rb
```diff
@@ -29,6 +29,7 @@
     address3 { Faker::Address.city }
     address4 { Faker::Address.state }
     postcode { Faker::Address.postcode }
+    scitt 'N'
```
benilovj:
```suggestion
    scitt { 'N' }
```
otherwise I'm not sure you can override this?

benilovj: Apologies, I think it was probably OK before, it needs the block if it's dynamic which it isn't in this case. Anyhoo the block doesn't hurt.

PR #46, opened 2019-01-21
changed field on course_site from text to date

MartinBentley:
### Context
This is the task to change the field "applications_accepted_from" from text to date.
Here is the card:
https://trello.com/c/w7kXfqpK/872-fix-datatype-for-applicationsacceptedfrom

Small change to the serializer


MartinBentley:
[Fix datatype for 'applications_accepted_from'](https://trello.com/c/w7kXfqpK/872-fix-datatype-for-applicationsacceptedfrom)

MartinBentley:
[Fix datatype for 'applications_accepted_from'](https://trello.com/c/w7kXfqpK/872-fix-datatype-for-applicationsacceptedfrom)

MartinBentley:
[Fix datatype for 'applications_accepted_from'](https://trello.com/c/w7kXfqpK/872-fix-datatype-for-applicationsacceptedfrom)

app/serializers/site_status_serializer.rb
```diff
@@ -26,12 +26,9 @@ def status
     object.status_before_type_cast
   end
 
-  # rubocop:disable Style/DateTime
   def course_open_date
-    # TODO applications_accepted_from should be a timestamp, not a string
-    DateTime.parse(object.applications_accepted_from).iso8601 if object.applications_accepted_from.present?
+    object.applications_accepted_from.iso8601 if object.applications_accepted_from.present?
```
dankmitchell: Do we need to convert to `iso8601`? I thought the data is stored in the correct format for us now?

dankmitchell:
@MartinBentley can you squash all the commits into one, please.

app/serializers/site_status_serializer.rb
MartinBentley: You're right. I don't think we need this especially as the data for this field has no time element.

MartinBentley:
> @MartinBentley can you squash all the commits into one, please.

I'll squash commit when I merge

app/serializers/site_status_serializer.rb
```diff
@@ -26,12 +26,9 @@ def status
     object.status_before_type_cast
   end
 
-  # rubocop:disable Style/DateTime
   def course_open_date
-    # TODO applications_accepted_from should be a timestamp, not a string
-    DateTime.parse(object.applications_accepted_from).iso8601 if object.applications_accepted_from.present?
+    object.applications_accepted_from if object.applications_accepted_from.present?
```
misaka: We can do away with the `if ...` guard, since we don't need to convert this anymore. The only possible values are Date or nil.

MartinBentley: Done

PR #48, opened 2019-01-22
Add an app-specific user to use when connecting to the db

misaka:
This keeps our DB-use isolated from other apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,14 +39,6 @@ On Linux:
 sudo apt install postgresql-9.6 postgresql-contrib-9.6
 ```
 
-### Creating the DB user
-
-Before creating the db create the dev user:
-
-```bash
-rails db:create_dev_user
-```
-
 ## Accessing API
 
 [See API Docs](https://github.com/DFE-Digital/manage-courses-backend/blob/master/docs/api.md)
--- a/README.md
+++ b/README.md
@@ -31,6 +31,14 @@ Then open http://localhost:3000 to see the app.
 
 ## Setting up the app for local (non-Docker) development
 
+### Installing prerequisites
+
+On Linux:
+
+```bash
+sudo apt install postgresql-9.6 postgresql-contrib-9.6
+```
+
 ### Creating the DB user
 
 We use a project-specific Postgres user which needs to be created before you create and use the databases.
--- a/README.md
+++ b/README.md
@@ -41,18 +41,10 @@ sudo apt install postgresql-9.6 postgresql-contrib-9.6
 
 ### Creating the DB user
 
-We use a project-specific Postgres user which needs to be created before you create and use the databases.
-
-On OSX:
+Before creating the db create the dev user:
 
 ```bash
-psql -U postgres < db/create_dev_user.sql
-```
-
-On Linux where there is a system `postgres` user:
-
-```bash
-sudo -u postgres psql < db/create_dev_user.sql
+rails db:create_dev_user
 ```
 
 ## Accessing API
--- a/README.md
+++ b/README.md
@@ -29,6 +29,24 @@ docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
 
 Then open http://localhost:3000 to see the app.
 
+## Setting up the app for local (non-Docker) development
+
+### Creating the DB user
+
+We use a project-specific Postgres user which needs to be created before you create and use the databases.
+
+On OSX:
+
+```bash
+psql -U postgres < db/create_dev_user.sql
+```
+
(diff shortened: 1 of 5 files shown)


config/database.yml
```diff
@@ -2,8 +2,8 @@ default: &default
   adapter: postgresql
   encoding: unicode
   pool: 5
-  username: <%= ENV['DB_USERNAME'] || 'postgres' %>
-  password: <%= ENV['DB_PASSWORD'] || '' %>
+  username: <%= ENV['DB_USERNAME'] || 'manage_courses_backend' %>
```
tvararu: I'm going to check if we need to update the docker setup to handle this, but we probably do.

README.md
```diff
@@ -29,6 +29,24 @@ docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
...
+## Setting up the app for local (non-Docker) development
+
+### Creating the DB user
+
+We use a project-specific Postgres user which needs to be created before you create and use the databases.
+
+On OSX:
+
+```bash
+psql -U postgres < db/create_dev_user.sql
```
dankmitchell: Why do we need this? Surely `rake db:setup` is enough?

config/database.yml
tvararu: Works fine. 👍

tvararu:
Ah, Travis is broken however:

```
FATAL:  role "manage_courses_backend" does not exist
Couldn't create 'manage_courses_backend_development' database. Please check your configuration.
rails aborted!
ActiveRecord::NoDatabaseError: FATAL:  role "manage_courses_backend" does not exist
```

docker-compose.yml
```diff
@@ -22,3 +22,4 @@ services:
       - db
     environment:
       - DB_HOSTNAME=db
+      - DB_USERNAME=postgres
```
tvararu: 👍

README.md
dankmitchell: You will have a default user on your OS, why create a new one?

defong:
Can some one dumb this down for me, its getting a bit too confusing looking at this, 

ie, 
there is a readme for creating user
there is a sql for creating user
there is various default user and passwords
there is a mixture of docker and non docker

timabell:
@defong let's do that as a further iteration

README.md
```diff
@@ -29,6 +29,24 @@ docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
...
 Then open http://localhost:3000 to see the app.
 
+## Setting up the app for local (non-Docker) development
+
+### Installing prerequisites
+
+On Linux:
+
+```bash
+sudo apt install postgresql-9.6 postgresql-contrib-9.6
```
dankmitchell: A requirement of the project is to have Postgres installed so I don’t think we need to specify this step for Linux users, we aren’t telling OSX users how to install Postgres....

misaka: I think the issue @timabell ran into was that he didn't realize he needed `postgresql-contrib-*`, and the error that comes up doesn't make it obvious.

The other thing we don't make clear is that we use PostgreSQL 9.6 in production, so that's what we should have installed in dev. That's not necessary when using the Docker route, of course, but all this is moot if we only worry about that.

README.md
misaka: Added user creating/dropping as part of the usual rake tasks. @dankmitchell this should address your concerns about non-standard setup.

db/schema.rb
```diff
@@ -177,21 +178,19 @@
...
     t.index ["subject_code"], name: "AK_ucas_subject_subject_code", unique: true
     t.index ["subject_code"], name: "IX_subject_subject_code", unique: true
   end
 
-  create_table "temp_tbl", id: false, force: :cascade do |t|
-    t.text "code"
-  end
-
-  create_table "user", id: :integer, force: :cascade do |t|
+  create_table "user", id: :integer, default: -> { "nextval('mc_user_id_seq'::regclass)" }, force: :cascade do |t|
```
dankmitchell: Does this run for you @tim. Running `rake db:schema:load` failed for me with these default values.

misaka: I'll have a look at this, but I think @timabell may have pushed this to the wrong PR ... just checking what's up now.

README.md
dankmitchell: I’ll let this go but I’d like to fix up the reader afteer this is merged as we lost the non-docker setup and prerequisites when the docker stuff was merged it. I didn’t realise that

dankmitchell: Not your fault @tvararu

PR #54, opened 2019-01-23
Add recruitment cycle

MartinBentley:
### Context
`recruitment cycle` endpoints

`/api/v1/:recruitment_cycle_year/xxx` 
where `recruitment_cycle_year` is either 2019 or 2020
where `xxx` is either courses, or, providers, or subjects.

Add `/api/v1/:recruitment_cycle_year/xxx` 
For now 2020 endpoint will be returning http status code 404

### Guidance to review
https://github.com/DFE-Digital/manage-courses-backend/blob/master/docs/api.md#preparation-for-the-next-recruitment-cycle-rollover

These will be new endpoints added, existing `/api/v1/xxx`  will stay

### DIFF
--- a/app/controllers/api/v1/courses_controller.rb
+++ b/app/controllers/api/v1/courses_controller.rb
@@ -2,7 +2,16 @@ module Api
   module V1
     class CoursesController < ApplicationController
       def index
-        @courses = Course.all.includes(:sites, :provider, :site_statuses)
+        recruitment_cycle_year = params[:recruitment_cycle_year]
+        if recruitment_cycle_year.present?
+          if recruitment_cycle_year == "2019"
+            @courses = Course.all.includes(:sites, :provider, :site_statuses)
+          else
+            return render plain: "courses for year '#{recruitment_cycle_year}' not found", status: :not_found
+          end
+        else
+          @courses = Course.all.includes(:sites, :provider, :site_statuses)
+        end
         paginate json: @courses
       end
     end
--- a/app/controllers/api/v1/providers_controller.rb
+++ b/app/controllers/api/v1/providers_controller.rb
@@ -2,7 +2,16 @@ module Api
   module V1
     class ProvidersController < ApplicationController
       def index
-        @providers = Provider.all
+        recruitment_cycle_year = params[:recruitment_cycle_year]
+        if recruitment_cycle_year.present?
+          if recruitment_cycle_year == "2019"
+            @providers = Provider.all
+          else
+            return render plain: "providers for year '#{recruitment_cycle_year}' not found", status: :not_found
+          end
+        else
+          @providers = Provider.all
+        end
         paginate json: @providers
       end
     end
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -2,7 +2,16 @@ module Api
   module V1
     class SubjectsController < ApplicationController
       def index
-        @subjects = Subject.all
+        recruitment_cycle_year = params[:recruitment_cycle_year]
+        if recruitment_cycle_year.present?
+          if recruitment_cycle_year == "2019"
+            @subjects = Subject.all
+          else
+            return render plain: "subjects for year '#{recruitment_cycle_year}' not found", status: :not_found
+          end
+        else
+          @subjects = Subject.all
+        end
         paginate json: @subjects
       end
     end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,24 +1,28 @@
 # == Route Map
 #
-#           Prefix Verb   URI Pattern                     Controller#Action
-# api_v1_providers GET    /api/v1/providers(.:format)     api/v1/providers#index
-#                  POST   /api/v1/providers(.:format)     api/v1/providers#create
-#  api_v1_provider GET    /api/v1/providers/:id(.:format) api/v1/providers#show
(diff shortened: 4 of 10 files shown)


defong:
[Add recruitment_cycle parameter to the UCAS API](https://trello.com/c/HGWsgI55/873-add-recruitmentcycle-parameter-to-the-ucas-api)

app/controllers/api/v1/courses_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
         paginate json: @courses
       end
 
-      def show
+      def show_by_recruitment_cycle
         @courses = Course.all.includes(:sites, :provider, :site_statuses)
```
dankmitchell: This will just render all courses... dont we need to check the `params` the year and then do something like

`@courses = Course.where(recruitment_cycle: params[:year])`

app/controllers/api/v1/providers_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
         paginate json: @providers
       end
 
-      def show
+      def show_by_recruitment_cycle
```
dankmitchell: I think `recruitment_cycle` is enough

app/controllers/api/v1/courses_controller.rb
misaka: @dankmitchell I thought we didn't have this info in the db yet, so until we sort that out (which is a bigger piece of work), we'd just dump it all.

app/controllers/api/v1/courses_controller.rb
```diff
@@ -5,6 +5,11 @@ def index
...
         paginate json: @courses
       end
+
+      def show_by_recruitment_cycle
```
misaka: `show` should be used when retrieving a single `Course`. Also, I don't think we need a separate method for this, we should be able to simply point this endpoint to `index`, and once we actually have more than one recruitment year's worth of data, we'll sort out the filtering then.

config/routes.rb
```diff
@@ -1,31 +1,38 @@
...
 
 Rails.application.routes.draw do
   namespace :api do
     namespace :v1 do
       resources :providers
       resources :subjects
       resources :courses
+      get '/:year/courses', to: 'courses#show_by_recruitment_cycle', year: /(2019|2020)/
+      get '/:year/subjects', to: 'subjects#show_by_recruitment_cycle', year: /(2019|2020)/
+      get '/:year/providers', to: 'providers#show_by_recruitment_cycle', year: /(2019|2020)/
```
misaka: I say we just point this to the `index`. Once we actually care about the year we can either add filtering to `index` or point it to a different endpoint if we think it's worth it.

app/controllers/api/v1/courses_controller.rb
```diff
@@ -2,6 +2,20 @@ module Api
...
+          else
+            return render plain: "courses for year '#{recruitment_cycle_year}' not found", status: :not_found
+          end
+        else
+          @courses = Course.all.includes(:sites, :provider, :site_statuses)
+        end
+        paginate json: @courses
+      end
+
+      def show_by_recruitment_cycle
```
defong: `show_by_recruitment_cycle` that can be removed

app/controllers/api/v1/courses_controller.rb
defong: `show_by_recruitment_cycle` that can be removed

app/controllers/api/v1/providers_controller.rb
defong: `show_by_recruitment_cycle` that can be removed

app/controllers/api/v1/subjects_controller.rb
```diff
@@ -2,6 +2,20 @@ module Api
...
+          else
+            return render plain: "subjects for year '#{recruitment_cycle_year}' not found", status: :not_found
+          end
+        else
+          @subjects = Subject.all
+        end
+        paginate json: @subjects
+      end
+
+      def show_by_recruitment_cycle
```
defong: `show_by_recruitment_cycle` that can be removed

config/routes.rb
```diff
@@ -1,31 +1,40 @@
...
+#                  GET    /api/v1/:year/providers(.:format) api/v1/providers#show {:year=>/(2019|2020)/}
+
 
 Rails.application.routes.draw do
   namespace :api do
     namespace :v1 do
       resources :providers
       resources :subjects
       resources :courses
+      scope "/(:recruitment_cycle_year)", constraints: { recruitment_cycle_year: /(2018|2019|2020)/ } do
```
defong: how did the past sneak in? naughty history

spec/requests/api/v1/course_spec.rb
```diff
@@ -46,74 +46,103 @@
...
-    it "returns http unauthorised" do
-      get "/api/v1/courses", headers: { 'HTTP_AUTHORIZATION' => ActionController::HttpAuthentication::Token.encode_credentials("foo") }
-      expect(response).to have_http_status(:unauthorized)
+      it "JSON body response contains expected course attributes" do
+        get "/api/v1/courses", headers: { 'HTTP_AUTHORIZATION' => ActionController::HttpAuthentication::Token.encode_credentials("bats") }
+        json = JSON.parse(response.body)
+        expect(json). to eq expected_json
+      end
+
+      it "returns http success" do
```
defong: there should be two describe blocks one for index as the default and index#parameters as the other

PR #66, opened 2019-01-25
Providers incremental fetch

misaka:
https://trello.com/c/HMAlga4l/852-implement-incremental-fetch-for-providers

### Context

We're adding a mechanism to deliver to API consumers only those providers that have changed since a given date-time.

This change adds a method to retrieve any provider that has been updated, or whose enrichment or site, has been updated since a given date-time.

### Guidance to review

This PR is an incremental change on the path to delivering this ticket. The defined method isn't accessible yet, that will come in future PRs.


app/models/provider.rb
```diff
@@ -40,7 +40,20 @@ class Provider < ApplicationRecord
   has_and_belongs_to_many :organisations, join_table: :organisation_provider
 
   has_many :sites
-  has_many :enrichments, foreign_key: :provider_code, primary_key: :provider_code, class_name: "ProviderEnrichment"
+  has_many :enrichments,
+           foreign_key: :provider_code,
+           primary_key: :provider_code,
+           class_name: "ProviderEnrichment"
+
+  scope :changed_since, ->(datetime) do
```
defong: correctiveness of `change_since`

so provider_enrichment is it `published` or `draft`?

if its `draft` is it intended  to state the provider has changed? given the aspect that of `publish`

misaka: You're right @defong, we should only be picking up published changes. Good catch!

misaka: @defong fixed, we hope.

timabell:
[Implement incremental fetch for providers.](https://trello.com/c/HMAlga4l/852-implement-incremental-fetch-for-providers)

app/models/provider.rb
```diff
@@ -46,11 +46,16 @@ class Provider < ApplicationRecord
...
   scope :changed_since, ->(datetime) do
-    joins(:sites, :enrichments).where(<<~EOSQL, since: datetime)
-      provider.updated_at >= :since
-        OR site.updated_at >= :since
-        OR provider_enrichment.updated_at >= :since
-    EOSQL
+    joins(:sites, :enrichments).where(
+      <<~EOSQL,
+        provider.updated_at >= :since
+          OR site.updated_at >= :since
```
defong: ummm test for that one?

timabell: done

timabell:
We have more to do to finish the card, but I'd like to get the incremental improvement we have so far merged in. Re-review please!

spec/requests/api/v1/provider_spec.rb
```diff
@@ -194,7 +194,7 @@
       end
     end
     context 'with changed_since' do
-      it 'JSON body response contains expected provider attributes' do
```
misaka: A part of me wants this to be:


```ruby
context 'with changed_since' do
  describe 'JSON body response' do
    it 'contains expected providers' do
      ...
```

PR #71, opened 2019-01-29
Include recruitment_year into routes

dankmitchell:
### Context
`recruitment_cycle`

Add `recruitment_cycle` into route

### Guidance to review
`http://localhost:3000/api/v1/2019/courses`
`http://localhost:3000/api/v1/2019/providers`
`http://localhost:3000/api/v1/2019/subjects`

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -57,4 +57,8 @@ class Course < ApplicationRecord
   has_and_belongs_to_many :subjects
   has_many :site_statuses
   has_many :sites, through: :site_statuses
+
+  def recruitment_cycle
+    "2019"
+  end
 end
--- a/app/serializers/course_serializer.rb
+++ b/app/serializers/course_serializer.rb
@@ -57,8 +57,7 @@ class CourseSerializer < ActiveModel::Serializer
     "Y" # we want to always create PDFs for applications coming in
   end
 
-  # TODO: make recruitment cycle dynamic
   def recruitment_cycle
-    "2019"
+    object.recruitment_cycle
   end
 end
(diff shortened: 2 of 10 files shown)


dankmitchell:
[Add recruitment_cycle parameter to the UCAS API](https://trello.com/c/HGWsgI55/873-add-recruitmentcycle-parameter-to-the-ucas-api)

app/models/course.rb
```diff
@@ -57,4 +57,10 @@ class Course < ApplicationRecord
   has_and_belongs_to_many :subjects
   has_many :site_statuses
   has_many :sites, through: :site_statuses
+
+  def recruitment_year
+    "2019"
+  end
+
+  scope :recruitment_cycle, ->(year) { select { |c| c.recruitment_year == year } if year.present? }
```
misaka: tests pls

app/serializers/course_serializer.rb
```diff
@@ -57,8 +57,7 @@ def copy_form_required
     "Y" # we want to always create PDFs for applications coming in
   end
 
-  # TODO: make recruitment cycle dynamic
   def recruitment_cycle
-    "2019"
+    object.recruitment_year
```
misaka: Is there a reason not to keep these two the same name?

app/models/course.rb
```diff
@@ -57,4 +57,8 @@ class Course < ApplicationRecord
   has_and_belongs_to_many :subjects
   has_many :site_statuses
   has_many :sites, through: :site_statuses
+
+  def recruitment_cycle
+    "2019"
```
timabell: should this be an int rather than a string given it's a year?

dankmitchell: Docs say string and UCAS accepts a string

PR #72, opened 2019-01-29
Simplify changed_since query param

dankmitchell:
### Context
`changed_since` query param

Move logic from controller into model

### Guidance to review
No change


timabell:
seems okay, will it still include the enrichment & sites in the output without the explicit join?

PR #74, opened 2019-01-30
Amended config to default to singularise table names

defong:
### Context
Table names

singularise table names


defong:
[Set config for singularise table names](https://trello.com/c/CXSH1yJz/919-set-config-for-singularise-table-names)

PR #81, opened 2019-01-31
Implement changed_since-based pagination for providers

tvararu:
This brings the API in line with our documentation, by querying the database with the necessary ordering, limiting, and then adding a custom response header with the correctly formed URL.

Also enables focusing by using `fit`, `fdescribe`, `fcontext` in rspec; see the first commit.

Paired with @timabell.

### DIFF
--- a/app/controllers/api/v1/providers_controller.rb
+++ b/app/controllers/api/v1/providers_controller.rb
@@ -1,9 +1,33 @@
 module Api
   module V1
     class ProvidersController < ApplicationController
+      # Potential edge case:
+      # It is possible for older updated_at values to written to the database after this API has been queried for changes. This would mean that these changes are missed when the client makes a subsequent request using the next-link.
+      # Possible causes of older updated_at values:
+      # - delay between c# calculating datetime.UtcNow and value being written to postgres
+      # - clock drift between servers
       def index
-        @providers = Provider.changed_since(params[:changed_since])
-        paginate json: @providers
+        page_size = params[:per_page] || 100
+        ActiveRecord::Base.transaction do
+          ActiveRecord::Base.connection.execute('LOCK provider, provider_enrichment, site IN SHARE UPDATE EXCLUSIVE MODE')
+          @providers = Provider.changed_since(params[:changed_since]).limit(page_size)
+        end
+        last_provider = @providers.last
+
+        response.headers['Link'] = if last_provider
+                                     next_link((last_provider.updated_at + 1.second).utc.iso8601, last_provider.id, page_size)
+                                   else
+                                     next_link(params[:changed_since], "", page_size)
+                                   end
+
+        render json: @providers
+      end
+
+    private
+
+      def next_link(changed_since, from_provider_id, per_page)
+        current_url = request.base_url + request.path
+        "#{current_url}?changed_since=#{changed_since}&from_provider_id=#{from_provider_id}&per_page=#{per_page}; rel=\"next\""
       end
     end
   end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -45,16 +45,18 @@ class Provider < ApplicationRecord
 
   scope :changed_since, ->(datetime) do
     if datetime.present?
-      left_joins(:sites, :enrichments).where(
-        <<~EOSQL,
-          provider.updated_at >= :since
-            OR site.updated_at >= :since
-            OR (provider_enrichment.status = :status
-                AND provider_enrichment.updated_at >= :since)
-        EOSQL
-        since: datetime,
-        status: ProviderEnrichment.statuses['published']
-      )
+      left_joins(:sites, :enrichments)
+        .where(
+          <<~EOSQL,
+            provider.updated_at >= :since
+              OR site.updated_at >= :since
+              OR (provider_enrichment.status = :status
+                  AND provider_enrichment.updated_at >= :since)
+          EOSQL
+          since: datetime,
+          status: ProviderEnrichment.statuses['published']
+        )
+        .order(:updated_at, :id)
     end
   end
 
--- a/spec/factories/providers.rb
+++ b/spec/factories/providers.rb
@@ -37,11 +37,12 @@ FactoryBot.define do
     region_code { ProviderEnrichment.region_codes['London'] }
 
     transient do
-      age          { nil }
-      site_count   { 1 }
-      sites        { build_list :site, site_count, provider: nil, age: age }
-      course_count { 2 }
-      enrichments  { [build(:provider_enrichment, age: age)] }
+      age                  { nil }
+      skip_associated_data { false }
+      site_count           { 1 }
+      sites                { build_list :site, site_count, provider: nil, age: age }
+      course_count         { 2 }
+      enrichments          { [build(:provider_enrichment, age: age)] }
     end
 
     after(:build) do |provider, evaluator|
--- a/spec/requests/api/v1/provider_spec.rb
+++ b/spec/requests/api/v1/provider_spec.rb
@@ -1,5 +1,10 @@
 require 'rails_helper'
 
+def get_provider_codes_from_body(body)
+  json = JSON.parse(body)
+  json.map { |provider| provider["institution_code"] }
+end
+
 describe 'Providers API', type: :request do
   describe 'GET index' do
     let(:credentials) do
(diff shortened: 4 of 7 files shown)


spec/requests/api/v1/provider_spec.rb
```diff
@@ -213,15 +218,78 @@
...
+      end
+
+      context "with many providers" do
+        before do
+          101.times do |i|
+            create(:provider, provider_code: "PROV#{i + 1}", age: (200 - i).minutes.ago)
+          end
+        end
+
+        it 'pages properly' do
```
tvararu: This test is a bit rambly and we briefly experimented with breaking it up but I don't know how to daisy chain the `next_url`. 🤔

spec/requests/api/v1/provider_spec.rb
```diff
@@ -213,15 +218,78 @@
...
+          get '/api/v1/providers',
+            headers: { 'HTTP_AUTHORIZATION' => credentials },
+            params: { changed_since: 201.minutes.ago.utc.iso8601 }
+
+          returned_provider_codes = get_provider_codes_from_body(response.body)
+
+          100.times { |i| expect(returned_provider_codes).to include "PROV#{i + 1}" }
```
tvararu: This appears to actually be quite slow, performance-wise...

tvararu:
[Implement incremental fetch for providers.](https://trello.com/c/HMAlga4l/852-implement-incremental-fetch-for-providers)

app/models/provider.rb
```diff
@@ -45,16 +45,19 @@ class Provider < ApplicationRecord
...
+            provider.updated_at >= :since
+              OR site.updated_at >= :since
+              OR (provider_enrichment.status = :status
+                  AND provider_enrichment.updated_at >= :since)
+          EOSQL
+          since: datetime,
+          status: ProviderEnrichment.statuses['published']
+        )
+        .order(:updated_at, :id)
+        .limit(100)
```
misaka: This feels like the wrong place to `limit` ... it's the controller's concern that this is being used for paging so it should be the one to impose a `limit`.

spec/requests/api/v1/provider_spec.rb
misaka: It'd probably makes sense to do:

```ruby
expected_provider_codes = 1..100.map { |n| "PROV#{n}" }
expect(returned_provider_codes).to match_array expected_provider_codes
```

spec/requests/api/v1/provider_spec.rb
```diff
@@ -213,15 +218,78 @@
...
+          params: { changed_since: provided_timestamp }
+
+        expected = /#{request.base_url + request.path}\?changed_since=#{provided_timestamp}&from_provider_id=; rel="next"$/
+        expect(response.headers["Link"]).to match expected
+      end
+
+      context "with many providers" do
+        before do
+          101.times do |i|
+            create(:provider, provider_code: "PROV#{i + 1}", age: (200 - i).minutes.ago)
```
misaka: I would expect creating 101 providers would be quite slow ... I would suggest that we either:

1. set a `per_page` option to something smaller, like 10, and then rely on other tests to test that the default `per_page` is set to 100
2. if that feels too artificial, then at least stop the provider from creating associated data like `courses`, `sites` and `enrichments`.

spec/requests/api/v1/provider_spec.rb
misaka: I know what you mean, but until we setup some kind of end-to-end tests it may as well live here. I'm thinking it may be an idea to have end-to-end tests run as smoke tests, using cucumber and mechanize, and then the tests here could be more granular. Up for discussion later.

spec/spec_helper.rb
```diff
@@ -53,6 +53,9 @@
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
 
+  config.filter_run focus: true
+  config.run_all_when_everything_filtered = true
```
misaka: 👍 

Also, another idea. What about we tag all the tests as `courses`, `providers`, etc, as per the endpoint they relate to, so that we change something that changes, say, the `providers` endpoint, we can limit test runs to only those tests that are relevant?

Not a blocker, this would be a nice to have.

spec/requests/api/v1/provider_spec.rb
tvararu: This does not change the performance but nonetheless is a nicer way to assert.

spec/factories/providers.rb
```diff
@@ -52,6 +53,8 @@
     end
 
     after(:create) do |provider, evaluator|
+      next if evaluator.skip_associated_data
```
misaka: `after` blocks aren't _just_ for creating associated data ... what if we just added `sites: [], enrichments: []`, etc, to the `create(:provider, ...)` below? Also, less machinery to add here.

spec/requests/api/v1/provider_spec.rb
```diff
@@ -256,7 +256,7 @@ def get_provider_codes_from_body(body)
       context "with many providers" do
         before do
           101.times do |i|
-            create(:provider, provider_code: "PROV#{i + 1}", age: (200 - i).minutes.ago)
+            create(:provider, provider_code: "PROV#{i + 1}", age: (200 - i).minutes.ago, skip_associated_data: true)
```
misaka: I still like the idea of separating out the test concerns:

1. a request test that tests that paging works with a `per_page` set to `10`
2. a unit test that only tests that `per_page` defaults to 100, perhaps using mocking if required

spec/spec_helper.rb
misaka: I'm coming from a place where it took 15-20 mins to run the entire test suite. We had a pretty clear separation of concerns we could tag the tests with, but the problem was that by the time we got there, it was hard to find the motivation to back-fill the tags.

misaka: I'm happy to not push on this for now, btw.

spec/factories/providers.rb
tvararu: Done

spec/requests/api/v1/provider_spec.rb
tvararu: ✅

app/controllers/api/v1/providers_controller.rb
```diff
@@ -2,8 +2,24 @@ module Api
   module V1
     class ProvidersController < ApplicationController
       def index
-        @providers = Provider.changed_since(params[:changed_since])
-        paginate json: @providers
+        page_size = params[:per_page] || 100
+        @providers = Provider.changed_since(params[:changed_since]).limit(page_size)
+        last_provider = @providers.last
+
+        response.headers['Link'] = if last_provider
```
misaka: I remembered something over the weekend. There's an edge case to be aware of, and that we might want to code for because it's likely to be a subtle and hard-to-diagnose bug. If there's a transaction in-flight when this request fires off, I think there's a chance that we will miss it here unless we acquire a write-lock or find some other way of mitigating it.

Let's just have a quick chat about this in-person, I'm not 100% sure how to put it down here.

tvararu: We've discussed this in person and have resolved to:

- Get a write-lock before performing this database fetch
- Document the transaction edge case we're dealing with here through a code comment, for posterity

timabell:
> There is a transaction for edge cases,
> but it doesn't cover all possible spaces,
> locks we obtain,
> not all quite in vain,
> So beware of the timestamps c# makeses

tvararu:
@misaka updated with db lock and helpful words

app/controllers/api/v1/providers_controller.rb
```diff
@@ -1,9 +1,33 @@
 module Api
   module V1
     class ProvidersController < ApplicationController
+      # Potential edge case:
+      # It is possible for older updated_at values to written to the database after this API has been queried for changes. This would mean that these changes are missed when the client makes a subsequent request using the next-link.
+      # Possible causes of older updated_at values:
+      # - delay between c# calculating datetime.UtcNow and value being written to postgres
+      # - clock drift between servers
```
misaka: LGTM

app/controllers/api/v1/providers_controller.rb
```diff
@@ -1,9 +1,33 @@
...
+                                   end
+
+        render json: @providers
+      end
+
+    private
+
+      def next_link(changed_since, from_provider_id, per_page)
+        current_url = request.base_url + request.path
+        "#{current_url}?changed_since=#{changed_since}&from_provider_id=#{from_provider_id}&per_page=#{per_page}; rel=\"next\""
```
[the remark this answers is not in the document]
tvararu: I tried this but couldn't figure out how to get rid of the escaping, tried `.undump` but 🤷‍♂️

PR #83, opened 2019-02-01
Amended docker to use alpine

defong:
### Context
Given that I'm the only human being and CI to use docker, as there is no chance of either of us installing ruby on our machine native

Those 1+ gb of ruby related rubbishness becomes a bigger eyesore. 

Improve ruby rubbish experience on docker

### Guidance to review
`docker-compose build`
Before

| repo        | tag           | mem  |
| ------------- |:-------------:| -----:|
| batdevcontainerregistry.azurecr.io/manage-courses-backend      | latest |   1.01GB| |
| ruby      | 2.5.3      |   871MB |

After

| repo        | tag           | mem  |
| ------------- |:-------------:| -----:|
| batdevcontainerregistry.azurecr.io/manage-courses-backend      | latest |   350MB |
| ruby      | 2.5.3-alpine      |   43.6MB |


defong:
@tvararu 

https://github.com/ettore26/wait-for-command

replacement for `wait_for_it .sh` as that is `bash`, 

alpine and me dislike `bash`, and would like to take the `b` of so it runs on `ash`.

You good with that?

Dockerfile
```diff
@@ -1,9 +1,11 @@
-FROM ruby:2.5.3
+FROM ruby:2.5.3-alpine
 
 ENV APP_HOME /app
 RUN mkdir $APP_HOME
 WORKDIR $APP_HOME
 
+RUN apk add --update build-base postgresql-dev tzdata
```
misaka: I think it's good practice to remove the packages once they're not needed, for example:

https://github.com/coderjoe/dockerfile-alpine-rails/blob/master/alpine-rails/Dockerfile

defong: hmmm... not too sure what you meant, 


pair these up with the gem `build-base` `postgresql-dev` `tzdata`?

or this `rm -rf /var/cache/apk/*`

?

tvararu:
@defong 👍

Dockerfile
misaka: Sorry, that linked-to Dockerfile isn't the best example.

Yes, it's partly about the cache, but also the unused packages that are installed. For example, when building this Docker container we see `g++` et al being installed:

```
(12/26) Installing gcc (6.4.0-r9)
(15/26) Installing g++ (6.4.0-r9)
(16/26) Installing make (4.2.1-r2)
```

these aren't needed at runtime so they should also be removed. Here's a better reference that has a sample `Dockerfile`:

https://www.chrisblunt.com/rails-on-docker-getting-started-docker-ruby-rails/

PR #85, opened 2019-02-05
Fixed duplicated provider that was exhibited in json response

defong:
### Context
Duplicated provider that was exhibited in json response

Fixed duplicated provider that was exhibited in json response

### DIFF
--- a/spec/requests/api/v1/provider_spec.rb
+++ b/spec/requests/api/v1/provider_spec.rb
@@ -290,6 +290,54 @@ describe 'Providers API', type: :request do
           expect(returned_provider_codes.size).to eq 0
         end
       end
+
+      context 'a single provider with multiple published enrichments' do
+        let!(:provider) do
+          create(:provider,
+                site_count: 0,
+                updated_at: 5.days.ago,
+                enrichments: [new_published_enrichment, old_published_enrichment])
+        end
+        let!(:site1) do
+          create(:site,
+                updated_at: 4.days.ago,
+                provider: provider)
+        end
+        let!(:site2) do
+          create(:site,
+                updated_at: 3.days.ago,
+                provider: provider)
+        end
+        let(:new_published_enrichment) do
+          build(:provider_enrichment,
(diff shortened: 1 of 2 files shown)


defong:
[Duplicate providers in the incremental Provider API](https://trello.com/c/Nn9OGiBP/938-duplicate-providers-in-the-incremental-provider-api)

benilovj:
The spec is very verbose. One of the major benefits of `factory_bot` is to be able to skip the parts of the test data that you don't care about and let `factory_bot` provide sensible defaults. Is there a way to strip the spec down to the actual attributes that are relevant for the particular scenario?

defong:
Yeah the factory can set most of it, I was verbose, using data from endpoint.

defong:
> > Yeah the factory can set most of it, I was verbose, using data from endpoint.
> 
> OK, could we strip the spec back to just the things that trigger the bug, so the clarity isn't lost? smiley

Sorted, as far as I can, not sure of the https://github.com/DFE-Digital/manage-courses-backend/blob/819c6fa18f59658e980f10b8cdf0d323ce610b70/spec/factories/providers.rb#L40

`age` is passed around its looks like a vintage singleton value, so sticking with hard coded dates.

spec/requests/api/v1/provider_spec.rb
```diff
@@ -290,6 +290,62 @@ def get_provider_codes_from_body(body)
...
+                postcode: 'enrichment1 postcode',
+                region_code: 6,
+                created_at: "2019-02-05T10:32:14.456414",
+                updated_at: "2019-02-05T10:32:22.95923",
+                status: 1)
+        end
+        let(:old_published_enrichment) do
+          build(:provider_enrichment,
+                region_code: 6,
+                created_at: "2018-09-26T15:25:33.191305",
```
dankmitchell: Might be easier to read using a date object i.e. `Date.yesterday`

defong: I was using the exact dates that was showing the bug.

PR #91, opened 2019-02-07
Rescue changed since param

dankmitchell:
### Context
`changed_since` query parameter

- Render `400` when `changed_since` is invalid

### Guidance to review
`http://localhost:3000/api/v1/2019/providers?changed_since=2019`

### DIFF
--- a/app/controllers/api/v1/providers_controller.rb
+++ b/app/controllers/api/v1/providers_controller.rb
@@ -22,7 +22,7 @@ module Api
 
         render json: @providers
       rescue ActiveRecord::StatementInvalid
-        render json: { status: 400, message: 'Invalid changed_since value, the format should be a iso8601 timestamp' }.to_json, status: 400
+        render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: 400
       end
 
     private
--- a/app/controllers/api/v1/providers_controller.rb
+++ b/app/controllers/api/v1/providers_controller.rb
@@ -21,6 +21,8 @@ module Api
                                    end
 
         render json: @providers
+      rescue ActiveRecord::StatementInvalid
+        render json: { status: 400, message: 'Invalid changed_since value, the format should be a iso8601 timestamp' }.to_json, status: 400
       end
 
     private
--- a/spec/controllers/api/v1/providers_controller_spec.rb
+++ b/spec/controllers/api/v1/providers_controller_spec.rb
@@ -20,5 +20,16 @@ RSpec.describe Api::V1::ProvidersController, type: :controller do
 
       get :index
     end
+
+    it 'renders a 400 when the changed_since param is not valid' do
+      allow(controller).to receive(:authenticate)
+
+      get :index, params: { changed_since: '2019' }
+      expect(response).to have_http_status(:bad_request)
+      json = JSON.parse(response.body)
+      expect(json). to eq(
+        'status' => 400, 'message' => 'Invalid changed_since value, the format should be a iso8601 timestamp'
+      )
+    end
   end
 end
--- a/spec/controllers/api/v1/providers_controller_spec.rb
+++ b/spec/controllers/api/v1/providers_controller_spec.rb
@@ -27,8 +27,8 @@ RSpec.describe Api::V1::ProvidersController, type: :controller do
       get :index, params: { changed_since: '2019' }
       expect(response).to have_http_status(:bad_request)
       json = JSON.parse(response.body)
-      expect(json). to eq(
-        'status' => 400, 'message' => 'Invalid changed_since value, the format should be a iso8601 timestamp'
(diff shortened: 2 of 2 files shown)


benilovj:
I'd have a preference for `Date.parse`ing (or similar) the param before passing it through, and bombing out early on in the action with a 400. That would avoid nesting...

dankmitchell:
> how do they request from scratch if the param is required?

@timabell that's a question for @benilovj. The story I picked up said this was a requirement

app/controllers/api/v1/providers_controller.rb
```diff
@@ -7,20 +7,24 @@ class ProvidersController < ApplicationController
...
+          response.headers['Link'] = if last_provider
+                                       next_link((last_provider.updated_at + 1.second).utc.iso8601, last_provider.id, page_size)
+                                     else
+                                       next_link(params[:changed_since], "", page_size)
+                                     end
 
-        render json: @providers
+          render json: @providers
+        rescue ActiveRecord::StatementInvalid => e
+          render json: { code: 400, status: 'Bad Request', message: 'Invalid changed_since value, the format should be a iso8601 timestamp' }.to_json, status: 400
```
defong: ` render json: { code: 400, status: 'Bad Request', message: 'Invalid changed_since value, the format should be a iso8601 timestamp' }.to_json, status: :bad_request`



--

` status: :bad_request` vs  `status: 400`

app/controllers/api/v1/providers_controller.rb
```diff
@@ -7,20 +7,24 @@ class ProvidersController < ApplicationController
       # - delay between c# calculating datetime.UtcNow and value being written to postgres
       # - clock drift between servers
       def index
-        page_size = params[:per_page] || 100
-        ActiveRecord::Base.transaction do
-          ActiveRecord::Base.connection.execute('LOCK provider, provider_enrichment, site IN SHARE UPDATE EXCLUSIVE MODE')
-          @providers = Provider.changed_since(params[:changed_since]).limit(page_size)
-        end
-        last_provider = @providers.last
+        begin
```
misaka: Don't need a `begin` block here, and personally I'm not sure it adds much.

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -20,5 +20,16 @@
...
     end
+
+    it 'renders a 400 when the changed_since param is not valid' do
+      allow(controller).to receive(:authenticate)
+
+      get :index, params: { changed_since: '2019' }
+      expect(response).to have_http_status(:bad_request)
+      json = JSON.parse(response.body)
+      expect(json). to eq(
+        'status' => 400, 'message' => 'Invalid changed_since value, the format should be a iso8601 timestamp'
```
tvararu: I don't think having the exact message letter for letter in the test is important, it couples the implementation unnecessarily, but it's a very minor thing

benilovj:
I still think something like this would be more explicit, if a bit more verbose:

```
changed_since = DateTime.iso8601(params[:changed_since]) rescue :invalid
if changed_since == :invalid
  render json: { code: 400, status: 'Bad Request', message: 'Invalid changed_since value, the format should be a iso8601 timestamp' }.to_json, status: 400
  return
end
```

app/controllers/api/v1/providers_controller.rb
```diff
@@ -21,6 +21,8 @@ def index
                                    end
 
         render json: @providers
+      rescue ActiveRecord::StatementInvalid
+        render json: { status: 400, message: 'Invalid changed_since value, the format should be a iso8601 timestamp' }.to_json, status: 400
```
tvararu:
```suggestion
        render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: 400
```

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -20,5 +20,16 @@
 
       get :index
     end
+
+    it 'renders a 400 when the changed_since param is not valid' do
+      allow(controller).to receive(:authenticate)
+
+      get :index, params: { changed_since: '2019' }
```
benilovj: [This post](https://www.myintervals.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/) reckons that `2019` is a valid ISO8601 date...

benilovj: Perhaps rather than tying ourselves up in knots, we should specify a more strict interface...

tvararu: It is a valid ISO8601 date, it is not a valid ISO8601 UTC datetime timestamp. This stuff is very pedantic. I think supplying a good example is enough to disambiguate.

benilovj: Fair enough...

PR #92, opened 2019-02-07
Editorconfig from The GDS Way

timabell:
https://gds-way.cloudapps.digital/manuals/programming-languages.html#code-style-guides


timabell:
@dankmitchell good call. Done.

PR #93, opened 2019-02-08
[950] Courses summary spike [WIP]

benilovj:
### Context
A spike implementation of https://manage-courses-prototype.herokuapp.com/history/status-logic-before-transition


fofr:
In https://manage-courses-prototype.herokuapp.com/quiz:
We also have "Opens on [date]" for applications.

defong:
looks about right to me, from data point of view

dankmitchell:
Looks good to me. It would be nice to clearly define the v1 and v2 serialisers but the logic is good and very clear 😍

app/controllers/api/v2/courses_controller.rb
```diff
@@ -0,0 +1,12 @@
+module Api
+  module V2
+    class CoursesController < ApplicationController
+      skip_before_action :authenticate
+
+      def index
```
dankmitchell: Should this be a `provider#show` instead?

benilovj:
> Looks good to me. It would be nice to clearly define the v1 and v2 serialisers but the logic is good and very clear 😍

We could move the existing v1 serialisers into a `UCAS` namespace (since the representation is quite UCAS-specific). That way we could reclaim the default serializer names.

dankmitchell:
> We could move the existing v1 serialisers into a `UCAS` namespace (since the representation is quite UCAS-specific). That way we could reclaim the default serializer names.

Great idea @benilovj. I'm onboard with that

dankmitchell:
[Create api/v2 for manage courses frontend](https://trello.com/c/ChXP2RQm/950-create-api-v2-for-manage-courses-frontend)

defong:
> In https://manage-courses-prototype.herokuapp.com/quiz:
> We also have "Opens on [date]" for applications.

Covered in this commit
https://github.com/DFE-Digital/manage-courses-backend/pull/93/commits/7b82d80ea6b501b7b3c5dd14502503310b94d972

Basically `Applications` is a ui concern of which backend needs to provide enough to determine what to exactly show ie `Open`|`Closed`|`Date`|`Vacancy`

benilovj:
Closing, will reopen with better branch name and cleaner PR.

benilovj:
Reopened as #118.

PR #94, opened 2019-02-08
Added v2 end points

defong:
### Context
v2 end points

Added v2 end points

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,25 +1,49 @@
 # == Route Map
 #
-#           Prefix Verb   URI Pattern                                         Controller#Action
-# api_v1_providers GET    /api/v1(/:recruitment_year)/providers(.:format)     api/v1/providers#index
-#                  POST   /api/v1(/:recruitment_year)/providers(.:format)     api/v1/providers#create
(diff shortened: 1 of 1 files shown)


defong:
[Create api/v2 for manage courses frontend](https://trello.com/c/ChXP2RQm/950-create-api-v2-for-manage-courses-frontend)

config/routes.rb
```diff
@@ -30,6 +54,12 @@
         resources :courses
       end
     end
+    namespace :v2 do
+      resources :providers
+      resources :subjects
+      resources :courses
+      resources :organisations
+    end
```
misaka: These don't have controllers (yet). That means if someone hits these endpoints they get a 500 not a 404. I think we should only add these routes in when we have a controller to respond to it.

defong: I was getting distracted with this...
https://github.com/rails-api/active_model_serializers/issues/144

regards to 404 vs 500

image

misaka: Ah, good spot. I still wonder about adding the routes first without having controllers for them, but at least it doesn't 500 then.

defong:
refocusing on depth vs breadth

PR #96, opened 2019-02-08
[850] Implement incremental fetching on CourseController index

tvararu:
### Context

This allows users to only fetch a specific set of courses from a passed in changed_since time.

- Pull in database changes based on the C# API changes
- Implement changed_since scope on the course model
- Implement changed_since param and Link header on course controller similarly to the one in the provider controller; Link header logic is abstracted into a concern

### Guidance to review

The tests are almost a complete carbon copy of the tests from the provider controller. This isn't DRY, but can be made if so requested, it just feels like the correct abstraction for this didn't naturally appear while writing them. Suggestions welcome.

The `rescue` call in the controller is ported from a previous PR by Dan, and is also not DRY. Suggestions welcome once more.

No docs changes were needed for this.

This entire PR will be rebased and squashed / reorganised after the PR stage. The commits should be considered WIP.

This was paired on with @timabell and @davidgisbey.

### DIFF
--- a/app/controllers/api/v1/courses_controller.rb
+++ b/app/controllers/api/v1/courses_controller.rb
@@ -20,6 +20,8 @@ module Api
         next_link_header("from_course_id", @courses.last, next_course, changed_since, per_page)
 
         render json: @courses
+      rescue ActiveRecord::StatementInvalid
+        render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: 400
       end
     end
   end
--- a/app/controllers/api/v1/courses_controller.rb
+++ b/app/controllers/api/v1/courses_controller.rb
@@ -1,9 +1,25 @@
 module Api
   module V1
     class CoursesController < ApplicationController
+      include NextLinkHeader
+      include FirstItemFromNextPage
+
       def index
-        @courses = Course.includes(:sites, :provider, :site_statuses, :subjects)
-        paginate json: @courses
+        per_page = (params[:per_page] || 100).to_i
+        changed_since = params[:changed_since]
+        from_course_id = params[:from_course_id].to_i
+
+        @courses = Course
+          .includes(:sites, :provider, :site_statuses, :subjects)
+          .changed_since(changed_since, from_course_id)
+
+        next_course = first_item_from_next_page(@courses, per_page)
+
+        @courses = @courses.limit(per_page)
+
+        next_link_header("from_course_id", @courses.last, next_course, changed_since, per_page)
+
+        render json: @courses
       end
     end
   end
--- a/app/controllers/api/v1/providers_controller.rb
+++ b/app/controllers/api/v1/providers_controller.rb
@@ -1,23 +1,28 @@
 module Api
   module V1
     class ProvidersController < ApplicationController
+      include NextLinkHeader
+
       # Potential edge case:
       # It is possible for older updated_at values to written to the database after this API has been queried for changes. This would mean that these changes are missed when the client makes a subsequent request using the next-link.
       # Possible causes of older updated_at values:
       # - delay between c# calculating datetime.UtcNow and value being written to postgres
       # - clock drift between servers
       def index
-        page_size = params[:per_page] || 100
+        per_page = params[:per_page] || 100
+        changed_since = params[:changed_since]
+
         ActiveRecord::Base.transaction do
           ActiveRecord::Base.connection.execute('LOCK provider, provider_enrichment, site IN SHARE UPDATE EXCLUSIVE MODE')
-          @providers = Provider.changed_since(params[:changed_since]).limit(page_size)
+          @providers = Provider.changed_since(changed_since).limit(per_page)
         end
+
         last_provider = @providers.last
 
         response.headers['Link'] = if last_provider
-                                     next_link((last_provider.last_published_at + 1.second).utc.iso8601, last_provider.id, page_size)
+                                     next_link((last_provider.last_published_at + 1.second).utc.iso8601, last_provider.id, per_page)
                                    else
-                                     next_link(params[:changed_since], "", page_size)
+                                     next_link(params[:changed_since], "", per_page)
                                    end
 
         render json: @providers
--- a/app/controllers/concerns/first_item_from_next_page.rb
+++ b/app/controllers/concerns/first_item_from_next_page.rb
@@ -1,0 +1,12 @@
+module FirstItemFromNextPage
+  extend ActiveSupport::Concern
+
+private
+
+  def first_item_from_next_page(collection, per_page)
+    has_next_page_available = collection.size > per_page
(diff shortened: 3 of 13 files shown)


tvararu:
WIP added so I don't forget to rebase and re-do the commits.

app/controllers/api/v1/courses_controller.rb
```diff
@@ -1,9 +1,19 @@
...
-        paginate json: @courses
+        per_page = params[:per_page] || 100
+        changed_since = params[:changed_since]
+
+        @courses = Course.includes(:sites, :provider, :site_statuses, :subjects).changed_since(changed_since).limit(per_page)
+
+        next_link_header("from_course_id", @courses.last, changed_since, per_page)
+
+        render json: @courses
+      rescue ActiveRecord::StatementInvalid
```
tvararu: @misaka is there a clean way to DRY this out, since it's also present in the provider controller?

tvararu:
[Implement incremental fetch for courses. - Rails](https://trello.com/c/DRtQR8bN/850-implement-incremental-fetch-for-courses-rails)

app/controllers/api/v1/courses_controller.rb
benilovj: Not sure there's much benefit to DRYing out two instances of a one-liner...

tvararu: Same, I agree with the [three strikes rule](http://wiki.c2.com/?ThreeStrikesAndYouRefactor), but there will be another I think... more thinking ahead.

app/controllers/concerns/next_link_header.rb
```diff
@@ -0,0 +1,18 @@
+module NextLinkHeader
+  extend ActiveSupport::Concern
+
+private
+
+  def next_link_header(from_key_string, last_model, changed_since, per_page)
+    response.headers['Link'] = if last_model
+                                 header_content(from_key_string, (last_model.updated_at + 1.second).utc.iso8601, last_model.id, per_page)
```
defong: whats the purpose of adding one second?

app/controllers/api/v1/courses_controller.rb
misaka: You could put it into a `before_action` hook. It might make sense to pull this param out and store the changed_since value in `@changed_since`, using `nil` as a signal that it's not set. Similar to how we might pull out an `id` value. Also, it would be less ambiguous than catching any and all `ActiveRecord::StatementInvalid` exceptions.

I'd wait to see what comes out of @dankmitchell's work to look at the parsing of `changed_since` before DRYing this up, though.

app/controllers/api/v1/providers_controller.rb
```diff
@@ -1,36 +1,28 @@
...
+          @providers = Provider.changed_since(changed_since).limit(per_page)
         end
-        last_provider = @providers.last
 
-        response.headers['Link'] = if last_provider
-                                     next_link((last_provider.updated_at + 1.second).utc.iso8601, last_provider.id, page_size)
-                                   else
-                                     next_link(params[:changed_since], "", page_size)
-                                   end
+        next_link_header("from_provider_id", @providers.last, changed_since, per_page)
```
misaka: :100:

app/controllers/concerns/next_link_header.rb
misaka: The idea is that we generate a new `changed_since` timestamp that is one second after the last object that is being returned as this page.

Having said that, I think that there is a bug here. I say we evaluate this PR independent of this aspect and address it later.

app/controllers/api/v1/courses_controller.rb
tvararu: Sensible suggestion 👍

app/controllers/concerns/next_link_header.rb
tvararu: What is the bug?

app/controllers/api/v1/providers_controller.rb
tvararu: @davidgisbey and I were also really pleased with how clean this turned out :)

app/controllers/concerns/next_link_header.rb
misaka: If the last object in our list is in the middle of a result set, i.e. there is another page of results, then it shouldn't be the time of update of the last object on the current page + 1 second, it should be the time of update of the next object on the next page. I think this is hard to explain in words, this is part of what Tim was alluding to with his comment about the design, but I think it can be simply illustrated.

app/controllers/api/v1/courses_controller.rb
dankmitchell: I'll revisit this once this PR is over the line.

app/controllers/concerns/next_link_header.rb
tvararu: Let me see if I understand correctly:

```
a { changed_since: 13:00:00 }
b { changed_since: 13:00:04 }
...
f { changed_since: 13:03:23 }
<page-break>
g { changed_since: 13:05:22 }
...
```

In our current implementation, if we return a page containing `a, b, ..., f`, the `changed_since` will be `f.changed_since + 1.second`, so `13:03:24`.

But you're saying the next item is at `13:05:22`, so that's what we should return. Why? It shouldn't change the functionality; all the items are sorted by their `changed_since` so querying from `13:03:24` and `13:05:22` should always return the same set of items.

misaka: Yes.

You say "VERY tiny bug", but in fact one of the things I thought we should protect for is a batch publishing of providers that all happen within the same second ... I'm not sure how often that'll happen but it doesn't sound too outlandish to me.

tvararu: Correct. :)

Let's fix it. 👍

tvararu: We've added functionality to the `changed_since` scope in order to address this issue, and we should backport this fix to Providers.

tvararu:
@misaka we've addressed the issue when querying for pages that end/start with the same timestamp, and have come up with a `concern` that does it in an alright way. Could you give the course model another look? If you're happy, we can port that to the `Provider` model and controller.

timabell:
Don't merge till https://github.com/DFE-Digital/manage-courses-api/pull/265 has been shipped to prod

app/controllers/concerns/first_item_from_next_page.rb
```diff
@@ -0,0 +1,12 @@
...
+
+private
+
+  def first_item_from_next_page(collection, per_page)
+    has_next_page_available = collection.size > per_page
+    if has_next_page_available
+      collection[per_page]
+    end
+  end
+end
```
misaka: Would it be worth merging `FirstItemFromNextPage` and `NextLinkHeader`? Maybe call it `ChangedSincePagination`? One-method per concern just feels a bit fine-grained to me, unless you're following some best-practice advice I haven't seen?

PR #101, opened 2019-02-11
953 change changed_since scope to use last_published_at

timabell:
#### Why are we doing this?

To ensure providers are exported whenever they, or any entities that indicate the the provider info has changed.

#### Background

To support the incremental results in the API providers endpoint we want to update the last_published_at column on the provider when a related entity, site or provider enrichments for example. This will be needed in the future once the Rails backend takes on the work of updating the providers and their related entities.

The C# has already been updated to populate this data. https://github.com/DFE-Digital/manage-courses-api/pull/267

* Switch to `provider.last_published_at` instead of `updated_at`
* Schema has been manually modified to match changes from .NET app.

# ~~do not merge~~

~~Until the c# schema change has shipped to prod https://github.com/DFE-Digital/manage-courses-api/pull/267~~

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -45,8 +45,10 @@ class Provider < ApplicationRecord
 
   scope :changed_since, ->(datetime) do
     if datetime.present?
-      where("updated_at >= ?", datetime).order(:updated_at, :id)
-    end
+      where("last_published_at >= ?", datetime)
+    else
+      where("last_published_at is not null")
+    end.order(:last_published_at, :id)
   end
 
   # TODO: filter to published enrichments, maybe rename to published_address_info
--- a/spec/controllers/api/v1/providers_controller_spec.rb
+++ b/spec/controllers/api/v1/providers_controller_spec.rb
@@ -33,7 +33,7 @@ RSpec.describe Api::V1::ProvidersController, type: :controller do
     end
 
     describe 'generated next link' do
-      let!(:oldy_provider) { create(:provider, last_published_at: 5.minute.ago.utc) }
+      let!(:old_provider) { create(:provider, last_published_at: 5.minute.ago.utc) }
       let!(:last_provider) { create(:provider, last_published_at: 1.minute.ago.utc) }
       let(:last_provider_id) { last_provider.id }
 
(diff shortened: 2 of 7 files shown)


timabell:
https://github.com/DFE-Digital/manage-courses-api/pull/267 has shipped so good to go

app/models/provider.rb
```diff
@@ -45,8 +45,10 @@ class Provider < ApplicationRecord
 
   scope :changed_since, ->(datetime) do
     if datetime.present?
-      where("updated_at >= ?", datetime).order(:updated_at, :id)
-    end
+      where("last_published_at >= ?", datetime)
+    else
+      where("last_published_at is not null")
```
dankmitchell: This feels odd but I assume you have a valid reason it.

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -31,5 +31,22 @@
         'status' => 400
       )
     end
+
+    describe 'generated next link' do
+      let!(:oldy_provider) { create(:provider, last_published_at: 5.minute.ago.utc) }
```
dankmitchell: Is this a typo or deliberate? `oldy_provider `

app/models/provider.rb
timabell: the `is not null` is there to prevent unpublished providers escaping into the ucas system prematurely, there's probably a nicer syntax or more DRY way of doing this, any ideas?

spec/controllers/api/v1/providers_controller_spec.rb
timabell: it was to make everything line up. too kitsch?

timabell:
rebased

spec/controllers/api/v1/providers_controller_spec.rb
timabell: renamed

app/models/provider.rb
timabell: went round the houses trying to optimise this. do later.

benilovj: this can be done as `where.not(last_published_at: nil)`

PR #102, opened 2019-02-11
960 Add models needed for users

dankmitchell:
### Context
We need to expose `users` for manage-courses-frontend

Add all the scaffolding needed


misaka:
Could we have a chat about how we're going to do all this? 

Although I agree with the overall implementation being added here, they're very standard Rails, I just wonder if we should really be adding All The API-Endpoints like this before we have a discussion of how we know how we're going to use them.

dankmitchell:
@misaka always happy to chat, FYI I’m striped this back to just `users` index and show only with the end goal being to render the user providers on the frontend. Still WIP at the moment

app/models/organisation.rb
```diff
@@ -0,0 +1,15 @@
...
+#  id     :integer          not null, primary key
+#  name   :text
+#  org_id :text
+#
+
+class Organisation < ApplicationRecord
+  has_many :organisation_users
+  has_many :users, through: :organisation_users
+  has_many :organisation_providers
+  has_many :providers, through: :organisation_providers
```
tvararu: Probably dumb question, but can't this be stripped to:

```ruby
	  has_many :users, through: :organisation_users
	  has_many :providers, through: :organisation_providers
```

Do we need the has_many: join model ones?

misaka: I think for the `has_many ... through` syntax you need an existing association, but here I think what we really want is a [`has_and_belongs_to_many` association](https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_and_belongs_to_many). So far as I can tell `organisation_user` and `organisation_provider` are simple join-tables.

tvararu: Yeah, I was thinking there is a terser way to do it, HABTM is probably what we want :)

app/models/organisation_provider.rb
```diff
@@ -0,0 +1,13 @@
+# == Schema Information
+#
+# Table name: organisation_provider
+#
+#  id              :integer          not null, primary key
+#  provider_id     :integer
+#  organisation_id :integer
+#
+
+class OrganisationProvider < ApplicationRecord
```
misaka: If this is just a join table I don't think you need or want this model, it's just noise.

app/models/organisation_user.rb
```diff
@@ -0,0 +1,13 @@
+# == Schema Information
+#
+# Table name: organisation_user
+#
+#  id              :integer          not null, primary key
+#  organisation_id :integer
+#  user_id         :integer
+#
+
+class OrganisationUser < ApplicationRecord
```
misaka: If this is just a join table I don't think you need or want this model, it's just noise.

spec/models/user_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+#  accept_terms_date_utc  :datetime
+#
+
+require 'rails_helper'
+
+RSpec.describe User, type: :model do
+  subject { create(:user) }
+
+  describe 'associations' do
+    it { should have_and_belong_to_many(:organisations) }
```
misaka: While you're here, may as well add:

```
it { should have_many(providers).through(:organisations) }
```

app/models/organisation.rb
```diff
@@ -0,0 +1,13 @@
...
+# Table name: organisation
+#
+#  id     :integer          not null, primary key
+#  name   :text
+#  org_id :text
+#
+
+class Organisation < ApplicationRecord
+  has_and_belongs_to_many :users
+  has_and_belongs_to_many :providers
```
misaka: Odd that there is only one unit test added for the `User`. I'd add tests for these since they're one-liners.

PR #106, opened 2019-02-12
[961] add migration file to add tables from dotnet [DO NOT MERGE]

misaka:
### Context

We are starting to add functionality that will only exist on the Rails side, so it doesn't make sense to be dependent of the .NET app to make DB modifications.

Add a Rails migration file that creates the currently existing DB structure.

### Guidance to review

This migration is a spike and is **not ready to merge!** This is just an idea of what the First Migration might look like if we take this approach of putting it into the Rails app. **We will have to regenerate it with a proper schema.rb!**

**What to review:** The general approach taken with `unless table_exists?` which are there to protect the deployed environments, comments as to whether that's a Good Idea and/or The Right Way are welcome.

### People who know about this

@misha, @timabell2

### DIFF
--- a/db/migrate/20190212161359_add_existing_tables_from_dotnet.rb
+++ b/db/migrate/20190212161359_add_existing_tables_from_dotnet.rb
@@ -1,0 +1,288 @@
+class AddExistingTablesFromDotnet < ActiveRecord::Migration[5.2]
+  def up
+    # These are extensions that must be enabled in order to support this database
+    # enable_extension "plpgsql"
+
+    # create_table "__EFMigrationsHistory", primary_key: "MigrationId", id: :string, limit: 150, force: :cascade do |t|
+    #   t.string "ProductVersion", limit: 32, null: false
+    # end
+
+    unless table_exists(:access_request)
+      create_table "access_request", id: :serial, force: :cascade do |t|
+        t.text "email_address"
+        t.text "first_name"
+        t.text "last_name"
+        t.text "organisation"
+        t.text "reason"
+        t.datetime "request_date_utc", null: false
+        t.integer "requester_id"
+        t.integer "status", null: false
+        t.text "requester_email"
+        t.index ["requester_id"], name: "IX_access_request_requester_id"
+      end
+    end
+
+    unless table_exists(:course)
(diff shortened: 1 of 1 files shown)


misaka:
[Move DB migrations to Rails app.](https://trello.com/c/oYqWlNMY/961-move-db-migrations-to-rails-app)

db/migrate/20190212161359_add_existing_tables_from_dotnet.rb
```diff
@@ -0,0 +1,288 @@
...
+      create_table "organisation_provider", id: :serial, force: :cascade do |t|
+        t.integer "provider_id"
+        t.integer "organisation_id"
+        t.index ["organisation_id"], name: "IX_organisation_provider_organisation_id"
+        t.index ["provider_id"], name: "IX_organisation_provider_provider_id"
+      end
+    end
+
+    unless table_exists(:organisation_user)
+      create_table "organisation_user", id: :serial, force: :cascade do |t|
```
misaka: This can probably be:

```
create_join_table(:organisation, :user)
```

or similar.

db/migrate/20190212161359_add_existing_tables_from_dotnet.rb
```diff
@@ -0,0 +1,288 @@
...
+    unless table_exists(:organisation)
+      create_table "organisation", id: :serial, force: :cascade do |t|
+        t.text "name"
+        t.text "org_id"
+        t.index ["org_id"], name: "IX_organisation_org_id", unique: true
+      end
+    end
+
+    unless table_exists(:organisation_provider)
+      create_table "organisation_provider", id: :serial, force: :cascade do |t|
```
misaka: This can probably be:

```
create_join_table(:organisation, :provider)
```

or similar.

misaka:
@defong Good idea. I wouldn't bother except that we're probably going to regenerate this migration anyway with a proper dump of the DB, at which point it'll be the same amount of work to do this more refined approach.

misaka:
I don't think we need this anymore. It looks like having `schema.rb` will be good enough.

PR #110, opened 2019-02-14
[914] authenticate requests

misaka:
### Context

Setup authentication so that the frontend can make requests on behalf of a user.

Add an authentication method which will be used by the v2 api.

### Guidance to review

Added config gem to integrate configuration with env vars.
Use JWT for the token.

This PR is based off of the branch `914-add-config-gem` which has it's own PR #111, since it uses the settings introduced there to configure authentication.


defong:
[Authenticate requests coming into manage courses backend](https://trello.com/c/vNmM5nOe/914-authenticate-requests-coming-into-manage-courses-backend)

app/controllers/application_controller.rb
```diff
@@ -1,15 +1,8 @@
 class ApplicationController < ActionController::API
   include ActionController::HttpAuthentication::Token::ControllerMethods
 
-  before_action :authenticate
```
benilovj: @misaka this might be a bit OTT defensive, but is it worth:

1. keeping the `before_action :authenticate`
2. have `authenticate` in this class throw an exception

The reason for that is:

* all endpoints in this app should and are behind authentication
* if someone inadvertently subclasses `ApplicationController`, rather than the `V*ApplicationController` it will be open by default

app/controllers/api/v1/v1_application_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Api
+  module V1
+    class V1ApplicationController < ApplicationController
```
misaka: This should be:

```
module Api
  module V1
```

I _think_ `::ApplicationController` will do the trick there, worth trying.

app/controllers/application_controller.rb
defong: hmmm literal leave this line in

`before_action :authenticate`

and dont have a `def authenticate` ?

defong: see https://github.com/DFE-Digital/manage-courses-backend/pull/110/commits/ac25d5dde884d78009e9fa2bb5dcc08957f8bb35

benilovj:
I'm not sure I understand the implication of adding the `config` gem on how we configure existing stuff (e.g. through initialisers), also it feels incidental to this PR.

misaka:
@benilovj The `config` gem gives us a way to define settings a bit more richly than just with env vars, while still allowing us to use them to override settings defined there. What are you wondering about specifically?

I'm happy to move it out to a separate PR but I did want to put some settings into there for this work.

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,13 @@
+module Api
+  module V1
+    class ApplicationController < ::ApplicationController
+      include ActionController::HttpAuthentication::Token::ControllerMethods
```
dankmitchell: This is included in the base application controller, so do we need it here too?

app/controllers/api/v2/application_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Api
+  module V2
+    class ApplicationController < ::ApplicationController
+      include ActionController::HttpAuthentication::Token::ControllerMethods
```
dankmitchell: This is included in the base application controller, so do we need it here too?

config/initializers/config.rb
```diff
@@ -46,5 +46,4 @@
   #   required(:age).maybe(:int?)
   #   required(:email).filled(format?: EMAIL_REGEX)
   # end
-
```
dankmitchell: Lets remove this file

app/controllers/api/v2/application_controller.rb
```diff
@@ -0,0 +1,25 @@
+module Api
+  module V2
+    class ApplicationController < ::ApplicationController
+      def authenticate
+        authenticate_or_request_with_http_token do |token|
+          if Settings.authentication.algorithm == 'plain-text'
+            # This method can be used in development mode to simplify querying
+            # the API with curl. It should allow us to do:
+            #
+            #    curl -H 'Authorization: Bearer user@education.gov.uk' http://localhost:3000/api/v2/providers
```
misaka: @benilovj ^^^

benilovj: 👍

PR #111, opened 2019-02-14
[914] add config gem with settings to pick up env vars

misaka:
### Context

As the app grows we'll have more configuration points to add, for example the settings used for token authentication.

Add a gem that gives us a richer way to create settings used in the app. These settings can still be overridden by env vars set by 'dotenv' or otherwise.


config/initializers/config.rb
```diff
@@ -0,0 +1,50 @@
...
+  config.env_parse_values = true
+
+  # Validate presence and type of specific config values. Check https://github.com/dry-rb/dry-validation for details.
+  #
+  # config.schema do
+  #   required(:name).filled
+  #   required(:age).maybe(:int?)
+  #   required(:email).filled(format?: EMAIL_REGEX)
+  # end
+
```
defong: You broke the build....

image

misaka:
@benilovj Any questions here before we merge?

misaka:
[Authenticate requests coming into manage courses backend](https://trello.com/c/vNmM5nOe/914-authenticate-requests-coming-into-manage-courses-backend)

PR #112, opened 2019-02-14
Add ping endpoint

timabell:
As per c# api.
Allows testing deployment without worrying about access tokens.
Will also be useful for health monitoring.

### Context
Deploying

### Testing

```
$ curl -i http://localhost:3000/ping
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
ETag: W/"04afe236f86d27cf0155166c38e27919"
Cache-Control: max-age=0, private, must-revalidate
X-Request-Id: 7140f427-09de-4b50-91a0-4dfa32aacc08
X-Runtime: 0.006820
Transfer-Encoding: chunked

{"course_count":4}
```

### DIFF
--- a/app/controllers/ping_controller.rb
+++ b/app/controllers/ping_controller.rb
@@ -1,0 +1,8 @@
+class PingController < ApplicationController
+  skip_before_action :authenticate
+
+  # this is for quick testing of api/db stability without needing to authorise
+  def index
+    render json: { course_count: Course.count }
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -22,6 +22,7 @@
 #        error_500 GET    /error_500(.:format)                                error#error_500
 
 Rails.application.routes.draw do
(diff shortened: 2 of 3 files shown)


dankmitchell:
We can use the [health_check](https://github.com/ianheggie/health_check) gem for this @timabell

dankmitchell:
@timabell we should probably add this feature in the rails boilerplate

timabell:
ahrr

timabell:
well at least I won't have to work out why my test has weird numbers

misaka:
Actually, I think that the `ping` endpoint can be and should be slightly different from the `health_check` one:

`health_check`
- complete check that includes external resources
- may incur a little extra load as it checks resources
- may block if certain resources aren't available

`ping`
- lightweight, just tells you the app is responding
- provides SHA of the latest commit and possibly other hard-coded values

I can see how these two can co-exist.

PR #115, opened 2019-02-15
[987] setup users endpoint

misaka:
### Context

Originally started as a spike to get a JSON API endpoint, and to provide a `/users` endpoint at the same time.

Add the `/users` endpoint with only one action, to show user info. This is the first endpoint to use JSON API.

### Guidance to review

We technically may not actually need the `/users` endpoint for any of the stories currently on the board, however we do think that we'll be building endpoints off of `/users` for user-scoped data and that we'll want to have the canonical endpiont `/users/:user_id` for user info at some point. So we've decided to proceed with creating this endpoint.


config/routes.rb
```diff
@@ -30,6 +30,10 @@
         resources :courses
       end
     end
+
+    namespace :v2 do
+      resources :users
```
dankmitchell: Should we restrict the routes to match the controller actions? i.e. `resources :users, only: :show`

misaka:
I guess that the tests for the JSON output of `/api/v2/users` are a bit different from what some of us are used to, but I'm trying it out since I'm not a big fan of doing comparisons of large hash structures like we've done with the other endpoints. What do people think?

config/routes.rb
misaka: I don't know. I agree that it would be more "correct", but I think the behaviour if we don't is fine (it'll still generate 404s, right?), and it looks a bit tidier and easier to read in the routes file.

The 2 really convincing arguments for restricting the routes are when you have a large number of routes where you might get a speed benefit, or when you have the actions available in the controller but you don't want to expose them for some reason.

misaka: I say we aim for legibility for now.

README.md
```diff
@@ -55,6 +57,41 @@ configured in `config/environments/development.rb`:
...
+
+or with curl:
+
+```bash
+curl http://localhost:3000/api/v2/users.json -H "Authorization: Bearer user@digital.education.gov.uk"
+```
+
+#### Production
+
+In production mode the bearer token is an HMAC-encrypted JWT with the JSON payload:
```
defong: I think encrypted is just fine, vs xyz encryped

benilovj:
> I guess that the tests for the JSON output of `/api/v2/users` are a bit different from what some of us are used to, but I'm trying it out since I'm not a big fan of doing comparisons of large hash structures like we've done with the other endpoints. What do people think?

I personally find the large hash structure specs better as documentation of what the response looks like, but your approach is readable enough.

misaka:
[Setup users endpoint on rails API:V2](https://trello.com/c/SPy13rTX/987-setup-users-endpoint-on-rails-apiv2)

README.md
misaka: Are you worried about leaking security info?

defong: Let's not open ourselves to that as well as tying us to something in particular for no reason

misaka:
@benilovj I've changed the test to test the JSON verbatim.

PR #116, opened 2019-02-15
953 add provider changed at & configure docker to run migrations in prod etc

timabell:
### Context
`provider.changed_at` is going to be used to track any change that would cause a re-export for the ucas apply system. It's an amalgamation of last_published and updated_at along with dependent record changes

Doing this as a separate PR as it's our first ever migration on the rails side to avoid tying it up with the functionality to update it.

* First migration - adds new column.
* Baseline schema.rb version (done automatically by the rails tools)

### Guidance to review

Test migration from clean setup of db, and with sanitized backup from prod. I've done both and it worked for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 ### Native
 
 1. Run `bundle install` to install the gem dependencies
-2. Run `rails db:setup` to create a development and testing database
+2. Run `bundle exec rails db:setup` to create a development and testing database
 3. Run `bundle exec rails server` to launch the app on http://localhost:3000.
 
 ### Docker
--- a/README.md
+++ b/README.md
@@ -22,6 +22,11 @@
 2. Run `rails db:setup` to create a development and testing database
 3. Run `bundle exec rails server` to launch the app on http://localhost:3000.
 
+#### Updating
+
+1. `git pull`
+2. `rails db:migrate`
+
 ### Docker
 
 Run this in a shell and leave it running:
@@ -43,6 +48,12 @@ docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
 ```
 
 Then open http://localhost:3000 to see the app.
+
+#### Updating
+
+1. `git pull`
+2. `docker-compose up --build --detach` - this will recreate the image if there's any changes
+3. `docker-compose exec web /bin/sh -c "bundle exec rails db:migrate"`
 
 ## Accessing API
 
--- a/README.md
+++ b/README.md
@@ -22,11 +22,6 @@
 2. Run `rails db:setup` to create a development and testing database
 3. Run `bundle exec rails server` to launch the app on http://localhost:3000.
 
-#### Updating
-
-1. `git pull`
-2. `rails db:migrate`
-
 ### Docker
 
 Run this in a shell and leave it running:
@@ -48,12 +43,6 @@ docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
 ```
 
 Then open http://localhost:3000 to see the app.
-
-#### Updating
-
-1. `git pull`
-2. `docker-compose up --build --detach` - this will recreate the image if there's any changes
-3. `docker-compose exec web /bin/sh -c "bundle exec rails db:migrate"`
 
 ## Accessing API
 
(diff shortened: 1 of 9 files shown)


timabell:
[Update `changed_at` when updating provider or related entities.](https://trello.com/c/CqS3w68r/953-update-changedat-when-updating-provider-or-related-entities)

timabell:
[Move DB migrations to Rails app.](https://trello.com/c/oYqWlNMY/961-move-db-migrations-to-rails-app)

misaka:
Is this change complete? I'm ok with the change, but I think we still need:

1. A way to run the migrations on deployment. This could be a separate PR but, again, might may be hard to test depending on what logs we get from the deploy. If we put it in here we'll get to test it by running this migration.
2. Something that updates the `changed_at`. Not strictly necessary but this change here seems so small that I wonder if it's worth shipping without something that uses it.

timabell:
wip - needs to add the mechanism for running the migrations in prod

timabell:
I'd like to get just the field in without the code to update it so we can prove the migrations approach works and use it as a base for other work.

docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
       cache_from:
         - batdevcontainerregistry.azurecr.io/manage-courses-backend:latest
     image: batdevcontainerregistry.azurecr.io/manage-courses-backend:latest
-    command: ash -c "rm -f tmp/pids/server.pid && bundle exec rails server -p 3000 -b '0.0.0.0'"
+    command: ash -c "rm -f tmp/pids/server.pid && bundle exec rails db:migrate && bundle exec rails server -p 3000 -b '0.0.0.0'"
```
misaka: I think we need this to be part of `Dockerfile` so that it's run in the Azure app .
server.

defong: refer to https://github.com/DFE-Digital/manage-courses-backend/blob/05e32186b351c61c823382dc86775fa9d09ef8ea/.travis.yml#L12

`bundle exec rails db:migrate` may just blow up as, basically `db` sever may be up but `db database` server in side of `db` may not be up

timabell: you are correct, sir. Doing that now...

timabell: cheers sau

davidgisbey:
Local testing against restore from prod.

```
docker run -e DB_HOSTNAME=host.docker.internal -e DB_DATABASE=mc  batdevcontainerregistry.azurecr.io/manage-courses-backend:latest 
== 20190213160920 AddChangedAtToProvider: migrating ===========================
-- add_column(:provider, :changed_at, :datetime, {:default=>#<Proc:0x0000561ea5375338@/app/db/migrate/20190213160920_add_changed_at_to_provider.rb:3 (lambda)>, :null=>false})
   -> 0.0018s
== 20190213160920 AddChangedAtToProvider: migrated (0.0019s) ==================

=> Booting Puma
=> Rails 5.2.2 application starting in development 
=> Run `rails server -h` for more startup options
Puma starting in single mode...
* Version 3.12.0 (ruby 2.6.1-p33), codename: Llamas in Pajamas
* Min threads: 5, max threads: 5
* Environment: development
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop


```

timabell:
We've tested all the paths through and they all seem to work. Added info for developers pulling migrations onto their own machine.

README.md
```diff
@@ -22,6 +22,11 @@
 2. Run `rails db:setup` to create a development and testing database
 3. Run `bundle exec rails server` to launch the app on http://localhost:3000.
 
+#### Updating
```
dankmitchell: We dont need this

README.md
```diff
@@ -44,6 +49,12 @@ docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
 
 Then open http://localhost:3000 to see the app.
 
+#### Updating
+
+1. `git pull`
+2. `docker-compose up --build --detach` - this will recreate the image if there's any changes
+3. `docker-compose exec web /bin/sh -c "bundle exec rails db:migrate"`
```
dankmitchell: I'd just run `migrate` by default

PR #118, opened 2019-02-18
[950] Courses api v2

benilovj:
Paired with @defong on this.

### Context
Course API needed for rendering course summary info in Publish.

Followed the patterns in #115.

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -1,0 +1,12 @@
+module API
+  module V2
+    class CoursesController < ApplicationController
+      def index
+        provider = Provider.find_by!(provider_code: params[:provider_code])
+
+        render jsonapi: provider.courses,
+               class: SERIALIZABLE_CLASSES
+      end
+    end
+  end
+end
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -34,4 +34,20 @@ class SiteStatus < ApplicationRecord
   def recruitment_cycle
     "2019"
   end
+
+  def findable?
+    status == "Running" && publish == 'Y'
+  end
+
+  def has_vacancies?
+    [
+      "Both full time and part time vacancies",
+      "Part time vacancies",
+      "Full time vacancies"
+    ].include?(vac_status)
+  end
+
+  def applications_being_accepted_now?
+    applications_accepted_from.present? && applications_accepted_from <= Time.now.utc
+  end
 end
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -33,16 +33,16 @@ RSpec.describe Course, type: :model do
     it { should have_many(:sites) }
   end
 
-  describe `no site statuses` do
+  describe 'no site statuses' do
     it { expect(subject.site_statuses).to be_empty }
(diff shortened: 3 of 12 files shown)


defong:
[Create api/v2 for manage courses frontend for course summary](https://trello.com/c/ChXP2RQm/950-create-api-v2-for-manage-courses-frontend-for-course-summary)

spec/models/course_spec.rb
```diff
@@ -24,12 +24,132 @@
...
+  let(:subject) { create(:course) }
   describe 'associations' do
     it { should belong_to(:provider) }
     it { should belong_to(:accrediting_provider) }
     it { should have_and_belong_to_many(:subjects) }
+    it { should have_many(:site_statuses) }
+    it { should have_many(:sites) }
+  end
+
+  describe `no site statuses` do
```
tvararu: Using backticks here is weird as it's also used to run inline shell commands, we should change (there are more in the rest of the file) these to single/double quotes.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -0,0 +1,12 @@
+module API
+  module V2
+    class CoursesController < ApplicationController
+      def index
+        provider = Provider.find_by!(provider_code: params[:provider_code])
```
misaka: We should be getting Pundit or similar in here to handle authorization.

misaka: Happy to do this as a later PR, however.

defong: this thing, https://github.com/varvet/pundit 

ok

app/models/site_status.rb
```diff
@@ -34,4 +34,20 @@ class SiteStatus < ApplicationRecord
   def recruitment_cycle
     "2019"
   end
+
+  def findable?
+    status == "Running" && publish == 'Y'
```
misaka: If we define an enum for "published", I'm pretty sure this can be achieved in SQL with a scope like this:

```
scope :findable, -> { Running.published }
```

(although, I say we rename `Running` to `running` ...)

And similarly, we should be able to create the same method as you have:

```
def findable?
  Running? && published?
end
```

misaka: We can do these later with the tickets:

1. rename enums to not be capitalised and without spaces
2. add enum for `publish` column
3. refactor `findable?` in `Course` and `SiteStatus` model to use the scopes generated by the enums

spec/models/course_spec.rb
```diff
@@ -24,12 +24,132 @@
...
+    describe 'applications_being_accepted_now?' do
+      context 'with at least one site status applications_being_accepted_now' do
+        context 'single site status applications_being_accepted_now as it open now' do
+          let(:subject) { create(:course, site_statuses: site_statuses) }
+
+          let(:site_statuses) {
+            [create(:site_status, :applications_being_accepted_now)]
+          }
+
+          it { expect(subject.site_statuses).not_to be_empty }
```
```
its(:site_statuses) { should_not be_empty }
```

app/models/site_status.rb
defong: Carded

https://trello.com/c/iUyp5KXW/995-ruby-enums-refactor\

PR #119, opened 2019-02-18
[998] Add providers endpoint and serializer

dankmitchell:
### Context
V2 Providers endpoint

Add v2 providers serialiser 

### Guidance to review
`api/v2/users/:user_id/providers`


dankmitchell:
[Add api/v2/providers endpoint](https://trello.com/c/erPNdfLH/998-add-api-v2-providers-endpoint)

config/routes.rb
```diff
@@ -39,6 +39,7 @@
 
     namespace :v2 do
       resources :users
+      resources :providers
       scope :providers do
```
defong: Can this be nested inside of the `resource` one?

misaka: I think @defong is right ... even though I think we should have `/providers`, I think that for the sake of the API we are actually requesting the user's list of providers, which means using `/users/:user_id/providers` to make that explicit.

app/controllers/api/v2/application_controller.rb
```diff
@@ -4,6 +4,7 @@ class ApplicationController < ::ApplicationController
       SERIALIZABLE_CLASSES = {
         User: API::V2::UserSerializable,
         Course: API::V2::CourseSerializable,
+        Provider: API::V2::ProviderSerializable
```
defong: Just wondering 
image

http://jsonapi-rb.org/guides/getting_started/rails.html 

does it mean we can omit it?

dankmitchell: Doing this in a separate PR @defong

app/policies/provider_policy.rb
```diff
@@ -20,6 +20,6 @@ def initialize(user, provider)
   end
 
   def index?
-    user.providers.include?(provider)
```
dankmitchell: @misaka this change is breaking the `/providers/:id/courses` request spec...

dankmitchell:
@misaka @benilovj i've been thinking about this and wondering if this PR should be changed to just `/providers` and then use the pundit scope to decide which providers the authenticated user can view? In the same way the `providers/:id/courses` endpoint is setup...

From my limited understanding of pundit we can easily add a scope to provider policy to allow an "admin" to view all the providers...

app/policies/provider_policy.rb
```diff
@@ -19,7 +19,14 @@ def initialize(user, provider)
...
+  def can_list_providers?
+    user.present?
+  end
+
+  def can_manage_provider?
     user.providers.include?(provider)
   end
+
+  alias_method :index?, :can_list_providers?
+  alias_method :show?, :can_manage_provider?
```
misaka: My first intuition was to do this like so:

```
def index?
  user.present?
end

def show?
   user.providers.include?(provider)
end

alias_method :can_list_courses?, :show?
```

benilovj: Fixed.

benilovj:
@misaka ready for another review 😅

spec/policies/user_policy_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+describe UserPolicy do
+  describe 'show?' do
```
misaka: Backfilling tests. :100:

benilovj:
> Sorry to pick but there's one new method that's not covered.
> 
> I can add the test for it though.

If you could add a spec, that'd help get this across the line 👍 thanks

misaka:
> > Sorry to pick but there's one new method that's not covered.
> > I can add the test for it though.
> 
> If you could add a spec, that'd help get this across the line 👍 thanks

Done. LGTM!

PR #122, opened 2019-02-19
[995] Course bugfix, enum and course improvements

benilovj:
### Context
* Rails enums generate methods by default. If the enum values are more ruby-esque in the first place, the helpers read much nicer
* It's best to keep strings as a presentation concern
* Code that uses ActiveRecord (and hence the database) is likely to be more performant that reading all the objects in memory and performing ruby operations on them

* rubify the enums
* re-implement certain `Course` methods using scopes

### Guidance to review
Came out of a review from #118.

**Update**: while cleaning up this PR, @defong and I found a bug in the implementation of #118. This is fixed here.


spec/models/site_status_spec.rb
```diff
@@ -18,126 +18,4 @@
     it { should belong_to(:site) }
     it { should belong_to(:course) }
   end
-
-  describe 'findable?' do
```
misaka: As discussed IRL, can we find a way to include these tests, but maybe tidy them up to make them more readable? I'm concerned that even though these are technically being tested elsewhere, it's going to be hard to rationalise the test coverage for this model in the future when we don't remember that these methods are being tested elsewhere.

An idea for a custom matcher:

```
RSpec::Matchers.define :have_site_status do |*expected|
  match do |actual|
    if :findable.in? expected
      expect(SiteStatus.findable).to include actual
    end
  end
end
```

And using it:

```
specify do
   expect(subject).to have_site_status(:findable).and_not(:open_for_applications)
end
```

benilovj:
@misaka we've re-instated the specs and found a bug in the process! 💯 Please have another look at this new and improved PR.

PR #123, opened 2019-02-19
[997] Authorisation with pundit

benilovj:
### Context
To prevent unauthorised publishers inadvertently gaining access to other publishers' data.

Introduce [Pundit](https://github.com/varvet/pundit), a commonly used authorisation framework and apply it to all `v1` and `v2` endpoints.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,9 @@ gem 'rails', '~> 5.2.2'
 # Use PostgreSQL as the database for Active Record
 gem 'pg'
 
+# Authorisation
+gem 'pundit'
+
 # Use Puma as the app server
 gem 'puma', '~> 3.11'
 
--- a/app/controllers/api/v2/application_controller.rb
+++ b/app/controllers/api/v2/application_controller.rb
@@ -1,6 +1,8 @@
 module API
   module V2
     class ApplicationController < ::ApplicationController
+      attr_reader :current_user
+
       SERIALIZABLE_CLASSES = {
         User: API::V2::UserSerializable,
         Course: API::V2::CourseSerializable,
@@ -21,8 +23,8 @@ module API
             payload = JSON.parse(json_payload)
             email = payload['email']
           end
-          @authenticated_user = User.find_by(email: email)
-          @authenticated_user.present?
+          @current_user = User.find_by(email: email)
+          @current_user.present?
         end
       end
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,8 +1,10 @@
 class ApplicationController < ActionController::API
   include ActionController::HttpAuthentication::Token::ControllerMethods
+  include Pundit
 
   # child must define authenticate method
   before_action :authenticate
+  after_action :verify_authorized
   rescue_from PG::ConnectionBad, with: :render_service_unavailable
 
 private
--- a/app/policies/provider_policy.rb
+++ b/app/policies/provider_policy.rb
@@ -1,0 +1,12 @@
+class ProviderPolicy
+  attr_reader :user, :provider
+
+  def initialize(user, provider)
+    @user = user
+    @provider = provider
+  end
+
+  def index?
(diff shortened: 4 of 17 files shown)


spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -18,20 +19,33 @@
...
+    }
     subject { response }
 
     before do
       get "/api/v2/providers/#{provider.provider_code}/courses",
           headers: { 'HTTP_AUTHORIZATION' => credentials }
     end
 
-    context 'when unauthorized' do
+    context 'when unauthenticated' do
```
defong: `unauthenticated` vs `unauthorized` you sure as http code returns `unauthorized`

Gemfile
```diff
@@ -9,6 +9,9 @@ gem 'rails', '~> 5.2.2'
 # Use PostgreSQL as the database for Active Record
 gem 'pg'
 
+# Authorisation
+gem "pundit"
```
dankmitchell: Single quotes / double 🤷‍♂️

app/controllers/api/v2/application_controller.rb
```diff
@@ -1,6 +1,8 @@
 module API
   module V2
     class ApplicationController < ::ApplicationController
+      attr_reader :current_user
```
dankmitchell: I wonder if we could use the new current attributes from rails 5.2 here?

spec/requests/api/v2/providers/courses_spec.rb
benilovj: I **think** this is right – there isn't a separate `unauthorized` HTTP status code to my knowledge.

As I understand it:

* an unauthenticated user isn't someone who we have on the system
* an unautorised user is someone who we have on the system, but shouldn't have access to the thing they're trying to access

Gemfile
benilovj: Good spot, fixed.

app/controllers/api/v2/application_controller.rb
benilovj: I've not come across this yet – anything you can point me towards?

dankmitchell: Create a class or concern like so -

```
class Current < ActiveSupport::CurrentAttributes
  attribute :user
end
```

then you can assign `Current.user = @user` anywhere which gives you access to `current_user` throughout the app.

Haven't used it myself, just read about it.

dankmitchell: Don't let this stop you merging...

spec/requests/api/v2/providers/courses_spec.rb
misaka: Ya, I think you're right. And as you point out, @benilovj, "Authorisation" seems to be the catch-all term, but if we can disambiguate, that seems to be the correct definition.

app/policies/provider_policy.rb
```diff
@@ -0,0 +1,12 @@
+class ProviderPolicy
```
misaka: The other thing we should implement here is the [policy scope](https://github.com/varvet/pundit#scopes). Were you saving this for a later PR?

app/controllers/application_controller.rb
```diff
@@ -1,13 +1,20 @@
...
 
 private
 
   def render_service_unavailable
     render json: { code: 503, status: 'Service Unavailable' }.to_json, status: 503
   end
+
+  def authorization_error
+    render json: { code: 401, status: 'Unauthorised user' }.to_json, status: 401
+  end
```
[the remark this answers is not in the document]
defong: see https://github.com/DFE-Digital/manage-courses-backend/pull/124 as example

benilovj:
@misaka I've had to tweak b147811da07d193a8025ce8cc33eb3ac70ddcd9e to get a couple of request specs to pass, they broke as a result of the change to exception handling. It's running green now...

PR #124, opened 2019-02-19
[875] handle PG connection errors in ActionDispatcher

misaka:
This was originally implemented in the controller, but we realised that this
will be hiding stack traces in development and is better implemented in
ActionDispatch (which is the way Sau oroginally wanted to do it1). We've
reverted to this implemention and added a super-simple (naive) configuration
spec to test it.


config/routes.rb
```diff
@@ -37,4 +37,5 @@
   end
 
   get 'error_500', to: 'error#error_500'
+  get 'error_nodb', to: 'error#error_nodb'
```
dankmitchell: Should this just be `503` ?

misaka: I don't think so. The intention was to test what happens when we get a db error, and yes it should be a 503, but that's not what we're explicitly triggering.

defong: `error_nosql_503`?

misaka:
> Approved but what happened to the test for double checking the config?

Oops! It's back now.

PR #125, opened 2019-02-19
[953] Set provider.changed_at

davidgisbey:
Will be used to return changed records in the UCAS API

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -28,6 +28,7 @@
 
 class Provider < ApplicationRecord
   include RegionCode
+  include ChangedAt
 
   enum provider_type: {
     "SCITT" => "B",
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -237,4 +237,18 @@ RSpec.describe Provider, type: :model do
       end
     end
   end
+
+  describe 'changed_at' do
+    it 'is set on create' do
+      expect(subject.changed_at).to be_present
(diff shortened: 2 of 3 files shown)


app/models/provider.rb
```diff
@@ -60,4 +62,10 @@ def address_info
       .attributes_before_type_cast
       .slice('address1', 'address2', 'address3', 'address4', 'postcode', 'region_code')
   end
+
+private
+
+  def update_changed_at
+    self.changed_at = Time.now
```
dankmitchell: `Time.now.utc` ?

spec/models/provider_spec.rb
```diff
@@ -237,4 +237,16 @@
...
   end
+
+  describe '#changed_at' do
+    context 'record is modified' do
+      let!(:provider) { create(:provider, changed_at: 1.hour.ago) }
+
+      before { provider.contact_name = "bob" }
+
+      it 'should have been updated'do
+        expect(provider.changed_at).to be_within(1.second).of(Time.now)
```
defong: `Time.now.utc` ?

davidgisbey: Thanks! Will change

app/models/provider.rb
davidgisbey: Thanks! Will change

misaka:
What do we think about patching into `timestamp_attributes_for_update` like this to update `changed_at` automagically?

PR #129, opened 2019-02-20
Clean up v2 serializable

dankmitchell:
- Clean up `SERIALIZABLE_CLASSES`

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -5,8 +5,7 @@ module API
         provider = Provider.find_by!(provider_code: params[:provider_code])
         authorize provider
 
-        render jsonapi: provider.courses,
-               class: { Course: API::V2::CourseSerializable }
+        render jsonapi: provider.courses
       end
     end
   end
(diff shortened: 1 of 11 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -6,7 +6,7 @@ def index
         authorize provider
 
         render jsonapi: provider.courses,
-               class: SERIALIZABLE_CLASSES
+               class: { Course: API::V2::CourseSerializable }
```
defong: I meant something just like this

```ruby
render jsonapi: provider.courses
```

omitting the need for `class`

see the hook
https://github.com/jsonapi-rb/jsonapi-rails/blob/14a9421658495fe916ec1886834b6adfd764b211/lib/jsonapi/rails/controller/hooks.rb#L13

The thing I would of expected is either conforming to a naming convention.

If you look at this test its don's specify a 'class' and the json returns contain the fields

misaka: > omitting the need for `class`

Good find @defong. 💯 I've made a change that should sort this out.

defong: nice one, I had no idea how to approach it, :upside_down_face:

PR #131, opened 2019-02-20
[953] update provider.changed_at when site changes

davidgisbey:
### Context

We need to track this for the incremental API so that it can include all related data that might have changed. I.e. if a site has changed then that provider needs to be included in the relevant `changed_since` range even though the provider record hasn't changed.

`provider.changed_at` is updated when site is updated, while `provider.updated_at` is unaffected.

We also change the implementation of the incremental load by `from_provider_id` from the `Next` link, and using sub-second values for `changed_since` manages to achieve the desired goal of ordering the set of changed providers with a resolution higher then seconds. This is supported by the `changed_at` column without any modifications, we just need to format the timestamp with the `strftime` format `%FT%T.%6N` instead of using `iso8601`.

The `courses` controller will have to be updated to also use sub-second values for `changed_since`

### DIFF
--- a/app/controllers/api/v1/providers_controller.rb
+++ b/app/controllers/api/v1/providers_controller.rb
@@ -27,6 +27,12 @@ module API
 
         last_provider = @providers.last
 
+        # When we extract the changed_at from the last provider, format it with
+        # sub-second timing information (micro-seconds) so that our incremental
+        # fetch can handle many records being updated within the same second.
+        #
+        # The strftime format '%FT%T.%6NZ' is similar to the ISO8601 standard,
+        # (equivalent to %FT%TZ) and adds micro-seconds (%6N).
         response.headers['Link'] = if last_provider
                                      next_link(last_provider.changed_at
                                                  .utc
--- a/app/controllers/api/v1/providers_controller.rb
+++ b/app/controllers/api/v1/providers_controller.rb
@@ -4,27 +4,37 @@ module API
       include NextLinkHeader
 
       # Potential edge case:
-      # It is possible for older updated_at values to written to the database after this API has been queried for changes. This would mean that these changes are missed when the client makes a subsequent request using the next-link.
+      #
+      # It is possible for older updated_at values to written to the database
+      # after this API has been queried for changes. This would mean that these
+      # changes are missed when the client makes a subsequent request using the
+      # next-link.
+      #
       # Possible causes of older updated_at values:
-      # - delay between c# calculating datetime.UtcNow and value being written to postgres
+      # - delay between c# calculating datetime.UtcNow and value being written
+      #   to postgres
       # - clock drift between servers
       def index
         per_page = params[:per_page] || 100
         changed_since = params[:changed_since]
-
         ActiveRecord::Base.transaction do
           ActiveRecord::Base.connection.execute('LOCK provider, provider_enrichment, site IN SHARE UPDATE EXCLUSIVE MODE')
-          @providers = Provider.opted_in.changed_since(changed_since).limit(per_page)
+          @providers = Provider
+                         .opted_in
+                         .changed_since(changed_since)
+                         .limit(per_page)
         end
 
         last_provider = @providers.last
 
         response.headers['Link'] = if last_provider
-                                     next_link((last_provider.last_published_at + 1.second).utc.iso8601, last_provider.id, per_page)
+                                     next_link(last_provider.changed_at
+                                                 .utc
+                                                 .strftime('%FT%T.%6NZ'),
+                                               per_page)
                                    else
-                                     next_link(params[:changed_since], "", per_page)
+                                     next_link(changed_since, per_page)
                                    end
-
         render json: @providers
       rescue ActiveRecord::StatementInvalid
         render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: 400
@@ -32,9 +42,9 @@ module API
 
     private
 
-      def next_link(changed_since, from_provider_id, per_page)
+      def next_link(changed_since, per_page)
         current_url = request.base_url + request.path
-        "#{current_url}?changed_since=#{changed_since}&from_provider_id=#{from_provider_id}&per_page=#{per_page}; rel=\"next\""
+        "#{current_url}?changed_since=#{changed_since}&per_page=#{per_page}; rel=\"next\""
       end
     end
   end
--- a/app/models/concerns/changed_at.rb
+++ b/app/models/concerns/changed_at.rb
@@ -4,6 +4,11 @@ module ChangedAt
   class_methods do
   private
 
+    # Hook into Rails' built-in mechanism to update `updated_at` by adding to
+    # it's list of columns that get updated when an object changes (by default
+    # this is 'updated_at' and 'updated_on'). This is simpler than using a
+    # before/after save hook and should allow our 'changed_at' to behave in
+    # exactly the same way as 'updated_at'.
     def timestamp_attributes_for_update
       super + %w[changed_at]
     end
--- a/app/models/concerns/changed_at.rb
+++ b/app/models/concerns/changed_at.rb
@@ -1,0 +1,11 @@
+module ChangedAt
+  extend ActiveSupport::Concern
+
+  class_methods do
+  private
+
+    def timestamp_attributes_for_update
+      super + %w[changed_at]
+    end
+  end
+end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -67,8 +67,9 @@ class Provider < ApplicationRecord
   end
 
   def update_changed_at(timestamp: Time.now.utc)
-    # We don't want `updated_at` to change, only `changed_at`. Otherwise what
-    # would be the point of separating the two?
+    # Changed_at represents changes to related records as well as provider
+    # itself, so we don't want to alter the semantics of updated_at which
+    # represents changes to just the provider record.
     update_columns changed_at: timestamp
   end
 end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -51,10 +51,10 @@ class Provider < ApplicationRecord
 
   scope :changed_since, ->(datetime) do
     if datetime.present?
-      where("last_published_at >= ?", datetime)
+      where("changed_at >= ?", datetime)
     else
-      where("last_published_at is not null")
-    end.order(:last_published_at, :id)
+      where("changed_at is not null")
+    end.order(:changed_at, :id)
   end
 
   scope :opted_in, -> { where(opted_in: true) }
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -29,6 +29,7 @@
 
 class Provider < ApplicationRecord
   include RegionCode
+  include ChangedAt
 
   enum provider_type: {
     scitt: "B",
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -49,9 +49,9 @@ class Provider < ApplicationRecord
            class_name: "ProviderEnrichment"
   has_many :courses
 
-  scope :changed_since, ->(datetime) do
-    if datetime.present?
-      where("changed_at >= ?", datetime)
+  scope :changed_since, ->(timestamp) do
+    if timestamp.present?
+      where("provider.changed_at > ?", timestamp)
     else
       where("changed_at is not null")
     end.order(:changed_at, :id)
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -65,4 +65,10 @@ class Provider < ApplicationRecord
       .attributes_before_type_cast
       .slice('address1', 'address2', 'address3', 'address4', 'postcode', 'region_code')
   end
+
+  def update_changed_at(timestamp: Time.now.utc)
+    # We don't want `updated_at` to change, only `changed_at`. Otherwise what
+    # would be the point of separating the two?
+    update_columns changed_at: timestamp
+  end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,8 +13,6 @@
 ActiveRecord::Schema.define(version: 2019_02_22_063449) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_buffercache"
-  enable_extension "pg_stat_statements"
   enable_extension "plpgsql"
 
   create_table "__EFMigrationsHistory", primary_key: "MigrationId", id: :string, limit: 150, force: :cascade do |t|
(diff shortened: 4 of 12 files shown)


app/models/concerns/changed_at.rb
```diff
@@ -0,0 +1,11 @@
+module ChangedAt
+  extend ActiveSupport::Concern
+
+  class_methods do
```
defong: `class_methods`

misaka: I'm pretty sure this is `class_methods &block`:

https://api.rubyonrails.org/classes/ActiveSupport/Concern.html#method-i-class_methods

defong: cool, just looked odd

misaka: I like that you're starting to get a feel for what looks wrong or right in Ruby. :) 🏅

app/controllers/api/v1/providers_controller.rb
```diff
@@ -4,37 +4,47 @@ class ProvidersController < ApplicationController
...
+      #
+      # It is possible for older updated_at values to written to the database
+      # after this API has been queried for changes. This would mean that these
+      # changes are missed when the client makes a subsequent request using the
+      # next-link.
+      #
       # Possible causes of older updated_at values:
-      # - delay between c# calculating datetime.UtcNow and value being written to postgres
+      # - delay between c# calculating datetime.UtcNow and value being written
+      #   to postgres
```
defong: Any ideas of the lag?

Maybe move the time setting to the database...

app/models/provider.rb
```diff
@@ -47,12 +48,12 @@ class Provider < ApplicationRecord
...
-    if datetime.present?
-      where("last_published_at >= ?", datetime)
+  scope :changed_since, ->(timestamp) do
+    if timestamp.present?
+      where("provider.changed_at > ?", timestamp)
     else
-      where("last_published_at is not null")
-    end.order(:last_published_at, :id)
+      where("changed_at is not null")
+    end.order(:changed_at, :id)
```
defong: `end.order` just looks odd

app/models/provider.rb
```diff
@@ -63,4 +64,10 @@ def address_info
       .attributes_before_type_cast
       .slice('address1', 'address2', 'address3', 'address4', 'postcode', 'region_code')
   end
+
+  def update_changed_at(timestamp: Time.now.utc)
+    # We don't want `updated_at` to change, only `changed_at`. Otherwise what
```
defong: so what's the definition?

When something is changed it's not an update?

app/models/provider.rb
```diff
@@ -63,4 +64,10 @@ def address_info
...
+    # would be the point of separating the two?
+    update_columns changed_at: timestamp
```
defong: ` update_columns changed_at:  Time.now.utc` instead?

The need for the parameter? Other then for test, maybe timecop?

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -21,20 +21,44 @@
...
     describe 'generated next link' do
-      let!(:old_provider) { create(:provider, last_published_at: 5.minute.ago.utc) }
-      let!(:last_provider) { create(:provider, last_published_at: 1.minute.ago.utc) }
-      let(:last_provider_id) { last_provider.id }
-
       before do
         allow(controller).to receive(:authenticate)
+      end
+
+      subject { Rack::Utils.parse_query(URI(response.headers['Link']).query) }
```
defong: maybe a test about `per_page` ...

spec/factories/providers.rb
```diff
@@ -40,6 +40,7 @@
     opted_in { true }
 
     transient do
+      changed_at           { nil }
```
defong: should it be transient?

spec/models/provider_spec.rb
```diff
@@ -28,14 +28,29 @@
...
+      provider = Provider.create
+      expect(provider.changed_at).to be_present
+      expect(provider.changed_at).to eq provider.updated_at
+    end
+
+    it 'is set on update' do
+      provider = create(:provider, updated_at: 1.hour.ago)
+      provider.touch
+      expect(provider.changed_at).to eq provider.updated_at
+      expect(provider.changed_at).not_to be_within(1.second).of(1.hour.ago)
```
defong: timecop?

app/controllers/api/v1/providers_controller.rb
misaka: We haven't witnessed the lag happening, we just hypothesized that it's possible. We try to protect against it by acquiring a lock on the table to make sure that there isn't an update in-flight that might cause this (line 21 below).

app/models/provider.rb
misaka: It's called functional programming. :) Do some lisp and it'll start to feel more comfortable.

Having said that, I'm a little reluctant to introduce an intermediary variable here, unless you think it would really help out? For example:

```
where_clause = if timestamp.present?
                 where("provider.changed_at > ?", timestamp)
               else
                 where("changed_at is not null")
               end
where_clause.order(:changed_at, :id)
```

app/models/provider.rb
misaka: `updated_at` - when the current object has been updated
`changed_at` - when anything on this, or related, object(s) have changed which will necessitate re-publishing to the apply system (via the incremental load system)

app/models/provider.rb
misaka: Not having a parameter her makes me itchy. It's trivial to do and it means no further change needed if we need it in the future. I don't think it gets in the way here, do you?

spec/controllers/api/v1/providers_controller_spec.rb
misaka: Good catch! Will add.

spec/factories/providers.rb
misaka: I needs to be for what happens on line 67 in the `after(:create)` block below.

spec/models/provider_spec.rb
misaka: Please expand. :)

timabell:
[Update `changed_at` when updating provider or related entities.](https://trello.com/c/CqS3w68r/953-update-changedat-when-updating-provider-or-related-entities)

app/models/provider.rb
defong: i was just think brackets...

vs reading out loud as `end order`

timabell: I'd keep the `end.order` if it's idiomatic rails, I could get used to it

app/models/provider.rb
defong: I'm not fussed, just stating.

spec/factories/providers.rb
defong: brain thought

`create(:provider, changed_at: time.now.utc)`

Comments 63 to 66 must make some sense, once its part of a chain of thought...

Rakefile
```diff
@@ -4,3 +4,7 @@
 require_relative 'config/application'
 
 Rails.application.load_tasks
+
+task lint: ['lint:ruby']
+task annotate: ['db:annotate']
+task default: %i[spec annotate lint]
```
timabell: :heart:

app/controllers/api/v1/providers_controller.rb
```diff
@@ -4,37 +4,47 @@ class ProvidersController < ApplicationController
...
+                         .limit(per_page)
         end
 
         last_provider = @providers.last
 
         response.headers['Link'] = if last_provider
-                                     next_link((last_provider.last_published_at + 1.second).utc.iso8601, last_provider.id, per_page)
+                                     next_link(last_provider.changed_at
+                                                 .utc
+                                                 .strftime('%FT%T.%6NZ'),
```
timabell: I'd like a comment explaining what this format is and why to save going to look it up, maybe with a url for the docs

app/models/concerns/changed_at.rb
```diff
@@ -0,0 +1,11 @@
...
+  private
+
+    def timestamp_attributes_for_update
+      super + %w[changed_at]
```
timabell: I think this could do with a code comment explaining what effect this has and why we needed to do it

spec/models/provider_spec.rb
defong: guessing this is the actual test subject

```ruby
provider.touch
``` inferring this

```ruby
 def touch_provider
    provider.update_changed_at
  end
```

hence the items in question


```ruby
expect(provider.changed_at).to eq provider.updated_at
expect(provider.changed_at).not_to be_within(1.second).of(1.hour.ago)
```

so this should be 
```ruby
expect(provider.updated_at).to be eq Time.now.utc
expect(provider.changed_at).to be eq Time.now.utc
```


therefore `TimeCop`

spec/controllers/api/v1/providers_controller_spec.rb
davidgisbey: Just added one.

db/schema.rb
```diff
@@ -13,8 +13,6 @@
 ActiveRecord::Schema.define(version: 2019_02_22_063449) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_buffercache"
```
dankmitchell: Is this change expected?

spec/requests/api/v1/provider_spec.rb
```diff
@@ -19,23 +19,23 @@ def get_provider_codes_from_body(body)
...
-              postcode: 'N1 5JN',
-              telephone: '020 812 345 678',
-              email: 'info@acmescitt.education.uk',
-              contact_name: 'Amy Smith',
-              region_code: :london,
-              accrediting_provider: 'Y',
-              scheme_member: 'Y',
-              last_published_at: DateTime.now.utc,
-              enrichments: [enrichment])
+               provider_name: 'ACME SCITT',
```
dankmitchell: Are these manual changes or changes from running `annotate` ?

db/schema.rb
timabell: it's not related to this work, so if we want to do it we should do that as a separate PR if we want it so we can easily roll it back

spec/requests/api/v1/provider_spec.rb
timabell: manual I think

dankmitchell: nvm. Annotate wont touch these files 👍

db/schema.rb
misaka: Right. This shouldn't be here. I'm sill not entirely clear on why it showed up. Good spot.

timabell: discussed IRL,
this isn't a migration so won't affect prod
happy to include this to eliminate the diff

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -21,20 +21,50 @@
...
+            should eq '100'
+          end
+        end
 
-        get :index, params: { changed_since: 10.minutes.ago.utc }
+        context 'using a changed_since before any providers have changed' do
+          let(:changed_since) { 10.minutes.ago.utc }
+
+          its(%w[changed_since]) do
+            should eq last_provider.changed_at.strftime('%FT%T.%6NZ')
```
timabell: it's a bit of a magic string, maybe a bit explanation or a named variable

spec/models/provider_spec.rb
misaka: Sure, we can do that.

spec/factories/providers.rb
timabell: eh?

spec/models/provider_spec.rb
```diff
@@ -248,4 +260,29 @@
       expect(Provider.opted_in).to match_array([opted_in_provider])
     end
   end
+
+  describe '#updated_changed_at' do
```
timabell: this name is a bit ambiguous, could we be more explicit? how about "changed_at is updated"

app/models/provider.rb
```diff
@@ -64,4 +65,10 @@ def address_info
       .attributes_before_type_cast
       .slice('address1', 'address2', 'address3', 'address4', 'postcode', 'region_code')
   end
+
+  def update_changed_at(timestamp: Time.now.utc)
+    # We don't want `updated_at` to change, only `changed_at`. Otherwise what
+    # would be the point of separating the two?
```
timabell: I'm not sure we should pose a question in the comments. How about explaining the separation
"changed_at represents changes to related records as well as provider itself, so we don't want to alter the semantics of updated_at which represents changes to just the provider record"

spec/models/provider_spec.rb
```diff
@@ -248,4 +260,29 @@
...
+
+    it 'sets changed_at to the current time' do
+      Timecop.freeze do
+        provider.update_changed_at
+        expect(provider.changed_at).to eq Time.now.utc
+      end
+    end
+
+    it 'sets changed_at to the given time' do
+      timestamp = Time.now.utc
```
timabell: is this at risk of co-incidental pass? maybe use 1 hour ago or something to differentiate from now

spec/models/provider_spec.rb
```diff
@@ -45,10 +45,12 @@
     end
 
     it 'is set on update' do
-      provider = create(:provider, updated_at: 1.hour.ago)
-      provider.touch
-      expect(provider.changed_at).to eq provider.updated_at
-      expect(provider.changed_at).not_to be_within(1.second).of(1.hour.ago)
+      Timecop.freeze do
```
dankmitchell: Nice!

spec/models/site_spec.rb
```diff
@@ -24,4 +24,22 @@
   describe 'associations' do
     it { should belong_to(:provider) }
   end
+
+  describe '#touch_provider' do
```
timabell: "touch provider sites" might be more accurate?

spec/requests/api/v1/provider_spec.rb
```diff
@@ -91,6 +92,8 @@ def get_provider_codes_from_body(body)
         expect(response).to have_http_status(:unauthorized)
       end
 
+      it 'includes correct next link in response headers'
```
timabell: missing do block? should this line be removed?

timabell: it's a todo

spec/models/site_spec.rb
dankmitchell: You are already in the context if a `site` so I think touching the provider is accurate

misaka:
@defong Are you happy for me to merge this now? I think I've addressed all your comments.

spec/models/site_spec.rb
misaka: Actually I think @timabell is right, since we don't actually call `touch_provider` here, we're testing the after `after_save` hook, and also by the `touch_provider` method coincidentally. But at this point we just need to get this merged so we can deal with this later, if at all.

PR #133, opened 2019-02-21
Build with docker compose

chelnak:
### Context
Switches the az pipelines build process to use docker-compose.

We now tag each image explicitly. This causes the cache_from command to fail as no image available with the latest tag on docker hub.

* Add -T switch to exec commands fix TTY error
* Align tagging with Travis steps
* Use --out parameter to push tests to an xml file on the host
* Remove docker-compose pull from build steps to avoid dependencies on latest
* Comment out cache_from

### Guidance to review
* Review changes / Make suggestions on how the process should work

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,7 +29,7 @@ steps:
   displayName: 'Build & setup'
 
 - script: |
-    docker-compose exec -T web /bin/sh -c 'bundle exec rake spec SPEC_OPTS="--format RspecJunitFormatter"' | sed -e 1d >> rspec-results.xml
+    docker-compose exec -T web /bin/sh -c 'bundle exec rake spec SPEC_OPTS="--format RspecJunitFormatter --out ./rspec-results.xml"'
   displayName: 'Execute tests'
 
 - script: |
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,6 +11,7 @@ pool:
 
 variables:
   imageName: 'manage-courses-backend-poc'
+  # Support multiple docker-compose files to allow travis and az pipelines to co exist: https://docs.docker.com/compose/extends/#multiple-compose-files
   dockerOverride: 'docker-compose -f docker-compose.yml -f docker-compose.az.yml'
 
 steps:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,14 +11,14 @@ pool:
 
 variables:
   imageName: 'manage-courses-backend-poc'
+  COMPOSE_INTERACTIVE_NO: 1
 
 steps:
 - script: |
     GIT_SHORT_SHA=$(echo $(Build.SourceVersion) | cut -c 1-7)
     docker_path=$(dockerHubUsername)/$(imageName)
     echo "##vso[build.updatebuildnumber]$GIT_SHORT_SHA"
-    echo "##vso[task.setvariable variable=tag;]$docker_path:$GIT_SHORT_SHA"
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -10,9 +10,9 @@ services:
   web:
     build:
       context: .
-      cache_from:
-        - batdevcontainerregistry.azurecr.io/manage-courses-backend:latest
-    image: batdevcontainerregistry.azurecr.io/manage-courses-backend:latest
+      #cache_from:
+      #  - govukdfe/manage-courses-backend-poc:latest
```
timabell: just delete it, we can always put it back

azure-pipelines.yml
```diff
@@ -15,26 +15,30 @@ variables:
...
+    docker-compose exec -T web /bin/sh -c 'bundle exec rake spec SPEC_OPTS="--format RspecJunitFormatter --out ./rspec-results.xml"'
   displayName: 'Execute tests'
 
 - script: |
-    docker run "$(tag)" /bin/sh -c "bundle exec govuk-lint-ruby app config db lib spec --format clang"
+    docker-compose exec -T web /bin/sh -c "bundle exec govuk-lint-ruby app config db lib spec --format clang"
   displayName: 'Execute linters'
 
 - script: |
     echo $password | docker login --username $(dockerHubUsername) --password-stdin
```
defong: Echoing password? Are you sure?

azure-pipelines.yml
```diff
@@ -15,26 +15,30 @@ variables:
...
-    echo "##vso[task.setvariable variable=tag;]$tag"
+    echo "##vso[task.setvariable variable=docker_path;]$docker_path"
   displayName: 'Set version number'
 
 - script: |
-    docker build -f Dockerfile -t $(tag) .
-  displayName: 'Build image'
+    docker-compose build
+    docker-compose up --no-build -d
+    docker-compose exec -T web /bin/sh -c "./wait-for-command.sh -c 'nc -z db 5432' -s 0 -t 20"
```
defong: sh vs ash for shell?

docker-compose.yml
chelnak: Removed

azure-pipelines.yml
chelnak: https://docs.docker.com/engine/reference/commandline/login/

See **Provide a password using STDIN**

chelnak:
@defong Interesting point re what shell to use.

ash is used inside docker-compose.yml however sh is used in the build (travis and az pipelines).

Probably a wider conversation?

defong:
> @defong Interesting point re what shell to use.
> 
> ash is used inside docker-compose.yml however sh is used in the build (travis and az pipelines).
> 
> Probably a wider conversation?

Let's use ps...

Can we just sh across the board, I don't think ash give us anything particaular

azure-pipelines.yml
```diff
@@ -11,30 +11,41 @@ pool:
 
 variables:
   imageName: 'manage-courses-backend-poc'
+  dockerOverride: 'docker-compose -f docker-compose.yml -f docker-compose.az.yml'
```
timabell: I'm not sure I understand what this does, which might be fine till I need to look after it :-)
Perhaps a comment on the line above explaining what it's for and why it's there, no need to repeat things I can read in the docs, though a link might be handy.

chelnak: Happy to link to the docs. This will go away when travis goes though.

chelnak: https://github.com/DFE-Digital/manage-courses-backend/pull/133/commits/15104ec9004137b8cee9aee2acbb7fc65b7b4ce4

PR #134, opened 2019-02-21
[1011] Update changed_at when provider enrichment is updated

davidgisbey:
### Context

Linked to the [953] changed_at pull request. 

https://github.com/DFE-Digital/manage-courses-backend/pull/131

Changes to the provider enrichment model now update changed_at.

- touch_provider implemented in the provider enrichment model
- touch_provider refactored into a concern
- Tests to check that changed_at is updated when provider enrichment is touched, while leaving updated_at unaffected.

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -30,7 +30,6 @@
 class Provider < ApplicationRecord
   include RegionCode
   include ChangedAt
-  include TouchProvider
 
   enum provider_type: {
     scitt: "B",
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -30,6 +30,7 @@
 class Provider < ApplicationRecord
   include RegionCode
   include ChangedAt
+  include TouchProvider
 
   enum provider_type: {
     scitt: "B",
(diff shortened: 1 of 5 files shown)


dankmitchell:
@davidgisbey if you change the base of this PR to point to #131 then this PR will be a lot smaller

misaka:
Ready for review, but cannot be merged #131 is merged.

app/models/provider.rb
```diff
@@ -29,6 +29,7 @@
 class Provider < ApplicationRecord
   include RegionCode
   include ChangedAt
+  include TouchProvider
```
misaka: Don't need this in `Provider`.

app/models/provider_enrichment.rb
```diff
@@ -17,11 +17,14 @@ class ProviderEnrichment < ApplicationRecord
   self.primary_key = "provider_code"
 
   include RegionCode
+  include TouchProvider
 
   enum status: { draft: 0, published: 1 }
 
   belongs_to :provider, foreign_key: :provider_code, primary_key: :provider_code
 
+  after_save :touch_provider
```
misaka: Can we move this into `TouchProvider`?

app/models/provider.rb
davidgisbey: Done 👍

app/models/provider_enrichment.rb
davidgisbey: Done 👍

PR #126, opened 2019-02-20
[1014] The Course's Qualifications

benilovj:
### Context
The course's qualifications

The possible qualification can only be one of the following
`QTS`
`PGCE with QTS`
`PGDE with QTS`
`PGCE`
`PGDE` 
based on the `profpost_flag` and `pdge_course` and is the course's subject is a `Further Education` in some combination

### Changes proposed in this pull request
Added the course's qualifications

#### Basic
image

#### Real Life
From the csharp side, the understanding of `qualification` is duplicated and disconnected and overloaded with usage.

duplicated as its spread across 3 projects 
disconnected due to mismatch compiled version of 
`GovUk.Education.SearchAndCompare.Domain.Models.Enums.IncludesPgce`
overloaded due to search and compare ui and api

From manage-courses-api
https://github.com/DFE-Digital/manage-courses-api/blob/89c801a71495fd06211e5e52e62598a042fb9fe8/src/ManageCourses.Api/Mapping/CourseMapper.cs#L206

https://github.com/DFE-Digital/manage-courses-api/blob/807e85297db2a73486cce824938e31aec159ffa8/src/ManageCourses.Domain/Models/CourseQualification.cs

based on the above as
```csharp
            {CourseQualification.Qts, IncludesPgce.No},
            {CourseQualification.QtsWithPgce, IncludesPgce.Yes},
            {CourseQualification.QtsWithPgde, IncludesPgce.QtsWithPgde},
            {CourseQualification.QtlsWithPgce, IncludesPgce.QtlsWithPgce},
            {CourseQualification.QtlsWithPgde, IncludesPgce.QtlsWithPgde}
```

disconnected
```csharp
            {CourseQualification.Qts, IncludesPgce.No},
            {CourseQualification.QtsWithPgce, IncludesPgce.Yes},
```

From search-and-compare-ui
https://github.com/DFE-Digital/search-and-compare-ui/blob/0f9513dfd1d622bcee33e72904555ebe093b4e15/shared/Views/Shared/Components/CourseDetails/_Qualification.cshtml

extracted from the above as (and is along with https://github.com/DFE-Digital/manage-courses-api/blob/89c801a71495fd06211e5e52e62598a042fb9fe8/src/ManageCourses.Api/Mapping/CourseMapper.cs#L46 and https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.UcasCourseImporter/importer/Mapping/QualificationMapper.cs as the final knowledge for the port to `ruby`)
```csharp
[...]

PR #135, opened 2019-02-21
[1014] Course qualifications

benilovj:
### Context

Currently, the possible qualifications at the end of a course are:

* `QTS`
* `PGCE with QTS`
* `PGDE with QTS`
* `PGCE`
* `PGDE` 

This is determined by:

* `Course#profpost_flag`
* At a `pdge_course` associated with that course (if there is one)
* Whether the course has been tagged with a `Further Education` subject

This is already implemented in the C# apps, you can see much more detail on #126.

Add course qualifications to the `Course` model and replace the `profpost_flag` attribute with a `qualifications` attribute on Course API v2.

#### Basic
image


benilovj:
We shouldn't merge this until we've performance tested this implementation with production-like data.

app/models/course.rb
```diff
@@ -79,4 +83,12 @@ def open_for_applications?
   def has_vacancies?
     site_statuses.with_vacancies.any?
   end
+
+  def qualifications
+    Qualifications.new(
+      profpost_flag: profpost_flag,
+      is_pgde: self.in?(Course.pgde),
+      is_fe: subjects.further_education.any?,
```
misaka: How about:

```
PgdeCourse.where(course_code: self.course_code, provider_code: provider.provider_code).exists?
```

app/models/pgde_course.rb
```diff
@@ -0,0 +1,11 @@
+# == Schema Information
+#
+# Table name: pgde_course
+#
+#  id            :integer          not null, primary key
+#  course_code   :text             not null
+#  provider_code :text             not null
+#
+
+class PgdeCourse < ApplicationRecord
```
misaka: How about:

```
class PGDECourses
```

?

app/models/course.rb
benilovj: Yup - fixed.

app/models/pgde_course.rb
benilovj: agreed – fixed

benilovj:
@misaka @defong ready for another review

app/models/qualifications.rb
```diff
@@ -0,0 +1,27 @@
...
+    qts_or_not + uni_qualification_or_not
+  end
+
+private
+
+  def qts_or_not
+    @is_further_education ? [] : [:qts]
+  end
+
+  def uni_qualification_or_not
```
defong: Maybe a comment or two with regards to `uni_qualification ` vs `further_education`

As layman terms,  uni and further education are synonymous to one and another in most sense.

Is uni_qualification just another way of stating it will be uni accreditation of a qualification?

app/models/qualifications.rb
```diff
@@ -0,0 +1,27 @@
...
+    if @is_pgde
+      [:pgde]
+    elsif @profpost_flag == "recommendation_for_qts"
```
defong: Any chance to rid of magic string?

config/initializers/inflections.rb
```diff
@@ -17,4 +17,5 @@
 
 ActiveSupport::Inflector.inflections do |inflect|
   inflect.acronym 'API'
+  inflect.acronym 'PGDE'
```
defong: Maybe use tablename instead so it close to the model

benilovj: Not sure what you mean? This is here so that autoloading tries to load `PGDECourse` instead of `PgdeCourse`.

app/models/qualifications.rb
benilovj: I can't find any. I think the fact that it's being used in the comparison with the `profpost_flag` should link it to that.

app/models/qualifications.rb
benilovj: I've renamed the methods for clarity – can you please have another look?

config/initializers/inflections.rb
defong: I was referring to 


```
class Mouse < ActiveRecord::Base
  self.table_name = "mice"
end
```

I guess this is something else

app/models/qualifications.rb
defong: == `"recommendation_for_qts"` 

old habits, never comfortable seeing comparing to a hardcoded string

defong: much better but what it is awar`e`ded

app/models/qualifications.rb
defong: much better but what it is awar`e`ded

benilovj: > much better but what it is awar`e`ded

fixed

config/initializers/inflections.rb
benilovj: Yeah, this is something else.

app/models/qualifications.rb
benilovj: I've gotten rid of the `Qualifications` class and put everything into a concern, so this point is now fixed: 89e834d

spec/factories/courses.rb
```diff
@@ -23,11 +23,45 @@
 
 FactoryBot.define do
   factory :course do
+    trait :in_further_education do
```
defong: Maybe just fussy but I liked it more when fields was at the top, and the traits was at the bottom

benilovj: The factory is using a trait that's defined further up...

defong:
```ruby 
Factory.define
  factory :course
#those fields
    sequence(:course_code) { |n| "C#{n}D3" }
    name { Faker::ProgrammingLanguage.name }
    qualification { 1 }
    association(:provider)
    resulting_in_pgce_with_qts

    transient do
      age { nil }
end

# lots of traits here
```
vs


```ruby 
Factory.define
  factory :course

# lots of traits here

#those fields
    sequence(:course_code) { |n| "C#{n}D3" }
    name { Faker::ProgrammingLanguage.name }
    qualification { 1 }
    association(:provider)
    resulting_in_pgce_with_qts

    transient do
      age { nil }
end

```


That was what I meant

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -15,7 +15,7 @@
     end
 
     let(:findable_open_course) {
-      create(:course,
+      create(:course, :resulting_in_pgce_with_qts,
```
defong: Ain't the default course  
has that?
`resulting_in_pgce_with_qts,`

```ruby 
Factory.define
  factory :course
#those fields
    sequence(:course_code) { |n| "C#{n}D3" }
    name { Faker::ProgrammingLanguage.name }
    qualification { 1 }
    association(:provider)
    resulting_in_pgce_with_qts

    transient do
      age { nil }
end

# maybe that is my justification for traits to be here ;P
```

spec/factories/courses.rb
benilovj: I got what you meant, but I didn't think that it would work to include a trait in a factory above the definition of the trait. However having tested it, it seems to be fine so I moved the traits to the bottom of the file and force-pushed.

spec/requests/api/v2/providers/courses_spec.rb
benilovj: The default course does indeed have that, but since we're asserting on `"qualifications" => %w[qts pgce],` further down, it's better IMHO to have an explicit connection between the input data and the output. It also means if the default qualifications in the course factory change, this spec won't break.

PR #137, opened 2019-02-22
[1015] Add opted_in for providers

dankmitchell:
### Context
UCAS Transition

Add `opted_in` to providers and scope v1/providers to only return when `true`


dankmitchell:
[incremental transition - tech work](https://trello.com/c/oH3VB9Zb/1015-incremental-transition-tech-work)

dankmitchell:
@benilovj do we know which providers are going to be "opted in"? 
I can knock up a rake task or migration to update those..

benilovj:
> @benilovj do we know which providers are going to be "opted in"?
> I can knock up a rake task or migration to update those..

Not yet – let's leave that as a separate task later, once we've decided and agreed the cohorts.

PR #138, opened 2019-02-22
[993] Create new endpoint in for user session

defong:
### Context
The ability to create a session and update the user last login time as will as the first and last name. 

Endpoint exists and frontend is able to "create" a session for new user login.
The last_login timestamp is incremented after this endpoint is accessed

### Guidance to review
This needs to work with frontend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,13 +129,13 @@ file
 ```
 
 ```bash
-export SETTINGS__FILE__BASED__SETTINGS_ENV1="machine wins"
+export SETTINGS__FILE__BASED__SETTINGS__ENV1="machine wins"
 ```
 
 ```ruby
 puts Settings.file.based.setting.env1
 
-# machine wins
+machine wins
 ```
 
 Any `Machine based env variables settings` that is not prefixed with `SETTINGS`.* are not considered for general consumption.
--- a/README.md
+++ b/README.md
@@ -115,6 +115,33 @@ Encoding the payload can be done with the [Ruby `jwt` gem](https://github.com/jw
 JWT.encode payload, SECRET, 'HS256'
 ```
 
+## Settings vs config vs Environment variables
+
+Refer to the [the config gem](https://github.com/railsconfig/config#accessing-the-settings-object) to understand the `file based settings` loading order.
+
+To override file based via `Machine based env variables settings`
+```bash
+cat config/settings.yml
+file
+  based
+    settings
+      env1: 'some file based value'
+```
+
+```bash
+export SETTINGS__FILE__BASED__SETTINGS_ENV1="machine wins"
+```
+
+```ruby
+puts Settings.file.based.setting.env1
+
+# machine wins
+```
+
+Any `Machine based env variables settings` that is not prefixed with `SETTINGS`.* are not considered for general consumption.
+
+
+
 ## CI variables
 
 You'll need to define the `AZURE_CR_PASSWORD` in Travis in order to successfully build and publish. This can be done using this command:
--- a/app/controllers/api/v2/sessions_controller.rb
+++ b/app/controllers/api/v2/sessions_controller.rb
@@ -1,0 +1,15 @@
+module API
+  module V2
+    class SessionsController < ApplicationController
+      def create
+        @current_user.update(
+          last_login_date_utc: Time.now.utc,
+          first_name: params[:first_name],
+          last_name: params[:last_name]
+        )
+
+        render jsonapi: @current_user
+      end
+    end
+  end
+end
--- a/spec/requests/api/v2/session_spec.rb
+++ b/spec/requests/api/v2/session_spec.rb
@@ -1,0 +1,88 @@
+require "rails_helper"
+
+describe '/api/v2/session', type: :request do
+  let(:user)    { create(:user) }
+  let(:payload) { { email: user.email } }
+  let(:token) do
+    JWT.encode payload.to_json,
+               Settings.authentication.secret,
+               Settings.authentication.algorithm
+  end
+  let(:credentials) do
+    ActionController::HttpAuthentication::Token.encode_credentials(token)
+  end
+
+  context 'when unauthenticated' do
+    let(:payload) { { email: 'foo@bar' } }
+
+    before do
+      get "/api/v2/users/#{user.id}",
+          headers: { 'HTTP_AUTHORIZATION' => credentials }
+    end
+
+    subject { response }
+
+    it { should have_http_status(:unauthorized) }
+  end
+
+  describe 'creating a session' do
+    let(:user)    { create(:user, last_login_date_utc: 10.days.ago) }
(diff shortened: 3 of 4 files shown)


app/controllers/api/v2/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+module API
+  module V2
+    class SessionsController < ApplicationController
```
defong: Currently flawed concept 

Endpoint needs to be a different auth mechanism

Cannot use `v2` auth as this end point is the precursor to being able to use `v2`

May need a `v3` auth to established client to client trusted communication vector

defong:
[Create new user sessions in API backend](https://trello.com/c/gHNepiOF/993-create-new-user-sessions-in-api-backend)

README.md
```diff
@@ -93,6 +93,45 @@ Encoding the payload can be done with the [Ruby `jwt` gem](https://github.com/jw
...
+```bash
+cat config/settings.yml
+file
+  based
+    setting
+      env1: 'some file based value'
+```
+
+```bash
+env SETTINGS__file__based__setting__env1=`machine wins"
```
dankmitchell: This reminds me of dotnet secrets 😭

app/controllers/api/v2/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+      def create
+        @current_user.update last_login_date_utc: Time.now.utc, first_name: params[:first_name], last_name: params[:last_name]
```
dankmitchell: Really picky but can we add a parentheses around the atttibutes please i.e. 
```
@current_user.update(last_login_date_utc: Time.now.utc, 
                     first_name: params[:first_name], 
                     last_name: params[:last_name])
```

README.md
defong: because it follows the same psychopathy, except the dotnet one is better as you cant accidentality commit it into repo

spec/requests/api/v2/session_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  describe 'creating a session' do
+    let(:user)    { create(:user, last_login_date_utc: 10.days.ago) }
+      # macOS is in not defintive in accuracy as its lacks nano seconds vs other OS
+      # TimeCop therefore has slight flaw in design
+    it 'saves the last login time' do
+      Timecop.freeze do
+        post '/api/v2/session',
+             headers: { 'HTTP_AUTHORIZATION' => credentials }
+
+        expect(user.reload.last_login_date_utc).to be_within(1.second).of Time.now.utc
```
misaka: Not quite sure what your above comment about Timecop is saying, but since you're using it you can simplify this test to be:

```
expect(user.reload.last_login_date_utc).to eq Time.now.utc
```

README.md
```diff
@@ -93,6 +93,45 @@ Encoding the payload can be done with the [Ruby `jwt` gem](https://github.com/jw
...
+config/environments/#{environment}.yml
+# load order 3, environment config settings based on one for the environments (either development|test|producation)
+printenv | grep SETTINGS
+# load order 4, Machine based env variables settings
+```
+
+All basic default settings are to be listed in `config/settings.yml`,
+which in order can be overridden by `config/settings/#{environment}.yml`,
+and in turn overrided by `config/environments/#{environment}.yml`,
+and finally overrided by `Machine based env variables settings`
```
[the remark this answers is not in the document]
defong: yeah I was just verbose, just to get my head around it, will link off to it instead.

Kinda leaning to removing the prefix altogether, what was the need for prefixing it at all again?

spec/requests/api/v2/session_spec.rb
defong: breeze thro this

https://github.com/travisjeffery/timecop/issues/97

defong:
```
1) /api/v2/session creating a session saves the last login time
     Failure/Error: expect(user.reload.last_login_date_utc).to eq Time.now.utc
     
       expected: 2019-02-26 12:52:26.530098940 +0000
            got: 2019-02-26 12:52:26.530098000 +0000
```

# macOS is in not defintive in accuracy as its lacks nano seconds vs other OS
# TimeCop therefore has slight flaw in design

I use linux vs you use mac.
TimeCop was most likely tied to mac defects.

Unless i wrapped it in wrong block...

defong: Now I'm just going to throw db in the equation as well.

README.md
```diff
@@ -115,6 +115,57 @@ Encoding the payload can be done with the [Ruby `jwt` gem](https://github.com/jw
...
+
+```ruby
+print Settings.file.based.setting.env1
+
+# machine wins
+```
+
+Any `Machine based env variables settings` that is not prefixed with `SETTINGS`.* are not considered for general consumption.
+
+## Linting
```
dankmitchell: No need to reference the linting here, it's already in the readme

defong: Argh bad merge

README.md
defong: readme amended

README.md
```diff
@@ -115,6 +115,33 @@ Encoding the payload can be done with the [Ruby `jwt` gem](https://github.com/jw
...
+```bash
+cat config/settings.yml
+file
+  based
+    settings
+      env1: 'some file based value'
+```
+
+```bash
+export SETTINGS__FILE__BASED__SETTINGS_ENV1="machine wins"
```
misaka: Need an extra double-score in front of `ENV1`: `...SETTINGS__ENV1`

README.md
```diff
@@ -115,6 +115,33 @@ Encoding the payload can be done with the [Ruby `jwt` gem](https://github.com/jw
...
+```
+
+```ruby
+puts Settings.file.based.setting.env1
+
+# machine wins
+```
```
misaka: Sorry to be picky, but if you're fixing the underscores above then also remove the `# ` since that wouldn't be printed out.

spec/requests/api/v2/session_spec.rb
misaka: But, if I change this test to test for equality (see my original comment), it works (tested locally). Have you tried it and had it fail?

misaka: Ok, I'm starting to understand what you mean Sau.

On Linux this test fails because the nanoseconds are different (Timecop hasn't frozen that, it seems). On OSX it doesn't fail, either because the nanoseconds are the same or maybe because `eq` doesn't compare them.

defong: To be correct in case anyone else wants to know

On a mac it nanosecond is zeroed out
On database when retrieving a date column nanosecond is not save, there for nanosecond is zeroed out.

Timecop do free the time inclusive of nanoseconds, but there seems to be other issues that question the overall functionality concerning time.

PR #139, opened 2019-02-23
Clean up hacky routes

dankmitchell:
### Context
- Provider code as URL

- Add `friendly_id` to use the `provider_code` as the slug inside of the `id`

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -2,7 +2,7 @@ module API
   module V2
     class CoursesController < ApplicationController
       def index
-        provider = Provider.find_by!(provider_code: params[:provider_code])
+        provider = Provider.friendly.find(params[:provider_code])
         authorize provider, :can_list_courses?
         authorize Course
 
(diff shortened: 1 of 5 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -2,7 +2,7 @@ module API
   module V2
     class CoursesController < ApplicationController
       def index
-        provider = Provider.find_by!(provider_code: params[:provider_code])
+        provider = Provider.friendly.find(params[:provider_id])
```
defong: `params[:provider_id]` so is that the `friendly_id` aka `provider_code`?

dankmitchell: Yes. We define what the slug/url in the model so the controllers and routes are standard rails with the except of the `friendly` helper in the controller to hook it all together.

dankmitchell:
@timabell I dont know where those commits came from but ive cleaned it up

defong:
@dankmitchell long live pr?

dankmitchell:
Closing as I know @misaka isn’t keen

PR #145, opened 2019-02-25
[1036] Amended default port to be used from 3000 to 3001

defong:
### Context
Post collision with its client, since client port change has external factors involved, its done here instead.

Amended default port to be used from 3000 to 3001

### Guidance to review
Client should now be in a position to be seeded with default values for development.

Soon rather than later it be nice to have the 2 rails chatting with one and another.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 1. Run `bundle install` to install the gem dependencies
 2. Run `bundle exec rails db:setup` to create a development and testing database
-3. Run `bundle exec rails server` to launch the app on http://localhost:3000.
+3. Run `bundle exec rails server` to launch the app on http://localhost:3001.
 
 ### Docker
 
@@ -43,7 +43,7 @@ The first time you run the app, you need to set up the databases. With the above
 docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
 ```
 
-Then open http://localhost:3000 to see the app.
+Then open http://localhost:3001 to see the app.
 
 ## Running specs, linter(without auto correct) and annotate models and serializers
 ```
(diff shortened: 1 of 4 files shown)


defong:
Port number is always going to be a debacle.
Just any number that's not 3xxx is fine, better have it out here so the overall communication.

80 is http, 433 is https, 3000, is neither 3080 or 3433, so the 3xxx is out of bounds.

defong:
normally default port number are simply not used

as when you wanna take a sample out to have a look see, you screwed, particular with favicon,

first 2048 port number is out of bounds due to system needs

so pick a number and see if someone wanna fight over it

i picked mine, so stick or twist.

defong:
[Backend port 3000 to 30000](https://trello.com/c/qBGXA3hE/1036-backend-port-3000-to-30000)

defong:
To make it easier the, below port number cannot chosen.

666
3xxx
5xxx
6xxx
43xxx

The passing mark, is xxx is details. 

So a quick question of xx thousand port is used by x default app.

misaka:
Why is 3xxx out?

With 7 services running now, I agree it'd be nice to have a different standard port for each, something like:

3001 - manage-courses-ui
3002 - manage-courses-ui
3003 - search-and-compare-ui
3004 - search-and-compare-api
3005 - manage-courses-support
3006 - manage-courses-frontend
3007 - manage-courses-backend

dankmitchell:
Okay, I give in. In this instance i'd just specify the port `rails s -p 3001` because as a rails developer I always expect the app to run over port `3000`.

Can we go with `3001`, please?

defong:
I'm reluctant to make wholesale changes, and additional dependencies will be incurred. 

ie
manage-courses-ui
manage-courses-frontend

I stick with `3001` for `manage-courses-backend` if there is no more debates

misaka:
@defong Had a think about this overnight and I think 3001 for `manage-courses-backend` is the right thing to do.

PR #147, opened 2019-02-25
[953] remove references to provider_id/course_id for incremental fetch from documentation

davidgisbey:
### Context

Incremental fetch has been refactored to use a sub-second changed_since solution instead of using provider_id/course_id.

References to using provider_id/course_id have been removed from the incremental fetch section of the documentation.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -73,10 +73,8 @@ incremental update should only be performed using the next-page urls provided
 in the response headers. With that in mind, these are the parameters you
 should expect to see:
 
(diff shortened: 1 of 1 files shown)


defong:
nothing wrong just fyi 

https://github.com/DFE-Digital/manage-courses-backend/pull/142#pullrequestreview-207342183

@benilovj

misaka:
I think this should be part of #131 since that one makes the change described here.

davidgisbey:
This has now been added as a commit to #131 as . requested

PR #154, opened 2019-02-27
[1043] change course incremental api to use sub-second changed_since

misaka:
This brings it inline with the implementation of the provider incremental api
endpoint. Changes include:

- remove from_course_id param from next link header and Course scope
- format changed_since param in next link header to include micro-seconds

### Guidance to review

We want to add some more thorough tests that exercise larger page sizes and counts, it seems the existing tests didn't catch certain failures, but the implementation seems solid and can be reviewed independently of those tests.

We plan to make another PR to use `changed_at` instead of `updated_at`, which is required to record updates done to course enrichments.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -57,7 +57,7 @@ class Course < ApplicationRecord
     if timestamp.present?
       where("course.updated_at > ?", timestamp)
     else
-      where("updated_at is not null")
+      where.not(updated_at: nil)
     end.order(:updated_at, :id)
   end
 
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -53,10 +53,12 @@ class Course < ApplicationRecord
   has_many :site_statuses
   has_many :sites, through: :site_statuses
 
-  scope :changed_since, ->(datetime, from_course_id = 0) do
-    if datetime.present?
-      where("course.updated_at >= ? AND course.id > ?", datetime, from_course_id).order(:updated_at, :id)
-    end
+  scope :changed_since, ->(timestamp) do
+    if timestamp.present?
+      where("course.updated_at > ?", timestamp)
+    else
+      where("updated_at is not null")
+    end.order(:updated_at, :id)
   end
 
   def recruitment_cycle
--- a/spec/controllers/api/v1/courses_controller_spec.rb
+++ b/spec/controllers/api/v1/courses_controller_spec.rb
@@ -1,0 +1,135 @@
+require 'rails_helper'
+
+describe API::V1::CoursesController, type: :controller do
+  describe '#index' do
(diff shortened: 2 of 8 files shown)


spec/controllers/api/v1/courses_controller_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+      let(:changed_since) { 10.minutes.ago.utc }
+
+      before do
+        controller.index
+      end
+
+      describe 'returned courses in JSON' do
+        subject { response.body }
+
+        # TODO: This isn't very clear, it should be "should_not have_courses"
```
benilovj: ?? is this an issue with the matcher?

spec/models/course_spec.rb
```diff
@@ -11,8 +11,8 @@
 #  qualification           :integer          not null
 #  start_date              :datetime
 #  study_mode              :text
-#  accrediting_provider_id :integer
-#  provider_id             :integer          default(0), not null
+#  accrediting_course_id :integer
+#  course_id             :integer          default(0), not null
```
benilovj: what are these changes related to?

spec/controllers/api/v1/courses_controller_spec.rb
misaka: I think there should be some way to get the matcher to behave better in this instance, just haven't looked yet.

spec/models/course_spec.rb
misaka: I think this was brought in by a search/replace. Good spot, will regenerate the annotations.

app/models/course.rb
```diff
@@ -53,10 +53,12 @@ class Course < ApplicationRecord
...
 
-  scope :changed_since, ->(datetime, from_course_id = 0) do
-    if datetime.present?
-      where("course.updated_at >= ? AND course.id > ?", datetime, from_course_id).order(:updated_at, :id)
-    end
+  scope :changed_since, ->(timestamp) do
+    if timestamp.present?
+      where("course.updated_at > ?", timestamp)
+    else
+      where("updated_at is not null")
```
benilovj: would `where.not(updated_at: nil)` be more idiomatic?

timabell: done, tests still pass, cheers

timabell:
> I think the API docs still has a reference to `from_course_id`

done @benilovj

spec/controllers/api/v1/courses_controller_spec.rb
timabell: suggested improvement: https://github.com/DFE-Digital/manage-courses-backend/pull/154/commits/bbbf446a8bd99c36e462044442061a49d3c11918
thoughts @misaka @benilovj  ?

PR #155, opened 2019-02-28
Add courses count to v2 providers endpoint

dankmitchell:
### Context
On the frontend we display the number of courses for each provider

Add `course_count` to v2 providers endpoint

### Guidance to review
`/api/v2/providers`

### DIFF
--- a/app/serializers/api/v2/serializable_provider.rb
+++ b/app/serializers/api/v2/serializable_provider.rb
@@ -10,6 +10,10 @@ module API
       attribute :institution_name do
         @object.provider_name
       end
+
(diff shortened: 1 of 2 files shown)


misaka:
I'd like to spend a tiny bit of time to see if we can't actually do this by actually including a list of the course URLs in the results (as opposed to the entire course info) of the API call to `providers` endpoint.

dankmitchell:
[Add course count to v2 providers endpoint?](https://trello.com/c/e3qctmlG/1054-add-course-count-to-v2-providers-endpoint)

PR #156, opened 2019-02-28
Add course scope for opted in providers

dankmitchell:
### Context
The courses endpoint should only return courses where their providers are opted in

Add scope to courses endpoint

### Guidance to review
`/api/v1/courses`


dankmitchell:
[Ensure courses are scoped by providers that have opted in](https://trello.com/c/vZLCmX1k/1053-ensure-courses-are-scoped-by-providers-that-have-opted-in)

PR #162, opened 2019-03-01
[1052] documentation updated to remove reference to updated_at

timabell:
`updated_at` is already inaccurate on provider and soon will be on course

Update the docs to be more future-proof on this.

Removed hand-wavy edge cases, it was a list of what we wanted to cover, but now we have some but not all so it's become just misleading.


timabell:
[Consider course changed when course enrichment is updated.](https://trello.com/c/eCY6BVcH/1052-consider-course-changed-when-course-enrichment-is-updated)

davidgisbey:
:shipit:

PR #164, opened 2019-03-04
1065 refactor provider next link

davidgisbey:
### Context
Course controller was previously refactored to use the next link header concern. This PR replicates these changes in the provider controller.

- Provider controller refactored to use the next link header concern
- Tests for provider controller now use custom matchers like those implemented in the course controller spec.


davidgisbey:
[Refactor provider to use same next link format as course](https://trello.com/c/mCoBXYpa/1065-refactor-provider-to-use-same-next-link-format-as-course)

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -1,66 +1,146 @@
...
+      failure_message do |body|
+        <<~STRING
+          expected provider codes #{providers.map(&:provider_code)}
+            to be found in body #{provider_codes(body)}
+        STRING
+      end
     end
 
-    it 'renders a 400 when the changed_since param is not valid' do
+    RSpec::Matchers.define :have_empty_response do
```
misaka: I think that having `have_empty_response` is a bit generic (what kind of response?), and probably duplicates existing functionality (e.g. `expect(JSON.parse(response.body)).to eq []`). I know that this is just copied over from the `courses_controller_spec.rb`, I played around with that tonight and came up with something that's a bit more generic that can be used as:

```ruby
should have_courses(old_course, new_course) # existing functionality
should have_courses                         # true if any courses are found
should_not have_courses                     # true if no courses are found
```

Here's what it looks like, it would be easy to translate to providers if you think it works:

```ruby
RSpec::Matchers.define :have_courses do |*courses|
      def course_codes(body)
        json = JSON.parse(body)
        json.map { |course| course["course_code"] }
      end

      match do |body|
        if courses.any?
          course_codes(body) == courses.map(&:course_code)
        else
          course_codes(body).any?
        end
      end

      failure_message do |body|
        if courses.any?
          <<~STRING
            expected course codes #{courses.map(&:course_code)}
              to be found in body #{course_codes(body)}
          STRING
        else
          'expected courses to be present, but no courses found'
        end
      end

      failure_message_when_negated do |body|
        if courses.any?
          <<~STRING
              expected course codes #{courses.map(&:course_code)}
            not to be found in body #{course_codes(body)}
          STRING
        else
          'expected no courses to be present, but some are'
        end
      end
    end
```

misaka: If we do keep `have_empty_response` then it should probably be moved into `spec/support` and included into `spec-helper.rb`

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -1,66 +1,146 @@
 require 'rails_helper'
 
 describe API::V1::ProvidersController, type: :controller do
-  describe "index" do
-    it "calls limit on the model with default value of 100" do
-      allow(controller).to receive(:authenticate)
-      expect(Provider).to receive_message_chain(:changed_since, :limit).with(100).and_return([])
+  describe '#index' do
+    RSpec::Matchers.define :have_providers do |*providers|
```
misaka: I just realised that this is a duplication of what's in `spec/requests/api/v1/provider_spec.rb`, so it should be moved into `spec/support` and included into `spec-helper.rb`.

timabell:
rebased onto master

timabell:
rebased again, `should_not` matcher done

spec/controllers/api/v1/providers_controller_spec.rb
timabell: done the alteration for negated, left DRY for later, made a note on the trello card.

spec/controllers/api/v1/providers_controller_spec.rb
timabell: separate PR to get this out?

timabell:
[Update course-changed to use changed_at as per provider](https://trello.com/c/eCY6BVcH/1052-update-course-changed-to-use-changedat-as-per-provider)

timabell:
[encoding difference in changed_since param](https://trello.com/c/72vJswGc/1094-encoding-difference-in-changedsince-param)

PR #167, opened 2019-03-04
1083 remove provider enrichment hook

davidgisbey:
### Context

Touch provider logic needs to be ported to V2 for provider enrichment

- Remove hook from the provider enrichment model
- Remove tests for hook
- Add an association test


davidgisbey:
[Remove hook from provider enrichment to updated provider changed at from Rails (will be in c#)](https://trello.com/c/VJYYj0BO/1083-remove-hook-from-provider-enrichment-to-updated-provider-changed-at-from-rails-will-be-in-c)

PR #169, opened 2019-03-04
[1050] Add mcb

misaka:
### Context

cmdline util to help use the app db and apis

Add a cmdline helper script to perform common functions on the API. For example:

* display the data presented by the v1 api for a given provider (pages through all the data to find the provider code given)
* display data help for a provider in the db including attributes, enrichments, courses, etc

The goal is to provide access to the data and functionality in a guided way that is safer and better sign-posted than using the raw Rails console. This can be useful for people who aren't as familiar with the data, or with certain complex operations that just aren't already packaged up in the app.

### Guidance to review

This was done as a quick stab to see if it's viable and useful. We should move commands out into separate modules in `lib` or similar.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,9 @@ gem 'config'
 # Nicer URL's
 gem 'friendly_id'
 
+# For building cmdline apps (mcb)
+gem 'cri'
+
 group :development, :test do
   # add info about db structure to models and other files
   gem 'annotate'
@@ -99,6 +102,9 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+
+  # Make HTTP requests fun again
+  gem 'httparty'
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,12 @@ gem 'friendly_id'
 # For building cmdline apps (mcb)
 gem 'cri'
 
+# Build pretty tables in the terminal
+#   table_print handles ActiveRecord objects and collections really nicely
+gem 'table_print'
+#   terminal-table is a bit more flexible allowing us to use a headers column
+gem 'terminal-table'
+
 group :development, :test do
   # add info about db structure to models and other files
   gem 'annotate'
--- a/lib/mcb/commands/apiv1/find.rb
+++ b/lib/mcb/commands/apiv1/find.rb
@@ -1,0 +1,25 @@
+name 'find'
+summary 'Find a particular provider entry'
+usage 'find [options] <code>'
+param :code
+
+run do |opts, args, _cmd|
+  # We only need httparty for API V1 calls
+  require 'httparty'
+
+  puts "looking for provider #{args[:code]}"
+
+  provider = each_v1_provider(opts).detect do |p|
+    p['institution_code'] == args[:code]
+  end
+
+  if provider.nil?
+    error "Provider with code '#{code}' not found"
+  else
+    campuses = provider.delete('campuses')
+    puts Terminal::Table.new rows: provider
+    puts ''
+    puts "Campuses:"
+    tp campuses
+  end
+end
--- a/lib/mcb/commands/apiv1/find.rb
+++ b/lib/mcb/commands/apiv1/find.rb
@@ -14,7 +14,7 @@ run do |opts, args, _cmd|
   end
 
   if provider.nil?
-    error "Provider with code '#{code}' not found"
+    error "Provider with code '#{args[:code]}' not found"
   else
     campuses = provider.delete('campuses')
     puts Terminal::Table.new rows: provider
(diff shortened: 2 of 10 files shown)


lib/mcb/commands/apiv1/find.rb
```diff
@@ -0,0 +1,25 @@
...
+  require 'httparty'
+
+  puts "looking for provider #{args[:code]}"
+
+  provider = each_v1_provider(opts).detect do |p|
+    p['institution_code'] == args[:code]
+  end
+
+  if provider.nil?
+    error "Provider with code '#{code}' not found"
```
heidar: `code` is not defined here

lib/mcb/commands/apiv1/find.rb
```diff
@@ -0,0 +1,25 @@
+name 'find'
```
heidar: I can't seem to get this one to work. What argument does it take? Provider code?

mcb
```diff
@@ -0,0 +1,99 @@
...
+
+MCB_ROOT_DIR = File.dirname($0)
+def init_rails
+  unless defined?(Rails)
+    exec(File.join(MCB_ROOT_DIR, 'bin', 'rails'), 'runner', $0, *ARGV)
+  end
+end
+
+
+def verbose(msg)
```
heidar: Can we move this to a helper file that we require?

lib/mcb/commands/apiv1/find.rb
```diff
@@ -0,0 +1,25 @@
...
+  # We only need httparty for API V1 calls
+  require 'httparty'
+
+  puts "looking for provider #{args[:code]}"
+
+  provider = each_v1_provider(opts).detect do |p|
+    p['institution_code'] == args[:code]
+  end
+
+  if provider.nil?
```
heidar: Can this be a guard clause?

mcb
```diff
@@ -0,0 +1,99 @@
...
+
+      verbose "Requesting page #{page_count + 1}: #{url}"
+      response = HTTParty.get(
+        url,
+        headers: {
+          authorization: "Bearer #{opts[:token]}"
+        }
+      )
+      providers_list = JSON.parse(response.body)
+      break if providers_list == []
```
heidar: Can we use `Array#empty?` here?

mcb
```diff
@@ -0,0 +1,99 @@
...
+  if $verbosity == :verbose
+    puts msg
+  end
+end
+
+def error(msg)
+  puts msg
+end
+
+def each_v1_provider(opts)
```
heidar: Would be nice to break this up a bit as it's a bit long maybe?

lib/mcb/commands/apiv1/find.rb
misaka: Should be `args[:code]`

Gemfile
```diff
@@ -97,6 +106,9 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+
+  # Make HTTP requests fun again
+  gem 'httparty'
```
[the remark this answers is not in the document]
misaka: We don't use Faraday directly, it's just an artifact of using jsonapi-rb, so I'm not sure it would add much to synchronize here. I just wanted something that let me set headers easily.

lib/mcb/commands/apiv1/find.rb
misaka: It assumes you have `manage-courses-backend` running locally on port 3001, and a db populated with providers. If you had that setup, what was your error?

lib/mcb/commands/apiv1/find.rb
misaka: Well, it's a block so we could do:

```ruby
  if provider.nil?
    error "Provider with code '#{args[:code]}' not found"
    next
  end
```

`next` doesn't feel very idiomatic, but it'd get the job done.

mcb
misaka: Good catch, I blame my copy-paste source for this one! ;)

mcb
misaka: I think this method is borderline, not sure it's worth the effort.

mcb
misaka: I'd look to move things out of `mcb` in a future iteration, when we see if it gets any uptake.

mcb
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
misaka: Ya, that's reasonable.

misaka:
@heidar I'm keen to not spend too much time on this as it's unproven whether it'll prove much value. Could you please let me know if any of the remaining issues are must-haves?

lib/mcb/commands/apiv1/find.rb
heidar: I got the error in the error message before, now that you have fixed it I get:

```
❯ bin/mcb apiv1 find A1
looking for provider A1
Provider with code 'A1' not found
```

misaka: Do you have a provider with the code `A1` in your DB? All the provider codes in the Real Data (tm) have 3 characters, did you add one with 'A1'?

PR #170, opened 2019-03-05
[1087] Course description

benilovj:
### Context
Courses have a consistently worded description (e.g. `PGCE with QTS full time with salary`) which are used in many places in the service. 

Add a `description` attribute to the `Course` model and expose it in the Courses API V2 for frontend consumption (most immediately in the courses table in Publish).

### Guidance to review
I debated whether the course description is a presentation or a backend concern. My conclusion is that this is a backend concern because:

- it's tightly coupled to the domain
- there is only one representation of this across the entire service
- it needs to be consistent across Publish and Find and hence is part of publishing

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -78,4 +78,8 @@ class Course < ApplicationRecord
   def has_vacancies?
     site_statuses.with_vacancies.any?
   end
+
+  def study_mode_description
+    study_mode.to_s.tr("_", " ")
+  end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -82,4 +82,17 @@ class Course < ApplicationRecord
   def study_mode_description
     study_mode.to_s.tr("_", " ")
   end
+
+  def program_type_description
+    if school_direct_salaried_training_programme? then " with salary"
+    elsif pg_teaching_apprenticeship? then " teaching apprenticeship"
+    else ""
+    end
+  end
+
+  def description
+    study_mode_string = (full_time_or_part_time? ? ", " : " ") +
+      study_mode_description
+    qualifications_description + study_mode_string + program_type_description
+  end
 end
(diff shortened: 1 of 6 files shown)


app/models/course.rb
```diff
@@ -78,4 +78,16 @@ def open_for_applications?
   def has_vacancies?
     site_statuses.with_vacancies.any?
   end
+
+  def study_mode_description
+    study_mode.to_s.tr("_", " ")
+  end
+
+  def description
```
dankmitchell: It is worth breaking this down? It's really hard to read...

benilovj: I've taken a stab at making it more readable, have another look.

app/models/course.rb
```diff
@@ -78,4 +78,19 @@ def open_for_applications?
   def has_vacancies?
     site_statuses.with_vacancies.any?
   end
+
+  def study_mode_description
+    study_mode.to_s.tr("_", " ")
```
tvararu: Minor suggestion: part of me would prefer a more explicit `case` here in which we opt into the strings that we want to translate, rather than automating it with `tr`, and set a `default` for unexpected cases. But this is more pragmatic and should work just fine.

spec/models/course_spec.rb
```diff
@@ -196,6 +196,56 @@
...
+
+    specs.each do |study_mode, expected_description|
+      context study_mode.to_s do
+        subject { create(:course, study_mode: study_mode) }
+        its(:study_mode_description) { should eq(expected_description) }
+      end
+    end
+  end
+
+  describe "#description" do
```
defong: missing combo?
image

defong: There is test for 
`PGDE with QTS full time with salary`
`PGCE with QTS full time with salary`

but not 
`QTS full time with salary` (ie the pic above)

but in any case no reasons to hold this PR back to be honesty

was a simple sanity check

PR #171, opened 2019-03-05
Re-add pg_* extensions to schema.rb

timabell:
These extensions are present in a recent pg_dump from production,
so they should be here too.

### Context
They re-appeared on my machine, seems everyone is in a different state. Realised that I'm running from a sanitized restore of prod, which brings the enabled extensions across.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -13,6 +13,8 @@
 ActiveRecord::Schema.define(version: 2019_03_06_122110) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "pg_buffercache"
+  enable_extension "pg_stat_statements"
   enable_extension "plpgsql"
 
   create_table "__EFMigrationsHistory", primary_key: "MigrationId", id: :string, limit: 150, force: :cascade do |t|


misaka:
So adding this in, and then doing a `db:schema:dump` took them back out. Do you know offhand what we need to do to install those extensions?

misaka:
Should probably include those instructions here either way.

timabell:
[EPIC Implement incremental fetch for courses. - Rails](https://trello.com/c/DRtQR8bN/850-epic-implement-incremental-fetch-for-courses-rails)

heidar:
Having looked at what these do, I don't think we need either of these? `pg_buffercache` seems to be an extension that has to be enabled on the server using `CREATE EXTENSION`, assuming the module was compiled. 

`pg_stat_statements` is a setting in the config of the server and I think it will be disabled if it is not enabled on the server. Changing this setting in the Postgres config requires a restart to take effect.

I think these are differences in the settings for our hosted Postgres instance. We could make our local instances match and merge this, or change the production one to not have these, unless we need them.

timabell:
@heidar this PR is based on a dump from prod, so brings the schema rb into line with prod.
I agree we might want to disable them in prod but didn't want take that task on right now.

timabell:
I've done some experimenting on mine machine and david's mac. It seems the extensions are controlled by the schema.rb and/or the sql dump. So if your `rails db:schema:dump` is causing a diff then you first need to check out master (or this branch if it's not merged), then run `rails db:reset` to bring your db into line, and then the dump will cease to generate a diff.
I can't think what I could add to the readme that would be useful in the long run once everyone has done a reset.

heidar:
Makes sense!

I tried resetting and dumping on this branch and didn't get a diff, despite my server config not having enabled `pg_stat_statements`. I guess the module is loaded, which is enough.

PR #172, opened 2019-03-06
[1052] Migration to set course.changed_at from updated_at

timabell:
### Context
Paging doesn't work without unique values.

Make values unique as per updated_at

### Guidance to review
:ship:


timabell:
[Update course-changed to use changed_at as per provider](https://trello.com/c/eCY6BVcH/1052-update-course-changed-to-use-changedat-as-per-provider)

PR #174, opened 2019-03-06
[1105] Return providers in alphabetical order

tvararu:
This is the ordering that the frontend needs to display them correctly.

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -8,7 +8,7 @@ module API
         providers = policy_scope(Provider)
         providers = providers.where(id: @user.providers) if @user.present?
 
-        render jsonapi: providers
+        render jsonapi: providers.in_order
       end
 
     private
--- a/spec/requests/api/v2/providers_spec.rb
+++ b/spec/requests/api/v2/providers_spec.rb
@@ -106,5 +106,22 @@ describe 'Providers API v2', type: :request do
         )
       end
     end
+
+    context 'with unalphabetical ordering in the database' do
+      let!(:second_alphabetical_provider) { create(:provider, provider_name: 'Zork', organisations: [organisation]) }
+
+      before do
+        provider.update(provider_name: 'Acme') # This moves it last in the order that it gets fetched by default.
+        get "/api/v2/users/#{user.id}/providers", headers: { 'HTTP_AUTHORIZATION' => credentials }
+      end
+
+      let(:provider_names_in_response) {
+        JSON.parse(response.body)["data"].map { |provider| provider["attributes"]["institution_name"] }
+      }
+
+      it 'returns them in alphabetical order' do
+        expect(provider_names_in_response).to eq(%w(Acme Zork))
+      end
+    end
   end
 end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -59,6 +59,8 @@ class Provider < ApplicationRecord
 
   scope :opted_in, -> { where(opted_in: true) }
 
+  scope :in_order, -> { order(:provider_name) }
+
   # TODO: filter to published enrichments, maybe rename to published_address_info
   def address_info
     (enrichments.latest_created_at.with_address_info.first || self)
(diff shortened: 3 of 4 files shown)


spec/requests/api/v2/providers_spec.rb
```diff
@@ -106,5 +106,18 @@
...
+    context 'with unalphabetical ordering in the database' do
+      let!(:second_alphabetical_provider) { create(:provider, provider_name: 'Zork', course_count: 0, site_count: 0, organisations: [organisation]) }
+      let!(:first_alphabetical_provider) { create(:provider, provider_name: 'Acme', course_count: 0, site_count: 0, organisations: [organisation]) }
+
+      before do
+        get "/api/v2/users/#{user.id}/providers", headers: { 'HTTP_AUTHORIZATION' => credentials }
+      end
+
+      it 'returns them in alphabetical order' do
+        expect(response.body).to match(/Acme(.*)Zork/)
```
tvararu: ¯\\\_(ツ)\_\/¯ not sure if this is the best way to match but it works?

tvararu:
[Sort providers in manage-courses-frontend alphabetically](https://trello.com/c/off65rte/1105-sort-providers-in-manage-courses-frontend-alphabetically)

spec/requests/api/v2/providers_spec.rb
benilovj: Alternatively:

```ruby
    context 'with unalphabetical ordering in the database' do
      let!(:first_alphabetical_provider) { create(:provider, provider_name: 'Acme', organisations: [organisation]) }

      before do
        provider.update(provider_name: 'Zork')
        get "/api/v2/users/#{user.id}/providers", headers: { 'HTTP_AUTHORIZATION' => credentials }
      end

      let(:provider_names_in_response) {
        JSON.parse(response.body)["data"].
          map {|provider| provider["attributes"]["institution_name"] }
      }

      it 'returns them in alphabetical order' do
        expect(provider_names_in_response).to eq(['Acme', 'Zork'])
      end
    end
```

app/controllers/api/v2/providers_controller.rb
```diff
@@ -6,7 +6,7 @@ class ProvidersController < ApplicationController
       def index
         authorize Provider
         providers = policy_scope(Provider)
-        providers = providers.where(id: @user.providers) if @user.present?
+        providers = providers.where(id: @user.providers).order(:provider_name) if @user.present?
```
benilovj: Placing the `order` on this line means that if `@user` isn't specified, the results are unordered...

app/controllers/api/v2/providers_controller.rb
```diff
@@ -6,7 +6,7 @@ class ProvidersController < ApplicationController
...
 
         render jsonapi: providers
```
dankmitchell: Can we create a scope? `render jsonapi: providers.in_order`

spec/requests/api/v2/providers_spec.rb
```diff
@@ -106,5 +106,18 @@
         )
       end
     end
+
+    context 'with unalphabetical ordering in the database' do
```
dankmitchell: Spec can just be `context '. in_order' do`

tvararu: It's a request/integration spec though, shouldn't it be more high-level in its verbiage, present the specs in terms closer to an actual user journey?

spec/requests/api/v2/providers_spec.rb
tvararu: A+

tvararu:
@misaka I've investigated and there's nothing built into `jsonapi-rb` or `jsonapi-rails` that would enable sorting out of the box. [JR supports sortable attributes](http://jsonapi-resources.com/v0.9/guide/resources.html#Sortable-Attributes) but it would be an entire new library to add and I'm not sure it's needed.

spec/requests/api/v2/providers_spec.rb
tvararu: Sorted, have added a model spec for the scope 👍

PR #175, opened 2019-03-07
1109 re add verify authorized

misaka:
### Context

Controller actions should be checking that authorisations has been performed to make sure we don't accidentally create controller actions that don't authorize.

Re-add this authorizaton verification, it was removed accidentally in #124.

### Guidance to review

Now that we-ve re-added authorizaton verification we need to explicitly skip
authorization for `SessionsController#create`.


timabell:
[Re-add check that authorisation is performed.](https://trello.com/c/aet1PT3b/1109-re-add-check-that-authorisation-is-performed)

PR #177, opened 2019-03-07
1101 refactor matcher logic

davidgisbey:
Context
There was duplication between the custom matchers in the request and controller spec files, for provider and courses.

Changes proposed in this pull request
For both course and provider:

Matchers moved into a matcher folder
Matcher refactored
Tests for pagination  refactored to use objects rather than strings


spec/controllers/api/v1/courses_controller_spec.rb
```diff
@@ -133,7 +96,7 @@ def format_timestamp(timestamp)
       describe 'returned courses in JSON' do
         subject { response.body }
 
-        it { should_not have_courses }
+        it { should have_courses([]) }
```
timabell: me no likey

defong: should be_empty ?

davidgisbey:
Additional commit added to use changed_at rather than updated_at in the pagination tests. Interestingly if you create an object using a timestamp with strftime on the updated_at or created_at columns it returns objects with microsecond differences. 

However, if you do this with the changed_at column is creates x objects with an identical timestamp. To  get around this we've used a 1/1000th second difference between the objects being created.

spec/controllers/api/v1/courses_controller_spec.rb
misaka: Why was `should_not have_courses` removed?

spec/controllers/api/v1/providers_controller_spec.rb
```diff
@@ -133,7 +96,7 @@ def format_timestamp(timestamp)
       describe 'returned providers in JSON' do
         subject { response.body }
 
-        it { should_not have_providers }
+        it { should have_providers([]) }
```
misaka: Why no `should_not have_providers`?

misaka:
> Additional commit added to use changed_at rather than updated_at in the pagination tests. Interestingly if you create an object using a timestamp with strftime on the updated_at or created_at columns it returns objects with microsecond differences.
>
> However, if you do this with the changed_at column is creates x objects with an identical timestamp. To get around this we've used a 1/1000th second difference between the objects being created.

That's interesting. I'm going to guess that that FactoryBot does special handling of `created_at` and `updated_at`. This would benefit from Tim's idea to make sure `changed_at` is unique with a unique index.

Also, the 1/1000th second bit is also interesting. I guess this is the issue @defong wrestled with for a bit, it may mean that extending the timestamp used in the HTTP header to be 6 digits is overdoing it, and 3 digits is all we can (easily) support.

spec/controllers/api/v1/courses_controller_spec.rb
davidgisbey: Have rewrote the logic so should_not works 👍

spec/controllers/api/v1/providers_controller_spec.rb
davidgisbey: Done 👍

spec/controllers/api/v1/courses_controller_spec.rb
defong: :rofl: should_not work :roll_eyes:

davidgisbey:
[Refactor controller spec matcher logic into spec helper](https://trello.com/c/M5O2d71f/1101-refactor-controller-spec-matcher-logic-into-spec-helper)

spec/support/matchers/have_providers.rb
```diff
@@ -0,0 +1,39 @@
+RSpec::Matchers.define :have_providers do |*providers|
+  providers = providers.flatten
```
heidar: You could use `Array#flatten!` to flatten in place here. Same in the other matcher.

PR #183, opened 2019-03-11
[1126] use fully-qualified ApplicationController class

misaka:
Trello ticket: https://trello.com/c/AiWztw9a/1126-tests-failing-when-run-in-a-certain-order

We have more than one `ApplicationController` class:

- ::ApplicationController
- ::API::V1::ApplicationController
- ::API::V2::ApplicationController

Because of this, Rails can get confused when trying to autoload constants. For
example:

  module API
    module V1

could pull in the wrong ApplicationController. At least, this has been the case
with RSpec tests where running the V2 specs before running the V1 specs were
causing most V1 specs to fail. This change fully quaifies ApplicationController
when used as a base class like this.

### DIFF
--- a/app/controllers/api/v1/courses_controller.rb
+++ b/app/controllers/api/v1/courses_controller.rb
@@ -1,6 +1,6 @@
 module API
   module V1
-    class CoursesController < ApplicationController
+    class CoursesController < API::V1::ApplicationController
       include NextLinkHeader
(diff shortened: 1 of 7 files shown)


misaka:
[Tests failing when run in a certain order](https://trello.com/c/AiWztw9a/1126-tests-failing-when-run-in-a-certain-order)

PR #185, opened 2019-03-12
[1127] Make provider lookup be case insensitive

tvararu:
Always uppercase the parameter for the database lookup as that is the canonical name.


tvararu:
[Organisation lookup is case sensitive on manage-courses-backend](https://trello.com/c/XKUHzUQP/1127-organisation-lookup-is-case-sensitive-on-manage-courses-backend)

PR #192, opened 2019-03-13
[1150] Handle user email matching case-insensitively

benilovj:
### Context

In line with the API design philosophy of "be permissive in what you accept and strict with what you return".

This change allows the V2 API to authenticate successfully in cases where the email address in the database (which is all lowercase by convention) doesn't match a mixed-case email coming through on the Authorization header from DfE Sign-in.

### DIFF
--- a/app/controllers/api/v2/application_controller.rb
+++ b/app/controllers/api/v2/application_controller.rb
@@ -17,7 +17,8 @@ module API
                                                  Settings.authentication.algorithm)
             email = payload['email']
           end
-          @current_user = User.find_by(email: email)
+          # match email addresses case-insensitively
+          @current_user = User.find_by("lower(email) = ?", email.downcase)
           @current_user.present?
         end
       end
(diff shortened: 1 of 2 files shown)


defong:
[DfE Sign-in authentication when there is an email address case mismatch](https://trello.com/c/LDQJDTcy/1150-dfe-sign-in-authentication-when-there-is-an-email-address-case-mismatch)

app/controllers/api/v2/application_controller.rb
```diff
@@ -17,7 +17,8 @@ def authenticate
                                                  Settings.authentication.algorithm)
             email = payload['email']
           end
-          @current_user = User.find_by(email: email)
+          # match email addresses case-insensitively
+          @current_user = User.find_by("lower(email) = ?", email.downcase)
```
misaka: Do we know if this will sidestep the index we have on the `email` column? I haven't checked yet but maybe if we can normalise emails in the db to be lower-case we can avoid running `lower(email)`?

misaka: Agreed with Jake that we'll look at addressing this later, as we can't yet ensure 100% that all emails entered in the DB are lower-case (because of multiple input sources).

benilovj: Reminder card created: https://trello.com/c/pZx8vqUM/212-dont-break-the-index-when-looking-for-users-in-the-db

PR #196, opened 2019-03-15
1131 site status validation

defong:
### Context
Validation for site status when updating. 

Added validation for site status.
Added skip validation traits for `course` & `site_status`
Seeding of database is more correct now

### Guidance to review
Ignore more or less everything... that's with `:skip_validate`

Just look at
- app/models/site_status.rb
- spec/factories/courses.rb
- spec/factories/site_statuses.rb
- spec/models/site_status_spec.rb
- db/seeds.rd
 

`:skip_validate` 

1. added to the existing tests as a stop gap as previous tests setup has no consideration of correctiveness of data context
2. nice to have in cases where we want to mock out true values stat is in database

The only redeeming thing about ruby/rails is really factorybot...

### DIFF
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -31,6 +31,10 @@ FactoryBot.define do
     study_mode { :full_time }
     resulting_in_pgce_with_qts
 
+    trait :skips_validate do
+      to_create { |instance| instance.save(validate: false) }
+    end
+
     transient do
       age { nil }
     end
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -30,7 +30,8 @@ FactoryBot.define do
     association(:provider, course_count: 0)
     study_mode { :full_time }
     resulting_in_pgce_with_qts
-    study_mode { :full_time_or_part_time }
+
+    with_study_mode_as_full_time
 
     trait :skips_validate do
       to_create { |instance| instance.save(validate: false) }
@@ -66,6 +67,40 @@ FactoryBot.define do
 
     trait :resulting_in_pgde do
       qualification { :pgde }
+    end
+
+    trait :with_study_mode_as_full_time do
+      study_mode { :full_time }
+    end
+
+    trait :with_study_mode_as_part_time do
+      study_mode { :part_time }
+    end
+
+    trait :with_study_mode_as_full_time_or_part_time do
+      study_mode { :full_time_or_part_time }
+    end
+
+    factory :course_with_site_statuses do
+      with_study_mode_as_full_time_or_part_time
+
+      transient do
+        site_statuses_traits {
+          [
+            [:findable],
+            [:with_any_vacancy],
+            [],
+            [:applications_being_accepted_now],
+            [:applications_being_accepted_in_future]
+          ]
+        }
+      end
+
+      after(:create) do |course, evaluator|
+        evaluator.site_statuses_traits.each do |traits|
+          create(:site_status, *traits, course: course)
+        end
+      end
     end
   end
 end
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -30,6 +30,7 @@ FactoryBot.define do
     association(:provider, course_count: 0)
     study_mode { :full_time }
     resulting_in_pgce_with_qts
+    study_mode { :full_time_or_part_time }
 
     trait :skips_validate do
       to_create { |instance| instance.save(validate: false) }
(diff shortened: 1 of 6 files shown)


defong:

benilovj:
Based on production stats (in terms of frequencies of the options), the factories should:

- default `course#study_mode` to full-time
- default `site_status#vac_status` to full-time

This combination is valid.

defong:
> I don't think that it's right to pollute the specs with `skips_validate`. Why can't we change the factories to produce valid constellations of data?

`skips_validate` was just my rabbit holing prevention before I embark on what is `valid constellations of data`


entering rabbit hole...
based on production
course#study_mode there is no `nil`

So a course can only be 1 of 3 study_mode
A course can have a study_mode of full_time
as well as its site_status#vac_status as part-time

misaka:
The factory should just worry about setting a valid combination, as Jake pasted, so why not just have this in the `course` factory:

```
  study_mode { :full_time }
```

and this in the `site_status` factory:

```
  vac_status { :full_time_vacancies }
```

You already have tests around the validations and we have a [ticket about cleaning up the existing bad data](https://trello.com/c/MTdYyHEA/1156-clean-up-inconsistent-vacancy-statuses).

spec/models/site_status_spec.rb
```diff
@@ -117,4 +117,60 @@
...
+        study_mode: :with_course_study_mode_as_part_time,
+        valid_states: %w[no_vacancies part_time_vacancies]
+      },
+      {
+        study_mode: :with_course_study_mode_as_full_time_or_part_time,
+        valid_states: all_vac_statuses
+      },
+    ].freeze
+
+    specs.each do |spec|
```
misaka: In Rails-speak what you're testing here is that an object is valid with certain combinations of values. I think these tests would be clearer if focused on that and didn't worry so much about making them DRY. This would be better, with 

```
  describe 'vac_status' do
    context 'full time course' do
      let(:course) { build(:course, study_mode: :full_time) }

      context 'vac_status set to full_time' do
        subject { create(:site_status, vac_status: :full_time_vacancies, course: course) }
        it { should be_valid }
      end

      context 'vac_status set to part_time' do
        subject { create(:site_status, :allow_invalid, vac_status: :part_time_vacancies, course: course) }
        it { should_not be_valid }
      end
    end
```

If we really want to DRY them up, I think a custom matcher would be better. It breaks the one-expectation per test guideline, but it is still readable:

```
    describe 'vac_status' do
      RSpec::Matchers.define :be_valid_vac_status_for_course do |course|
        match do |vac_status|
          site_status = create(:site_status,
                               :allow_invalid,
                               vac_status: vac_status,
                               course: course)
          site_status.valid?
        end
      end

      context 'full time course' do
        let(:full_time_course) { create(:course, study_mode: :full_time) }

        it 'only allows full_time or no vacancies' do
          expect(:no_vacancies).to be_valid_vac_status_for_course(full_time_course)
          expect(:full_time_vacancies).to be_valid_vac_status_for_course(full_time_course)
          expect(:part_time_vacancies).not_to be_valid_vac_status_for_course(full_time_course)
          expect(:both_full_time_and_part_time_vacancies).not_to be_valid_vac_status_for_course(full_time_course)
        end
      end
    end
```

Here is the `allow_invalid` trait for `site_status`. I was a little surprised we need it, but I guess it makes sense for factories to insist that the objects they create are valid, and it might make sense to bypass the factory to create invalid objects in model tests.

```
    trait :allow_invalid do
      to_create { |instance| instance.save }
    end
```

spec/factories/courses.rb
```diff
@@ -62,5 +68,39 @@
...
+      study_mode { :full_time_or_part_time }
+    end
+
+    factory :course_with_site_statuses do
+      with_study_mode_as_full_time_or_part_time
+
+      transient do
+        site_statuses_traits {
+          [
+            [:findable],
```
heidar: How come these need to be arrays? Can we not make them just symbols and not splat it when we call the factory?

spec/models/site_status_spec.rb
```diff
@@ -117,4 +117,49 @@
...
+      context "#{spec[:study_mode].to_s.humanize(capitalize: false)} course" do
+        let(:course) { build(:course, study_mode: spec[:course_study_mode]) }
+
+        spec[:valid_states].each do |state|
+          context "vac_status set to #{state}" do
+            subject { create(:site_status, vac_status: state, course: course) }
```
heidar: Do we have to `create` here? Can we not just `build`? Same below on line 158.

benilovj:
Alternative WIP approach here: https://github.com/DFE-Digital/manage-courses-backend/pull/202

heidar:
Closing in favour of #202

PR #200, opened 2019-03-18
1131 add policy for updating course

defong:
### Context
policy for updating course

Added policy for updating course

### Guidance to review
As long as the user has access to the course, for example to view it, then inherently they can update it as well.

### DIFF
--- a/spec/policies/course_policy_spec.rb
+++ b/spec/policies/course_policy_spec.rb
@@ -21,15 +21,11 @@ describe CoursePolicy do
               organisations: [organisation])
     }
 
-    it 'permits when the user belongs to the organisation' do
-      should permit(user, course)
-    end
+    it { should permit(user, course) }
 
     context 'with a user outside the organisation' do
       let(:other_user) { create(:user) }
-      it 'does not permit' do
-        should_not permit(other_user, course)
-      end
+      it { should_not permit(other_user, course) }
     end
   end
 end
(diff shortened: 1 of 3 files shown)


defong:

spec/policies/course_policy_spec.rb
```diff
@@ -35,6 +39,10 @@
       it 'allows the :show action' do
         expect(subject.show?).to be_falsey
       end
+
+      it 'allows the :update action' do
```
misaka: The words aren't quite right here.

spec/policies/course_policy_spec.rb
```diff
@@ -27,6 +27,10 @@
       it 'allows the :show action' do
         expect(subject.show?).to be_truthy
       end
+
+      it 'allows the :update action' do
+        expect(subject.update?).to be_truthy
+      end
     end
 
     context 'when accessing a course of a provider that the user does not have' do
```
misaka: Just realised, there are custom matchers to test policies:

https://github.com/varvet/pundit#policy-specs

spec/policies/course_policy_spec.rb
defong: allows the :show action
Vs 
allows the :update action
Vs
... 

Not sure what's you mean...

defong: https://github.com/varvet/pundit/blob/master/spec/policies/post_policy_spec.rb


Actually the custom mishy matcher will get rid of my bad English

defong: Guess I need to look at this as well.

https://github.com/varvet/pundit/blob/master/spec/spec_helper.rb

Cheers misha

PR #202, opened 2019-03-19
[1131] Alternative implementation of site status validation

benilovj:
### Context
An alternative approach to #196.

Add validation to site statuses and fix the specs/factories that break as a result.

### DIFF
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -14,6 +14,9 @@
 class SiteStatus < ApplicationRecord
   self.table_name = "course_site"
 
+  validate :vac_status_must_be_consistent_with_course_study_mode,
+    if: Proc.new { |s| s.course&.study_mode.present? }
+
   enum vac_status: {
     both_full_time_and_part_time_vacancies: "B",
     part_time_vacancies: "P",
@@ -43,4 +46,27 @@ class SiteStatus < ApplicationRecord
   }
   scope :with_vacancies, -> { where.not(vac_status: :no_vacancies) }
   scope :open_for_applications, -> { findable.applications_being_accepted_now.with_vacancies }
+
+private
+
+  def vac_status_must_be_consistent_with_course_study_mode
+    unless vac_status_consistent_with_course_study_mode?
+      errors.add(:vac_status, "(#{vac_status}) must be consistent with course study mode #{course.study_mode}")
+    end
+  end
+
+  def vac_status_consistent_with_course_study_mode?
+    case vac_status
+    when 'no_vacancies'
+      true
+    when 'full_time_vacancies'
+      course.full_time? || course.full_time_or_part_time?
+    when 'part_time_vacancies'
+      course.part_time? || course.full_time_or_part_time?
+    when 'both_full_time_and_part_time_vacancies'
+      course.full_time_or_part_time?
+    else
+      false
+    end
+  end
 end
--- a/spec/factories/site_statuses.rb
+++ b/spec/factories/site_statuses.rb
@@ -13,7 +13,7 @@
 
 FactoryBot.define do
   factory :site_status do
-    association(:course)
+    association :course, study_mode: :full_time_or_part_time
     association(:site)
     publish { 'N' }
     vac_status { :full_time_vacancies }
--- a/spec/factories/site_statuses.rb
+++ b/spec/factories/site_statuses.rb
@@ -18,6 +18,10 @@ FactoryBot.define do
     publish { 'N' }
     vac_status { :full_time_vacancies }
 
+    transient do
+      any_vancancy { false }
+    end
+
     trait :published do
       publish { :published }
     end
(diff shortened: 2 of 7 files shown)


benilovj:
I've still got to:

1. add specs for the validation implementation itself
2. fix the failures in `spec/models/course_spec.rb` where site statuses are being created before the course.

app/models/site_status.rb
```diff
@@ -43,4 +49,20 @@ class SiteStatus < ApplicationRecord
   }
   scope :with_vacancies, -> { where.not(vac_status: :no_vacancies) }
   scope :open_for_applications, -> { findable.applications_being_accepted_now.with_vacancies }
+
+private
+
+  def vac_status_consistent_with_course_study_mode?
+    if no_vacancies?
```
dankmitchell: A case statement might be easier to read...

benilovj: The linter objects to case statements without a subject, unfortunately.

In any case, we think we're gonna back out the actual validation from this PR and just take the spec changes forward, so #196 will have the actual code.

app/models/site_status.rb
```diff
@@ -14,6 +14,12 @@
 class SiteStatus < ApplicationRecord
   self.table_name = "course_site"
 
+  validate if: Proc.new { |s| s.course&.study_mode.present? } do
+    unless vac_status_consistent_with_course_study_mode?
+      errors.add(:vac_status, "(#{vac_status}) must be consistent with course study mode #{course.study_mode}")
+    end
+  end
```
misaka: I was trying to say earlier before my daughter a bit rudely interrupted 🤢 ... it'd be nice to move the body of this to a method. I know it already uses `vac_status_consistent_with_course_study_mode?`, which is good, but I think it be cleaner to keep this part of the model free of code (esp multi-line block) and just have declarations here.

Worth considering if you're going to move this to the other PR.

spec/factories/site_statuses.rb
```diff
@@ -13,11 +13,15 @@
 
 FactoryBot.define do
   factory :site_status do
-    association(:course)
+    association :course, study_mode: :full_time_or_part_time
     association(:site)
     publish { 'N' }
     vac_status { :full_time_vacancies }
```
misaka: This status isn't always going to be valid, why not do the `with_any_vacancy` logic by default?

app/models/site_status.rb
dankmitchell: @benilovj could you just use `case vac_status: no_vacancies` instead?

misaka: Dunno if it's clear, but I think @dankmitchell means:

```
case vac_status
when 'no_vacancies'
  ...
```

dankmitchell: Thanks @misaka 💯

benilovj: Fixed.

app/models/site_status.rb
benilovj: Done.

spec/factories/site_statuses.rb
benilovj: I think a static default value is simpler as it bypasses the `after(:build)` hook and is therefore less magic. Plus a bunch of stuff breaks if I make `with_any_vacancy` the default, so it didn't feel like the right time to switch this.

benilovj:
I can leave the validation implementation in or take it out – happy to be steered by the reviewers.

app/models/site_status.rb
```diff
@@ -43,4 +46,27 @@ class SiteStatus < ApplicationRecord
...
+
+private
+
+  def vac_status_must_be_consistent_with_course_study_mode
+    unless vac_status_consistent_with_course_study_mode?
+      errors.add(:vac_status, "(#{vac_status}) must be consistent with course study mode #{course.study_mode}")
+    end
+  end
+
+  def vac_status_consistent_with_course_study_mode?
```
dankmitchell: Nice! We’ve got this same method on my frontend PR so good to have here instead 👌

app/models/site_status.rb
```diff
@@ -43,4 +46,27 @@ class SiteStatus < ApplicationRecord
   }
   scope :with_vacancies, -> { where.not(vac_status: :no_vacancies) }
   scope :open_for_applications, -> { findable.applications_being_accepted_now.with_vacancies }
+
+private
+
+  def vac_status_must_be_consistent_with_course_study_mode
+    unless vac_status_consistent_with_course_study_mode?
+      errors.add(:vac_status, "(#{vac_status}) must be consistent with course study mode #{course.study_mode}")
```
defong: Think it's needs message

errors.add(:vac_status,
                 :invalid,
                 message:

app/models/site_status.rb
heidar: Let's keep this for consistency then

app/models/site_status.rb
heidar: As it is works fine, but you highlighted that the tests don't check for the validation error, which is really important!

defong:
[Site status validation](https://trello.com/c/fKdscaJb/1187-site-status-validation)

PR #203, opened 2019-03-19
[1165] Check for accept_terms_date_utc on every request

tvararu:
### Context

We need to return a 403 Forbidden if the user hasn't accepted the terms and conditions. The frontend  can then use this to redirect them to manage-courses-ui so that they can choose  to accept.

Refactors the email retrieval logic and then adds the check for the `accept_terms_date_utc` field.

### Guidance to review

~Not sure if this is where this bit of logic should go; I've thought about creating another `before_action` but then I need to get the token again, which is cumbersome. It's easier to insert the new logic into this existing check.~

### DIFF
--- a/app/controllers/api/v2/application_controller.rb
+++ b/app/controllers/api/v2/application_controller.rb
@@ -3,23 +3,47 @@ module API
     class ApplicationController < ::ApplicationController
       attr_reader :current_user
 
+      before_action :check_terms_accepted
+
       def authenticate
         authenticate_or_request_with_http_token do |token|
-          if Settings.authentication.algorithm == 'plain-text'
-            # This method can be used in development mode to simplify querying
-            # the API with curl. It should allow us to do:
-            #
-            #    curl -H 'Authorization: Bearer user@education.gov.uk' http://localhost:3001/api/v2/providers
-            email = token
-          else
-            (payload, _options) = JWT.decode(token,
-                                                 Settings.authentication.secret,
-                                                 Settings.authentication.algorithm)
-            email = payload['email']
-          end
-          # match email addresses case-insensitively
+          email = email_from_token(token)
+
           @current_user = User.find_by("lower(email) = ?", email.downcase)
+
           @current_user.present?
+        end
+      end
+
+    private
+
+      def check_terms_accepted
+        return if @current_user&.accept_terms_date_utc.present?
+
+        error_body = {
+          errors: [
+            {
+              status: 403,
+              title: 'Forbidden',
+              detail: 'The user has not accepted terms and conditions.'
+            }
+          ]
+        }
+        render json: error_body, status: :forbidden
+      end
+
+      def email_from_token(token)
+        if Settings.authentication.algorithm == 'plain-text'
+          # This method can be used in development mode to simplify querying
+          # the API with curl. It should allow us to do:
+          #
+          #    curl -H 'Authorization: Bearer user@education.gov.uk' http://localhost:3001/api/v2/providers
+          token
+        else
+          (payload, _options) = JWT.decode(token,
+                                               Settings.authentication.secret,
+                                               Settings.authentication.algorithm)
+          payload['email']
         end
       end
     end
--- a/spec/requests/api/v2/providers/courses_spec.rb
+++ b/spec/requests/api/v2/providers/courses_spec.rb
@@ -43,6 +43,19 @@ describe 'Courses API v2', type: :request do
       it { should have_http_status(:unauthorized) }
     end
 
+    context 'when user has not accepted terms' do
+      let(:user_without_terms_accepted) { create(:user, accept_terms_date_utc: nil) }
+      let(:organisation) { create(:organisation, users: [user, user_without_terms_accepted]) }
+      let(:payload) { { email: user_without_terms_accepted.email } }
+
+      before do
+        get "/api/v2/providers/#{provider.provider_code}/courses/#{findable_open_course.course_code}",
+            headers: { 'HTTP_AUTHORIZATION' => credentials }
+      end
+
+      it { should have_http_status(:forbidden) }
+    end
+
     context 'when unauthorised' do
       let(:unauthorised_user) { create(:user) }
       let(:payload) { { email: unauthorised_user.email } }
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -19,5 +19,6 @@ FactoryBot.define do
     email { Faker::Internet.email }
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
+    accept_terms_date_utc { Faker::Date.backward(1) }
   end
 end


app/controllers/api/v2/application_controller.rb
```diff
@@ -5,23 +5,35 @@ class ApplicationController < ::ApplicationController
...
+          #
+          #    curl -H 'Authorization: Bearer user@education.gov.uk' http://localhost:3001/api/v2/providers
+          email = token
+        else
+          (payload, _options) = JWT.decode(token,
+                                               Settings.authentication.secret,
+                                               Settings.authentication.algorithm)
+          email = payload['email']
+        end
+        email
```
defong: is this line really needed?

i thought that if blocks return and else block returns you get the email anyway

db/schema.rb
```diff
@@ -192,6 +192,10 @@
     t.index ["subject_code"], name: "AK_subject_subject_code", unique: true
   end
 
+  create_table "temp_tbl", id: false, force: :cascade do |t|
```
defong: what is this?

app/controllers/api/v2/application_controller.rb
tvararu: It's not, but I think it's a bit more explicit... not fussed with either way though.

db/schema.rb
tvararu: Garbage, didn't mean to commit it.

app/controllers/api/v2/application_controller.rb
benilovj: you could do

```ruby
def get_email_from_token(token)
  if Settings.authentication.algorithm == 'plain-text'
    # ...
    token
  else
    ...
    payload['email']
  end
end
```

app/controllers/api/v2/application_controller.rb
```diff
@@ -5,23 +5,34 @@ class ApplicationController < ::ApplicationController
...
-            #    curl -H 'Authorization: Bearer user@education.gov.uk' http://localhost:3001/api/v2/providers
-            email = token
-          else
-            (payload, _options) = JWT.decode(token,
-                                                 Settings.authentication.secret,
-                                                 Settings.authentication.algorithm)
-            email = payload['email']
-          end
-          # match email addresses case-insensitively
+          email = get_email_from_token(token)
```
heidar: How about calling this `email_from_token` as `get_` is not very Rubyish?

app/controllers/api/v2/application_controller.rb
```diff
@@ -5,23 +5,34 @@ class ApplicationController < ::ApplicationController
...
+
           @current_user = User.find_by("lower(email) = ?", email.downcase)
+
+          if @current_user.present? && @current_user.accept_terms_date_utc.nil?
```
heidar: This could be rewritten using the safe navigation operator.

`head 403 unless @current_user&.accept_terms_date_utc`

tvararu: Thing is I want to add a message and not just the `HEAD`, so I will use the safe navigation operator but keep it as an `if` block.

app/controllers/api/v2/application_controller.rb
```diff
@@ -5,23 +5,34 @@ class ApplicationController < ::ApplicationController
...
+            head 403
```
misaka: Will this work? If `@current_user.present?` is true then the controller action will continue as before, no?

Maybe this would be better off as a hook that triggers after the `authenticate` hook, so that it can access `@current_user` and perform it's work in isolation.

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -2,7 +2,8 @@
 
 describe 'Courses API v2', type: :request do
   let(:user) { create(:user) }
-  let(:organisation) { create(:organisation, users: [user]) }
+  let(:user_without_terms_accepted) { create(:user, accept_terms_date_utc: nil) }
+  let(:organisation) { create(:organisation, users: [user, user_without_terms_accepted]) }
```
misaka: You could move `user_without_terms_accepted` and this definition of `organisation` (with a `user_without_terms_accepted` into the `context` below). That would describe the context better there and avoid polluting this global context.

app/controllers/api/v2/application_controller.rb
tvararu: Oh, we can actually access `@current_user` after the previous hook... that would be really good

app/controllers/api/v2/application_controller.rb
```diff
@@ -5,23 +5,34 @@ class ApplicationController < ::ApplicationController
...
+          email = email_from_token(token)
+
           @current_user = User.find_by("lower(email) = ?", email.downcase)
+
           @current_user.present?
         end
       end
+
+      def check_terms_accepted
+        render body: '{ error: 403, message: "Forbidden. The user has not accepted terms and conditions." }', status: :forbidden if @current_user&.accept_terms_date_utc.nil?
```
tvararu: I think this should be one of these: https://jsonapi.org/format/#error-objects

tvararu:
[New publishers should tick Ts & Cs](https://trello.com/c/z1KwaS8B/1165-new-publishers-should-tick-ts-cs)

app/controllers/api/v2/application_controller.rb
```diff
@@ -5,23 +5,45 @@ class ApplicationController < ::ApplicationController
...
-          # match email addresses case-insensitively
+          email = email_from_token(token)
+
           @current_user = User.find_by("lower(email) = ?", email.downcase)
+
           @current_user.present?
         end
       end
+
+      def check_terms_accepted
```
defong: As discussed: How about making these private methods?

app/controllers/api/v2/application_controller.rb
```diff
@@ -5,23 +5,45 @@ class ApplicationController < ::ApplicationController
...
+          error_body = {
+            errors: [
+              {
+                status: 403,
+                title: 'Forbidden',
+                detail: 'The user has not accepted terms and conditions.'
+              }
+            ]
+          }
+          render body: error_body.to_json, status: :forbidden
```
defong: As discussed: maybe use a guard clause here to avoid nesting?

app/controllers/api/v1/application_controller.rb
```diff
@@ -8,6 +8,10 @@ def authenticate
           ActiveSupport::SecurityUtils.secure_compare(token, Rails.application.config.authentication_token)
         end
       end
+
+    private
+
+      def check_terms_accepted; end
```
defong: Do we need this on both controllers? Can this not just be on the v2 controller, along with the before action?

app/controllers/api/v2/application_controller.rb
```diff
@@ -3,25 +3,49 @@ module V2
...
           @current_user = User.find_by("lower(email) = ?", email.downcase)
+
           @current_user.present?
         end
       end
+
+    private
+
+      def check_terms_accepted
+        return unless @current_user&.accept_terms_date_utc.nil?
```
misaka: Can we use this please?

```
return if @current_user&.accept_terms_date_utc.present?
```

My mind is weak and the `unless` with a `nil?` always catches me out. :)

app/controllers/api/v2/application_controller.rb
```diff
@@ -3,25 +3,49 @@ module V2
...
+        error_body = {
+          errors: [
+            {
+              status: 403,
+              title: 'Forbidden',
+              detail: 'The user has not accepted terms and conditions.'
+            }
+          ]
+        }
+        render json: error_body, status: :forbidden
```
misaka: I'm not sure why, but this also caught me out. I expected there to be a double-`render` triggered, but maybe that's old Rails behaviour ... I tested it out and placing the `render` in here appears to prevent the controller action from running. 💯

app/controllers/api/v2/application_controller.rb
tvararu: I don't like `unless` either 👍

PR #204, opened 2019-03-19
[1131] Update course changed_at when updating site status

defong:
### Context
Course changed_at is updated when updating site status

Added ability to update change at for course when site statuses changes.
This is done via a reusable concern, through a call back.

### DIFF
--- a/spec/requests/api/v1/course_spec.rb
+++ b/spec/requests/api/v1/course_spec.rb
@@ -55,6 +55,8 @@ describe "Courses API", type: :request do
           applications_accepted_from: "2018-10-09 00:00:00",
           course: course,
           site: site)
+
+        course.update_attributes changed_at: 2.hours.ago
       end
 
       it "returns http success" do
--- a/spec/requests/api/v1/course_spec.rb
+++ b/spec/requests/api/v1/course_spec.rb
@@ -56,7 +56,7 @@ describe "Courses API", type: :request do
           course: course,
           site: site)
 
-        course.update_attributes changed_at: 2.hours.ago
+        course.update changed_at: 2.hours.ago
       end
 
       it "returns http success" do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -114,4 +114,7 @@ RSpec.configure do |config|
   Kernel.srand config.seed
 =end
   config.include JSONAPI::RSpec
+
+  # Load shared examples
+  Dir['./spec/support/shared_examples/**/*.rb'].each { |f| require f }
 end
--- a/spec/support/shared_examples/touch_course_spec.rb
+++ b/spec/support/shared_examples/touch_course_spec.rb
@@ -1,0 +1,21 @@
+require 'rails_helper'
+
+shared_examples 'Touch course' do
+  describe '#touch_course' do
+    let!(:model) { create(described_class.model_name.singular.to_sym) }
+
+    it 'sets changed_at on the parent course to the current time' do
+      Timecop.freeze do
+        model.save
+        expect(model.course.changed_at).to eq Time.now.utc
+      end
+    end
+
+    it 'leaves updated_at on the parent course unchanged' do
+      timestamp = 1.hour.ago
+      model.course.update updated_at: timestamp
+      model.save
+      expect(model.course.updated_at).to eq timestamp
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


defong:
[Site status validation](https://trello.com/c/fKdscaJb/1187-site-status-validation)

defong:

spec/requests/api/v1/course_spec.rb
```diff
@@ -49,12 +49,13 @@ def get_course_codes_from_body(body)
           provider: provider,
           age: 2.hours.ago)
 
-        course.site_statuses.first.update(
+        course.site_statuses.first.update_attributes(
```
heidar: These changes were needed because the `changed_at` gets updated implicitly when creating the course with the factory.

spec/spec_helper.rb
```diff
@@ -114,4 +114,7 @@
   Kernel.srand config.seed
 =end
   config.include JSONAPI::RSpec
+
+  # Load shared examples
+  Dir['./spec/support/shared_examples/**/*.rb'].each { |f| require f }
```
heidar: Are we all happy to store shared examples in this folder?

spec/requests/api/v1/course_spec.rb
misaka: `#update_attributes` is an alias for `#update`, and looks like it's actually been deprecated in favour of the later:

https://github.com/rails/rails/blob/35efe44f5379b031e6d4abab7b0a8c280e31b640/activerecord/lib/active_record/persistence.rb#L641-L642

```ruby
    alias update_attributes update
    deprecate update_attributes: "please, use update instead"
```

Is there a reason for changing this to `update_attributes`?

spec/requests/api/v1/course_spec.rb
```diff
@@ -49,12 +49,13 @@ def get_course_codes_from_body(body)
...
           vac_status: :full_time_vacancies,
           publish: 'Y',
           status: :running,
           applications_accepted_from: "2018-10-09 00:00:00"
         )
+        course.update_attributes changed_at: 2.hours.ago
```
misaka: I think it might be good to use `update_columns` here, which bypasses validations and hooks. Although, and this might echo your own comments to me about how do our factories, could we also create the `SiteStatus` as part of the create of the `Course`, maybe by passing it in instead of the `site`? That would avoid having to reset `changed_at` I think.

spec/spec_helper.rb
misaka: Works for me. Generally I, and you may already agree with me, would prefer to avoid shared example, preferring custom matchers where possible. But I do agree here that it may be the better approach.

spec/support/shared_examples/touch_course_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
+
+shared_examples 'Touch course' do
+  describe '#touch_course' do
+    let!(:model) { create(described_class.model_name.singular.to_sym) }
```
misaka: This feels like a potential gotcha. Would passing in an object as a parameter to this shared example be worse?

spec/requests/api/v1/course_spec.rb
heidar: Ah yes. We should configure our linter to tell us about that. It's a default in Rubocop, not sure why it is disabled.

spec/support/shared_examples/touch_course_spec.rb
heidar: Yeah agreed, that would be easier to understand!

spec/requests/api/v1/course_spec.rb
heidar: I think `#update` is fine for that, it doesn't fire any of those. As for changing the creation, it still causes the callback to fire, so the `changed_at` will be the current time.

I feel like the root cause is doing this via callback instead of explicitly when we want to publish the course to UCAS.

heidar:
I've addressed all the comments @misaka :) Thanks for the review!

spec/requests/api/v1/course_spec.rb
dankmitchell: I with `govuk-lint` forced us to use parenthesis here i.e. `course.update(changed_at: 2.hours.ago)`

misaka: This should changed to `#update`, then. I wonder if the approach I outlined in #209 would benefit here.

heidar: Oops, my bad, done now.

heidar:
@misaka all sorted now.

PR #205, opened 2019-03-20
[1172] add ucas preferences to providers

misaka:
https://trello.com/c/1R5mTZjg/1172-provider-preferences-modify-provider-model

### Context

Adding support for provider UCAS preferences as per ticket.

Add migration and model for ProviderUCASPreference objects

### DIFF
--- a/app/models/provider_ucas_preference.rb
+++ b/app/models/provider_ucas_preference.rb
@@ -2,12 +2,14 @@
 #
 # Table name: provider_ucas_preference
 #
-#  id                      :bigint(8)        not null, primary key
-#  provider_id             :integer          not null
-#  type_of_gt12            :text
-#  send_application_alerts :text
-#  created_at              :datetime         not null
-#  updated_at              :datetime         not null
+#  id                        :bigint(8)        not null, primary key
+#  provider_id               :integer          not null
+#  type_of_gt12              :text
+#  send_application_alerts   :text
+#  application_alert_email   :text
+#  gt12_response_destination :text
+#  created_at                :datetime         not null
+#  updated_at                :datetime         not null
 #
 
 class ProviderUCASPreference < ApplicationRecord
--- a/app/models/provider_ucas_preference.rb
+++ b/app/models/provider_ucas_preference.rb
@@ -1,0 +1,31 @@
+# == Schema Information
+#
+# Table name: provider_ucas_preference
+#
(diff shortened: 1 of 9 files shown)


db/migrate/2019031920190319163755_add_provider_ucas_preferences.rb
```diff
@@ -0,0 +1,11 @@
+class AddProviderUCASPreferences < ActiveRecord::Migration[5.2]
+  def change
+    create_table :provider_ucas_preference do |t|
+      t.integer :provider_id, null: false
+      t.text :name, null: false
+      t.text :value
```
heidar: I'm not sure I'm a big fan of `value` as a name for an attribute, very ambiguous.

spec/models/provider_ucas_preference_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    expect(subject)
+      .to define_enum_for(:name)
+            .backed_by_column_of_type(:text)
+            .with_values(
+              type_of_gt12: 'type_of_gt12',
+            )
+  end
+
+  describe 'type_of_gt12 values' do
+    let(:provider) { create :provider }
```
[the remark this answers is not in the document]
misaka: This is gone now.

db/migrate/2019031920190319163755_add_provider_ucas_preferences.rb
misaka: Ya, I decided this whole abstraction was unnecessary, and this approach has been redone.

app/models/provider_ucas_preference.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+#  id                      :bigint(8)        not null, primary key
+#  provider_id             :integer          not null
+#  type_of_gt12            :text
+#  send_application_alerts :text
+#  created_at              :datetime         not null
+#  updated_at              :datetime         not null
+#
+
+class ProviderUCASPreference < ApplicationRecord
```
dankmitchell: Can this just be `UCASPreference` #namingishard

misaka: I wanted the table and model to be clearly linked to the provider. The relationship strips the `provider_` part off, since it's redundant when applied to a provider there.

What do others think?

PR #208, opened 2019-03-20
[1177] add contacts for providers

davidgisbey:
https://trello.com/c/XanBgbYi/1177-provider-contacts-add-model-migrations

### Context
Adding support for provider UCAS contacts as per ticket.

Add migration and model for Contact objects

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -4,7 +4,7 @@
 #
 #  id          :bigint(8)        not null, primary key
 #  provider_id :integer          not null
-#  type        :text
+#  type        :text             not null
 #  name        :text
 #  email       :text
 #  telephone   :text
@@ -13,10 +13,12 @@
 #
 
 class Contact < ApplicationRecord
+  self.inheritance_column = '_unused'
+
   belongs_to :provider
 
   enum type: {
-         admin:  'admin',
+         admin: 'admin',
          utt: 'utt',
          web_link: 'web_link',
          fraud: 'fraud',
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -1,0 +1,26 @@
+# == Schema Information
+#
+# Table name: contact
+#
+#  id          :bigint(8)        not null, primary key
+#  provider_id :integer          not null
+#  type        :text
+#  name        :text
+#  email       :text
+#  telephone   :text
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+
+class Contact < ApplicationRecord
+  belongs_to :provider
+
+  enum type: {
+         admin:  'admin',
+         utt: 'utt',
+         web_link: 'web_link',
+         fraud: 'fraud',
+         finance: 'finance'
+       },
+       _suffix: 'contact'
+end
--- a/db/migrate/20190320144534_create_contact.rb
+++ b/db/migrate/20190320144534_create_contact.rb
@@ -1,0 +1,15 @@
+class CreateContact < ActiveRecord::Migration[5.2]
+  def change
+    create_table :contact do |t|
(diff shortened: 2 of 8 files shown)


spec/factory_specs/contacts_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+describe 'Contact factory' do
+  let(:contact) { create(:contact) }
+
+  it 'creates ProviderUCASPreference object' do
+    expect(contact).to be_instance_of(Contact)
+    expect(contact).to be_valid
+  end
```
defong: describe 'associations' do
end

??

app/models/contact.rb
```diff
@@ -0,0 +1,26 @@
...
+#  telephone   :text
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+
+class Contact < ApplicationRecord
+  belongs_to :provider
+
+  enum type: {
+         admin:  'admin',
```
heidar: Extra space here :p

spec/factory_specs/contacts_spec.rb
misaka: This is a factory spec that proves that the factory works, not a model test so no need to test associations.

app/models/contact.rb
```diff
@@ -0,0 +1,26 @@
...
+#  email       :text
+#  telephone   :text
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+
+class Contact < ApplicationRecord
+  belongs_to :provider
+
+  enum type: {
```
heidar: Can we all this `contact_type` maybe? `type` is a special column in active record for single table inheritance.

db/migrate/20190320144534_create_contact.rb
```diff
@@ -0,0 +1,15 @@
+class CreateContact < ActiveRecord::Migration[5.2]
+  def change
+    create_table :contact do |t|
+      t.integer :provider_id, null: false
+      t.text :type
```
heidar: Should all of these columns be nullable?

spec/models/contact_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require 'rails_helper'
+
+describe Contact, type: :model do
+  it { should belong_to(:provider) }
+
+  describe 'type' do
+    it 'is an enum' do
+      expect(subject)
+        .to define_enum_for(:type)
```
heidar: Is there value in testing this? Feels like testing config, but I'm not sure.

spec/models/contact_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+              .backed_by_column_of_type(:text)
+              .with_values(
+                admin:  'admin',
```
heidar: Extra space again :o

db/migrate/20190320144534_create_contact.rb
davidgisbey: Done. Thanks for the feedback.

app/models/contact.rb
misaka: I'd prefer to just rename the `inheritance_column` since this seems like the right name to use for this value.

One test I use for this is to ask myself, if we were going to use STI, would the values in this column likely be the ones to decide the class name? And I think the answer here is yes. What do you think?

spec/models/contact_spec.rb
misaka: This is testing config that is checked-in in the code. I'd be against testing config that's changeable outside the system.

app/models/contact.rb
heidar: Makes sense!

PR #209, opened 2019-03-21
[1173] add ucas preferences serialiser

misaka:
https://trello.com/c/kezSZblP/1173-provider-preferences-serialiser

### Context

We need to expose the provider preferences that are in the DB in the V1 API, right now they are being semi-randomly chosen.

This change pulls these values from the DB in the serialiser, and fixes a bunch of related tests.


tvararu:
[Provider preferences - serialiser](https://trello.com/c/kezSZblP/1173-provider-preferences-serialiser)

PR #212, opened 2019-03-22
[1174] load mcb sub-commands automatically

misaka:
### Context

We will be adding commands to `mcb` for importing, this change makes is simpler and less error-prone to do this.

Automatically load all the commands and sub-commands in `lib/mcb/commands` instead of having to list them all out in `bin/mcb` manually.

### Guidance to review

Hey look! A spec for mcb! And a module too!!


spec/bin/mcb_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
```
defong: Is there a point with `rails_helper`?

Just thinking its a console app no longer a rails app

defong:
[Provider preferences - mcb importer](https://trello.com/c/6Iz0uWm8/1174-provider-preferences-mcb-importer)

spec/bin/mcb_spec.rb
misaka: Good spot.

Some commands _will_ load the Rails environment, though, but I guess we can worry about those later.

spec/lib/mcb_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'spec_helper'
+require "#{Rails.root}/lib/mcb"
+
+describe 'mcb command' do
+  describe '#load_commands' do
+    include FakeFS::SpecHelpers
```
defong: is there something we can about 
` include FakeFS::SpecHelpers`

spec/spec_helper.rb
```diff
@@ -5,6 +5,7 @@
 require "super_diff/rspec"
 require 'support/matchers/have_courses'
 require 'support/matchers/have_providers'
+require 'fakefs/spec_helpers'
```
defong: That was the thing i was thinking about for 

` include FakeFS::SpecHelpers`

PR #213, opened 2019-03-22
1179 add provider contact serializer

davidgisbey:
### Context

We need to expose the provider contacts/application alert recipient  in the V1 API.

- Attributes added to the contact factory bot
- Creates a contact serialiser 
- Removes dummy data logic from the provider serializer
- Adds application alert recipient as an attribute to the provider serialiser
- Adds contacts as an attribute to the alert serialiser

### Guidance to review

Application alert recipient is an anomalous contact as it is simply an email address that is contacted on application receipt. Due to this it has been added a column by UCAS preferences model and is exposed in the api via adding it manually to the Provider Serialiser contacts attribute.

Lines 38-43 of provider serialiser (see bold)

attribute :contacts do
    object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') } + **[{
      type: 'application_alert_recipient',
      email: object.ucas_preferences.application_alert_email,
      }]**
  end

### DIFF
--- a/app/serializers/contact_serializer.rb
+++ b/app/serializers/contact_serializer.rb
@@ -21,19 +21,5 @@
 class ContactSerializer < ActiveModel::Serializer
   belongs_to :provider
 
-  attribute :type do
-    object.type
-  end
-
-  attribute :name do
-    object.name
-  end
-
-  attribute :email do
-    object.email
-  end
-
-  attribute :telephone do
-    object.telephone
-  end
+  attributes :type, :name, :email, :telephone
 end
--- a/app/serializers/contact_serializer.rb
+++ b/app/serializers/contact_serializer.rb
@@ -1,0 +1,39 @@
+# == Schema Information
+#
+# Table name: contact
+#
+#  id          :bigint(8)        not null, primary key
+#  provider_id :integer          not null
+#  type        :text             not null
+#  name        :text
+#  email       :text
+#  telephone   :text
+#  created_at  :datetime         not null
+#  updated_at  :datetime         not null
+#
+
+# application alert recipient has been added as a contact manually in the provider
+# contact serializer (../provider_serializer.rb:38). It has not been added into
+# the enum in the contact model as it's does not share the name and email attribute.
+# It is simply a contact email address and is more suited to sit in the ucas
+# preferences model.
+
+class ContactSerializer < ActiveModel::Serializer
+  belongs_to :provider
+
+  attribute :type do
+    object.type
+  end
+
+  attribute :name do
+    object.name
+  end
+
+  attribute :email do
+    object.email
+  end
+
+  attribute :telephone do
+    object.telephone
+  end
+end
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -33,7 +33,14 @@ class ProviderSerializer < ActiveModel::Serializer
   attributes :institution_code, :institution_name, :institution_type, :accrediting_provider,
              :address1, :address2, :address3, :address4, :postcode, :region_code, :scheme_member,
              :contact_name, :email, :telephone, :recruitment_cycle, :utt_application_alerts,
-             :type_of_gt12
+             :type_of_gt12, :application_alert_recipient
+
+  attribute :contacts do
+    object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') } + [{
+      type: 'application_alert_recipient',
+      email: object.ucas_preferences.application_alert_email,
+      }]
+  end
 
   def institution_code
     object.provider_code
@@ -87,22 +94,8 @@ class ProviderSerializer < ActiveModel::Serializer
     @object.ucas_preferences.type_of_gt12_before_type_cast
   end
 
-  attribute :contacts do
-    %w[
-      admin
-      utt
-      web_link
-      fraud
-      finance
-      application_alert_recipient
-    ].map do |type|
-      {
-        type: type,
-        name: "#{type.humanize.titleize} Contact #{@object.provider_code}",
-        email: @object.email&.sub(/.*@/, "#{type}@"),
-        telephone: @object.telephone,
-      }
-    end
+  def application_alert_recipient
+    @object.ucas_preferences.application_alert_email
   end
 
 private
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -33,13 +33,16 @@ class ProviderSerializer < ActiveModel::Serializer
   attributes :institution_code, :institution_name, :institution_type, :accrediting_provider,
              :address1, :address2, :address3, :address4, :postcode, :region_code, :scheme_member,
              :contact_name, :email, :telephone, :recruitment_cycle, :utt_application_alerts,
-             :type_of_gt12, :application_alert_recipient
+             :type_of_gt12
+
+             # application alert reciepient has not been added into the enum in the contact model
+             # as it's does not share the name and email attribute. It is simply a contact email
+             # address and is more suited to sit in the ucas  preferences model. However, in the docs
+             # it is exposed in the contacts so has been added directly into the contacts attribute
+             # see lines 109-120 for logic.
 
   attribute :contacts do
-    object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') } + [{
-      type: 'application_alert_recipient',
-      email: object.ucas_preferences.application_alert_email,
-      }]
+    generate_provider_contacts + add_application_alert_recipient
   end
 
   def institution_code
@@ -94,10 +97,6 @@ class ProviderSerializer < ActiveModel::Serializer
     @object.ucas_preferences.type_of_gt12_before_type_cast
   end
 
-  def application_alert_recipient
-    @object.ucas_preferences.application_alert_email
-  end
-
 private
 
   def select_value_for_provider(provider_code, values)
@@ -105,5 +104,18 @@ private
     # hash. ex. 'A1'.to_i(36) == 361
     index = provider_code.to_i(36) % values.count
     values[index]
+  end
+
+  def generate_provider_contacts
+    object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') }
+  end
+
+  def add_application_alert_recipient
+    [{
+      type: 'application_alert_recipient',
+      name: '',
+      email: object.ucas_preferences.application_alert_email,
+      telephone: ''
+      }]
   end
 end
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -35,14 +35,14 @@ class ProviderSerializer < ActiveModel::Serializer
              :contact_name, :email, :telephone, :recruitment_cycle, :utt_application_alerts,
              :type_of_gt12
 
-             # application alert reciepient has not been added into the enum in the contact model
-             # as it's does not share the name and email attribute. It is simply a contact email
-             # address and is more suited to sit in the ucas  preferences model. However, in the docs
+             # application alert recipient has not been added into the enum in the contact model
+             # as it does not share the name and email attribute. It is simply a contact email
+             # address and is more suited to sit in the ucas preferences model. However, in the docs
              # it is exposed in the contacts so has been added directly into the contacts attribute
-             # see lines 109-120 for logic.
+             # see generate_provider_contacts and return_application_alert_recipient for logic.
 
   attribute :contacts do
-    generate_provider_contacts + add_application_alert_recipient
+    generate_provider_contacts + application_alert_recipient
   end
 
   def institution_code
@@ -110,7 +110,7 @@ private
     object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') }
   end
 
-  def add_application_alert_recipient
+  def application_alert_recipient
     [{
       type: 'application_alert_recipient',
       name: '',
--- a/spec/factories/contacts.rb
+++ b/spec/factories/contacts.rb
@@ -20,4 +20,24 @@ FactoryBot.define do
     email { Faker::Internet.email }
     telephone { Faker::PhoneNumber.phone_number }
   end
+
+  trait :admin_contact do
+    type { :admin }
+  end
+
+  trait :utt_contact do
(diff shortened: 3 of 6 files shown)


app/serializers/contact_serializer.rb
```diff
@@ -0,0 +1,39 @@
...
+# the enum in the contact model as it's does not share the name and email attribute.
+# It is simply a contact email address and is more suited to sit in the ucas
+# preferences model.
+
+class ContactSerializer < ActiveModel::Serializer
+  belongs_to :provider
+
+  attribute :type do
+    object.type
+  end
```
misaka: When we aren't actually changing anything about these attributes, we don't need to define a block. This could be simply be:

```ruby
attributes :type, :name, :email, :telephone
```

app/serializers/provider_serializer.rb
```diff
@@ -87,22 +94,8 @@ def type_of_gt12
...
-    ].map do |type|
-      {
-        type: type,
-        name: "#{type.humanize.titleize} Contact #{@object.provider_code}",
-        email: @object.email&.sub(/.*@/, "#{type}@"),
-        telephone: @object.telephone,
-      }
-    end
+  def application_alert_recipient
+    @object.ucas_preferences.application_alert_email
```
misaka: The API spec doesn't list this attribute on the `providers` endpoint, so probably not worth exposing it here.

spec/requests/api/v1/provider_spec.rb
```diff
@@ -174,9 +227,7 @@ def get_provider_codes_from_body(body)
                   },
                   {
                     'type' => 'application_alert_recipient',
-                    'name' => 'Application Alert Recipient Contact A123',
-                    'email' => 'application_alert_recipient@acmescitt.education.uk',
-                    'telephone' => '020 812 345 678'
+                    'email' => 'application_alert_recipient@acmescitt.education.uk'
```
misaka: I feel like it'd be more cautious to list `email` and `telephone` here, but just have them set to empty strings.

app/serializers/provider_serializer.rb
```diff
@@ -33,7 +33,14 @@ class ProviderSerializer < ActiveModel::Serializer
...
              :address1, :address2, :address3, :address4, :postcode, :region_code, :scheme_member,
              :contact_name, :email, :telephone, :recruitment_cycle, :utt_application_alerts,
-             :type_of_gt12
+             :type_of_gt12, :application_alert_recipient
+
+  attribute :contacts do
+    object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') } + [{
+      type: 'application_alert_recipient',
+      email: object.ucas_preferences.application_alert_email,
+      }]
```
misaka: I know I'm the one that contributed this bit of mess, but how about we use a transient variable here to help explain what `object.contacts.map { ... }` is all about.

It would probably also help to have a comment to explain what's going on here.

spec/serializers/provider_serializer_spec.rb
```diff
@@ -84,4 +84,78 @@
...
+    describe 'admin contact' do
+      let(:provider) do
+        create :provider,
+               contacts: [admin]
+      end
+      let(:admin) { create(:contact, type: 'admin') }
+
+      subject { serialize(provider)['contacts'].find { |c| c[:type] == 'admin' } }
+
+      its([:name]) { should eq admin.name }
```
defong: whats with the square brackets?

`its(:name)` ?

spec/serializers/provider_serializer_spec.rb
```diff
@@ -84,4 +84,78 @@
...
+      its([:email]) { should eq admin.email }
+      its([:telephone]) { should eq admin.telephone }
+    end
+
+    describe 'utt contact' do
```
defong: context vs describe

spec/serializers/provider_serializer_spec.rb
```diff
@@ -84,4 +84,78 @@
 
     it { should eq provider.ucas_preferences.send_application_alerts_before_type_cast }
   end
+
+  describe 'contacts' do
+    describe 'admin contact' do
```
defong: should this test be closer to the model? then here?

spec/serializers/provider_serializer_spec.rb
```diff
@@ -84,4 +84,78 @@
...
+      let(:provider) do
+        create :provider,
+               contacts: [admin]
```
defong: I will probably just the put 
contacts: [admin, utt, banana, bananas]

and just loop it ...

spec/serializers/provider_serializer_spec.rb
davidgisbey: :name is returned by a hash rather than a method (:name) just returns an indifferent hash error.

misaka: `its(:name)` -> `subject.name`
`its([:name])` -> `subject[:name]`

spec/serializers/provider_serializer_spec.rb
misaka: imo

* **describe**: this is what we're testing (kind of like a noun)
* **context**: this is what state the system is in (kind of like an adjective)

spec/serializers/provider_serializer_spec.rb
defong: or 

```
	the_types.each do |type|
		describe '#{the_type} contact' do
		  let(:provider) do
			create :provider,
				   contacts: [contact]
		  end

		  let(:contact) { create(:contact, type: the_type) }

		  subject { serialize(provider)['contacts'].find { |c| c[:type] == the_type } }

		  its([:name]) { should eq contact.name }
		  its([:email]) { should eq contact.email }
		  its([:telephone]) { should eq contact.telephone }
		end
	end
```

spec/serializers/provider_serializer_spec.rb
defong: Then I'm just getting a feeling there is a lack of intent or the understanding of significant of what it is `testing` or its halfway point, from line 88 to 160 (Nevermind the dupes).

Therefore its more aligned with `context` then `describe`.

spec/serializers/provider_serializer_spec.rb
```diff
@@ -84,4 +84,78 @@
...
+      end
+      let(:admin) { create(:contact, type: 'admin') }
```
defong: Why?

spec/factories/contacts.rb
```diff
@@ -16,5 +16,8 @@
   factory :contact do
     provider
     type { 'admin' }
```
defong: traits

app/serializers/contact_serializer.rb
davidgisbey: Done 👍

spec/serializers/provider_serializer_spec.rb
misaka: I think if I put all your comments together, mix them up, and then reduce them on the hob for a bit, stirring constantly, I come up with something like:

> [You don't need to test every type of contact here, since the code behaves the same for each type and doesn't differentiate, so why bother?]

So we should be able to test for a single contact type, which makes sense to me. 😸

defong: 👍

app/serializers/provider_serializer.rb
davidgisbey: Done and refactored the tests so to ensure it only appears once (in contacts not prov pref)

spec/requests/api/v1/provider_spec.rb
davidgisbey: Done 👍

spec/serializers/provider_serializer_spec.rb
davidgisbey: Just for clarity on what is being tested.

spec/serializers/provider_serializer_spec.rb
davidgisbey: Thanks for all the feedback. Will do.

spec/serializers/provider_serializer_spec.rb
davidgisbey: Addressed in the comment below.

app/serializers/provider_serializer.rb
davidgisbey: Okay have done! Thanks for all the feedback 👍

spec/serializers/provider_serializer_spec.rb
```diff
@@ -84,4 +84,18 @@
 
     it { should eq provider.ucas_preferences.send_application_alerts_before_type_cast }
   end
+
+  describe 'contacts' do
+    let(:provider) do
+      create :provider,
+             contacts: [utt]
+    end
+    let(:utt) { create(:contact, type: 'utt') }
```
defong: made the trait use the trait

`create(:contact, :utt_contact)`

spec/factories/contacts.rb
```diff
@@ -16,5 +16,28 @@
...
+    name { Faker::Name.name }
+    email { Faker::Internet.email }
+    telephone { Faker::PhoneNumber.phone_number }
+  end
+
+  trait :admin_contact do
```
defong: maybe `:xxx_type` rather than `xxx_contact` as it object itself is a contact

spec/factories/contacts.rb
```diff
@@ -16,5 +16,28 @@
   factory :contact do
     provider
     type { 'admin' }
```
defong: may as well

`admin_contact` instead of `type { 'admin' }`

app/serializers/provider_serializer.rb
```diff
@@ -35,6 +35,16 @@ class ProviderSerializer < ActiveModel::Serializer
              :contact_name, :email, :telephone, :recruitment_cycle, :utt_application_alerts,
              :type_of_gt12
 
+             # application alert reciepient has not been added into the enum in the contact model
+             # as it's does not share the name and email attribute. It is simply a contact email
+             # address and is more suited to sit in the ucas  preferences model. However, in the docs
+             # it is exposed in the contacts so has been added directly into the contacts attribute
+             # see lines 109-120 for logic.
```
defong: drop lines number just refer to those methods...
spacing of comment blocks you referring to `attributes ` or `attribute :contacts do`

ie line numbers will jump around

typo `reciepient` 
grammer `it's` does not share
double space ucas`  `preferences model

app/serializers/provider_serializer.rb
```diff
@@ -113,4 +105,17 @@ def select_value_for_provider(provider_code, values)
     index = provider_code.to_i(36) % values.count
     values[index]
   end
+
+  def generate_provider_contacts
+    object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') }
+  end
+
+  def add_application_alert_recipient
```
defong: maybe not `add_something` as its not really adding....

spec/serializers/provider_serializer_spec.rb
davidgisbey: Done 👍

spec/factories/contacts.rb
davidgisbey: and done

spec/factories/contacts.rb
davidgisbey: and done

app/serializers/provider_serializer.rb
davidgisbey: And done. Thanks for the thorough feedback @defong.

app/serializers/provider_serializer.rb
```diff
@@ -113,4 +105,17 @@ def select_value_for_provider(provider_code, values)
     index = provider_code.to_i(36) % values.count
     values[index]
   end
+
+  def generate_provider_contacts
+    object.contacts.map { |c| c.attributes.slice('type', 'name', 'email', 'telephone') }
+  end
+
+  def return_application_alert_recipient
```
misaka: How about just simply `application_alert_recipient`? i.e. just a noun without attaching a verb/action to it, which is a bit more idiomatic Ruby. I don't think that clashes with anything, but if it does you could use `application_alert_recipient_record` or similar.

davidgisbey: application_alert_recipient worked fine 😄

spec/serializers/provider_serializer_spec.rb
```diff
@@ -84,4 +84,18 @@
 
     it { should eq provider.ucas_preferences.send_application_alerts_before_type_cast }
   end
+
+  describe 'contacts' do
+    let(:provider) do
+      create :provider,
+             contacts: [utt]
```
defong: do we care its an `utt` if not just `contact`

same with creating it `create(:contact)` as that is just `admin` which is now the default

then this bit can just be refactored as 
`|c| c[:type] == contact.type`

which altogether goes back to 
`describe 'contacts'` 

the subject in question is being able to find a thing of `type` 

where as 
```
  attribute :contacts do
    generate_provider_contacts + application_alert_recipient
  end```

so what is being tested, finding a single item from an array? as there is no other items in array or we dont care about `application_alert_recipient`

Comment (davidgisbey) [reply spec/serializers/provider_serializer_spec.rb]: How's this @defong? Better or worse?
```
describe 'contacts' do
    describe 'generate_provider_contacts returns the correct values' do
      let(:provider) do
        create :provider,
               contacts: [contact]
      end
      let(:contact) { create(:contact) }

      subject { serialize(provider)['contacts'].find { |c| c[:type] == contact.type } }

      its([:name]) { should eq contact.name }
      its([:email]) { should eq contact.email }
      its([:telephone]) { should eq contact.telephone }
    end

    describe 'the application alert recipient has been serialized upon instantiation of the contact object' do

      subject { serialize(provider)['contacts'].find { |c| c[:type] == 'application_alert_recipient' } }

      its([:name]) { should eq '' }
      its([:email]) { should eq provider.ucas_preferences.application_alert_email }
      its([:telephone]) { should eq ''}
    end
  end

```

defong: so it was 2 things 👍

davidgisbey: Yea it is added when the contacts serializer instantiates the contact object. generate_provider_contacts grabs each contacts name, type email etc. from provider.contacts. Then the application_alert_recipient is added separately.

PR #217, opened 2019-03-26
[1133] Set up ManageCoursesAPI service

tvararu:
### Context

This will be used to trigger republishing on the C# API.

Depends on https://github.com/DFE-Digital/manage-courses-backend/pull/215. (ignore the first commit in this)

Related to https://github.com/DFE-Digital/manage-courses-api/pull/283.

Creates a service to encapsulate the request to the API.


tvararu:
[Config in Manage Courses Backend for bearer token auth with Manage Courses API](https://trello.com/c/i9USTN0k/1133-config-in-manage-courses-backend-for-bearer-token-auth-with-manage-courses-api)

app/services/manage_courses_api.rb
```diff
@@ -0,0 +1,36 @@
...
+      def publish_course(email, provider_code, course_code)
+        response = api.post("/api/Publish/internal/course/#{provider_code}/#{course_code}", email: email)
+        begin
+          if response.status == 200
+            JSON.parse(response.body)["result"]
+          else
+            false
+          end
+        rescue StandardError => e
+          raise StandardError("Could not publish course #{response.status}, #{response.body}, #{e}")
```
defong: Maybe 
`provider_code` 
`course_code`

as well

defong: May a test for this part as well since its system to system integration as well

config/settings.yml
```diff
@@ -3,3 +3,8 @@ authentication:
   # Set this in the env! The below ensures that we are un-authenticatable if we
   # forget to do this in production.
   secret: <%= SecureRandom.base64 %>
+manage_api:
+  # URL of the C# API app (manage-courses-api)
+  base_url: http://localhost:6001
+  # Auth token used between backend and C# API to authorize requests.
+  secret: banana
```
defong: Feeding 🍌

spec/spec_helper.rb
```diff
@@ -114,6 +114,8 @@
   # as the one that triggered the failure.
   Kernel.srand config.seed
 =end
+  require 'webmock/rspec'
```
defong: shouldn't this be at top of file?

app/services/manage_courses_api.rb
```diff
@@ -0,0 +1,36 @@
...
+        faraday.headers['Content-Type'] = 'application/json'
+        faraday.headers['Authorization'] = "Bearer #{TOKEN}"
+      end
+    end
+  end
+
+  class Request
+    class << self
+      def publish_course(email, provider_code, course_code)
+        response = api.post("/api/Publish/internal/course/#{provider_code}/#{course_code}", email: email)
```
defong: Maybe just a sanity check before posting rather than reallying 3 party to tell you something don't exist since both touches the same database

app/services/manage_courses_api.rb
```diff
@@ -0,0 +1,36 @@
...
+        faraday.response :logger unless Rails.env.test?
+        faraday.adapter Faraday.default_adapter
+        faraday.headers['Content-Type'] = 'application/json'
+        faraday.headers['Authorization'] = "Bearer #{TOKEN}"
+      end
+    end
+  end
+
+  class Request
+    class << self
```
defong: Is this inheritance?

app/services/manage_courses_api.rb
```diff
@@ -0,0 +1,36 @@
+module ManageCoursesAPI
+  class Connection
+    BASE_URL = Settings.manage_api.base_url
+    TOKEN = Settings.manage_api.secret
```
defong: wondering is there a need for 2 variable....

app/services/manage_courses_api.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def self.api
+      Faraday.new(url: BASE_URL) do |faraday|
+        faraday.response :logger unless Rails.env.test?
+        faraday.adapter Faraday.default_adapter
+        faraday.headers['Content-Type'] = 'application/json'
+        faraday.headers['Authorization'] = "Bearer #{TOKEN}"
```
defong: Maybe something like this

https://github.com/DFE-Digital/manage-courses-frontend/blob/91efb2d6d592e34a9ea78625f44c790a7e528a17/app/controllers/application_controller.rb#L70

spec/spec_helper.rb
tvararu: Doesn't matter ¯\\\_(ツ)\_\/¯

app/services/manage_courses_api.rb
tvararu: This is making us not have to specify `def self.method` for every method underneath; basically saying "everything in this block is `static`"

app/services/manage_courses_api.rb
tvararu: ¯\\\_(ツ)\_\/¯ there isn't but I like to hoist them at the top in one place instead of scattered in the file.

app/services/manage_courses_api.rb
tvararu: Perfect! 👍

app/services/manage_courses_api.rb
tvararu: Fair point, but should this library do that? I figure it doesn't need to concern itself with that, only with the request...

app/services/manage_courses_api.rb
tvararu: Nevermind, I don't think I can add it here like that.

app/services/manage_courses_api.rb
tvararu: Done and done

app/services/manage_courses_api.rb
defong: Yeah, not there there, but definitely somewhere.

config/settings.yml
misaka: Y U NO HAS NYOULINE?

tvararu: vs code default settings are **** that's why

spec/spec_helper.rb
misaka: Unless this `require` is conditional, I think it would be surprising to see it here. Wouldn't necessary think to look for it somewhere other than the top.

tvararu: I'll move it, but I think this is where it's in MCFE, but that was probably added in the wrong spot also by me

spec/services/manage_courses_api_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        before do
+          stub_request(:post, Settings.manage_api.base_url + "/api/Publish/internal/course/#{provider_code}/#{course_code}")
+            .with { |req| req.body == body }
+            .to_return(
+              status: 200,
+              body: '{ "result": true }'
+            )
+        end
+
+        it "returns the body" do
```
misaka: Returns the body or returns true?

tvararu: I refactored and forgot, thanks

PR #222, opened 2019-03-27
Switch to Rubocop for linting Ruby

heidar:
GOV.UK Lint is being deperacated, this is the new way of using it.

### Context

GOV.UK Lint is being deprecated, it's recommended to be used in this way instead.

I haven left the SCSS listing in there, they recommend using node based tooling for that instead.

Switch from govuk-lint-ruby to Rubocop configured to use govuk-lint's rules.

### Guidance to review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
   - docker-compose exec web /bin/sh -c "bundle exec rails db:setup"
 script:
   - docker-compose exec web /bin/sh -c "bundle exec rake spec"
-  - docker-compose exec web /bin/sh -c "bundle exec govuk-lint-ruby app config db lib spec Gemfile --format clang"
+  - docker-compose exec web /bin/sh -c "bundle exec rubocop app config db lib spec Gemfile --format clang"
 deploy:
   provider: script
   script: bash docker-push.sh
--- a/Gemfile
+++ b/Gemfile
@@ -92,6 +92,9 @@ group :development, :test do
   gem 'rspec-json_matchers'
   gem 'rspec-rails'
 
+  # A Ruby static code analyzer and formatter
+  gem 'rubocop', require: false
+
   # Make diffs of Ruby objects much more readable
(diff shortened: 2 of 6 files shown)


heidar:
> Do we need to update travis and Azure pipeline too?

No sir, the rake task is unchanged and can still be run directly or via default `rake`.

dankmitchell:
@heidar 🤔 our travis config runs `docker-compose exec web /bin/sh -c "bundle exec govuk-lint-ruby app config db lib spec Gemfile --format clang"` so don't we need to update this to `rubocop` instead?

heidar:
Oh, good catch...figured it would use the rake task :/

misaka:
`govuk-ruby-lint` is also in the Gemfile. Should we remove it?

dankmitchell:
@misaka we use the rules from govuk-lint

misaka:
> @misaka we use the rules from govuk-lint

Ya, I forgot that we do this in `.rubocop.yml`. Thing is, as we've found, this gem is being replaced by the gem `govuk-rubocop` in the PR https://github.com/alphagov/govuk-rfcs/pull/100. However this is still in RFC.

I think it's premature to be retiring `govuk-link` right now.

heidar:
> > @misaka we use the rules from govuk-lint
> 
> Ya, I forgot that we do this in `.rubocop.yml`. Thing is, as we've found, this gem is being replaced by the gem `govuk-rubocop` in the PR [alphagov/govuk-rfcs#100](https://github.com/alphagov/govuk-rfcs/pull/100). However this is still in RFC.
> 
> I think it's premature to be retiring `govuk-link` right now.

We aren't retiring it, just using it via Rubocop. At least this way we can use Rubocop and maintain the same functionality of govuk-lint, making it a smoother transition when it does get fully replaced.

dankmitchell:
I think we should get this in so we’re consistent with the frontend but watch that RFC and see what GDS create govuk_rubocop.

PR #227, opened 2019-03-28
[1181] use admin contact in provider record if not in contacts table

misaka:
https://trello.com/c/WRv7HyW0/1181-provider-contacts-use-table-for-admin-contact-with-fallback

### Context

To facilitate the gradual transition of providers to the separation of front-office contacts (for candidates) and back-office admin contact (for UCAS), we want to support either retrieving the admin contact from the provider record (our old model) or retrieving the admin contact from the contacts table (the new model).

When serialising the provider use the admin contact from the provider record if present.

### DIFF
--- a/spec/serializers/provider_serializer_spec.rb
+++ b/spec/serializers/provider_serializer_spec.rb
@@ -87,13 +87,10 @@ describe ProviderSerializer do
 
   describe 'contacts' do
     describe 'generate provider object returns the providers contacts' do
-      let(:provider) do
-        create :provider,
-               contacts: [contact]
-      end
-      let(:contact) { create(:contact) }
+      let(:contact)  { create :contact }
+      let(:provider) { create :provider, contacts: [contact] }
 
-      subject { serialize(provider)['contacts'].find { |c| c[:type] == contact.type } }
+      subject { serialize(provider)['contacts'].first }
 
       its([:name]) { should eq contact.name }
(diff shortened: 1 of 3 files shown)


misaka:
[Provider contacts - use table for admin contact (with fallback)](https://trello.com/c/WRv7HyW0/1181-provider-contacts-use-table-for-admin-contact-with-fallback)

spec/serializers/provider_serializer_spec.rb
```diff
@@ -87,13 +87,10 @@
...
-      let(:provider) do
-        create :provider,
-               contacts: [contact]
-      end
-      let(:contact) { create(:contact) }
+      let(:contact)  { create :contact }
+      let(:provider) { create :provider, contacts: [contact] }
 
-      subject { serialize(provider)['contacts'].find { |c| c[:type] == contact.type } }
+      subject { serialize(provider)['contacts'].first }
```
defong: ain't this an admin?

misaka: Is it? This test isn't about admin and we should be able to change the default type in the factory without affecting this test.

If anything, we could get rid of this test, but then we might be getting rid of the `admin` test below at some point (when we no longer have the fallback behaviour) and not have any testing for contact serialisation in the provider serialiser.

In the end, I just didn't think it was worth worrying too much about. What do you think we should do?

PR #229, opened 2019-03-29
[1205] mcb for azure

timabell:
# Changes proposed in this pull request

* Make it possible to upload provider preferences to dev environment.
* Couple of minor improvements in separate commits

# Guidance to review

## read the updated docs

	bin/mcb -h
	bin/mcb az -h
	bin/mcb providers -h

## existing local behaviour

	bin/mcb providers list

	Providers:
	id | provider_code | provider_name | provider_type | postcode
	---|---------------|---------------|---------------|---------
	1  | A01           | Acme SCITT    |               |
	2  | A02           | Acme Alliance |               |
	...

note the above is seed data

	bin/mcb providers ucas_preferences show A01
	no preferences found

### find the app name

        bin/mcb az apps

### list providers on that app

	bin/mcb providers list --webapp bat-dev-manage-courses-api-app

	Providers:
	id    | provider_code | provider_name                  | provider_type | postcode
	------|---------------|--------------------------------|---------------|---------
	12955 | 27X           | XXXXXXXXX Junior School All... | lead_school   | XXXXXXX
	13033 | 213           | XXXXXXXXXLincolnshire Educa... | lead_school   | XXXXXXXX

### try out preference upload/download

	bin/mcb providers --webapp bat-dev-manage-courses-api-app ucas_preferences show 15W
	no preferences found

I don't have a test file for upload. Supply your own.

## new azure commands

* `bin/mcb az login` / `bin/mcb az logout` - run az login/logout for convenience
* `bin/mcb az apps` - list webapps - this was built on the way to using the data for finding the relevant database
* `bin/mcb az subs` - list subscriptions

## known issues

* The az commands are run twice due to the reload of the script with rails, this can be improved on but it works as is and the impact is only speed.
* untested with prod webapps / subscription

I say we merge as is then iterate.

----

@misaka please go ahead and merge this if you're happy with it


timabell:
I wonder if there's a better way of handling azure secrets and config that we should be using that the DevOps folks could enlighten us with. This might not be bash but it still feels a bit pyjama-code grabbing config from the webapp. Perhaps @defong or @OutKa5t might have contributions. (Caveat needing to get something that works shipped, I don't think we should block this merge even if there's a better way).

lib/mcb.rb
```diff
@@ -19,6 +19,8 @@ def self.init_rails
     unless defined?(Rails)
       app_root = File.expand_path(File.join(File.dirname($0), '..'))
       exec_path = File.join(app_root, 'bin', 'rails')
+      ENV["DISABLE_SPRING"] = "true" # prevent caching of environment variables by spring
```
dankmitchell: Not sure what the rest of the team think but I’d be happy to remove spring completely

lib/mcb/commands/providers/ucas_preferences/show.rb
```diff
@@ -8,7 +8,9 @@
 
   provider = Provider.find_by(provider_code: code)
 
-  if provider.ucas_preferences
+  if !provider
```
dankmitchell: Is there a cleaner way to right this?
If not, else if, else doesn’t read very well IMO

heidar:
How about some them tests? 🤔

lib/mcb.rb
misaka: Spring can be a PITA, but it does reduce start-up time. I'm on the fence on this one.

lib/mcb/commands/providers/ucas_preferences/show.rb
misaka: Not sure what a better alternative is.

heidar: `if provider.blank?` maybe? `unless` might be nicer but I tend to not like it when there's `else`/`elsif` involved.

misaka: :+1: for `if provider.blank?`.

lib/mcb.rb
misaka: Does rspec use Spring? If so, I'd be really tempted to keep it.

lib/mcb/azure/azure.rb
```diff
@@ -0,0 +1,32 @@
...
+    JSON.parse(raw_json)
+  end
+
+  def self.get_config(app, rgroup)
+    cmd = "az webapp config appsettings list -g #{rgroup} -n #{app}"
+    verbose(cmd)
+    raw_json = `#{cmd}`
+    config = JSON.parse(raw_json)
+    config.map{|c| k,v = [c["name"],c["value"]]}.to_h
+  end
```
misaka: Looks like some of the above 3 methods could be in a `private` method.

misaka: CodeClimate would flag this sort of thing. Wonder if it's worth integrating.

misaka:
Sorry, technical issues locally here, didn't mean to delete this branch.

OutKa5t:
> I wonder if there's a better way of handling azure secrets and config that we should be using that the DevOps folks could enlighten us with. This might not be bash but it still feels a bit pyjama-code grabbing config from the webapp. Perhaps @defong or @OutKa5t might have contributions. (Caveat needing to get something that works shipped, I don't think we should block this merge even if there's a better way).

Whats the question here? 
FYI @smneal @EwanNoble @chelnak

misaka:
> > I wonder if there's a better way of handling azure secrets and config that we should be using that the DevOps folks could enlighten us with. This might not be bash but it still feels a bit pyjama-code grabbing config from the webapp. Perhaps @defong or @OutKa5t might have contributions. (Caveat needing to get something that works shipped, I don't think we should block this merge even if there's a better way).
> 
> Whats the question here?
> FYI @smneal @EwanNoble @chelnak

@OutKa5t this app runs the `az` CLI to grab app settings to use locally. This is handy because it hands off user auth to `az`:

https://github.com/DFE-Digital/manage-courses-backend/blob/a33be5121d4a3178a51f44b847ebdfc6f76f1fac/lib/mcb/azure.rb#L15

I think @timabell's concern is that maybe there is a better way of getting these values, maybe out of VSTS secrets? I'm not sure myself and am happy to leave it as it is.

OutKa5t:
Yeah that doesnt sound sexy. Lets talk about it.

PR #232, opened 2019-03-31
[1202] add tag for bat environment for mcb

davidgisbey:
### Context

When error messages come through on sentry, there is no quick way to determine which env (dev, stag, prod) the error has originated from.

- Adds a bat_environment tag which shows the SENTRY_ENVIRONMENT variable at the top of the message. This env variable has been set in azure.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,5 +4,6 @@ class ApplicationController < ActionController::API
 
   # child must define authenticate method
   before_action :authenticate
+  before_action :bat_environment
   after_action :verify_authorized
 end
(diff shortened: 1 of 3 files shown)


davidgisbey:
Thanks Dan 👍. Caught me a sec before i fixed it 🤦‍♂️.

misaka:
As @dankmitchell is on the warpath to fix our `RAILS_ENV` settings, this may ultimately be unnecessary. Truth be told, that is the better approach, ++@dankmitchell, so we should hold fire on this PR.

dankmitchell:
Can now be closed.

PR #234, opened 2019-04-01
[1160] publish course

misaka:
### Context

When a course is updated, e.g. vacancy statuses are changed, it needs to be pushed out to the `search_and_compare` system.

Implement an API endpoint that can be used to trigger the syncing of a course to `search_and_compare`.

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -2,6 +2,7 @@ module API
   module V2
     class CoursesController < API::V2::ApplicationController
       before_action :build_provider
+      before_action :build_course, except: :index
 
       def index
         authorize @provider, :can_list_courses?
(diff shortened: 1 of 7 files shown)


misaka:
[Create endpoint in mcb to sync course with search and compare](https://trello.com/c/klynNtYt/1160-create-endpoint-in-mcb-to-sync-course-with-search-and-compare)

fofr:
Just to double check – can I confirm that this will have no effect on unpublished enrichments?

misaka:
> Just to double check – can I confirm that this will have no effect on unpublished enrichments?

This is the goal here, true, and we're just using the endpoint added here in this PR: https://github.com/DFE-Digital/manage-courses-api/pull/283 for the Trello card [Manage Courses API endpoint for triggering course-republishing](https://trello.com/c/ooKtcota/1132-manage-courses-api-endpoint-for-triggering-course-republishing). So the question should be asked of that endpoint, @defong can you enlighten us?

misaka:
@defong has verbally confirmed that unpublished changes won't be synced to search-and-compare.

defong:
> @defong has verbally confirmed that unpublished changes won't be synced to search-and-compare.

It calls this, https://github.com/DFE-Digital/manage-courses-api/blob/4e7319f8ca2cf5277b16b09c554c0e3937084b0b/src/ManageCourses.Api/Services/Publish/SearchAndCompareService.cs#L111


To build the final course, it only grabs the published enrichment of courses or nothing and it only grabs the published provider or the minimum details.

PR #235, opened 2019-04-01
[1213] suppress application alert recipient in provider serialiser if…

davidgisbey:
### Context
Previously, the provider serialiser added the application alert recipient if nil. This PR fixes that.

- Added a conditional which only adds the application alert recipient if the email is not nil

### Guidance to review
- I'm feeling really unsure about the change i made to Jake's previous test. If i left the test for "handles the missing data gracefully"  as:
`expect(subject['contacts'].detect { |c| c['type'] == 'application_alert_recipient' }['email']).to be_nil`
rather than: 
`expect(subject['contacts'].detect { |c| c['type'] == 'application_alert_recipient' }).to be_nil`

it fails, as it returns an empty array.

Would also appreciate some input on possibly renaming the generate_provider_contacts, generate_contacts and application_alert recipient method names.

### DIFF
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -41,7 +41,11 @@ class ProviderSerializer < ActiveModel::Serializer
              # see generate_provider_contacts and return_application_alert_recipient for logic.
 
   attribute :contacts do
-    generate_provider_contacts + application_alert_recipient
+    if application_alert_recipient
+      generate_provider_contacts + application_alert_recipient
+    else
+      generate_provider_contacts
+    end
   end
 
   def institution_code
@@ -96,11 +100,13 @@ private
   end
 
   def application_alert_recipient
-    [{
-      type: 'application_alert_recipient',
-      name: '',
-      email: object.ucas_preferences&.application_alert_email,
-      telephone: ''
-      }]
+    if object.ucas_preferences&.application_alert_email
+      [{
+        type: 'application_alert_recipient',
+        name: '',
+        email: object.ucas_preferences&.application_alert_email,
+        telephone: ''
+        }]
+    end
   end
 end
--- a/spec/requests/api/v1/provider_spec.rb
+++ b/spec/requests/api/v1/provider_spec.rb
@@ -80,7 +80,7 @@ describe 'Providers API', type: :request do
         build(:ucas_preferences,
               type_of_gt12: :coming_or_not,
               send_application_alerts: :none,
-              application_alert_email: 'application_alert_recipient@acmeuniversity.education.uk')
+              application_alert_email: nil)
       end
       let(:contacts2) do
         [
@@ -269,12 +269,6 @@ describe 'Providers API', type: :request do
                   'name' => 'Finance Contact B123',
                   'email' => 'finance@acmeuniversity.education.uk',
                   'telephone' => '01273 345 678'
-                },
-                {
-                  'type' => 'application_alert_recipient',
-                  'name' => '',
-                  'email' => 'application_alert_recipient@acmeuniversity.education.uk',
-                  'telephone' => ''
                 }
               ]
             }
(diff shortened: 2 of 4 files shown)


app/serializers/provider_serializer.rb
```diff
@@ -103,4 +103,12 @@ def application_alert_recipient
       telephone: ''
       }]
   end
+
+  def generate_contacts
+    if object.ucas_preferences&.application_alert_email != nil
```
dankmitchell: Don't think we need the `!= nil` here. Can we check?

spec/serializers/provider_serializer_spec.rb
```diff
@@ -57,7 +57,25 @@
     it { should eq provider.ucas_preferences.send_application_alerts_before_type_cast }
   end
 
-  context "when UCAS preferences are missing" do
+  describe 'application alert recipient' do
+    context 'if not nil the application alert recipient is serialized upon instantiation of the contact object' do
```
heidar: This doesn't feel like a context. Could we also reorganise this test to make it easier to follow?
```
context 'when set' do
  it 'does not serialise the application alert recipient' do
...
  end
end
```

spec/serializers/provider_serializer_spec.rb
```diff
@@ -57,7 +57,25 @@
...
+      subject { serialize(provider)['contacts'].find { |c| c[:type] == 'application_alert_recipient' } }
```
heidar: It might be good to break this up a bit? Readability feels like it could be better perhaps.
```
let(:application_alert_recipient) do
  serialize(provider)['contacts'].find { |contact| contact[:type] == 'application_alert_recipient' }
end

subject { application_alert_recipient }
```

app/serializers/provider_serializer.rb
davidgisbey: You're right. Thanks Dan.

spec/serializers/provider_serializer_spec.rb
davidgisbey: Have actioned both your changes. Thanks for your feedback Heidar 👍

app/serializers/provider_serializer.rb
```diff
@@ -41,7 +41,7 @@ class ProviderSerializer < ActiveModel::Serializer
              # see generate_provider_contacts and return_application_alert_recipient for logic.
 
   attribute :contacts do
-    generate_provider_contacts + application_alert_recipient
+    generate_contacts
```
misaka: Not sure it's worth having this as separate method at this point, unless it's used somewhere else?

spec/serializers/provider_serializer_spec.rb
```diff
@@ -57,16 +57,44 @@
...
+      it { should_not include 'application_alert_recipient' }
+    end
+  end
+
   context "when UCAS preferences are missing" do
     before do
       provider.ucas_preferences.destroy
       provider.reload
     end
+# no need to test for the application as it will not have been instantiated into the contacts object due to being nil
```
misaka: minor tick: indentation

spec/requests/api/v1/provider_spec.rb
```diff
@@ -80,7 +80,7 @@ def get_provider_codes_from_body(body)
         build(:ucas_preferences,
               type_of_gt12: :coming_or_not,
               send_application_alerts: :none,
-              application_alert_email: 'application_alert_recipient@acmeuniversity.education.uk')
+              application_alert_email: nil)
```
misaka: Why remove this from the test? Would it make sense to test this instead of the `nil` case?

davidgisbey: In this test two providers are parsed into JSON. Of these two, one has an application alert recipient and one doesn't. So it now tests both cases in one test.

PR #238, opened 2019-04-02
[1216] Add content status

benilovj:
### Context
The course enrichment status is shown to publishers in the course table

Add `content_status` to Course API V2.

### Guidance to review
This does a query per-course, which means N+1 queries on the courses table. We may need to consider performance going forward.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -104,4 +104,15 @@ class Course < ApplicationRecord
       study_mode_description
     qualifications_description + study_mode_string + program_type_description
   end
+
+  def enrichments
+    # This isn't implemented with pure ActiveRecord associations because for historic reasons,
+    # CourseEnrichment uses `course_code` and `provider_code` instead of
+    # `course_id` and `provider_id`. This should be fixed when the UCAS data model
+    # is revisited.
+    CourseEnrichment.where(
+      provider_code: self.provider.provider_code,
+      ucas_course_code: self.course_code
+    )
+  end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -55,6 +55,12 @@ class Course < ApplicationRecord
   has_many :site_statuses
   has_many :sites, through: :site_statuses
 
+  has_many :enrichments,
+           ->(course) { where(provider_code: course.provider.provider_code) },
+           foreign_key: :ucas_course_code,
+           primary_key: :course_code,
+           class_name: 'CourseEnrichment'
+
   scope :changed_since, ->(timestamp) do
     if timestamp.present?
       where("course.changed_at > ?", timestamp)
@@ -103,17 +109,6 @@ class Course < ApplicationRecord
     study_mode_string = (full_time_or_part_time? ? ", " : " ") +
       study_mode_description
     qualifications_description + study_mode_string + program_type_description
-  end
-
-  def enrichments
-    # This isn't implemented with pure ActiveRecord associations because for historic reasons,
-    # CourseEnrichment uses `course_code` and `provider_code` instead of
-    # `course_id` and `provider_id`. This should be fixed when the UCAS data model
-    # is revisited.
-    CourseEnrichment.where(
-      provider_code: self.provider.provider_code,
-      ucas_course_code: self.course_code
-    )
   end
 
   def content_status
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -115,4 +115,18 @@ class Course < ApplicationRecord
       ucas_course_code: self.course_code
     )
   end
+
+  def content_status
+    newest_enrichment = enrichments.order('created_at desc').first
+
+    if newest_enrichment.nil?
+      :empty
+    elsif newest_enrichment.published?
+      :published
+    elsif newest_enrichment.has_been_published_before?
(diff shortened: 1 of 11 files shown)


app/models/course_enrichment.rb
```diff
@@ -0,0 +1,23 @@
+# == Schema Information
+#
+# Table name: course_enrichment
+#
+#  id                           :integer          not null, primary key
+#  created_by_user_id           :integer
+#  created_at                   :datetime         not null
+#  provider_code                :text             not null
+#  json_data                    :jsonb
```
defong: Don't we need to map this back out?

defong:
Looking good, now why has travis gone on holiday...

app/models/course_enrichment.rb
benilovj: We don’t need it in this PR.

app/models/course.rb
```diff
@@ -104,4 +104,29 @@ def description
...
     qualifications_description + study_mode_string + program_type_description
   end
+
+  def enrichments
+    # This isn't implemented with pure ActiveRecord associations because for historic reasons,
+    # CourseEnrichment uses `course_code` and `provider_code` instead of
+    # `course_id` and `provider_id`. This should be fixed when the UCAS data model
+    # is revisited.
+    CourseEnrichment.where(
+      provider_code: self.provider.provider_code,
```
heidar: Redundant `self`

app/models/course.rb
```diff
@@ -104,4 +104,29 @@ def description
...
+      ucas_course_code: self.course_code
```
heidar: Redundant `self`

app/models/course.rb
```diff
@@ -104,4 +104,29 @@ def description
...
+    )
+  end
+
+  def content_status
+    newest_enrichment = self.enrichments.order('created_at desc').first
```
heidar: Redundant `self`

spec/factory_specs/course_spec.rb
```diff
@@ -1,18 +1,31 @@
...
-describe "Course Factory" do
-  let(:course) { create(:course) }
+describe "Course factory" do
+  subject { create(:course) }
 
-  it "created course" do
-    expect(course).to be_instance_of(Course)
-    expect(course).to be_valid
-  end
+  it { should be_instance_of(Course) }
```
heidar: Do we really need to test for this?

spec/models/course_enrichment_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+#  last_published_timestamp_utc :datetime
+#  status                       :integer          not null
+#  ucas_course_code             :text             not null
+#  updated_by_user_id           :integer
+#  updated_at                   :datetime         not null
+#
+
+require 'rails_helper'
+
+describe CourseEnrichment, type: :model do
```
heidar: We've set up RSpec to infer types, not that I mind it being explicit.

spec/models/course_enrichment_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  context 'initial draft' do
```
heidar: I feel like these contexts don't read very well. I think it would be clearer if it was "when the status is initial draft" or something like that?

spec/models/course_spec.rb
```diff
@@ -333,4 +333,31 @@
       it { should_not include(course) }
     end
   end
```
heidar: I don't see a unit test for `#enrichments`, which is a public method.

spec/factory_specs/course_spec.rb
misaka: It may be pedantic, but I like having tests that prove the factories work. Also, it's CHEAP!

spec/models/course_enrichment_spec.rb
benilovj: All the other specs specify type so I'd prefer to be consistent and remove type wholesale on a separate PR, if we want to do that.

app/models/course.rb
benilovj: This whole method has been removed.

app/models/course.rb
```diff
@@ -55,6 +55,12 @@ class Course < ApplicationRecord
   has_many :site_statuses
   has_many :sites, through: :site_statuses
 
+  has_many :enrichments,
```
heidar: very nice

app/models/course.rb
benilovj: This whole method has been removed.

spec/models/course_spec.rb
benilovj: added

spec/models/course_enrichment_spec.rb
benilovj: fixed

app/models/course.rb
benilovj: fixed

benilovj:
@heidar I've addressed all your comments. Ready for re-review.

@misaka I backed out your association additions to `CourseEnrichment` because they were causing the specs to fail (and anyway aren't needed as part of this change).

PR #240, opened 2019-04-02
[1234] optionally include sites with provider show endpoint

dankmitchell:
### Context
Training locations.

We need to expose `sites` on the provider#show endpoint in order to allow users access to update the `sites` aka training locations

- Optionally include `sites` on provider#show endpoint

### Guidance to review
Example url to test `localhost:3001/api/v2/providers/A9?include=sites`


dankmitchell:
[Providers API V2 can include sites](https://trello.com/c/Ac5ol3Aa/1234-providers-api-v2-can-include-sites)

PR #241, opened 2019-04-03
[1205] request confirmation when running against Azure app

misaka:
### Context

When running against a remote app's db there is always the possibility that an unexpected result would be had should the wrong environment be targeted.

Require the user to enter in the correct `RAILS_ENV` for the application that is being targeted.

### DIFF
--- a/spec/lib/providers_spec.rb
+++ b/spec/lib/providers_spec.rb
@@ -1,0 +1,46 @@
+require 'spec_helper'
+load 'bin/mcb'
+
+describe 'mcb providers' do
+  describe '--webapp option' do
+    let(:app_config) do
+      {
+        'RAILS_ENV' => 'aztest'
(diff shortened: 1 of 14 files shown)


spec/lib/providers_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        $mcb.run(%w[providers --webapp=banana])
+      end
+    end
+
+    it 'configures the database' do
+      subject
+
+      expect(MCB::Azure).to have_received(:get_config).with('banana')
+      expect(MCB::Azure).to have_received(:configure_database)
+                              .with('banana', app_config: app_config)
```
tvararu: nice 🍌

defong: 🍌 🍌 🍌

PR #242, opened 2019-04-03
[1210] Add ucas_status on course model and expose on serializer

tvararu:
### Context

:shipit:

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -79,6 +79,10 @@ class Course < ApplicationRecord
     site_statuses.findable.any?
   end
 
+  def new?
+    site_statuses.status_new_status.any?
+  end
+
   def open_for_applications?
     site_statuses.open_for_applications.any?
   end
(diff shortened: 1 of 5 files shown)


tvararu:
[EPIC: Courses table in manage-courses-frontend](https://trello.com/c/Ju2Kfadq/1210-epic-courses-table-in-manage-courses-frontend)

app/models/course.rb
```diff
@@ -124,4 +128,11 @@ def content_status
       :draft
     end
   end
+
+  def ucas_status
+    return :running if findable?
```
heidar: How come a course is running if it's findable? Are they the same concept?

misaka: Findable means it shows up on Find / Search and compare.
Running probably means it's on UCAS' Apply system.

tvararu: @benilovj instructed this and can explain better I imagine 😬

PR #243, opened 2019-04-03
1210] Allow include parameter on courses index lookup

tvararu:
This will be used to fetch accrediting providers in the frontend.

Also reverse the has_one to belongs_to on the serializable_course, because that's more logical.


tvararu:
[EPIC: Courses table in manage-courses-frontend](https://trello.com/c/Ju2Kfadq/1210-epic-courses-table-in-manage-courses-frontend)

PR #245, opened 2019-04-04
[1259] Expose provider in courses

heidar:
### Context

We need to expose `provider` and `accrediting_provider` on the course endpoint.

Rather than hard coding the course endpoint to return `site_statuses` and `sites`, we now rely on the `include` parameter.

Added a bunch of tests for the serializer to make sure we can render out everything we need.

### Guidance to review

I originally included a request spec in this that brings in provider and accredited_provider, but decided it was overkill in the end.

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -12,7 +12,7 @@ module API
       end
 
       def show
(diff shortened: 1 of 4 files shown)


heidar:
Let's not merge this until I've updated the vacancies controller on the frontend to explicitly request site statuses and sites from the backend, as this changes them to no longer be included by default.

PR #250, opened 2019-04-07
[1266] add address to sites serializer

dankmitchell:
### Context
sites aka training locations

- Add address data to sites serializer
- Setup `sites#index` and `sites#show`

### Guidance to review
e.g. `http://localhost:3001/api/v2/providers/T92?include=sites`

```ruby
{
  "id": "10324702",
  "type": "sites",
  "attributes": {
    "code": "O",
    "location_name": "Oakthorpe Primary School",
    "address1": "Tile Kiln Lane",
    "address2": "",
    "address3": "",
    "address4": "",
    "postcode": "N13 6BY",
    "region_code": "london"
  }
}
```

### DIFF
--- a/spec/requests/api/v2/providers/sites_spec.rb
+++ b/spec/requests/api/v2/providers/sites_spec.rb
@@ -1,0 +1,136 @@
+require 'rails_helper'
+
+describe 'Sites API v2', type: :request do
+  let(:user) { create(:user) }
+  let(:organisation) { create(:organisation, users: [user]) }
+  let(:payload) { { email: user.email } }
+  let(:token) { build_jwt :apiv2, payload: payload }
+  let(:credentials) do
+    ActionController::HttpAuthentication::Token.encode_credentials(token)
+  end
+
+  let(:site1) { create :site, location_name: 'Main site 1' }
+  let(:site2) { create :site, location_name: 'Main site 2' }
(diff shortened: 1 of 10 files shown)


spec/requests/api/v2/providers/sites_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+           sites: sites,
+           organisations: [organisation])
+  }
+
+  subject { response }
+
+  describe 'GET show' do
+    let(:show_path) do
+      "/api/v2/providers/#{provider.provider_code}" +
+        "/sites/#{site1.id}"
```
benilovj: Just had a thought about `site.id` . I wonder whether we may need to use campus code here, like we use provider code and campus code for the frontend URLs. I think @fofr’s prototype uses these in the URL for the site show/edit page. 

This introduces complications because the campus code data is inconsistent ie some main site campus codes are `-` and others are an empty string ` `, which would need translation in the URL :-(

Not 100% sure what the right answer is

dankmitchell: Exactly that @benilovj. We don't have unique `location_name`'s at present so I went with `id` to be safe.

benilovj: Ok lets go with that for now, can easily tweak if needed.

PR #253, opened 2019-04-08
[1265] Endpoint for updating sites

heidar:
### Context

Provider needs to be able to update site information.

And update endpoint for sites that allows updating all attributes, except `code`.

### DIFF
--- a/app/models/concerns/postcode_normalize.rb
+++ b/app/models/concerns/postcode_normalize.rb
@@ -2,7 +2,11 @@ module PostcodeNormalize
   extend ActiveSupport::Concern
   included do
     def postcode=(str)
-      super UKPostcode.parse(str).to_s
+      if str
+        super UKPostcode.parse(str).to_s
+      else
+        super str
+      end
     end
   end
 end
--- a/app/models/concerns/postcode_normalize.rb
+++ b/app/models/concerns/postcode_normalize.rb
@@ -1,0 +1,8 @@
+module PostcodeNormalize
+  extend ActiveSupport::Concern
+  included do
+    def postcode=(str)
+      super UKPostcode.parse(str).to_s
+    end
+  end
+end
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -23,11 +23,11 @@ class Site < ApplicationRecord
 
   belongs_to :provider
 
-  validates :location_name, uniqueness: { scope: :provider }
-  validates :postcode, postcode: true
+  validates :location_name, uniqueness: { scope: :provider_id }
   validates :location_name,
             :address1,
             :address3,
             :postcode,
             presence: true
+  validates :postcode, postcode: true
 end
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -21,4 +21,11 @@ class Site < ApplicationRecord
   include TouchProvider
 
   belongs_to :provider
+
+  validates :location_name, uniqueness: { scope: :provider }
+  validates :location_name,
+            :address1,
+            :address3,
+            :postcode,
+            presence: true
 end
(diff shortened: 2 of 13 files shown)


heidar:
Don't merge this yet, @tvararu is doing validations for the frontend that we can re-use in this PR. Should have that in tomorrow, otherwise we can merge this and get the validations in later. :)

heidar:
> LGTM
> 
> We will need separate PRs:
> * add validations

We're doing validations in this one, we'll do a second PR for the API call.

app/models/site.rb
```diff
@@ -21,4 +21,11 @@ class Site < ApplicationRecord
   include TouchProvider
 
   belongs_to :provider
+
+  validates :location_name, uniqueness: { scope: :provider }
```
defong: 👍

app/models/concerns/postcode_normalize.rb
```diff
@@ -0,0 +1,12 @@
+module PostcodeNormalize
+  extend ActiveSupport::Concern
+  included do
+    def postcode=(str)
+      if str
+        super UKPostcode.parse(str).to_s
+      else
+        super str
```
misaka: MINOR: It looks like we're guarding against running `UKPostcode.parse(nil)` here, but it's not tested below. How important is it to guard against this and is it worth testing?

spec/factories/sites.rb
```diff
@@ -24,7 +24,7 @@
     address2 { Faker::Address.community }
     address3 { Faker::Address.city }
     address4 { Faker::Address.state }
-    postcode { Faker::Address.postcode }
+    postcode { 'SW1A 2AA' }
```
misaka: Maybe set the Faker locale for this? Seems easy to do: https://github.com/stympy/faker/blob/master/lib/locales/README.md

app/models/concerns/postcode_normalize.rb
misaka: Explained by @tvararu: if we don't guard against `nil` here then many tests fail, so it's tested indirectly. LGTM.

heidar: I added a unit test anyway

spec/factories/sites.rb
tvararu: DOne

spec/rails_helper.rb
```diff
@@ -12,6 +12,8 @@
 require 'rspec/rails'
 # Add additional requires below this line. Rails is not loaded until this point!
 
+Faker::Config.locale = 'en-GB'
```
misaka: 💯

spec/requests/api/v2/providers/sites_spec.rb
```diff
@@ -133,4 +133,217 @@
...
+        )
+      end
+
+      it 'updates the location name of the site' do
+        expect { subject }.to change { site1.reload.location_name }
+          .from(site1.location_name)
+          .to(location_name)
+      end
+
+      it 'updates the code of the site' do
```
misaka: ITYM: DOES NOT update the code of the site :)

spec/requests/api/v2/providers/sites_spec.rb
```diff
@@ -133,4 +133,217 @@
...
+              expect(response.body).to include('Invalid postcode')
+              expect(response.body).to include("Postcode can't be blank")
+            end
+
+            it 'checks the postcode is present' do
+              expect(response.body).to include('Invalid postcode')
+              expect(response.body).to include('Postcode not recognised as a UK postcode')
+            end
+
+            it 'checks the region_code is present' do
```
misaka: Maybe use `xit` to flag this as a pending test?

PR #255, opened 2019-04-09
[1276] Vacancies should be derived only from open sites

benilovj:
### Context
Course vacancy logic currently looks at suspended and discontinued sites.

Fix the above, and make some spec improvements.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -88,7 +88,7 @@ class Course < ApplicationRecord
   end
 
   def has_vacancies?
-    site_statuses.with_vacancies.any?
+    site_statuses.open_for_applications.with_vacancies.any?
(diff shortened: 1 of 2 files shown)


fofr:
Good catch.

But what about the situation where applications are not yet open:
Screen Shot 2019-04-09 at 09 22 52
https://manage-courses-prototype.herokuapp.com/quiz

Technically there are vacancies they just can't be filled yet. 🤔

benilovj:
> Good catch.
> 
> https://manage-courses-prototype.herokuapp.com/quiz
> 
> Technically there are vacancies they just can't be filled yet. 🤔

@fofr thanks, fixed in c593945 #traincoding

PR #256, opened 2019-04-09
[1195] ucas integration testing

misaka:
### Context

The mcb command was missing some functionality that would let us poke and prod the APIs of production environments.

Add new mcb commands to further querying of APIs in production environments.

- Added `mcb apiv1 providers list` to list providers using the V1 API.
- Added `mcb apiv1 token` to retrieve the V1 token from deployed environments.

And modify existing commands to support querying production environments.

### Guidance to review

The test coverage is a bit sparse here, sorry, I'm happy to back-fill specs in the future but for now the focus needs to remain on doing integration testing.


misaka:
[Integration testing with UCAS Apply](https://trello.com/c/1TbQ5ogV/1195-editing-vacancies-integration-testing-with-ucas-apply)

PR #261, opened 2019-04-10
[1248] Expose course enrichments

dankmitchell:
### Context
Course enrichment data, this is required for the frontend course details view

- Optional include on courses endpoint

### Guidance to review
example: `http://localhost:3001/api/v2/providers/2AT/courses/35L8?include=enrichments`

TODO: Specs

### DIFF
--- a/app/serializers/api/v2/serializable_course_enrichment.rb
+++ b/app/serializers/api/v2/serializable_course_enrichment.rb
@@ -1,0 +1,11 @@
+module API
+  module V2
+    class SerializableCourseEnrichment < JSONAPI::Serializable::Resource
+      type 'course_enrichments'
+
+      attributes :json_data, :status, :has_been_published_before?
+
+      belongs_to :course
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


dankmitchell:
@benilovj good point. Let's discuss

defong:
max of 2, latest published, latest draft

app/serializers/api/v2/serializable_course_enrichment.rb
```diff
@@ -0,0 +1,11 @@
+module API
+  module V2
+    class SerializableCourseEnrichment < JSONAPI::Serializable::Resource
+      type 'course_enrichments'
+
+      attributes :json_data, :status, :has_been_published_before?
```
misaka: Should we really be exposing this as `json_data`? I think we should be hiding this particular implementation detail and exposing the attributes explicitly.

misaka: Can we add tests to `spec/serializers/api/v2/serializable_course_spec.rb` to ensure that we serialize the expected attributes?

misaka: The more I speak & think about it, it feels like the right thing to do here is to protect `manage-courses-frontend` from the enrichments as much as possible. I think we should explore exposing the course details in a 'published' form and a 'current' form (which would be the latest published or draft version).

PR #263, opened 2019-04-10
[1289] Publish course when location is edited

heidar:
### Context

Courses should be published to find when we edit a site.

Call the service to publish the course once a site has been successfully edited.

### Guidance to review

We don't care if this fails, so we silently handle it.


tvararu:
[Publish courses when a location is edited](https://trello.com/c/FirRqBbf/1289-publish-courses-when-a-location-is-edited)

PR #265, opened 2019-04-11
[1272] fix region codes in factories

davidgisbey:
### Context
The factories were calling a region_code method on the provider enrichment model with `ProviderEnrichment.region_code['London']`. As it was returning `nil` the serialiser tests were still passing. 

The provider factory has a default site count of 1. This causes factory bot issues that don't occur on live.
When the provider serialiser attempts to build the course it cannot convert the enum key ie. london into an integer.

This bug only affects our tests and not live.

- Changes the factories to generate region code by the enum key ie. london, eastern etc.
- Alters the providers in the provider serialiser spec to have `site_count 0` allowing serialisation without errors occurring 

### Guidance to review

Bit more context.

In the console 

`s = Site.create(region_code: 1)`
`p = Provider.create(site: [s])`
`p.sites.first.region_code`
returns 
`'london'` 
and 
`p.sites.first.region_code_before_type_cast` 
returns 
`1`

When you replicate this using factory bots create(:object) it returns 'london' for region_code and region_code_before_type_cast.

### DIFF
--- a/spec/serializers/provider_serializer_spec.rb
+++ b/spec/serializers/provider_serializer_spec.rb
@@ -30,7 +30,7 @@
 require "rails_helper"
 
 describe ProviderSerializer do
-  let(:provider) { create :provider }
+  let(:provider) { create :provider, site_count: 0 }
 
   subject { serialize(provider) }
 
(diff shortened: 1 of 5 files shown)


spec/serializers/provider_serializer_spec.rb
```diff
@@ -41,7 +41,7 @@
   it { should include(address3: provider.address3) }
   it { should include(address4: provider.address4) }
   it { should include(postcode: provider.postcode) }
-  it { should include(region_code: provider.region_code) }
+  it { should include(region_code: "%02d" % provider.region_code_before_type_cast) }
```
heidar: Can you use `.to_s` here instead?

heidar: oh never mind, totally misread what is happening here.

PR #269, opened 2019-04-12
[1269] subject mapper

defong:
### Context
This is a port of https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Api/Mapping/SubjectMapper.cs

Added Subject mapper

### Guidance to review
Barometer tis the port of https://github.com/DFE-Digital/manage-courses-api/blob/master/tests/ManageCourses.Tests/UnitTesting/SubjectMapperTests.cs


dankmitchell:
I had no idea it was this complex 🤯

Do we need to keep this logic around forever? 

Can we add a new column i.e. `find_course_name` (or rename the existing one to `ucas_course_name` and populate `course_name` with the mapped name) to the subject table and do a one off database migration?

benilovj:
@dankmitchell this can be simplified a lot, yes. When this was written originally in C#, I don't think we had the complete list of subjects so that's why some of the pattern-matching stuff is in place. We can replace it with static mappings, which will make things easier.

We should aim to replace this with a migration to DfE subjects imminently.

tvararu:
[Course - Basic detail tab SEND and Subject fields](https://trello.com/c/FN14LoBW/1269-course-basic-detail-tab-send-and-subject-fields)

tvararu:
[Course - Basic detail tab SEND and Subject fields](https://trello.com/c/FN14LoBW/1269-course-basic-detail-tab-send-and-subject-fields)

spec/services/subject_mapper_spec.rb
```diff
@@ -0,0 +1,235 @@
+require "spec_helper"
+require 'csv'
```
dankmitchell: Single vs double quotes 🤷🏻‍♂️👀

app/services/subject_mapper.rb
```diff
@@ -0,0 +1,239 @@
...
+                   "english literature"]
+
+  @ucas_primary = ["early years",
+                   "upper primary",
+                   "primary",
+                   "lower primary"]
+
+  @ucas_unexpected = [
+    "construction and the built environment",
+    # "history of art",
```
dankmitchell: Are these comments intentional?

spec/services/subject_mapper_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+        test_case: "secondary ESOL"
+      },
+      {
+        course_title: "PCET ESOL",
+        ucas_subjects: ["further education", "english as a second or other language"],
+        expected_subjects: ["Further education"],
+        test_case: "secondary ESOL"
+      },
+    ]
+        # public void MapToSearchAndCompareCourse(string courseTitle","string commaSeparatedUcasSubjects","string commaSeparatedExpectedSubjects)
```
dankmitchell: Is this line needed?

app/services/subject_mapper.rb
```diff
@@ -0,0 +1,239 @@
...
+    %w[Primary] + primary_subject_mappings.map { |mapping|
+      mapping.to_s if mapping.applicable_to?(ucas_subjects)
+    }.compact
+  end
+
+  def self.get_subject_list(course_title, ucas_subjects)
+    ucas_subjects = ucas_subjects.map { |subject| (subject.strip! || subject).downcase }
+    course_title = (course_title.strip! || course_title).downcase
+    # if unexpected throw.
+    if (ucas_subjects & @ucas_unexpected).any?
```
dankmitchell: Should this and be `&&`?

dankmitchell: Actually this is everywhere so I assume it’s correct. Never seen a single `&` before so I’ve learnt something new :)

app/services/subject_mapper.rb
defong: Comments are intentional, I guess they have previously been deliberated. Also it's in the original code base.

app/services/subject_mapper.rb
defong: It's intersection

spec/services/subject_mapper_spec.rb
defong: Good spot

spec/services/subject_mapper_spec.rb
defong: Argh @benilovj

app/services/subject_mapper.rb
defong: `[1,2,3] & [3,4,5]` == [3]`

app/services/subject_mapper.rb
defong: Given that there is a huge list of hardcoded list that needs to be considered for mapping/transforming/ignoring, I'm going to keep as is until we revisit this subject matter.

dankmitchell:
Can we squash some the commits? 26 😱

PR #271, opened 2019-04-15
[1305] Publish endpoint for courses

timabell:
Stub implementation of publishing.

This is enough for frontend to build on. The data modification is to
follow.

The request spec is a duplication of the sync spec as the same
constraints apply.

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -25,6 +25,12 @@ module API
         head response ? :ok : :internal_server_error
       end
 
+      def publish
+        # todo: set running
+        # todo: set enrichment published
+        sync_with_search_and_compare
+      end
+
     private
 
       def build_provider
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -28,7 +28,14 @@ module API
       def publish
         # todo: set running
         # todo: set enrichment published
-        sync_with_search_and_compare
+
+        response = ManageCoursesAPI::Request.sync_course_with_search_and_compare(
+          @current_user.email,
+          @provider.provider_code,
+          @course.course_code
+        )
+
+        head response ? :ok : :internal_server_error
       end
 
     private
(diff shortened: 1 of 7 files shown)


timabell:
[Rebuild course publishing in Rails](https://trello.com/c/wsUCpyJp/1305-rebuild-course-publishing-in-rails)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -25,6 +25,12 @@ def sync_with_search_and_compare
         head response ? :ok : :internal_server_error
       end
 
+      def publish
+        # todo: set running
+        # todo: set enrichment published
+        sync_with_search_and_compare
```
misaka: This smells, to me, is the intention to change this when the other logic is applied?

I don't think I've seen an action call another action like this. Probably because each action needs to build it's own response to an action, otherwise it should be an alias.

timabell: I'm not sure what an alias would look like, could you link to an example for my edification?

Intention is to try and get a PR approved :laughing: - but seriously, this is a bit of a straw-man, not sure what the best way of DRY'ing this would be or whether we should DRY it at all. We could duplicate the logic from the sync method.

misaka: I would duplicate the logic. One way to do this is to have a service that encapsulates it (e.g. `CourseSyncWithSearchAndCompareService`) independent of anything else like the params coming, how to generate a response, etc. Even in this case where the logic simply just defers to `ManageCoursesAPI::Request.sync_course_with_search_and_compare`, it would still be better, notionally, as a separate service.

tvararu:
[Course - basic detail tab](https://trello.com/c/izY1pEUm/1253-course-basic-detail-tab)

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -154,6 +154,75 @@
     end
   end
 
+  describe 'POST publish' do
+    let(:api_status)   { 200 }
+    let(:api_response) { '{ "result": true }' }
+    let!(:stubbed_manage_courses_api_request) do
```
defong: Why let! Instead of let ?

misaka: Good catch. Also, I'd say, `let` should be used for memoized values, not for setup. Especially if the setup is non-optional like here.

heidar:
I've rebased this and fixed the problem @defong pointed out. Can this be merged now, @timabell?

app/controllers/api/v2/courses_controller.rb
timabell: done the duplication for now as we discussed

PR #273, opened 2019-04-15
[1247] add update endpoint to user

dankmitchell:
### Context
In readiness for the transition screen (and t&c's) that will require a user to "update" their record.

- Add `users#update` controller action

### Guidance to review
Testing "time" is hard so had to remove the json match and check for individual attributes. Then using the `be_within` matcher is impossible to test when the json output is a string 🤷‍♂️ 

Open to feedback/suggestions..

### DIFF
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -1,12 +1,37 @@
 module API
   module V2
     class UsersController < API::V2::ApplicationController
-      def show
-        user = User.find(params[:id])
-        authorize user
+      before_action :build_user
+      deserializable_resource :user, only: :update
 
-        render jsonapi: user,
+      def show
+        render jsonapi: @user,
                include: params[:includes]
+      end
+
+      def update
+        @user.update(user_params)
+        render jsonapi: @user
+      end
+
+    private
+
+      def build_user
+        @user = authorize User.find(params[:id])
+      end
+
+      def user_params
+        params.require(:user).permit(
+          :email,
+          :first_name,
+          :last_name,
+          :first_login_date_utc,
+          :last_login_date_utc,
+          :sign_in_user_id,
+          :welcome_email_date_utc,
+          :invite_date_utc,
+          :accept_terms_date_utc,
+        )
       end
     end
   end
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -10,8 +10,11 @@ module API
       end
 
       def update
-        @user.update(user_params)
-        render jsonapi: @user
+        if @user.update(user_params)
+          render jsonapi: @user
+        else
+          render jsonapi_errors: @user.errors, status: 422
+        end
       end
 
     private
--- a/spec/requests/api/v2/users_spec.rb
+++ b/spec/requests/api/v2/users_spec.rb
@@ -72,7 +72,7 @@ describe '/api/v2/users', type: :request do
       Timecop.return
     end
 
-    context 'with accept_terms_date_utc' do
+    context 'when authenticated and authorised' do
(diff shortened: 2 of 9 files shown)


dankmitchell:
[Transition Screen](https://trello.com/c/LwxgvYQP/1247-transition-screen)

spec/requests/api/v2/users_spec.rb
```diff
@@ -48,21 +48,103 @@
...
+
+      subject { perform_request }
+
+      it 'updates email on the user' do
+        expect { subject }.to(change { user.reload.email }
+          .from(user.email).to(email))
+      end
+
+      it 'updates accept_terms_date_utc on the user' do
+        expect { subject }.to(change { user.reload.accept_terms_date_utc }
```
defong: Can't you set a specific time on user object, json stringify user, then check time?

dankmitchell: That is possible, yes 👍

app/controllers/api/v2/users_controller.rb
```diff
@@ -1,13 +1,38 @@
...
+      before_action :build_user
+      deserializable_resource :user, only: :update
 
-        render jsonapi: user,
+      def show
+        render jsonapi: @user,
                include: params[:includes]
       end
+
+      def update
```
misaka: How about we follow the pattern set out here, just on principle?

https://github.com/DFE-Digital/manage-courses-backend/blob/master/app/controllers/api/v2/sites_controller.rb#L16

We don't have validations yet but there could still be errors and, to some extent, it would be nice for the API to be built out a bit more like a platform instead of only feature-by-feature.

spec/requests/api/v2/users_spec.rb
```diff
@@ -48,21 +48,103 @@
...
+          .from(user.accept_terms_date_utc).to(be_within(1.second)
+          .of(accept_terms_date_utc)))
```
misaka: I found this indentation confusing. How about:

```
         expect { subject }.to(change { user.reload.accept_terms_date_utc }
           .from(user.accept_terms_date_utc)
           .to(be_within(1.second).of(accept_terms_date_utc)))
```

I'd even argue the other tests below and above could benefit from having the `from` and the `to` on separate lines ... my rule of thumb is that if the line is too long to chain all on one line, go all out and put each `.method` on a separate line (within reason, of course).

spec/requests/api/v2/users_spec.rb
misaka: I also got this to work like this:

```ruby
        expect { subject }.to(change { user.reload.accept_terms_date_utc.to_s }
                                .from(user.accept_terms_date_utc.to_s)
                                .to(accept_terms_date_utc.to_s))
```

app/controllers/api/v2/users_controller.rb
dankmitchell: Done

spec/requests/api/v2/users_spec.rb
dankmitchell: Done

spec/requests/api/v2/users_spec.rb
```diff
@@ -48,21 +48,121 @@
...
+      expect(data_attributes['first_name']).to eq(user.first_name)
+      expect(data_attributes['last_name']).to eq(user.last_name)
+    end
+  end
+
+  describe 'PATCH update' do
+    let(:params) { {} }
+
+    def perform_request
+      Timecop.freeze
```
heidar: I don't think we actually need this 🤔

heidar: Maybe it's safe to leave it in, just in case though. we can change it to use the block syntax. `Timecop.freeze` always scares the crap out of me. :P

PR #274, opened 2019-04-16
[1251] Add state machine to users for terms

dankmitchell:
### Context
To track when a user has accepted terms & conditions and confirmed transition information

- Add [State machine](https://github.com/aasm/aasm)
- Setup "event" that can be called when a user accepted the t&c's `@current_user.accept_terms`
- I propose we get this merged then move the t&c's into this stack

> For the transition screen, we can simply add a new event to set the state to `accepted_transition` once the user has hit continue

### Guidance to review
```ruby
u = User.find_by(email: "tim.abell+4@digital.education.gov.uk")
u.accept_terms
u

{
    "data": {
        "id": "10396",
        "type": "users",
        "attributes": {
            "first_name": "Tim4",
            "last_name": "Abell",
            "email": "tim.abell+4@digital.education.gov.uk",
            "aasm_state": "accepted_terms"
        }
    },
    "jsonapi": {
        "version": "1.0"
    }
}

```

### DIFF
--- a/app/serializers/api/v2/serializable_user.rb
+++ b/app/serializers/api/v2/serializable_user.rb
@@ -4,6 +4,10 @@ module API
       type 'users'
 
       attributes :first_name, :last_name, :email, :accept_terms_date_utc
+
+      attribute :state do
+        @object.aasm_state
+      end
     end
   end
 end
--- a/db/migrate/20190416093524_add_aasm_state_to_users.rb
+++ b/db/migrate/20190416093524_add_aasm_state_to_users.rb
@@ -1,0 +1,5 @@
+class AddAasmStateToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :user, :aasm_state, :string
+  end
+end
--- a/db/migrate/20190416190949_initialize_state_for_users.rb
+++ b/db/migrate/20190416190949_initialize_state_for_users.rb
@@ -1,0 +1,18 @@
+class InitializeStateForUsers < ActiveRecord::Migration[5.2]
+  def up
+    User.find_each do |user|
+      user.activated! if user.accept_terms_date_utc.present?
+    end
+  end
+
+  def down
+    User.find_each do |user|
+      next unless user.active?
+
+      user.update(
+        accept_terms_date_utc: Time.current,
+        aasm_state:            'new'
+      )
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_04_09_230916) do
+ActiveRecord::Schema.define(version: 2019_04_16_093524) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_buffercache"
@@ -246,6 +246,7 @@ ActiveRecord::Schema.define(version: 2019_04_09_230916) do
     t.datetime "welcome_email_date_utc"
     t.datetime "invite_date_utc"
     t.datetime "accept_terms_date_utc"
+    t.string "aasm_state"
     t.index ["email"], name: "IX_user_email", unique: true
   end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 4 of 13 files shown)


dankmitchell:
Done @benilovj but once we've run that migration we need to think about future users coming onto the app and accepting the t&c's.

Can we prioritize moving t&c's into this app 🤓 ?

db/schema.rb
```diff
@@ -246,6 +246,7 @@
     t.datetime "welcome_email_date_utc"
     t.datetime "invite_date_utc"
     t.datetime "accept_terms_date_utc"
+    t.string "aasm_state"
```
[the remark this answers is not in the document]
dankmitchell: `aasm_state` is the default. Can change to `state` just means adding a line of code into the model to look for this instead of the default; I was going for less code but happy to update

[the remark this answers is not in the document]
defong: out of the box it be then

aasm => adding a state machine

db/migrate/20190416190949_initialize_state_for_users.rb
```diff
@@ -0,0 +1,11 @@
+class InitializeStateForUsers < ActiveRecord::Migration[5.2]
+  def change
+    User.find_each do |user|
```
heidar: This cannot live in a `#change` as this cannot be rolled back

db/migrate/20190416093524_add_aasm_state_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddAasmStateToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :user, :aasm_state, :string
```
heidar: Should this be nullable?

heidar:
> Done @benilovj but once we've run that migration we need to think about future users coming onto the app and accepting the t&c's.
> 
> Can we prioritize moving t&c's into this app 🤓 ?

Is this currently all done in the C# apps, which only updates the old DB columns?

db/migrate/20190416093524_add_aasm_state_to_users.rb
dankmitchell: No. The initial state set in the model covers that.

dankmitchell:
@heidar at present the c# UI just adds a timestamp into the `accept_terms_date_utc` column so we need to move the feature into this repo so we can use the state machine stuff in this PR.

app/serializers/api/v2/serializable_user.rb
```diff
@@ -3,7 +3,7 @@ module V2
     class SerializableUser < JSONAPI::Serializable::Resource
       type 'users'
 
-      attributes :first_name, :last_name, :email
+      attributes :first_name, :last_name, :email, :aasm_state
```
heidar: Not a big fan of this attribute name to be honest :(

heidar:
@misaka does that WIP flag mean we're not merging this until we have moved the functionality over?

db/migrate/20190416190949_initialize_state_for_users.rb
```diff
@@ -0,0 +1,18 @@
...
+  def down
+    User.find_each do |user|
+      next unless user.active?
+
+      user.update(
+        accept_terms_date_utc: Time.current,
+        aasm_state:            'new'
+      )
+    end
+  end
```
misaka: As @heidar pointed out elsewhere, here's a good opportunity to add `data_migrate` and `strong_migrations` gems, and to move this into a data migration.

tvararu: No, let's do that as part of a separate PR

heidar:
What are we doing with this PR? Are we not merging it until we manage this data from the frontend? @benilovj @misaka @dankmitchell

benilovj:
@heidar we want to pause on this one until #284 is merged, because:

- #284 one's more urgently needed for functionality we need to ship this week
- there's dependencies on `manage-courses-ui` for the T&C work that we haven't resolved yet

We can rebase this PR later on this week.

heidar:
Seems like 

> @heidar we want to pause on this one until #284 is merged, because:
> 
> * #284 is more urgently needed for functionality we need to ship this week
> * there's dependencies on `manage-courses-ui` for the T&C work that we haven't resolved yet
> 
> We can rebase this PR later on this week.

Seems #284 reimplements this slightly differently. Can we close this in favour of #284 then?

heidar:
Closing in favour of #284. @tvararu for king.

PR #279, opened 2019-04-17
[1315] Add applications_open_from to course data

tvararu:
### Context

We're exposing more course information for the courses show view.

- Add applications_open_from to course data

### Guidance to review

🚢

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -90,6 +90,17 @@ class Course < ApplicationRecord
     site_statuses.open_for_applications.any?
   end
 
+  def applications_open_from
+    site_statuses
+      .open_for_applications
+      .order("applications_accepted_from ASC")
+      .first
+      &.applications_accepted_from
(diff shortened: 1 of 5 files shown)


tvararu:
[Course basic details: applications open (backend)](https://trello.com/c/4hgkoc6J/1315-course-basic-details-applications-open-backend)

app/models/course.rb
```diff
@@ -90,6 +90,17 @@ def open_for_applications?
     site_statuses.open_for_applications.any?
   end
 
+  def applications_open_from
+    site_statuses
+      .open_for_applications
```
fofr: Does this need to be tested? eg Something showing that it picks the earliest application open date?

tvararu: I didn't test that particular branch, because I don't think it's at a high risk of breaking… if more people feel strongly for it I can add some.

PR #282, opened 2019-04-18
[1269] Simplify subject mapper

benilovj:
## Context

The `SubjectMapper` is difficult to understand and reason about.

## Proposed changes

- extract `*Mapping` classes for particularly gnarly bits of logic so they're a lot more encapsulated
- bring the code closer to common ruby conventions (it started its life as C#)

## Code review guidance

image

### DIFF
--- a/app/services/subject_mapper.rb
+++ b/app/services/subject_mapper.rb
@@ -87,7 +87,7 @@ class SubjectMapper
     },
   }.freeze
 
-  class GroupedSubjectMapping
+  class StaticMapping
     def initialize(included_ucas_subjects, resulting_dfe_subject)
       @included_ucas_subjects = included_ucas_subjects
       @resulting_dfe_subject = resulting_dfe_subject
@@ -174,7 +174,7 @@ class SubjectMapper
 
   def self.map_to_secondary_subjects(course_title, ucas_subjects)
     secondary_subject_mappings = MAPPINGS[:secondary].map do |ucas_input_subjects, dfe_subject|
-      GroupedSubjectMapping.new(ucas_input_subjects, dfe_subject)
+      StaticMapping.new(ucas_input_subjects, dfe_subject)
     end
 
     secondary_subject_mappings += [
@@ -207,7 +207,7 @@ class SubjectMapper
 
   def self.map_to_primary_subjects(ucas_subjects)
     primary_subject_mappings = MAPPINGS[:primary].map do |ucas_input_subjects, dfe_subject|
-      GroupedSubjectMapping.new(ucas_input_subjects, dfe_subject)
+      StaticMapping.new(ucas_input_subjects, dfe_subject)
     end
 
     %w[Primary] + primary_subject_mappings.map { |mapping|
--- a/app/services/subject_mapper.rb
+++ b/app/services/subject_mapper.rb
@@ -98,11 +98,6 @@ class SubjectMapper
     },
   }.freeze
 
-  def self.is_further_education(subjects)
-    subjects = subjects.map { |subject| (subject.strip! || subject).downcase }
-    (subjects & SUBJECT_LEVEL[:ucas_further_education]).any?
-  end
-
   class GroupedSubjectMapping
     def initialize(included_ucas_subjects, resulting_dfe_subject)
       @included_ucas_subjects = included_ucas_subjects
--- a/app/services/subject_mapper.rb
+++ b/app/services/subject_mapper.rb
@@ -20,10 +20,6 @@ class SubjectMapper
                       "law"]
   }.freeze
 
-  @ucas_english = ["english",
-                   "english language",
-                   "english literature"]
-
   MAPPINGS = {
     primary: {
       ["english", "english language", "english literature"] => "Primary with English",
@@ -152,8 +148,15 @@ class SubjectMapper
       end
     end
 
+    # TODO: remove this bonkers logic once course mapping is done by one app!
+    # The user need for this is unclear
+    #
     # Does the subject list mention english, and it's mentioned in the title (or it's the only subject we know for this course)?
-    if (ucas_subjects & @ucas_english).any?
+    ucas_english = ["english",
+                    "english language",
+                    "english literature"]
+
+    if (ucas_subjects & ucas_english).any?
       if secondary_subjects.none? || course_title.index("english") != nil
         secondary_subjects.push("English")
       end
--- a/app/services/subject_mapper.rb
+++ b/app/services/subject_mapper.rb
@@ -134,6 +134,31 @@ class SubjectMapper
     end
   end
 
+  # TODO: remove this bonkers logic once course mapping is done by one app!
+  # The user need for this is unclear
+  class SecondaryEnglishMapping
+    def initialize(course_title)
+      @course_title = course_title
+    end
+
+    def applicable_to?(ucas_subjects)
+      (ucas_subjects & ucas_english).any? &&
+        @course_title.index("english") != nil
+    end
+
+    def to_s
+      "English"
+    end
+
+  private
+
+    def ucas_english
+      ["english",
+       "english language",
+       "english literature"]
+    end
+  end
+
   def self.map_to_secondary_subjects(course_title, ucas_subjects)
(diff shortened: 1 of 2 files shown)


spec/services/subject_mapper_spec.rb
```diff
@@ -136,7 +73,7 @@
       },
       {
         course_title: "",
-        ucas_subjects: %w[secondary welsh],
+        ucas_subjects: %w[welsh],
```
defong:
```suggestion
ucas_subjects: %w[secondary welsh],
```

put that back in as its a port from csharp

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,112 @@ def to_s
...
-      if secondary_subjects.none? || course_title.index("english") != nil
-        secondary_subjects.push("English")
-      end
+    def to_s
+      "English"
+    end
+
+  private
+
+    def ucas_english
```
defong: Can we shove that in  the `Mapping` and remove SecondaryEnglishMapping class altogether?

app/services/subject_mapper.rb
```diff
@@ -108,18 +87,7 @@ class SubjectMapper
...
-  end
-
-  def self.map_to_subject_name(ucas_subject)
-    res = (@ucas_rename[ucas_subject] || ucas_subject).capitalize
-
-    (res.sub "english", "English" || res)
-  end
-
-  class GroupedSubjectMapping
+  class StaticMapping
```
heidar: Can we keep it at one class per file? I also feel like `SubjectMapper::StaticMapping` doesn't tell us much. Can we use naming that makes it clear what we are mapping?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
     end
   end
 
-  def self.map_to_secondary_subjects(course_title, ucas_subjects)
-    secondary_subject_mappings = MAPPINGS[:secondary].map do |ucas_input_subjects, dfe_subject|
-      GroupedSubjectMapping.new(ucas_input_subjects, dfe_subject)
+  #  Does the subject list mention languages but hasn't already been covered?
+  class MFLOtherMapping
```
heidar: Can we keep it at one class per file? Also, would be nice to avoid jargon like `MFL` - at least it's very ambiguous to me what it means.

spec/services/subject_mapper_spec.rb
benilovj: Welsh logic completely excised in f7b1d63 (it's no longer needed)

app/services/subject_mapper.rb
benilovj: unfortunately not yet – some of the prod data specs fail when I tried it, I believe. I'd prefer to do this simplification on both the C# and the Rails side in a short while.

defong: It's is a deeper rabbit hole, another time, 

Basically `English` mapping relies on other mapping to firstly before attempting to map `English`

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
+    def applicable_to?(ucas_subjects)
+      (ucas_subjects & language_categories).any? &&
```
heidar: Would it be more readable if these were extracted into 3 methods?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
+      "Modern languages (other)"
     end
 
-    secondary_subjects = []
+  private
 
-    secondary_subjects += secondary_subject_mappings.map { |mapping|
-      mapping.to_s if mapping.applicable_to?(ucas_subjects)
-    }.compact
+    def language_categories
```
heidar: Is there any reason to not make these constants?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
+
+    def mandarin
+      %w[chinese mandarin]
+    end
 
-    ucas_language_cat = ["languages",
-                         "languages (african)",
-                         "languages (asian)",
-                         "languages (european)"]
+    def mfl_main
```
heidar: Modern foreign languages main?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
+       "spanish"]
+    end
+  end
 
-    ucas_mfl_mandarin = %w[chinese mandarin]
+  # TODO: remove this bonkers logic once course mapping is done by one app!
+  # The user need for this is unclear
+  #
+  # Does the subject list mention english, and it's mentioned in the title (or it's the only subject we know for this course)?
+  class SecondaryEnglishMapping
```
heidar: Can we keep it at one class per file?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
 
-    ucas_mfl_main = ["english as a second or other language",
-                     "french",
-                     "german",
-                     "italian",
-                     "japanese",
-                     "russian",
-                     "spanish"]
+    def applicable_to?(ucas_subjects)
+      (ucas_subjects & ucas_english).any? &&
```
heidar: Can this logic be extracted to private methods to make this more readable?

app/services/subject_mapper.rb
heidar: Any reason not to make this a constant?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
+       "english literature"]
     end
+  end
 
-    # Does the subject list mention english, and it's mentioned in the title (or it's the only subject we know for this course)?
-    if (ucas_subjects & @ucas_english).any?
-      if secondary_subjects.none? || course_title.index("english") != nil
-        secondary_subjects.push("English")
-      end
+  def self.map_to_secondary_subjects(course_title, ucas_subjects)
```
heidar: Map....what? `map_ucas_subject_to_dfe_subject`?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
+
+    secondary_subjects = secondary_subject_mappings.map { |mapping|
+      mapping.to_s if mapping.applicable_to?(ucas_subjects)
+    }.compact
+
+    # TODO: remove this bonkers logic once course mapping is done by one app!
+    # There is absolutely no user need for it!
+    #
+    # Does the subject list mention a subject we are happy to translate if the course title contains a mention?
+    ucas_needs_mention_in_title = {
```
heidar: This feels like a constant?

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
-    if secondary_subjects.none? && (ucas_subjects & %w[welsh]).any?
-      secondary_subjects.push("Welsh")
+    secondary_subject_mappings += [
+      MFLOtherMapping.new,
+      SecondaryEnglishMapping.new(course_title),
+    ]
+
+    secondary_subjects = secondary_subject_mappings.map { |mapping|
+      mapping.to_s if mapping.applicable_to?(ucas_subjects)
+    }.compact
```
heidar: It's not entirely clear to me what the intent is with the lines above here.

app/services/subject_mapper.rb
```diff
@@ -134,62 +102,101 @@ def to_s
...
+
+    # TODO: remove this bonkers logic once course mapping is done by one app!
+    # There is absolutely no user need for it!
+    #
+    # Does the subject list mention a subject we are happy to translate if the course title contains a mention?
+    ucas_needs_mention_in_title = {
+      "humanities" => /humanities/,
+      "science" => /(?<!social |computer )science/,
+    }
+    (ucas_subjects & %w[humanities science]).each do |ucas_subject|
```
heidar: Does this need to be a loop?

heidar:
Looks like I missed the train again, oh well.

benilovj:
@heidar I'll do a follow-up PR to address your comments.

PR #283, opened 2019-04-18
[1327] Added level to the course serializer

defong:
### Context
Course level can only be really one of 3, `primary`, `secondary` and `further education`

Added course level derived from course subject and/or course title

### DIFF
--- a/app/services/subject_mapper.rb
+++ b/app/services/subject_mapper.rb
@@ -198,6 +198,7 @@ class SubjectMapper
   end
 
   def self.get_subject_level(ucas_subjects)
+    ucas_subjects = ucas_subjects.map { |subject| (subject.strip! || subject).downcase }
     if (ucas_subjects & SUBJECT_LEVEL[:ucas_unexpected]).any?
       "found unsupported subject name(s): #{(ucas_subjects & SUBJECT_LEVEL[:ucas_unexpected]) * ', '}"
     elsif (ucas_subjects & SUBJECT_LEVEL[:ucas_primary]).any?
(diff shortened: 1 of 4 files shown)


spec/serializers/api/v2/serializable_course_spec.rb
```diff
@@ -93,4 +93,30 @@
       it { expect(subject["attributes"]).to include("is_send?" => true) }
     end
   end
+
+  context "subjects & level" do
+    context 'with no subjects' do
+      let(:course) { create(:course, subject_count: 0) }
+      it { expect(subject["attributes"]).to include("level" => "secondary") }
```
tvararu: Is this the default we want?

defong: No idea, but previously it would of been a secondary welsh corpse.

There should be more subject matter work, but current aim is to make it work and refine later.

@benilovj to clarify

defong:
[Course - Basic detail tab Level field (backend)](https://trello.com/c/8hP7I8ic/1327-course-basic-detail-tab-level-field-backend)

spec/serializers/api/v2/serializable_course_spec.rb
defong: checked in meatsphere, its fine

app/services/subject_mapper.rb
```diff
@@ -198,6 +198,7 @@ def self.map_to_primary_subjects(ucas_subjects)
   end
 
   def self.get_subject_level(ucas_subjects)
+    ucas_subjects = ucas_subjects.map { |subject| (subject.strip! || subject).downcase }
```
benilovj:
```suggestion
    ucas_subjects = ucas_subjects.map(&:strip).map(&:downcase)
```

PR #284, opened 2019-04-18
[1251] Implement state machine for user, add transitioned state

tvararu:
### Context

This is more or less a fork of https://github.com/DFE-Digital/manage-courses-backend/pull/274.

- Implements the state machine pattern on the users model, with initial states around the users `transitioned`

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,7 +12,7 @@
 #  welcome_email_date_utc :datetime
 #  invite_date_utc        :datetime
 #  accept_terms_date_utc  :datetime
-#  state                  :string
+#  state                  :string           not null
 #
 
 class User < ApplicationRecord
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -12,10 +12,12 @@
 #  welcome_email_date_utc :datetime
 #  invite_date_utc        :datetime
 #  accept_terms_date_utc  :datetime
-#  aasm_state             :string
+#  state                  :string
 #
 
 class User < ApplicationRecord
+  include AASM
+
   has_and_belongs_to_many :organisations
   has_many :providers, through: :organisations
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -24,4 +24,13 @@ class User < ApplicationRecord
   validates :email, presence: true
 
   audited
+
+  aasm column: 'state' do
+    state :new, initial: true
+    state :transitioned
+
+    event :accept_transition_screen do
+      transitions from: :new, to: :transitioned
+    end
+  end
 end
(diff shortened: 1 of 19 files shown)


db/data/20190418114011_initialise_state_for_users.rb
```diff
@@ -0,0 +1,15 @@
...
+    User.find_each do |user|
+      user.save!
+    end
+  end
+
+  def down
+    User.find_each do |user|
+      next unless user.new?
+
+      user.update(aasm_state: '')
```
dankmitchell: Not sure this is right. There will always be an initial state set in the model i.e. `state :new, initial: true`

db/data/20190418114011_initialise_state_for_users.rb
```diff
@@ -0,0 +1,15 @@
+class InitialiseStateForUsers < ActiveRecord::Migration[5.2]
+  def up
+    User.find_each do |user|
```
dankmitchell: Could just be a oneliner `User.find_each(&:save)`

tvararu: Done already by rubocop

db/data/20190418114011_initialise_state_for_users.rb
tvararu: If we do that then it's not a full reverse-migration; the initial database state is empty string. 🤔

app/models/user.rb
```diff
@@ -12,13 +12,25 @@
...
+
   has_and_belongs_to_many :organisations
   has_many :providers, through: :organisations
 
   validates :email, presence: true
 
   audited
+
+  aasm column: 'state' do
+    state :new, initial: true
```
heidar: Would it be better to name these states based on what we use them for? I think we only use them for terms and conditions?

db/data/20190418114011_initialise_state_for_users.rb
```diff
@@ -0,0 +1,13 @@
+class InitialiseStateForUsers < ActiveRecord::Migration[5.2]
+  def up
+    User.find_each(&:save!)
```
heidar: It'd be more explicit to make this set `state` to `User.aasm.initial_state`. Not a dealbreaker though. :)

app/models/user.rb
tvararu: Discussed offline, it's going to be used for Ts and Cs as well but it's ok.

db/data/20190418114011_initialise_state_for_users.rb
tvararu: Yes, agree

PR #286, opened 2019-04-18
[1305] course publishing implementation

timabell:
# Context

Port from C#. Pairing David+Tim

part of this was in #271 which I think we'll close now

## Why are we doing this?

* To support the rebuilding of the course enrichment page
* and turning off the C# apps.

Prototype: https://manage-courses-prototype.herokuapp.com/course/2B2/2TG7

This page has a "Publish" button.

## Success

- when the course publishing is triggered, draft enrichments are marked as published (feature parity with C#)
- when the course publishing is triggered, any new site statuses are set to running and publish='Y' (feature parity with C#)
- when the course publishing is triggered, any suspended or discontinued site statuses are ignored (feature parity with C#)
- trigger the existing POST to search and compare
- `changed_at` is updated so that changes show up in the api for ucas
- don't need to do validations as part of this work that's in another card

## Existing code/behaviour

* Publish button posts to https://bat-dev-manage-courses-ui-app.azurewebsites.net/organisation/2cg/course/self/2z28 which is implemented by `ShowPublish()` https://github.com/DFE-Digital/manage-courses-ui/blob/master/src/ui/Controllers/CourseController.cs#L61
  * there's a lot of logic in here including validation
* this posts to manage api, calling `PublishCourseToSearchAndCompare()` https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Api/Controllers/PublishController.cs#L39 which sets courses to running, sets up the enrichment publish status, and calls the push-to-search method (confusingly named Save)
  * this publishes the enrichments https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Api/Services/EnrichmentService.cs#L107

# Changes proposed in this pull request

* Setup enrichment & site statuses when the publish method is called.
* Allow the factory to create older enrichments to ensure tests are not coincidentally passing.

# Guidance to review

The tests work apart from a known issue with the factories resulting in stale data that breaks one of the tests

The style of the tests needs input from a more experienced rails dev.

Draft until the rspec is green.


spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
       its(:content_status) { should eq(:published_with_unpublished_changes) }
     end
   end
+
+  describe "#publish_enrichments" do
```
dankmitchell: I think this should live in its own file `course_publish_spec.rb`

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -151,6 +151,101 @@
     end
   end
 
+  describe 'POST publish' do
```
dankmitchell: Same with this. Move to its own file

app/models/course.rb
```diff
@@ -148,4 +148,31 @@ def funding
...
+
+  def publish_sites
+    update_new_to_running
+    update_published_if_running
+    # as update_all is called and doesn't trigger a save this has to be manually tested
+    update(changed_at: Time.now.utc)
+  end
+
+  def publish_enrichment(current_user_id)
+    enrichments.where(status: 'draft')
```
defong: potential concurrency issues

change it to make it pick up the latests draft only

```csharp
                var enrichmentPublishRecord = _context.ProviderEnrichments
                    .Where(ie => ie.ProviderCode == providerCode && ie.Status == EnumStatus.Published)
                    .OrderByDescending(x => x.Id)
                    .FirstOrDefault();

```

timabell:
[Rebuild course publishing in Rails](https://trello.com/c/wsUCpyJp/1305-rebuild-course-publishing-in-rails)

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -151,6 +151,101 @@
...
+               provider: provider,
+               with_site_statuses: [:new],
+               with_enrichments: [:initial_draft],
+               age: 17.days.ago)
+      }
+      it 'publishes a course' do
+        # run the POST
+        response = subject
+
+        expect(course.site_statuses.first.status).to eq 'running'
```
defong: `course.content_status`

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+        ].each do |test_case|
+          site = create(:site_status, status: test_case[:before][:status], publish: test_case[:before][:publish])
+          subject = create(:course, site_statuses: [site])
+          subject.publish_sites
+          expect(subject.site_statuses.first.status).to eq test_case[:after][:status]
+          expect(subject.site_statuses.first.publish).to eq test_case[:after][:publish]
+        end
+      end
+    end
+    # xcontext 'course with multiple sites in different states'
```
defong: leftover

app/models/course.rb
timabell: concurrency issues live on the c# side that generates the draft. There is only ever one draft so no issue here.

spec/requests/api/v2/providers/courses_spec.rb
timabell: independent concern

timabell:
current failure due to factory issue:

```
Failures:

  1) Course publish_sites course with a single site correctly applies all state transitions
     Failure/Error: expect(subject.site_statuses.first.status).to eq test_case[:after][:status]

       expected: "running"
            got: "new_status"

       (compared using ==)
       
     # ./spec/models/course_spec.rb:529:in `block (5 levels) in <top (required)>'
     # ./spec/models/course_spec.rb:525:in `each'
     # ./spec/models/course_spec.rb:525:in `block (4 levels) in <top (required)>'
     # ./spec/rails_helper.rb:61:in `block (3 levels) in <top (required)>'
     # /home/tim/.rvm/gems/ruby-2.5.1/gems/database_cleaner-1.7.0/lib/database_cleaner/generic/base.rb:16:in `cleaning'
     # /home/tim/.rvm/gems/ruby-2.5.1/gems/database_cleaner-1.7.0/lib/database_cleaner/base.rb:100:in `cleaning'
     # /home/tim/.rvm/gems/ruby-2.5.1/gems/database_cleaner-1.7.0/lib/database_cleaner/configuration.rb:86:in `block (2 levels) in cleaning'
     # /home/tim/.rvm/gems/ruby-2.5.1/gems/database_cleaner-1.7.0/lib/database_cleaner/configuration.rb:87:in `cleaning'
     # ./spec/rails_helper.rb:60:in `block (2 levels) in <top (required)>'

Finished in 30.36 seconds (files took 2.06 seconds to load)
1785 examples, 1 failure, 2 pending

Failed examples:

rspec ./spec/models/course_spec.rb:515 # Course publish_sites course with a single site correctly applies all state transitions
```

spec/models/course_spec.rb
misaka: I like the pattern where the contents of this describe block would go into:

```
app/models/course/publish_enrichments_spec.rb
```

Naming it after the method being focused on makes it super easy to find. Thoughts? Is it too granular?

spec/requests/api/v2/providers/courses_spec.rb
misaka: And likewise, here I'd really like to see:

```
spec/requests/api/v2/providers/courses/publish_spec.rb
```

Even more so than for the model ... there's likely to be more tests here. :)

app/models/course.rb
```diff
@@ -148,4 +148,31 @@ def funding
...
+                last_published_timestamp_utc: Time.now.utc,
+                updated_by_user_id: current_user_id
+                )
+    update(changed_at: Time.now.utc)
+  end
+
+private
+
+  def update_new_to_running
+    site_statuses.where(status: 'new_status').update_all(status: 'running')
```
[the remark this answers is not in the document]
misaka: This should already exist as a scope because of the `status` enum on `site_status`.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -25,6 +25,14 @@ def sync_with_search_and_compare
         head response ? :ok : :internal_server_error
       end
 
+      def publish
+        if @provider.opted_in
+          @course.publish_sites
+        end
+        @course.publish_enrichment(@current_user.id)
+        sync_with_search_and_compare # todo: make a service?
```
misaka: Definitely should not be calling an action, and should probably be a service.

misaka: As agreed in person, for this PR how about we just duplicate what's in `sync_with_search_and_compare` here and we can (maybe) turn it into a service later.

app/models/course.rb
misaka: Use the scope created by the enum.

app/models/course.rb
```diff
@@ -148,4 +148,31 @@ def funding
...
+    update(changed_at: Time.now.utc)
+  end
+
+  def publish_enrichment(current_user_id)
+    enrichments.where(status: 'draft')
+    .update_all(status: 'published',
+                updated_at: Time.now.utc,
+                last_published_timestamp_utc: Time.now.utc,
+                updated_by_user_id: current_user_id
+                )
```
misaka: I don't think `update_all` is the right thing to use here, as it skips callbacks and validations (as you discovered with `changed_at`). I'm curious if there isn't a way to do this with a plain `update`, but it doesn't look like there is, so you'll probably have to iterate over each and run `update`.

Alternatively, as you point out there should only ever be one course enrichment in draft, you could just take the first one but I'd advise some way of guarding against there being multiples (raise an error if there is more than one course enrichment in draft).

app/models/course.rb
```diff
@@ -148,4 +148,31 @@ def funding
       'fee'
     end
   end
+
+  def publish_sites
+    update_new_to_running
+    update_published_if_running
```
misaka: Is it really worth having these as two separate updates? I say inline the updates here in as a single update.

app/models/course.rb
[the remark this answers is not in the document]
misaka: Also, same as before don't use `update_all`.

app/models/course.rb
```diff
@@ -148,4 +148,31 @@ def funding
...
+  end
+
+private
+
+  def update_new_to_running
+    site_statuses.where(status: 'new_status').update_all(status: 'running')
+  end
+
+  def update_published_if_running
+    site_statuses.where(status: 'running', publish: 'unpublished').update_all(publish: 'published')
```
misaka: Also, same as before don't use `update_all`.

misaka: You can use a method generated by the enum here too: `published_on_ucas!`

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+            changed_at: 10.minutes.ago,
+          with_enrichments: [
+          [:initial_draft, created_at: 1.day.ago, updated_at: 20.minutes.ago],
+        ])
+      end
+      let(:user) { create(:user) }
+
+      it 'should publish the draft' do
+        course.publish_enrichment(user.id)
+        expect(course.enrichments.first.status).to eq 'published'
```
misaka: Not saying this has to change, but you could also:

```
         expect(course.enrichments.first).to be_published
```

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+      let(:course) do
+        create(:course,
+            changed_at: 10.minutes.ago,
+          with_enrichments: [
+          [:initial_draft, created_at: 1.day.ago, updated_at: 20.minutes.ago],
+        ])
+      end
+      let(:user) { create(:user) }
+
+      it 'should publish the draft' do
```
misaka: How about nixing the `should`: `it 'publishes the draft'`. For here and below.

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+      end
+
+      it 'should update course changed_at to the current time' do
+        course.publish_enrichment(user.id)
+        expect(course.changed_at).to be_within(1.second).of Time.now.utc
+      end
+
+      it 'should touch enrichment updated_at to the current time' do
+        course.publish_enrichment(user.id)
+        expect(course.enrichments.first.updated_at).to be_within(1.second).of Time.now.utc
```
misaka: This can be tidied up like this:

```
      describe 'course enrichment' do
        before { course.publish_enrichment(user.id) } # This may be bubbled up?

        subject { course.enrichments.first }

        its(:updated_at) { should be_within(1.second).of Time.now.utc }
      end
```

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+        expect(course.enrichments.first.last_published_timestamp_utc).to be_within(1.second).of Time.now.utc
+      end
+
+      it 'should updated updated_by to the current user' do
+        course.publish_enrichment(user.id)
+        expect(course.enrichments.first.updated_by_user_id).to eq user.id
+      end
+    end
+
+    context 'on a course draft and published enrichments' do
```
misaka: Maybe: on a course with a draft enrichment and previously-published

Whatever it may be, I think a couple more words in here will help. :)

app/models/course.rb
```diff
@@ -148,4 +148,31 @@ def funding
...
   end
+
+  def publish_sites
+    update_new_to_running
+    update_published_if_running
+    # as update_all is called and doesn't trigger a save this has to be manually tested
+    update(changed_at: Time.now.utc)
+  end
+
+  def publish_enrichment(current_user_id)
```
misaka: I feel like, assuming you'll always have it, the user object should be here instead of the id.

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+        create(:course, with_enrichments: [
+            [:published, created_at: 5.days.ago],
+            [:published, created_at: 3.days.ago],
+            [:subsequent_draft, created_at: 1.day.ago],
+        ])
+      end
+      let(:user) { create(:user) }
+
+      it 'should publish the draft' do
+        course.publish_enrichment(user)
```
misaka: How is this test working when a user object is being passed, not the id?

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+        course.enrichments.each do |enrichment|
+          expect(enrichment.status).to eq 'published'
```
```
expect(enrichment).to be_published
```

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+      it 'updates course changed_at attribute to the current time' do
+        subject.publish_sites
+        expect(subject.changed_at).to be_within(1.second).of Time.now.utc
+      end
+    end
+
+    context 'course with a single site' do
+      let(:subject) { create(:course, with_site_statuses: [:new]) }
+
+      context 'site status is new site'do
```
misaka: 1. There isn't anything different in this context, does it add anything?
2. Spacing?

spec/models/course_spec.rb
```diff
@@ -434,4 +434,103 @@
...
+        [
+          { before: { publish: 'published', status: 'new_status' }, after: { publish: 'published', status: 'running' } },
+          { before: { publish: 'published', status: 'running' }, after: { publish: 'published', status: 'running' } },
+          { before: { publish: 'published', status: 'suspended' }, after: { publish: 'published', status: 'suspended' } },
+          { before: { publish: 'published', status: 'discontinued' }, after: { publish: 'published', status: 'discontinued' } },
+          { before: { publish: 'unpublished', status: 'new_status' }, after: { publish: 'published', status: 'running' } },
+          { before: { publish: 'unpublished', status: 'running' }, after: { publish: 'published', status: 'running' } },
+          { before: { publish: 'unpublished', status: 'suspended' }, after: { publish: 'unpublished', status: 'suspended' } },
+          { before: { publish: 'unpublished', status: 'discontinued' }, after: { publish: 'unpublished', status: 'discontinued' } },
+        ].each do |test_case|
```
misaka: I'm not sure about this. Even though I like the idea, I think that the errors generated here will be a little hard to parse and, at the end of the day, isn't really ["unsurprising"](http://principles-wiki.net/principles:principle_of_least_surprise).

This might be clearer with a [custom matcher](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/custom-matchers). Or, dare I say it, [shared examples](https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples).

Here's a quick stab at using shared examples:

```
let(:new_published_site_status) { create :site_status, :new, :published }
let(:running_published_site_status) { create :site_status, :running, :published }

it 'changes all the site statuses' do
  expect(create :site_status, :new, :published).to be_published_and_have_status('running')
  expect(create :site_status, :running, :published).to be_published_and_have_status('running')
end
```

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -151,6 +151,101 @@
     end
   end
 
+  describe 'POST publish' do
+    let(:api_status)   { 200 }
+    let(:api_response) { '{ "result": true }' }
+    let!(:stubbed_manage_courses_api_request) do
```
misaka: This is setup, not a memoised value.

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -151,6 +151,101 @@
...
+      it { should have_http_status(:forbidden) }
+    end
+
+    context 'when unauthorised' do
+      let(:unauthorised_user) { create(:user) }
+      let(:payload)           { { email: unauthorised_user.email } }
+
+      it "raises an error" do
+        expect { subject }.to raise_error Pundit::NotAuthorizedError
+      end
```
misaka: Can be a one-liner.

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -151,6 +151,101 @@
...
+    end
+
+    context 'when course and provider is not related' do
+      let(:course) { create(:course) }
+
+      it 'raises an error' do
+        expect { subject }.to raise_error ActiveRecord::RecordNotFound
+      end
```
misaka: Can be a one-liner.

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -151,6 +151,101 @@
...
+      let!(:course) {
+        create(:course,
+               provider: provider,
+               with_site_statuses: [:new],
+               with_enrichments: [:initial_draft],
+               age: 17.days.ago)
+      }
+      it 'publishes a course' do
+        # run the POST
+        response = subject
```
misaka: Ya, I agree that this is weird. I'm not sure how to suggest sorting it. :-/ Actually, I'm just a bit PRed out.

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -151,6 +151,101 @@
...
+    end
+
+    context 'when the api responds with result: false' do
+      let(:api_response) { '{ "result": false }' }
+
+      it { should have_http_status(:internal_server_error) }
+    end
+
+    context 'when the api sets http status to 500' do
+      let(:api_status)   { 500 }
```
misaka: Is this used by search_and_compare? Is it easy to make that clearer here?

app/models/course.rb
```diff
@@ -159,4 +159,21 @@ def funding
...
+    site_statuses.status_new_status.each(&:status_running!)
+    site_statuses.status_running.unpublished_on_ucas.each(&:published_on_ucas!)
+
+    # as update_all is called and doesn't trigger a save this has to be manually tested
+    update(changed_at: Time.now.utc)
+  end
+
+  def publish_enrichment(current_user)
+    enrichments.draft
+    .update(status: 'published',
```
heidar: Could this be moved to an instance method on `Enrichment` called `#published` or something similar?

app/models/course.rb
```diff
@@ -159,4 +159,21 @@ def funding
...
+            updated_at: Time.now.utc,
+            last_published_timestamp_utc: Time.now.utc,
+            updated_by_user_id: current_user.id)
+    update(changed_at: Time.now.utc)
```
heidar: Same here regarding `changed_at`, see comment above.

spec/models/course_spec.rb
```diff
@@ -434,4 +434,110 @@
       its(:content_status) { should eq(:published_with_unpublished_changes) }
     end
   end
+
+  describe "#publish_enrichments" do
+    context 'on a course with only a draft enrichment' do
+      let(:course) do
+        create(:course,
+            changed_at: 10.minutes.ago,
+          with_enrichments: [
```
heidar: The indentation doesn't look quite right here

spec/models/course_spec.rb
heidar: Could also throw the `reload` in there: `expect(course.reload.enrichments.first).to be_published` and get rid of the extra line.

spec/models/course_spec.rb
```diff
@@ -434,4 +434,110 @@
...
+      it 'should publish the draft' do
+        course.publish_enrichment(user)
+        course.reload
+        expect(course.enrichments.first.status).to eq 'published'
+      end
+
+      it 'should update course changed_at to the current time' do
+        course.publish_enrichment(user)
+        course.reload
+        expect(course.changed_at).to be_within(1.second).of Time.now.utc
```
heidar: Same here regarding `reload`. It can be done in the expectation since there's only one expectation here.

Also for the time, you can do:

```
expect(course.reload.changed_at.change(usec: 0)).to eq Time.now.utc.change(usec: 0)
```

spec/models/course_spec.rb
```diff
@@ -434,4 +434,110 @@
...
+      it 'should touch enrichment updated_at to the current time' do
+        course.publish_enrichment(user)
+        course.reload
+        expect(course.enrichments.first.updated_at).to be_within(1.second).of Time.now.utc
+      end
+
+      it 'should updated last_published to the current time' do
+        course.publish_enrichment(user)
+        course.reload
+        expect(course.enrichments.first.last_published_timestamp_utc).to be_within(1.second).of Time.now.utc
```
heidar: Same here, see my and @misaka's previous comments

spec/models/course_spec.rb
```diff
@@ -434,4 +434,110 @@
...
+      end
+
+      it 'should updated updated_by to the current user' do
+        course.publish_enrichment(user)
+        course.reload
+        expect(course.enrichments.first.updated_by_user_id).to eq user.id
```
heidar: Same here, see my and @misaka's previous comments

app/controllers/api/v2/courses_controller.rb
timabell: done

spec/models/course_spec.rb
```diff
@@ -434,4 +434,110 @@
...
+      end
+    end
+  end
+
+  describe 'publish_sites' do
+    context 'a pre-existing course' do
+      let(:subject) { create(:course, changed_at: 10.minutes.ago) }
+      it 'updates course changed_at attribute to the current time' do
+        subject.publish_sites
+        expect(subject.changed_at).to be_within(1.second).of Time.now.utc
```
heidar: Same applies here as before with the time comparisons, removing microseconds makes the times equal.

Feels like this might be better represented with an `expect { subject.publish_sites }.to change { ... }` maybe?

spec/models/course_spec.rb
```diff
@@ -434,4 +434,110 @@
...
+      end
+    end
+
+    context 'course with a single site' do
+      let(:subject) { create(:course, with_site_statuses: [:new]) }
+
+      context 'site status is new site'do
+        it 'sets sites to running and published' do
+          subject.publish_sites
```
heidar: Maybe move this to a `before` block?

app/models/course.rb
timabell: part 1 done
part 2 - not so sure about this at the moment as if there is an issue the current code would allow the user to continue unhindered, I'm starting to change my mind a bit on the fail early vs not breaking user journeys.

spec/requests/api/v2/providers/courses_spec.rb
heidar: Doesn't `response` get magically set when doing a request in request specs?

app/models/course.rb
```diff
@@ -159,4 +159,21 @@ def funding
       'fee'
     end
   end
+
+  def publish_sites
+    site_statuses.status_new_status.each(&:status_running!)
```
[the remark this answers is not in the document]
timabell: interestingly we had that to start with. I defer to @misaka

app/models/course.rb
```diff
@@ -159,4 +159,21 @@ def funding
...
+    site_statuses.status_running.unpublished_on_ucas.each(&:published_on_ucas!)
+
+    # as update_all is called and doesn't trigger a save this has to be manually tested
```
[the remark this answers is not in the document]
timabell: thanks, comment removed, it didn't quite make sense anyway

app/models/course.rb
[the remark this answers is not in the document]
heidar: If we need the `updated_at` timestamp to be updated, we can keep this as-is, otherwise I see no reason not to use `update_all`.

Since we don't have many site statuses per course, I don't think this will make a big difference though.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -26,8 +26,10 @@ def sync_with_search_and_compare
       end
 
       def publish
-        # todo: set running
-        # todo: set enrichment published
+        if @provider.opted_in
+          @course.publish_sites
```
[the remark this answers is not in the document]
timabell: brill, done

app/models/course.rb
```diff
@@ -159,4 +159,21 @@ def funding
...
+    update(changed_at: Time.now.utc)
+  end
+
+  def publish_enrichment(current_user)
+    enrichments.draft
+    .update(status: 'published',
+            updated_at: Time.now.utc,
```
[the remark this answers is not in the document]
timabell: good point. done, thanks

app/models/course.rb
timabell: done :+1:

app/models/course.rb
timabell: good idea. done

app/models/course.rb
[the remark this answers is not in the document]
timabell: there is an issue with `update_all` not calling the `TouchCourse` concern and not updating `updated_at`

spec/models/course_spec.rb
timabell: rearranged, thanks

spec/models/course_spec.rb
timabell: done, thanks

spec/models/course_spec.rb
timabell: reload inlining done, thanks

app/models/course.rb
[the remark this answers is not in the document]
heidar: Ah ok, good reason to stick with something that fires callbacks then. 👍

app/models/course_enrichment.rb
```diff
@@ -41,4 +43,10 @@ class CourseEnrichment < ApplicationRecord
   def has_been_published_before?
     last_published_timestamp_utc.present?
   end
+
+  def publish(current_user)
+    update(status: 'published',
```
defong: :published instead?

spec/models/course_enrichment_spec.rb
```diff
@@ -31,18 +31,50 @@
...
+
+    context 'when the enrichment is an initial draft' do
+      subject { create(:course_enrichment, :initial_draft, created_at: 1.day.ago, updated_at: 20.minutes.ago) }
+
+      before do
+        subject.publish(user)
+      end
+
+      its(:status) { should eq 'published' }
+      its(:updated_at) { should be_within(1.second).of Time.now.utc }
```
defong: timecop?

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -217,8 +223,30 @@
...
+    context 'unpublished course with draft enrichment' do
+      let!(:course) {
+        create(:course,
+               provider: provider,
+               with_site_statuses: [:new],
+               with_enrichments: [:initial_draft],
+               age: 17.days.ago)
+      }
+      it 'publishes a course' do
+        # run the POST
```
defong: leftovers?

app/models/course_enrichment.rb
timabell: you mean like this?
```
    update(status: :published,
          last_published_timestamp_utc: Time.now.utc,
          updated_by_user_id: current_user.id)
```
The tests still pass with the above but I'm not inclined to change it unless there's a compelling reason, we've spent a lot of time on this PR already. If that's not what you mean please post full example code so I can see what you are suggesting.

spec/models/course_enrichment_spec.rb
timabell: I think we had an issue with unique constraints on the db

spec/models/course_spec.rb
timabell: done, thanks

app/models/course_enrichment.rb
defong: yes that, symbol always seems the better practice, https://stackoverflow.com/questions/255078/whats-the-difference-between-a-string-and-a-symbol-in-ruby

spec/models/course_spec.rb
timabell: we're using user throughout now so we'll never know for sure what was going on

spec/models/course_spec.rb
davidgisbey: Done. Thanks for the feedback

spec/models/course_spec.rb
davidgisbey: Done. Thanks!

spec/requests/api/v2/providers/courses_spec.rb
davidgisbey: Tests now added.

spec/requests/api/v2/providers/courses_spec.rb
timabell: I'm not sure I follow

spec/models/course_spec.rb
timabell: to be_published done

spec/models/course_spec.rb
timabell: done, good spot

spec/models/course_spec.rb
timabell: We've brought it closer to this but we're dealing with both enrichments and course as a subject so I've picked course and moved on.

spec/models/course_spec.rb
timabell: rewritten, thanks for your help on this. Hope you like the replacement!

spec/models/course_spec.rb
heidar: I'm not a massive fan of this. I would find it confusing if the spec file matching the course model file doesn't have all the tests.

This is a side effect of course becoming a bit of a god object, we should look at decomposing it. One way to do that would be to move some code into a concern, then testing that in its own file.

I think this would be too granular and ultimately confusing.

spec/requests/api/v2/providers/courses_spec.rb
timabell: there's some oddness where this assignment causes the POST to actually be run, bit of a hack that needs revisiting. We'll have another look as it isn't right how it is. We'll tackle this after breaking the files up.

spec/requests/api/v2/providers/courses_spec.rb
davidgisbey: Done. Thanks for the feedback!

spec/models/course_spec.rb
timabell: Given this requires further discussion I suggest we merge this as-is then make improvements as future PRs.

spec/requests/api/v2/providers/courses_spec.rb
timabell: I think I need more guidance with this

spec/requests/api/v2/providers/courses_spec.rb
timabell: This makes the test fail, not sure how to make it work:

```
-      it "raises an error" do
-        expect { subject }.to raise_error Pundit::NotAuthorizedError
-      end
+      it { should raise_error Pundit::NotAuthorizedError }
```
failure:
```
$ bundle exec rspec ./spec/requests/api/v2/providers/courses/publish_spec.rb:58                                                                            {1}
Run options: include {:focus=>true, :locations=>{"./spec/requests/api/v2/providers/courses/publish_spec.rb"=>[58]}}
F

Failures:

  1) Publish API v2 POST publish when unauthorised 
     Failure/Error: authorize @course

     Pundit::NotAuthorizedError:
       not allowed to publish? this #<Course id: 1, age_range: nil, course_code: "C1D3", name: "AngelScript", profpost_flag: nil, program_type: "higher_education_programme", qualification: "pgce_with_qts", start_date: nil, study_mode: "full_time", accrediting_provider_id: nil, provider_id: 1, modular: nil, english: nil, maths: nil, science: nil, created_at: "2019-04-25 10:15:29", updated_at: "2019-04-25 10:15:29", changed_at: "2019-04-25 10:15:29">
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/pundit-2.0.1/lib/pundit.rb:221:in `authorize'
     # ./app/controllers/api/v2/courses_controller.rb:49:in `build_course'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/audited-4.8.0/lib/audited/sweeper.rb:14:in `around'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/sentry-raven-2.9.0/lib/raven/integrations/rails/controller_transaction.rb:7:in `block in included'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/audited-4.8.0/lib/audited/sweeper.rb:14:in `around'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-2.0.7/lib/rack/etag.rb:25:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-2.0.7/lib/rack/conditional_get.rb:38:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-2.0.7/lib/rack/head.rb:12:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/railties-5.2.3/lib/rails/rack/logger.rb:38:in `call_app'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/railties-5.2.3/lib/rails/rack/logger.rb:26:in `block in call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/railties-5.2.3/lib/rails/rack/logger.rb:26:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-2.0.7/lib/rack/runtime.rb:22:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-2.0.7/lib/rack/sendfile.rb:111:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/sentry-raven-2.9.0/lib/raven/integrations/rack.rb:51:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/railties-5.2.3/lib/rails/engine.rb:524:in `call'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-test-1.1.0/lib/rack/mock_session.rb:29:in `request'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-test-1.1.0/lib/rack/test.rb:266:in `process_request'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rack-test-1.1.0/lib/rack/test.rb:119:in `request'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/rails-controller-testing-1.0.4/lib/rails/controller/testing/integration.rb:13:in `block (2 levels) in <module:Integration>'
     # ./spec/requests/api/v2/providers/courses/publish_spec.rb:37:in `block (3 levels) in <top (required)>'
     # ./spec/requests/api/v2/providers/courses/publish_spec.rb:58:in `block (4 levels) in <top (required)>'
     # ./spec/rails_helper.rb:64:in `block (3 levels) in <top (required)>'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/database_cleaner-1.7.0/lib/database_cleaner/generic/base.rb:16:in `cleaning'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/database_cleaner-1.7.0/lib/database_cleaner/base.rb:100:in `cleaning'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/database_cleaner-1.7.0/lib/database_cleaner/configuration.rb:86:in `block (2 levels) in cleaning'
     # /home/tim/.rvm/gems/ruby-2.6.1/gems/database_cleaner-1.7.0/lib/database_cleaner/configuration.rb:87:in `cleaning'
     # ./spec/rails_helper.rb:63:in `block (2 levels) in <top (required)>'

Finished in 0.80168 seconds (files took 1.45 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/requests/api/v2/providers/courses/publish_spec.rb:58 # Publish API v2 POST publish when unauthorised 

Coverage report generated for RSpec to /home/tim/repo/dfe/manage/backend/coverage. 641 / 807 LOC (79.43%) covered.
```

spec/requests/api/v2/providers/courses_spec.rb
timabell: see above comment

spec/requests/api/v2/providers/courses_spec.rb
timabell: it's actually the manage-api status. done. thanks

app/models/course.rb
timabell: @misaka what do you want to do about the guard thing if anything?

timabell: @misaka @benilovj  I think in hindsight we modelled this wrong. Because draft and published are different things, they should (in a normalized database) be in different tables, then we could have applied proper database design and constraints to prevent bad data getting in. Worth bearing in mind as we look at reworking the structures later on. The draft could have been a field in the parent object, naturally enforcing the singular nature of them.

spec/models/course_spec.rb
timabell: thanks

* reloads have been tidied up
* I think `to be_within(1.second)` is more expressive

spec/models/course_spec.rb
timabell: thanks for the suggestions but I'm inclined to leave these for now

spec/models/course/publish_enrichments_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    let(:another_course) {
+      create(:course, with_enrichments: [
+        [:published, created_at: 5.days.ago],
+      ])
+    }
+
+    its(:size) { should eq(3) }
+
+    it "doesn't overlap with enrichments from another course" do
```
defong: overkill but we can live with that

spec/models/course/publish_enrichments_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+  end
+
+  describe "#content_status" do
+    context "for a course without any enrichments" do
+      subject { create(:course, with_enrichments: []) }
+      its(:content_status) { should eq(:empty) }
+    end
+
+    context "for a course an initial draft enrichments" do
```
defong:
```suggestion
    context "for a course an initial draft enrichment" do
```

singular

app/models/course_enrichment.rb
```diff
@@ -44,6 +46,12 @@ def has_been_published_before?
     last_published_timestamp_utc.present?
   end
 
+  def publish(current_user)
+    update(status: 'published',
```
defong:
```suggestion
    update(status: :published,
```

spec/requests/api/v2/providers/courses_spec.rb
davidgisbey: done 👍

spec/requests/api/v2/providers/courses_spec.rb
timabell: discussed in person, we could come back to this if we have more ideas

PR #287, opened 2019-04-18
[1318] Show languages used in modern language courses on the API

heidar:
This service class extracts languages, given a course name and a list of
UCAS subject names.

It only works on modern language courses.

### Context

We need to list the languages a course uses on the backend API.

Add a service to extract languages used in modern language courses and use it on the backend.

### Guidance to review

This is a draft.

### DIFF
--- a/app/services/list_languages_taught_in_modern_language_course.rb
+++ b/app/services/list_languages_taught_in_modern_language_course.rb
@@ -1,17 +1,15 @@
 class ListLanguagesTaughtInModernLanguageCourse
-  LANGUAGES = [
-    'english as a foreign language',
-    'french',
-    'german',
-    'italian',
-    'japanese',
-    'mandarin',
-    'russian',
-    'spanish',
-    'urdu'
+  LANGUAGE_SUBJECTS = [
+    'English as a foreign language',
+    'French',
+    'German',
+    'Italian',
+    'Japanese',
+    'Mandarin',
+    'Russian',
+    'Spanish',
+    'Urdu'
   ].freeze
-
-  MODERN_LANGUAGE_COURSE_PATTERN = /mfl|modern language|modern foreign language/.freeze
 
   attr_reader :course_name, :ucas_subject_names
 
@@ -25,20 +23,17 @@ class ListLanguagesTaughtInModernLanguageCourse
   end
 
   def call
-    return [] unless modern_language_course?
-
-    LANGUAGES.select do |language|
-      language.in? normalized_ucas_subject_names
+    dfe_subject_names.select do |subject_name|
+      subject_name.in? LANGUAGE_SUBJECTS
     end
   end
 
 private
 
-  def normalized_ucas_subject_names
-    @normalized_ucas_subject_names ||= ucas_subject_names.strip.downcase
-  end
-
-  def modern_language_course?
-    course_name.downcase.match?(MODERN_LANGUAGE_COURSE_PATTERN)
+  def dfe_subject_names
+    SubjectMapper.get_subject_list(
+      course_name,
+      ucas_subject_names
+    )
   end
 end
--- a/app/services/list_languages_taught_in_modern_language_course.rb
+++ b/app/services/list_languages_taught_in_modern_language_course.rb
@@ -1,0 +1,44 @@
+class ListLanguagesTaughtInModernLanguageCourse
+  LANGUAGES = [
+    'english as a foreign language',
+    'french',
+    'german',
+    'italian',
+    'japanese',
+    'mandarin',
+    'russian',
+    'spanish',
+    'urdu'
+  ].freeze
+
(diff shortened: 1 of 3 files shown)


heidar:
https://trello.com/c/U6DCVn4d

heidar:
I've only added the service class to figure out what the languages for the course are. Still need to hook it up to the course and set the attribute in the serialiser.

Just want to know if this is on the right path, I'm not sure about any of this.

app/services/list_languages_taught_in_modern_language_course.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+private
+
+  def normalized_ucas_subject_names
+    @normalized_ucas_subject_names ||= ucas_subject_names.strip.downcase
+  end
+
+  def modern_language_course?
+    course_name.downcase.match?(MODERN_LANGUAGE_COURSE_PATTERN)
```
fofr: Modern languages should be any course tagged to a language subject.

fofr: For example:
https://find-postgraduate-teacher-training.education.gov.uk/course/2DO/3DQB (which is tagged to Languages (European), French, Languages, Secondary)

spec/services/list_languages_taught_in_modern_language_course_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+          'French, German, Languages, Languages (European), Secondary'
+        end
+
+        subject { described_class.call(course_name, ucas_subject_names) }
+
+        it { should eq([]) }
+      end
+    end
+
+    context "when the course name contains MFL" do
```
defong: subject mapper dont really touch course name for language subject

spec/services/list_languages_taught_in_modern_language_course_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      let(:course_name) { 'MFL - Spanish' }
+
+      context 'with two language subjects' do
+        let(:ucas_subject_names) do
+          'French, German, Languages, Languages (European), Secondary'
```
defong: this should be an array

see
https://github.com/DFE-Digital/manage-courses-backend/blob/590a40c71e630d979d7938464a2a639e7e26b6f7/app/serializers/api/v2/serializable_course.rb#L20

app/services/list_languages_taught_in_modern_language_course.rb
heidar: So what you're saying is that it's not restricted to modern languages courses, but any "secondary" course?

I'm not sure how to differentiate the two, as both can have languages.

heidar: I'm guessing `age_range = 'S'` (secondary)?

spec/services/list_languages_taught_in_modern_language_course_spec.rb
heidar: I think this is what me and @fofr are discussing above.

spec/services/list_languages_taught_in_modern_language_course_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        it { should eq([]) }
+      end
+    end
+
+    context "when the course name contains MFL" do
+      let(:course_name) { 'MFL - Spanish' }
+
+      context 'with two language subjects' do
+        let(:ucas_subject_names) do
```
defong: any items that this touches 
https://github.com/DFE-Digital/manage-courses-backend/blob/590a40c71e630d979d7938464a2a639e7e26b6f7/app/services/subject_mapper.rb#L106

we set the get_subjects_list should return either `"Modern languages" or` "Modern languages (other)"

app/services/list_languages_taught_in_modern_language_course.rb
heidar: @defong has answered my question - thanks

spec/services/list_languages_taught_in_modern_language_course_spec.rb
```diff
@@ -0,0 +1,20 @@
+describe ListLanguagesTaughtInModernLanguageCourse do
+  describe '#call' do
+    context 'when the course is for Modern Languages (Spanish with German)' do
```
heidar: I'm unable to write test cases for other contexts as I'm not clear what courses are applicable and which aren't.

heidar:
I've done another iteration of this which uses the subject mapper service. I'm still unclear what courses should have languages listed and which shouldn't. Right now, if the subject mapper lists languages as subjects, they get listed. I don't think this is the intended behaviour though?

heidar:
Closing this but keeping the branch for whoever picks up this card.

PR #290, opened 2019-04-23
[1332] Add data migrate to docker setup

tvararu:
This will allow it to run in the pipeline.


tvararu:
[Track and expose which users have seen the transition screen](https://trello.com/c/KaeVi9jt/1251-track-and-expose-which-users-have-seen-the-transition-screen)

PR #291, opened 2019-04-23
[1345] Set enrichments as draft for new courses when opting in

heidar:
### Context

After transition, a bunch (50+) courses will have a UCAS status 'new' but content status of 'published'. This state inconsistency has arisen because pre-transition, these steps were carried out on separate systems.

Since the "publish" button only appears on courses that are in "draft" or "published with unpublished changes" states, these aforementioned courses won't have a publish button, meaning it's not obvious how to publish them.

When 'mcb provider optin' is run, it unpublishes the content enrichments for any course that's in the 'new' with 'published' enrichments state.

### Guidance to review

### DIFF
--- a/lib/mcb/commands/providers/optin.rb
+++ b/lib/mcb/commands/providers/optin.rb
@@ -9,6 +9,11 @@ run do |opts, args, _cmd|
       provider = Provider.find_by!(provider_code: provider_code)
       verbose "updating provider #{provider_code}"
       provider.update(opted_in: true)
+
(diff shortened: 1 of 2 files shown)


heidar:
https://trello.com/c/9oRBDrMp

heidar:
['mcb provider optin' should fix inconsistent course enrichments](https://trello.com/c/9oRBDrMp/1345-mcb-provider-optin-should-fix-inconsistent-course-enrichments)

PR #292, opened 2019-04-23
[1351] Only mark latest enrichment as draft

heidar:
### Context

When we opt in providers that have a course that is new in UCAS, we mark all of the enrichments as draft. We only want to mark the latest enrichment as draft, if it's published.

- Only latest enrichment gets marked as draft, not all of them
- Some text is output to the screen to indicate what happened

### Guidance to review

There's a failing test, will fix tomorrow. I have no idea why it fails. :(

### DIFF
--- a/app/models/course_enrichment.rb
+++ b/app/models/course_enrichment.rb
@@ -38,6 +38,8 @@ class CourseEnrichment < ApplicationRecord
              foreign_key: :ucas_course_code,
              primary_key: :course_code
 
+  scope :latest_first, -> { order(created_at: :desc) }
+
   def has_been_published_before?
     last_published_timestamp_utc.present?
   end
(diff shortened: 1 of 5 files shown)


benilovj:
@heidar: @misaka identified the the issue with the failing spec, it was related to matching on whitespace and the extra spec that you added making the factory-generated provider code go from being 3-character to 4-character. Should be fixed with 3399d53561d8a211d31a1981afa6de2b7beb1236 now.

heidar:
> @heidar: @misaka identified the issue with the failing spec, it was related to matching on whitespace and the extra spec that you added making the factory-generated provider code go from being 3-character to 4-character. Should be fixed with [3399d53](https://github.com/DFE-Digital/manage-courses-backend/commit/3399d53561d8a211d31a1981afa6de2b7beb1236) now.

Ahh, thanks! Seems I also used `puts` by accident, whoops.

PR #296, opened 2019-04-24
[1333] Added migration to change jsonb field name

defong:
### Context
Related to https://github.com/DFE-Digital/manage-courses-api/commit/ccb55b40e513ad0e51553c3882719339f77374f5#diff-d7d07a6731c9899247d0139cf93697f5

Change the actual stored jsonb data for accrediting provider to use `UcasProviderCode` instead of `UcasInstitutionCode`

### Guidance to review
Has no direct impact in this project as I far as I can see thru we have `provider_enrichment`

This change should solve 

- missing accrediting provider enrichment provided by the related provider on the manage-courses-api on the edit page for about your organisation
- missing accrediting body text for the manage-courses-ui for the course details preview page
- missing accrediting body text for the search-and-compare-ui for a course details page
 
#### Workflow
Courses are published via manage-courses-api to search-and-compare-api. (webjob, or user interaction, publishing organisation enrichment (publishes courses) or publishing a course)

Users looks at manage-courses-ui and search-and-compare-ui and the missing text is surfaced.

#### Justification
Migration needs to be done here, as manage-courses-backend pwnd the database.

It needs to be prefixed with `Ucas` as that is the expectation of manage-courses-api and manage-courses-ui.

The sql statement converts jsonb to text and replaces the field key name than converts it back to jsonb (cheap option) otherwise the whole jsonb needs to be deconstructed and amended then reconstructed as its a nested array field key name (head bang option).


defong:
[💣 Fix missing org 'about your accredited body' text 💣](https://trello.com/c/0PfhtzPM/1333-%F0%9F%92%A3-fix-missing-org-about-your-accredited-body-text-%F0%9F%92%A3)

db/migrate/20190424094003_fix_provider_enrichment_accrediting_provider_field_change.rb
```diff
@@ -0,0 +1,17 @@
+class FixProviderEnrichmentAccreditingProviderFieldChange < ActiveRecord::Migration[5.2]
+  def self.up
+    execute "UPDATE   provider_enrichment
+                SET   json_data = Replace(json_data :: text, '\"UcasInstitutionCode\":',
+                              '\"UcasProviderCode\":') ::
+                              jsonb
```
misaka: Turning the `jsonb` into `text` just to do a search-and-replace feels a bit risky. Is there no way to do this by using the built-in jsonb functions in PostgreSQL?

db/schema.rb
```diff
@@ -122,9 +122,6 @@
     t.index ["subject_id"], name: "IX_course_subject_subject_id"
   end
 
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
-  end
-
```
misaka: It might be worth running `rails data:migrate` on your development db just so that this doesn't get removed ... otherwise it's going to keep appearing and disappearing.

defong: do i need to run `rails db:migrate` as well as that was what changed `schema.rb`

db/migrate/20190424094003_fix_provider_enrichment_accrediting_provider_field_change.rb
defong: you asking for the head bang option...

db/schema.rb
misaka: Running `rails data:migrate` should just add this table to your db, and hopefully also dump the schema with it.

defong: dump the migration file to the `data` directory and run `rails data:migrate` I think that is right...

db/migrate/20190424094003_fix_provider_enrichment_accrediting_provider_field_change.rb
misaka: I just looked at the structure of the data that's being changed. 🙀 😭 What a nightmare. But yes, you are right, the structure of the JSON data is a bit too complex to be able to operate on using the json functions built-into Postgres.

defong: :upside_down_face:

PR #298, opened 2019-04-24
[1358] add current user id to Raven user context for Sentry

misaka:
### Context

We are getting sporadic Pundit::NotAuthorisedError. This change tries to up our visibility to try to understand the cause.

Add the id for the current user to the Raven user context.

### Guidance to review

Wasn't sure this change was worth testing, the rest of the tests should (hopefully) ensure we don't break anything.

### DIFF
--- a/app/controllers/api/v2/application_controller.rb
+++ b/app/controllers/api/v2/application_controller.rb
@@ -10,8 +10,14 @@ module API
           email = email_from_token(token)
 
           @current_user = User.find_by("lower(email) = ?", email.downcase)
-
-          @current_user.present?
+          if @current_user.present?
+            Raven.user_context(id: @current_user.id)
+            true
+          else
+            # Once DFE Sign-In ID is passed in, add that to the Raven user
+            # context.
+            false
+          end
         end
       end
 


misaka:
[Investigate pundit errors on transition page.](https://trello.com/c/iZxbET7v/1358-investigate-pundit-errors-on-transition-page)

app/controllers/api/v2/application_controller.rb
```diff
@@ -10,8 +10,14 @@ def authenticate
           email = email_from_token(token)
 
           @current_user = User.find_by("lower(email) = ?", email.downcase)
-
-          @current_user.present?
+          if @current_user.present?
+            Raven.user_context(id: @current_user.id)
+            true
+          else
+            # Once DFE Sign-In ID is passed in, add that to the Raven user
```
defong: current_user.sign_in_user_id 

thats the dfe sigh in id

misaka: Not available in the backend, we're still waiting for @heidar to add it so that it's passed through in the payload.

PR #300, opened 2019-04-24
[xxxx] Fixed Mr Lottery

defong:
### Context
Got strung by a Smithy
https://travis-ci.org/DFE-Digital/manage-courses-backend/builds/523905764?utm_source=github_status&utm_medium=notification

Make it user under test have unique user info

### DIFF
--- a/spec/requests/api/v2/users_spec.rb
+++ b/spec/requests/api/v2/users_spec.rb
@@ -85,9 +85,9 @@ describe '/api/v2/users', type: :request do
       end
       let(:user_params)           { params.dig :_jsonapi, :data, :attributes }
       let(:accept_terms_date_utc) { DateTime.now.utc }
-      let(:email)                 { 'dave.smith@gmail.com' }
-      let(:first_name)            { 'Dave' }
-      let(:last_name)             { 'Smith' }
+      let(:first_name)            { Faker::Name.unique.first_name }
+      let(:last_name)             { Faker::Name.unique.last_name }
+      let(:email)                 { Faker::Internet.email("#{first_name}.#{last_name}") }
       let(:json_data)             { JSON.parse(response.body)['data'] }
 
       before do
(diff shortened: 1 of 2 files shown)


spec/requests/api/v2/users_spec.rb
```diff
@@ -85,9 +85,9 @@ def perform_request
       end
       let(:user_params)           { params.dig :_jsonapi, :data, :attributes }
       let(:accept_terms_date_utc) { DateTime.now.utc }
-      let(:email)                 { 'dave.smith@gmail.com' }
-      let(:first_name)            { 'Dave' }
-      let(:last_name)             { 'Smith' }
+      let(:first_name)            { Faker::Name.first_name }
+      let(:last_name)             { Faker::Name.unique.last_name }
+      let(:email)                 { Faker::Internet.unique.email("#{first_name}.#{last_name}") }
```
misaka: You're missing a `unique` on the `first_name`. :) I still think you should just have this:

```
let(:user) { create :user, first_name: 'Bob', last_name: 'Kim', email: 'bob.kim@local' }
```

inside this block, to circumvent Faker for this test.

defong: Opps this simple `unique` thing is getting tenacity 

I settling for 

- user factory to using unique, due to the fact that building a list of user there maybe a clash
- for the test setup for setting a particular user details using unique ie first_name, last_name, email

vs hardcoding the user variable and/or the individual variables of first_name, last_name, email as that introduce confidence by lack of variation vs confidence by various

I care about mechanic of the action and that the original and intended is not the same.
I don't care explicitly what has changed

misaka: @defong The point is that this is a much smaller change that clearly communicates what the test is trying to do. Changing the names and emails to use Faker only communicates that they're meant to be different from "Dave Smith" as a side-effect.

defong: Apologies, point taken, was seething with Mr Lottery and Agent Smith and my jackpot.

PR #301, opened 2019-04-24
[1345] optin should unpublish to initial draft

benilovj:
### Context
There are two flavours of draft, it's appropriate for optin to set published
enrichments for new courses back to initial draft.

- encapsulate unpublishing on the `CourseEnrichment` model
- `optin` should unpublish back to initial draft

### DIFF
--- a/app/models/course_enrichment.rb
+++ b/app/models/course_enrichment.rb
@@ -43,4 +43,10 @@ class CourseEnrichment < ApplicationRecord
   def has_been_published_before?
     last_published_timestamp_utc.present?
   end
+
+  def unpublish(initial_draft: true)
+    data = { status: :draft }
+    data[:last_published_timestamp_utc] = nil if initial_draft
+    update(data)
+  end
 end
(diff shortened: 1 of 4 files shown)


app/models/course_enrichment.rb
```diff
@@ -43,4 +43,9 @@ class CourseEnrichment < ApplicationRecord
   def has_been_published_before?
     last_published_timestamp_utc.present?
   end
+
+  def unpublish(initial_draft: true)
+    update(status: :draft)
+    update(last_published_timestamp_utc: nil) if initial_draft
```
misaka: Is it worth it to hit the DB twice here?

spec/models/course_enrichment_spec.rb
```diff
@@ -57,4 +57,42 @@
...
+    let(:course) { create(:course, provider: provider) }
+    let(:last_published_timestamp_utc) { Date.new(2017, 1, 1) }
+    subject {
+      create(:course_enrichment, :published,
+        last_published_timestamp_utc: last_published_timestamp_utc,
+        course: course,
+        provider: provider)
+    }
+
+    context "to initial draft" do
```
misaka: `context` should describe a new setup/initial state (fixtures/memoized values), not a new target state.

app/models/course_enrichment.rb
benilovj: fixed

spec/models/course_enrichment_spec.rb
benilovj: fixed

PR #302, opened 2019-04-24
[1337] users who updated their emails in dfe signin cannot use publish

heidar:
### Context

We should be signing in users based on their DfE-SignIn ID first, then falling back to email if it doesn't exist.

- Moved authentication logic to a service object
- Added lookup by DfE-SignIn ID
- Added guarding against empty string emails (I don't know how sound our data is)

### Guidance to review

Frontend still needs a matching PR for this.

Sorry for refactoring in a feature PR. I had added the functionality before I realised. :(

### DIFF
--- a/app/controllers/api/v2/application_controller.rb
+++ b/app/controllers/api/v2/application_controller.rb
@@ -7,7 +7,7 @@ module API
 
       def authenticate
         authenticate_or_request_with_http_token do |token|
-          @current_user = Authentication.call(token)
+          @current_user = AuthenticationService.call(token)
           assign_sentry_contexts
           @current_user.present?
         end
--- a/app/controllers/api/v2/application_controller.rb
+++ b/app/controllers/api/v2/application_controller.rb
@@ -7,17 +7,9 @@ module API
 
       def authenticate
         authenticate_or_request_with_http_token do |token|
-          email = email_from_token(token)
-
-          @current_user = User.find_by("lower(email) = ?", email.downcase)
-          if @current_user.present?
-            Raven.user_context(id: @current_user.id)
-            true
-          else
-            # Once DFE Sign-In ID is passed in, add that to the Raven user
-            # context.
-            false
-          end
+          @current_user = Authentication.call(token)
+          assign_sentry_contexts
+          @current_user.present?
         end
       end
 
@@ -38,19 +30,9 @@ module API
         render json: error_body, status: :forbidden
       end
 
-      def email_from_token(token)
-        if Settings.authentication.algorithm == 'plain-text'
-          # This method can be used in development mode to simplify querying
-          # the API with curl. It should allow us to do:
-          #
-          #    curl -H 'Authorization: Bearer user@education.gov.uk' http://localhost:3001/api/v2/providers
-          token
-        else
-          (payload, _options) = JWT.decode(token,
-                                               Settings.authentication.secret,
-                                               Settings.authentication.algorithm)
-          payload['email']
-        end
+      def assign_sentry_contexts
+        Raven.user_context(id:              @current_user&.id)
+        Raven.tags_context(sign_in_user_id: @current_user&.sign_in_user_id)
       end
     end
   end
--- a/app/services/authentication.rb
+++ b/app/services/authentication.rb
@@ -1,0 +1,61 @@
+class Authentication
+  attr_accessor :encoded_token, :user
+
+  def self.call(encoded_token)
+    new(encoded_token).call
+  end
+
+  def initialize(encoded_token)
+    @encoded_token = encoded_token
+  end
+
+  def call
+    @user = user_by_sign_in_user_id || user_by_email
+    update_user_email if user_email_does_not_match_token?
+
+    user
+  end
+
+private
+
+  def decoded_token
+    @decoded_token ||= JWT.decode(
+      encoded_token,
+      Settings.authentication.secret,
+      Settings.authentication.algorithm
+    )
+    (decoded_token_payload, _algorithm) = @decoded_token
+
+    decoded_token_payload
+  end
+
+  def email_from_token
+    decoded_token['email']&.downcase
+  end
+
+  def sign_in_user_id_from_token
+    decoded_token['sign_in_user_id']
+  end
+
+  def user_by_email
+    return unless email_from_token.present?
+
+    User.find_by("lower(email) = ?", email_from_token)
+  end
+
+  def user_by_sign_in_user_id
+    return unless sign_in_user_id_from_token.present?
+
+    User.find_by(sign_in_user_id: sign_in_user_id_from_token)
+  end
+
+  def user_email_does_not_match_token?
+    return unless user
+
+    user.email&.downcase != email_from_token
+  end
+
+  def update_user_email
+    user.update(email: email_from_token)
+  end
+end
--- a/spec/controllers/api/v2/application_controller_spec.rb
+++ b/spec/controllers/api/v2/application_controller_spec.rb
@@ -2,107 +2,43 @@ require 'rails_helper'
 
 describe API::V2::ApplicationController, type: :controller do
   describe '#authenticate' do
-    let(:user) { create(:user) }
-    let(:payload) { { email: user.email } }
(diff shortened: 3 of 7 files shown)


heidar:
[Users who updated their emails in DfE Signin can't use Publish](https://trello.com/c/PT3mxHeE/1337-users-who-updated-their-emails-in-dfe-signin-cant-use-publish)

spec/services/authentication_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        {
+          email:           user.email,
+          sign_in_user_id: user.sign_in_user_id
+        }
+      end
+
+      it { should eq user }
+    end
+
+    context 'with a DfE-SignIn ID but no email' do
```
defong: I'm thinking there is one more combo

`with a DfE-SignIn ID but mismatch email`

mismatch email get updated in db

spec/services/authentication_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      let(:payload) { { sign_in_user_id: user.sign_in_user_id } }
+
+      it { should eq user }
+    end
+
+    context 'with an email but no DfE-SignIn ID' do
+      let(:payload) { { email: user.email } }
+
+      it { should eq user }
```
defong: I don't think this should ever be allowed, as the user must signed in via dfe sign in to progress

The original use case for the user email address was used as a `whitelist` but they must of signed in via dfe sign
If the user email is a mismatch from db but has dfe sign in id let them though and update their original email address on end.

spec/services/authentication_spec.rb
heidar: Correct, I'll add this.

heidar: @benilovj says that's a separate piece of work, shouldn't be done as a part of this.

spec/services/authentication_spec.rb
heidar: Gotcha, the correct case here is if the user has a DfE-SignIn ID but it doesn't match our records, we fall back to the email. But if there's no DfE-SignIn ID we shouldn't let them in.

When we fall back to the email, we should update the DfE-SignIn ID in the database however. I'll make these changes now.

spec/services/authentication_spec.rb
defong: oh i thought this was the mismatch work, also thinking about this dfe sign in always need to give email

heidar: DfE sign in always gives an email I think.

app/services/authentication.rb
```diff
@@ -0,0 +1,45 @@
+class Authentication
```
misaka: What about calling it `AuthenticationService`? `Authentication` feels too generic.

app/controllers/api/v2/application_controller.rb
```diff
@@ -38,19 +30,9 @@ def check_terms_accepted
...
-          token
-        else
-          (payload, _options) = JWT.decode(token,
-                                               Settings.authentication.secret,
-                                               Settings.authentication.algorithm)
-          payload['email']
-        end
+      def assign_sentry_contexts
+        Raven.user_context(id:              @current_user&.id)
+        Raven.tags_context(sign_in_user_id: @current_user&.sign_in_user_id)
```
misaka: Would it make sense to use `Raven.user_context` here, assuming it works? Then maybe it'd show up in the `user` section at the end of the tags:

image

app/services/authentication.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+private
+
+  def decoded_token
+    (decoded_token, _algorithm) = JWT.decode(
+      @encoded_token,
+      Settings.authentication.secret,
+      Settings.authentication.algorithm
+    )
```
misaka: I don't know how expensive JWT.decode is, but would it be worth memoizing `decoded_token` here? It just seems to me that this could potentially run a few times in the span of a request.

spec/controllers/api/v2/application_controller_spec.rb
```diff
@@ -2,107 +2,42 @@
...
-    it 'saves the user for use by the action' do
-      controller.authenticate
-
-      expect(assigns(:current_user)).to eq user
-    end
-
-    context 'algorithm is not plain-text' do
-      it { should be true }
-
+    context 'successful authentication' do
```
misaka: Sorry to keep banging on about this, but `context` should describe a setup or state, not an outcome. So here, I think "user is valid" or something similar would better the describe the state where we would expect "it saves the user for use by the action".

spec/controllers/api/v2/application_controller_spec.rb
```diff
@@ -2,107 +2,42 @@
...
     end
 
-    context 'algorithm is set to plain-text' do
-      let(:bearer_token) { "Bearer #{user.email}" }
-
-      before do
-        allow(Settings).to receive_message_chain(:authentication, :algorithm)
-                             .and_return('plain-text')
-      end
+    context 'unsuccessful authentication' do
```
misaka: Same comment about use of `context` here.

app/controllers/api/v2/application_controller.rb
heidar: It doesn't accept custom values as far as I can tell, whereas tags are a key/value store.

app/services/authentication.rb
heidar: We don't do that with any of our other service objects, so I opted for consistency.

app/services/authentication.rb
heidar: It's not possible to use restructuring and memoizing at the same time. So I opted for this, as it keeps the readability, which would otherwise suffer. What do you think? Got any better ideas?

misaka: How about doing it in `initialize`?

app/services/authentication.rb
misaka: I think this is a mistake, personally. I think there's way too much opportunity to clash. In Rails it's only the models that don't have a suffix ...

app/services/authentication.rb
heidar: I'm not sure how that solves the issue? Also shouldn't do "work" in initialisers.

app/services/authentication.rb
heidar: Yeah. I can rename this one and put another PR through to rename the others.

app/services/authentication.rb
misaka: That could be solved in `initialize` by setting `@decoded_token`, but I kind of agree with not doing work in the initialiser. So, alternatively:

```
   def decoded_token
     @decoded_token ||= JWT.decode(
      encoded_token,
      Settings.authentication.secret,
      Settings.authentication.algorithm
    ).first
  end
```

The destructuring is a nice to have, but ultimately disposable I think.

app/controllers/api/v2/application_controller.rb
misaka: 😿

app/services/authentication.rb
heidar: Yeah, I opted for the destructuring because it's more readable and it should be quite rare to have to decode this many times, and it seems like a relatively cheap operation so I can't see it being an issue. But I don't have strong feelings either way

spec/services/authentication_service_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      it 'does not authenticate the user based on a nil match' do
+        expect(subject).not_to eq user
+      end
+    end
+
+    context 'with an empty payload' do
+      let(:payload) {}
+
+      it 'raises an error' do
+        expect { subject }.to raise_error NoMethodError
```
heidar: Does this test actually provide any value? I pulled it over from the controller tests. @misaka

misaka: Ya, I agree, if we're not actually expecting this error anywhere it's probably useless. Maybe it's a vestigial test I left behind when I wanted to catch that *something* was thrown.

heidar:
This is actually backwards compatible, so merging this.

PR #307, opened 2019-04-25
[1337] Make development authentication consistent with production

heidar:
### Context

Previously we handled authentication without JWT decoding on the backend when running in development.

If you need to create a secret for some reason, then there's an mcb command for that.

Now it works the same as prod, with the secret set to `secret`.
Also removed the MCB command option for plain-text tokens.

### Guidance to review

Tested on Mac OS Mojave (10.14.4)

Frontend still needs updating for this to work.

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -1,3 +1,2 @@
 authentication:
-  algorithm: plain-text
-  secret: nil
+  secret: secret
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -82,19 +82,12 @@ module MCB
     require 'jwt'
 
     payload = { email: email }
-    if encoding == 'plain-text'
-      unless secret.nil?
-        raise 'Secret provided, only valid when encoding is NOT plain-text'
-      end
 
-      payload.to_json
-    else
-      if secret.nil?
-        raise 'Secret not provided, only valid when encoding is plain-text'
-      end
-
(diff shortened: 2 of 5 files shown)


dankmitchell:
I think we have some stuff in the reader about this and settings too... can we remove?

dankmitchell:
MCB is referring to "plain-text" in a few places. Can we clean that up too?

app/services/authentication_service.rb
```diff
@@ -19,6 +19,8 @@ def call
 private
 
   def decoded_token
+    return encoded_token if Rails.env.development?
```
benilovj: I think this needs to be

```suggestion
    return { 'email' => encoded_token } if Rails.env.development?
```

heidar:
[Users who updated their emails in DfE Signin can't use Publish](https://trello.com/c/PT3mxHeE/1337-users-who-updated-their-emails-in-dfe-signin-cant-use-publish)

heidar:
[Users who updated their emails in DfE Signin can't use Publish](https://trello.com/c/PT3mxHeE/1337-users-who-updated-their-emails-in-dfe-signin-cant-use-publish)

heidar:
[Users who updated their emails in DfE Signin can't use Publish](https://trello.com/c/PT3mxHeE/1337-users-who-updated-their-emails-in-dfe-signin-cant-use-publish)

PR #310, opened 2019-04-28
[1362] Add endpoint for support  requests to add additional users

davidgisbey:
### Context

Currently, the endpoint for the support app is currently on mc-api. This PR moves this functionality over to rails. The support app will need to be reworked to use the same authentication as the v2 api. This will follow in a separate card.

* Designs: https://manage-courses-prototype.herokuapp.com/request-access
* Existing implementation https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Api/Controllers/AdminController.cs#L28 (two actions)

- Approve and decline endpoint setup.
- Logic ported over from C#

### Guidance to review

Currently just a draft PR. As i worked on it this weekend solo i wanted to open this just to make sure i haven't done anything glaringly wrong.

Unit tests to follow tomorrow am.


app/models/access_request.rb
```diff
@@ -0,0 +1,51 @@
...
+  belongs_to :user, foreign_key: :requester_id, primary_key: :id
+
+  enum status: {
+    requested: 0,
+    approved: 1,
+    completed: 2,
+    declined: 3,
+  }
+
+  def admin?(current_user)
```
benilovj: This belongs on the `User` model, I think.

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -0,0 +1,34 @@
+module API
+  module V2
+    class AccessRequestsController < API::V2::ApplicationController
+      before_action :build_access_request
+      before_action :build_requesting_user
+
+      def approve
+        if @access_request.admin?(@current_user) && requesting_user_belongs_to_oranisation?
```
benilovj: This logic should be pushed into the policy.

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+          @access_request.completed!
+          @access_request.update_access(@access_request, @requesting_user)
```
benilovj: I think it'd be a little safer to `update_access` and then call `completed!`, in the unlikely case that `update_access` bombs out.

timabell:
[Add endpoint to approve user access.](https://trello.com/c/AaW5wsnh/1362-add-endpoint-to-approve-user-access)

app/policies/access_request_policy.rb
```diff
@@ -0,0 +1,14 @@
+class AccessRequestPolicy
+  attr_reader :user, :access_request
+
+  def initialize(user, access_request)
+    @user = user
+    @access_request = access_request
+  end
+
+  def approve?
+    @user.organisations.map(&:name).include?(@access_request.organisation)
```
benilovj: I believe that this should be `@user.admin?`

app/models/user.rb
```diff
@@ -20,6 +20,8 @@ class User < ApplicationRecord
 
   has_and_belongs_to_many :organisations
   has_many :providers, through: :organisations
+  has_many :access_requests, foreign_key: :requester_id, primary_key: :id
+  has_many :providers, through: :organisations
```
dankmitchell: Duplicate

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+        end
+      end
+
+      def decline
+        @access_request.declined! if @access_request.admin?(@current_user)
```
heidar: This postcondition should also be moved into a policy.

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+    private
+
+      def build_access_request
+        @access_request = AccessRequest.find_by(id: params[:id])
```
heidar: Not a big deal or anything but this can be one line
`@access_request = authorise AccessRequest.find_by(id: params[:id])`

app/models/access_request.rb
```diff
@@ -0,0 +1,51 @@
...
+#  requester_id     :integer
+#  status           :integer          not null
+#  requester_email  :text
+#
+
+class AccessRequest < ApplicationRecord
+  belongs_to :user, foreign_key: :requester_id, primary_key: :id
+
+  enum status: {
+    requested: 0,
```
heidar: You don't need to map to the integers, it happens by default. I think the recommendation is to not do it unless you need to.

app/models/access_request.rb
```diff
@@ -0,0 +1,51 @@
...
+    declined: 3,
+  }
+
+  def admin?(current_user)
+    # should probably be using regex here?
+    admin_users = User.where("email ILIKE (?)", "%education.gov.uk")
+    admin_users.include?(current_user)
+  end
+
+  def update_access(access_request, requesting_user)
```
heidar: Can this be named differently to give a better idea what is happening here?

app/models/access_request.rb
```diff
@@ -0,0 +1,51 @@
...
+    requested_user = User.find_by(email: access_request.email_address)
+    if requested_user == nil
+      access_request.create_requested_user(access_request, requesting_user, requested_user)
```
heidar: How come we pass in `nil` here? Would it be better to be explicit about passing in `nil` to avoid confusion?

app/models/access_request.rb
heidar: Also, `requested_user` and `requesting_user` are quite hard to differentiate, are there any better names we can use?

app/models/access_request.rb
```diff
@@ -0,0 +1,51 @@
...
+    else
+      requested_user.update(organisations: requesting_user.organisations)
```
heidar: What is the reason for doing this?

app/models/user.rb
davidgisbey: Done & thanks!

app/models/access_request.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+  end
+
+  def create_requested_user(access_request, requesting_user, _requested_user)
```
heidar: This might be nice to keep in a service object.

Regarding the arguments, how come we have an unused argument?

app/models/user.rb
```diff
@@ -38,3 +39,9 @@ def opted_in?
     providers.any?(&:opted_in?)
   end
 end
+# organisation     :text
```
heidar: Is this meant to be here? I thought this was always at the top of files?

config/routes.rb
```diff
@@ -74,6 +74,11 @@
 
       resource :sessions
       resources :site_statuses, only: :update
+
+      resources :access_requests, only: :update do
+        post :approve, on: :member
```
heidar: Could DRY  the `on: :member` by doing
```
member do
  post :approve
  post :decline
end
```
Not important!

spec/factories/access_requests.rb
```diff
@@ -0,0 +1,30 @@
...
+#  status           :integer          not null
+#  requester_email  :text
+#
+
+FactoryBot.define do
+  factory :access_request do
+    user
+    status { :requested }
+    request_date_utc { Time.now.utc }
+    email_address { 'testuser.org.uk' }
```
heidar: Shouldn't this be an email?

app/models/access_request.rb
misaka: This method signature doesn't quite feel right to me, actually.

1. `access_request` and `self` are the same thing, so there is no need to pass it in
2. `requester` could and probably should be a `belongs_to` relationship
3. given the above 2 points, this method doesn't need any arguments (and a better name would probably also help)

app/models/access_request.rb
misaka: If we are requesting access for a user, and that user already exists, we should be copying the requester's access to this user.

BUT, @davidgisbey, I think this is wrong here. This is replacing the requested user's organisations when I think that the requester's organisations should be *added*.

app/models/access_request.rb
```diff
@@ -0,0 +1,47 @@
...
+    requested
+    approved
+    completed
+    declined
+  ].freeze
+
+  def update_access(access_request, requesting_user)
+    requested_user = User.find_by(email: access_request.email_address)
+    if requested_user == nil
+      access_request.create_requested_user(access_request, requesting_user)
```
misaka: `access_request` is self, so again, no need to specify.

app/models/access_request.rb
```diff
@@ -0,0 +1,47 @@
...
+      requested_user.update(organisations: requesting_user.organisations)
+    end
+  end
+
+  def create_requested_user(access_request, requesting_user)
+    User.create(
+      email: access_request.email_address,
+      first_name: access_request.first_name,
+      last_name: access_request.last_name,
+      invite_date_utc: Time.now.utc,
```
misaka: Is this right? Should `invite_date_utc` not be the same as `request_date_utc`?

benilovj: For context, that kinda doesn't matter too much – the `invite_date_utc` is not used anywhere and should actually be dropped middle-to-long-term.

app/models/user.rb
```diff
@@ -37,4 +38,11 @@ class User < ApplicationRecord
   def opted_in?
     providers.any?(&:opted_in?)
   end
+
+  def admin?
+    return true if email.end_with? "@education.gov.uk"
```
benilovj: You could condense this method to:

```suggestion
    email.end_with? "@education.gov.uk" || email.end_with? "@digital.education.gov.uk"
```

app/models/access_request.rb
```diff
@@ -0,0 +1,28 @@
...
+#  reason           :text
+#  request_date_utc :datetime         not null
+#  requester_id     :integer
+#  status           :integer          not null
+#  requester_email  :text
+#
+
+class AccessRequest < ApplicationRecord
+  belongs_to :user, foreign_key: :requester_id, primary_key: :id
+
```
benilovj: we probably want to add auditing to this class.

app/models/user.rb
dankmitchell: or `return true if email.include?("education.gov.uk")`

defong: maybe not include

education.gov.uk@somewhere.com

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -0,0 +1,19 @@
+module API
+  module V2
+    class AccessRequestsController < API::V2::ApplicationController
+      before_action :build_access_request
+
+      def approve
+        result = AccessRequestApprovalService.call(@access_request)
+
+        render status: 200, json: { result: result }
```
defong: `render jsonapi: result` maybe

spec/factories/users.rb
```diff
@@ -26,5 +26,9 @@
     trait :opted_in do
       organisations { [create(:organisation, providers: [create(:provider, opted_in: true)])] }
     end
+
+    trait :admin do
+      email { "admin@education.gov.uk" }
```
defong:
```suggestion
      email { "admin@" + ["digital.education.gov.uk",  "education.gov.uk"].sample }
```

spec/factories/access_requests.rb
```diff
@@ -0,0 +1,29 @@
...
+FactoryBot.define do
+  factory :access_request do
+    requester { build(:user, :opted_in) }
+    status { :requested }
+    request_date_utc { Time.now.utc }
+    email_address { Faker::Internet.email }
+    first_name { requester.first_name }
+    last_name { requester.last_name }
+    organisation { create(:organisation) }
+    reason { 'things' }
```
defong:
```suggestion
    reason { Faker::Lorem.sentence.to_s}
```

spec/models/user_spec.rb
```diff
@@ -58,4 +58,24 @@
       its(:opted_in?) { should be_falsey }
     end
   end
+
+  describe '#admin?' do
+    context 'user has an education.gov.uk email' do
+      subject { create(:user, email: 'test@education.gov.uk') }
```
defong:
```suggestion
      subject { create(:user, :admin }
```

heidar: I think we're testing the implementation here, so we should hard code it. The trait is useful for other areas where we need an admin.

app/controllers/api/v2/access_requests_controller.rb
heidar: Going to follow this: https://discuss.jsonapi.org/t/how-to-you-send-a-success-message/280

defong: cool, 

fyi 
> if the request does not exist it returns not found

https://trello.com/c/AaW5wsnh/1362-add-endpoint-to-approve-user-access

heidar: Rails should do that for us

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+
+    private
+
+      def build_access_request
+        @access_request = authorize AccessRequest.find_by(id: params[:id])
```
defong: do we need to query for not approved?

spec/factories/access_requests.rb
```diff
@@ -0,0 +1,29 @@
...
+#  requester_email  :text
+#
+
+FactoryBot.define do
+  factory :access_request do
+    requester { build(:user, :opted_in) }
+    status { :requested }
+    request_date_utc { Time.now.utc }
+    email_address { Faker::Internet.email }
+    first_name { requester.first_name }
```
defong:
```suggestion
    first_name { Faker::Name.first_name }
```

I think its should be the new user not the person requesting

heidar:
There are still some things that can be improved, like the request spec but I think we should get this out the door as it's needed and it's getting big. It's been open for a while now.

I'm not 100% sure what the endpoint should be returning, I think we can always address that when the frontend is further on and the requirements are clearer. Right now it follows the same pattern as the other apps do with `result :true`.

We should probably do this in a JSONAPI way though, like we do with errors.

spec/factories/access_requests.rb
```diff
@@ -0,0 +1,29 @@
...
+    last_name { requester.last_name }
```
defong:
```suggestion
    last_name { Faker::Name.last_name }
```

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+                  requester_id: requesting_user.id,
+                  organisation: organisation.name)
+        }
+
+        before do
+          post "/api/v2/access_requests/#{new_user_access_request.id}/approve",
+          headers: { 'HTTP_AUTHORIZATION' => credentials }
+        end
+
+        it 'creates a new account for a new user and gives access to the right orgs' do
```
defong: some test for 

```
                  first_name: 'test',
                  last_name: 'user',
```

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      context 'when the user requested user already exists' do
+        it 'gives a pre existing user access to the right organisations' do
+          expect(requested_user.organisations).to eq requesting_user.organisations
+        end
+      end
+
+      context 'when email address does not belong to a user' do
+        let(:new_user_access_request) {
+          create(:access_request,
```
defong: maybe use the default factory? and override the required bits?

davidgisbey:
🚢

spec/requests/api/v2/access_request_spec.rb
heidar: I'll add a Trello card to address this.

spec/requests/api/v2/access_request_spec.rb
heidar: I'll add a Trello card to address this.

PR #312, opened 2019-04-29
[1354] Remove sensitive information from Sentry

heidar:
### Context

The parameters we filter out for sensitive information in logs, etc. isn't respected by Sentry.

Make Sentry filter out the same attributes as Rails does.

### Guidance to review

Need to test if this actually works.


heidar:
[Sentry/Slack/Azure/Logging Not to contain PPI ie email address](https://trello.com/c/hVd4NwKD/1354-sentry-slack-azure-logging-not-to-contain-ppi-ie-email-address)

PR #316, opened 2019-04-29
[1400] Add guard

tvararu:
### Context

This allows us to automatically run rspec on files that have changed.

### Guidance to review

This runs pretty slowly, about 3 seconds for a model spec with 2 assertions. Suggestions welcome on how to improve this.

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -1,0 +1,68 @@
+# A sample Guardfile
+# More info at https://github.com/guard/guard#readme
+
+## Uncomment and set this to only include directories you want to watch
+# directories %w(app lib config test spec features) \
+#  .select{|d| Dir.exist?(d) ? d : UI.warning("Directory #{d} does not exist")}
+
+## Note: if you are using the `directories` clause above and you are not
+## watching the project directory ('.'), then you will want to move
+## the Guardfile to a watched dir and symlink it back, e.g.
+#
+#  $ mkdir config
+#  $ mv Guardfile config/
+#  $ ln -s config/Guardfile .
+#
+# and, you'll have to watch "config/Guardfile" instead of "Guardfile"
+
+# Note: The cmd option is now required due to the increasing number of ways
+#       rspec may be run, below are examples of the most common uses.
+#  * bundler: 'bundle exec rspec'
+#  * bundler binstubs: 'bin/rspec'
+#  * spring: 'bin/rspec' (This will use spring if running and you have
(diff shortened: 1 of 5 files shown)


tvararu:
[Add guard to manage-courses-backend](https://trello.com/c/6uslCKWD/1400-add-guard-to-manage-courses-backend)

heidar:
my hero

misaka:
Just wanted to add rubocop to it.

misaka:
I haven't managed to get `pry` to run nicely in guard, unfortunately. Anyone have any ideas? I'm currently trying to compile Ruby with readline.

Guardfile
```diff
@@ -0,0 +1,74 @@
...
+    # Rails files
+    rails = dsl.rails(view_extensions: %w(erb))
+    dsl.watch_spec_files_for(rails.app_files)
+    dsl.watch_spec_files_for(rails.views)
+
+    watch(rails.controllers) do |m|
+      [
+        rspec.spec.call("routing/#{m[1]}_routing"),
+        rspec.spec.call("controllers/#{m[1]}_controller"),
+        rspec.spec.call("acceptance/#{m[1]}")
```
heidar: We probably want to change this to run `spec/requests/` but we don't map these files clearly. We could change that so we can run the relevant request specs every time we modify controllers?

Guardfile
```diff
@@ -0,0 +1,74 @@
...
+      ]
+    end
+
+    # Rails config changes
+    watch(rails.spec_helper)     { rspec.spec_dir }
+    watch(rails.routes)          { "#{rspec.spec_dir}/routing" }
+    watch(rails.app_controller)  { "#{rspec.spec_dir}/controllers" }
+
+    # Capybara features specs
```
heidar: I don't think we have any of these

Guardfile
misaka: Noticed this too. I'm wondering if we need to sort this now though, or can we slowly improve this as we go?

misaka: Also, seeing as this is just for us devs, could alternatively just make the changes you feel need to be done to add to this PR. ;)

Guardfile
tvararu: done

tvararu:
Have made this significantly faster by using spring.

misaka:
+1 for `--no-interactions`!

PR #319, opened 2019-04-30
[1346] add mcb config system

misaka:
https://trello.com/c/11v4CXhF/1346-mcb-provider-optin-should-audit-which-user-ran-the-command

### Context

When admins make changes via `mcb` we want to ensure there is an audit trail for who did the changes. This also requires some kind of a config system to keep track of the admin's email address, and which will also be useful for future features.

This PR introduces a config mechanism that will be used in a near-future PR to add the audit stuff. It also does a little bit of specs tidying up that fell out of the work of adding the config stuff.

### Guidance to review

Config mechanism has it's own PR since it's grown a bit big. Another PR for auditing system to follow later.

### DIFF
--- a/bin/mcb
+++ b/bin/mcb
@@ -18,6 +18,7 @@ lib_dir = File.expand_path 'lib'
 $LOAD_PATH << lib_dir
 require 'mcb'
 require 'mcb/azure'
+require 'mcb/config'
 
 tp.set :capitalize_headers, false
 
@@ -35,6 +36,11 @@ $mcb = Cri::Command.define do
   flag :h, :help, 'show help for this command' do |_value, cmd|
     puts cmd.help
     exit 0
+  end
+
+  option :c, :config, 'use provided config file',
+         argument: :required do |config_file|
+    MCB.config_file = config_file
   end
 end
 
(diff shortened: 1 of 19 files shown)


misaka:
['mcb provider optin' should audit which user ran the command](https://trello.com/c/11v4CXhF/1346-mcb-provider-optin-should-audit-which-user-ran-the-command)

misaka:
Whoops. Included too much in this PR. Re-cutting.

bin/mcb
```diff
@@ -36,6 +37,11 @@ $mcb = Cri::Command.define do
     puts cmd.help
     exit 0
   end
+
+  option :c, :config, 'use provided config file',
```
defong: can you throw in an example config file

misaka: Not really meant to be used this way, as in ... we shouldn't have to add the config stuff by hand. mcb commands will use this to save stuff, and if the user has to do something special they can prompt the user to do so (e.x. the new audit command will [ask the user to add their email](https://github.com/DFE-Digital/manage-courses-backend/pull/324/files#diff-fabca55fa119b555a7942669c431eafeR212))

PR #321, opened 2019-04-30
[XXX] mcb add new course [early feedback needed]

benilovj:
Raising to get early feedback about the approach, I'll re-do this PR when we've agreed on the way forward.


lib/mcb/commands/courses/create.rb
```diff
@@ -0,0 +1,184 @@
...
+        start_date: wizard.start_date,
+        study_mode: wizard.study_mode,
+        accrediting_provider: wizard.accredited_body,
+        modular: nil,
+        english: 3,
+        maths: 3,
+        science: nil,
+        subjects: wizard.ucas_subjects,
+        sites: wizard.sites,
+      )
```
misaka: As a general pattern, would it maybe make sense to construct a `Course` object inside of `NewCourseWizard` so that instead of saving values in `@name`, etc, it would be stored directly on, say, `@course`? You could immediately get the benefits of validations and other calculated values, for example, and wouldn't need this extra mapping that has to happen here, and if implemented a certain might even contribute to edit functionality (e.g. a system allows users to pick what they want to edit, and automatically reports on anything not-yet-valid as a guide of what's missing).

lib/mcb/commands/courses/create.rb
```diff
@@ -0,0 +1,184 @@
...
+      "Modular" => @course.modular,
+      "English" => @course.english,
+      "Maths" => @course.maths,
+      "Science" => @course.science,
+      "Subjects" => @course.dfe_subjects.join(", "),
+      "Level" => @course.dfe_level,
+      "Training locations" => @course.sites.map(&:location_name).join(", "),
+    }
+  end
+end
```
misaka: Feels like this is an opportunity to establish a pattern of reuse for the `show` commands we've already been adding.

misaka:
👍 💯 I think this looks great.

lib/mcb/commands/courses/create.rb
benilovj: what did you have in mind?

misaka: Just to reuse `CoursePreview` here and in `mcb courses show`, so maybe it needs to be `CourseShow` in `lib/mcb/course_show.rb` or something like that.

lib/mcb/commands/courses/create.rb
benilovj: Agreed – I've introduced a new `NewCourseWizard` model class which encapsulates all the course creation logic.

lib/mcb/commands/courses/create.rb
benilovj: Sure – I extracted the `CourseShow` as part of #326.

timabell:
[Missing course and location editing through mcb](https://trello.com/c/QGQiBD7b/1407-missing-course-and-location-editing-through-mcb)

PR #323, opened 2019-04-30
[1407] Course model improvements

benilovj:
### Context
The `Course` model will be called from `mcb` in order to build and create new courses.

- pull subject-related logic (`dfe_subjects`, `level` and `is_send?`) into model code out of the serializer, so it can be reused
- add a validation for course codes to reduce the risk of duplicates being added

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -160,6 +160,18 @@ class Course < ApplicationRecord
     end
   end
 
+  def dfe_subjects
+    SubjectMapperService.get_subject_list(name, subjects.map(&:subject_name))
+  end
+
+  def level
+    SubjectMapperService.get_subject_level(subjects.map(&:subject_name))
+  end
+
+  def is_send?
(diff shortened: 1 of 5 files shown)


app/models/course.rb
```diff
@@ -28,6 +28,7 @@ class Course < ApplicationRecord
 
   has_associated_audits
   audited except: :changed_at
+  validates :course_code, uniqueness: { scope: :provider_id }
```
misaka: Looking at `schema.rb` I don't think `course_code` is indexed. This is going to be a bit of a performance hit, I think we should add it.

We might also consider adding a unique index on the combination of `course_code` and `provider_id`.

misaka: My mistake, the index is there I just missed it.

defong:
[New course creation through mcb](https://trello.com/c/QGQiBD7b/1407-new-course-creation-through-mcb)

PR #327, opened 2019-05-01
[1407] Add enums for GCSE entry requirement attributes

benilovj:
### Context

See design write-ups here for the domain context:

- https://bat-design-history.herokuapp.com/publish-teacher-training/minimum-course-requirements-logic
- https://bat-design-history.herokuapp.com/publish-teacher-training/specific-requirements

This makes it easier to reason about them and reduces the likelihood of mistakes.

- add enums to `maths`, `english` and `science` fields on `Course`.
- update V1 course serializer (to maintain the same output) and the request spec


benilovj:
> Wherever we have used enums, we've added unit tests for them. I don't personally mind, but shall we do the same here for consistency?

Added

PR #328, opened 2019-05-01
[1362] Access requests shouldn't duplicate orgs for a user

benilovj:
### Context
Only add new orgs to the target user. This comes up in production because publishers lodge duplicate access requests that get approved.

- fix unit test that was obscuring the issue
- fix issue

### DIFF
--- a/app/services/access_request_approval_service.rb
+++ b/app/services/access_request_approval_service.rb
@@ -14,7 +14,8 @@ class AccessRequestApprovalService
       user.invite_date_utc = Time.now.utc
     end
 
-    target_user.organisations << @access_request.requester.organisations
+    orgs_missing_on_target_user = @access_request.requester.organisations - target_user.organisations
+    target_user.organisations << orgs_missing_on_target_user
     @access_request.approve
   end
 end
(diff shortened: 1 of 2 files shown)


heidar:
> fix unit test that was obscuring the issue

Did you do this?

Fix looks good. Shall we add some DB constraints to prevent this from happening?

spec/services/access_request_approval_service_spec.rb
```diff
@@ -81,7 +81,7 @@
         let!(:target_user) do
           create(
             :user,
-            :opted_in,
+            email: access_request.email_address,
```
defong: Nice catch


Can change this test as well, as it should of caught it out, 


expect(target_user.organisations).to(
             match_array(target_user.organisations.uniq)
          )

heidar: Ugh, I messed this up when refactoring the test. This is what happens when rushing. :(

heidar: This test does that @defong

app/services/access_request_approval_service.rb
```diff
@@ -14,7 +14,8 @@ def call
       user.invite_date_utc = Time.now.utc
     end
 
-    target_user.organisations << @access_request.requester.organisations
```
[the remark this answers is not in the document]
heidar: I think we prefer what Jake has done, I believe they do the same thing in the end. His version is nice and easy to understand.

benilovj:
> Can forgo array union but not the test, removed one obscurity but there is another one.

image

PR #329, opened 2019-05-01
[1417] Remove opted_in logic

dankmitchell:
### Context
All providers are now `opted_in`

Remove `opted_in`

# Guidance
Frontend needs to be deployed first - https://github.com/DFE-Digital/manage-courses-frontend/pull/244


benilovj:
I think we need to remove opt-in logic from the frontend before we merge this change, otherwise the frontend will break.

dankmitchell:
Description updated @benilovj ;-)

defong:
[Remove opted_in logic from mcb](https://trello.com/c/6GTCfroB/1417-remove-optedin-logic-from-mcb)

PR #332, opened 2019-05-03
[1431] Handle not found errors

heidar:
### Context

Backend API doesn't handle 404s gracefully.

Rescue from `ActiveRecord::RecordNotFound` and show an empty JSONAPI response with an HTTP status of 404.

### DIFF
--- a/app/controllers/api/v2/application_controller.rb
+++ b/app/controllers/api/v2/application_controller.rb
@@ -3,6 +3,8 @@ module API
     class ApplicationController < ::ApplicationController
       attr_reader :current_user
 
+      rescue_from ActiveRecord::RecordNotFound, with: :jsonapi_404
+
       before_action :check_terms_accepted
 
       def authenticate
@@ -11,6 +13,10 @@ module API
           assign_sentry_contexts
           @current_user.present?
         end
+      end
+
+      def jsonapi_404
+        render jsonapi: nil, status: 404
       end
 
     private
--- a/spec/requests/api/v2/application_spec.rb
+++ b/spec/requests/api/v2/application_spec.rb
@@ -1,0 +1,32 @@
+require "rails_helper"
+
+describe 'API v2', type: :request do
+  describe 'Error 404' do
+    let(:user) { create(:user, organisations: [organisation]) }
+    let(:organisation) { create(:organisation) }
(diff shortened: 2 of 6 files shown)


heidar:
[Handle not found errors with 404 responses](https://trello.com/c/CZFmryNP/1431-handle-not-found-errors-with-404-responses)

app/controllers/api/v2/application_controller.rb
```diff
@@ -13,6 +15,10 @@ def authenticate
         end
       end
 
+      def jsonapi_404
+        render jsonapi: nil, status: 404
+      end
```
misaka: I'm not a fan of this.

This works so long as what's failing is finding the resource specified in the URL, but as soon as a piece of logic further along fails to find something, we'll be at risk of getting 404 errors with nothing being logged to Sentry and no way to debug this without replicating this locally.

I think that instead of doing this, our `build_*` methods in the controllers should catch a `RecordNotFound` error and return the error. e.g.:

```
def build_access_request
  @access_request = authorize AccessRequest.requested.find(params[:id])
rescue ActiveRecord::RecordNotFound => ex
  render jsonapi_errors: ex.message, status: :not_found
end
```
NB. the `jsonapi_errors` bits are a guess, not 100% of the correct way to form that.

heidar: I've never had an issue with this or seen this become a problem. I've never seen a project that doesn't handle 404s in this way actually. But I don't really care either way.

defong: titbits

'out of the box experience'
show_detailed_exceptions

`ex.message be careful not to leak something that should not be leaked`
render `jsonapi_errors: ex.message`, status: :not_found

I have no ideas...

but had experience a number of handling exceptions (including the leaky bit as well)

dankmitchell: @misaka so the ask here is to move the rescue out of the application controller and into the individual controller actions, case by case?

heidar: The goal of this PR was just getting 404s to work, they can have an empty body. We should do another PR to introduce error objects as per the JSONAPI spec.

I'm think `render_errors` is meant for validation errors by the way, needs investigating.

PR #335, opened 2019-05-03
[1406] Improve course enrichment and site validation errors

tvararu:
### Context

This removes part of the work that the frontend has to do to display validation errors.

Add `custom_error_message` as a dependency. It allows us to prefix error messages with `^` when they should not use the attribute as a prefix for the full_message.

This modifies the `course#validate_enrichment` method. Before, all enrichment errors would be hoisted under the `latest_enrichment` key. This isn't enough information to be able to determine which field this error belongs to. The new key is computed as `latest_enrichment__FIELD`.

Updated tests and adds better validation messages for course enrichments and sites via `en.yml` overrides.

### Guidance to review

I really tried to find a way to do this without `custom_error_message` which is a very dirty monkey patch of an unmaintained library, but I couldn't due to the double `full_messages` call that gets run on the nested `course_enrichment` validations... maybe it would be cleaner to reimplement the `jsonapi_error` renderer? Neither option seems particularly fun...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN apk add --update --no-cache tzdata && \
     echo "Europe/London" > /etc/timezone
 
 RUN apk add --update --no-cache --virtual runtime-dependances \
- postgresql-dev
+ postgresql-dev git
 
 ENV APP_HOME /app
 RUN mkdir $APP_HOME
--- a/Gemfile
+++ b/Gemfile
@@ -78,6 +78,9 @@ gem 'aasm'
 # Handle data migrations
 gem 'data_migrate'
 
+# Allows writing of error full_messages for validations that don't start with the attribute name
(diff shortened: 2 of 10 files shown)


tvararu:
Does Travis not know how to bundle git sha references?

```bash
Retrying `git clone 'https://github.com/nanamkim/custom-err-msg.git' "/usr/local/bundle/cache/bundler/git/custom-err-msg-0b16c72d578bcce543bbd21a4d5cc9123a4cbdc8" --bare --no-hardlinks --quiet` due to error (2/4): Bundler::Source::Git::GitCommandError Git error: command `git clone 'https://github.com/nanamkim/custom-err-msg.git' "/usr/local/bundle/cache/bundler/git/custom-err-msg-0b16c72d578bcce543bbd21a4d5cc9123a4cbdc8" --bare --no-hardlinks --quiet` in directory /app has failed.sh: git: not found
```

PR #336, opened 2019-05-05
1364 Expose list of access requests in mcbe to mc support

davidgisbey:
### Context

MC support currently hits the db directly for to list the access requests.  This pr creates an endpoint on MCBE for the support app to hit instead

- endpoint created
- list of 'requested' access requests returned when endpoint is hit

### Guidance to review
The support app still needs to be refactored to hit the mcbe endpoint.

### DIFF
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -1,8 +1,7 @@
 module API
   module V2
     class AccessRequestsController < API::V2::ApplicationController
-      before_action :build_access_request, except: :list
-      before_action :build_list_of_access_requests, except: :approve
+      before_action :build_access_request, only: :approve
 
       def approve
         result = AccessRequestApprovalService.call(@access_request)
@@ -10,18 +9,17 @@ module API
         render status: 200, json: { result: result }
       end
 
-      def list
-        render status: 200, json: @list_of_access_requests
+      def index
+        authorize AccessRequest
+        @access_requests = AccessRequest.requested
+
+        render status: 200, json: @access_requests
       end
 
     private
 
       def build_access_request
         @access_request = authorize AccessRequest.requested.find(params[:id])
-      end
-
-      def build_list_of_access_requests
-        @list_of_access_requests = authorize AccessRequest.requested.all
       end
     end
   end
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -13,7 +13,7 @@ module API
         authorize AccessRequest
         @access_requests = AccessRequest.requested
 
-        render status: 200, json: @access_requests
+        render jsonapi: @access_requests
       end
 
     private
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -1,7 +1,8 @@
 module API
   module V2
     class AccessRequestsController < API::V2::ApplicationController
-      before_action :build_access_request
+      before_action :build_access_request, except: :list
+      before_action :build_list_of_access_requests, except: :approve
 
       def approve
         result = AccessRequestApprovalService.call(@access_request)
@@ -9,10 +10,18 @@ module API
         render status: 200, json: { result: result }
       end
 
+      def list
+        render status: 200, json: @list_of_access_requests
+      end
+
     private
 
       def build_access_request
         @access_request = authorize AccessRequest.requested.find(params[:id])
+      end
+
+      def build_list_of_access_requests
+        @list_of_access_requests = authorize AccessRequest.requested.all
       end
     end
   end
--- a/app/models/access_request.rb
+++ b/app/models/access_request.rb
@@ -24,6 +24,10 @@ class AccessRequest < ApplicationRecord
     declined
   ].freeze
 
+  def recipient
+    User.new(first_name: first_name, last_name: last_name, email: email_address)
+  end
+
   alias_method :approve, :completed!
 
   audited
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -75,8 +75,9 @@ Rails.application.routes.draw do
       resource :sessions
       resources :site_statuses, only: :update
 
-      resources :access_requests, only: :update do
+      resources :access_requests, only: %i[update show] do
         post :approve, on: :member
+        get :list, on: :member
       end
     end
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -75,9 +75,8 @@ Rails.application.routes.draw do
       resource :sessions
       resources :site_statuses, only: :update
 
-      resources :access_requests, only: %i[update show] do
+      resources :access_requests, only: %i[update index] do
         post :approve, on: :member
-        get :list, on: :member
       end
     end
   end
--- a/spec/factories/access_requests.rb
+++ b/spec/factories/access_requests.rb
@@ -22,8 +22,24 @@ FactoryBot.define do
     email_address { Faker::Internet.email }
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
-    organisation { create(:organisation) }
+    organisation { requester.organisations.first }
     reason { Faker::Lorem.sentence.to_s }
     requester_email { requester.email }
+  end
+
+  trait :requested do
+    status { :requested }
+  end
+
+  trait :approved do
+    status { :approved }
+  end
+
+  trait :completed do
+    status { :completed }
+  end
+
+  trait :declined do
+    status { :declined }
   end
 end
--- a/spec/policies/access_request_policy_spec.rb
+++ b/spec/policies/access_request_policy_spec.rb
@@ -3,19 +3,19 @@ require "rails_helper"
 describe AccessRequestPolicy do
   subject { described_class }
 
-  permissions :approve? do
+  permissions :approve?, :index? do
     let(:access_request) { build(:access_request) }
 
     context 'non-admin user' do
       let(:user) { build(:user) }
 
-      it { should_not permit(user, access_request) }
+      it { should_not permit(user) }
     end
 
     context 'admin user' do
       let(:user) { build(:user, :admin) }
 
-      it { should permit(user, access_request) }
+      it { should permit(user) }
     end
   end
 end
--- a/spec/requests/api/v2/access_request_spec.rb
+++ b/spec/requests/api/v2/access_request_spec.rb
@@ -24,15 +24,15 @@ describe 'Access Request API V2', type: :request do
 
   subject { response }
 
-  describe 'GET #list' do
-    let(:list_access_requests_route) do
-      get "/api/v2/access_requests/#{admin_user.id}/list",
+  describe 'GET #index' do
+    let(:access_requests_index_route) do
+      get "/api/v2/access_requests",
           headers: { 'HTTP_AUTHORIZATION' => credentials }
     end
 
     context 'when unauthenticated' do
       before do
-        list_access_requests_route
+        access_requests_index_route
       end
 
       let(:payload) { { email: 'foo@bar' } }
@@ -44,7 +44,7 @@ describe 'Access Request API V2', type: :request do
       let(:unauthorised_user) { create(:user) }
       let(:payload) { { email: unauthorised_user.email } }
       let(:unauthorised_user_route) do
-        get "/api/v2/access_requests/#{unauthorised_user.id}/list",
+        get "/api/v2/access_requests",
             headers: { 'HTTP_AUTHORIZATION' => credentials }
       end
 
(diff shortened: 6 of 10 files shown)


app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -1,19 +1,28 @@
...
+      before_action :build_access_request, except: :list
+      before_action :build_list_of_access_requests, except: :approve
 
       def approve
         result = AccessRequestApprovalService.call(@access_request)
 
         render status: 200, json: { result: result }
       end
 
+      def list
```
dankmitchell: As this is the collection the rails convention dictates it should be `index`

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -1,19 +1,28 @@
...
+      end
+
     private
 
       def build_access_request
         @access_request = authorize AccessRequest.requested.find(params[:id])
       end
+
+      def build_list_of_access_requests
+        @list_of_access_requests = authorize AccessRequest.requested.all
```
dankmitchell: This instance variable should be `@access_requests` and you shouldn’t need the `all`

config/routes.rb
```diff
@@ -75,8 +75,9 @@
       resource :sessions
       resources :site_statuses, only: :update
 
-      resources :access_requests, only: :update do
+      resources :access_requests, only: %i[update show] do
```
dankmitchell: `show is not part of this PR so no need for that here but I would add `index` then remove the line you added below

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -24,6 +24,115 @@
 
   subject { response }
 
+  describe 'GET #list' do
```
dankmitchell: list to index

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -24,6 +24,115 @@
...
+          [
+            {
+              "id" => first_access_request.id,
+              "email_address" => "first_user@test.com",
+              "first_name" => "test",
+              "last_name" => "user",
+              "requester_email" => "second_user@test.com",
+              "requester_id" => second_user.id,
+              "status" => "requested",
+              "organisation" => "First Organisation",
```
dankmitchell: For most of these values you can pull them from the parent variable I.e. `first_organisation.name`

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -24,6 +24,115 @@
...
+               organisation: second_organisation.name,
+               reason: 'Leaving current role',
+               request_date_utc: '2019-05-05 00:10:48 UTC',
+               status: 'requested')
+      }
+      before do
+        list_access_requests_route
+      end
+
+      it 'JSON has a data section with the correct attributes' do
```
benilovj: there should be a spec (this one or a separate one) that tests the filtering down to `requested` in the controller i.e. you'd seed the DB with an actioned access request and ensure that it's not returned in the response.

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -24,6 +24,115 @@
 
   subject { response }
 
+  describe 'GET #list' do
+    let(:list_access_requests_route) do
+      get "/api/v2/access_requests/#{admin_user.id}/list",
```
benilovj: Once you make the change as Dan's suggested, this will change to `get "/api/v2/access_requests"`. You don't want the admin user ID in there as they're unnecessary, the admin user details are already in the `Authorisation` header.

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -1,19 +1,28 @@
...
+        render status: 200, json: @list_of_access_requests
+      end
+
     private
 
       def build_access_request
         @access_request = authorize AccessRequest.requested.find(params[:id])
       end
+
+      def build_list_of_access_requests
```
benilovj: I wouldn't bother extracting this method just yet, as it's only used in one place.

app/controllers/api/v2/access_requests_controller.rb
benilovj: I think it'd be better to split this into two lines:

```ruby
authorize AccessRequest
@access_requests = AccessRequest.requested
```

The first line is a slightly different way to do the authorisation because lists of things are authorised slightly differently to individual things. See `API::V2::ProvidersController#index` and `API::V2::CoursesController#index` that follow the pattern of authorising the class, rather than the collection that you get back from ActiveRecord.

benilovj:
A comment regarding the structure of the API:

```ruby
            {
              "id" => first_access_request.id,
              "email_address" => "first_user@test.com",
              "first_name" => "test",
              "last_name" => "user",
              "requester_email" => "second_user@test.com",
              "requester_id" => second_user.id,
              "status" => "requested",
              "organisation" => "First Organisation",
              "reason" => "Need additional support",
              "request_date_utc" => "2019-05-05T00:10:47.000Z"
            },
```
I think it could be made more expressive, and harder to misuse, if `requester` and `recipient` were explicit associations of the access request of type `users`. Requester is already a proper foreign key. Recipient could be fed into the serialiser from the model in this way:

```ruby
class AccessRequest
  ...
  def recipient
    User.new(first_name: first_name, last_name: last_name, email: email_address)
  end
end
```

davidgisbey:
Thanks for the feedback guys.  Will action it tonight/tomorrow.

app/controllers/api/v2/access_requests_controller.rb
davidgisbey: Have actioned this. Thanks Dan.

app/controllers/api/v2/access_requests_controller.rb
davidgisbey: What's funny about this is i initially had it as @access_requests, but changed it due to Heidar's voice popping into my head saying it wasn't very clear. Have actioned this. Thanks for the feedback.

config/routes.rb
davidgisbey: Done. Thanks.

spec/requests/api/v2/access_request_spec.rb
davidgisbey: Done. Thanks

spec/requests/api/v2/access_request_spec.rb
davidgisbey: Done. Thanks for the feedback.

spec/requests/api/v2/access_request_spec.rb
davidgisbey: Done. Thanks

davidgisbey: There is one exception. When JSON parses a datetime into a string it seems to output it in a slightly different format. However, it is the correct time so i've manually made that attribute by inputting the output value in json. 

before 

`'2019-05-05 00:10:48 UTC'`
after
`'2019-05-05T00:10:47.000Z'`
🤷‍♂

spec/requests/api/v2/access_request_spec.rb
davidgisbey: Added this into the (now) #index req spec test by let!'ing an approved access request prior to hitting the endpoint. It's not being parsed into JSON 👍

davidgisbey:
Regarding your comments on the structure of the API. I'll pick this up tomorrow (today i guess as it's past midnight) as it will involve refactoring the previous access request work. As it looks like a fairly substantial piece of work, is it worth opening a separate PR or are you happy for me to change  approve endpoint work in this PR?

app/controllers/api/v2/access_requests_controller.rb
davidgisbey: Cool 👍

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -1,19 +1,29 @@
...
+      before_action :build_access_request, except: :index
+      before_action :build_access_requests, except: :approve
 
       def approve
         result = AccessRequestApprovalService.call(@access_request)
 
         render status: 200, json: { result: result }
       end
 
+      def index
```
dankmitchell: Jake has mentioned this but I think it was missed.

This method should be consistent with our other controllers and read like the below so you can remove the build method. We only use a `before_action` to DRY the code up when we're repeating something on multiple actions or to make it read easier.

```
    def index
        authorize AccessRequest
        @access_requests = AccessRequest.requested

        render jsonapi: @access_requests
    end
```

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -1,19 +1,29 @@
 module API
   module V2
     class AccessRequestsController < API::V2::ApplicationController
-      before_action :build_access_request
+      before_action :build_access_request, except: :index
```
dankmitchell: Might be more explicit to say use `only` rather than `except` i.e. `before_action :build_access_request, only: :show`

app/controllers/api/v2/access_requests_controller.rb
davidgisbey: Yea sorry i missed that one. Have now actioned it. Thanks Dan.

app/controllers/api/v2/access_requests_controller.rb
davidgisbey: Great. Just changed it to 
` before_action :build_access_request, only: :approve`
Thanks Dan.

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -24,6 +24,117 @@
...
+      end
+
+
+      it "should raise an error" do
+        expect { unauthorised_user_route }.to raise_error Pundit::NotAuthorizedError
+      end
+    end
+
+    context 'when authorised' do
+      let(:first_organisation) { create(:organisation, name: 'First Organisation') }
```
heidar: Do we need all this setup? Feel like a lot of this can come from the factories instead of being explicitly set.

spec/policies/access_request_policy_spec.rb
```diff
@@ -18,4 +18,20 @@
...
     end
   end
+
+  permissions :index? do
+    let(:access_request) { build(:access_request) }
+
+    context 'non-admin user' do
+      let(:user) { build(:user) }
+
+      it { should_not permit(user, access_request) }
```
heidar: I don't think you need the `access_request` here. You should be able to just do 

`it { should_not permit(user) }`

That's because the `index?` method doesn't need an access request - we only need the user to see if it can get the list of access requests.

heidar: Separately, if the `index?` and `approve?` would be exactly the same, you can do `permissions :approve?, :index? do ...`

See https://github.com/varvet/pundit#policy-specs

Just figured I'd share, I had forgotten myself :D

app/models/access_request.rb
```diff
@@ -24,6 +24,10 @@ class AccessRequest < ApplicationRecord
     declined
   ].freeze
 
+  def recipient
```
heidar: What is this used for?

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -24,6 +24,79 @@
...
+              "last_name" => second_access_request.recipient.last_name,
+              "requester_email" => second_access_request.requester.email,
+              "requester_id" => second_access_request.requester.id,
+              "organisation" => second_access_request.organisation,
+              "reason" => second_access_request.reason,
+              "request_date_utc" => '2019-05-05T00:10:48.000Z',
+              "status" => second_access_request.status
+            }
+          ]
+       )
```
misaka: Is there a reason this is not a JSON API response? I would expect it to be, offhand ...

spec/policies/access_request_policy_spec.rb
misaka: > That's because the index? method doesn't need an access request

I think it needs `AccessRequest` since technically we're checking the user against the model class, and not (as you say) an individual access request.

heidar: hmm I thought that was just for the magic in `authorize` to work

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -1,14 +1,21 @@
...
         result = AccessRequestApprovalService.call(@access_request)
 
         render status: 200, json: { result: result }
       end
 
+      def index
+        authorize AccessRequest
+        @access_requests = AccessRequest.requested
+
+        render status: 200, json: @access_requests
```
heidar: oh yeah, good catch @misaka - this should be `jsonapi: @access_requests`

spec/factories/access_requests.rb
```diff
@@ -22,8 +22,24 @@
     email_address { Faker::Internet.email }
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
-    organisation { create(:organisation) }
+    organisation { requester.organisations.first.name }
```
defong: is meant to be `name` and not `organisation`?

spec/policies/access_request_policy_spec.rb
heidar: Just noticed that we alias `index?` to `approve?` so maybe let's test these the way I suggested before:

`permissions :approve?, :index? do ...`

By modifying the test for `approve?` that is. Then we can get rid of this one.

spec/factories/access_requests.rb
davidgisbey: You are correct. Done. Thanks.

spec/policies/access_request_policy_spec.rb
davidgisbey: Cool. Done. Thanks guys

misaka: Spoke in-person, makes sense to me assuming `AccessRequest` really isn't needed.

PR #339, opened 2019-05-07
Add newqa env

smneal:
### Context
Add new QA env and config to support readiness testing

Addition of various configs to support new_qa


dankmitchell:
Is this just for testing @smneal ?

PR #342, opened 2019-05-08
[1363] Add seeds for access request table

timabell:
Without this it's hard to test the support app being able to action them
while connected to local development backend app.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -168,6 +168,7 @@ AccessRequest.create!(
   first_name: Faker::Name.first_name,
   last_name: Faker::Name.last_name,
   requester: access_requester_user,
+  requester_email: access_requester_user.email,
   request_date_utc: Time.now - 1.week,
   status: :requested,
 )
@@ -177,6 +178,7 @@ AccessRequest.create!(
   first_name: Faker::Name.first_name,
   last_name: Faker::Name.last_name,
   requester: access_requester_user,
+  requester_email: access_requester_user.email,
   request_date_utc: Time.now - 2.weeks,
   status: :completed,
   )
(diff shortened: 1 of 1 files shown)


timabell:
[Allow mc-support to grant access requests through mc-be](https://trello.com/c/VdmjSTgi/1363-allow-mc-support-to-grant-access-requests-through-mc-be)

db/seeds.rb
```diff
@@ -132,3 +141,42 @@
 
   user.organisations << organisation
 end
+
+access_requester_user = User.create!(
+  first_name: 'Jane',
```
dankmitchell: Can we use Faker? i.e. `Faker::Name.unique.name`

timabell: good idea, will do

db/seeds.rb
```diff
@@ -132,3 +141,44 @@
...
+  org_id: '12345',
+  providers: [
+    accrediting_provider,
+    provider2,
+  ],
+  users: [
+    access_requester_user,
+    admin_user,
+  ],
+)
```
misaka: Can we simplify this a little? I think that makes the seeds.rb file harder to read and reason about, it would be nice to just use the already-created users and providers.

db/seeds.rb
```diff
@@ -132,3 +141,25 @@
 
   user.organisations << organisation
 end
+
+access_requester_user = (User.select { |u| !u.admin? }).sample
```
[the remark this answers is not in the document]
timabell: @dankmitchell we couldn't make it work, didn't seem to think reject was defined for

    User.reject(&:admin?)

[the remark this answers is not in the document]
dankmitchell: Ah yes, you probably need to call `.all` first. I'll sort it in repayment for the API work ;-)

PR #345, opened 2019-05-08
[1413] Fix email already in use error during user sign in

heidar:
### Context

The database constraint for email uniqeness on users causes ActiveRecord
to raise an error whenever we try to update the email to one that
already exists.

This happens for some users that have a DfE-SignIn ID on a user record
with an email address different from the one that we get from
DfE-SignIn, and the email from DfE-SignIn matches another record.

When we look up the user by the ID and the email doesn't match the
user's email, we want to update their email to the one we get from
DfE-SignIn.

Sometimes though, there's already another user on the system with the
same email, so we get an error. Instead of just not updating and moving
on with authenticatation, we want to send an exception to Sentry with
the details of the users.

This allows us to start collecting some data about the users affected by
this, which will help us come up with a plan to resolve this, wheter
that is manually or automatically.

This also authenticates the user as normal, rather than failing.

### Guidance to review

I think we need to test the Sentry reporting on QA?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/authentication_service.rb
+++ b/app/services/authentication_service.rb
@@ -11,7 +11,11 @@ class AuthenticationService
 
   def call
     @user = user_by_sign_in_user_id || user_by_email
-    update_user_email if update_user_email_needed?
+    update_user_email if user_email_does_not_match_token?
+
+    user
+  rescue DuplicateUserError => e
+    Raven.capture(e)
 
     user
   end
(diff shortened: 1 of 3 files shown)


heidar:
> Looks fine to me, but was this a bug that was introduced by ruby/recently?

Yeah, it started happening as soon as we started updating emails when users authenticate.

PR #349, opened 2019-05-09
[1363] Add more access_request records when seeding

timabell:
Once you've actioned one it's gone, so useful to have more for testing.

Cleaner code suggestions would be appreciated.


timabell:
[Allow mc-support to grant access requests through mc-be](https://trello.com/c/VdmjSTgi/1363-allow-mc-support-to-grant-access-requests-through-mc-be)

PR #350, opened 2019-05-09
[1457] Add prettier

tvararu:
### Context

This adds prettier as a project dependency, allowing us to run editor plugins to automatically indent / format files as we go.

- Add prettier as an npm dependency (because the `gem 'prettier'` version gets ignored by my editor Atom for some reason... probably works for other editors, but please test)
- Add instructions for how to run prettier
- Disable rubocop rules that conflict with prettier

### Guidance to review

Try this locally! Configure your editor, go to any file (`app/models/provider.rb` is good), and save it / format it. I highly recommend using the `format on save` option and being as lazy as humanly possible about indentation; let the tool do it.

I've stopped short of formatting *all* the files because in some niche situations I think the changes deserve manual review.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu:
[Use prettier to auto-format Ruby and ERB](https://trello.com/c/PKPT87Lt/1457-use-prettier-to-auto-format-ruby-and-erb)

README.md
```diff
@@ -73,6 +75,17 @@ or
 docker-compose exec web /bin/sh -c "bundle exec rubocop app config db lib spec Gemfile --format clang"
 ```
 
+### Prettier
+
+Auto-formatting is available using [prettier](https://github.com/prettier/plugin-ruby). To run on the entire project:
+
+```bash
+yarn
+./node_modules/.bin/prettier --write '{app,config,lib,spec}/**/*.rb'
```
defong: can we not bundle with `bundle exec rake`

tvararu: I don't think we should; I don't think the build should fail if it fails formatting.

defong: cool, so no more linting failed build reason? does that mean we need t change CI/travis/vsts?

tvararu: Nah, I think we should keep the current rubocop setup, but not add prettier to rake at this stage.

tvararu:
I actually now propose parking this, as I think Rubocop deserves a bit more exploration, based on my chats with @heidar and @misaka.

Will leave the branch open.

PR #353, opened 2019-05-10
[1407] Fix mcb azure integration on Windows

benilovj:
### Context
When attempting to run `mcb` on windows, spaces in the subscription name were tripping the command up on Windows.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


benilovj:
Now with actual working tests! 🎉

PR #355, opened 2019-05-10
[1310] Fixed additional validation issues

defong:
### Context
Misplaced validation
Missing validation for `course_length`
Validation message to be the same as the csharp version

Fixed misplaced validation
Fixed missing validation for `course_length`
Fixed validation message to be the same as the csharp version

### Guidance to review

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


defong:
[Rebuild course publishing validations in Rails](https://trello.com/c/g7MtJU3T/1310-rebuild-course-publishing-validations-in-rails)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -16,7 +16,20 @@ def show
       end
 
       def sync_with_search_and_compare
+        response = ManageCoursesAPIService::Request.sync_course_with_search_and_compare(
+          @current_user.email,
+          @provider.provider_code,
+          @course.course_code
+        )
+
+        head response ? :ok : :internal_server_error
```
misaka: For the future ... this is going to swallow up errors and not log them. We should not do that.

misaka: Added a note to expose errors to the ticket about syncing courses (which will re-write this part).

PR #356, opened 2019-05-10
[1470] add mcb user show

misaka:
### Context

When debugging production problems it's handy to be able to view the user info in production without having to make a db connection.

Add commands to mcb:
- `mcb user list`
- `mcb user show <id>`
- `mcb user audit <id>`

### Guidance for review

Not sure why CodeClimate is failing, ignoring for now.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/commands/users/list.rb
+++ b/lib/mcb/commands/users/list.rb
@@ -4,7 +4,7 @@ run do |opts, args, _cmd|
   MCB.init_rails(opts)
 
   users = if args.any?
-            User.where(id: args.to_a)
+            args.map { |id| MCB.find_user_by_identifier id }
           else
             User.all
           end
@@ -12,5 +12,11 @@ run do |opts, args, _cmd|
   tp.set :capitalize_headers, false
 
   puts "\nUsers:"
-  tp users, 'id', 'email', 'first_name', 'last_name', 'last_login_date_utc'
+  puts Tabulo::Table.new(users,
+                         :id,
+                         :email,
+                         :sign_in_user_id,
+                         :first_name,
+                         :last_name,
+                         :last_login_date_utc).pack(max_table_width: nil)
 end
--- a/lib/mcb/commands/users/list.rb
+++ b/lib/mcb/commands/users/list.rb
@@ -1,0 +1,16 @@
+summary 'List users in the tb'
+
+run do |opts, args, _cmd|
+  MCB.init_rails(opts)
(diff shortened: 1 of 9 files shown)


misaka:
[List users in mcb](https://trello.com/c/S44WFX3D/1470-list-users-in-mcb)

lib/mcb/commands/users/list.rb
```diff
@@ -0,0 +1,16 @@
+summary 'List users in the tb'
+
+run do |opts, args, _cmd|
+  MCB.init_rails(opts)
+
+  users = if args.any?
+            User.where(id: args.to_a)
+          else
+            User.all
```
benilovj: Whats the use case for getting all users listed?

lib/mcb/commands/users/list.rb
```diff
@@ -0,0 +1,16 @@
+summary 'List users in the tb'
+
+run do |opts, args, _cmd|
+  MCB.init_rails(opts)
+
+  users = if args.any?
+            User.where(id: args.to_a)
```
benilovj: Do we want one of these for dfe sign-in id? Maybe also by email address?

lib/mcb/commands/users/list.rb
misaka: I imagine it would be useful for searching for a user if you weren't sure of their userid, for example, just pass the output into less and you can search it easily, etc. Or even just to test that things are working ...

lib/mcb/commands/users/list.rb
misaka: Ya, I think that those would be nice to haves, although I'd be more likely to add that to the `show` command since I imagine that's more likely to be used to pull out individual users. At the very least maybe we should be exposing the DFE sign-in ids here.

lib/mcb/commands/users/list.rb
misaka: Just a thought, you're not mixing up the `list` and `show` commands are you?

lib/mcb/commands/users/list.rb
benilovj: Quite right, I meant the `show` command.

lib/mcb/commands/users/list.rb
benilovj: 👍

PR #357, opened 2019-05-10
[1472] Add course.toggle_site method, and more

tvararu:
### Context

These are various model methods that are used by the forthcoming new MCB scripts.

This allows us to toggle sites on courses through MCB as well as present different kinds of data, such as the potential sites that can be attached to a course.

Implements the new methods, alongside tests.

### Guidance to review

Easiest reviewed commit by commit.

I have not ticked the "Cleaned commit history" because I don't know if I should add more context to the existing 6 commits, or squash them into one. They all do one small method each which I find pleasing but maybe not all reviewers agree.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -227,6 +227,16 @@ class Course < ApplicationRecord
     end
   end
 
+  def add_site!(site:)
+    is_course_new = new? # persist this before we change anything
+    site_status = site_statuses.find_or_create_by!(site: site)
+    site_status.start! unless is_course_new
+  end
+
+  def remove_site!(site:)
+    site_statuses.find_by!(site: site).suspend!
+  end
+
   def sites_not_associated_with_course
     provider.sites - sites
   end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -226,4 +226,8 @@ class Course < ApplicationRecord
       enrichment.publish(current_user)
     end
   end
+
+  def sites_not_associated_with_course
+    provider.sites - sites
+  end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -217,7 +217,7 @@ class Course < ApplicationRecord
   end
 
   def publish_sites
-    site_statuses.status_new_status.each(&:status_running!)
+    site_statuses.status_new_status.each(&:start!)
     site_statuses.status_running.unpublished_on_ucas.each(&:published_on_ucas!)
   end
 
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -16,6 +16,9 @@ class SiteStatus < ApplicationRecord
 
   self.table_name = "course_site"
 
+  after_initialize :set_defaults
+  before_validation :set_vac_status
+
   audited associated_with: :course
 
   validate :vac_status_must_be_consistent_with_course_study_mode,
@@ -69,6 +72,16 @@ class SiteStatus < ApplicationRecord
   end
 
 private
+
+  def set_defaults
+    self.status ||= :new_status
+    self.applications_accepted_from ||= Date.today
+    self.publish ||= :unpublished
+  end
+
+  def set_vac_status
+    self.vac_status ||= self.class.default_vac_status_given(study_mode: course.study_mode)
+  end
 
   def vac_status_must_be_consistent_with_course_study_mode
     unless vac_status_consistent_with_course_study_mode?
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -51,6 +51,19 @@ class SiteStatus < ApplicationRecord
   scope :with_vacancies, -> { where.not(vac_status: :no_vacancies) }
   scope :open_for_applications, -> { findable.applications_being_accepted_now.with_vacancies }
 
+  def self.default_vac_status_given(study_mode:)
+    case study_mode
+    when "full_time"
+      :full_time_vacancies
+    when "part_time"
+      :part_time_vacancies
+    when "full_time_or_part_time"
+      :both_full_time_and_part_time_vacancies
+    else
+      raise "Unexpected study mode #{study_mode}"
+    end
+  end
+
   def description
     "#{site.description} – #{status}/#{publish}"
   end
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -51,6 +51,10 @@ class SiteStatus < ApplicationRecord
   scope :with_vacancies, -> { where.not(vac_status: :no_vacancies) }
   scope :open_for_applications, -> { findable.applications_being_accepted_now.with_vacancies }
 
+  def description
+    "#{site.description} – #{status}/#{publish}"
+  end
+
 private
 
   def vac_status_must_be_consistent_with_course_study_mode
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -13,6 +13,7 @@
 
 class SiteStatus < ApplicationRecord
   include TouchCourse
+  include AASM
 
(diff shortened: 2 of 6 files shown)


tvararu:
[Ability to "toggle" training locations on courses in code](https://trello.com/c/ULEPU2K8/1472-ability-to-toggle-training-locations-on-courses)

benilovj:
@tvararu I made a couple of tweaks:

- replaced the `course.actual_and_potential_site_statuses` method with the far simpler `sites_not_associated_with_course` method
- broke up the specs for the state machine and added a little more coverage

tvararu:
Thanks @benilovj 👍

app/models/course.rb
```diff
@@ -226,4 +226,26 @@ def publish_enrichment(current_user)
       enrichment.publish(current_user)
     end
   end
+
+  def toggle_site(site_to_toggle)
+    is_course_new = new? # persist this before we change anything
+
+    site_status = site_statuses.find_or_initialize_by(site: site_to_toggle) do |ss|
+      ss.vac_status = SiteStatus.default_vac_status_given(study_mode: study_mode)
```
benilovj: I just realised there may be a better way to do this: move this block into an `after_initialize` filter on `SiteStatus`.

benilovj: Fixed in bb0cbf9

app/models/course.rb
```diff
@@ -226,4 +226,20 @@ def publish_enrichment(current_user)
...
+    is_course_new = new? # persist this before we change anything
+
+    site_status = site_statuses.find_or_initialize_by(site: site_to_toggle)
+    if !site_status.new_record?
+      site_status.toggle_status!
+    else
+      site_status.save!
+      site_status.start! unless is_course_new
+    end
+  end
```
misaka: This isn't very clear to me wrt what it's actually doing. Discussed IRL so I'll leave it with you. :)

benilovj: I've replaced `toggle` with `add_site!` and `remove_site!`

spec/models/site_status_spec.rb
```diff
@@ -168,4 +170,35 @@
...
+      expect(subject.default_vac_status_given(study_mode: 'part_time')).to eq :part_time_vacancies
+      expect(subject.default_vac_status_given(study_mode: 'full_time_or_part_time')).to eq :both_full_time_and_part_time_vacancies
+      expect { subject.default_vac_status_given(study_mode: 'foo') }.to raise_error("Unexpected study mode foo")
+    end
+  end
+
+  describe "status changes" do
+    describe "when suspending a running, published site status" do
+      subject { create(:site_status, :running, :published).tap(&:suspend!) }
+      its(:status) { should eq("suspended") }
```
misaka: Just a suggestion, but this would read a little better and also hide the implementation detail:

```
it { should be_status_suspended }
```

spec/models/site_status_spec.rb
```diff
@@ -168,4 +170,35 @@
...
+      its(:publish) { should eq("unpublished") }
```
misaka:
```
it { should be_published_on_ucas }
```

spec/models/site_status_spec.rb
```diff
@@ -168,4 +170,35 @@
...
+    end
+
+    %i[new suspended discontinued].each do |status|
+      describe "when starting a #{status}, unpublished site status" do
+        subject { create(:site_status, status, :unpublished).tap(&:start!) }
+        its(:status) { should eq("running") }
```
misaka:
```
it { should be_status_running }
```

spec/models/site_status_spec.rb
```diff
@@ -168,4 +170,35 @@
...
+        its(:publish) { should eq("published") }
```
misaka:
```
it { should be_published_on_ucas }
```

app/models/site_status.rb
```diff
@@ -35,6 +39,27 @@ class SiteStatus < ApplicationRecord
...
+    after_all_transitions :update_publish_flag
+
+    event :start do
+      transitions from: %i[new_status suspended discontinued], to: :running
+    end
+
+    event :suspend do
+      transitions from: :running, to: :suspended
+    end
+  end
```
misaka: Seeing as we have an `enum` defined below here, maybe we should move this `aasm` definition and the `update_publish_flag` method definition down a little?

spec/models/course_spec.rb
```diff
@@ -421,4 +405,67 @@
...
+    let(:first_site) { create(:site) }
+    let(:second_site) { create(:site) }
+    let(:provider) { create(:provider, sites: [first_site, second_site]) }
+    let(:first_site_status) { create(:site_status, :findable, site: first_site) }
+    subject { create(:course, provider: provider, site_statuses: [first_site_status]) }
+
+    its(:sites_not_associated_with_course) { should eq([second_site]) }
+  end
+
+  context "adding and removing sites on a course" do
```
misaka: At the risk of sounding picky ... can we make this a `describe`? We should reserve `context` to define the setup for a test, and `describe` is what is being tested. If that makes any sense. :)

spec/models/course_spec.rb
```diff
@@ -421,4 +405,67 @@
...
+    let(:provider) { create(:provider) }
+    let(:new_site) { create(:site, provider: provider) }
+    let(:existing_site) { create(:site, provider: provider) }
+    let(:new_site_status) { subject.site_statuses.find_by!(site: new_site) }
+    subject { create(:course, site_statuses: [existing_site_status]) }
+
+    describe "for running courses" do
```
misaka: This seems to be the `context` here, describing the setup within which tests are running.

app/models/site_status.rb
```diff
@@ -35,6 +39,27 @@ class SiteStatus < ApplicationRecord
...
+
+  def update_publish_flag
+    self.publish = (aasm.to_state == :running ? :published : :unpublished)
```
misaka: Is this change being saved? I don't see where and the only tests I've managed to find (admittedly with only a cursory look) seem to test the `site_status` object without reloading it, which could be prone to giving a false positive.

spec/models/course_spec.rb
benilovj: fixed

spec/models/course_spec.rb
benilovj: fixed

app/models/site_status.rb
benilovj: fixed

spec/models/site_status_spec.rb
benilovj: fixed

spec/models/site_status_spec.rb
benilovj: fixed

spec/models/site_status_spec.rb
benilovj: fixed

spec/models/site_status_spec.rb
benilovj: fixed

app/models/site_status.rb
benilovj: My understanding from [how we're supposed to use the `aasm` library](https://github.com/aasm/aasm/blob/master/README.md#bang-events) is that we don't save explicitly; instead we modify the underlying object and call the `!` method (e.g. `start!`) if we want to auto-save.

I've added `.reload` to the site status specs to ensure that we're not testing stale objects, and the specs are still passing.

tvararu:
🎉

PR #359, opened 2019-05-13
[1469] Training location code: validations and automatic assignment

benilovj:
### Context
The UCAS Apply system uses training location codes and so each training location assigned to a provider needs one. There's a maximum of 37 single-character codes per provider: `0-9`, `A-Z`, `-`

When a provider creates a new location:
- we need to assign that location a valid campus code
- we should also validate that the code is one that will be accepted downstream
- we want to assign easily-confused codes last

- `Site` model sets `code` value by default
- active record validations added

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -22,6 +22,8 @@ class Site < ApplicationRecord
   include TouchProvider
 
   POSSIBLE_CODES = (('A'..'Z').to_a + ('0'..'9').to_a + ['-']).freeze
+  EASILY_CONFUSED_CODES = %w[1 I 0 O -].freeze # these ought to be assigned last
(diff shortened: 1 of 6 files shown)


app/models/site.rb
```diff
@@ -21,6 +21,12 @@ class Site < ApplicationRecord
   include RegionCode
   include TouchProvider
 
+  POSSIBLE_CODES = (('A'..'Z').to_a + ('0'..'9').to_a + ['-']).freeze
+  EASILY_CONFUSED_CODES = %w[1 I 0 O -].freeze # these ought to be assigned last
```
fofr: 👌

PR #360, opened 2019-05-13
Add help for all subcommands with `-v -h`

timabell:
It's getting harder to find where a command is given the depth and
number of things. Run `bin/mcb -v -h` to see the full list.

Selection_613

### DIFF
--- a/bin/mcb
+++ b/bin/mcb
@@ -9,6 +9,7 @@
 
 require 'ap'
 require 'cri'
+require 'rainbow'
 require 'pry'
 require 'table_print'
 require 'terminal-table'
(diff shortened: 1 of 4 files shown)


bin/mcb
```diff
@@ -36,6 +37,11 @@ $mcb = Cri::Command.define do
 
   flag :h, :help, 'show help for this command' do |_value, cmd|
     puts cmd.help
+    if $verbosity == :verbose
+      puts
+      puts Rainbow("SUBCOMMANDS").red.bright
+      show_all_commands(cmd, "")
+    end
```
misaka: Just a guess, but this won't work if the command switches are done in the order: -h -v, will it? That's a bother, if true. :(

timabell: You are correct sir, I have no idea what to do about it....

timabell: Kicks can down road... :-)

PR #371, opened 2019-05-16
[1493] Add create endpoint for sites

tvararu:
### Context

Adds a create endpoint which is similar to the other endpoints.

The validations and response are not exhaustively tested as the `update` endpoint already covers them and this endpoint does not differ significantly.

- Adds the `sites#create` controller action, route, authorisation

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu:
[Create create-location endpoint in mc-be](https://trello.com/c/3BTKJtyd/1493-create-create-location-endpoint-in-mc-be)

PR #373, opened 2019-05-17
[1496] Use site.save instead of save! when creating

tvararu:
### Context

Adding new locations.

This stop the backend from blowing up and instead sending down the validation errors.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/sites_controller.rb
+++ b/app/controllers/api/v2/sites_controller.rb
@@ -18,7 +18,7 @@ module API
         @site.provider = @provider
         authorize @site
 
-        if @site.save!
+        if @site.save
           render jsonapi: @site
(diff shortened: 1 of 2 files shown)


tvararu:
[New training location form: validations](https://trello.com/c/L5dCY5nT/1496-new-training-location-form-validations)

benilovj:
Are we missing specs that would have caught this?

tvararu:
@benilovj yeah, I deliberately didn't copy+paste from the update tests because I thought "oh, everything is the same, why test it twice..."

tvararu:
@benilovj added a spec

PR #374, opened 2019-05-17
[1497] Add can_add_more_sites? to provider serializer

tvararu:
### Context

We need to hide the button to add new locations when this is false.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -140,4 +140,16 @@ describe Provider, type: :model do
 
     its(:unassigned_site_codes) { should eq(expected_unassigned_codes) }
   end
+
+  describe "#can_add_more_sites?" do
+    context "when provider has less sites than max allowed" do
+      subject { create(:provider, site_count: 0) }
+      its(:can_add_more_sites?) { should be_truthy }
+    end
+
+    context "when provider has the max sites allowed" do
+      subject { create(:provider, site_count: Site::POSSIBLE_CODES.size) }
+      its(:can_add_more_sites?) { should be_falsey }
+    end
+  end
 end
--- a/spec/requests/api/v2/providers_spec.rb
+++ b/spec/requests/api/v2/providers_spec.rb
@@ -45,6 +45,7 @@ describe 'Providers API v2', type: :request do
               "provider_code" => provider.provider_code,
               "provider_name" => provider.provider_name,
               "accredited_body?" => false,
+              "can_add_more_sites?" => true
             },
             "relationships" => {
               "sites" => {
@@ -81,6 +82,7 @@ describe 'Providers API v2', type: :request do
               "provider_code" => provider.provider_code,
               "provider_name" => provider.provider_name,
               "accredited_body?" => false,
+              "can_add_more_sites?" => true
             },
             "relationships" => {
               "sites" => {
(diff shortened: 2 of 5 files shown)


tvararu:
[Publishers shouldn't be able to add more than 37 training locations](https://trello.com/c/1HgAUmLX/1497-publishers-shouldnt-be-able-to-add-more-than-37-training-locations)

spec/requests/api/v2/providers_spec.rb
```diff
@@ -236,6 +240,46 @@
...
+      let!(:provider) { create(:provider, course_count: 0, site_count: Site::POSSIBLE_CODES.size, organisations: [organisation]) }
+
+      before do
+        get "/api/v2/providers/#{provider.provider_code}",
+          headers: { 'HTTP_AUTHORIZATION' => credentials }
+      end
+
+      it 'has a data section with the correct attributes' do
+        json_response = JSON.parse(response.body)
+        expect(json_response).to eq(
```
benilovj: I wonder whether we should only assert the `can_add_more_sites?` on this response, to make the spec more explicit and less prone to breakage if unrelated things change?

tvararu: Agree

spec/models/provider_spec.rb
```diff
@@ -140,4 +140,16 @@
...
+    it "should return true when provider has less sites than max allowed" do
+      provider = create(:provider, site_count: 0)
+      expect(provider.can_add_more_sites?).to eq(true)
+    end
+
+    it "should return false when provider has the max sites allowed" do
+      provider = create(:provider, site_count: Site::POSSIBLE_CODES.size)
+      expect(provider.can_add_more_sites?).to eq(false)
+    end
+  end
```
benilovj: I'd have a slight preference for the following style which is what we've been doing in other places:

```ruby
  describe "#can_add_more_sites?" do
    context "when provider has less sites than max allowed" do
      subject { create(:provider, site_count: 0) }
      its(:can_add_more_sites?) { should be_truthy }
    end

    context "when provider has the max sites allowed" do
      subject { create(:provider, site_count: 37) }
      its(:can_add_more_sites?) { should be_falsey }
    end
  end
```

@misaka et al, thoughts?

spec/requests/api/v2/providers_spec.rb
```diff
@@ -236,6 +240,20 @@
       end
     end
 
+    context "with the maximum number of sites" do
+      let!(:provider) { create(:provider, course_count: 0, site_count: Site::POSSIBLE_CODES.size, organisations: [organisation]) }
```
misaka: Minor thing, but you don't need `let!` here as the `get` request below already references `provider`. It can also get in the way in the future by creating extra data you don't need.

tvararu: There's a `let!` above and I don't know why

benilovj: It was probably copied from the spec above that one :trollface:

tvararu:
[Publishers shouldn't be able to add more than 37 training locations](https://trello.com/c/1HgAUmLX/1497-publishers-shouldnt-be-able-to-add-more-than-37-training-locations)

PR #375, opened 2019-05-17
[1491] Create endpoint returns the created access request

davidgisbey:
### Context
The create endpoint should return an access request when hit. This is needed in the support app to hit the approve endpoint.

Returns the access request created via jsonapi.

### DIFF
--- a/spec/requests/api/v2/access_request_spec.rb
+++ b/spec/requests/api/v2/access_request_spec.rb
@@ -221,6 +221,23 @@ describe 'Access Request API V2', type: :request do
         Timecop.return
       end
 
+      it 'returns the correct id' do
+        string_id = JSON.parse(response.body)["data"]['id']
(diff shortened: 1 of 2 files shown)


spec/requests/api/v2/access_request_spec.rb
```diff
@@ -60,9 +60,15 @@
       let!(:second_access_request) { create(:access_request) }
 
       before do
+        Timecop.freeze
```
misaka: I don't see anything testing time below ... why do we need to freeze time here?

davidgisbey: Without Timecop line 88

```                "request_date_utc" => first_access_request.request_date_utc.iso8601,
```
and line 106

```               "request_date_utc" => second_access_request.request_date_utc.iso8601,
```
was flagging on the dev ops build was failing.

PR #376, opened 2019-05-17
[1464] raise error when assigning to unpermitted attributes

misaka:
This changes the behaviour from the default which ignores mass-assigns to
unpermitted attribute, to one that raises an error. For an API it seems
sensible to be more strict about what's coming in and going out this way.

### Context

This really starts to make sense for course updates where we are only allowed to update attributes that are in the enrichments, but even so there are fields such as code that we won't allow updating.

By raising an exceptions here it'll be more obvious in our testing that we haven't enabled it. If this gets noisy in production with false negatives, then we can always revert to just logging there.

Change configuration to raise when an unpermitted parameter is being set. Change tests that broke as a result.

### Guidance to review

Some testing was done locally to ensure that sign-ins, location updates and vacancy updates still worked. So far everything appears to be good.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/sessions_controller.rb
+++ b/app/controllers/api/v2/sessions_controller.rb
@@ -22,7 +22,13 @@ module API
       def create_params
         validate_jsonapi_type(params, 'sessions')
 
-        params.require(:session).permit(:first_name, :last_name)
+        params
+          .require(:session)
+          .except(:id, :type)
+          .permit(
+            :first_name,
+            :last_name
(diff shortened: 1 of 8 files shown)


misaka:
[[M] Course update action in mc-be](https://trello.com/c/1lZpQcIX/1464-m-course-update-action-in-mc-be)

PR #380, opened 2019-05-20
[1501] Subject mapper improvements

benilovj:
### Context

The subject mapper code was hastily ported from C# and there were [code review comments left after the initial port](https://github.com/DFE-Digital/manage-courses-backend/pull/282) that haven't been done yet.

Follow-up to suggestions left on #282. These are pure refactorings. I've moved code out of the service class into model classes.

### Guidance to review
This isn't adding any extra test coverage but we already have almost perfect coverage from the production data spec which is in `subject_mapper_service_spec`. There's more I want to do in this area of the code but I think this change already makes this better.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


config/initializers/inflections.rb
```diff
@@ -19,4 +19,6 @@
   inflect.acronym 'API'
   inflect.acronym 'PGDE'
   inflect.acronym 'UCAS'
+  inflect.acronym 'MFL' # Modern foreign languages
+  inflect.acronym 'DFE' # Department for Educatino
```
dankmitchell:
```suggestion
  inflect.acronym 'DFE' # Department for Education
```

benilovj: Ta

benilovj:
This PR makes CodeClimate happier:

image

app/models/subjects/secondary_english_mapping.rb
```diff
@@ -0,0 +1,28 @@
+# TODO: remove this bonkers logic once course mapping is done by one app!
+# The user need for this is unclear
+#
+# Does the subject list mention english, and it's mentioned in the title (or it's the only subject we know for this course)?
```
defong: is there a test to cover this nuget of knowledge?

app/models/subjects/modern_foreign_languages_other_mapping.rb
```diff
@@ -0,0 +1,39 @@
+# Does the subject list mention languages but hasn't already been covered?
+# TODO: this should be replaced with an explicit static mapping
```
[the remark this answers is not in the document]
benilovj: no, this comment still stands – this whole class needs to go away...

app/models/subjects/secondary_english_mapping.rb
benilovj: Yes, https://github.com/DFE-Digital/manage-courses-backend/blob/master/spec/services/subject_mapper_service_spec.rb#L136

benilovj: It's also covered by the production data test.

benilovj: I've removed the redundant comment in 6944134

PR #382, opened 2019-05-20
[1491] Add deserializable_resource call to AccessRequestsController

tvararu:
This is necessary to correctly interpret the parameters sent in by the support app.

Paired with @davidgisbey.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -1,6 +1,7 @@
 module API
   module V2
     class AccessRequestsController < API::V2::ApplicationController
+      deserializable_resource :access_request, only: %i[create]
       before_action :build_access_request, only: :approve
 
       def approve


tvararu:
[Switch manual access_request creation api calls to new backend endpoint](https://trello.com/c/ksqmnJYW/1491-switch-manual-accessrequest-creation-api-calls-to-new-backend-endpoint)

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -1,6 +1,7 @@
 module API
   module V2
     class AccessRequestsController < API::V2::ApplicationController
+      deserializable_resource :access_request, only: %i[create]
```
dankmitchell:
```suggestion
      deserializable_resource :access_request, only: :create
```

tvararu: Ehhh if we add another one then Rubocop will be like "use `%i[`, it's fine.

dankmitchell: Yep. Ship it

PR #386, opened 2019-05-21
Implement layer caching

smneal:
### Context
Reduce build time by implementing layer caching

Add task to push master builds as latest tags
Add task to pull latest tag for each build

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,7 +71,7 @@ steps:
 
 - task: Docker@1
   displayName: Push tagged image (latest) if master
-  condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/implement_layer_caching'))
+  condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
   inputs:
     command: Push an image
     imageName: "$(docker_path):latest"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,6 +22,9 @@ steps:
     echo "##vso[task.setvariable variable=docker_path;]$docker_path"
   displayName: 'Set version number'
 
+- script: docker pull $(docker_path):latest || true
+  displayName: "Pull latest docker image to cache"
+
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -22,6 +22,9 @@ steps:
     echo "##vso[task.setvariable variable=docker_path;]$docker_path"
   displayName: 'Set version number'
 
+- script: docker pull $(docker_path):latest || true
```
misaka: Why do we need `|| true`? Does `docker pull` return false negatives?

azure-pipelines.yml
```diff
@@ -22,6 +22,9 @@ steps:
...
+  displayName: "Pull latest docker image to cache"
```
misaka: nit-picking, but this is "from cache", right? ;)

smneal: I think we pull it to the cache don't we as there is nothing in there until we populate it? So probably from Docker Hub to cache, and then from cache during the build? Hence cache_from in the az.yml?

azure-pipelines.yml
smneal: It's to ensure the first run does't fail, also ensures any non-zero return codes don't cause the whole build to fail. It just means the build won't end up using layer caching so would take a bit longer.

misaka: I get the mechanics of `|| true`, my point is that with it if there's a failure in pulling the image we won't get an error here and it would be confusing and harder to debug. But I think you're saying that we need it because the first time we run, there may not be a `latest` so there won't be anything to pull and cache from?

azure-pipelines.yml
misaka: Ah, ok. Ya, that makes sense!

azure-pipelines.yml
smneal: Yes that's it.

PR #389, opened 2019-05-21
[1508] add Provider#external_contact_info

misaka:
### Context

The contact info on the provider record is for internal/UCAS use. We need to expose the contact info from the enrichment if it's there.

Expose the contact info from the latest published provider enrichment as "external_contact_info", if exists.

### Guidance to review

Although this method should do more (fall-back to the contact info in the provider for example), this is a temporary measure to just get it in there for course syncing and preview generation. Another PR will be created for the rest of the work.

Also, ignore the wrong card number in the branch name. Doh!

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -100,7 +100,7 @@ class Provider < ApplicationRecord
   end
 
   def external_contact_info
-    enrichments.published.order(created_at: :desc).first
+    enrichments.published.order(last_published_at: :desc).first
       .attributes
       .slice(
         'address1',
@@ -111,6 +111,7 @@ class Provider < ApplicationRecord
         'region_code',
         'telephone',
         'email',
+        'website'
       )
   end
 end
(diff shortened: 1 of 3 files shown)


misaka:
[Expose contacts in provider enrichment as external_contact_info](https://trello.com/c/Dur537nb/1508-expose-contacts-in-provider-enrichment-as-externalcontactinfo)

app/models/provider.rb
```diff
@@ -98,4 +98,19 @@ def unassigned_site_codes
   def can_add_more_sites?
     sites.size < Site::POSSIBLE_CODES.size
   end
+
+  def external_contact_info
+    enrichments.published.order(created_at: :desc).first
+      .attributes
+      .slice(
+        'address1',
```
dankmitchell: Can we add `website`, please?

FYI on `Provider` it's called `url`.

defong: Good call, you wanna reject it?

dankmitchell: Done @defong

misaka: Done. You guys look like circling sharks looking for a reason to reject PRs. :) I'll worry about the fallback to `url` stuff later.

dankmitchell: He made me do it @misaka

PR #395, opened 2019-05-21
[1516] Further subject mapper improvements

benilovj:
### Context
#380 left the code in a better state than before it, but still not good enough to leave for a few weeks. It's valuable to feed the context into the codebase while it's still fresh in the mind.

- extend the mapping domain specific language in `SubjectMapperService` so that the custom mapping classes can be deleted
- extract mapping-processing and course level logic into separate model classes
- make the primary, secondary and further education mappings work in a much more similar way than before
- add model specs for extracted classes
- rename methods and classes to reduce ambiguity

### Guidance to review
It may be easier to follow the simplification/refactoring process if reading this PR commit-by-commit.

The mappings can be changed/simplified once the "sync to Find" is done by this app too. For now, the logic needs to be consistent between the Ruby and the C# implementations.

Also, I've had enough of looking at this code now.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/models/subjects/ucas_subject_to_dfe_subject_mappings.rb
```diff
@@ -0,0 +1,21 @@
+module Subjects
+  class UCASSubjectToDFESubjectMappings
```
misaka: I had to go through and `UCASSubjectToDFESubjectMappings` carefully to see how these are different. :) What about `UCASSubjectToDFESubjectMappingCollection`? Oof, it is a bit long ...

benilovj: Thanks, fixed in 0e4eec2

app/models/subjects/ucas_subject_to_dfe_subject_mapping.rb
```diff
@@ -0,0 +1,40 @@
+module Subjects
+  class UCASSubjectToDFESubjectMapping
+    def initialize(ucas_input_subjects_or_options_hash, resulting_dfe_subject)
```
heidar: I still find this a bit confusing. Given the name of this, I'd expect to be able to pass in a UCAS subject name and have a DfE subject name returned. Could we do that and change how we call this?

app/models/subjects/ucas_subject_to_dfe_subject_mapping_collection.rb
```diff
@@ -0,0 +1,21 @@
+module Subjects
+  class UCASSubjectToDFESubjectMappingCollection
+    def initialize(config:)
```
heidar: It feels like the majority of the work is done via the config. Should the config live in here and we just pass in the level?

heidar: Also, maybe make this take a collection and then iterate over it, using the other class to get the result one by one rather than using a collection again?

app/models/subjects/ucas_subject_to_dfe_subject_mapping.rb
```diff
@@ -0,0 +1,40 @@
+module Subjects
+  class UCASSubjectToDFESubjectMapping
```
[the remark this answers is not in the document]
benilovj: Good idea – done

app/models/subjects/ucas_subject_to_dfe_subject_mapping_collection.rb
benilovj: My reasoning for splitting the data and the functionality was:

- to keep this class easier to unit test and reason about
- have all the specific "settings" defined in the config

kinda like the model we have with the `settings.yml` files.

I can combine them if you think it's absolutely needed.

app/models/subjects/ucas_subject_to_dfe_subject_mapping.rb
benilovj: Sure – I will see how I can rework this to make it clearer.

benilovj: Having just spoken in IRL, I want to build on top of this PR so I'll merge this and address the feedback in a separate PR, since this seems like an improvement over what we had previously.

PR #396, opened 2019-05-21
[1503] Add website to provider

dankmitchell:
### Context
All of our provider enrichment attributes match our provider attributes names so falling back when data is missing is straight forward except for `url` / `website`

- Add provider `website` and copy all data from `url`
- Once our c# app is decommisioned then the `url` column can  be removed

### Guidance to review
Note: The V1 API doesn’t use it so this shouldn't have any upstream issues

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -24,6 +24,7 @@
 #  accrediting_provider :text
 #  last_published_at    :datetime
 #  changed_at           :datetime         not null
+#  website              :text
 #
 
 class Provider < ApplicationRecord
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -24,6 +24,7 @@
 #  accrediting_provider :text
 #  last_published_at    :datetime
 #  changed_at           :datetime         not null
+#  website              :text
 #
 
 class ProviderSerializer < ActiveModel::Serializer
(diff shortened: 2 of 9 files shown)


dankmitchell:
[Expose provider enrichment data](https://trello.com/c/xHIRy6de/1503-expose-provider-enrichment-data)

db/migrate/20190521214615_rename_url_to_website_on_provider.rb
```diff
@@ -0,0 +1,5 @@
+class RenameUrlToWebsiteOnProvider < ActiveRecord::Migration[5.2]
+  def change
+    rename_column :provider, :url, :website
```
misaka: Isn't there a risk that with a change like this, the C# app might have a fit?

dankmitchell: Yes :(

dankmitchell: I have another idea... watch this space (tomorrow morning)

dankmitchell: Solved

misaka:
This feels a little messy ... can we hold off on doing this a little? If the `&.attributes` approach works for now, I'd like to have a think about how we should fall-back, and also what is our model going to be for the editable data on a course. I think we had an assumption that this was all going to be moved to the course enrichment, but I'm not so sure that that's the case now after speaking with Paul H.

dankmitchell:
@misaka it's messy because I don't want to touch the c# app's. This change gets us moving on the rails app and `url` can be dropped once we have fully migrated away from manage courses ui...

dankmitchell:
Closing in favour of https://github.com/DFE-Digital/manage-courses-backend/pull/398

PR #402, opened 2019-05-22
[1367] Add show endpoint to access requests

tvararu:
### Context

This is needed by the support app.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -10,6 +10,13 @@ module API
         render status: 200, json: { result: result }
       end
 
+      def show
+        authorize AccessRequest
+        @access_request = AccessRequest.find(params[:id])
+
+        render jsonapi: @access_request, include: [:requester]
+      end
+
       def index
         authorize AccessRequest
         @access_requests = AccessRequest.requested.includes(:requester)
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -19,7 +19,7 @@ module API
 
       def index
         authorize AccessRequest
-        @access_requests = AccessRequest.requested.includes(:requester)
+        @access_requests = AccessRequest.requested.includes(:requester).by_request_date
 
         render jsonapi: @access_requests, include: params[:include]
       end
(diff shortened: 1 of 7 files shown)


tvararu:
[Change mc-support to list access requests from mc-be](https://trello.com/c/xgZjvefX/1367-change-mc-support-to-list-access-requests-from-mc-be)

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -187,6 +302,14 @@
   end
 
   describe 'POST #create' do
+    before do
```
heidar: Can't we just wrap the request itself in Timecop.do .. end?

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -10,11 +10,18 @@ def approve
...
+        authorize AccessRequest
+        @access_request = AccessRequest.find(params[:id])
+
+        render jsonapi: @access_request, include: [:requester]
+      end
+
       def index
         authorize AccessRequest
-        @access_requests = AccessRequest.requested
+        @access_requests = AccessRequest.requested.includes(:requester).order(request_date_utc: :asc)
```
benilovj: 🎉 the ordering will hopefully fix the flaky test

spec/requests/api/v2/access_request_spec.rb
tvararu: That's what we did before, we changed to this because we thought maybe it's flaking on VSTS but it's probably not. Maybe we shouldn't bother now.

app/controllers/api/v2/access_requests_controller.rb
dankmitchell: Bonus points for a scope i.e. `by_request_date` rather than ordering in the controller

tvararu:
[Change mc-support to list access requests from mc-be](https://trello.com/c/xgZjvefX/1367-change-mc-support-to-list-access-requests-from-mc-be)

spec/requests/api/v2/access_request_spec.rb
tvararu: Nevermind, removed this

PR #404, opened 2019-05-23
[1520] Financial incentives

benilovj:
### Context
We need financial incentive information for rendering the course details page preview. Currently this information resides only on Find.

- extract a `DFESubject` plain-old-Ruby-object model
- define financial incentives data as a Ruby structure on `DFESubject` (this would only get updated at most once a year so should be fine being hard-coded)
- expose `has_bursary?` and `has_scholarship_and_bursary?` on `Course` through the Course API V2

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -396,6 +396,19 @@ RSpec.describe Course, type: :model do
         its(:is_send?) { should be_truthy }
       end
     end
+
+    describe "bursaries and scholarships" do
+      let(:subjects) {
+        [
+          build(:subject, subject_name: 'mathematics'),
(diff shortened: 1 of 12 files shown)


spec/models/course_spec.rb
```diff
@@ -396,6 +396,19 @@
         its(:is_send?) { should be_truthy }
       end
     end
+
+    describe "bursaries and scholarships" do
+      let(:subjects) {
+        [
+          create(:subject, subject_name: 'mathematics'),
```
heidar: Can these just be `build`?

benilovj: yup, done

PR #405, opened 2019-05-23
[1527] Training locations without vacancies should be suspended

benilovj:
### Context
If a training location has no vacancies, it's still selectable on UCAS Apply
but the candidate is informed several steps into their application journey
that it's full.

This change makes the training location suspended in the V1 API. The end effect
of this change is that the location is no longer listed on UCAS Apply, so candidates
don't inadvertently select them anymore.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/serializers/site_status_serializer.rb
+++ b/app/serializers/site_status_serializer.rb
@@ -23,7 +23,11 @@ class SiteStatusSerializer < ActiveModel::Serializer
   end
 
   def status
-    object.status_before_type_cast
+    if object.no_vacancies?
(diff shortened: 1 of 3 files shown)


fofr:
When an applicant has selected a training location, then before they submit their application the location is suspended – what do they see? what happens?

fofr:
> When the applicant goes to submit the application it will inform them that the course is no longer available and they will be able to change the course to a new code/location at the same provider.

PR #407, opened 2019-05-23
[1528] Fix manual access requests

tvararu:
### Context

Manual access requests give users God access.

Read the `requester_email` from the params, and pass it into the update. Do not use the super admin. Oops.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -25,11 +25,13 @@ module API
       end
 
       def create
-        authorize AccessRequest # todo, generalise admin auth
         access_request = AccessRequest.new(access_request_params)
-        access_request.update(requester: @current_user,
-                              request_date_utc: Time.now.utc,
-                              status: :requested)
+        authorize access_request
+
+        access_request.requester        = User.find_by(email: access_request.requester_email)
+        access_request.request_date_utc = Time.now.utc
+        access_request.status           = :requested
+        access_request.save!
 
         render jsonapi: access_request
       end
@@ -41,7 +43,14 @@ module API
       end
 
       def access_request_params
-        params.require(:access_request).permit(:email_address, :first_name, :last_name, :organisation, :reason)
+        params.require(:access_request).permit(
+          :email_address,
+          :first_name,
+          :last_name,
+          :organisation,
+          :reason,
+          :requester_email
+        )
       end
     end
   end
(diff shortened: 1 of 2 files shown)


tvararu:
[Manual access requests on support app gives God-user](https://trello.com/c/m7ZOKMo2/1528-manual-access-requests-on-support-app-gives-god-user)

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -25,11 +25,13 @@ def index
...
         access_request = AccessRequest.new(access_request_params)
-        access_request.update(requester: @current_user,
-                              request_date_utc: Time.now.utc,
-                              status: :requested)
+        authorize access_request
+
+        access_request.requester        = User.find_by(email: access_request.requester_email)
+        access_request.request_date_utc = Time.now.utc
+        access_request.status           = :requested
+        access_request.save!
```
defong: Does it meant to do the bang?

heidar: yeah. it should blow up since we dont handle validation errors

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -25,11 +25,13 @@ def index
...
 
       def create
-        authorize AccessRequest # todo, generalise admin auth
         access_request = AccessRequest.new(access_request_params)
-        access_request.update(requester: @current_user,
-                              request_date_utc: Time.now.utc,
-                              status: :requested)
+        authorize access_request
+
+        access_request.requester        = User.find_by(email: access_request.requester_email)
```
heidar: This code is beautiful, incredible work friend!

misaka: Should this be happening in a `before_create` hook? Feels funny having this in the model ... if we ever create an access request via mcb, for example, it wouldn't do the right thing.

heidar: Yeah, this code isn't ideal in many ways, but this is the smallest change to fix the bug. I think anything else should be addressed separately.

tvararu: Yeah, let's fix the actual issue so that our support team can go back to using this first... @misaka ok with it?

misaka: @tvararu @heidar Now that it's fixed and shipping ... can we raise a PR to fix it "properly"? Assuming this wouldn't be part of a bigger feature/story, I think we'd rather avoid "tech debt" stories if they're not clearly attached to something else.

PR #413, opened 2019-05-24
[1506] Grant user access to provider

benilovj:
### Context
We need to be able to add new users to organisations. It's possible to use access requests for this, but only as long as there is an existing user in the organisation.

- Add an `mcb` command which adds new users to an organisation, or gives existing users access to that organisation.
- Add some validations on `User` to help catch any mistakes while using `mcb`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,7 +27,8 @@ class User < ApplicationRecord
   scope :non_admins, -> { where.not('email ~ ?', DFE_EMAIL_PATTERN) }
   scope :active, -> { where.not(accept_terms_date_utc: nil) }
 
-  validates :email, presence: true
+  validates :email, presence: true, format: { with: /@/, message: "must contain @" }
+  validate :email_is_lowercase
 
   audited
 
@@ -46,5 +47,13 @@ class User < ApplicationRecord
 
   def to_s
     "#{first_name} #{last_name} <#{email}>"
+  end
+
+private
+
+  def email_is_lowercase
+    if email.present? && email.downcase != email
+      errors.add(:email, "must be lowercase")
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


app/models/user.rb
```diff
@@ -27,7 +27,8 @@ class User < ApplicationRecord
   scope :non_admins, -> { where.not('email ~ ?', DFE_EMAIL_PATTERN) }
   scope :active, -> { where.not(accept_terms_date_utc: nil) }
 
-  validates :email, presence: true
+  validates :email, presence: true, format: { with: /@/, message: "must contain @" }
```
defong: is that too wide

image

benilovj: Given this is going to be used by internal users at the moment, I don't think we need to be that strict.

defong: :cool:

app/models/user.rb
```diff
@@ -27,7 +27,8 @@ class User < ApplicationRecord
...
+  validate :email_is_lowercase
```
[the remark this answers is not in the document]
benilovj: image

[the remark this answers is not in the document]
defong: didn't know sql can be like that :nerd_face:

spec/models/user_spec.rb
```diff
@@ -18,14 +18,18 @@
...
 describe User, type: :model do
-  subject { create(:user) }
+  subject { create(:user, first_name: 'Jane', last_name: 'Smith', email: 'jsmith@scitt.org') }
 
   describe 'associations' do
     it { should have_and_belong_to_many(:organisations) }
     it { should have_many(:providers).through(:organisations) }
   end
 
   it { is_expected.to validate_presence_of(:email) }
```
defong: with_message("must contain @")

https://github.com/thoughtbot/shoulda-matchers/blob/ab69fab50033adf7d4f38ca02c9dba429e8bf569/lib/shoulda/matchers/active_model/validate_presence_of_matcher.rb#L103

benilovj: Done

PR #417, opened 2019-05-24
[1483] Add early_career_payments to course endpoint

dankmitchell:
### Context
Extend course financial support

- Add `has_early_career_payments`
- Add `bursary_amount`
- Add `scholarship_amount`

### Guidance to review
Example -`/api/v2/providers/1CS/courses/2DH4`

### Checklist
- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/dfe_subject.rb
+++ b/app/models/dfe_subject.rb
@@ -25,6 +25,10 @@ class DFESubject
     has_bursary? && has_scholarship?
   end
 
+  def has_early_career_payments?
+    financial_support&.fetch(:early_career_payments).present?
+  end
+
   def to_s
     @subject_name
   end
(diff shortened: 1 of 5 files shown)


app/models/dfe_subject.rb
```diff
@@ -25,6 +25,10 @@ def has_scholarship_and_bursary?
     has_bursary? && has_scholarship?
   end
 
+  def has_early_career_payments?
```
benilovj: it's worth adding tests for this in `dfe_subject_spec.rb`

dankmitchell: Done

PR #422, opened 2019-05-29
[1540] include site_status sites when generating api v1 courses

misaka:
### What

Change the query used to generate the results for `/api/v1/2019/courses` to include the `sites` that belong to the `site_statuses`.

### Why

This speeds up the generation of these results by about 33%.

### Background

This is an N+1 query that can be fixed by simply including the data being used when making the original SQL request.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


misaka:
[Make /api/v1/2019/courses load faster](https://trello.com/c/xpj2enuT/1540-make-api-v1-2019-courses-load-faster)

PR #424, opened 2019-05-29
[1542] add mcb apiv1 course find command

misaka:
###Why are we doing this?

To allow admins to view what a course looks like as it's being sent to UCAS.

###What are we adding or changing?

A command to`mcb` to find a course in the output generated by the api v1.

###Background

Useful for investigating issues with the system.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -99,99 +99,13 @@ module MCB
   end
 
   def self.each_v1_course(opts)
-    # We only need httparty for API V1 calls
-    require 'httparty'
-
-    url = URI.join(opts[:url], opts[:endpoint])
-
-    token = opts.fetch(:token) { apiv1_token(opts.slice(:webapp, :rgroup)) }
-
-    process_opt_changed_since(opts, url)
-    page_count = 0
-    max_pages = opts.fetch(:'max-pages', '30').to_i
-    all_pages = opts.fetch(:all, false)
-
-    Enumerator.new do |y|
-      loop do
-        if page_count > max_pages
-          raise "too many page requests, stopping at #{page_count}" \
-                " as a safeguard. Use --max-pages to increase max page count" \
-                " if necessary."
-        end
-
-        verbose "Requesting page #{page_count + 1}: #{url}"
-        response = HTTParty.get(
-          url.to_s,
-          headers: { authorization: "Bearer #{token}" }
-        )
-        courses_list = JSON.parse(response.body)
-        break if courses_list.empty?
-
-        next_url = response.headers[:link].sub(/;.*/, '')
-
-        # Send each provider to the consumer of this enumerator
-        courses_list.each do |course|
-          y << [course, {
-            page: page_count,
-                  url: url,
-                  next_url: next_url
-          }]
-        end
-
-        break unless all_pages
-
-        url = next_url
-        page_count += 1
-      end
-    end
+    # This methods can actually be entirely driven by the args provided it. auto-configure!
+    iterate_v1_endpoint(**opts)
   end
 
   def self.each_v1_provider(opts)
-    # We only need httparty for API V1 calls
-    require 'httparty'
-
-    url = URI.join(opts[:url], opts[:endpoint])
-
-    token = opts.fetch(:token) { apiv1_token(opts.slice(:webapp, :rgroup)) }
-
-    process_opt_changed_since(opts, url)
-    page_count = 0
-    max_pages = opts.fetch(:'max-pages', '30').to_i
-    all_pages = opts.fetch(:all, false)
-
-    Enumerator.new do |y|
-      loop do
-        if page_count > max_pages
-          raise "too many page requests, stopping at #{page_count}" \
-                " as a safeguard. Use --max-pages to increase max page count" \
-                " if necessary."
-        end
-
-        verbose "Requesting page #{page_count + 1}: #{url}"
-        response = HTTParty.get(
-          url.to_s,
-          headers: { authorization: "Bearer #{token}" }
-        )
-        providers_list = JSON.parse(response.body)
-        break if providers_list.empty?
-
-        next_url = response.headers[:link].sub(/;.*/, '')
-
-        # Send each provider to the consumer of this enumerator
-        providers_list.each do |provider|
-          y << [provider, {
-            page: page_count,
-                  url: url,
-                  next_url: next_url
-          }]
-        end
-
-        break unless all_pages
-
-        url = next_url
-        page_count += 1
-      end
-    end
+    # This methods can actually be entirely driven by the args provided it. auto-configure!
+    iterate_v1_endpoint(**opts)
   end
 
   def self.config_dir=(dir)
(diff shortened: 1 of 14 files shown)


misaka:
[Allow mcb to find a course using api v1](https://trello.com/c/9ZGeC80F/1542-allow-mcb-to-find-a-course-using-api-v1)

lib/mcb/commands/apiv1.rb
```diff
@@ -7,3 +7,8 @@
 option :t, 'token', 'set the authorization token',
        argument: :required,
        default: 'bats'
+option :P, 'max-pages', 'maximum number of pages to request',
+       default: 200,
```
defong: sensible value?

lib/mcb/commands/apiv1/courses/find.rb
```diff
@@ -0,0 +1,62 @@
+summary 'Find a particular provider course entry'
+description <<~EODESCRIPTION
+  Searches for a course with the given provider and course code by iterating
+  through all the pages of results provided from the course endpoint, outputing
```
defong:
```suggestion
  through all the pages of results provided from the course endpoint, outputting
```

typo `outputing`

lib/mcb/commands/apiv1/courses/find.rb
```diff
@@ -0,0 +1,62 @@
...
+  the found record.
```
defong:
```suggestion
  the found records.
```

grammar?

lib/mcb/commands/apiv1.rb
misaka: I know, seems large right? But we have 17k+ courses and we get 100 courses back per page, so ya as it turns out this is sensible.

lib/mcb/commands/apiv1/courses/find.rb
misaka: Maybe search-and-replace gone bad? Anyway, thanks!

lib/mcb.rb
```diff
@@ -99,97 +99,13 @@ def self.generate_apiv2_token(email:, encoding:, secret: nil)
...
-          }]
-        end
-
-        break unless all_pages
-
-        url = next_url
-        page_count += 1
-      end
-    end
+    # This methods can actually be entirely driven by the args provided it. auto-configure!
```
defong:
```suggestion
    # This method can actually be entirely driven by the args provided, it is auto-configure!
```

methods?

it. auto-configure!

lib/mcb.rb
```diff
@@ -99,97 +99,13 @@ def self.generate_apiv2_token(email:, encoding:, secret: nil)
...
```
defong:
```suggestion
    # This method can actually be entirely driven by the args provided, it is auto-configure!
```

methods?

it. auto-configure!

lib/mcb.rb
```diff
@@ -234,6 +150,56 @@ def get_user_from_config
...
+
+      if requesting_remote_connection?(**opts)
+        opts[:url] = MCB::Azure.get_urls(**opts).first
+        opts[:token] = MCB::Azure.get_config(**opts)['AUTHENTICATION_TOKEN']
+      end
+
+      opts
+    end
+
+    def iterate_v1_endpoint(url:, endpoint:, **opts)
```
defong: just wondering can this be moved to it own, as this file is a bit big

this mcb file seems to forever dealing with doing prerequisites like, fetching azure/config/cli args then the actionable bits

misaka: Sounds reasonable to me. Will to.

misaka: I just had a look and I think that this refactoring should be a larger bit of work to move out all the API V1 related methods into a separate module. Good call, but I think it's worth putting into it's own PR.

lib/mcb/commands/apiv1/courses/find.rb
```diff
@@ -0,0 +1,62 @@
...
+  through all the pages of results provided from the course endpoint, outputting
+  the found records.
+EODESCRIPTION
+usage 'find [options] <provider_code> <course_code>'
+param :provider_code
+param :course_code
+option :j, 'json', 'show the returned JSON response'
+
+
+run do |opts, args, _cmd|
```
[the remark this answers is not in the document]
misaka: An underscore prefix denotes an argument that is not being used. rubocop, for example, says this when it's missing:

```
W: Lint/UselessAssignment: Useless assignment to variable - cmd. Use _ or _cmd as a variable name to indicate that it won't be used. (https://github.com/rubocop-hq/ruby-style-guide#underscore-unused-vars)
```

spec/serializers/api/v2/serializable_access_request_spec.rb
```diff
@@ -46,7 +46,11 @@
           .and(have_id(requester.id.to_s))))
     end
   end
+
   describe 'has the correct attributes' do
+    before { Timecop.freeze }
```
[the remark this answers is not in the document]
misaka: Ya, as we said in the stand-up when it's something simple we'd just fix it on the fly.

lib/mcb/commands/apiv1/courses/find.rb
```diff
@@ -0,0 +1,62 @@
...
+  if opts[:all]
+    puts 'All pages searched.'
+  else
+    puts 'Only first page of results searched (use -a to retrieve all).'
+  end
+  puts "To continue searching use the url: #{last_context[:next_url]}"
+end
+
+def find_course(provider_code, course_code, opts)
+  MCB.each_v1_course(opts).detect do |course, _context|
```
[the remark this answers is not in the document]
misaka: See above.

lib/mcb/commands/apiv1/courses/find.rb
```diff
@@ -0,0 +1,62 @@
...
+    course['provider']['institution_code'] == provider_code &&
+      course['course_code'] == course_code
```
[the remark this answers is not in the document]
misaka: The V1 API returns JSON that we process as hashes ... it's not JSONAPI and we don't try to turn them into objects, so we can't use dot-notation.

lib/mcb/commands/apiv1/courses/find.rb
```diff
@@ -0,0 +1,62 @@
...
+    puts 'All pages searched.'
+  else
+    puts 'Only first page of results searched (use -a to retrieve all).'
+  end
+  puts "To continue searching use the url: #{last_context[:next_url]}"
+end
+
+def find_course(provider_code, course_code, opts)
+  MCB.each_v1_course(opts).detect do |course, _context|
+    course['provider']['institution_code'] == provider_code &&
```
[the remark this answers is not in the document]
misaka: The V1 API returns JSON that we process as hashes ... it's not JSONAPI and we don't try to turn them into objects, so we can't use dot-notation.

lib/mcb/commands/apiv1/courses/find.rb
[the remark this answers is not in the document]
misaka: An underscore prefix denotes an argument that is not being used. rubocop, for example, says this when it's missing:

```
W: Lint/UselessAssignment: Useless assignment to variable - context. Use _ or _context
```

lib/mcb/commands/apiv1/courses/find.rb
```diff
@@ -0,0 +1,62 @@
+summary 'Find a particular provider course entry'
+description <<~EODESCRIPTION
+  Searches for a course with the given provider and course code by iterating
+  through all the pages of results provided from the course endpoint, outputting
+  the found records.
```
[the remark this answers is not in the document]
misaka: https://github.com/DFE-Digital/manage-courses-backend/pull/424#discussion_r289295640

Make up your mind. :) It should be singular, I guess I just trusted you when you made the original suggestion.

PR #430, opened 2019-05-31
[1520] no scholarship or bursary for courses with another subject

dankmitchell:
### Context
Funding overrides

Re-implementing what our c# app does

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


fofr:
Do you have a link to the corresponding C# code?

tvararu:
@fofr https://github.com/DFE-Digital/search-and-compare-ui/blob/4e8cd271631e186361a6eb3db3b92eeb5f52dd4f/shared/ViewFormatters/CourseExtensions.cs#L109-L178

tvararu:
[Financial support info in course API v2](https://trello.com/c/9oQ4VBt4/1520-financial-support-info-in-course-api-v2)

app/models/course.rb
```diff
@@ -263,10 +263,22 @@ def sites_not_associated_with_course
   end
 
   def has_bursary?
+    # If course name contains "with"
+    return false if name.include?("with")
```
fofr: Primary with mathematics has a bursary though:
https://getintoteaching.education.gov.uk/funding-my-teacher-training/bursaries-and-scholarships-for-teacher-training

fofr: Live example showing bursary: https://find-postgraduate-teacher-training.education.gov.uk/course/2AT/2T89#section-financial-support

app/models/course.rb
```diff
@@ -263,10 +263,40 @@ def sites_not_associated_with_course
   end
 
   def has_bursary?
+    # https://bat-design-history.herokuapp.com/find-teacher-training/financial-incentives-multiple-subjects
+    # If course name contains "with"
+    return true if name.include?("with")
```
benilovj: I'm not convinced that this logic is right – I don't think the unit tests from the C# side will pass (see https://github.com/DFE-Digital/search-and-compare-ui/pull/335). In any case we should port those over.

app/models/course.rb
```diff
@@ -263,10 +263,40 @@ def sites_not_associated_with_course
   end
 
   def has_bursary?
```
heidar: Can we not add these as database fields and populate them based on this (or whatever the correct logic is)?

PR #432, opened 2019-05-31
What is the use case?

defong:
### Context
What is the use case?

```ruby
def sync_with_search_and_compare
```
almost the same as

```ruby
def publish
```

Eviction from codebase

### Guidance to review
I got tripped by this not so long ago

https://github.com/DFE-Digital/manage-courses-backend/pull/355/commits/f3cdbf4cf8b8e4f33a4cb2c7a0e5194478dd4443

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/controllers/api/v2/courses_controller.rb
```diff
@@ -19,16 +19,6 @@ def show
...
-      def sync_with_search_and_compare
-        response = ManageCoursesAPIService::Request.sync_course_with_search_and_compare(
-          @current_user.email,
-          @provider.provider_code,
-          @course.course_code
-        )
-
-        head response ? :ok : :internal_server_error
-      end
-
```
misaka: How about instead of killing it, and as a result whatever is calling it, we just call `sync_with_search_and_compare` from `publish` below?

misaka:
We aren't doing this. We might re-factor to make it clearer.

PR #433, opened 2019-05-31
[1517] Add/remove/suspend course sites

timabell:
### Context

Pairing @davidgisbey & @timabell & @tvararu 

* Extend the `course#update` endpoint to support supplying a list of sites.
* Remove some dead code that was in the way after discussing with @benilovj 
* Hide more of the exposed `site_status` inside `sites` in the API surface.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -56,6 +56,9 @@ module API
         enrichment.assign_attributes(update_params)
         enrichment.save
 
+        site_ids = params[:course][:sites_ids]
+        @course.sites = @provider.sites.where(id: site_ids) if site_ids.present?
+
         if @course.valid?
           render jsonapi: @course.reload
         else
@@ -78,7 +81,7 @@ module API
       def update_params
         params
           .require(:course)
-          .except(:id, :type)
+          .except(:id, :type, :sites_ids, :sites_types)
           .permit(
             :about_course,
             :course_length,
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -260,10 +260,6 @@ class Course < ApplicationRecord
     new? ? site_status.destroy! : site_status.suspend!
   end
 
-  def sites_not_associated_with_course
-    provider.sites - sites
-  end
-
   def has_bursary?
     dfe_subjects.any?(&:has_bursary?)
   end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -251,13 +251,26 @@ class Course < ApplicationRecord
 
   def add_site!(site:)
     is_course_new = new? # persist this before we change anything
-    site_status = site_statuses.find_or_create_by!(site: site)
+    site_status = site_statuses.find_or_initialize_by(site: site)
     site_status.start! unless is_course_new
+    site_status.save! if persisted?
   end
 
   def remove_site!(site:)
     site_status = site_statuses.find_by!(site: site)
     new? ? site_status.destroy! : site_status.suspend!
+  end
+
+  def sites=(desired_sites)
+    existing_sites = sites
+
+    to_add = desired_sites - existing_sites
+    to_add.each { |site| add_site!(site: site) }
+
+    to_remove = existing_sites - desired_sites
+    to_remove.each { |site| remove_site!(site: site) }
+
+    sites.reload
   end
 
   def has_bursary?
--- a/spec/requests/api/v2/providers/courses/update_sites_spec.rb
+++ b/spec/requests/api/v2/providers/courses/update_sites_spec.rb
@@ -1,0 +1,85 @@
+require "rails_helper"
+
+describe 'PATCH /providers/:provider_code/courses/:course_code with sites' do
+  let(:organisation) { create :organisation }
+  let(:provider)     { create :provider, organisations: [organisation] }
+  let(:user)         { create :user, organisations: [organisation] }
+  let(:payload)      { { email: user.email } }
+  let(:token)        { build_jwt :apiv2, payload: payload }
+  let(:credentials) do
+    ActionController::HttpAuthentication::Token.encode_credentials(token)
+  end
+
+  let(:course) { create :course, provider: provider }
+  let(:site_to_add) { create :site, provider: provider }
+  let(:unwanted_site) { create :site, provider: provider }
+  let(:existing_site) { create :site, provider: provider }
+
+  before do
+    course.add_site!(site: existing_site)
+    course.add_site!(site: unwanted_site)
+  end
+
+  let(:jsonapi_data) do
+    {
+      "_jsonapi" => {
+        "data" => {
+          "id" => course.id.to_s,
+          "type" => "courses",
+          "relationships" => {
+            "sites" => {
+              "data" => [
+                { "type" => "sites", "id" => existing_site.id.to_s },
+                { "type" => "sites", "id" => site_to_add.id.to_s }
+              ]
+            }
+          }
+        }
+      }
+    }
+  end
+
+  def perform_request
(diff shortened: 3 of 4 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,24 @@ def update
         enrichment.assign_attributes(update_params)
         enrichment.save
 
+        # todo: something more canonical
+        # todo: validate sites actually belong to this org
+        desired_site_ids = params["_jsonapi"]["data"]["relationships"]["sites"]["data"].map{|s| s["id"].to_i}
```
defong: think you missing something to do  `deserializable_resource`

app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,24 @@ def update
...
+
+        existing_site_ids = @course.sites.map{|s| s["id"]}
+
+        to_add = desired_site_ids - existing_site_ids
+        to_add.each do |id|
```
defong: this is basically 
select one drop one looped....
instead of `.find(id)` `.find(to_add)`?

app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,24 @@ def update
...
+          site = @provider.sites.find(id)
+          @course.add_site!(site: site)
```
defong: then you can rework this one

app/controllers/api/v2/courses_controller.rb
defong: use `hash.dig`?

app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,27 @@ def update
         enrichment.assign_attributes(update_params)
         enrichment.save
 
+        # todo: something more canonical
+        # todo: validate sites actually belong to this org
+        sites = params["_jsonapi"]["data"]["relationships"]["sites"]
```
misaka: Are we not able to use the deserializer here? As in, use `params['sites']`?

app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,27 @@ def update
...
+          to_add.each do |id|
+            site = @provider.sites.find(id)
+            @course.add_site!(site: site)
+          end
+
+          to_remove = existing_site_ids - desired_site_ids
+          to_remove.each do |id|
+            site = @provider.sites.find(id)
+            @course.remove_site!(site: site)
+          end
```
misaka: I haven't tested this, but I can't help but feel like we should be able to do:

```
sites = Site.find sites["data"].map &:to_i
@course.sites = sites
```

app/controllers/api/v2/courses_controller.rb
tvararu: Sorta; how about `params[:course][:sites_ids]`?

spec/requests/api/v2/providers/courses/update_sites_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      expect(course.reload.sites.exists?(unwanted_site.id)).to be(false)
+    end
+
+    context "course is running" do
+      before do
+        course.publish_sites
+      end
+
+      it "suspends an unwanted site" do
+        expect(course.reload.sites.exists?(unwanted_site.id)).to be(false)
```
timabell: On second thoughts @tvararu perhaps we do want to dive into `site_statuses` here and check it's suspended? Now that everything's working I'm not sure I'm happy with the two "unwanted site" tests here, they're a bit surprising/magical.

timabell:
[Course basic details - Edit locations mc-be](https://trello.com/c/3zQZm3zS/1517-course-basic-details-edit-locations-mc-be)

timabell:
@tvararu do the more extensive changes to the API surface make this a breaking change? How are we going to ship this & the UI to not have downtime?

benilovj:
> @tvararu do the more extensive changes to the API surface make this a breaking change? How are we going to ship this & the UI to not have downtime?

Good point @timabell, I'd suggest that adding `sites` is done on this PR and removing `site_statuses` is done on a separate PR once the `frontend` has been switched from `site_statuses` to `sites`.

tvararu:
@benilovj split out ✅

app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,14 @@ def update
         enrichment.assign_attributes(update_params)
         enrichment.save
 
+        site_ids = params[:course][:sites_ids]
+        if site_ids.present?
+          new_sites = site_ids
+            .map(&:to_i)
+            .map { |site_id| @provider.sites.find { |site| site.id == site_id } }
```
```suggestion
            .map { |site_id| @provider.sites.find { |site| site.id == site_id.to_i } }
            
            
```

Thro should ` site_ids = params[:course][:sites_ids]` be a list of ints already?

app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,14 @@ def update
         enrichment.assign_attributes(update_params)
         enrichment.save
 
+        site_ids = params[:course][:sites_ids]
+        if site_ids.present?
+          new_sites = site_ids
+            .map(&:to_i)
```
defong: drop that and just map once...

app/controllers/api/v2/courses_controller.rb
misaka: @defong They wouldn't be, these are params taken verbatim from the query string, I'm pretty sure they don't get converted.

Also, I prefer converting them as a separate step, as done in the pr, so that there's less to reason about on this line.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -56,6 +56,14 @@ def update
         enrichment.assign_attributes(update_params)
         enrichment.save
 
+        site_ids = params[:course][:sites_ids]
+        if site_ids.present?
+          new_sites = site_ids
```
misaka: Would this work too?

```suggestion
          new_sites = @provider.sites.where(id: site_ids.map(&:to_i))
```

misaka: If that does, you could to the `to_i` conversion on line 59, which lends itself to @defong's suggestion too.

app/models/course.rb
```diff
@@ -258,8 +258,16 @@ def remove_site!(site:)
...
+  def sites=(desired_sites)
+    existing_sites = sites
+
+    to_add = desired_sites - existing_sites
+    to_add.each { |site| add_site!(site: site) }
+
+    to_remove = existing_sites - desired_sites
+    to_remove.each { |site| remove_site!(site: site) }
+
+    sites
```
misaka: I don't see any tests for this method in the model spec. Can you add some? For example, I'm wondering whether `sites` needs to be reloaded since the methods `add_site!` and `remove_site!` act on the `site_status` objects directly.

spec/requests/api/v2/providers/courses/update_sites_spec.rb
tvararu: Thanks, that uncovered a bug, fixed

app/controllers/api/v2/courses_controller.rb
tvararu: FWIW seems to work fine without the `.to_i` conversion.

app/models/course.rb
tvararu: Is `sites.reload` a thing?..

misaka: I think so, ya.

PR #439, opened 2019-06-04
[1517] Send down sites courses serializer

tvararu:
### Context

These are needed for the locations feature in the frontend. Also, filter only the running and new sites.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -70,7 +70,9 @@ class Course < ApplicationRecord
   has_many :course_subjects
   has_many :subjects, through: :course_subjects
   has_many :site_statuses
-  has_many :sites, through: :site_statuses
+  has_many :sites,
+           -> { merge(SiteStatus.where(status: %i[new_status running])) },
+           through: :site_statuses
 
   has_many :enrichments,
            ->(course) { where(provider_code: course.provider.provider_code) },
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -68,6 +68,19 @@ RSpec.describe Course, type: :model do
     its(:has_vacancies?) { should be false }
   end
 
+  describe "#sites" do
+    let(:first_site) { create(:site) }
+    let(:first_site_status) { create(:site_status, :running, site: first_site) }
+    let(:second_site) { create(:site) }
+    let(:second_site_status) { create(:site_status, :suspended, site: second_site) }
+
+    subject { create(:course, site_statuses: [first_site_status, second_site_status]) }
+
+    it "should only return new and running sites" do
+      expect(subject.sites.to_a).to eq([first_site])
+    end
+  end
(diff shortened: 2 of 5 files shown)


tvararu:
[Course basic details - Edit locations mc-be](https://trello.com/c/3zQZm3zS/1517-course-basic-details-edit-locations-mc-be)

fofr:
> Also, filter only the running and new sites.
Why?

tvararu:
@fofr those are used in the course details tab on frontend, currently the only place where this is used

Screenshot 2019-06-04 at 15 12 49

benilovj:
@fofr the suspended and discontinued sites never need to be shown by the frontend

fofr:
@benilovj @tvararu My specific worry was that on the edit locations screen we'll need to show all sites.

tvararu:
@fofr those are taken from the provider, which does have the full list 😉

app/models/course.rb
```diff
@@ -70,7 +70,9 @@ class Course < ApplicationRecord
   has_many :course_subjects
   has_many :subjects, through: :course_subjects
   has_many :site_statuses
-  has_many :sites, through: :site_statuses
+  has_many :sites,
+           -> { merge(SiteStatus.where(status: %i[new_status running])) },
+           through: :site_statuses
```
misaka: Do we test anywhere that this is working as expected?

tvararu: Not yet 👍

misaka:
@fofr @tvararu I think it's a good question ... if the API is a platform, then should it be filtering automatically for us. I guess the question is what is the data model we want to move to. Unfortunately, we don't quite know for sure yet, but our best guess is, so far, that sites will have a binary association with a course ... it's either associated or it isn't, and the rest is internal details. I'm good with this if that's the reasoning behind it.

tvararu:
Turns out we do need to send down the site_statuses still because otherwise the vacancies won't work.

misaka:
Ya, that's the other side of the coin ... we're often not quite in a position to make data model changes yet. :(

app/models/course.rb
tvararu: binding.pry

spec/models/course_spec.rb
```diff
@@ -68,6 +68,19 @@
     its(:has_vacancies?) { should be false }
   end
 
+  context "with sites" do
```
misaka: Sorry if this comes off as picky, but to continue the crusade to get our specs to actually _sound_ right 😛 ... we should have a describe here, for example `describe '#sites'` or `describe 'sites association'`. And I'm not sure the `context` is adding much here, since we only have one ... if you really want it then it should be within the describe block I think.

PR #440, opened 2019-06-04
1489 sink it

defong:
### Context
Publishing to Search and Compare api

Syncing via Pub

### Guidance to review
app/serializers/search_and_compare/course_serializer.rb
Changes made was flushed out using proper data that is available

`provider_enrichment&` are optional data.
`attribute(:StartDate)` &   `attribute(:ApplicationsAcceptedFrom)` seems to be optional data.
`course_enrichment.xxx.to_i,` needs to be int not float/decimal

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/search_and_compare_api_service.rb
+++ b/app/services/search_and_compare_api_service.rb
@@ -1,0 +1,44 @@
+module SearchAndCompareAPIService
+  class Connection
+    class << self
+      def api
+        Faraday.new(url: Settings.search_api.base_url) do |faraday|
+          faraday.response :logger unless Rails.env.test?
+          faraday.adapter Faraday.default_adapter
+          faraday.headers['Content-Type'] = 'application/json; charset=utf-8;'
+          faraday.headers['Authorization'] = "Bearer #{Settings.search_api.secret}"
+        end
+      end
+    end
+  end
+
+  class Request
+    class << self
+      # NOTE:
+      #   HTTP PUT "/api/courses" accepts a list of courses
+      #   It will only add or update, no DELETION.
+      def sync(course)
+        response = api.put(
+          "/api/courses/",
+          serialize(course)
+        )
+
+        response.success?
+      end
+
+      def api
+        Connection.api
+      end
+
+    private
+
+      def serialize(payload)
+        ActiveModel::Serializer::CollectionSerializer.new(
+          payload,
+          serializer: SearchAndCompare::CourseSerializer,
+          adapter: ActiveModel::Serializer::Adapter::JsonApi
+        ).to_json
+      end
+    end
+  end
+end
--- a/spec/services/search_and_compare_api_service_spec.rb
+++ b/spec/services/search_and_compare_api_service_spec.rb
@@ -1,0 +1,66 @@
+require "rails_helper"
+
+describe SearchAndCompareAPIService do
+  describe "Connection.api" do
+    subject { described_class::Connection.api }
+
+    it "exposes an Faraday connection" do
+      should be_instance_of(Faraday::Connection)
+    end
+
+    it "uses the configured URL as the base" do
+      expect(subject.url_prefix.to_s).to eq(Settings.search_api.base_url + "/")
+    end
+
+    it "uses the configured secret for the bearer token" do
+      expect(subject.headers["Authorization"]).to eq("Bearer #{Settings.search_api.secret}")
+    end
+  end
+
+  describe "Request" do
+    let(:request) { described_class::Request }
+    let(:body) do
+      ActiveModel::Serializer::CollectionSerializer.new(
+        [course],
+        serializer: SearchAndCompare::CourseSerializer,
+        adapter: :attributes
+      )
+    end
+
+    let(:published_enrichment) do
+      build :course_enrichment, :published,
+            created_at: 5.days.ago
+    end
+
+    let(:course) { create :course, enrichments: [published_enrichment] }
+
+    let(:course_code) { course.course_code }
+    let(:provider_code) { course.provider.provider_code }
+
+    let(:status) { 200 }
+
+    before do
+      stub_request(:put, "#{Settings.search_api.base_url}/api/courses/")
+        .with { |req| req.body == body.to_json }
+        .to_return(
+          status: status
+        )
+    end
+
+    describe 'sync' do
+      subject {
+        request.sync(
+          [course]
+        )
+      }
+      describe "with a normal response" do
+        it { should eq true }
+      end
+
+      describe "with a bad response" do
+        let(:status) { 400 }
+        it { should eq false }
+      end
+    end
+  end
+end


defong:
[[M] Sync course to search-and-compare](https://trello.com/c/wXvgCtNG/1489-m-sync-course-to-search-and-compare)

app/services/search_and_compare_api_service.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  class Request
+    class << self
+      # NOTE:
+      #   HTTP PUT "/api/courses" accepts a list of courses
+      #   It will only add or update, no DELETION.
+      # WARNING:
+      #   HTTP POST "/api/courses" accepts a list of courses
+      #   It will DELETE ALL EXISTING COURSES beforehand.
```
timabell: remove till code added

timabell:
* code climate is unhappy
* tidy up the PUT/POST comment to match the code
then I'll have another look

defong:
>     * code climate is unhappy

`code climate` complaints is the reviewer judgement ie you, to see if it needs resolving.

imo, its a `wontfix` as the other file its complaining will be removed in due course

app/services/search_and_compare_api_service.rb
defong: removed

app/serializers/search_and_compare/course_serializer.rb
```diff
@@ -41,7 +41,7 @@ class CourseSerializer < ActiveModel::Serializer
     attribute(:Name)                                  { object.name }
     attribute(:ProgrammeCode)                         { object.course_code }
     # using server date time not utc, so it's local date time?
-    attribute(:StartDate)                             { object.start_date.utc.strftime('%Y-%m-%dT%H:%M:%S') }
+    attribute(:StartDate)                             { object.start_date&.utc&.strftime('%Y-%m-%dT%H:%M:%S') }
```
misaka: Why would `.utc` return `nil`? I don't think you need that second guard. And does this return the right thing when `start_date` is nil? I'm guessing 'null'?

app/serializers/search_and_compare/course_serializer.rb
```diff
@@ -71,7 +71,7 @@ class CourseSerializer < ActiveModel::Serializer
     # Campuses_related_Mapping
     attribute(:Campuses)                              { campuses }
     # using server date time not utc, so it's local date time?
-    attribute(:ApplicationsAcceptedFrom)              { object.applications_open_from.to_date.strftime('%Y-%m-%dT%H:%M:%S') }
+    attribute(:ApplicationsAcceptedFrom)              { object.applications_open_from&.to_date&.strftime('%Y-%m-%dT%H:%M:%S') }
```
misaka: Similar to the guard on `.utc`, I would avoid guarding against `.to_date` returning nil ... if it does, I think that's an error and we should find out about it.

app/services/search_and_compare_api_service.rb
```diff
@@ -0,0 +1,42 @@
+module SearchAndCompareAPIService
+  class Connection
+    def self.api
```
misaka: You use `class << self` below in lieu of `self.` like here. Is it worth keeping it consistent?

defong: borrowed that from 'ManageCoursesAPIService'

but you mean do it like the `Request` one?

```ruby 
class Request
    class << self

```

app/serializers/search_and_compare/course_serializer.rb
```diff
@@ -184,9 +184,9 @@ def fees
         }
       else
         {
-          Uk: course_enrichment.fee_uk_eu,
-          Eu: course_enrichment.fee_uk_eu,
-          International: course_enrichment.fee_international,
+          Uk: course_enrichment.fee_uk_eu.to_i,
+          Eu: course_enrichment.fee_uk_eu.to_i,
+          International: course_enrichment.fee_international.to_i,
```
misaka: I don't see any tests changed to reflect this, are these covered?

misaka:
> `code climate` complaints is the reviewer judgement ie you, to see if it needs resolving.

Ya, I agree, it would be smart if the reviewer had a look and decided whether to enforce what CodeClimate is saying or not.

misaka:
> > Similar blocks of code found in 2 locations. Consider refactoring.
> >
> > ```
> >   class Connection
> >     def self.api
> >       Faraday.new(url: Settings.search_api.base_url) do |faraday|
> >         faraday.response :logger unless Rails.env.test?
> >         faraday.adapter Faraday.default_adapter
> > ```
> >
> > Severity: Minor
> > Found in app/services/search_and_compare_api_service.rb and 1 other location - About 45 mins to fix
> > 
> > Similar blocks of code found in 2 locations. Consider refactoring. New
> > 
> > ```
> >   class Connection
> >     def self.api
> >       Faraday.new(url: Settings.manage_api.base_url) do |faraday|
> >         faraday.response :logger unless Rails.env.test?
> >         faraday.adapter Faraday.default_adapter
> > ```
> > 
> > Severity: Minor
> > Found in app/services/manage_courses_api_service.rb and 1 other location - About 45 mins to fix 

> imo, its a `wontfix` as the other file its complaining will be removed in due course

Ya, also agreed that this doesn't need to be fixed since one of them is going away anyway, although it could've been nice to have a standard way to build a Faraday connection.

app/services/search_and_compare_api_service.rb
misaka: I'm guilty of mix-and-matching this in `lib/mcb.rb` ... but I think I mistakenly started that with `self.` when it really should've been `class << self`. I plan on fixing that one sometime rsn.

app/serializers/search_and_compare/course_serializer.rb
defong: :cry:

app/services/search_and_compare_api_service.rb
defong: gotcha, 

```ruby

return reject if self == misha
```

app/serializers/search_and_compare/course_serializer.rb
defong: this one is not a error

`object.applications_open_from`

that derives from campus/site/site_status/location 

whereby the user can set the `actual application open date`

so the output i want is either the date or nil

@benilovj to confirm

app/serializers/search_and_compare/course_serializer.rb
misaka: I think I was partly confused and forgot when commenting that you need `&.` after `utc` since otherwise it'll try to run `nil.utc` if `objects.start_date` is nil. I guess this is one of the reasons `&.` is considered an anti-pattern by some.

app/serializers/search_and_compare/course_serializer.rb
misaka: Similar to above, I think my comment was partly down to forgetting how `&.` works exactly.

defong:
I'm going hold of from merging til next week, kept getting distracted, plus would like to elicit some basic validation to prevent `ruby` going pub then get run over with `c#` 

maybe another pr
but items of interest

course must 
- have dfe subjects
- have findable sites

defong:
see #508 see #507

PR #444, opened 2019-06-05
[XXX] Update the QA environment config for MCB

benilovj:
### Context
We're switching over to use a new QA environment (that's managed by VSTS).

Update `mcb` to connect to the new environment instead of the Travis-managed one.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


benilovj:
An existing flaky test is breaking on VSTS ☹️ merging anyway.

PR #445, opened 2019-06-05
[1585] fix use of max-pages and page reporting in apiv1 mcb commands

misaka:
### Background

In many of the mcb commands that query the apiv1, the max-pages isn't respected and hard-set to 1.

### What is being changed

Fix this. This was fixed in 1542 for the `mcb apiv1 course find` command, the same approach needs to be copied to the other apiv1 commands.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -161,6 +161,31 @@ module MCB
       opts
     end
 
+    def display_pages_received(page:, max_pages:, next_url:)
+      # (current) page starts at 0
+      pages = page + 1
+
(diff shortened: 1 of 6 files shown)


misaka:
[Fix use of max-pages option in mcb apiv1 commands](https://trello.com/c/j2W6vOS6/1585-fix-use-of-max-pages-option-in-mcb-apiv1-commands)

PR #446, opened 2019-06-06
[1586] mcb courses edit: edit title

benilovj:
### Context
Until we've built all course editing features for publishers to self-serve, we need support tooling to be able to action support requests for changes.

- add a new `mcb courses edit` command
- allow editing individual courses or all courses for a provider
- some basic validation on the provider, course codes and the user making the changes
- editing of course title

### Guidance to review

I will add the editing of other aspects of course in subsequent PRs, I wanted to lay the groundwork with this one and get feedback.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -11,8 +11,19 @@ module MCB
     end
 
     def run
+      finished = false
       puts "Editing #{course_codes.join(', ')}"
-      edit_title
+      until finished
+        choice = @cli.choose do |menu|
+          menu.choice(:exit) { finished = true }
+          menu.choice("edit title")
+        end
+
+        case choice
+        when "edit title"
+          edit_title
+        end
+      end
     end
 
   private
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -11,21 +11,32 @@ module MCB
     end
 
     def run
+      puts "Editing #{course_codes.join(', ')}"
       edit_title
     end
 
   private
 
     def edit_title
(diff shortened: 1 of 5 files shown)


lib/mcb/courses_editor.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    def ask_title
+      @cli.ask("New course title?  ")
+    end
+
+    def check_authorisation
+      @courses.each { |course| raise Pundit::NotAuthorizedError unless can_update?(course) }
+    end
+
+    def print_existing(attribute_name)
```
tvararu: I expect these methods will see more use in future PRs with the rest of the fields :+1:

spec/lib/mcb/commands/courses/edit_spec.rb
```diff
@@ -0,0 +1,70 @@
+require 'mcb_helper'
+
+describe 'mcb courses edit' do
+  def edit(provider_code, course_codes, *input_cmds)
+    stderr = nil
+    output = with_stubbed_stdout(stdin: input_cmds.join("\n"), stderr: stderr) do
+      cmd.run([provider_code] + course_codes)
+    end
+    [output, stderr]
```
misaka: Looking at this, it feels like `with_stubbed_stdout` should return `(stdin, stdout)`.

benilovj: I've stolen this pattern from a few of the other `mcb` specs, we should fix this across the piece on a separate PR.

PR #447, opened 2019-06-06
[1580] Add validations to edit locations

tvararu:
### Context

This will allow us to display an appropriate error message on the frontend when the user submits an empty array of sites.

Pushes a validation message to `@course.errors` and checks for its presence before rendering the response.

### Guidance to review

The validation is done at the controller level instead of the model. This is because `sites = []` is something that we can validate against, but we can't actually revert easily from what I can tell because of the `remove_site!` side effects that occur when it's called.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -53,8 +53,13 @@ module API
 
         site_ids = params[:course][:sites_ids]
         @course.sites = @provider.sites.where(id: site_ids) if site_ids.present?
+        # This validation is done at the controller level instead of the model.
+        # This is because sites = [] is something that we can validate against,
+        # but we can't actually revert easily from what I can tell because of the
+        # remove_site! side effects that occur when it's called.
+        @course.errors[:sites] << "^You must choose at least one location" if site_ids == []
 
-        if @course.valid?
+        if @course.errors.empty? && @course.valid?
           render jsonapi: @course.reload
         else
           render jsonapi_errors: @course.errors, status: :unprocessable_entity
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -47,12 +47,7 @@ module API
       end
 
       def update
-        enrichment = if @course.enrichments.draft.any?
-                       @course.enrichments.draft.first
-                     else
-                       @course.enrichments.new(status: 'draft')
-                     end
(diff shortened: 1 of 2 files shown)


tvararu:
[[M] Course basic details - Edit locations validations](https://trello.com/c/QkdCYsJL/1580-m-course-basic-details-edit-locations-validations)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -47,19 +47,15 @@ def publish
...
-                       @course.enrichments.new(status: 'draft')
-                     end
-
+        enrichment = first_draft_or_new_enrichment
         enrichment.assign_attributes(update_params)
         enrichment.save
 
         site_ids = params[:course][:sites_ids]
         @course.sites = @provider.sites.where(id: site_ids) if site_ids.present?
+        @course.errors[:sites] << "^You must choose at least one location" if site_ids == []
```
```suggestion
        @course.errors[:sites] << "^You must choose at least one location" if site_ids.blank?
```

app/controllers/api/v2/courses_controller.rb
```diff
@@ -68,6 +64,14 @@ def update
 
     private
 
+      def first_draft_or_new_enrichment
+        if @course.enrichments.draft.any?
+          @course.enrichments.draft.first
+        else
+          @course.enrichments.new(status: 'draft')
```
defong: do you not need to clone the the last published enrichment?

tvararu: This is logic that already existed, I only moved it around because codeclimate is complaining

app/controllers/api/v2/courses_controller.rb
tvararu: This won't work, because it catches `nil` too:

```ruby
[1] pry(main)> nil.blank?
=> true
[2] pry(main)> nil == []
=> false
```

I thought about doing `!site_ids.nil? && site_ids.empty?` but `== []` seems more explicit...

misaka:
See #448 for a way to replace `add_site!` and `remove_site!` with callback functionality ... if feasible, this would allow us to move these validations to the model, which I would greatly prefer.

tvararu:
@misaka I greatly prefer that too, however in the interest of shipping this sooner than later, would prefer to work on that as a follow-up.

PR #448, opened 2019-06-07
[1580] use callbacks to add or remote sites to course

misaka:
- this replaces the functionality of add_site! and remove_site!

### Context

We should be able to add or remove sites to a course without having to resort to custom methods like `add_site!` or `remove_site!`.

Replace the functionality in those methods with callbacks in `SiteStatus`.

### Guidance to review

This code was done as a proof of concept and, on account of it being pretty late and me not being able to see straight, could probably use a little tidying up. It also doesn't remove the mentioned methods, that might be slightly more work.

This change could probably happen independent of `1580-edit-locations-validations`, if we think it's a good idea.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 # == Schema Information
 #
 # Table name: course_site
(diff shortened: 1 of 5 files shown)


app/models/site_status.rb
```diff
@@ -19,6 +20,21 @@ class SiteStatus < ApplicationRecord
 
   after_initialize :set_defaults
   before_validation :set_vac_status
+  before_create :set_new_status
+
+  def set_new_status
+    if !course.site_statuses.empty? && !course.site_statuses.all?(&:status_new_status?)
+      self.status = 'R'
```
benilovj:
```suggestion
      self.status = : running
```

defong: Just wondering that this change will affect the factory...

misaka:
I'm cool with this being separate. I'll prep it up now, and probably give it a new card.

app/models/site_status.rb
misaka: @defong I'll have a look.

misaka:
Will re-cut this PR from another branch

PR #450, opened 2019-06-07
[1587] Create enrichment in courses#update only when necessary

tvararu:
### Context

Currently the courses#update action always creates an enrichment when called. This leads to two bugs:

- When updating the sites, the course is set to draft
- When the user opens the enrichment pages but makes no changes and
  hits "Save", the course changes state

This wraps the enrichment logic in a guard so that it isn't run unnecessarily.

The enrichment and site update logic are both wrapped in private methods to make the update action a bit easier to parse.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -47,17 +47,8 @@ module API
       end
 
       def update
-        enrichment = first_draft_or_new_enrichment
-        enrichment.assign_attributes(update_params)
-        enrichment.save
-
-        site_ids = params[:course][:sites_ids]
-        @course.sites = @provider.sites.where(id: site_ids) if site_ids.present?
-        # This validation is done at the controller level instead of the model.
-        # This is because sites = [] is something that we can validate against,
-        # but we can't actually revert easily from what I can tell because of the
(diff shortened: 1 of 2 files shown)


tvararu:
[Editing locations for a published course puts it into draft](https://trello.com/c/rPp9Udhw/1587-editing-locations-for-a-published-course-puts-it-into-draft)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -47,27 +47,34 @@ def publish
       end
 
       def update
+        update_enrichment if enrichment_params.values.any?
+        update_sites
```
benilovj: Should this be
```suggestion
        update_sites if params[:course][:sites_ids].present?
```
for symmetry?

tvararu: Yeah, was thinking about putting that in its own `sites_params` method 👍 I'll update

tvararu: Nevermind; this makes codeclimate unhappy. Have used a guard clause instead.

misaka:
Nice one.

PR #451, opened 2019-06-07
[1598] use callbacks to add or remote sites to course

misaka:
### Context

We should be able to add or remove sites to a course without having to resort to custom methods like add_site! or remove_site!.

Replace the functionality in those methods with callbacks in SiteStatus.

### Guidance to review

This code was done as a proof of concept and, on account of it being pretty late and me not being able to see straight, could probably use a little tidying up. It also doesn't remove the mentioned methods, that might be slightly more work.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -19,6 +19,21 @@ class SiteStatus < ApplicationRecord
 
   after_initialize :set_defaults
   before_validation :set_vac_status
+  before_create :set_new_status
+
+  def set_new_status
+    if !course.site_statuses.empty? && !course.site_statuses.all?(&:status_new_status?)
+      self.status = :running
+    end
+  end
+
+  def destroy
+    if course.site_statuses.any?(&:status_new_status?)
+      self.delete
+    else
+      self.status_suspended!
+    end
+  end
 
   audited associated_with: :course
 
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -205,7 +205,15 @@ RSpec.describe Course, type: :model do
                  ])
         }
 
-        its(:has_vacancies?) { should be false }
+        before do
+          SiteStatus.skip_callback(:create, :before, :set_new_status)
+        end
+
+        after do
+          SiteStatus.set_callback(:create, :before, :set_new_status)
+        end
+
+        its(:has_vacancies?) { should be(false) }
       end
     end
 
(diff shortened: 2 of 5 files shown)


heidar:
What's the plan with this?

misaka:
Been reviewing code all day, will try to pick this up once I finish something else ... RSN!

heidar:
[Refactor: use callbacks to add or remove sites to a course](https://trello.com/c/JUYc3Wnd/1598-refactor-use-callbacks-to-add-or-remove-sites-to-a-course)

misaka:
:shipit: 🍜 🥇

tvararu:
@misaka @heidar I've added an additional test to check courses that have both a `new` and `running` site. If we're happy with the PR we squash and merge.

spec/models/site_status_spec.rb
```diff
@@ -38,6 +38,179 @@
     it { should be_audited.associated_with(:course) }
   end
 
+  describe 'creation' do
+    context 'when course has a running site' do
+      let(:site1)       { create(:site) }
```
defong: `create(:site) `

is flawed 

```ruby
 let(:site1)       { create(:site) } # site1.provider.id == 1
 let(:site2)       { create(:site) }  # site1.provider.id == 2

```

`site` model has a validate method that sets the `code` based on the associated `provider`

`code` needs to be unique for the `provider`

see
https://github.com/DFE-Digital/manage-courses-backend/blob/38c05dadca27976919a5271797f9ddd1736bbf6b/app/models/site.rb#L46



maybe something like

```ruby
 let(:provider)  { build(:provider, sites: [site1, site2]) } 
 let(:site1)       { build(:site) } 
 let(:site2)       { build(:site) } 
 let(:site_status1) { build(:site_status, :running, site: site1) }
 let(:course)     {create(:course, provider: provider, site_statuses: [site_status1])}

```

please try to defer creation at the latest point in time and only have one `create` method call

Otherwise its ends up like https://github.com/DFE-Digital/manage-courses-backend/pull/472 for the pain factory

spec/models/course_spec.rb
```diff
@@ -474,142 +482,4 @@
...
-        expect(new_site_status.status).to eq("running")
-      end
-
-      it "suspends the site when an existing site is removed" do
-        expect { subject.remove_site!(site: existing_site) }.
-          to change { existing_site_status.reload.status }.from("running").to("suspended")
-      end
-    end
-
-    describe '#accrediting_provider_description' do
```
davidgisbey: I think this should be in a different block and needs to stay in the tests. It's currently nested when it shouldn't be @misaka

misaka: Good spot. Thanks!

misaka: Interesting. I wonder who put it there in the first place ...

image

-- https://github.com/DFE-Digital/manage-courses-backend/blame/84ea318ac52bfe7620be61fa82349762847ae234/spec/models/course_spec.rb#L563

defong: I can only wonder who would approve such a change...

heidar: Damn! Good spot David. I did wonder why there was a merge conflict with this code...

heidar: It's back in, resolving

app/models/site_status.rb
```diff
@@ -19,6 +19,21 @@ class SiteStatus < ApplicationRecord
...
+  before_create :set_new_status
+
+  def set_new_status
+    if !course.site_statuses.empty? && !course.site_statuses.all?(&:status_new_status?)
+      self.status = :running
+    end
+  end
+
+  def destroy
+    if course.site_statuses.any?(&:status_new_status?)
```
benilovj: @tvararu Is this logic correct for mixed courses? I notice that the spec for mixed courses only covers adding new locations, not deleting them.

tvararu: Good spot... I was going to add one but I got distracted and forgot. Will rectify 👍

PR #452, opened 2019-06-07
Add new staging env and settings

smneal:
### Context
Add config bits for new staging env

Add config for database and rails env

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


config/settings/new_staging.yml
```diff
@@ -0,0 +1,6 @@
+manage_api:
```
defong: remember to change me

PR #454, opened 2019-06-07
1460 refactor factories to simplify relationships

davidgisbey:
### Context
The factories are full of dependancies. This is a very rough first attempt at ironing **some** out.

- Remove with_site_statuses, subject_count, site_count & course_count
- Rebuilt some of the tests to remove 

### Guidance to review

This is an initial first draft at removing **some** not all of the dependencies. Right now a bunch of the tests are flagging. I will have to go through each spec manually and later creates to builds or rebuild how the objects are created where necessary.

Just looking for general feedback. Thanks!
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/factories/course_enrichments.rb
+++ b/spec/factories/course_enrichments.rb
@@ -16,8 +16,8 @@
 
 FactoryBot.define do
   factory :course_enrichment do
-    provider
     course
+    provider { course.provider }
     status { :draft }
 
     about_course { Faker::Books::Dune.quote }
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -28,8 +28,7 @@ FactoryBot.define do
     name { Faker::ProgrammingLanguage.name }
     qualification { :pgce_with_qts }
     with_higher_education
-
-    association(:provider)
+    provider
 
     study_mode { :full_time }
     resulting_in_pgce_with_qts
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -33,15 +33,16 @@ FactoryBot.define do
 
     study_mode { :full_time }
     resulting_in_pgce_with_qts
-
     transient do
-      subject_count      { 1 }
-      subjects           { build_list(:subject, subject_count) }
-      with_site_statuses { [] }
       with_enrichments   { [] }
       age                { nil }
       enrichments        { [] }
     end
+
+    trait :with_subject do
+      subjects { [build(:subject)] }
+    end
+
 
     after(:build) do |course, evaluator|
       if evaluator.age.present?
@@ -52,19 +53,6 @@ FactoryBot.define do
     end
 
     after(:create) do |course, evaluator|
-      course.subjects << evaluator.subjects.map { |subject|
-        subject.is_a?(Subject) ? subject : create(*subject)
-      }
-
-      evaluator.with_site_statuses.each do |traits|
-        attrs = { course: course }
-        if traits == [:default]
-          create(:site_status, attrs)
-        else
-          create(:site_status, *traits, attrs)
-        end
-      end
-
       evaluator.with_enrichments.each do |trait, attributes = {}|
         defaults = {
           ucas_course_code: course.course_code,
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -34,9 +34,8 @@ FactoryBot.define do
     study_mode { :full_time }
     resulting_in_pgce_with_qts
     transient do
-      with_enrichments   { [] }
-      age                { nil }
-      enrichments        { [] }
+      age         { nil }
+      enrichments { [] }
     end
 
     trait :with_subject do
@@ -50,21 +49,21 @@ FactoryBot.define do
         course.changed_at = evaluator.age
       end
     end
-
+    #
     after(:create) do |course, evaluator|
-      evaluator.with_enrichments.each do |trait, attributes = {}|
-        defaults = {
-          ucas_course_code: course.course_code,
-          provider_code: course.provider.provider_code,
-        }
-        if evaluator.age.present?
-          defaults = defaults.merge(
-            created_at: evaluator.age,
-            updated_at: evaluator.age,
-          )
-        end
-        create(:course_enrichment, trait, attributes.merge(defaults))
-      end
+      # evaluator.with_enrichments.each do |trait, attributes = {}|
+      #   defaults = {
+      #     ucas_course_code: course.course_code,
+      #     provider_code: course.provider.provider_code,
+      #   }
+      #   if evaluator.age.present?
+      #     defaults = defaults.merge(
+      #       created_at: evaluator.age,
+      #       updated_at: evaluator.age,
+      #     )
+      #   end
+        # create(:course_enrichment, trait, attributes.merge(defaults))
+      # end
 
       course.enrichments += evaluator.enrichments.map do |enrichment|
         enrichment.tap { |e| e.provider_code = course.provider.provider_code }
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -43,7 +43,6 @@ FactoryBot.define do
       subjects { [build(:subject)] }
     end
 
-
     after(:build) do |course, evaluator|
       if evaluator.age.present?
         course.created_at = evaluator.age
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -49,22 +49,7 @@ FactoryBot.define do
         course.changed_at = evaluator.age
       end
     end
-    #
     after(:create) do |course, evaluator|
-      # evaluator.with_enrichments.each do |trait, attributes = {}|
-      #   defaults = {
-      #     ucas_course_code: course.course_code,
-      #     provider_code: course.provider.provider_code,
-      #   }
-      #   if evaluator.age.present?
-      #     defaults = defaults.merge(
-      #       created_at: evaluator.age,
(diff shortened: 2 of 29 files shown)


davidgisbey:

spec/factories/course_enrichments.rb
```diff
@@ -16,8 +16,8 @@
 
 FactoryBot.define do
   factory :course_enrichment do
-    provider
     course
+    provider { course.provider }
```
defong: I don't think `provider` & `course` are actually part of `course_enrichment`

relation is defined by `provider_code` & `ucas_course_code`

It be nice to remove

spec/factories/courses.rb
```diff
@@ -33,16 +33,16 @@
...
     transient do
-      subject_count      { 1 }
-      subjects           { build_list(:subject, subject_count) }
-      with_site_statuses { [] }
       with_enrichments   { [] }
       age                { nil }
       enrichments        { [] }
     end
 
+    trait :with_subject do
```
defong: feels a bit over the top

may have that

`subjects { [build(:subject)] }` as the default and callee can override it with blank if not needed

spec/factories/provider_enrichments.rb
```diff
@@ -16,12 +16,8 @@
 
 FactoryBot.define do
   factory :provider_enrichment do
-    transient do
-      age { nil }
-    end
-
+    association :provider
```
defong: maybe in `after(:build)`?


given a provider have a trait to create it `with_enrichment`...

```ruby

    after(:build) do |enrichment, evaluator|
      if enrichment.provider.blank?
```

spec/factory_specs/providers_spec.rb
```diff
@@ -7,9 +7,4 @@
     expect(provider).to be_instance_of(Provider)
     expect(provider).to be_valid
   end
-
-  it "creates a provider with enrichments" do
```
defong: why was the test removed?

spec/lib/mcb/commands/apiv1/courses/find_spec.rb
```diff
@@ -8,6 +8,8 @@
 
     site_status2 = create(:site_status)
     course2 = site_status2.course
+    create_subject = create(:subject)
```
defong: it feels like the apiv1 specs are a good barometer to illustrate the whole of the course/site/subject 

extract of the code
```ruby
# no lonnger makes sense  # The site_status factory is an easy way to create a course and it's site
    site_status1 = create(:site_status)
    course1 = site_status1.course

    site_status2 = create(:site_status)
    course2 = site_status2.course
    create_subject = create(:subject)
    course2.update(subjects: [create_subject]) # was there really a need for this but?
    subject2 = course2.subjects.first

```

spec/lib/mcb/commands/courses/show_spec.rb
```diff
@@ -13,7 +13,9 @@
 
     site_status2 = create(:site_status)
     course2      = site_status2.course
-    subjects2    = course2.subjects
+    new_subject = create(:subject)
+    course2.subjects = [new_subject]
```
defong: same as above

spec/factories/course_enrichments.rb
misaka: What do you mean @defong? These relationships just use those codes to associate the records, what do you want to remove?

spec/factories/courses.rb
misaka: This may have come from me ... I was saying that our design philosophy should be that only those associations that are _required_ by an object should be created in the factory. Anything else should be expressed explicitly. Can a `course` object exist with a `subject`?

spec/factories/provider_enrichments.rb
misaka: @defong what's the benefit over doing it this way?

@davidgisbey is `association` needed here? Is there a reason not just to use `provider` here?

spec/factory_specs/providers_spec.rb
misaka: Sounds like it should be converted to a test of the `:with_enrichment` trait. Except that we tend not to test traits separately.

spec/lib/mcb/commands/apiv1/courses/find_spec.rb
misaka: @defong course2's subject is used later on when testing that it gets rendered out.

@davidgisbey does it make sense to have both `create_subject` and `subject2` here? We could probably just use `subject` instead of `create_subject`.

spec/requests/api/v2/providers/courses/publish_spec.rb
```diff
@@ -20,18 +20,23 @@
     end
 
     before do
+      organisation.users = [user]
```
misaka: This is already happening on line 5 ... is it an issue of having to `reload`?

spec/factories/course_enrichments.rb
defong: opps guess it was defined 

so maybe 

``ruby
ucas_course_code { course.course_code }
provider_code { course.provider.provider_code }
```

since 
```ruby

course
```

as the driver

spec/factories/provider_enrichments.rb
defong: @misaka don't this link from `providers`

```ruby
   trait :with_enrichment do
      enrichments { [build(:provider_enrichment)] }
    end

```

creates two provider

when you 
do either

```ruby

build(:provider, :with_enrichments)
```

```ruby
create(:provider, :with_enrichments)

```

spec/models/course_spec.rb
```diff
@@ -129,54 +135,73 @@
...
+          create(:site_status,
+                 :findable,
+                 :no_vacancies,
+                 course: site_status_with_vacancies.course)
         }
 
+        before do
+          site_status_with_vacancies
+          findable_site_status_with_no_vacancies
+          findable_site_status_with_no_vacancies
```
misaka: This doesn't actually do the same as the original ... it looks like it creates two `site_statuses`, but it doesn't since `let` gives us memoized values. BUT ... I can't see why we needed two in the first place, so how about we just get rid of one and call it a day?

spec/models/course_spec.rb
```diff
@@ -99,27 +99,33 @@
   end
 
   context 'with site statuses' do
+    let(:findable_site_status) { create(:site_status, :findable) }
+    let(:site_status_with_any_vacancy) { create(:site_status, :with_any_vacancy, course: course) }
+    let(:default_site_status) { create(:site_status, course: course) }
+    let(:applications_being_accepted_now_site_status) { create(:site_status, :applications_being_accepted_now, course: course) }
+    let(:applications_being_accepted_in_future_site_status) { create(:site_status, :applications_being_accepted_in_future, course: course) }
+    let(:site_status_with_no_vacancies) { create(:site_status, :with_no_vacancies, course: course) }
+    let(:course) { findable_site_status.course }
```
misaka: I know I originally started the pattern of creating a `site_status` and use the course it creates, but I'm not sure this is a great idea, and even it it was then, it's probably not a great idea here, unfortunately. Reason being, I think that we should communicate to the reader that the course is central to the test. Can we create the course in it's own `let` and add it to `findable_site_status` just like we've done with the other `site_statuses`?

I think there'll be an added benefit below, too.

spec/factories/course_enrichments.rb
misaka: If I understand this correctly, and by "this" I mean "factories", by doing `provider { course }` we ensure that we just re-use the course's provider, instead of creating a separate one.

misaka: Had a convo with @defong IRL, and he is now happy with this.

spec/factories/courses.rb
misaka: So what's come out of this is that a `course` should not exist with a `subject`. I say we change this here when that there changes in the model.

spec/factories/provider_enrichments.rb
misaka: Discussed IRL and @defong now accepts that this works.

misaka: Actually, @davidgisbey, is the `association` bit needed?

spec/lib/mcb/commands/apiv1/courses/find_spec.rb
misaka: It might make sense to create the `course` first so that we focus on that. Again, I started the pattern of creating `site_status` first, mea culpa ... I'm not sure that was a good idea any more.

spec/factories/course_enrichments.rb
davidgisbey: Ok great.

spec/lib/mcb/commands/apiv1/courses/find_spec.rb
davidgisbey: @misaka sounds like a plan. I'll go through my tests and sort them (there are a lot 😬 😆)

spec/factories/provider_enrichments.rb
davidgisbey: There is not. I'll standardise the factories to remove the associations.

spec/lib/mcb/commands/courses/show_spec.rb
davidgisbey: Done 👍

spec/models/course_spec.rb
davidgisbey: Good shout. Have removed the duplicate findable site status.

PR #455, opened 2019-06-07
[1606] raise error when sync to sac fails

misaka:
### Context

We're getting errors when we try to sync a course with search and compare.

Ensure an error is raised when we try to sync a course so we have a better understanding of what's gone wrong.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -1,3 +1,5 @@
+# coding: utf-8
+
 module API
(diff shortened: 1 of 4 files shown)


misaka:
[Expose errors when sync/publish to S&C fails](https://trello.com/c/ewX43wO5/1606-expose-errors-when-sync-publish-to-sc-fails)

PR #456, opened 2019-06-10
[1586] mcb courses edit part 3

benilovj:
### Context
Follow-up to #446 and #449.

Until we've built all course editing features for publishers to self-serve, we need support tooling to be able to action support requests for changes.

- `mcb courses edit` prints (no more than 2) courses being edited so the support agent can see what they're doing
- edit qualifications
- edit study mode
- edit accredited body
- edit start date
- edit application acceptance date
- convenience function for syncing courses to Find

### Guidance to review

This should cover all the editing features currently on the extra tooling branch. New course creating will be in subsequent PRs.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


benilovj:
Code climate thinks that the `CoursesEditor` class has too many methods. Any proposals from anyone as to how best split it up?

dankmitchell:
@benilovj we can either spilled the class up or change the CodeClimate defaults - https://docs.codeclimate.com/docs/default-analysis-configuration#section-per-language-mass-threshold-defaults

tvararu:
@benilovj you can also tell it we don't care for this file:

Screenshot 2019-06-10 at 09 58 54

lib/mcb/courses_editor.rb
```diff
@@ -118,13 +118,20 @@ def edit_accredited_body
...
+        rescue ActiveRecord::RecordNotFound
+          puts "Can't find accredited body; please enter one that exists."
+        end
       end
       new_accredited_body
     end
 
+    def ask_accredited_body_once
+      code = @cli.ask("Provider code of accredited body (leave blank if self-accredited)  ")
+      code.present? ? Provider.find_by!(provider_code: code) : @provider
```
defong: upcase?

benilovj: yeah, good call

benilovj: Now fixed in 3bd1fde, thanks

lib/mcb/courses_editor.rb
```diff
@@ -71,10 +84,81 @@ def ask_route
       end
     end
 
+    def edit_qualifications
+      print_existing(:qualification)
+      update(qualification: ask_qualifications)
+    end
```
misaka: Ya, I think that there's scope to merge these `edit_*` methods into one, or potentially with the related `ask_*`, although I understand the separation of concerns you implemented there.(

benilovj: I've now:

- generalised the `edit_*` methods in cfacaa5
- extracted a `CoursesEditorCLI` class in ca0d01a

and that seems to satisfy Code Climate.

lib/mcb/courses_editor.rb
```diff
@@ -59,22 +78,63 @@ def ask_gcse_subject(subject)
...
+    def ask_accredited_body_once
+      code = @cli.ask "Provider code of accredited body (leave blank if self-accredited)  ", ->(str) { str.upcase }
+      code.present? ? Provider.find_by!(provider_code: code) : @provider
+    end
+
+    def ask_start_date
+      Date.parse(@cli.ask("Start date?  ") { |q| q.default = "September #{@courses.first.recruitment_cycle}" })
+    end
+
+    def ask_application_opening_date
```
defong: is this related to 

https://github.com/DFE-Digital/manage-courses-backend/blob/2794cc115207c201a45b06fccda2d138689bbe67/app/models/course.rb#L150

if so can we rename the method 'ask_applications_open_from'

benilovj: > if so can we rename the method 'ask_applications_open_from'

The different names is a deliberate choice – the `ask_*` method names refer to the domain logic naming and the language we're using in the frontend, and these map to the database names using [this lovely mapping](https://github.com/DFE-Digital/manage-courses-backend/blob/ca0d01a3cb49fec9c6c867edd6cf35b42c5f4dd9/lib/mcb/courses_editor.rb#L3-L14) that I'm introducing as part of this PR.

I'm hoping that we eventually can rename the database fields to be in line with the language that we're using in the frontend, so I don't want the database names to be embedded everywhere, but that's a personal preference.

> 
> btw how does that one date get associated to a single site?

This PR defines a setter on `Course` here: https://github.com/DFE-Digital/manage-courses-backend/blob/ca0d01a3cb49fec9c6c867edd6cf35b42c5f4dd9/app/models/course.rb#L163-L165

benilovj: I'm gonna merge this PR so we can keep dismantling the extra tooling branch. Happy to discuss naming in person tomorrow and can make naming consistent with the database field names if that's the consensus opinion. Ultimately it shouldn't matter in the long run because:

1. this code will reduce and eventually disappear completely in future, as more and more functionality is moved to be self-serve
2. a lot of the `site_status` malarkey will also disappear as we simplify the data model.

PR #459, opened 2019-06-10
[1608] Sync courses when updating courses and site_statuses

tvararu:
### Context

The frontend doesn't necessarily know when a course is due to be synced with search-and-compare, the concern for that knowledge should be in the backend.

Call `ManageCoursesAPIService` in courses update.

### Guidance to review

🚢 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -20,11 +20,7 @@ module API
       end
 
       def sync_with_search_and_compare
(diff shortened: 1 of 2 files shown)


tvararu:
[Sync a course to search-and-compare when changing vacancies or locations](https://trello.com/c/2APzFeBv/1607-sync-a-course-to-search-and-compare-when-changing-vacancies-or-locations)

fofr:
> We think the majority of courses only have one site
But some will have 30.

PR #460, opened 2019-06-11
[1599] fix apiv1 provider find campuses rendering

misaka:
### Context

The previous commit to add `mcb courses show` tweaked and broke rendering for `mcb apiv1 provider find`.

Re-add campus table rendering.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/render/apiv1.rb
+++ b/lib/mcb/render/apiv1.rb
@@ -4,6 +4,20 @@ module MCB
       class << self
         include MCB::Render
 
+        def campuses_table(campuses, name: 'Campuses')
+          return if campuses.nil?
+
+          [
+            "#{name}:",
+            render_table_or_none(hashes_to_ostructs(campuses),
+                                 campuses_table_columns)
+          ]
+        end
+
+        def campuses_table_columns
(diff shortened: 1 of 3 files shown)


misaka:
[Fix "mcb apiv1 provider find" campuses table rendering](https://trello.com/c/m6QczJr9/1599-fix-mcb-apiv1-provider-find-campuses-table-rendering)

spec/factories/sites.rb
```diff
@@ -24,7 +24,7 @@
     address3 { Faker::Address.city }
     address4 { Faker::Address.state }
     postcode { Faker::Address.postcode }
-    region_code { 'london' }
+    region_code { 1 }
```
benilovj: why do we need this change? `london` is more descriptive than `1`.

benilovj: Nothing seems to break if I put this change back, so I'm gonna amend the commit that introduced this change.

benilovj:
I'm gonna merge this change because I would like to use the command that it fixes.

PR #465, opened 2019-06-11
Allow exiting requirements selection

timabell:
### Context

Accidentally set a requirement because 1 is usually exit.
Not spending too long on this.

Make exit an option

### Guidance to review

Suggestions for why exit isn't a symbol would be nice, also how to apply it to the other editing bits of course (route etc) as they don't follow the same pattern.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -47,7 +47,7 @@ module MCB
       database_attribute = LOGICAL_NAME_TO_DATABASE_NAME_MAPPING[logical_attribute]
       print_existing(database_attribute)
       user_response_from_cli = @cli.send("ask_#{logical_attribute}".to_sym)
-      if user_response_from_cli != "exit"
+      unless user_response_from_cli.nil?
         update(database_attribute => user_response_from_cli)
       end
     end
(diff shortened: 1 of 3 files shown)


benilovj:
@timabell I added some unit specs and fixed a subtle bug in the code by replacing a call of `menu.choice(:exit)` to `menu.choice("exit")` which was not matching the guard.

lib/mcb/courses_editor.rb
```diff
@@ -47,7 +47,9 @@ def edit(logical_attribute)
       database_attribute = LOGICAL_NAME_TO_DATABASE_NAME_MAPPING[logical_attribute]
       print_existing(database_attribute)
       user_response_from_cli = @cli.send("ask_#{logical_attribute}".to_sym)
-      update(database_attribute => user_response_from_cli)
+      unless user_response_from_cli.nil?
```
timabell: like it :+1:

spec/lib/mcb/courses_editor_spec.rb
```diff
@@ -41,24 +41,52 @@ def run_editor(*input_cmds)
...
-        expect { run_editor("edit maths", "equivalence_test", "exit") }.to change { course.reload.maths }.
-          from("must_have_qualification_at_application_time").to("equivalence_test")
+      describe "(maths)" do
+        it 'updates the maths setting when that is valid' do
+          expect { run_editor("edit maths", "equivalence_test", "exit") }.to change { course.reload.maths }.
+            from("must_have_qualification_at_application_time").to("equivalence_test")
+        end
+
+        it "doesn't change the setting if the user exits" do
+          expect { run_editor("edit maths", "exit", "exit") }.to_not change { course.reload.maths }.
```
timabell: A++ would read again

PR #466, opened 2019-06-11
[1460] Amended the usage of the provider and provider enrichments

defong:
### Context
Provider and provider enrichments factories usage

Amended the usage of the provider and provider enrichments

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -456,7 +456,7 @@ RSpec.describe Course, type: :model do
   end
 
   describe "#applications_open_from=" do
-    let(:provider) { create(:provider, sites: [sites]) }
+    let(:provider) { build(:provider, sites: [sites]) }
     let(:sites) { build_list(:site, 3) }
     let!(:existing_site_status) {
       sites.each do |site|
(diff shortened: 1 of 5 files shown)


defong:

spec/models/course_spec.rb
```diff
@@ -560,7 +560,7 @@
 
         context 'with published enrichment' do
           let(:provider_enrichment) { build(:provider_enrichment, :published, last_published_at: 1.day.ago) }
-          let(:provider) { create(:provider, enrichments: [provider_enrichment]) }
+          let(:provider) { build(:provider, enrichments: [provider_enrichment]) }
```
misaka: Does this really need to be `build`?

spec/models/course_spec.rb
```diff
@@ -578,10 +578,10 @@
 
             let(:accrediting_provider_enrichments) { [accrediting_provider_enrichment] }
             let(:provider_enrichment) do
-              build(:provider_enrichment,
-                    :published,
-                    last_published_at: 1.day.ago,
-                    accrediting_provider_enrichments: accrediting_provider_enrichments)
+              create(:provider_enrichment,
```
misaka: This is going to create a `provider` since 1) it's not used here, and 2) it's being created.

spec/models/course_spec.rb
davidgisbey: Changed

spec/models/course_spec.rb
davidgisbey: Done. Thanks for the feedback.

PR #468, opened 2019-06-11
[1460] Removed subjects from course factory

defong:
### Context

Subject creation in course factory

Removed unneeded chaos of having subject_count and its related usage in course factory 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/lib/mcb/commands/courses/show_spec.rb
+++ b/spec/lib/mcb/commands/courses/show_spec.rb
@@ -9,11 +9,10 @@ describe '"mcb courses show"' do
   end
 
   it 'displays the course info' do
-    _site_status1 = create(:site_status)
+    subject2 = build(:subject)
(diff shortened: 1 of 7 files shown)


defong:

spec/lib/mcb/commands/courses/show_spec.rb
```diff
@@ -9,11 +9,10 @@
   end
 
   it 'displays the course info' do
-    _site_status1 = create(:site_status)
+    subject2 = build(:subject)
+    site_status2 = build(:site_status)
```
misaka: I think the only reason `2` was appended to these was to differentiate it from `_site_status1`. If we're going to get rid of that, which isn't a bad call, then may as well remove the `2`.

PR #474, opened 2019-06-12
[1619] mcb console with environmental config

benilovj:
### Context

Part of supporting the live product involves making data changes on production. In order to allow making safe, audited changes we need to be able to connect the Rails console to production environments.

Add an `mcb` command for launching the Rails console:

```ruby
$ bundle exec mcb console -E qa -v
As a safety measure, please enter the expected RAILS_ENV for bat-qa-mcbe-as: qa
Loading qa environment (Rails 5.2.3)
irb(main):001:0>
```

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/initializers/mcb_audit.rb
+++ b/config/initializers/mcb_audit.rb
@@ -1,0 +1,10 @@
+if ENV['MCB_SETUP_AUDIT_USER'].present?
+  require 'mcb'
+  require 'mcb/config'
+
+  def verbose(msg)
+    Rails.logger.info msg
+  end
+
+  MCB.configure_audited_user if MCB.connecting_to_remote_db?
+end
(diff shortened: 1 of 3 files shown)


config/initializers/mcb_audit.rb
```diff
@@ -0,0 +1,10 @@
+if ENV.key?('MCB_AUDITED_USER_EMAIL')
```
misaka: So do we supply the email here or not? If not, we can give this a more appropriate name.

benilovj: oops - fixed

benilovj:
I just made a small tweak to the command to keep the linter happy.

PR #480, opened 2019-06-13
[1626] accomodate empty attributes when updating from jsonapi

misaka:
### Context

JSONAPI requests can come in with an empty set of attributes, if there are no attributes to update. When this happens, the deserializer doesn't even bother to create the root parameter (e.g. `course` in this case).

Default the course update attributes to an empty hash instead of requiring it to be present. i.e. we usually expect the params to come in as:

```
{
  course: {
    about_course: "This is a great course.",
    ...
  }
}
```

so we can do `params.require(:course)`. But now, when `course:` isn't present, we default it to `{}`.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -118,7 +118,7 @@ module API
 
       def enrichment_params
         params
-          .require(:course)
+          .fetch(:course, {})
           .except(:id, :type, :sites_ids, :sites_types)
           .permit(
             :about_course,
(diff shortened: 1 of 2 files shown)


misaka:
[Cannot update a course enrichment is nothing changes](https://trello.com/c/0J2v5VmW/1626-cannot-update-a-course-enrichment-is-nothing-changes)

misaka:
Why hacky? It's not what we're used to with HTML forms that submit the entire thing on POST/PUT, but this is an API so that's not a surprise. `fetch` serves this purpose perfectly by converting the `{}` into an `ActionController::Parameters` object for us.

PR #481, opened 2019-06-13
[1586] mcb courses edit: part 4

benilovj:
### Context
Follow-up to #446, #449 and #456 

Until we've built all course editing features for publishers to self-serve, we need support tooling to be able to action support requests for changes and new courses.

- extract out a method for handling multiple choice questions in `CoursesEditorCLI`

As a precursor to the new course creation:
- edit the age range on a record
- edit the course code

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -12,6 +12,7 @@ module MCB
       start_date: :start_date,
       application_opening_date: :applications_open_from,
       age_range: :age_range,
+      course_code: :course_code,
     }.freeze
 
     def initialize(provider:, requester:, course_codes: [])
@@ -66,7 +67,7 @@ module MCB
       puts "Existing values for course #{attribute_name}:"
       table = Tabulo::Table.new @courses.order(:course_code) do |t|
         t.add_column(:course_code, header: "course\ncode", width: 4)
-        t.add_column(attribute_name)
+        t.add_column(attribute_name) unless attribute_name == :course_code
       end
       puts table.pack(max_table_width: nil), table.horizontal_rule
     end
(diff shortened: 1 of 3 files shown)


lib/mcb/courses_editor_cli.rb
```diff
@@ -90,5 +97,23 @@ def ask_start_date
     def ask_application_opening_date
       Date.parse(@cli.ask("Applications opening date?  ") { |q| q.default = Date.today.to_s })
     end
+
+    def ask_course_code
+      @cli.ask("Course code?  ") do |q|
+        q.whitespace = :strip_and_collapse
+        q.validate = /\S+/
```
defong: Upper case it?

benilovj: Agreed - 40cc7cb

lib/mcb/courses_editor.rb
```diff
@@ -11,6 +11,8 @@ class CoursesEditor
       accredited_body: :accrediting_provider,
       start_date: :start_date,
       application_opening_date: :applications_open_from,
+      age_range: :age_range,
+      course_code: :course_code,
```
defong: was there a reason to add `course_code` only for 

` unless attribute_name == :course_code`

benilovj: Yes - when editing course code, the table library doesn’t like having two columns with the same name.

benilovj: I've changed that part of the code now so it's no longer in the hash: b0709d8

defong: Yeah i saw, the the more I ask the more magical it gets

benilovj: Then it’s easy - just ask fewer questions before approving :trollface:

PR #484, opened 2019-06-14
[1455] Use JSONAPI matchers instead of equality

heidar:
### Context

This test was too specific, checking each and every attribute. This
caused it to be too brittle.

I've replaced comparing the complete JSON output to using JSONAPI RSpec
matchers to check if the correct objects are rendered out, with the
relationships and attributes we expect.

There's no need to test this further than that, the serializer is unit
tested to make sure it gives us the correct data.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/access_request_spec.rb
+++ b/spec/requests/api/v2/access_request_spec.rb
@@ -57,97 +57,49 @@ describe 'Access Request API V2', type: :request do
     end
 
     context 'when authorised' do
-      let!(:first_access_request) { create(:access_request) }
-      let!(:second_access_request) { create(:access_request) }
+      let!(:access_request_1) { create(:access_request) }
+      let!(:access_request_2) { create(:access_request) }
 
       before do
-        Timecop.freeze
         access_requests_index_route
       end
 
-      after do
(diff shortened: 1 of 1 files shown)


heidar:
[Fix specs that are failing intermittently](https://trello.com/c/HbjiIiaV/1455-fix-specs-that-are-failing-intermittently)

heidar:
[flapping access request spec](https://trello.com/c/YS5rBj32/1499-flapping-access-request-spec)

misaka:
I think this is a good change overall, but I've avoided making this change myself because I think we lose a couple things with this approach to testing our JSONAPI responses:

1. we don't have a full view of what the JSONAPI response looks like, which makes understanding the response very easy
2. we loosen the contract on what the API can return ... if we add things to it we don't get a prompt with a failing spec

I'm not sure how to mitigate these ... I'm a little reluctant to throw them both away so I'd like to have some ideas for how we can address them. For example, we could satisfy 1 with a new mcb command: `mcb apiv2 access_request show --json`.

PR #488, opened 2019-06-14
[1638] site status fixes

benilovj:
### Context
- Un-suspending a site on a course creates duplicate site statuses
- Taking sites off a course with new and running site statuses deletes running site statuses (these should be suspended instead)

- a fix to comment flagged here: https://github.com/DFE-Digital/manage-courses-backend/pull/451/files#r293711836

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -28,7 +28,7 @@ class SiteStatus < ApplicationRecord
   end
 
   def destroy
-    if course.site_statuses.any?(&:status_new_status?)
+    if self.status_new_status?
       self.delete
     else
       self.status_suspended!
--- a/spec/models/site_status_spec.rb
+++ b/spec/models/site_status_spec.rb
@@ -39,10 +39,12 @@ RSpec.describe SiteStatus, type: :model do
   end
 
   describe 'creation' do
+    let(:provider)    { build(:provider) }
+    let(:course) { create(:course, provider: provider) }
+    let(:site1) { build(:site, provider: provider) }
+    let(:site2) { build(:site, provider: provider) }
+
     context 'when course has a running site' do
-      let(:provider)    { build(:provider) }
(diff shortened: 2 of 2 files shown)


app/models/site_status.rb
```diff
@@ -28,7 +28,7 @@ def set_new_status
   end
 
   def destroy
-    if course.site_statuses.any?(&:status_new_status?)
+    if self.status_new_status?
```
tvararu: 😞

spec/models/site_status_spec.rb
```diff
@@ -148,6 +141,30 @@
...
+
+      describe "the status" do
+        it "is set to running if a completely new site is added" do
+          new_site_status = SiteStatus.create course: course, site: site3
+
+          expect(new_site_status).to be_status_running
+        end
+
+        it "is set to running if the suspended site is re-added" do
+          course.sites << [site1]
```
misaka: So, if we have a look at `course.sites` before we append to it:

```ruby
[3] pry> course.sites
=> []
```

Because ... `sites` is only running sites. I think that this is the source of the problem, if we have a look at `site_statuses`:

```ruby
[4] pry> course.site_statuses
=> [#<SiteStatus:0x00007fabe7950908 id: 1, applications_accepted_from: Fri, 14 Jun 2019, course_id: 1, publish: "published", site_id: 1, status: "suspended", vac_status: "full_time_vacancies">]
```

I had a funny feeling about `sites` implicitly scoping our list of sites to `running`, and now I know why. I think `sites` should be un-scoped and we should use `sites.running` when that's what we really mean. Or maybe `running_sites`, but I'm not sure if there's any benefit to adding another named scope.

benilovj:
Closing since #490 has been merged.

PR #490, opened 2019-06-16
Revert "Merge pull request #451 from DFE-Digital/1598-use-callbacks-to-add-or-remove-course-sites"

misaka:
This reverts commit cc1eadf5c099861f9ac2efe272fd6e742c2fc6bc, reversing
changes made to 07c22b2d36931e6a18fb56ba557c4e2ddc186aae.

A nasty, gnarly bug was discovered which hadn't been tested for. When using
callbacks and the built-in has-many-through mechanism it seems to be impossible,
in a practical sense, to modify existing sites statuses back to "running" when
being created instead of instantiating new ones. As we are likely to be
modifying the data model to remove the use of these statuses anyway, reverting
this and focusing on revising the data model seems like the best use of time.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


misaka:
Ya, I spent a good portion of Friday evening delving into some of the internals of `ActiveRecord` to come to this conclusion. I was a little surprised and disappointed that this kind of thing didn't work, though:

```ruby
  has_many(:active_site_statuses,
           -> { active },
           class_name: 'SiteStatus') do
    def build(attributes = {}, &block)
      existing_site_status = target.find { |ss| ss.site_id == attributes[:site_id] }
      existing_site_status || super
    end
  end
```

I think the best approach would be, if possible, to remove the use of the status column completely and just use presence to determine whether the site is new/running, however I'm not sure if we have a requirement to send discontinued/suspended site_statuses to UCAS or not.

timabell:
[Changing locations on courses creates duplicate site statuses](https://trello.com/c/XChPfqwl/1638-changing-locations-on-courses-creates-duplicate-site-statuses)

timabell:
LGTM, couldn't see any new usages that have crept in

PR #494, opened 2019-06-17
[1530] Move create logic to controller

davidgisbey:
### Context

The access request controller currently has the logic for adding the requester_id, request_date_utc and the status to the access_request object upon instantiation. 

- Move updating the above 3 attributes from the controller to the model.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/access_request_spec.rb
+++ b/spec/models/access_request_spec.rb
@@ -75,4 +75,32 @@ describe AccessRequest, type: :model do
       expect(AccessRequest.by_request_date.last).to eq access_request1
     end
   end
+
+  describe '#add_additonal_attributes' do
+    let(:user) { create(:user, organisations: [organisation]) }
+    let(:organisation) { build(:organisation) }
+    let(:access_request) {
+      build(:access_request,
+            organisation: user.organisations.first.name,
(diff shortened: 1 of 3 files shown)


davidgisbey:
Good point 👍

davidgisbey:
[[S] Move access request logic to model callbacks](https://trello.com/c/rcXML6pF/1530-s-move-access-request-logic-to-model-callbacks)

spec/models/access_request_spec.rb
```diff
@@ -75,4 +75,27 @@
...
+    }
+
+    before do
+      access_request.add_additonal_attributes(access_request.requester_email)
+    end
+
+    subject { access_request }
+
+    its(:requester)         { should eq user }
+    its(:request_date_utc)  { should be_within(1.second).of Time.now.utc }
```
benilovj: I'd use Timecop around this whole describe block to pin the time to a particular timestamp; then you should be able to do an exact match.

davidgisbey: Thanks. Have added it 👍

PR #495, opened 2019-06-18
[1624] Add deserializable site class

tvararu:
### Context

json-api will auto-generate one for us, but it's better to use an explicit one instead because that way we have more finely grained control over what attributes get exposed.

Add DeserializableSite and use it in the site_controller.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/sites_controller.rb
+++ b/app/controllers/api/v2/sites_controller.rb
@@ -45,7 +45,7 @@ module API
 
       def site_params
         params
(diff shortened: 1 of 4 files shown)


tvararu:
[Display validations when saving enrichments](https://trello.com/c/zqMPpbp7/1624-display-validations-when-saving-enrichments)

PR #497, opened 2019-06-18
[1624] Remove latest_enrichment prefix on enrichment errors

tvararu:
### Context

These were initially added with the idea that without them, the frontend will not be able to differentiate which errors belong to the course and which errors belong to the course enrichment.

In practice, they actually end up confusing json-api, which will not automatically assign `source: { pointer }`s which are used to then associate validation errors to the appropriate field.

Because the frontend doesn't need to know about enrichments, it's better if we simply remove this prefix.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -287,14 +287,11 @@ private
 
   def add_enrichment_errors(enrichment)
     enrichment.errors.messages.map do |field, _error|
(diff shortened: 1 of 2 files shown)


tvararu:
[Display validations when saving enrichments](https://trello.com/c/zqMPpbp7/1624-display-validations-when-saving-enrichments)

PR #506, opened 2019-06-19
[1640] Add recruitment cycle to courses

dankmitchell:
### Context
Add the 2019 recruitment cycle to ALL courses

- Setup association between course and recruitment cycle
- Add data migration to add 2019 recruitment cycle to all courses

### Guidance to review
Run `bundle exec rake db:migrate:with_data`

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -20,6 +20,7 @@
 #  created_at              :datetime         not null
 #  updated_at              :datetime         not null
 #  changed_at              :datetime         not null
+#  recruitment_cycle_id    :integer
 #
 
 class Course < ApplicationRecord
@@ -69,6 +70,7 @@ class Course < ApplicationRecord
 
   belongs_to :provider
   belongs_to :accrediting_provider, class_name: 'Provider', optional: true
+  belongs_to :recruitment_cycle
   has_many :course_subjects
   has_many :subjects, through: :course_subjects
   has_many :site_statuses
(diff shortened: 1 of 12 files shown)


dankmitchell:
[[M] Backend - Add recruitment cycles to courses](https://trello.com/c/z4vIdLCw/1640-m-backend-add-recruitment-cycles-to-courses)

app/models/course.rb
```diff
@@ -20,6 +20,7 @@
 #  created_at              :datetime         not null
 #  updated_at              :datetime         not null
 #  changed_at              :datetime         not null
+#  recruitment_cycle_id    :integer
```
defong: nullable?

misaka: Ya, this will have to be nullable for the first migration at least, unless we give it a default value (in the SQL), which is probably not the best. Otherwise:

1. Add `course.recruitment_cycle_id`, nullable
2. Populate `recruitment_cycle_id` for all the courses with a data migration
3. Make `course.recruitment_cycle_id` not nullable.

defong: aint that the default?

`Add2019RecruitmentCycleToCourses`

dankmitchell: We can't make it not null until we've ran the data migration so I suggest doing a follow up PR for this.

defong: Guess I got baited to reject this, will fall for it everytime.

PR #509, opened 2019-06-19
[1657] Improved string representation of models

benilovj:
### Context
The support tooling displays sites and subjects to the support agents in various places.

- remove `SiteStatus#description` (no longer used)
- add `Site#to_s` and `Subject#to_s`

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -283,6 +283,10 @@ class Course < ApplicationRecord
     dfe_subjects&.first&.scholarship_amount
   end
 
+  def to_s
+    "#{name} (#{course_code})"
+  end
+
 private
 
   def add_enrichment_errors(enrichment)
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -46,7 +46,7 @@ class Site < ApplicationRecord
     self.code ||= pick_next_available_code(available_codes: provider&.unassigned_site_codes)
   end
 
-  def description
+  def to_s
     "#{location_name} (code: #{code})"
   end
 
--- a/spec/models/subject_spec.rb
+++ b/spec/models/subject_spec.rb
@@ -10,7 +10,8 @@
 require 'rails_helper'
 
 RSpec.describe Subject, type: :model do
-  subject { create(:subject) }
+  subject { create(:subject, subject_name: 'Mathematics') }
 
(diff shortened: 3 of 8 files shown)


app/models/course.rb
```diff
@@ -283,6 +283,10 @@ def scholarship_amount
     dfe_subjects&.first&.scholarship_amount
   end
 
+  def to_s
+    "#{name} (#{course_code})"
```
defong: Just wondering provider code as well since that's the unique combo?

app/models/site.rb
```diff
@@ -46,7 +46,7 @@ def assign_code
     self.code ||= pick_next_available_code(available_codes: provider&.unassigned_site_codes)
   end
 
-  def description
+  def to_s
     "#{location_name} (code: #{code})"
```
defong: Keep it the same as course?

Ie 

Main (-) 

Vs 

Main (code: -) 

?

spec/models/subject_spec.rb
```diff
@@ -10,7 +10,8 @@
 require 'rails_helper'
 
 RSpec.describe Subject, type: :model do
-  subject { create(:subject) }
+  subject { create(:subject, subject_name: 'Mathematics') }
 
   it { should have_many(:courses).through(:course_subjects) }
+  its(:to_s) { should eq('Mathematics') }
```
defong: Just wondering humanize it?

app/models/course.rb
benilovj: I'm just following the pattern used in Publish:

image

Also it'll just look redundant and noisy in mcb:

image

So I'd prefer to leave as is.

spec/models/subject_spec.rb
benilovj: Que?

app/models/site.rb
benilovj: I think I started with that but then `Foo (code: 1)` looked really weird when the code is 0-9: `Foo (1)'

spec/models/subject_spec.rb
defong: C'est ce que je voulais dire ...

https://apidock.com/rails/String/humanize

PR #510, opened 2019-06-19
[1657] Course defaults and aligned data

benilovj:
### Context
When new courses are created, we need to ensure that the correct downstream data is derived. In future we should move these attributes to the API V1 serialisers and drop the underlying columns.

- add a default for `Course#modular`
- set `Course#profpost_flag` when `Course#qualification` is set

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/concerns/with_qualifications.rb
+++ b/app/models/concerns/with_qualifications.rb
@@ -39,13 +39,7 @@ module WithQualifications
     # subjects this course was tagged to.
     #
     # Defined here: https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Domain/Models/CourseQualification.cs
-    enum qualification: %i[
-      qts
-      pgce_with_qts
-      pgde_with_qts
-      pgce
-      pgde
-    ]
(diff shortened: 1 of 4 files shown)


app/models/concerns/with_qualifications.rb
```diff
@@ -69,5 +63,10 @@ def qualifications
     def qualifications_description
       qualifications.map(&:upcase).sort.join(" with ")
     end
+
+    def qualification=(value)
+      super(value)
+      self.profpost_flag = qts? ? :recommendation_for_qts : :postgraduate
```
defong: `qts?` that bit looks really magically guess it comes from 

```ruby
enum qualification: %i[qts pgce_with_qts pgde_with_qts pgce pgde]
```

benilovj: yup that's right

benilovj:
> Those `self` im sure those rails dev will think you a c# dev... must of being wearing wrong onesie pajamas

I **think** you need the `slef`, otherwise Ruby will define a local variable instead of calling the setter.

PR #518, opened 2019-06-21
1651 mcb revoke god user

timabell:
### Context

We have users with access to many orgs, when they leave it takes too long to loop through removing access. We can't just kill their sign-in acc because they might use it for other things, we need to control our own access.

cri params can't be optional, but options can so the provider code is now a named parameter - `-p XYZ` so the usage of the existing `users revoke` has changed from `users revoke email@address XY1` to `users revoke email@address -p XY1`

Make provider code for revoke optional and named. If not supplied all access will be removed.

### Guidance to review

```
tim@fox:~/repo/dfe/manage/backend(1651-mcb-revoke-god-user)
$ be bin/mcb user revoke x@y      
User:
+------------------------+-----+
| id                     | 114 |
| email                  | x@y |
| first_name             | x   |
| last_name              | y   |
| first_login_date_utc   |     |
| last_login_date_utc    |     |
| sign_in_user_id        |     |
| welcome_email_date_utc |     |
| invite_date_utc        |     |
| accept_terms_date_utc  |     |
| state                  | new |
+------------------------+-----+

Has access to providers:
+-----+--------------+------+
|  id |     name     | code |
+-----+--------------+------+
| 125 | ACME SCITT 1 | A1   |
| 127 | ACME SCITT 3 | A3   |
+-----+--------------+------+

You're revoking access for x@y to:
Organisation access being revoked:
+----+-------+
| id |  name |
+----+-------+
| 92 | ACME1 |
| 94 | ACME3 |
+----+-------+
Provider access being revoked:
+-----+--------------+------+
|  id |     name     | code |
+-----+--------------+------+
| 125 | ACME SCITT 1 | A1   |
| 127 | ACME SCITT 3 | A3   |
+-----+--------------+------+
Agree?  y
tim@fox:~/repo/dfe/manage/backend(1651-mcb-revoke-god-user)
$ be bin/mcb user show x@y  
User:
+------------------------+-----+
| id                     | 114 |
| email                  | x@y |
| first_name             | x   |
| last_name              | y   |
| first_login_date_utc   |     |
| last_login_date_utc    |     |
| sign_in_user_id        |     |
| welcome_email_date_utc |     |
| invite_date_utc        |     |
| accept_terms_date_utc  |     |
| state                  | new |
+------------------------+-----+

Has access to providers:
-none-
```

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -49,6 +49,11 @@ class User < ApplicationRecord
     "#{first_name} #{last_name} <#{email}>"
   end
 
+  # accepts array or single organisation
+  def remove_access_to(organisations_to_remove)
+    self.organisations = self.organisations - [*organisations_to_remove]
+  end
+
 private
 
   def email_is_lowercase
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -51,7 +51,7 @@ class User < ApplicationRecord
 
   # accepts array or single organisation
   def remove_access_to(organisations_to_remove)
-    self.organisations = self.organisations - [*organisations_to_remove]
+    self.organisations = self.organisations - Array(organisations_to_remove)
   end
 
 private
--- a/lib/mcb/commands/users/revoke.rb
+++ b/lib/mcb/commands/users/revoke.rb
@@ -6,16 +6,11 @@ usage 'revoke <user_id/email/sign_in_user_id> [--provider-code <provider_code>]'
 
 run do |opts, args, _cmd|
   MCB.init_rails(opts)
-
   cli = HighLine.new
 
-  provider = nil
-  if opts[:provider_code]
-    provider = Provider.find_by!(provider_code: opts[:provider_code])
-  end
-
-
+  provider = Provider.find_by(provider_code: opts[:provider_code])
   user = MCB.find_user_by_identifier args[:id_or_email_or_sign_in_id]
+
   if user == nil
     puts "#{args[:id_or_email_or_sign_in_id]} does not exist."
   else
--- a/lib/mcb/commands/users/revoke.rb
+++ b/lib/mcb/commands/users/revoke.rb
@@ -1,13 +1,19 @@
 summary 'Remove a user from an organisation/provider in the DB.'
 param :id_or_email_or_sign_in_id
-param :provider_code, transform: ->(code) { code.upcase }
-usage 'revoke <user_id/email/sign_in_user_id> <provider_code>'
+option :p, 'provider_code', 'provider code',
+       argument: :optional, transform: ->(code) { code.upcase }
+usage 'revoke <user_id/email/sign_in_user_id> [--provider-code <provider_code>]'
 
 run do |opts, args, _cmd|
   MCB.init_rails(opts)
 
   cli = HighLine.new
-  provider = Provider.find_by!(provider_code: args[:provider_code])
+
+  provider = nil
+  if opts[:provider_code]
+    provider = Provider.find_by!(provider_code: opts[:provider_code])
+  end
+
 
   user = MCB.find_user_by_identifier args[:id_or_email_or_sign_in_id]
   if user == nil
--- a/lib/mcb/revoke_access_wizard.rb
+++ b/lib/mcb/revoke_access_wizard.rb
@@ -7,25 +7,35 @@ module MCB
     end
 
     def run
-      fetch_organisation
+      fetch_organisations
       puts MCB::Render::ActiveRecord.user @user
       confirm_and_remove_user_from_organisation
     end
 
   private
 
-    def fetch_organisation
-      @organisation = @provider.organisations.first # a provider should only ever be associated with one organisation
+    def fetch_organisations
+      @organisations = if @provider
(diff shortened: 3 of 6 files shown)


timabell:
[Un-deify old team-members in production](https://trello.com/c/C2qN5dN6/1381-un-deify-old-team-members-in-production)

lib/mcb/revoke_access_wizard.rb
```diff
@@ -7,25 +7,35 @@ def initialize(cli, user, provider)
...
-        puts MCB::Render::ActiveRecord.providers_table @organisation.providers, name: "Provider access being revoked"
-        @organisation.users.delete(@user) if @cli.agree("Agree?  ")
+        puts "You're revoking access for #{@user.email} to:"
+        puts MCB::Render::ActiveRecord.organisations_table @organisations, name: "Organisation access being revoked"
+        puts MCB::Render::ActiveRecord.providers_table @organisations.map(&:providers).flatten, name: "Provider access being revoked"
+        unless @cli.agree("Agree?  ")
+          return
+        end
+
+        @user.organisations = @user.organisations - @organisations
```
benilovj: I think this should live on the `User` model.

timabell:
will there be extra brownies?

benilovj:
> will there be extra brownies?

Yes... fine you've brow-beaten me into it! It'll be there on Monday...

timabell:
woohoo! @defong would be proud

defong:
> woohoo! @defong would be proud

Harassing a rails dev is always fun, they not quite sharp. ;)

lib/mcb/revoke_access_wizard.rb
```diff
@@ -7,25 +7,35 @@ def initialize(cli, user, provider)
...
     def confirm_and_remove_user_from_organisation
-      if @user.in?(@organisation.users)
+      if (@user.organisations & @organisations).any?
         puts "\n"
-        puts "You're revoking access for #{@user.email} to organisation #{@organisation.name}."
-        puts MCB::Render::ActiveRecord.providers_table @organisation.providers, name: "Provider access being revoked"
-        @organisation.users.delete(@user) if @cli.agree("Agree?  ")
+        puts "You're revoking access for #{@user.email} to:"
+        puts MCB::Render::ActiveRecord.organisations_table @organisations, name: "Organisation access being revoked"
+        puts MCB::Render::ActiveRecord.providers_table @organisations.map(&:providers).flatten, name: "Provider access being revoked"
```
misaka: Don't think we need the `.flatten` here.

```suggestion
        puts MCB::Render::ActiveRecord.providers_table @organisations.map(&:providers), name: "Provider access being revoked"
```

timabell:
```
Failures:

  1) mcb users revoke one provider when the user exists and has access to the provider revokes organisation membership to that user
     Failure/Error: table.pack(max_table_width: nil),

     NoMethodError:
       undefined method `id' for #<ActiveRecord::Associations::CollectionProxy []>
       Did you mean?  ids
```

lib/mcb/revoke_access_wizard.rb
timabell: moved to model, see newest commit

timabell:
All comments addressed, @benilovj if you want to :+1: when you get a chance I'll merge it

app/models/user.rb
```diff
@@ -49,6 +49,11 @@ def to_s
     "#{first_name} #{last_name} <#{email}>"
   end
 
+  # accepts array or single organisation
+  def remove_access_to(organisations_to_remove)
+    self.organisations = self.organisations - [*organisations_to_remove]
```
misaka: This is just golfing, but if you're trying to ensure that `organisations_to_remove` is an Array here, I think `Array(organisations_to_remove)` is more idiomatic.

lib/mcb/commands/users/revoke.rb
```diff
@@ -1,13 +1,19 @@
...
 run do |opts, args, _cmd|
   MCB.init_rails(opts)
 
   cli = HighLine.new
-  provider = Provider.find_by!(provider_code: args[:provider_code])
+
+  provider = nil
+  if opts[:provider_code]
+    provider = Provider.find_by!(provider_code: opts[:provider_code])
+  end
```
misaka: More golfing ... I think that this would work as well, and be shorter:

```
    provider = Provider.find_by(provider_code: opts[:provider_code])
```

Basically, just allow `find_by` to return a nil, which is what you'll get if you run `find_by(nil)`. Embrace the `nil`!!

app/models/user.rb
timabell: done, I saw both and picked one at random :-) I think yours is clearer in intent now I think about it

lib/mcb/commands/users/revoke.rb
timabell: much tidier. I hadn't understood that it was just the bang making it go bang!

all done, rebased & pushed, cheers for those

PR #519, opened 2019-06-21
[1624] Match word count validator to frontend

tvararu:
### Context

The word count component in the frontend isn't doing anything as sophisticated as the word count library, which causes a mismatch between the frontend and the backend when submitting e.g. markdown tokens.

This changes to the same [RegExp used in the frontend](https://github.com/alphagov/govuk-frontend/blob/063cd8e2470b62b824c6e50ca66342ac7a95d2d8/src/govuk/components/character-count/character-count.js#L80), and removes the words_counted library.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/validators/words_count_validator.rb
+++ b/app/validators/words_count_validator.rb
@@ -2,7 +2,7 @@ class WordsCountValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.blank?
 
-    if WordsCounted.count(value).token_count > options[:maximum]
+    if value.scan(/\S+/).size > options[:maximum]
       record.errors[attribute] << (options[:message] || "^Reduce the word count for #{attribute.to_s.humanize(capitalize: false)}")
     end
   end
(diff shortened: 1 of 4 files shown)


tvararu:
[[S] Validate word count correctly on enrichments when words are split by newlines](https://trello.com/c/j92XEVGc/1635-s-validate-word-count-correctly-on-enrichments-when-words-are-split-by-newlines)

app/validators/words_count_validator.rb
```diff
@@ -2,7 +2,7 @@ class WordsCountValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.blank?
 
-    if WordsCounted.count(value).token_count > options[:maximum]
+    if value.scan(/\S+/).size > options[:maximum]
```
fofr: Wonder if this deserves a comment and a link to the frontend validation.

PR #524, opened 2019-06-21
[1645] Return dynamic recruitment year for V1 courses

dankmitchell:
### Context
RE: The `courses` endpoint on the V1 API. 
At present, we return a hardcoded year stored in the course model but we now have a recruitment cycle model.

Replace hardcoded value with the dynamic value from `course.recruitment_cycle`

### Guidance to review
`/api/v1/2019/courses`

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/serializers/course_serializer.rb
+++ b/app/serializers/course_serializer.rb
@@ -74,6 +74,6 @@ class CourseSerializer < ActiveModel::Serializer
   end
 
   def recruitment_cycle
-    object.recruitment_cycle_year
+    object.recruitment_cycle.year
   end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -24,8 +24,6 @@
 #
 
 class Course < ApplicationRecord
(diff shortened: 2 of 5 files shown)


dankmitchell:
[[S] Switch apiv1 to return the actual recruitment cycle year in JSON](https://trello.com/c/rMhfhfiY/1645-s-switch-apiv1-to-return-the-actual-recruitment-cycle-year-in-json)

app/serializers/course_serializer.rb
```diff
@@ -74,6 +74,6 @@ def copy_form_required
   end
 
   def recruitment_cycle
-    object.recruitment_cycle_year
+    object.recruitment_cycle.year
```
misaka: Can we add a quick spec for this to `spec/serializers/course_serializer_spec.rb`?

dankmitchell: Done

PR #525, opened 2019-06-21
[1658] Fixed issues found from testing

defong:
### Context
c# vs rails pub testing

Campus address needs commas instead of new line, and space separated postcode

Accrediting Provider Enrichment descriptions always return empty string

### Guidance to review

https://github.com/DFE-Digital/manage-courses-api/blob/4b851b14c504dca8d11e77b8dcce9ac514612d46/src/ManageCourses.Api/Mapping/CourseMapper.cs#L251

https://github.com/DFE-Digital/manage-courses-api/blob/4b851b14c504dca8d11e77b8dcce9ac514612d46/src/ManageCourses.Api/Mapping/CourseMapper.cs#L229

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/serializers/search_and_compare/course_serializer.rb
+++ b/app/serializers/search_and_compare/course_serializer.rb
@@ -139,7 +139,7 @@ module SearchAndCompare
        },
        {
          Name: "about this training provider accrediting",
-         Text: object.accrediting_provider_description
+         Text: object.accrediting_provider_description.to_s
        },
        {
          Name: "training with disabilities",
@@ -280,11 +280,15 @@ module SearchAndCompare
       [address1, address2, address3, address4, postcode].reject(&:blank?).join("\n")
     end
 
+    def campuses_full_address(address1:, address2:, address3:, address4:, postcode:)
+      [address1, address2, address3, address4].reject(&:blank?).join(", ") + (postcode.present? ? ' ' + postcode : '')
+    end
+
     def campuses
       object.site_statuses.findable.map do |site_status|
         raw_address = { address1: site_status.site.address1, address2: site_status.site.address2, address3: site_status.site.address3, address4: site_status.site.address4, postcode: site_status.site.postcode }
 
-        address = full_address(raw_address)
+        address = campuses_full_address(raw_address)
 
         {
           **default_campus_value,
--- a/spec/serializers/search_and_compare/course_serializer_spec.rb
+++ b/spec/serializers/search_and_compare/course_serializer_spec.rb
@@ -39,7 +39,6 @@ describe SearchAndCompare::CourseSerializer do
           # that they're missing the following bits of data.
           c.site_statuses.each do |site_status|
             site_status.site.address2 = ''
-            site_status.site.address3 = ''
             site_status.site.address4 = ''
             site_status.site.postcode = ''
             site_status.site.save validate: false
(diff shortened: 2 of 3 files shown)


defong:
[[M] Sync course to search-and-compare serializer testings](https://trello.com/c/aJhqeHqF/1658-m-sync-course-to-search-and-compare-serializer-testings)

app/serializers/search_and_compare/course_serializer.rb
```diff
@@ -139,7 +139,7 @@ def description_sections
        },
        {
          Name: "about this training provider accrediting",
-         Text: object.accrediting_provider_description
+         Text: object.accrediting_provider_description.present? ? object.accrediting_provider_description : ''
```
misaka: Also an option to avoid a ternary:

```suggestion
         Text: object.accrediting_provider_description.to_s
```

app/serializers/search_and_compare/course_serializer.rb
```diff
@@ -280,11 +280,15 @@ def full_address(address1:, address2:, address3:, address4:, postcode:)
       [address1, address2, address3, address4, postcode].reject(&:blank?).join("\n")
     end
 
+    def campuses_full_address(address1:, address2:, address3:, address4:, postcode:)
+      [address1, address2, address3, address4].reject(&:blank?).join(", ") + (postcode.present? ? ' ' + postcode : '')
```
misaka: Just to check, do we really want the postcode tacked onto the end without a comma?

spec/serializers/search_and_compare/course_serializer_spec.rb
```diff
@@ -237,9 +237,8 @@
                 site_status.site.address1,
                 site_status.site.address2,
                 site_status.site.address3,
-                site_status.site.address4,
-                site_status.site.postcode
-              ].reject(&:blank?).join('\n')
+                site_status.site.address4
+              ].reject(&:blank?).join('\n') + (site_status.site.postcode.present? ? ' ' + site_status.site.postcode : '')
```
misaka: Is this right? Doesn't the change above join with a comma (`,`)? So why is this test passing looking for newlines?

app/serializers/search_and_compare/course_serializer.rb
defong:
```c#
        private string MapAddress(Site school)
        {
            var addressFragments = new List<string>{
                school.Address1,
                school.Address2,
                school.Address3,
                school.Address4
            }.Where(x => !string.IsNullOrWhiteSpace(x));

            var postCode = school.Postcode ?? "";

            return addressFragments.Any()
                ? String.Join(", ", addressFragments) + " " + postCode
                : postCode;
}
```

it then goes thro the scary geocoder

defong: we should keep it as in as geocoder works well enough with it as well as it keeps  cost and effort down, the lat/lang is keyed with the address as far as i remember on search and compare.

Maybe another PR once we get rails working without dependencies on its c# counterpart

spec/serializers/search_and_compare/course_serializer_spec.rb
defong: fixed, good catch at the time I did noticed test was passing without any changes, as the one and only json we pluck out has an invalid site saved.

PR #526, opened 2019-06-21
generate spreadsheet for allocations report

misaka:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/models/course.rb
```diff
@@ -172,6 +174,45 @@ def allocations_report_fields
...
+  # Outputs an allocations XLSX with a randomly generated suffix to `public/`.
+  # The first argument is an optional prefix for the filename to make it
+  # easier to differentiate among ~200 other similar XLSX files. Defaults to
+  # the provider.provider_code.
+  # The filename will have a UUID prefix so that it can be hosted
+  # and linked to from an Azure bucket, but that the names can't be
+  # simply guessed.
+  # Example usage:
+  #    $ bin/rails c
+  #    > Course.save_xlsx(Course.first(20))
```
```suggestion
  #    > Course.save_allocations_report(Course.first(20))
```

app/models/course.rb
```diff
@@ -172,6 +174,45 @@ def allocations_report_fields
...
+  def self.save_allocations_report(file_name_prefix = all.first.provider.provider_code)
+    file_data = all.to_xlsx
+
+    file_name_suffix = SecureRandom.uuid
```
defong: timestamp instead?

upload-allocations.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "Getting access key"
+access_key=`az storage account keys list --account-name sadfebatallocations --subscription "DFE BAT Development" --query "[?keyName == 'key1'].value | [0]"`;
+
+for f in $files
+do
+  echo "Uploading $f"
+
+  filename=$(echo $f | sed 's/public\///g')
+
+  az storage blob upload --container-name find-allocations --name "$filename" --file "$f" --connection-string "DefaultEndpointsProtocol=https;AccountName=sadfebatallocations;AccountKey=$access_key;EndpointSuffix=core.windows.net" > /dev/null
```
defong: you sure?

`AccountName=sad...`

app/models/course.rb
```diff
@@ -172,6 +174,45 @@ def allocations_report_fields
...
+  def spreadsheet_columns
+    [
+      ['Academic Year', '2020/21'],
+      ['Requested By (Name)', accrediting_provider&.provider_name],
+      ['Requested by (UKPRN)', ''],
+      ['Partner ITT Provider (Name)', provider.provider_name],
+      ['Partner ITT provider (UKPRN)', provider.organisations.first&.school_nctl_organisation&.urn],
+      ['Allocation Subject', allocation_subjects.join(' | ')],
+      ['Route', program_type],
+      ['Course Aim', qualification],
```
tvararu: Please double check these for sanity, they are massaged from the `allocations_report_fields` above

app/models/course.rb
```diff
@@ -172,6 +174,45 @@ def allocations_report_fields
...
+      ['Other Institution', '']
+    ]
+  end
+
+  # Outputs an allocations XLSX with a randomly generated suffix to `public/`.
+  # The first argument is an optional prefix for the filename to make it
+  # easier to differentiate among ~200 other similar XLSX files. Defaults to
+  # the provider.provider_code.
+  # The filename will have a UUID prefix so that it can be hosted
+  # and linked to from an Azure bucket, but that the names can't be
```
defong: are you implying its a public visible bucket?

tvararu: It's set to the `blob` setting which means yes, individual files are accessible using the link, but it's not enumerable / you can't access other files easily.

app/models/course.rb
tvararu: I will never write comments again!

app/models/course.rb
tvararu: Timestamp could be ~the same for an export of 200 or so spreadsheets, reducing its effectiveness in making the URL harder to guess 🤔

app/models/course.rb
defong: so why make it public but obfuscated?

benilovj: we will distribute publicly available individual deep links but don't want to make it feasible to guess other file links if you don't have it.

defong: maybe? https://blogs.msdn.microsoft.com/jpsanders/2017/10/12/easily-create-a-sas-to-download-a-file-from-azure-storage-using-azure-storage-explorer/

benilovj: > maybe? https://blogs.msdn.microsoft.com/jpsanders/2017/10/12/easily-create-a-sas-to-download-a-file-from-azure-storage-using-azure-storage-explorer/

Yes that's definitely an option, although probably more work than above implementation, given we need to do this for 250-1000 files.

PR #529, opened 2019-06-21
[AL-6] Generate a separate report line for each course subjects

misaka:
*!!! This PR depends on #528 which needs to be merged before this one is !!!*

When generating the subject allocation request report, we need to output one line per subject being requested.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


misaka:
Argh, merged into the wrong branch. 😡

PR #532, opened 2019-06-22
[1670] Introduce enum for Provider#accrediting_provider

benilovj:
### Context
Providers are either accredited bodies or they're not. There is information stored on the record to interrogate this.

Replace an existing helper method with an enum, in order to get the scope and both possible values.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -41,6 +41,11 @@ class Provider < ApplicationRecord
     invalid_value: "0", # there is only one of these in the data
   }
 
+  enum accrediting_provider: {
+    accredited_body: 'Y',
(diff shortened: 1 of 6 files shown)


dankmitchell:
Will this impact the frontend in anyway?

benilovj:
> Will this impact the frontend in any way?

There should be no impact at all. The serialisers for the V1 API have been updated to keep the interface the V1 interface the same, and V2 is similarly unaffected because the `accredited_body?` method now comes courtesy of Rails magic from the enum definition.

PR #536, opened 2019-06-24
[1656] Always run enrichment save validations

tvararu:
### Context

Currently, it is possible to save enrichments even though they do not pass validations. This is incorrect behaviour when compared to the C# app.

The changes the `:save` scoped validations to always run whenever we validate enrichments.

### Guidance to review

Review commit by commit.

TODO:

- [x] Clean up `course#saveable?` method, probably can be replaced with `valid?` everywhere
- [x] Uncomment commented specs
- [x] Fix the update hack
- [x] Remove `latest_first` invocations in the rest of the `course` model
- [x] Squash commits that don't add much value

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -84,7 +84,11 @@ class Course < ApplicationRecord
            primary_key: :course_code,
            class_name: 'CourseEnrichment' do
     def find_or_initialize_draft
-      latest_draft_enrichment = latest_first.draft.first
+      # This is a ruby search as opposed to an AR search, because calling `draft`
+      # will return a new instance of a CourseEnrichment object which is different
+      # to the ones in the cached `enrichments` association. This makes checking
+      # for validations later down non-trivial.
+      latest_draft_enrichment = select(&:draft?).last
 
       if latest_draft_enrichment.present?
         latest_draft_enrichment
@@ -323,7 +327,7 @@ private
   end
 
   def validate_enrichment(validation_scope)
-    latest_enrichment = enrichments.latest_first.first
+    latest_enrichment = enrichments.select(&:draft?).last
     return unless latest_enrichment.present?
 
     latest_enrichment.valid? validation_scope
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -123,7 +123,7 @@ class Course < ApplicationRecord
 
   validates :enrichments, presence: true, on: :publish
   validate :validate_enrichment_publishable, on: :publish
-  validate :validate_enrichment_saveable, on: :save
+  validate :validate_enrichment
 
   def accrediting_provider_description
     return nil if accrediting_provider.blank?
@@ -146,10 +146,6 @@ class Course < ApplicationRecord
 
   def publishable?
     valid? :publish
-  end
-
-  def saveable?
-    valid? :save
   end
 
   def findable?
@@ -326,7 +322,7 @@ private
     end
   end
 
-  def validate_enrichment(validation_scope)
+  def validate_enrichment(validation_scope = nil)
     latest_enrichment = enrichments.select(&:draft?).last
     return unless latest_enrichment.present?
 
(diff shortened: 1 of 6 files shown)


tvararu:
[[?] Enrichments get saved regardless of validation errors](https://trello.com/c/VT6BCuiV/1656-enrichments-get-saved-regardless-of-validation-errors)

app/models/course_enrichment.rb
```diff
@@ -42,27 +42,50 @@ class CourseEnrichment < ApplicationRecord
...
+  validates :qualifications, presence: true, on: :publish
+  validates :qualifications, words_count: { maximum: 100 }
 
   validates :salary_details, presence: true, on: :publish, unless: :is_fee_based?
-  validates :salary_details, words_count: { maximum: 250 }, on: %i[save publish], unless: :is_fee_based?
+  validates :salary_details, words_count: { maximum: 250 }, unless: :is_fee_based?
 
   def is_fee_based?
-    course.is_fee_based?
+    course&.is_fee_based?
```
defong: Can it ever be nil?

tvararu: I couldn't figure out how to get model tests to pass without this, because it is indeed nil...

defong: the test setup is wrong then... missing the course on the course enrichment

tvararu: Investigated together, and it's still not clear why this is happening, but it's probably because of the `belongs_to` relationship between a `CourseEnrichment` and `Course`. The `'belongs to a course'` spec fails because the `expect(subject).to belong_to(:course)` assertion seems to remove the course association, which leads to `undefined method `is_fee_based?' for nil:NilClass`.

spec/models/course_enrichment_spec.rb
```diff
@@ -210,9 +210,9 @@
 
     context 'fee based course' do
       it { should validate_presence_of(:fee_uk_eu).on(:publish) }
-      # it { should validate_presence_of(:about_course).on(:publish) }
+      it { should validate_presence_of(:about_course).on(:publish) }
```
tvararu: @misaka were these commented out because they should be removed? I uncommented them because they pass, but they seem to duplicate part of the above tests.

misaka: Ya, I commented them out just because they duplicate. Wasn't sure about their value but am happy to leave them if you think they have value.

tvararu: No, let's remove them then.

app/models/course.rb
```diff
@@ -79,12 +79,19 @@ class Course < ApplicationRecord
            through: :site_statuses
 
   has_many :enrichments,
-           ->(course) { where(provider_code: course.provider.provider_code) },
+           ->(course) {
+             where(provider_code: course.provider.provider_code)
+               .order(created_at: :asc, id: :asc)
+           },
```
misaka: I'm not sure about ordering in the association here. Aside from the potentially trivial performance hit, this could lead to surprising results if you try to order the association. For example, suppose we wanted to grab the first draft (thinking about the code we already have), you might try this to keep things fast:

```
enrichments.order(:status).first
```

This wouldn't work as expected since AR merges `order` statements and `order(:status)` would be the primary ordering criteria and `status` would only be used where `created_at` matched. Since this will never happen, using a subsequent `order` on this association would never work.

Can we create an association that orders `enrichments`? e.g.:

```
  has_many :enrichments,
           ->(course) { where(provider_code: course.provider.provider_code) },
           foreign_key: :ucas_course_code,
           primary_key: :course_code,
           class_name: 'CourseEnrichment' do
    ...
  end

  def ordered_enrichments
    enrichments.order(created_at: :asc, id: :asc)
  end
```

tvararu: Hmm, I think the example is flawed but basis of the argument I agree with. I'll give this a go :+1:

PR #548, opened 2019-06-26
[1687] 'mcb courses edit' should pick the right course to edit

benilovj:
### Context
`mcb courses edit` picks the wrong course to edit when there are two courses with the same course code across two different providers.

Use the provider when filtering courses to edit.

### Guidance to review
😞 

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/lib/mcb/courses_editor_spec.rb
+++ b/spec/lib/mcb/courses_editor_spec.rb
@@ -258,6 +258,38 @@ describe MCB::CoursesEditor do
       end
     end
 
+    context 'when there are several courses with the same course code' do
+      let(:another_provider) { create(:provider) }
+      let!(:another_course_with_the_same_course_code) {
+        create(:course,
+               provider: another_provider,
+               course_code: course.course_code,
+               name: 'Another name')
+      }
+
+      it 'edits the course from the specified provider' do
+        expect { run_editor("edit title", "Mathematics", "exit") }.
+          to change { course.reload.name }.
+          from("Original name").to("Mathematics")
+      end
+    end
+
+    context "when trying to edit a course code that doesn't exist on this provider but exists on another one" do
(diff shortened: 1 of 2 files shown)


spec/lib/mcb/courses_editor_spec.rb
```diff
@@ -258,6 +258,22 @@ def run_editor(*input_cmds)
...
+
+      it 'edits the course from the specified provider' do
+        expect { run_editor("edit title", "Mathematics", "exit") }.
+          to change { course.reload.name }.
+          from("Original name").to("Mathematics")
+      end
+    end
+
     describe 'tries to edit a non-existent course' do
       let(:course_codes) { [course_code, "ABCD"] }
```
timabell: could extend this to test in case where `ABCD` does exist with another provider

benilovj: added and force-pushed

spec/lib/mcb/courses_editor_spec.rb
```diff
@@ -258,6 +258,38 @@ def run_editor(*input_cmds)
...
+      let!(:another_course_with_another_provider) {
+        create(:course,
+               provider: another_provider,
+               course_code: 'XYZ1',
+               name: 'Another name')
+      }
+      subject { described_class.new(provider: provider, course_codes: %w{XYZ1}, requester: requester) }
+
+      it 'raises an error' do
+        expect { subject }.to raise_error(ArgumentError, /Couldn't find course XYZ1/)
```
defong: is it a complete dead end if

```subject { described_class.new(provider: provider, course_codes: %w{XYZ1, I_EXIST}, requester: requester) }```

PR #550, opened 2019-06-26
[XXX] Bump govuk-lint to 3.11.4

benilovj:
### Context
Since the rubocop version bump, the Rails cops have been broken.

- bump the `govuk-lint` gem to pick up the re-introduction of the Rails cops
- auto-fix and manually fix flagged issues
- disable the `Rails/HasManyOrHasOneDependent` for now to get back to green

### Guidance to review
Is it right to disable the `Rails/HasManyOrHasOneDependent` cop?

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


dankmitchell:
I think we should add `dependant` to the necessary associations but perhaps another PR or commit if this is going to get too big...

benilovj:
> I think we should add `dependant` to the necessary associations but perhaps another PR or commit if this is going to get too big...

Agreed, let's discuss tomorrow in the office.

app/models/course.rb
```diff
@@ -82,6 +82,7 @@ class Course < ApplicationRecord
            ->(course) { where(provider_code: course.provider.provider_code) },
            foreign_key: :ucas_course_code,
            primary_key: :course_code,
+           inverse_of: 'course',
```
fofr: What does this do?

benilovj: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/InverseOf

PR #554, opened 2019-06-27
[1682] remove rails data:migrate from Dockerfile

misaka:
### Context

We've removed the data migrations and gem, but not it being run.

Remove 'rails data:migrate' from Dockerfile.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


misaka:
[Remove data migrations](https://trello.com/c/4haxChLF/1682-remove-data-migrations)

PR #556, opened 2019-06-27
[1690] Add word count validations for personal qualities and other requirements

fofr:
### Context

We didn’t have validations for word count for:

- personal qualities
- other qualifications

This means users could save long answers to these.

- Add missing validations
- Group validations by enrichment page, and order based on order they appear in form – this means the error messages shown to users will also appear in the correct order


fofr:
[Missing word count validations on personal qualities and other requirements](https://trello.com/c/er4YK4Si/1690-missing-word-count-validations-on-personal-qualities-and-other-requirements)

PR #560, opened 2019-06-28
[1685] Remove unnecessary "Enrichments is invalid" message

tvararu:
### Context

Because of the `has_many :enrichments` relationship, Rails will add a `Enrichments is invalid` message on the `Course` model when a new `CourseEnrichment` is created but fails to be saved due to validation issues. This behaviour can occur when users attempt to save a new enrichment (as opposed to updating an already draft one) which has errors.

Adds an `after_validation` to manually clear the unnecessary validation.

### Guidance to review

I can't seem to find a more idiomatic way of doing this.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -124,6 +124,8 @@ class Course < ApplicationRecord
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_enrichment
 
+  after_validation :remove_unnecessary_enrichments_validation_message
+
   def accrediting_provider_description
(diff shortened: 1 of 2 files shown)


tvararu:
[When a user tries to publish a course without any enrichment data the validation message is wrong](https://trello.com/c/yZcPwzdX/1685-when-a-user-tries-to-publish-a-course-without-any-enrichment-data-the-validation-message-is-wrong)

PR #549, opened 2019-06-26
[1679] The course enrichments can be 2019 or 2020 recruitment cycle

defong:
### Context
The course enrichment can be 2019 or 2020 recruitment cycle

### Changes proposed in this pull request
The course enrichment is now associated with a course, making the need for `ucas_course_code` & `provider_code` redundant as well as making it possible deterministic to get a 2019 or 2020 recruitment cycle for course enrichments.

### Guidance to review
`ucas_course_code` & `provider_code` needs to stay for backwards compatibility with its C# counterpart. 
`ucas_course_code` & `provider_code` original purpose was due to the course records being dropped and imported therefore the `course_id` was not usable.

### Checklist
- [x] update models
- [x] update the record creation
- [x] await #536 to complete

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #561, opened 2019-06-28
[1677] The provider enrichments 2019 vs 2020 recruitment cycle

defong:
### Context
The provider enrichment can be 2019 or 2020 recruitment cycle
### Guidance to review

The provider enrichment is now associated with a provider, making the need for provider_code redundant as well as making it possible deterministic to get a 2019 or 2020 recruitment cycle for provider enrichments.

### Checklist

There is no provider_enrichment creation...

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -51,14 +51,10 @@ class Provider < ApplicationRecord
 
   has_many :sites
   has_many :enrichments,
-           foreign_key: :provider_code,
-           primary_key: :provider_code,
            class_name: "ProviderEnrichment",
            inverse_of: 'provider'
   has_one :latest_published_enrichment,
           -> { published.latest_published_at },
-          foreign_key: :provider_code,
-          primary_key: :provider_code,
           class_name: "ProviderEnrichment",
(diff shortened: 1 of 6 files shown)


defong:
[2019 vs 2020 provider enrichments](https://trello.com/c/rrmGmfRm/1677-2019-vs-2020-provider-enrichments)

defong:
@misaka with this change we can start looking at remove `recruitment cycle` from `course` since this provider is the aggregate root to retrieve a course

defong:
not merging as #549 is giving me a headache

defong:
For product review, there should not be any rows, if there then `DON'T DEPLOY`

```
SELECT *
FROM   (SELECT *,
               (SELECT provider.id
                FROM   provider
                WHERE  provider_enrichment.provider_code =
                       provider.provider_code) AS
               provider_id
        FROM   provider_enrichment) AS pr
WHERE  provider_id IS NULL  
```

defong:
fyi @misaka

PR #564, opened 2019-07-01
[1657] mcb courses edit subjects

benilovj:
### Context

New course creation is currently being done through user support using `mcb` dev tooling. The tooling itself is sitting on a code branch that isn't the `master` branch and needs constant updating/rebasing.

The code is also missing automated tests, meaning it needs to be manually tested after every change.

Introduce editing subjects to `mcb courses edit` which is both useful for editing courses, as well as being a stepping stone towards the new course creation.

## Comments for reviewers

This PR is best reviewed commit by commit.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -24,20 +24,39 @@ module MCB
       puts "Editing #{course_codes.join(', ')}"
       print_at_most_two_courses
       until finished
-        choice = @cli.main_loop
+        choice = main_loop
 
-        if choice.start_with?("edit")
+        if choice.nil?
+          finished = true
+        elsif choice.start_with?("edit")
           attribute = choice.gsub("edit ", "").gsub(" ", "_").to_sym
           edit(attribute)
         elsif choice =~ /sync .* to Find/
           sync_courses_to_find
-        elsif choice == "exit"
-          finished = true
         end
       end
     end
 
   private
+
+    def main_loop
+      choices = [
+        "edit title",
+        "edit course code",
+        "edit maths",
+        "edit english",
+        "edit science",
+        "edit route",
+        "edit qualifications",
+        "edit study mode",
+        "edit accredited body",
+        "edit start date",
+        "edit application opening date",
+        "edit age range",
+        "sync course(s) to Find"
+      ]
+      @cli.ask_multiple_choice(prompt: "What would you like to edit?", choices: choices)
+    end
 
     def edit(logical_attribute)
       database_attribute = LOGICAL_NAME_TO_DATABASE_NAME_MAPPING[logical_attribute] || logical_attribute
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -21,9 +21,9 @@ module MCB
 
     def run
       finished = false
-      puts "Editing #{course_codes.join(', ')}"
-      print_at_most_two_courses
       until finished
+        puts "Editing #{course_codes.join(', ')}"
+        print_at_most_two_courses
         choice = main_loop
 
         if choice.nil?
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -28,6 +28,8 @@ module MCB
 
         if choice.nil?
           finished = true
+        elsif choice == 'edit subjects'
+          edit_subjects
         elsif choice.start_with?("edit")
           attribute = choice.gsub("edit ", "").gsub(" ", "_").to_sym
           edit(attribute)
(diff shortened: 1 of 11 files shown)


benilovj:
CodeClimate doesn't like my code ¯\_(ツ)_/¯
Seems reasonably understandable to me but If others feel passionately about it, I can change it.

lib/mcb/courses_editor.rb
```diff
@@ -21,24 +21,46 @@ def initialize(provider:, requester:, course_codes: [])
...
+        "edit route",
+        "edit qualifications",
+        "edit study mode",
+        "edit accredited body",
+        "edit start date",
+        "edit application opening date",
+        "edit age range",
+        "edit subjects",
+        "sync course(s) to Find"
+      ].grep_v(->(c) { c.in?(["edit subjects"]) && @courses.count != 1 })
```
timabell: I can see what this does (having read https://ruby-doc.org/core-2.5.1/Enumerable.html#method-i-grep_v ) but I'm not clear from the code why it's needed.

spec/lib/mcb/courses_editor_cli_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  def run_with_input_commands(*input_cmds)
+    stderr = nil
+    output = with_stubbed_stdout(stdin: input_cmds.join("\n"), stderr: stderr) do
+      yield
+    end
+    [output, stderr]
+  end
+
+  let(:provider) { nil } # not needed for these specs
+  subject { described_class.new(provider) }
```
timabell: why do these two lines have to be here?

spec/lib/mcb/courses_editor_spec.rb
```diff
@@ -207,6 +211,30 @@ def run_editor(*input_cmds)
...
+            to change { course.subjects.reload.sort_by(&:subject_name) }.
+            from([biology, secondary]).to([secondary])
+        end
+
+        context "when more than 1 course is being edited" do
+          let(:another_course) { create(:course, provider: provider) }
+          let(:course_codes) { [course_code, another_course.course_code] }
+
+          it "does not allow editing subjects" do
+            output, = run_editor("exit")
```
timabell: what's the comma for after `output`?

lib/mcb/courses_editor_cli.rb
```diff
@@ -105,7 +84,23 @@ def ask_course_code
       end
     end
 
-  private
+    def multiselect(initial_items:, possible_items:)
```
timabell: I think I agree with code climate on this one, it's a lot of nesting to comprehend in one go and I did get an "uhoh" feeling when I looked at it. Perhaps just factor out a method or two from the inner parts

lib/mcb/courses_editor.rb
```diff
@@ -21,24 +21,46 @@ def initialize(provider:, requester:, course_codes: [])
 
     def run
```
timabell: "Method run has a Cognitive Complexity of 6 (exceeds 5 allowed)."
I think this reads fine. Ignore.

spec/lib/mcb/courses_editor_cli_spec.rb
benilovj: They don't – I've condensed them into one.

spec/lib/mcb/courses_editor_spec.rb
benilovj: The `run_editor` method returns a list with two elements. Since I don't use the second element, I threw it away by initially having `output, _` (which in my opinion is a bit clearer) but Rubocop complained that it should be `output,`.

lib/mcb/courses_editor.rb
benilovj: I've spun out the `filter_single_course_options_if_necessary` which should hopefully make things clearer.

lib/mcb/courses_editor_cli.rb
benilovj: I've extracted a method but I'm not sure it's made it easier to read.

spec/lib/mcb/courses_editor_spec.rb
```diff
@@ -15,6 +15,9 @@ def run_editor(*input_cmds)
   let(:email) { 'user@education.gov.uk' }
   let(:provider) { create(:provider, provider_code: provider_code) }
   let(:accredited_body) { create(:provider, :accredited_body) }
+  let!(:mathematics) { create(:subject, subject_name: "Mathematics") }
```
defong: can we start seeding these into factory?

benilovj: yup good idea, see 86e8ef9

PR #567, opened 2019-07-02
[1657] mcb courses edit: recruitment cycle

benilovj:
### Context
We need to be able to create new courses against different course cycles. 

- Fix the `-c` flag for the V1 API mcb commands (it was broken when `rubocop-rails` was introduced)
- Print the recruitment cycle when printing the course
- Allow the courses editor to edit recruitment cycle (this will be reused as part of the new course wizard).

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/recruitment_cycle.rb
+++ b/app/models/recruitment_cycle.rb
@@ -15,4 +15,9 @@ class RecruitmentCycle < ApplicationRecord
   has_many :sites
 
   validates :year, presence: true
+
+  def to_s
+    following_year = Date.new(year.to_i, 1, 1) + 1.year
+    "#{year}/#{following_year.strftime('%y')}"
+  end
 end
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -354,7 +354,7 @@ module MCB
           CGI.unescape(opts[:'changed-since']),
           '%FT%T.%NZ'
         ) rescue nil
-        changed_since ||= DateTime.zone.parse(opts[:'changed-since']).utc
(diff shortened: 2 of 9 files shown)


app/models/recruitment_cycle.rb
```diff
@@ -15,4 +15,9 @@ class RecruitmentCycle < ApplicationRecord
   has_many :sites
 
   validates :year, presence: true
+
+  def to_s
+    following_year = Date.new(year.to_i, 1, 1) + 1.year
```
defong: why did we store year as string?

lib/mcb.rb
```diff
@@ -354,7 +354,7 @@ def process_opt_changed_since(opts, url)
           CGI.unescape(opts[:'changed-since']),
           '%FT%T.%NZ'
         ) rescue nil
-        changed_since ||= DateTime.zone.parse(opts[:'changed-since']).utc
+        changed_since ||= DateTime.parse(opts[:'changed-since']) # rubocop:disable Rails/TimeZone
```
defong: don't you need the enable bit afterwards?

app/models/recruitment_cycle.rb
benilovj: `¯\_(ツ)_/¯`

lib/mcb.rb
benilovj: [The docs](https://docs.rubocop.org/en/stable/configuration/#disabling-cops-within-source-code) seem to imply that for single line disabling, it's not necessary to re-enable explicitly.

PR #569, opened 2019-07-03
1489 Sync a single course to S&C from mc-be

davidgisbey:
### Context
 **Part 1** 
We want to do all the course publishing in Rails instead of C#. This PR focuses on syncing a single course to search and compare.

- Uses the new S&C service to hit S&C directly rather than via ManageAPI
- Adds validations to only sync valid courses (findable & a DfE subject)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -22,16 +22,12 @@ module API
       end
 
       def sync_with_search_and_compare
-        if @course.syncable?
-          response = SearchAndCompareAPIService::Request.sync(
-            [@course]
+        if has_synced?
+          head :ok
+        else
+          raise RuntimeError.new(
+            'error received when syncing with search and compare'
           )
-          if response
-          else
-            raise RuntimeError.new(
-              'error received when syncing with search and compare'
-            )
-          end
         end
       end
 
@@ -39,22 +35,14 @@ module API
         if @course.publishable?
           @course.publish_sites
           @course.publish_enrichment(@current_user)
-          if @course.syncable?
-            has_synced = SearchAndCompareAPIService::Request.sync(
-              [@course]
-            )
-            if has_synced
-              head :ok
-            else
-              raise RuntimeError.new(
-                'error received when syncing with search and compare'
-              )
-            end
+          if has_synced?
+            head :ok
           else
             raise RuntimeError.new(
-              'course is not syncable'
+              'error received when syncing with search and compare'
             )
           end
+
         else
           render jsonapi_errors: @course.errors, status: :unprocessable_entity
         end
@@ -71,6 +59,7 @@ module API
       def update
         update_enrichment
         update_sites
+        has_synced? if site_ids.present?
 
         if @course.errors.empty? && @course.valid?
           render jsonapi: @course.reload
@@ -98,8 +87,6 @@ module API
         # but we can't actually revert easily from what I can tell because of the
         # remove_site! side effects that occur when it's called.
         @course.errors[:sites] << "^You must choose at least one location" if site_ids.empty?
-
-        sync_with_search_and_compare if site_ids.any?
       end
 
       def build_provider
@@ -134,6 +121,16 @@ module API
 
       def site_ids
         params.fetch(:course, {})[:sites_ids]
+      end
+
+      def has_synced?
+        if @course.syncable?
+          SearchAndCompareAPIService::Request.sync([@course])
+        else
+          raise RuntimeError.new(
+            'course is not syncable'
+          )
+        end
       end
     end
   end
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -22,14 +22,16 @@ module API
       end
 
       def sync_with_search_and_compare
-        response = sync_courses
-
-        if response
-          head :ok
-        else
-          raise RuntimeError.new(
-            'error received when syncing with search and compare'
+        if @course.syncable?
+          response = SearchAndCompareAPIService::Request.sync(
+            [@course]
           )
+          if response
+          else
+            raise RuntimeError.new(
+              'error received when syncing with search and compare'
+            )
+          end
         end
       end
 
@@ -37,16 +39,20 @@ module API
         if @course.publishable?
           @course.publish_sites
           @course.publish_enrichment(@current_user)
-
-          response = SearchAndCompareAPIService::Request.sync(
-            [@course]
-          )
-
-          if response
-            head :ok
+          if @course.syncable?
+            has_synced = SearchAndCompareAPIService::Request.sync(
+              [@course]
+            )
+            if has_synced
+              head :ok
+            else
+              raise RuntimeError.new(
+                'error received when syncing with search and compare'
+              )
+            end
           else
             raise RuntimeError.new(
-              'error received when syncing with search and compare'
+              'course is not syncable'
             )
           end
         else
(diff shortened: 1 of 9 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -22,33 +22,37 @@ def show
...
-        if response
-          head :ok
-        else
-          raise RuntimeError.new(
-            'error received when syncing with search and compare'
+        if @course.syncable?
+          response = SearchAndCompareAPIService::Request.sync(
+            [@course]
           )
+          if response
```
defong: response is a `bool`

app/controllers/api/v2/courses_controller.rb
```diff
@@ -22,33 +22,37 @@ def show
...
-
-          response = ManageCoursesAPIService::Request.sync_course_with_search_and_compare(
-            @current_user.email,
-            @provider.provider_code,
-            @course.course_code
-          )
-
-          if response
-            head :ok
+          if @course.syncable?
```
defong: maybe call `sync_with_search_and_compare`

app/controllers/api/v2/courses_controller.rb
```diff
@@ -95,7 +91,7 @@ def update_sites
         # remove_site! side effects that occur when it's called.
         @course.errors[:sites] << "^You must choose at least one location" if site_ids.empty?
 
-        sync_courses if site_ids.any?
+        sync_with_search_and_compare if site_ids.any?
```
defong: this call looks wrong

davidgisbey:
[[M] Sync course to search-and-compare](https://trello.com/c/3H9koXL1/1659-m-sync-course-to-search-and-compare)

app/controllers/api/v2/courses_controller.rb
davidgisbey: Done

app/controllers/api/v2/courses_controller.rb
davidgisbey: Done

app/controllers/api/v2/courses_controller.rb
davidgisbey: Is this what you were after?

defong: yes, and you can drop `if @course.syncable?`

davidgisbey: Cool. All done 👍

app/models/course.rb
```diff
@@ -338,4 +342,13 @@ def set_defaults
   def remove_unnecessary_enrichments_validation_message
     self.errors.delete :enrichments if self.errors[:enrichments] == ['is invalid']
   end
+
+  def validate_course_syncable
+    if findable?.blank?
```
defong: dont think you need blank as `findable?` is a bool right?

davidgisbey: Discussed. We need to check for false rather than true. It will be left how it is now.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -37,20 +43,7 @@ def publish
...
-          )
-
-          if response
-            head :ok
-          else
-            raise RuntimeError.new(
-              'error received when syncing with search and compare'
-            )
-          end
+          sync_with_search_and_compare
```
misaka: `sync_with_search_and_compare` is an action according to `routes.rb`. I don't think we should be calling an action from another action like this, so `sync_with_search_and_compare` is still being used then we should be using whatever it uses to sync ... either `SearchAndCompareAPIService::Request.sync` or a new, `private` method in this controller.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -37,20 +43,7 @@ def publish
...
           @course.publish_enrichment(@current_user)
-
-          response = ManageCoursesAPIService::Request.sync_course_with_search_and_compare(
-            @current_user.email,
-            @provider.provider_code,
-            @course.course_code
-          )
-
-          if response
-            head :ok
```
misaka: Should this not be returning `head :ok`? Or is that being handled elsewhere? I don't see where if so.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -59,6 +52,7 @@ def publish
       def update
         update_enrichment
         update_sites
+        sync_with_search_and_compare if site_ids.present?
```
misaka: Similar to the comment about `publish`, I don't think we should be relying on code in an existing action here.

spec/models/course_spec.rb
```diff
@@ -741,4 +741,28 @@
...
   end
+
+  describe '#syncable?' do
+    let(:subjects) { [build(:subject, subject_name: "primary")] }
+    let(:site_status) { build(:site_status, :findable) }
+
+    subject { create(:course, subjects: subjects, site_statuses: [site_status]) }
+
+    context 'course which is valid' do
+      let(:subject) { create(:course, subjects: subjects, site_statuses: [site_status]) }
```
misaka: You don't need this because of the `subject` above. You don't need to `let(:subject) {...}`, this is enough: `subject {...}`.

misaka: In fact, I'm not sure you even really need the `context` block here, since you could just say that by default, the course is valid, which seems like a reasonable assumption.

spec/models/course_spec.rb
```diff
@@ -741,4 +741,28 @@
...
+    end
+
+    context'invalid courses' do
+      context 'course which has a dfe subject, but no findable site statuses' do
+        let(:site_status) { build(:site_status, :suspended) }
+        its(:syncable?) { should be_falsey }
+      end
+
+      context 'course which has a findable site status, but no dfe_subject' do
+        let(:subjects) { [build(:subject, subject_name: "secondary")] }
```
misaka: Does `let(:subjects)` even work? I would have assumed that only `subject` is supported.

davidgisbey: This was just a work around to avoid subject. I'll change it to `non_dfe_subject` for clarity.

spec/models/course_spec.rb
davidgisbey: Have actioned.

app/controllers/api/v2/courses_controller.rb
davidgisbey: Okay. Have addressed this and the below comment. Let me know if you're happy with the solution.

app/controllers/api/v2/courses_controller.rb
davidgisbey: Same as above.

spec/requests/api/v2/providers/courses/update_sites_spec.rb
```diff
@@ -10,7 +10,9 @@
     ActionController::HttpAuthentication::Token.encode_credentials(token)
   end
 
-  let(:course) { create :course, provider: provider }
+  let(:course) { create :course, provider: provider, site_statuses: [site_status], subjects: [subject] }
+  let(:site_status) { build(:site_status) }
+  let(:subject) { build(:subject, subject_name: 'primary') }
```
misaka: On the back of the other comments, just noticed this use of `subject` here. Could we use `primary_subject` here?

davidgisbey: Done. Thanks for the review Misha.

PR #572, opened 2019-07-03
[1664] Override reverse_mapping in deserializable_course

tvararu:
### Context

Publish validations are hacky on the frontend because the backend does not give us source pointers which map the errors to their appropriate fields.

`reverse_mapping` is responsible for generating the `pointer` attributes in a JSON API response which contains validation errors. The way it works is `jsonapi-rails` will inspect the `params` and based on the attributes provided, populate `reverse_mapping`. This means that the only attributes that can be associated with errors are ones that get passed in as part of the request.

When we do the `publish` action, we don't pass in any attributes, because they already exist on the model. This means that without help, `jsonapi-rails` can't figure out to which attributes the fields we're showing errors on should map back to. In this scenario, it silently moves on without setting pointers.

We need validation errors to always have pointers, so it makes sense to force it to always include them by overriding the default `reverse_mapping`.

The qualifications are also added to the `attributes` call, to allow them to correctly map.

### Guidance to review

🤕 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/deserializers/api/v2/deserializable_course_spec.rb
+++ b/spec/deserializers/api/v2/deserializable_course_spec.rb
@@ -22,4 +22,14 @@ describe API::V2::DeserializableCourse do
       expect(subject[:qualifications]).to eq('test')
     end
   end
+
+  describe "reverse_mapping" do
+    subject { described_class.new({}).reverse_mapping }
+
+    it "always contains all attributes" do
+      API::V2::DeserializableCourse::COURSE_ATTRIBUTES.each do |attribute|
+        expect(subject[attribute.to_sym]).to eq("/data/attributes/#{attribute}")
+      end
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


tvararu:
[Errors that display on publish are persisted when users move to about/fees/requirements/salary](https://trello.com/c/TbGUzGvY/1664-errors-that-display-on-publish-are-persisted-when-users-move-to-about-fees-requirements-salary)

spec/deserializers/api/v2/deserializable_course_spec.rb
```diff
@@ -22,4 +22,15 @@
       expect(subject[:qualifications]).to eq('test')
     end
   end
+
+  describe "reverse_mapping" do
+    subject { described_class.new({}).reverse_mapping }
+
+    it "always contains all attributes" do
+      all_attributes = described_class.attr_blocks.keys
```
benilovj: This feels a bit too much like you're just repeating the implementation in the spec. One way to avoid it would be to:

1. Extract the attributes list in `DeserializableCourse` into a constant
2. Use the constant here.

That way you're avoiding having to maintain the list of attributes in multiple places.

PR #577, opened 2019-07-04
[1387] Add a rolled_over state for users

fofr:
### Context

- Use user state to determine if they've seen the rollover screen
- They must have seen the transition screen first

Goes with https://github.com/DFE-Digital/manage-courses-frontend/pull/430

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -38,9 +38,14 @@ class User < ApplicationRecord
   aasm column: 'state' do
     state :new, initial: true
     state :transitioned
+    state :rolled_over
 
     event :accept_transition_screen do
       transitions from: :new, to: :transitioned
+    end
+
+    event :accept_rollover_screen do
+      transitions from: :transitioned, to: :rolled_over
     end
   end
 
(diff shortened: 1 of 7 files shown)


fofr:
[[S] Rollover Screen - first log-in for users when next cycle is available](https://trello.com/c/CqD9fZa4/1387-s-rollover-screen-first-log-in-for-users-when-next-cycle-is-available)

app/models/user.rb
```diff
@@ -38,10 +38,15 @@ class User < ApplicationRecord
   aasm column: 'state' do
     state :new, initial: true
     state :transitioned
+    state :rolled_over
```
misaka: Is it premature to use `rolled_over_2020` here?

fofr: Next year we can always do `rolled_over_2021` if we need to. Not sure it's needed yet.

fofr:
[[M] Rollover screen - record first log-in for users when next cycle is available](https://trello.com/c/Qyn9ZuKT/1545-m-rollover-screen-record-first-log-in-for-users-when-next-cycle-is-available)

PR #578, opened 2019-07-04
[1664] Add #publishable endpoint on courses API

tvararu:
### Context

Deep linking validations.

This checks whether a course is publishable and replies with either a 200 or a 422 with the errors. Unlike `publish`, it doesn't actually publish.

### Guidance to review

This maybe should be a `GET` and not a `POST` but `jsonapi-rb` will not populate `source: { pointer }`s with a `GET`... I've spent too much time in that library already so am feeling weary about figuring out why.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -7,7 +7,7 @@ module API
       before_action :build_course, except: :index
 
       deserializable_resource :course,
-                              only: %i[update publish],
+                              only: %i[update publish publishable],
                               class: API::V2::DeserializableCourse
 
       def index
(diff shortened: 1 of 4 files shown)


tvararu:
[Errors that display on publish are persisted when users move to about/fees/requirements/salary](https://trello.com/c/TbGUzGvY/1664-errors-that-display-on-publish-are-persisted-when-users-move-to-about-fees-requirements-salary)

tvararu:
This needs to be merged/deployed AFTER https://github.com/DFE-Digital/manage-courses-backend/pull/578

PR #580, opened 2019-07-04
[1657] mcb courses edit: edit training locations

benilovj:
### Context
Part of creating a new course through `mcb` involves selecting which training locations should be added to it. It's also necessary to to add or remove training locations on an existing course when all sites are suspended, or all sites need to be suspended.

- add editing of training locations to `mcb courses edit`
- clean up the `CoursesEditorCLI` class by extracting some generic functionality out into `BaseCLI` (which will be useful in subsequent PRs)

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


lib/mcb/courses_editor.rb
```diff
@@ -59,12 +60,14 @@ def main_loop
     end
 
     def filter_single_course_options_if_necessary(choices)
-      choices.grep_v(->(c) { c.in?(["edit subjects"]) && @courses.count != 1 })
+      choices.grep_v(->(c) { c.in?(["edit subjects", "edit training locations"]) && @courses.count != 1 })
```
misaka: Would it be worth using something more idiomatic like:

```suggestion
      choices.reject { |c| c.in?(["edit subjects", "edit training locations"]) && @courses.count != 1 }
```

The mechanism with `grep_v` seems to rely on `Proc#===` which apparently invokes `Proc#call` with the object being compared-to. Handy, but not as well understood as plain-old `reject`. :)

benilovj: thanks – done

PR #583, opened 2019-07-07
[1737] mcb providers edit

benilovj:
### Context

Provider editing is currently being done through user support using `mcb` dev tooling. The tooling itself is sitting on a code branch that isn't the `master` branch and needs constant updating/rebasing.

The code is also missing automated tests, meaning it needs to be manually tested after every change.

Add an `mcb providers edit` command. The new command:
- can edit the provider name
- has a convenience function for selecting multiple courses and invoking `mcb courses edit` on them
- prints outs the provider in the same way that `mcb courses edit` prints courses

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


lib/mcb/commands/providers/edit.rb
```diff
@@ -0,0 +1,12 @@
+name 'show'
```
misaka: I think this is meant to be `edit`?

misaka: Pro tip: We don't actually need command names in these files, `Cri` takes it from the filename.

PR #590, opened 2019-07-09
[1657] Add 'mcb courses create' subcommand for creating new courses

benilovj:
### Context

New course creation is currently being done through user support using `mcb` dev tooling. The tooling itself is sitting on a code branch that isn't the `master` branch and needs constant updating/rebasing.

The tooling branch code is also missing automated tests, meaning it needs to be manually tested after every change.

Add an `mcb` command for creating new courses on a given provider.

### Guidance to review

It was a slog getting here 😅 
~This PR ought to be reviewed but probably not merged until the Rollover stuff has been merged, because the recruitment cycle stuff will conflict.~ The `mcb` command adds a new course to the current cycle by default.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -43,8 +43,7 @@ module MCB
 
       puts "\nAbout to create the following course:"
       print_at_most_two_courses
-      if @cli.confirm_creation?
-        try_saving_course
+      if @cli.confirm_creation? && try_saving_course
         edit_subjects
         edit_sites
         edit(:application_opening_date)
@@ -165,9 +164,11 @@ module MCB
       if course.valid?
         puts "Saving the course"
         course.save!
+        true
       else
         puts "Course isn't valid:"
         course.errors.full_messages.each { |error| puts " - #{error}" }
+        false
       end
     end
 
--- a/lib/mcb/courses_editor.rb
+++ b/lib/mcb/courses_editor.rb
@@ -9,12 +9,12 @@ module MCB
       application_opening_date: :applications_open_from,
     }.freeze
 
-    def initialize(provider:, requester:, course_codes: [])
+    def initialize(provider:, requester:, course_codes: [], courses: nil)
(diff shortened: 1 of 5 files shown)


lib/mcb/courses_editor.rb
```diff
@@ -34,6 +34,25 @@ def run
       end
     end
 
+    def new_course_wizard
+      %i[title qualifications study_mode accredited_body start_date route maths
```
fofr: Are these processed in order – we might want to put recruitment_cycle first?

benilovj: Yes, these are currently processed in order.

However I'm not expecting this to be merged before the Rollover branch gets merged in, whereby the recruitment cycle choice step will be rebased out and pulled up to the top level (i.e. specified on the command line, before the interactive part).

Do we want any of the other steps in a different order?

benilovj:
This is ready for review now.

lib/mcb/courses_editor.rb
```diff
@@ -118,16 +142,42 @@ def print_at_most_two_courses
...
+      @cli.enter_to_continue
+    end
+
+    def try_saving_course
+      if course.valid?
+        puts "Saving the course"
+        course.save!
+      else
+        puts "Course isn't valid:"
+        course.errors.full_messages.each { |error| puts " - #{error}" }
```
misaka: If this fails, the user will go straight to editing subjects and sites, which will both fail, which will fail, I think, since `course` isn't valid. Would it make more sense to kick them back up to the top of the wizard to try again before proceeding?

benilovj: I've added 30c563c, which does the slightly simpler thing of aborting the wizard if the course fails validation. From that point onward, it's easy for the support agent to just restart the process by invoking the previous command and copying their previous answers from the terminal history.

PR #593, opened 2019-07-10
1745 refactor provider on v1 api

davidgisbey:
### Context
The V1 API is currently not cycle aware. 

- exposes recruitment cycle in v1 provider serialiser
- creates a scope for all courses in a specific cycle
- refactors the provider controller to use the recruitment year in the url if there
- if not defaults to current recruitment cycle

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/controllers/api/v1/providers_controller.rb
```diff
@@ -16,10 +16,7 @@ class ProvidersController < API::V1::ApplicationController
       # - clock drift between servers
       def index
         # only return 2019 courses until rollover is supported
-        if params[:recruitment_year].present? && params[:recruitment_year] != '2019'
-          render json: [], status: :not_found
-          return
-        end
+        build_recruitment_cycle
```
defong: i thought you wanted the `before_action`

also the comments above is kinda not making sense any morish

app/controllers/api/v1/providers_controller.rb
```diff
@@ -16,10 +18,6 @@ class ProvidersController < API::V1::ApplicationController
       # - clock drift between servers
       def index
         # only return 2019 courses until rollover is supported
```
defong: rogue comment?

app/models/provider.rb
```diff
@@ -74,6 +74,10 @@ class Provider < ApplicationRecord
 
   scope :in_order, -> { order(:provider_name) }
 
+  scope :by_recruitment_cycle, ->(recruitment_year) {
+    joins(:recruitment_cycle).merge(RecruitmentCycle.where(year: recruitment_year))
```
defong: is there a need for a join?

given that a provider belongs to a recruitment year?

spec/requests/api/v1/provider_spec.rb
```diff
@@ -275,6 +279,42 @@ def get_provider_codes_from_body(body)
...
+
+          it 'defaults to the current cycle when year' do
+            returned_provider_codes = get_provider_codes_from_body(response.body)
+
+            expect(returned_provider_codes).not_to include provider2.provider_code
+            expect(returned_provider_codes).to include provider.provider_code
+          end
+        end
+        context 'with a future recruitment cycle specified in the route' do
+          let(:get_index) { get '/api/v1/2020/providers', headers: { 'HTTP_AUTHORIZATION' => credentials } }
```
defong: wondering should this be condense with 

for current
`'/api/v1/providers'`
`'/api/v1/#{current}/providers'`

for future
`'/api/v1/#{future}/providers'`

app/controllers/api/v1/providers_controller.rb
davidgisbey: Done 👍

app/models/provider.rb
davidgisbey: Done 👍

spec/requests/api/v1/provider_spec.rb
davidgisbey: Seems like it's longer than just having 2020 though right?

davidgisbey:
[[S] Change apiv1 course to return 2020 recruitment cycle](https://trello.com/c/ss65HD8u/1647-s-change-apiv1-course-to-return-2020-recruitment-cycle)

davidgisbey:
[Refactor V1 Provider controller/serializers to reflect the new data model](https://trello.com/c/K4L4hzC1/1745-refactor-v1-provider-controller-serializers-to-reflect-the-new-data-model)

davidgisbey:
[Allow recruitment cycle to be specified as a query parameter for backwards compatibility](https://trello.com/c/xgqaAf6E/1721-allow-recruitment-cycle-to-be-specified-as-a-query-parameter-for-backwards-compatibility)

PR #596, opened 2019-07-11
[1760] Removed pgde course

defong:
### Context
Pgde course model
Removed unneeded file
Removed pgde course as well

### Guidance to review
:ship: 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


db/migrate/20190711085633_drop_pgde_course.rb
```diff
@@ -0,0 +1,9 @@
+class DropPGDECourse < ActiveRecord::Migration[5.2]
+  def up
+    drop_table :pgde_course
+  end
+
+  def down
```
dankmitchell: Could make it a `change` rather than `up/down`

defong:
```
drop_table(without block) is not reversible.
```

defong:
[Remove pgde course from database](https://trello.com/c/nfxGa1xn/1760-remove-pgde-course-from-database)

db/migrate/20190711085633_drop_pgde_course.rb
```diff
@@ -0,0 +1,9 @@
...
+    # There is no going back as it was legacy.
```
misaka: Wouldn't it make sense to just have the table there, at least? In case we have to rollback?

misaka: Just discussed and @defong tells me that there's data that is lost in the drop table. So maybe we should remove the `def down` so that this migration can't be rolled back, or if that doesn't work find a way to raise an error to block this from going backwards (since we wouldn't be able to run the migrations again if we did).

defong: done

```ruby

raise ActiveRecord::IrreversibleMigration, "explain"

```

misaka:
🚢 :shipit: 🚀

defong:
@timabell will dropping a table be an issue for https://github.com/DFE-Digital/manage-courses-api

Till answered hanging this to dry...

timabell:
Holding off merging till all the rollover churn is out of the way and the dust has settled.

PR #597, opened 2019-07-11
[1743] Make V1 API  course controller cycle aware

davidgisbey:
### Context

The V1 API course controller is currently not cycle aware.

- Refactor v1 course controller to build recruitment cycle
- Use scope to return the correct years courses
- Defaults to current cycles courses
- Makes `set_next_link_header_using_changed_since_or_last_object`  return the recruitment year in the path rather than in the params

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/controllers/api/v1/courses_controller.rb
```diff
@@ -18,21 +19,26 @@ def index
...
-                       .by_recruitment_cycle(recruitment_year)
+                       .by_recruitment_cycle(@recruitment_cycle.year)
                        .limit(per_page)
         end
 
         set_next_link_header_using_changed_since_or_last_object(
           @courses.last,
           changed_since: changed_since,
           per_page: per_page,
-          recruitment_year: recruitment_year
```
misaka: Dropping this here makes me nervous, even though it feels like it *should* work, and there is another PR coming that should fix this. Can we actually make one PR for all the course changes, and another for all the provider changes.

davidgisbey: Now part of this PR. See below.

app/controllers/api/v1/courses_controller.rb
```diff
@@ -18,21 +19,27 @@ def index
                                  :subjects,
                                  site_statuses: [:site])
                        .changed_since(changed_since)
-                       .by_recruitment_cycle(recruitment_year)
+                       .by_recruitment_cycle(@recruitment_cycle.year)
```
misaka: If you look at the implementation for `by_recruitment_cycle`:

```ruby
scope :by_recruitment_cycle, ->(recruitment_year) {
  joins(provider: :recruitment_cycle).merge(RecruitmentCycle.where(year: recruitment_year))
}
```

It seems a bit odd to be instantiate a recruitment cycle in the controller action using the year, then passing the year value just to instantiate another recruitment cycle object using the year.

Not a big deal as this only happens once per index, but maybe this would be better expressed like this:

```ruby
@courses = @recruitment_cycle
             .courses
             .includes(:provider
                       :site_statuses,
                       :accrediting_provider,
                       :subjects,
                       site_statuses: [:site])
             .changed_since(changed_since)
             .limit(per_page)
```

But not a blocker.

app/controllers/concerns/next_link_header.rb
```diff
@@ -5,12 +5,21 @@ module NextLinkHeader
...
-      params[:changed_since] =
+      next_url_params[:changed_since] =
         incremental_load_timestamp_format last_object.changed_at
     end
 
-    response.headers['Link'] = "#{url_for(params: params)}; rel=\"next\""
+    if params[:recruitment_year].nil?
+      response.headers['Link'] = "#{url_for(recruitment_year: Settings.current_recruitment_cycle, params: next_url_params)}; rel=\"next\""
+      return
+    end
```
misaka: I find the return here a little confusing, and also had to check to see if the string value for the Link header changes between the two commands. Can I suggest:

```ruby
cycle_year = params.fetch(:recruitment_year, Settings.current_recruitment_cycle)

response.headers['Link'] = "#{url_for(recruitment_year: cycle_year, params: next_url_params)}; rel=\"next\""
```

misaka:
None of the comments are blockers, both are nice-to-haves.

spec/requests/api/v1/course_spec.rb
```diff
@@ -201,7 +243,21 @@ def get_course_codes_from_body(body)
...
+      end
+
+      it 'includes correct next link when there is an empty set' do
+        provided_timestamp = 5.seconds.ago.utc.iso8601
+
+        get '/api/v1/2020/courses',
+            headers: { 'HTTP_AUTHORIZATION' => credentials },
+            params: { changed_since: provided_timestamp }
+
+        url = url_for(recruitment_year: 2020, params: {
```
misaka: Oh, one more nice-to-have, how about making this the full explicit URL? So it's clear exactly what this should look like.

spec/requests/api/v1/course_spec.rb
```diff
@@ -186,11 +226,13 @@ def get_course_codes_from_body(body)
...
-                        changed_since: timestamp_of_last_course.utc.strftime('%FT%T.%6NZ'),
-                        per_page: 100,
-                        recruitment_year: 2019
-                      })
+        url = url_for(
+          recruitment_year: 2019,
+          params: {
+            changed_since: timestamp_of_last_course.utc.strftime('%FT%T.%6NZ'),
+            per_page: 100
+          }
```
misaka: Oh, one more nice-to-have, how about making this the full explicit URL? So it's clear exactly what this should look like.

app/controllers/api/v1/courses_controller.rb
davidgisbey: Done.

spec/requests/api/v1/course_spec.rb
davidgisbey: Have experimented. Can't get it work due to string matching not working. Asked Sau. Both couldn't figure it out. Don't want to waste more time on a nice to have.

spec/requests/api/v1/course_spec.rb
davidgisbey: Have experimented. Can't get it work due to string matching not working. Asked Sau. Both couldn't figure it out. Don't want to waste more time on a nice to have.

app/controllers/concerns/next_link_header.rb
davidgisbey: Your suggestion did not return the correct cycle when params[:recruitment_year] was nil so i've added the following:
```

  if params[:recruitment_year].nil?
      cycle_year = Settings.current_recruitment_cycle
      response.headers['Link'] = "#{url_for(recruitment_year: cycle_year, params: next_url_params)}; rel=\"next\""
    else
      response.headers['Link'] = "#{url_for(recruitment_year: params[:recruitment_year], params: next_url_params)}; rel=\"next\""
    end

```

davidgisbey:
[Refactor course V1 controller controller/serializers to reflect the new data model](https://trello.com/c/yi5k5MeS/1743-refactor-course-v1-controller-controller-serializers-to-reflect-the-new-data-model)

davidgisbey:
[Allow recruitment cycle to be specified as a query parameter for backwards compatibility](https://trello.com/c/xgqaAf6E/1721-allow-recruitment-cycle-to-be-specified-as-a-query-parameter-for-backwards-compatibility)

PR #602, opened 2019-07-11
[1586] Improvements to applications open from

fofr:
### Context

A draft PR indicating potential improvements to Course#applications_open_from based on @benilovj's branch.

"Applications open from" should be the earliest date in any of the sites. Currently it's restricted to looking at just the sites with open applications.

This means that no date is available if applications haven't opened yet.
The wrong date might show if some of the sites have no vacancies.
No date available if the site is new and not yet running.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -71,7 +71,7 @@ class Course < ApplicationRecord
   has_many :subjects, through: :course_subjects
   has_many :site_statuses
   has_many :sites,
-           -> { merge(SiteStatus.where(status: %i[new_status running])) },
+           -> { merge(SiteStatus.findable_or_new) },
            through: :site_statuses
 
   has_many :enrichments,
@@ -149,7 +149,7 @@ class Course < ApplicationRecord
 
   def applications_open_from
     site_statuses
-      .findable
+      .findable_or_new
       .with_vacancies
       .order("applications_accepted_from ASC")
       .first
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -158,6 +158,10 @@ class Course < ApplicationRecord
       &.iso8601
   end
 
+  def applications_open_from=(new_date)
+    site_statuses.each { |ss| ss.update(applications_accepted_from: new_date) }
+  end
+
   def has_vacancies?
     site_statuses.findable.with_vacancies.any?
   end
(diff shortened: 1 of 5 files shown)


fofr:
[Applications open from shouldn't be restricted to just sites that are open to applications](https://trello.com/c/l0t1Oun3/1699-applications-open-from-shouldnt-be-restricted-to-just-sites-that-are-open-to-applications)

app/serializers/search_and_compare/course_serializer.rb
```diff
@@ -71,7 +71,7 @@ class CourseSerializer < ActiveModel::Serializer
     # Campuses_related_Mapping
     attribute(:Campuses)                              { campuses }
     # using server date time not utc, so it's local date time?
-    attribute(:ApplicationsAcceptedFrom)              { object.applications_open_from.to_date.strftime('%Y-%m-%dT%H:%M:%S') }
+    attribute(:ApplicationsAcceptedFrom)              { object.applications_open_from.strftime('%Y-%m-%dT%H:%M:%S') }
```
defong:
```
def campuses
   object.site_statuses.findable.map do |site_status|
```

I guess that needs to change as well?

app/models/course.rb
```diff
@@ -149,13 +149,17 @@ def open_for_applications?
 
   def applications_open_from
     site_statuses
-      .open_for_applications
+      .findable_or_new
+      .with_vacancies
```
defong: > This means that no date is available if applications haven't opened yet.
The wrong date might show if some of the sites have no vacancies.
No date available if the site is new and not yet running.

Not reading last time or this time properly....

But I guess 

```
Get me the applications open from date from the course
where at least one site statuses is both
findable or new
with vacancies

and grab the earliest
```

hmmm 
why find or `new`?

Given that there `new` is just a transitional state that gets flipped when this is triggered?

https://github.com/DFE-Digital/manage-courses-backend/blob/6fb4326919b4ba1f1c51af633abcf87ee0ac8297/app/controllers/api/v2/courses_controller.rb#L37

misaka:
@dankmitchell Can we actually remove the sidebar stuff from this PR and do it with a separate ticket? Just because of the logic behind published/not published, etc. We're starting to ticket up the provider enrichment stuff now anyway.

NyWeb:
Closing this as it's outdated. We'll have to do the underlying applications_open_from eventually.

PR #603, opened 2019-07-11
[1731] Add option to change recruitment year for list commands

faissalMT:
### Context
Rollover wasn't yet reflected in MCB.

Allow the specification of the current recruitment year, or default to the current recruitment year.

### Guidance to review
  
Main concern is test style.  
 [Add option to specify recruitment cycle to mcb commands.](https://trello.com/c/T5oYvmB1/1731-add-option-to-specify-recruitment-cycle-to-mcb-commands)  
  
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -247,6 +247,9 @@ module MCB
     def remote_connect_options
       envs = env_to_azure_map.keys.join(', ')
       Proc.new do
+        option :R, 'recruitment_year',
+               "Set the recruitment year, defaults to the current recruitment year",
+               argument: :required
         option :E, 'env',
                "Connect to a pre-defined environment: #{envs}",
                argument: :required
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -113,13 +113,33 @@ module MCB
   end
 
   def self.each_v1_course(opts)
+    endpoint = "/api/v1/#{get_recruitment_year(opts)}/courses"
+
     # This method can actually be entirely driven by the args provided, it is auto-configure!
-    iterate_v1_endpoint(**opts)
+    iterate_v1_endpoint(endpoint: endpoint, **opts)
   end
 
   def self.each_v1_provider(opts)
+    endpoint = "/api/v1/#{get_recruitment_year(opts)}/providers"
+
     # This method can actually be entirely driven by the args provided, it is auto-configure!
-    iterate_v1_endpoint(**opts)
+    iterate_v1_endpoint(endpoint: endpoint, **opts)
+  end
+
+  def self.get_recruitment_year(opts)
+    raise RuntimeError, 'Rails has not been initialised' if !defined? Rails
+
+    opts[:'recruitment-year'] || RecruitmentCycle.current_recruitment_cycle.year
+  end
+
+  def self.get_recruitment_cycle(opts)
+    raise RuntimeError, 'Rails has not been initialised' if !defined? Rails
+
+    if opts.key? :'recruitment-year'
+      RecruitmentCycle.find_by(year: opts[:'recruitment-year'])
+    else
+      RecruitmentCycle.current_recruitment_cycle
+    end
   end
 
   def self.config_dir=(dir)
@@ -161,7 +181,6 @@ module MCB
       # the following lines are necessary to make opts work with double **splats and default values
       # See the change introduced in https://github.com/ddfreyne/cri/pull/99 (cri 2.15.8)
       opts[:url] = opts[:url]
-      opts[:endpoint] = opts[:endpoint]
       opts[:'max-pages'] = opts[:'max-pages']
       opts[:token] = opts[:token]
       opts[:all] = opts[:all]
@@ -205,7 +224,7 @@ module MCB
       # We only need httparty for API V1 calls
       require 'httparty'
 
-      endpoint_url = process_opt_changed_since(opts, URI.join(url, endpoint))
+      endpoint_url = add_url_params_from_opts(opts, URI.join(url, endpoint))
       token = opts.fetch(:token) { apiv1_token(opts.slice(:webapp, :rgroup)) }
 
       # Safeguard to ensure we don't go off the deep end.
@@ -247,7 +266,7 @@ module MCB
     def remote_connect_options
       envs = env_to_azure_map.keys.join(', ')
       Proc.new do
-        option :R, 'recruitment_year',
+        option :r, 'recruitment-year',
                "Set the recruitment year, defaults to the current recruitment year",
                argument: :required
         option :E, 'env',
@@ -403,7 +422,7 @@ module MCB
       end
     end
 
-    def process_opt_changed_since(opts, url)
+    def add_url_params_from_opts(opts, url)
       new_url = url.dup
       if opts.key? :'changed-since'
         changed_since = DateTime.strptime(
@@ -414,6 +433,7 @@ module MCB
         changed_since_param = CGI.escape(changed_since.strftime('%FT%T.%6NZ'))
         new_url.query = "changed_since=#{changed_since_param}"
       end
+
       new_url
     end
   end
--- a/lib/mcb/commands/providers/list.rb
+++ b/lib/mcb/commands/providers/list.rb
@@ -12,7 +12,7 @@ run do |opts, args, _cmd|
                       end
 
   providers = recruitment_cycle.providers
-  providers = providers.where(provider_code: args.to_a) if args.any?
+  providers = providers.where(provider_code: args.map(&:upcase)) if args.any?
 
   output = [
     'Providers:',
--- a/lib/mcb/commands/providers/list.rb
+++ b/lib/mcb/commands/providers/list.rb
@@ -5,11 +5,7 @@ summary 'List providers in db'
 run do |opts, args, _cmd|
   MCB.init_rails(opts)
 
-  recruitment_cycle = if opts[:recruitment_year].nil?
-                        RecruitmentCycle.current_recruitment_cycle
-                      else
-                        RecruitmentCycle.find_by(year: opts[:recruitment_year])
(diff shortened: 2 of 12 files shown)


lib/mcb.rb
```diff
@@ -247,6 +247,9 @@ def iterate_v1_endpoint(url:, endpoint:, **opts)
     def remote_connect_options
       envs = env_to_azure_map.keys.join(', ')
       Proc.new do
+        option :R, 'recruitment_year',
+               "Set the recruitment year, defaults to 2019",
+               argument: :required
```
defong: reads kinda wrong...
argument being required yet defaulting?

misaka: I think it's the `recruitment_year` value that's required.

So this is valid:

```
mcb providers list
mcb providers list -R 2020
```

But not this:

```
mcb providers list -R
```

lib/mcb.rb
```diff
@@ -247,6 +247,9 @@ def iterate_v1_endpoint(url:, endpoint:, **opts)
     def remote_connect_options
       envs = env_to_azure_map.keys.join(', ')
       Proc.new do
+        option :R, 'recruitment_year',
+               "Set the recruitment year, defaults to 2019",
```
misaka: I think the default is the current recruitment cycle, which happens to be 2019 now but will change in the future. Might be better to use that wording.

faissalMT:
[Add option to specify recruitment cycle to mcb commands.](https://trello.com/c/T5oYvmB1/1731-add-option-to-specify-recruitment-cycle-to-mcb-commands)

spec/lib/mcb/commands/apiv1/courses/list_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    expect(output).to have_text_table_row('Code',
+                                          'Name',
+                                          'Provider Code',
+                                          'Provider Name')
+    expect(output).to have_text_table_row(course1.course_code, course1.name)
+    expect(output).to have_text_table_row(course2.course_code, course2.name)
+  end
+
+  it 'lists courses for a given recruitment year' do
+    url = "http://localhost:3001/api/v1/2019/courses?recruitment_year=#{next_cycle.year}"
```
defong: http://localhost:3001/api/v1/`2019`/courses?recruitment_year=`#{next_cycle.year}`

so who wins? `2019` vs `#{next_cycle.year}`

spec/lib/mcb/commands/apiv1/providers/list_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    expect(output).to have_text_table_row('Code',
+                                          'Name')
+    expect(output).to have_text_table_row(provider1.provider_code, provider1.provider_name)
+    expect(output).to have_text_table_row(provider2.provider_code, provider2.provider_name)
+  end
+
+  it 'lists providers for a given recruitment year' do
+    url = "http://localhost:3001/api/v1/2019/providers?recruitment_year=#{next_cycle.year}"
```
defong: same as above who wins? and should it be possible?

spec/lib/mcb/commands/providers/list_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        command_output = list(provider1.provider_code)[:stdout]
+
+        expect(command_output).to include(provider1.provider_code)
+        expect(command_output).to include(provider1.provider_name)
+
+        expect(command_output).not_to include(provider2.provider_code)
+        expect(command_output).not_to include(provider2.provider_name)
+      end
+
+      it 'displays all specified providers' do
```
defong: why is this one a `it` and this is a `context`

```
context 'when provider is unspecified' do
```

lib/mcb.rb
```diff
@@ -247,6 +247,9 @@ def iterate_v1_endpoint(url:, endpoint:, **opts)
     def remote_connect_options
       envs = env_to_azure_map.keys.join(', ')
       Proc.new do
+        option :R, 'recruitment_year',
```
misaka: Can we use the more standard `recruitment-year` here?

lib/mcb.rb
```diff
@@ -411,6 +414,11 @@ def process_opt_changed_since(opts, url)
         changed_since_param = CGI.escape(changed_since.strftime('%FT%T.%6NZ'))
         new_url.query = "changed_since=#{changed_since_param}"
       end
+
+      if opts.key? :recruitment_year
+        new_url.query = [new_url.query, "recruitment_year=#{opts[:recruitment_year]}"].compact.join('&')
+      end
```
misaka: How about generating a hash and using `to_query` to generate a query from it? Just thinking it'd be a tidier.

```
$ rails c
> { a: 'a', b: 'b' }.to_query
=> "a=a&b=b"
```

misaka:
Loving the new option and the new tests. 😍

spec/lib/mcb/commands/providers/list_spec.rb
faissalMT: This one is still specific, I'll rephrase it

spec/lib/mcb/commands/apiv1/courses/find_spec.rb
```diff
@@ -10,7 +10,7 @@
     course2 = create(:course, subjects: [subject], site_statuses: [site_status])
 
 
-    url = 'http://localhost:3001/api/v1/2019/courses'
+    url = 'http://localhost:3001/api/v1/courses'
```
misaka: Just to check, does this simulate the URL that UCAS is currently using? Or is it meant to be the primary way to access the API? To me this looks like it's simulating the URL that UCAS should have stored from the current version of the API.

spec/lib/mcb/commands/apiv1/providers/find_spec.rb
```diff
@@ -9,7 +9,7 @@
   let(:provider2) { create(:provider, sites: [site2], contacts: [contact2]) }
 
   it 'displays the info for the given provider' do
-    url = 'http://localhost:3001/api/v1/2019/providers'
+    url = 'http://localhost:3001/api/v1/providers'
```
misaka: Same question as for the api v1 courses list.

spec/lib/mcb/commands/apiv1/providers/list_spec.rb
```diff
@@ -0,0 +1,91 @@
+require 'mcb_helper'
+
+describe '"mcb apiv1 providers list"' do
+  let(:current_cycle) { find_or_create(:recruitment_cycle, year: '2019') }
+  let(:next_cycle) { find_or_create(:recruitment_cycle, year: '2020') }
+  let(:provider1) { create(:provider, recruitment_cycle: current_cycle) }
+  let(:provider2) { create(:provider, recruitment_cycle: next_cycle) }
+
+  it 'lists providers' do
+    url = 'http://localhost:3001/api/v1/providers'
```
misaka: I'm not sure this test makes sense. For API V1 we don't expect to ever get a request without a recruitment cycle, and is documented to only work with the recruitment cycle:

https://github.com/DFE-Digital/manage-courses-backend/blob/master/docs/api.md

I would expect there to be a test to show that if the `-r` option is not provider, only providers from the current cycle are displayed.

spec/lib/mcb/commands/providers/list_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        expect(command_output).to include(provider1.provider_code)
+        expect(command_output).to include(provider1.provider_name)
+
+        expect(command_output).to include(provider2.provider_code)
+        expect(command_output).to include(provider2.provider_name)
+      end
+    end
+  end
+
+  context 'when recruitment cycle is specified' do
```
misaka: How about a test that when recruitment-cycle is not specified, it defaults to the current one?

spec/lib/mcb/commands/courses/list_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+        expect(command_output).to include(course1.course_code)
+        expect(command_output).to include(course1.name)
+
+        expect(command_output).to include(course2.course_code)
+        expect(command_output).to include(course2.name)
+      end
+    end
+  end
+
+  context 'when recruitment cycle is specified' do
```
misaka: How about a test that when recruitment-cycle is not specified, it defaults to the current one?

lib/mcb/commands/providers/list.rb
```diff
@@ -5,10 +5,10 @@
 run do |opts, args, _cmd|
   MCB.init_rails(opts)
 
-  recruitment_cycle = if opts[:recruitment_year].nil?
+  recruitment_cycle = if !opts.key? :'recruitment-year'
                         RecruitmentCycle.current_recruitment_cycle
                       else
-                        RecruitmentCycle.find_by(year: opts[:recruitment_year])
+                        RecruitmentCycle.find_by(year: opts[:'recruitment-year'])
```
misaka: nneeeeewww method!!! :)

lib/mcb.rb
```diff
@@ -113,13 +113,29 @@ def self.generate_apiv2_token(email:, encoding:, secret: nil)
...
 
   def self.each_v1_provider(opts)
+    endpoint = "/api/v1/#{get_recruitment_year(opts)}/providers"
+
     # This method can actually be entirely driven by the args provided, it is auto-configure!
-    iterate_v1_endpoint(**opts)
+    iterate_v1_endpoint(endpoint: endpoint, **opts)
+  end
+
+  def self.get_recruitment_year(opts)
```
misaka: If this method gets called before Rails is initialized it'll fail in potentially a non-obvious way. Can we make it more obvious to the user that Rails is required?

PR #604, opened 2019-07-11
[1763] change accrediting provider assocation to use code

misaka:
### Context

Because when we link it by ID, we link it to a provider in a specific recruitment cycle, which makes rolling-over a bit more complicated. By changing this to provider_code it becomes a simpler relationship that doesn't risk being out-of-sync.

Change `accrediting_provider_id` to `accrediting_provider_code` in `Course`.

### Guidance to review

Because `Course#accrediting_provider` is now a dynamic association, we can't pre-load it, which leads to a degradation in performance which will probably be most felt on the courses endpoint for api v1. However, with local testing it doesn't seem to bad since Rails can cache these SQL lookups when we're looking up the same provider.

```
[active_model_serializers]   Provider Load (0.2ms)  SELECT  "provider".* FROM "provider" WHERE "provider"."provider_code" = $1 AND "provider"."recruitment_cycle_id" = $2 LIMIT $3  [["provider_code", "N42"], ["recruitment_cycle_id", 1], ["LIMIT", 1]]
[active_model_serializers]   ↳ app/controllers/api/v1/courses_controller.rb:31
[active_model_serializers]   CACHE Provider Load (0.0ms)  SELECT  "provider".* FROM "provider" WHERE "provider"."provider_code" = $1 AND "provider"."recruitment_cycle_id" = $2 LIMIT $3  [["provider_code", "N42"], ["recruitment_cycle_id", 1], ["LIMIT", 1]]
[active_model_serializers]   ↳ app/controllers/api/v1/courses_controller.rb:31
```

Even if for whatever reason we don't get as many cache hits as we'd hope, say if the courses are updated in such a way so that each course on a page of 100 is accredited by a different provider, the timings to perform the lookup on a single one seems to be fast enough to not be an issue (0.2ms in the example above, which looks representative looking at the console output on my laptop).

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/db/migrate/20190711074433_change_course_accrediting_provider_id_to_code.rb
+++ b/db/migrate/20190711074433_change_course_accrediting_provider_id_to_code.rb
@@ -1,0 +1,44 @@
+class ChangeCourseAccreditingProviderIdToCode < ActiveRecord::Migration[5.2]
+  class Course < ApplicationRecord
+    belongs_to :accrediting_provider, class_name: 'Provider', optional: true
+  end
+
+  def change
+    add_column :course, :accrediting_provider_code, :text
+    add_index :course, :accrediting_provider_code
+
+    reversible do |dir|
+      say_with_time 'updating accrediting provider on courses' do
+        Course.all.each do |course|
+          dir.up {
+            course.update accrediting_provider_code:
+                            course.accrediting_provider&.provider_code
+          }
+          # manage_courses_backend_development=# select count(*) from course where accrediting_provider_id is null;
+          #   count
+          # -------
+          #   3403
+
+          # manage_courses_backend_development=# select count(*) from course where accrediting_provider_code is null;
+          #   count
+          # -------
+          #   3403
+          dir.down {
+            course.update accrediting_provider_id:
+                            course.accrediting_provider&.id
+          }
+        end
+      end
+    end
+
+    revert do
+      add_column :course, :accrediting_provider_id, :integer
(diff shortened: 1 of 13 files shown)


misaka:
[Link course to accrediting provider using provider code.](https://trello.com/c/Ur9oOSxD/1763-link-course-to-accrediting-provider-using-provider-code)

db/migrate/20190711074433_change_course_accrediting_provider_id_to_code.rb
```diff
@@ -0,0 +1,44 @@
...
+    add_index :course, :accrediting_provider_code
+
+    reversible do |dir|
+      say_with_time 'updating accrediting provider on courses' do
+        Course.all.each do |course|
+          dir.up {
+            course.update accrediting_provider_code:
+                            course.accrediting_provider&.provider_code
+          }
+          # manage_courses_backend_development=# select count(*) from course where accrediting_provider_id is null;
```
defong: why 

 select count(*) from course where accrediting_provider_id `is null`;

same below

misaka:
> Should `address3` stuff be here?

No, ugh, not sure how that got in here. Thanks.

misaka:
I see what happened here ... I merged 1765 into this branch. Not sure that was the right thing to do, but since that's approved I'll just merge both into master now.

PR #605, opened 2019-07-11
[1765] remove site address3 validation

misaka:
### Context

We have a bunch of legacy sites in the db that are invalid because they're missing a value for `address3`, and it's becoming increasingly difficult to work-around this validation. Fox example for the rollover, copying the sites is failing because they're invalid, and even creating the data in a spec is painful.

Remove the presence validation we have on `site.address3`.

### Guidance to review

I'm looking for some validation that this is a good idea, let me know if you think that there's a better approach to this, or if we should just suck it up and do the work to disable validations for tests & rollover.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -34,7 +34,6 @@ class Site < ApplicationRecord
   validates :location_name, uniqueness: { scope: :provider_id }
   validates :location_name,
             :address1,
-            :address3,
             :postcode,
             presence: true
   validates :postcode, postcode: true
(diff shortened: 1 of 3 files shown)


misaka:
[Remove validation on site.address3 field](https://trello.com/c/LKiyzfLD/1765-remove-validation-on-siteaddress3-field)

defong:
Syncing it should be fine. 

Better check with @fofr existing data being invalid is one thing but going forward and allowing more invalid is another.

```
DfE
SW1P 3BT
```

```
20 Great Smith St
SW1P 3BT
```

vs

```
DfE
London
SW1P 3BT
```

```
20 Great Smith St
London
SW1P 3BT
```

Bear in mind that the the address is geo code, so not enforcing `town/city` may throw a wobble,

misaka:
This idea was seeded by @fofr originally ... re: syncing, the thing is that we already have a ton of sites that don't have `address3`, around 3371 out of 8257 total.

PR #608, opened 2019-07-12
[1564] re-add accrediting_provider_id on courses

misaka:
### Context

Just to keep the C# app happy for now, we'll remove it once we've deployed the
change to use accrediting_provider_code.

Re-add the `accrediting_provider_id` to the `courses` table until we've re-deployed the new C# changes to not use it. It's values will be populated using the associated `accrediting_provider` on a course.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/db/migrate/20190711074433_change_course_accrediting_provider_id_to_code.rb
+++ b/db/migrate/20190711074433_change_course_accrediting_provider_id_to_code.rb
@@ -30,15 +30,5 @@ class ChangeCourseAccreditingProviderIdToCode < ActiveRecord::Migration[5.2]
         end
       end
     end
-
-    revert do
-      add_column :course, :accrediting_provider_id, :integer
-      add_foreign_key :course,
-                      :provider,
-                      column: :accrediting_provider_id,
-                      name: "FK_course_provider_accrediting_provider_id"
-      add_index :course, %w[accrediting_provider_id],
-                name: "IX_course_accrediting_provider_id"
-    end
   end
 end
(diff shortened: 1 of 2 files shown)


timabell:
If it were me I'd modify the existing migration. I don't much like the idea of rolling out a breakage immediately followed by a fix. Your call.

misaka:
> If it were me I'd modify the existing migration. I don't much like the idea of rolling out a breakage immediately followed by a fix. Your call.

I think you're right, I was already starting to think the same thing myself.

db/migrate/20190711074433_change_course_accrediting_provider_id_to_code.rb
```diff
@@ -30,15 +30,5 @@ def change
...
-
-    revert do
-      add_column :course, :accrediting_provider_id, :integer
-      add_foreign_key :course,
-                      :provider,
-                      column: :accrediting_provider_id,
-                      name: "FK_course_provider_accrediting_provider_id"
-      add_index :course, %w[accrediting_provider_id],
-                name: "IX_course_accrediting_provider_id"
-    end
```
timabell: reverting the inverse of adding a revert. or something.
confusing but I think this is right to remove. qa will need resetting

misaka: Ya, I buggered QA by trying to alter it in-place. Doh.

misaka:
> Is it me or does that looks wrong

`revert` will revert the commands in the block there. So those commands were removing things like `accrediting_provider_id`. Is that not what we want?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_07_11_074433) do
+ActiveRecord::Schema.define(version: 2019_07_12_152312) do
```
defong: `ActiveRecord::Schema.define(version: 2019_07_12_152312) do` todays date?

misaka: Good spot! Fixed.

PR #627, opened 2019-07-18
[1797] Don't sync courses not in current recycle

tvararu:
### Context

This is to allow users to manage their course sites without seeing an error due to the sync currently failing.

Checks for the recruitment cycle in the controller. Adds a helper method to the model to check if it's current.

### Guidance to review

I fruitlessly attempted to write a request spec for this but could not figure out why the request would result in a 404 when I would set the provider to the 2020 cycle.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/recruitment_cycle_spec.rb
+++ b/spec/models/recruitment_cycle_spec.rb
@@ -13,7 +13,7 @@
 require 'rails_helper'
 
 describe RecruitmentCycle, type: :model do
-  subject { create(:recruitment_cycle, year: "2019") }
+  subject { RecruitmentCycle.find_by(year: "2019") }
 
   its(:to_s) { should eq("2019/20") }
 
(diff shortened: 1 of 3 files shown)


tvararu:
['Save' action update for managing locations](https://trello.com/c/BVftGS5s/1797-save-action-update-for-managing-locations)

spec/models/recruitment_cycle_spec.rb
```diff
@@ -13,7 +13,7 @@
 require 'rails_helper'
 
 describe RecruitmentCycle, type: :model do
-  subject { create(:recruitment_cycle, year: "2019") }
+  subject { RecruitmentCycle.find_by(year: "2019") }
```
tvararu: There's already a 2019 one; `creating` one as we did before would result in **two** 2019 recruitment cycles.

PR #635, opened 2019-07-22
[1826] Fix mismatch between api v1 docs and api v1 implementation

faissalMT:
### Context
https://github.com/DFE-Digital/manage-courses-backend/blob/master/docs/api.md#example-response-body-2  
This document specifies 'admin_contact' be the admin type however we are producing 'admin'.  

This pull request serialises the 'admin' type as 'admin_contact'.  

### Guidance to review
Because MCB's displayed types are coupled to the serialiser this has had he knock-on effect of changing the type displayed for MCB but I believe this is minor.  

### Checklist

- [x] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -95,13 +95,17 @@ private
 
   def generate_provider_contacts
     provider_contacts = object.contacts.map do |c|
-      c.attributes.slice('type', 'name', 'email', 'telephone')
+      contact = c.attributes.slice('type', 'name', 'email', 'telephone')
+      if c.type == 'admin'
+        contact['type'] = 'admin_contact'
+      end
+      contact
     end
 
-    has_admin_contact = provider_contacts.any? { |c| c['type'] == 'admin' }
+    has_admin_contact = provider_contacts.any? { |c| c['type'] == 'admin_contact' }
     unless has_admin_contact
       provider_contacts.prepend(
-        type: 'admin',
+        type: 'admin_contact',
         name: object.contact_name,
         email: object.email,
         telephone: object.telephone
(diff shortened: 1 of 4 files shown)


faissalMT:
[Fix mismatch between course contacts and documented API](https://trello.com/c/OvPYy1cf/1826-fix-mismatch-between-course-contacts-and-documented-api)

faissalMT:
This fix is just so that what's documented as being returned by v1 of the API matches what actually happens, the model shouldn't change.

app/serializers/provider_serializer.rb
```diff
@@ -95,13 +95,17 @@ def select_value_for_provider(provider_code, values)
 
   def generate_provider_contacts
     provider_contacts = object.contacts.map do |c|
-      c.attributes.slice('type', 'name', 'email', 'telephone')
+      contact = c.attributes.slice('type', 'name', 'email', 'telephone')
+      if c.type == 'admin'
```
defong: So is the data store as admin or admin_contact as the if looks like a workaround  code.

faissalMT: This is for all intents and purposes workaround code, the type field is stored as 'admin' and translated to 'admin_contact' on serialisation.

misaka: > So is the data store as admin or admin_contact as the if looks like a workaround code.

Ya, we decided to call it `admin` in our "clean" data model, but then forgot to document it that way. So now we're just adjusting it for the v1 api for UCAS.

defong: fine enough, workaround code it be

PR #636, opened 2019-07-22
[1747] Rename course enrichment qualifications field

dankmitchell:
### Context
Our course model has `qualifications` as well as `qualifications` from the enrichment model which is confusing and caused us to add some "workarounds"

Rename `qualifications` to `required_qualifications` in the course enrichment model

### Guidance to review
- Run the migration then check a course enrichment 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


dankmitchell:
[Fix inconsistent qualifications/required_qualifications field names](https://trello.com/c/BnKmktuz/1747-fix-inconsistent-qualifications-requiredqualifications-field-names)

app/models/course_enrichment.rb
```diff
@@ -32,7 +32,7 @@ class CourseEnrichment < ApplicationRecord
                  interview_process: [:string, store_key: 'InterviewProcess'],
                  other_requirements: [:string, store_key: 'OtherRequirements'],
                  personal_qualities: [:string, store_key: 'PersonalQualities'],
-                 qualifications: [:string, store_key: 'Qualifications'],
+                 required_qualifications: [:string, store_key: 'RequiredQualifications'],
```
defong: https://github.com/DFE-Digital/manage-courses-api/blob/7419131f3fb92e41c8f131281afaba96e2abb1ad/src/ManageCourses.Api/Model/CourseEnrichmentModel.cs#L16

are you sure?

defong:
This PR calls for babysitting @misaka

tvararu:
What @defong suggested makes sense; we could do this as a more increment change which is easier to deploy and less error prone

misaka:
We could also just focus on moving the background publishing job to Rails and then change it then.

misaka:
@dankmitchell Maybe this PR should just add a comment to not try changing the name of this attribute until kill C#. :)

db/migrate/20190722124603_rename_course_enrichment_qualifications.rb
```diff
@@ -0,0 +1,17 @@
+class RenameCourseEnrichmentQualifications < ActiveRecord::Migration[5.2]
+  def up
+    execute "UPDATE   course_enrichment
+                SET   json_data = Replace(json_data :: text, '\"Qualifications\":',
+                              '\"RequiredQualifications\":') ::
+                              jsonb
+              WHERE   json_data :: text LIKE '%\"Qualifications\":%';"
+  end
```
misaka: I'm a little confused by this. Isn't the model now expecting `required_qualifications` to be the JSON field `Qualifications`? And isn't this replacing `"Qualifications"` with `"RequiredQualifications"`? Which would then break `required_qualifications`?

defong: I don't think the migration file is needed. Dan probably forgot about it.

ie, Data should be stored as `qualification` and then get hydrated as `required_qualification`

misaka: Ok, so this needs to be removed, I think.

misaka:
:shipit:

PR #631, opened 2019-07-18
1804 Add not null validation error for gcse_requirments

davidgisbey:
### Context

**Should not be merged until nil data has been cleansed**   

The backend would currently update gcse_requirements to nil if passed them from the frontend. This is not desirable behaviour

### Changes proposed in this pull request
- Add validations to check to ensure nil cannot be passed in.
- Small refactor of request specs

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #642, opened 2019-07-24
[1804] Add validations for GCSE entry requirements

fofr:
### Context

A follow up to #631.

- Add validations to maths, english and science
  - Maths and english must always be set, science only on Primary courses
- New validations make some existing courses invalid, include a data migration to correct those courses (22 per cycle)
- Don't include "not_set" option for Maths and English when creating courses in MCB (this can't use the gcse_subjects_required method as the course level is determined by subjects which are only added after creation)
- DRY up the existing spec

There's a few gotchas found while working on this, which led to some frustrating rabbit holes:
- When a enum is assigned to nil, (not_set: nil), a `presence` validation won't work as expected – 'not_set' gets passed through and becomes nil after validations run
- When creating a course in MCB we do so before assigning subjects, subjects determine the course level, which in turn determine if science is required or not
- When trying to assign an invalid option not in the requirements enum, Rails throws an ArgumentError rather than using validations if present

### Guidance to review

- Review commit by commit
- Double check the data migration

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -125,6 +125,9 @@ class Course < ApplicationRecord
     ENTRY_REQUIREMENT_OPTIONS.reject { |option| option == :not_set }.keys.map(&:to_s)
   end
 
+  validates :maths,   inclusion: { in: entry_requirement_options_without_nil_choice }
+  validates :english, inclusion: { in: entry_requirement_options_without_nil_choice }
+  validates :science, inclusion: { in: entry_requirement_options_without_nil_choice }, if: :gcse_science_required?
   validates :enrichments, presence: true, on: :publish
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_enrichment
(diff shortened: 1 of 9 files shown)


fofr:
[Add validations to GCSE requirements](https://trello.com/c/OAmzR18z/1804-add-validations-to-gcse-requirements)

app/models/course.rb
```diff
@@ -370,6 +381,16 @@ def next_recruitment_cycle?
     recruitment_cycle.year > RecruitmentCycle.current_recruitment_cycle.year
   end
 
+  # Ideally this would just use the validation, but:
+  # https://github.com/rails/rails/issues/13971
+  def entry_requirements_assignable(course_params)
+    course_params.slice(:maths, :english, :science)
+      .to_h
+      .select { |_subject, value| value && !ENTRY_REQUIREMENT_OPTIONS.key?(value.to_sym) }
```
misaka: Should this be `value.valid?`? Could it be an empty string?

fofr: It can be nil but not an empty string. Passing in an empty string returns an error as expected.

PR #643, opened 2019-07-24
[1835] sync courses to find active job

misaka:
### Context

We need to background this task.

Use ActiveJob to sync a course to find. This will allow us to plugin a background job processor when ready.

### Guidance to review

This doesn't include a proper background job processor, but it's a step above what we have (it'll background the task to sync the course), even if it suffers from the flaw that it could lose jobs when the app is restarted. The next PR should add a background job processor, so we may want to couple the two PRs together for release.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,0 +1,5 @@
+class ApplicationJob < ActiveJob::Base
+  retry_on ActiveRecord::Deadlocked
+
+  discard_on ActiveJob::DeserializationError
+end
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -7,5 +7,5 @@ Rails.application.initialize!
 Rails.application.configure do
   config.action_controller.action_on_unpermitted_parameters = :raise
 
-  config.active_job.queue_adapter = :async
+  config.active_job.queue_adapter = :inline
 end
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -6,4 +6,6 @@ Rails.application.initialize!
 
 Rails.application.configure do
   config.action_controller.action_on_unpermitted_parameters = :raise
+
+  config.active_job.queue_adapter = :async
 end
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -23,26 +23,18 @@ module API
       end
 
       def sync_with_search_and_compare
-        if has_synced?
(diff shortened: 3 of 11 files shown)


misaka:
[Add job to publish course to find](https://trello.com/c/m8sj7qOt/1835-add-job-to-publish-course-to-find)

app/jobs/application_job.rb
```diff
@@ -0,0 +1,7 @@
+class ApplicationJob < ActiveJob::Base
+  # Automatically retry jobs that encountered a deadlock
+  retry_on ActiveRecord::Deadlocked
+
+  # Most jobs are safe to ignore if the underlying records are no longer available
+  discard_on ActiveJob::DeserializationError
```
tvararu: This file is so terse, Rails is good at this stuff. I don't think the comments add much 🤔

misaka: Fair 'nuf. They were boilerplate and I was feeling lazy. :)

misaka:
Just to add, I've tested this locally with both `inline` and `async` modes:

- **`inline`** - when I publish a course (after modifying it), I get an error triggered by the ActiveJob failing to connect with search-and-compare (I don't have that running locally)
- **`async`** - when I publish a course (after modifying it), I get no visible error through the front-end, but I see the same connection error as with `inline` in the logs

These behaviours are expected, since I'm not running `search-and-compare`, so I can't integration test this locally. If anyone wants to give it a whirl I'd appreciate it.

misaka:
Looking to voluntell someone to test this locally with search-and-compare running. @timabell? @defong? All you have to do is run mcbe, mcfe and the search-and-compare stack, and try publishing a course.

tvararu:
I have tested this locally with both search API and UI running, and can confirm that changes propagate correctly end-to-end. :+1:

misaka:
Adding "do not merge" as we need to change the queue mechanism to be `inline`.

config/environment.rb
```diff
@@ -6,4 +6,6 @@
 
 Rails.application.configure do
   config.action_controller.action_on_unpermitted_parameters = :raise
+
+  config.active_job.queue_adapter = :async
```
misaka: This should be `:inline`

richrace: This has now been set.

PR #645, opened 2019-07-24
[1813] delete a course

NyWeb:
### Context

* Allowing deletion of courses through the Ruby gem discard.
* :sweat_smile: damn hot https://www.youtube.com/watch?v=DpktBGInl60
* Scorchio :hot_pepper: https://www.youtube.com/watch?v=kZMUAd7OJc8

Pairing @timabell & @NyWeb 

- Include gem
- Used the gems callbacks for throwing error when a course can't be deleted
- Forced kept as a scope for provider
- Minor clean-up of test
- Added tests

### Guidance to review

* Simplest implementation of if `discard is allowed` was to throw an exception in `before_discard`
* Feedback on the test setup always welcomed.
* Turn off whitespace to see the change to `spec/requests/api/v2/providers/courses_spec.rb` more clearly (indented large json block because it's wrapped in a new block)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master x10
- [x] Cleaned commit history x100
- [x] Tested by running locally x20

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -24,6 +24,7 @@
 #
 
 class Course < ApplicationRecord
+  include Discard::Model
   include WithQualifications
   include ChangedAt
 
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -31,6 +31,10 @@ class Course < ApplicationRecord
 
   after_initialize :set_defaults
 
+  before_discard do
+    raise "You cannot delete a running course (Course: #{self}, Provider: #{provider_id})" if ucas_status != :new
+  end
+
   has_associated_audits
   audited
   validates :course_code, uniqueness: { scope: :provider_id }
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -11,6 +11,7 @@
 #  qualification             :integer          not null
 #  start_date                :datetime
 #  study_mode                :text
+#  accrediting_provider_id   :integer
 #  provider_id               :integer          default(0), not null
 #  modular                   :text
 #  english                   :integer
@@ -20,7 +21,7 @@
 #  updated_at                :datetime         not null
 #  changed_at                :datetime         not null
 #  accrediting_provider_code :text
-#  accrediting_provider_id   :integer
+#  discarded_at              :datetime
 #
 
 class Course < ApplicationRecord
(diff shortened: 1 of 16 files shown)


NyWeb:
[[S] Course - Soft Delete a course (MCBE)](https://trello.com/c/YyJxBHzh/1813-delete-a-course)

timabell:
Theses commits need a bit more cleaning up before we merge

app/models/course.rb
```diff
@@ -20,15 +21,20 @@
...
 
 class Course < ApplicationRecord
+  include Discard::Model
   include WithQualifications
   include ChangedAt
 
   after_initialize :set_defaults
 
+  before_discard do
+    raise "You cannot delete a running course" if ucas_status != :new
```
dankmitchell: Should we pass any course info into this error? i.e. name and code

NyWeb: Perhaps if we implement different ways of using it in the future, like bulk deleting. Otherwise, the system and user should be fully aware of what course it's calling `discard` on.

NyWeb: I guess for logging purposes, that'd be neat.

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -351,4 +365,54 @@ def perform_request
...
+      "/api/v2/providers/#{provider.provider_code}" +
+        "/courses/#{course.course_code}"
+    end
+
+    let(:course) do
+      course = create(:course, provider: provider)
+      create(:site_status, :new, course: course)
+      course
+    end
+
```
davidgisbey: If you change this to 

`
let(:course) { create(:course, provider: provider, site_statuses: [site_status) }
let(:site_status) { build(:site_status, :new) }
    
before do 
  course
end 
`
It won't create the extra course. It only creates the additional course if it is not being built into another object using build.

NyWeb: I wasn't aware, thanks a lot for that.

We might be talking past each other here though, in regards to the issue in the [Trello card](https://trello.com/c/ZHWQr2Wq/1839-spec-factories-remove-automatic-creation-of-course-from-sitestatus). I don't think this is creating an extra course. It's older specs unrelated to this PR that are accidentally creating extra courses.

app/models/course.rb
NyWeb: @dankmitchell providing provider_id and course name and code (Course.to_s). Will that be enough?

timabell:
Force push:

* filtered out merge of apv1 change and related patches (can be shipped independently)

PR #649, opened 2019-07-25
[1711] Edit the outcome of a course

davidgisbey:
### Context
Providers can't currently edit the outcome of a course
Goes with https://github.com/DFE-Digital/manage-courses-frontend/pull/483

- Adds the ability to edit the outcome of a course
- Adds validations to make sure it is updated to a valid qualification

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


fofr:
[Basic Details - Edit Outcomes (MCBE)](https://trello.com/c/DpJqEOdV/1711-basic-details-edit-outcomes-mcbe)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -167,7 +167,8 @@ def course_params
           .permit(
             :english,
             :maths,
-            :science
+            :science,
+            :qualification
```
defong: whats the diff between `qualification` vs `qualifications`

fofr: A subtle one. I mentioned it a bit more in the frontend PR:

> Removes the only frontend dependency on `course.qualifications` (`['qts', 'pgce']`), favouring `course.qualification` (`pgce_with_qts`).

Qualifications is an array.
Qualification is a string that combines the array.

Frontend will set the string, and use the string, and doesn't need the array – we store as array in case we need to do different things for QTS/PGCE in the future.

PR #641, opened 2019-07-23
[1822] Don't send courses marked 'new' to UCAS

faissalMT:
### Context
When UCAS requests course data we should not provide courses that have been marked 'new' as these haven't yet been published (and also have the potential to be discarded instead).

### Changes proposed in this pull request
This PR ensures that courses with a status of 'new' are not emitted by the v1 courses endpoint.  

### Guidance to review
This has the side effect of causing empty courses not to be sent, but this is correct behaviour. I had to correct tests that were dependent on the previous behaviour.

### Checklist

- [X] Make sure all information from the Trello card is in here
- [X] Attach to Trello card
- [X] Rebased parent
- [X] Cleaned commit history
- [X] Tested by running locally

PR #650, opened 2019-07-25
[1822] Don't send courses marked 'new' to UCAS

faissalMT:
### Context
When UCAS requests course data we should not provide courses that have been marked 'new' as these haven't yet been published (and also have the potential to be discarded instead).

This PR ensures that courses with a status of 'new' are not emitted by the v1 courses endpoint.  

### Guidance to review
This has the side effect of causing empty courses not to be sent, but this is correct behaviour. I had to correct tests that were dependent on the previous behaviour.  
This PR is a redo of #641 for merging into master.  

### Checklist

- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -121,6 +121,12 @@ class Course < ApplicationRecord
     end.order(:changed_at, :id)
   end
 
+  scope :not_new, -> do
+    includes(site_statuses: %i[site course])
+      .where
+      .not(SiteStatus.table_name => { status: SiteStatus.statuses[:new_status] })
+  end
+
   def self.entry_requirement_options_without_nil_choice
     ENTRY_REQUIREMENT_OPTIONS.reject { |option| option == :not_set }.keys.map(&:to_s)
   end
--- a/lib/tasks/govuk_lint.rake
+++ b/lib/tasks/govuk_lint.rake
@@ -3,7 +3,7 @@ namespace :lint do
   task :ruby do
     puts 'Linting...'
     unless system 'bundle exec rubocop app config db lib spec Gemfile --format clang -a'
-      exit $?
+      exit $?.exitstatus
     end
   end
 end
--- a/spec/requests/api/v1/course_spec.rb
+++ b/spec/requests/api/v1/course_spec.rb
@@ -25,7 +25,6 @@ describe "Courses API", type: :request do
                         enrichments: [])
     end
 
-
     context "without changed_since parameter" do
       before do
         site = FactoryBot.create(:site, code: "-", location_name: "Main Site", provider: provider)
@@ -123,16 +122,22 @@ describe "Courses API", type: :request do
       end
 
       it 'includes correct next link in response headers' do
-        create(:course,
-               course_code: "LAST1",
-               age: 10.minutes.ago,
-               provider: provider)
+        Timecop.freeze(10.minutes.ago) do
+          first_course = create(:course,
+                                course_code: 'LAST1',
+                                provider: provider)
+
+          create(:site_status, :published, course: first_course)
+        end
 
         timestamp_of_last_course = 2.minutes.ago
-        _last_course_in_results = create(:course,
-                                         course_code: "LAST2",
-                                         age: timestamp_of_last_course,
-                                         provider: provider)
+
+        Timecop.freeze(timestamp_of_last_course) do
+          last_course_in_results = create(:course,
+                                          course_code: 'LAST2',
+                                          provider: provider)
+          create(:site_status, :published, course: last_course_in_results)
+        end
 
         get '/api/v1/2019/courses',
             headers: { 'HTTP_AUTHORIZATION' => credentials }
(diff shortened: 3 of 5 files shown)


faissalMT:
[Implement deleted-course behaviour in APIv1 for UCAS](https://trello.com/c/kNRLJZQy/1822-implement-deleted-course-behaviour-in-apiv1-for-ucas)

app/models/course.rb
```diff
@@ -121,6 +121,12 @@ def new_draft_attributes
     end.order(:changed_at, :id)
   end
 
+  scope :not_new, -> do
```
timabell: Ronseal, does what it says on the tin. :+1:

lib/tasks/govuk_lint.rake
```diff
@@ -3,7 +3,7 @@ namespace :lint do
   task :ruby do
     puts 'Linting...'
     unless system 'bundle exec rubocop app config db lib spec Gemfile --format clang -a'
-      exit $?
+      exit $?.exitstatus
```
timabell: I think the commit message should explain why this change was necessary

spec/requests/api/v1/course_spec.rb
```diff
@@ -123,16 +122,22 @@ def get_course_codes_from_body(body)
...
-               course_code: "LAST1",
-               age: 10.minutes.ago,
-               provider: provider)
+        Timecop.freeze(10.minutes.ago) do
+          first_course = create(:course,
+                                course_code: 'LAST1',
+                                provider: provider)
+
+          create(:site_status, :published, course: first_course)
+        end
```
timabell: Re: the diff on this file; it's not clear to me from the commit message why this patch on this file is so large. I'm sure it's a good improvement, but I can't tell from reading it without a long time getting into your head why it was needed or desired for this particular PR.

Generally I'd suggest either:

* explanation in the commit message as to why you needed/wanted to improve the file
* overall improvments to the spec that are not specific to the task or that are improvements to be built on should be pulled out into a separate base commit or maybe even a separate PR

spec/factories/site_statuses.rb
```diff
@@ -22,6 +22,7 @@
     transient do
       any_vancancy { false }
       provider { build :provider }
+      age { nil }
```
defong: whats the `age` for?

spec/requests/api/v1/course_spec.rb
```diff
@@ -398,5 +410,31 @@ def get_next_courses(link, params = {})
...
+      let(:course2) { create(:course, study_mode: 'full_time', profpost_flag: 'postgraduate', program_type: 'higher_education_programme', provider: provider1) }
+
+      let(:status1) { create(:site_status, status: :new_status, course: course1, vac_status: 'full_time_vacancies', applications_accepted_from: Date.new(2019, 7, 23)) }
+      let(:status2) { create(:site_status, status: :new_status, course: course2, vac_status: 'full_time_vacancies', applications_accepted_from: Date.new(2019, 7, 24)) }
+      let(:status3) { create(:site_status, status: :running, course: course2, vac_status: 'full_time_vacancies', applications_accepted_from: Date.new(2019, 7, 24)) }
+
+      it 'does not send courses marked new' do
+        status1
+        status2
+        status3
```
timabell: use `let!` instead of `let` to cause immediate execution of the block, this will eliminate the need for calling them at the start of the `it` block

spec/factories/site_statuses.rb
faissalMT: It's a remnant of a fix I was attempting, I ended up just going with TimeCop in the end.

app/models/course.rb
```diff
@@ -121,6 +121,12 @@ def new_draft_attributes
...
+    includes(site_statuses: %i[site course])
+      .where
+      .not(SiteStatus.table_name => { status: SiteStatus.statuses[:new_status] })
```
```suggestion
      .not(SiteStatus.table_name => { status: :new_status })
```

Can it not be like that?

spec/requests/api/v1/course_spec.rb
faissalMT: These changes were done because the tests break without it. In these tests the courses have no `site_statuses` and so this PR causes those courses not to show, since this is correct behaviour it means those courses must be given `site_statuses`. Since creating these associations updates certain things I've used TimeCop to make those guarantees.  
I've improved the commit message to clarify.

app/models/course.rb
faissalMT: https://stackoverflow.com/questions/36487900/rails-using-enum-in-a-where-query

spec/requests/api/v1/course_spec.rb
defong: or a before block, its confusing looking at

```ruby

it 'test something' do
do_lots_of_setup

do_actual_thing

expect(actual).to eq (do_lots_of_setup)

```

PR #652, opened 2019-07-25
[1734] Add completions for MCB

faissalMT:
### Context
For the convenience of users, especially those new to MCB, we should provide a way for commands to be completed.  

Add tab autocomplete for main commands.  

### Guidance to review  
Currently stderr spills into the terminal during tests, I'm not sure how to disable it in `with_stubbed_stdout`. This also only performs completions for main commands as Readline is not context-aware.  

### Checklist

- [x] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history


faissalMT:
[Add command completion to mcb repl](https://trello.com/c/asf8rjaD/1734-add-command-completion-to-mcb-repl)

PR #656, opened 2019-07-26
[1843] mcb specs fails intermittently due to pseudo-random test-data

NyWeb:
### Context
MCB spec randomly fails as string checks where a bit wobbly. We're using Faker for generating strings (programming languages). D & C gives false positives, and there are only so many languages.

This got me and @timabell many times when we were trying to push forward the Delete Courses Backend work, and left us rerunning TravisCI over and over, delaying the eventual merge. 

Gave explicit names to courses to give fewer false positives, making us

### Guidance to review
I could've created more randomness in the factory too, but that seemed like the wrong way.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/lib/mcb/commands/courses/list_spec.rb
+++ b/spec/lib/mcb/commands/courses/list_spec.rb
@@ -11,75 +11,76 @@ describe 'mcb providers list' do
     let(:provider2) { create(:provider, recruitment_cycle: current_cycle) }
     let(:provider3) { create(:provider, recruitment_cycle: additional_cycle) }
 
-    let(:course1) { create(:course, provider: provider1) }
-    let(:course2) { create(:course, provider: provider2) }
-    let(:course3) { create(:course, provider: provider3) }
+    # Explicitly setting the course names, Course Factory isn't random enough
+    let(:course_present) { create(:course, provider: provider1, name: 'English') }
+    let(:course_present2) { create(:course, provider: provider2, name: 'Maths') }
+    let(:course_not_present) { create(:course, provider: provider3, name: 'Swedish') }
 
     it 'lists all courses for the default recruitment cycle' do
-      course1
-      course2
-      course3
+      course_present
+      course_present2
+      course_not_present
 
       command_output = list.execute[:stdout]
 
-      expect(command_output).to include(course1.course_code)
-      expect(command_output).to include(course1.name)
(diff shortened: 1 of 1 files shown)


NyWeb:
[[S] Travis randomly fails on mcb courses list spec](https://trello.com/c/Q2kyykcw/1843-s-travis-randomly-fails-on-mcb-courses-list-spec)

defong:
ummm what exactly is the problem?
Why does hard coding the name solve its?

defong:
Faker::Lorem.characters(10) for the length if that is the issue

NyWeb:
Thanks for looking @defong!

The factory for courses uses `name { Faker::ProgrammingLanguage.name }` - i.e. programming language names. C occurs, as does D, which gives plenty of false positives. There's also only so many, so courses may end up with the same names.

We could have a more random string generator from faker too, but that *could* give false positive matches too. Here it's more explicit what we're testing.

spec/lib/mcb/commands/courses/list_spec.rb
```diff
@@ -11,75 +11,71 @@
...
+      course_present
+      course_not_present
 
       command_output = list.execute[:stdout]
 
-      expect(command_output).to include(course1.course_code)
-      expect(command_output).to include(course1.name)
-
-      expect(command_output).to include(course2.course_code)
-      expect(command_output).to include(course2.name)
```
misaka: I think the reason this was here originally is to prove that the list contains _all_ the courses, by default (as per the name of the test). Was there a specific reason you removed it?

NyWeb: Adding it back! Mistook it to look like it tested the same thing twice.

defong:
Oh right, gotcha, can we leave a comment there as well in case we stumble on it and changed it etc.

And maybe use a more real life examples.

ie actual course name

A course can be called `English` a provider has multiple Course called `English`

The comb for uniqueness is Provider code and course code.

ie

xxx, yyyy, English
xxx, zzzz, English

yyy, yyyy, English
yyy, zzzz, English

NyWeb:
Thanks guys! I'll incorporate your comment @defong

NyWeb:
Tidied up commits, that's why the fp

PR #658, opened 2019-07-26
[1844] Add new age ranges to course

dankmitchell:
### Context
Better age range for courses

Add new age range course for V2 so that we don't impact V1(UCAS)

### Guidance to review
I see us handling this in the same way we do for course length, have several hardcoded radion buttons on the frontend and just send a string value to the backend. An ENUM would have been nicer but the "other" makes this difficult.

🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/providers/courses_spec.rb
+++ b/spec/requests/api/v2/providers/courses_spec.rb
@@ -23,7 +23,8 @@ describe 'Courses API v2', type: :request do
            enrichments: [enrichment],
            maths: :must_have_qualification_at_application_time,
            english: :must_have_qualification_at_application_time,
-           science: :must_have_qualification_at_application_time)
+           science: :must_have_qualification_at_application_time,
+           age_range_in_years: '3 to 7')
   }
 
   let(:courses_site_status) {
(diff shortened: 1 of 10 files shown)


dankmitchell:
[Add age range field to basic details (MCBE)](https://trello.com/c/C5nRx1qg/1844-add-age-range-field-to-basic-details-mcbe)

spec/requests/api/v2/providers/courses_spec.rb
```diff
@@ -23,7 +23,8 @@
            enrichments: [enrichment],
            maths: :must_have_qualification_at_application_time,
            english: :must_have_qualification_at_application_time,
-           science: :must_have_qualification_at_application_time)
+           science: :must_have_qualification_at_application_time,
+           age_range_in_years: '3 to 7')
```
defong: should it be an enum instead of string?

dankmitchell: Did you read the above? 👀

defong: opps :face_with_head_bandage:

PR #660, opened 2019-07-26
[1792] Add provider publish and publishable actions

tvararu:
### Context

These will be called by the frontend to publish the latest provider enrichment.

Adds `#publish` and `#publishable` actions to the provider controller. `#publish` is a no-op.

### Guidance to review

Does not actually implement the publish logic, that will be done as a separate PR; this is just to facilitate the frontend validations work.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider_enrichment.rb
+++ b/app/models/provider_enrichment.rb
@@ -46,6 +46,11 @@ class ProviderEnrichment < ApplicationRecord
   validates :train_with_us, words_count: { maximum: 250 }
   validates :train_with_disability, words_count: { maximum: 250 }
 
+  validates :email, :website, :telephone,
+            :address1, :address2, :address3, :address4,
+            :postcode, :train_with_us, :train_with_disability,
+            presence: true, on: :publish
+
   def has_been_published_before?
     last_published_at.present?
   end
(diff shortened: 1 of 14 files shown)


tvararu:
[Provider #publish and #publishable](https://trello.com/c/CFEeSKyP/1851-provider-publish-and-publishable)

spec/requests/api/v2/providers/publish_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+               data: {
+                 attributes: {},
+                 type: "provider"
+               }
+             }
+           }
+      response
+    end
+
+    context 'when unauthenticated' do
```
fofr: We should DRY up these specs if they're needed across multiple tests.

app/models/provider_enrichment.rb
```diff
@@ -46,6 +46,17 @@ class ProviderEnrichment < ApplicationRecord
   validates :train_with_us, words_count: { maximum: 250 }
   validates :train_with_disability, words_count: { maximum: 250 }
 
+  validates :email, presence: true, on: :publish
```
dankmitchell: Could we put these into a one-liner or is there future benefit from splitting them out?

tvararu: One-liner should be fine 👍

spec/requests/api/v2/providers/publish_spec.rb
tvararu: Will do 👍

PR #664, opened 2019-07-29
[1851] Remove validation for provider enrichment address2

tvararu:
I accidentally added this when I refactored something in this area.


tvararu:
[Provider #publish and #publishable](https://trello.com/c/CFEeSKyP/1851-provider-publish-and-publishable)

PR #666, opened 2019-07-30
🗡 fix dev role creation for db:create (and dropping for db:drop) 

misaka:
### Context

We've taken a couple stabs :hocho: at fixing how we create the use role when we create the db, esp on dev. :imp: 

* Create the user before the dbs are created, and then set the user's password afterwards. This sequence seems to be the right way to get things done.
* Unholy deeds :latin_cross: 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [x] Bug squashed

### DIFF
--- a/lib/tasks/prepare_postgres.rake
+++ b/lib/tasks/prepare_postgres.rake
@@ -1,34 +1,41 @@
 desc "Prepare db for dev environments"
 namespace :db do
-  task create: [:prepare_postgres]
+  task create: [:create_dev_role]
 
-  task :prepare_postgres do
-    if package_missing? 'postgres'
-      case RUBY_PLATFORM
-      when /darwin/
-        darwin_install_package 'postgres'
+  task :create_dev_role do
+    if exec_present? 'createuser'
+      system "createuser -d -s -w #{username}"
+      puts "Created user '#{username}'"
+    else
+      puts <<~EOTEXT
+        No 'createuser' command. Please ensure postgres is installed:
 
-      else
-        puts <<~EOTEXT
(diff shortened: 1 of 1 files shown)


lib/tasks/prepare_postgres.rake
```diff
@@ -48,13 +55,27 @@ private
...
 
-  def package_missing?(package)
-    command = 'which ' + package
-
-    `#{command}`[0] != '/'
+  def exec_present? name
+    system("which #{name} 2>&1 >/dev/null")
   end
 
   def darwin_install_package(package)
```
NyWeb: As we're moving away from a `one-click-installer`, let's delete this method to not have it slushing around perhaps?

NyWeb: As we're moving away from a "one-click-installer", let's delete the method `darwin_install_package` to not have it slushing around perhaps?

misaka: Ya, I had a problem when it tried to install postgres for me:

a) I already had `postgresql@9.6` installed, which matches the version in production, so I didn't really want the latest Postgres installed.
b) it hung for some reason and didn't complete, for some reason ... maybe brew doesn't like running that way? Not sure.

I agree we should probably just remove the method.

NyWeb: I tested it, and it just gave an error message in the build log, but still succeeded - seems floppy, so better to remove it. Mac devs should know how to use brew, and if not, it's in the error msg.

PR #667, opened 2019-07-30
[1641] Added additional tests to illustrate manage-courses-api integration

defong:
### Context
Manage course api calling manage courses backend

Pass thro user info in order to allow for authenticated calls to backend.

Api will be proxying an authenticated user to backend similar aspect as to the frontend.

### Guidance to review

Api user is already authenticated the test is to illustrate the expected behaviour of api generating a token that can be consumed by backend. 

Related to 
https://github.com/DFE-Digital/manage-courses-api/pull/325

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/controllers/api/v2/application_controller_spec.rb
+++ b/spec/controllers/api/v2/application_controller_spec.rb
@@ -2,10 +2,11 @@ require 'rails_helper'
 
 describe API::V2::ApplicationController, type: :controller do
   describe '#authenticate' do
+    let(:secret) { Settings.authentication.secret }
     let(:encoded_token) do
       JWT.encode(
         payload,
-        Settings.authentication.secret,
+        secret,
         Settings.authentication.algorithm
       )
     end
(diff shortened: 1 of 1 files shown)


defong:
[[M] Switch provider publishing to use mc-be in mc-api](https://trello.com/c/mCSPuez8/1641-m-switch-provider-publishing-to-use-mc-be-in-mc-api)

spec/controllers/api/v2/application_controller_spec.rb
```diff
@@ -2,10 +2,13 @@
 
 describe API::V2::ApplicationController, type: :controller do
   describe '#authenticate' do
+    let(:secret) {
```
faissalMT: When putting a block on multiple lines it's Ruby convention to use `do` `end`, this isn't too long though so I'd just put it on one line.

spec/controllers/api/v2/application_controller_spec.rb
```diff
@@ -41,5 +42,33 @@
...
+      let(:existing_user) { create(:user, email: email, sign_in_user_id: sign_in_user_id) }
+      let(:secret) { Settings.authentication.secret = "SETTINGS:MANAGE_BACKEND:SECRET" }
+
+      before do
+        existing_user
+        Settings.authentication.secret = "SETTINGS:MANAGE_BACKEND:SECRET"
+      end
+
+      # NOTES:
+      # This is hardcode as the ruby version of JWT encoding is not asymmetrical with csharp and does not applies
```
misaka: > ... is not asymmetrical ...

Does this mean it is symmetrical? That might be easier to read.

spec/controllers/api/v2/application_controller_spec.rb
```diff
@@ -41,5 +42,33 @@
           .to eq('Token realm="Application"')
       end
     end
+
+    context 'manage course api integration' do
```
misaka: Is "manage course api integration" a description of the state of the system? Sounds like what we're testing, so `describe` would be more appropriate.

spec/controllers/api/v2/application_controller_spec.rb
```diff
@@ -41,5 +42,33 @@
...
+      end
+
+      # NOTES:
+      # This is hardcode as the ruby version of JWT encoding is not symmetrical with csharp and does not applies
+      #
+      # RECOMMENDED (Notational Conventions) for the "typ" (Type) Header Parameter see (https://tools.ietf.org/html/rfc7519#section-5.1)
+      # as default.
+      # In any case as long as it can be verified it is fine.
+      let(:encoded_token) do
+        "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzaWduX2luX3VzZXJfaWQiOiJtYW5hZ2VfY291cnNlc19hcGkiLCJlbWFpbCI6Im1hbmFnZV9jb3Vyc2VzQGFwaS5jb20ifQ.fs0tM5Lc_5vA2w94PhlDMnq50NUn2K5SMxggdAApp_w"
```
misaka: How is this generated? It would be really nice to see that in the spec so we can understand it.

defong: > How is this generated?

https://github.com/DFE-Digital/manage-courses-api/pull/325/commits/6286fef6d2d55729301394bee07df77607ad1838

As explained 
`This is hardcode as the ruby version of JWT encoding is not symmetrical with csharp and does not applies...`

ie 
```
      JWT.encode(
        payload,
        Settings.authentication.secret,
        Settings.authentication.algorithm
      )
```

`JWT.encode` == `_manageCoursesBackendJwtService.GetCurrentUserToken()`

should of been equal except left hand side do `too little` and the right hand side do `too much`

configuring them to do the exact requires too much as the actual structure and ordering of items needs to exact, ie the header's items and the payload items

misaka: Am I right to think that mcbe needs to send this across to mcapi (in the `Authorization` by the looks of it)? How does mcbe get it's hands on it? Is it in app's settings / environment?

defong: > Am I right to think that mcbe needs to send this across to mcapi (in the Authorization by the looks of it)?

No

defong: Its the otherway round...

mcapi will call mcbe 
mcapi http call Authorization header will contain the token in itself has generated

mcbe will authorise the call

misaka: 👍

PR #677, opened 2019-08-01
[1874] Use latest enrichment for provider contact info

tvararu:
### Context

If we use the latest published enrichment, the frontend will not display the latest values when we are in the editing flow.

Use `enrichments.last` instead of `latest_published_enrichment`.

Update any relevant tests as this already has quite a bit of coverage.

### Guidance to review

:shipit:

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu:
[Provider enrichments: Updating isn't working correctly](https://trello.com/c/hvkKRNd1/1874-provider-enrichments-updating-isnt-working-correctly)

defong:
> when we are in the editing flow.

Does it affects any other flow for frontend?

Do front end need to know the difference between a published vs a draft version, particularly on the course preview?

tvararu:
@defong it does afect course previewing. Discussed IRL with @misaka and @fofr and there is some follow-up work we can do to make this better, but for now it's okay.

PR #680, opened 2019-08-02
[1808] accrediting provider enrichments

timabell:
### Context

Full read/create/update for accrediting provider enrichments.

Because of all the questions about feasibility & shape of update when working on just the list of names/codes we gave up on splitting the card in three and this PR is for the whole feature.

This is api work is to support re-implement https://bat-qa-mcui-as.azurewebsites.net/organisation/1dq/about in rails

* `providers#show`: Include code/name, & enrichment if available, of accrediting bodies for a provider.
* `providers#update`: Update enrichments in database when accredited body info is posted as part of provider#update

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Test coverage for read
- [x] Test coverage updating existing enrichment

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -29,6 +29,7 @@ module API
       def update
         authorize @provider, :update?
         update_enrichment
+        update_accrediting_enrichment
 
         if @provider.valid?
           render jsonapi: @provider.reload, include: params[:include]
@@ -103,21 +104,61 @@ module API
         @user = User.find(params[:user_id])
       end
 
+      def update_accrediting_enrichment
+        return if accredited_bodies_params.values.none?
+
+        enrichment = @provider.enrichments.find_or_initialize_draft
+
+        enrichment.accrediting_provider_enrichments =
+          accredited_bodies_params["accredited_bodies"].map do |accredited_body|
+            {
+              UcasProviderCode: accredited_body["provider_code"],
+              Description: accredited_body["description"],
+            }
+          end
+
+        set_provider_code enrichment
+        enrichment.save
+      end
+
       def update_enrichment
         return unless enrichment_params.values.any?
 
         enrichment = @provider.enrichments.find_or_initialize_draft
         enrichment.assign_attributes(enrichment_params)
 
+        set_provider_code enrichment
+        enrichment.save
+      end
+
+      def set_provider_code(enrichment)
         # Note: provider_code is only here to support c# counterpart, until provide_code is removed from database
         enrichment.provider_code = @provider.provider_code if enrichment.provider_code.blank?
-        enrichment.save
+      end
+
+      def accredited_bodies_params
+        params
+          .fetch(:provider, {})
+          .except(
+            :train_with_us,
+            :train_with_disability,
+            :email,
+            :telephone,
+            :website,
+            :address1,
+            :address2,
+            :address3,
+            :address4,
+            :postcode,
+            :region_code
+          )
+          .permit(accredited_bodies: %i[provider_code provider_name description])
       end
 
       def enrichment_params
         params
           .fetch(:provider, {})
-          .except
+          .except(:accredited_bodies)
           .permit(
             :train_with_us,
             :train_with_disability,
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -273,10 +273,9 @@ class Provider < ApplicationRecord
 
   def accredited_bodies
     accrediting_providers.map do |ap|
-      accrediting_provider_entry = latest_enrichment&.accrediting_provider_enrichments&.find do |ape|
-        ape['UcasProviderCode'] == ap.provider_code
-      end
+      accrediting_provider_entry = latest_enrichment&.accrediting_provider_enrichment(ap.provider_code)
 
+      # map() to this hash:
       {
         provider_name: ap.provider_name,
         provider_code: ap.provider_code,
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -99,7 +99,7 @@ class Provider < ApplicationRecord
            primary_key: :provider_code,
            inverse_of: :accrediting_provider
 
-
+  has_many :accrediting_providers, -> { distinct }, through: :courses
 
   scope :changed_since, ->(timestamp) do
     if timestamp.present?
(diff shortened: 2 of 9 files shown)


timabell:
[Expose accredited provider name/code list in backend api](https://trello.com/c/lhmNp5HF/1808-expose-accredited-provider-name-code-list-in-backend-api)

app/controllers/api/v2/providers_controller.rb
```diff
@@ -114,10 +133,29 @@ def update_enrichment
...
+            :region_code
+          )
+          .permit(accredited_bodies: %i[provider_code provider_name description])
+      end
+
       def enrichment_params
         params
           .fetch(:provider, {})
-          .except
+          .except(:accredited_bodies)
```
misaka: `accredited_bodies` is also being permitted below, shouldn't it be one or the other?

misaka:
Looking good! I didn't delve into the specs yet, will wait for them to be done, but the implementation looks good.

app/controllers/api/v2/providers_controller.rb
```diff
@@ -103,6 +104,24 @@ def get_user
         @user = User.find(params[:user_id])
       end
 
+      def update_accrediting_enrichment
+        return unless accredited_bodies_params.values.any?
```
misaka: Ruby also provides a `none?` that might be better here:

```suggestion
        return if accredited_bodies_params.values.none?
```

tvararu: TIL 😲

misaka: Checking what versions of ruby has it:

```
|       | `Enumerable#none?` | `Array#none?` |
| 2.5.x |         ✅         |       ❌      |
| 2.6.x |         ✅         |      ✅       | 
```

defong: tests passes, @misaka  vindicated... should really check 1.8 it was so much less fluffy...

timabell:
Squashed, rebased & force-pushed

app/controllers/api/v2/providers_controller.rb
```diff
@@ -103,6 +104,24 @@ def get_user
         @user = User.find(params[:user_id])
       end
 
+      def update_accrediting_enrichment
+        return if accredited_bodies_params.values.none?
+
+        enrichment = @provider.enrichments.find_or_initialize_draft
+
+        accredited_body_enrichments = accredited_bodies_params["accredited_bodies"].map do |accredited_body|
```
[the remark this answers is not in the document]
timabell: good spot, done & squashed

app/models/provider.rb
```diff
@@ -271,6 +271,20 @@ def publishable?
     valid? :publish
   end
 
+  def accredited_bodies
+    accrediting_providers.map do |ap|
+      accrediting_provider_entry = latest_enrichment&.accrediting_provider_enrichments&.find do |ape|
```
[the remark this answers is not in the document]
timabell: factored out for readability, I don't think there's a dupe check in there, it was confusing to read as it's all inside a `map()`

misaka:
Just did a test to see how slow this is on what I think is the worst-case scenario: a provider with 9 accredited bodies:

```
  Provider Load (0.7ms)  SELECT  "provider".* FROM "provider" WHERE "provider"."id" = $1 LIMIT $2  [["id", 11637], ["LIMIT", 1]]
  RecruitmentCycle Load (0.4ms)  SELECT  "recruitment_cycle".* FROM "recruitment_cycle" WHERE "recruitment_cycle"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Provider Load (44.2ms)  SELECT DISTINCT "provider".* FROM "provider" INNER JOIN "course" ON "provider"."provider_code" = "course"."accrediting_provider_code" WHERE "course"."provider_id" = $1 AND "course"."discarded_at" IS NULL AND "provider"."recruitment_cycle_id" = $2  [["provider_id", 11637], ["recruitment_cycle_id", 1]]
  ProviderEnrichment Load (0.7ms)  SELECT  "provider_enrichment".* FROM "provider_enrichment" WHERE "provider_enrichment"."provider_id" = $1 LIMIT $2  [["provider_id", 11637], ["LIMIT", 1]]
```

@timabell I don't see any reason to worry about optimising this, yet.

app/models/provider_enrichment.rb
```diff
@@ -58,4 +58,10 @@ def has_been_published_before?
   def publish(current_user)
     update(status: 'published', last_published_at: Time.now.utc, updated_by_user_id: current_user.id)
   end
+
+  def accrediting_provider_enrichment(provider_code)
+    accrediting_provider_enrichments&.find do |enrichment|
+      enrichment['UcasProviderCode'] == provider_code
```
misaka: Some `accrediting_provider_enrichments` don't include the key `UcasProviderCode` here:

```
[33] pry(main)> Provider.find(12233).latest_enrichment.accrediting_provider_enrichments
  Provider Load (40.8ms)  SELECT  "provider".* FROM "provider" WHERE "provider"."id" = $1 LIMIT $2  [["id", 12233], ["LIMIT", 1]]
  ProviderEnrichment Load (0.3ms)  SELECT  "provider_enrichment".* FROM "provider_enrichment" WHERE "provider_enrichment"."provider_id" = $1 LIMIT $2  [["provider_id", 12233], ["LIMIT", 1]]
=> [{"Description"=>
   "Essex and Thames Primary SCITT is one of the UK's most established providers, having sustained an excellent reputation for 20 years.  Working in partnership with Canterbury Christ Church University the courses offer excellent value for money and an opportunity to train with some of the very best in the sector."}]
```

This isn't the first one I've seen, either.

misaka:
I'm happy with what's here, aside from handling the cases where there is no `UcasProviderCode`.

app/controllers/api/v2/providers_controller.rb
```diff
@@ -103,21 +104,61 @@ def get_user
...
         return unless enrichment_params.values.any?
 
         enrichment = @provider.enrichments.find_or_initialize_draft
         enrichment.assign_attributes(enrichment_params)
 
+        set_provider_code enrichment
+        enrichment.save
+      end
+
+      def set_provider_code(enrichment)
```
misaka: Can we have this as an `after_create` hook on the model? Can be a new ticket, of course.

timabell: https://trello.com/c/3KknUZq1/1903-refactor-move-setprovidercodeenrichment-to-hook

app/models/provider_enrichment.rb
timabell: test coverage added for this case

timabell: data investigation https://trello.com/c/jlK06Apv/1912-investigate-bad-data-in-accrediting-provider-enrichments

PR #681, opened 2019-08-02
[1858]  Update a courses subjects 

davidgisbey:
### Context

A provider needs the ability to edit  a courses subject. 

- Allow the course update endpoint to update subjects. 

### Guidance to review

After speaking to Misha it seems like this may not need updates as the drop down logic should sit on the front end. Am open to suggestions.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -60,6 +60,7 @@ module API
         update_course
         update_enrichment
         update_sites
+        update_subjects
         should_sync = site_ids.present? && @course.recruitment_cycle.current?
         has_synced? if should_sync
 
@@ -105,6 +106,12 @@ module API
         @course.errors[:sites] << "^You must choose at least one location" if site_ids.empty?
       end
 
+      def update_subjects
+        return if subject_params.nil?
+
+        @course.subjects = Subject.where(subject_name: subject_params)
+      end
+
       def build_provider
         @provider = @recruitment_cycle.providers.find_by!(
           provider_code: params[:provider_code].upcase
@@ -134,7 +141,8 @@ module API
                   :maths,
                   :science,
                   :qualification,
-                  :age_range_in_years)
+                  :age_range_in_years,
+                  :subjects)
           .permit(
             :about_course,
             :course_length,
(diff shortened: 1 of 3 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -180,6 +189,10 @@ def course_params
           )
       end
 
+      def subject_params
+        params.fetch(:course, {})[:subjects]
```
defong: Something like that 

. Permit(subjects : [])

benilovj:
I’m not convinced that we should be exposing the current `Subject` class for editing, as this is a legacy data structure that’s not in line with our application’s designs.

We have `DFESubject` which is in line with our designs, but at the moment we have logic to map between the two concepts.

I would propose to bite the bullet and move to the DFE subject taxonomy, rather than persist with the legacy taxonomy. Here’s a sketch of what that could look like technically:

- create a reference database table for dfe subjects and populate it with the subjects that are on Find
- re-visit the mapping logic and simplify it (do away with mappings that depend on course titles, for instance) - @fofr is up for that
- store the mapping between Subject and DFESubject in the database using foreign keys
- get rid of the mapping ruby code
- keep returning Subject thru the V1 API

cc @misaka

fofr:
I vote for doing the simplification work now.

app/controllers/api/v2/courses_controller.rb
davidgisbey: Discussed with Misha. If you do it the way you suggest you have to exclude all the other attributes (somewhere around 30). If you just want 1 param returned fetch seems preferable.

spec/requests/api/v2/providers/courses/update_subject_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      it "does not change the subjects" do
+        expect(course.reload.subjects).to eq(subjects[:subjects])
+      end
+    end
+    context "with no values passed into the params" do
+      let(:subjects) { {} }
+
+      before do
+        @subjects = course.subjects
```
faissalMT: I'd avoid using instance variables for tests

defong:
can we close this for the time being, till we simplify it? @davidgisbey

PR #683, opened 2019-08-06
[1798] mcb providers create

richrace:
### Context

Need to provide a command via `mcb` to create a Provider (and organisation).

- Add create command for Providers 
- Automatically set `accrediting_provider` and `scitt` based on Provider Type
- `User.admins` scope

### Guidance to review

- ~~Should this belong here: https://github.com/DFE-Digital/manage-courses-backend/pull/683/files#diff-5c08de53fa6bea147e31fe1d5930422aR30-R32~~

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/recruitment_cycle.rb
+++ b/app/models/recruitment_cycle.rb
@@ -28,6 +28,10 @@ class RecruitmentCycle < ApplicationRecord
     all.order(:year).second
   end
 
+  def next
+    RecruitmentCycle.find_by(year: year.to_i + 1)
+  end
+
   def current?
     RecruitmentCycle.current_recruitment_cycle == self
   end
--- a/app/models/recruitment_cycle.rb
+++ b/app/models/recruitment_cycle.rb
@@ -25,7 +25,7 @@ class RecruitmentCycle < ApplicationRecord
   end
 
   def self.next_recruitment_cycle
-    all.order(:year).second
+    current_recruitment_cycle.next
   end
 
   def next
--- a/lib/mcb/provider_editor.rb
+++ b/lib/mcb/provider_editor.rb
@@ -78,6 +78,16 @@ module MCB
       # add god users to the org, for any that aren't already in it
       organisation.users << (User.admins - organisation.users)
 
+      provider.sites.create(
+        location_name: @cli.ask_name_of_first_location,
+        address1: provider.address1,
+        address2: provider.address2,
+        address3: provider.address3,
+        address4: provider.address4,
+        postcode: provider.postcode,
+        region_code: provider.region_code
+      )
+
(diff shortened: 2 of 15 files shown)


richrace:
[mcb providers create](https://trello.com/c/dXGRMUmo/1798-mcb-providers-create)

benilovj:
Looking good! I'd still be tempted to move 

```
provider.scheme_member = 'Y'
provider.year_code = provider.recruitment_cycle.year
...
provider.changed_at = Time.now.utc
```

into the `Provider` model but I don't think it's a deal-breaker for me.

app/models/recruitment_cycle.rb
```diff
@@ -28,6 +28,10 @@ def self.next_recruitment_cycle
     all.order(:year).second
   end
 
+  def next
+    RecruitmentCycle.find_by(year: year.to_i + 1)
+  end
```
misaka: Even though this is the right implementation, I feel like this muddies the waters a bit ... wrt `RecruitmentCycle#next_recruitment_cycle`. I know this is a pretty big PR already, but is there anything simple we can do to not have this split-view of what qualifies as the "next" cycle?

lib/mcb/provider_editor.rb
```diff
@@ -69,7 +171,8 @@ def update(attrs)
     end
 
     def check_authorisation
-      raise Pundit::NotAuthorizedError unless ProviderPolicy.new(@requester, @provider).update?
+      action = @provider.persisted? ? :update? : :create?
+      raise Pundit::NotAuthorizedError unless ProviderPolicy.new(@requester, @provider).send(action)
```
misaka: Minor improvement, but this use's Pundit's built-in mechanism to find `ProviderPolicy`.

```suggestion
      raise Pundit::NotAuthorizedError unless Pundit.policy(@requester, @provider).send(action)
```

app/models/recruitment_cycle.rb
benilovj: Would it help if we changed:

```ruby
  def self.next_recruitment_cycle
    current_recruitment_cycle.next
  end
```

misaka: Ya, I thought that would work too. 👍

lib/mcb/provider_editor.rb
benilovj: fixed and force-pushed.

app/models/recruitment_cycle.rb
benilovj: Right-o, fixed in 4d25757

PR #684, opened 2019-08-06
[1905] expose applications accepted from

danielburnley:
## Context

Courses have a date that applications are accepted from. This date is currently modelled onto the `SiteStatus` instead of the course, so it's duplicated a lot.

- Add new attribute `applications_open_from`, replacing the current method which pulls from the site statuses
- Default to next recruitment cycle
- Update current references to `applications_open_from`

### Guidance to review

Am open to suggestions on any changes/incorrect assumptions I have made given I'm still getting my head around the data model :tada:

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/db/migrate/20190806102256_add_applications_open_from_to_courses.rb
+++ b/db/migrate/20190806102256_add_applications_open_from_to_courses.rb
@@ -1,0 +1,16 @@
+class AddApplicationsOpenFromToCourses < ActiveRecord::Migration[5.2]
+  def change
+    add_column :course, :applications_open_from, :date
+
+    say_with_time 'Setting applications_open_from to one held within sites' do
+      Course.all.each do |course|
+        if course.recruitment_cycle.current?
+          applications_open_from = course.site_statuses.order("applications_accepted_from ASC").first.applications_accepted_from
+          course.update!(applications_open_from: applications_open_from)
+        else
+          course.update!(applications_open_from: course.recruitment_cycle.application_start_date)
+        end
+      end
+    end
+  end
+end
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -23,6 +23,7 @@
 #  accrediting_provider_code :text
 #  discarded_at              :datetime
 #  age_range_in_years        :string
+#  applications_open_from    :date
 #
 
 require 'rails_helper'
@@ -563,26 +564,6 @@ RSpec.describe Course, type: :model do
     end
   end
 
-  describe "#applications_open_from=" do
-    let(:provider) { build(:provider, sites: [sites]) }
-    let(:sites) { build_list(:site, 3) }
-    let!(:existing_site_status) {
-      sites.each do |site|
-        create(:site_status,
-               :running,
-               site: site,
-               course: subject,
-               applications_accepted_from: Date.new(2018, 10, 9))
-      end
-    }
-
-    it "updates the applications_accepted_from date on the site statuses" do
-      expect { subject.applications_open_from = Date.new(2018, 10, 23) }.
-        to change { subject.reload.site_statuses.pluck(:applications_accepted_from).uniq }.
-        from([Date.new(2018, 10, 9)]).to([Date.new(2018, 10, 23)])
-    end
-  end
-
   describe "adding and removing sites on a course" do
     let(:provider) { build(:provider) }
       #this code will be removed and fixed properly in the next pr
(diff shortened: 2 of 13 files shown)


fofr:
I prefer `applications_open_from` to `applications_accepted_from`. "accepted" is a term that can be misleading – "applications accepted" could be interpreted to refer to a successful application rather than a received one.

db/migrate/20190806102256_add_applications_accepted_from_to_courses.rb
```diff
@@ -0,0 +1,15 @@
+class AddApplicationsAcceptedFromToCourses < ActiveRecord::Migration[5.2]
+  def change
+    add_column :course, :applications_accepted_from, :date
+
+    say_with_time 'Setting applications_accepted_from to one held within sites' do
+      Course.all.each do |course|
+        if course.recruitment_cycle.current?
+          course.update(applications_accepted_from: course.site_statuses.first.applications_accepted_from)
```
fofr: We don't want the first status’s open date, we want the earliest in the set of statuses.

db/migrate/20190806102256_add_applications_open_from_to_courses.rb
```diff
@@ -0,0 +1,16 @@
+class AddApplicationsOpenFromToCourses < ActiveRecord::Migration[5.2]
+  def change
+    add_column :course, :applications_open_from, :date
+
+    say_with_time 'Setting applications_open_from to one held within sites' do
+      Course.all.each do |course|
+        if course.recruitment_cycle.current?
+          applications_open_from = course.site_statuses.order("applications_accepted_from ASC").first.applications_accepted_from
```
defong: shouldn't it be the whole of this?

```
  site_statuses
      .open_for_applications
      .order("applications_accepted_from ASC")
      .first
      &.applications_accepted_from
```

this gives the site status that is open for application as it is findable and application being accepted now with vacancies

spec/models/course_spec.rb
```diff
@@ -1056,4 +1036,21 @@
...
+  describe '#applications_open_from' do
+    context 'a new course with a given date' do
+      let(:applications_open_from) { Date.today }
+      let(:subject) { create(:course, applications_open_from: applications_open_from) }
+
+      its(:applications_open_from) { should eq applications_open_from }
+    end
+
+    context 'a new course within a recruitment cycle' do
+      let(:recruitment_cycle) { create :recruitment_cycle, :next }
```
```suggestion
      let(:recruitment_cycle) { build :recruitment_cycle, :next }
```

spec/models/course_spec.rb
```diff
@@ -1056,4 +1036,21 @@
...
+      let(:provider)          { create :provider, recruitment_cycle: recruitment_cycle }
```
```suggestion
      let(:provider)          { build :provider, recruitment_cycle: recruitment_cycle }
```

db/migrate/20190806102256_add_applications_open_from_to_courses.rb
defong: nvm
https://github.com/DFE-Digital/manage-courses-backend/pull/684#discussion_r311033152

danielburnley: Theo and I chatted about it but we made the assumption that the check as to whether or not they're open isn't necessary, though this was back when we were creating a new field for it. 

Given we're now replacing the field - it would probably be worth keeping the current behaviour, i'll update this :tada:

db/migrate/20190806102256_add_applications_open_from_to_courses.rb
```diff
@@ -0,0 +1,16 @@
...
+          course.update(applications_open_from: applications_open_from)
+        else
+          course.update(applications_open_from: course.recruitment_cycle.application_start_date)
+        end
+      end
```
misaka: I forget what Rails does when you try to roll this back, does it work (i.e. just drop the column and not complain)?

danielburnley: Yep - it'll just drop it silently, this doesn't change any existing data so we don't need to worry about creating a more in-depth up/down migration

misaka: Also, as you pointed out yourself, we should probably use `update!` here.

db/migrate/20190806102256_add_applications_open_from_to_courses.rb
defong: https://trello.com/c/l0t1Oun3/1699-applications-open-from-shouldnt-be-restricted-to-just-sites-that-are-open-to-applications

spec/serializers/api/v2/serializable_course_spec.rb
```diff
@@ -24,7 +25,7 @@
   it { should have_attribute :ucas_status }
   it { should have_attribute :funding }
   it { should have_attribute :subjects }
-  it { should have_attribute :applications_open_from }
+  it { should have_attribute(:applications_open_from).with_value(date_today.to_s) }
```
defong: can you do the same for `start_date` since its now set

PR #687, opened 2019-08-07
1837 add background job arm template

rizzkhan1:
### Context
Automate deployment of background jobs docker images onto an Azure Container Instance (ACI). ARM template has been created to deploy the ACI infrastructure using Azure Devops Release pipeline. 

New additional ARM template for for background job creation. This ARM template has no dependency and does not replace existing template for Backend app services (template.json). 

### Guidance to review
New file created under Azure folder

### DIFF
--- a/azure/backgroundJobTemplate.json
+++ b/azure/backgroundJobTemplate.json
@@ -8,6 +8,12 @@
         "description": "The name of the Azure Container Instances resource. This will be its identifier in Azure and can be different from the image name."
       }
     },
+    "containerGroupName": {
+      "type": "string",
+      "metadata": {
+        "description": "The name of the Azure Container Instances resource. This will be its identifier in Azure and can be different from the image name."
+      }
+    },
     "authenticationToken": {
       "type": "string",
       "metadata": {
@@ -92,7 +98,6 @@
     "deploymentUrlBase": "https://raw.githubusercontent.com/DFE-Digital/bat-platform-building-blocks/130-create-a-release-pipeline-for-background-web-app/templates/"
   },
   "resources": [
-
     {
       "name": "container-instances",
       "type": "Microsoft.Resources/deployments",
@@ -105,6 +110,9 @@
         },
         "parameters": {
           "containerName": {
+            "value": "[parameters('containerName')]"
+          },
+          "containerGroupName": {
             "value": "[parameters('containerName')]"
           },
           "imageName": {
--- a/azure/backgroundJobTemplate.json
+++ b/azure/backgroundJobTemplate.json
@@ -89,7 +89,7 @@
     }
   },
   "variables": {
-    "deploymentUrlBase": "https://raw.githubusercontent.com/DFE-Digital/bat-platform-building-blocks/ARM-Template-for-Azure-Container-Instances/templates/"
+    "deploymentUrlBase": "https://raw.githubusercontent.com/DFE-Digital/bat-platform-building-blocks/130-create-a-release-pipeline-for-background-web-app/templates/"
   },
   "resources": [
 
--- a/azure/backgroundJobTemplate.json
+++ b/azure/backgroundJobTemplate.json
@@ -81,12 +81,6 @@
         "description": "Auth secret for manage back-end."
       }
     },
-    "settingsManageApiSecret": {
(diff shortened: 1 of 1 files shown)


defong:
do you need all those parameters

EwanNoble:
Secrets should probably be passed in as `securestring`s

rizzkhan1:
> do you need all those parameters
No, only 3 parameters are required to create an Azure container instance. Most of the parameters are for the application environment variables. 

> Secrets should probably be passed in as `securestring`s
Secret are being passed through pipeline variables in a clear text and are converted to secure values under environmental variables section (not parameter). This is identical to all other web apps application settings. Using secure string will double encrypt the secret and application may not support that.

misaka:
> > do you need all those parameters
> 
> No, only 3 parameters are required to create an Azure container instance. Most of the parameters are for the application environment variables.

We've gotten all of these because we've done a wholesale copy of the `mcbe` app ... which kind of makes sense, although:

1. most of these aren't (probably) won't be needed, so we could reduce clutter by stripping them out
2. for the ones that are needed, if they ever change we'll need to ensure we update both apps

We could get two birds stoned at once if we found a way to re-use values between `mcbe` and this app, how much work would that be?

rizzkhan1:
> 1. most of these aren't (probably) won't be needed, so we could reduce clutter by stripping them out 
> 2. for the ones that are needed, if they ever change we'll need to ensure we update both apps

Happy to declutter them once we know which ones aren't needed.
I could merge mcbe and mcbg arm templates so we won't have to change values in multiple places but then we lose lots of flexibility in terms of release pipeline and deployments.

misaka:
@rizzkhan1 Do we take the values for these params from a central place that is shared with `mcbe` at least? That would probably be good enough, in which case I'm not sure I mind the clutter for the sake of expediency.

rizzkhan1:
@misaka values are passed from AzureDevops pipelines, most of them have been grouped so they are quite easy to manage.

azure/backgroundJobTemplate.json
```diff
@@ -0,0 +1,177 @@
...
+            "value": "[parameters('imageName')]"
+          },
+          "environmentVariables": {
+            "value": [
+              {
+                "name": "RAILS_ENV",
+                "value": "[parameters('railsEnv')]"
+              },
+              {
+                "name": "AUTHENTICATION_TOKEN",
```
misaka: I just had a look through the code for how this is used, and it's only used in authenticating requests for API V1. If we remove this then I *think* the authentication token will be blank, and while we don't expose a way to reach API V1 here, if we ever did using this configuration we'd be leaking data.

At the risk of being paranoid, I say we just leave it here. What we really outta be doing is setting this kind of thing to a secure string (read: random) by default, with overrides per environment.

azure/backgroundJobTemplate.json
```diff
@@ -0,0 +1,177 @@
...
+              {
+                "name": "SECRET_KEY_BASE",
+                "secureValue": "[parameters('secretKeyBase')]"
+              },
+              {
+                "name": "SENTRY_DSN",
+                "secureValue": "[parameters('sentryDSN')]"
+              },
+              {
+                "name": "SETTINGS__AUTHENTICATION__SECRET",
```
misaka: This one, like `AUTHENTICATION_TOKEN`, is used for communicating with an API ... V2 in this case. Like that one, I say we leave this one until we have it set to a reasonably secure default.

azure/backgroundJobTemplate.json
```diff
@@ -0,0 +1,177 @@
...
+                "secureValue": "[parameters('settingsAuthenticationSecret')]"
+              },
+              {
+                "name": "SETTINGS__MANAGE_API__SECRET",
```
misaka: This is used when communicating with `manage-courses-api`. I believe the only outgoing connection we'll make in this app will be to search-and-compare so this is one we can probably get rid of.

azure/backgroundJobTemplate.json
rizzkhan1: We are already using this configuration in rest of the other ARM templates. Authentication token value goes in application settings which takes only string values.

PR #689, opened 2019-08-07
[1798] Make sure an Org Name has been entered via MCB providers create

richrace:
### Context

Providers were created with an Organisation without a name.

Before: Was able to create/assign Providers to organisation with an empty name
After: Must provide an Organisation name to create a new Provider

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/provider_cli.rb
+++ b/lib/mcb/provider_cli.rb
@@ -40,7 +40,7 @@ module MCB
     end
 
     def ask_organisation_name
-      @cli.ask("What organisation should the new provider be added to?  ")
+      @cli.ask("What Organisation should the new provider be added to? (Enter Org Name)  ")
     end
 
     def ask_name_of_first_location
--- a/lib/mcb/provider_editor.rb
+++ b/lib/mcb/provider_editor.rb
@@ -88,22 +88,22 @@ module MCB
 
     def find_or_create_organisation
       finished_picking_organisation = false
-      org_name_input = nil
 
-      until finished_picking_organisation && org_name_input.present?
-        org_name_input = @cli.ask_organisation_name
+      until finished_picking_organisation
+        organisation_name = @cli.ask_organisation_name
 
-        if org_name_input.present?
-          organisation = Organisation.find_or_initialize_by(name: org_name_input)
-
-          if organisation.persisted?
-            finished_picking_organisation = true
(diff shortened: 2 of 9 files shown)


richrace:
[mcb providers create](https://trello.com/c/dXGRMUmo/1798-mcb-providers-create)

benilovj:
CodeClimate is complaining about complexity; I think it's possible to fix this issue without adding more nesting:

```ruby
    def find_or_create_organisation
      finished_picking_organisation = false
      until finished_picking_organisation
        organisation_name = @cli.ask_organisation_name

        if organisation_name.blank?
          puts "Organisation name cannot be blank."
          next
        end

        organisation = Organisation.find_or_initialize_by(name: organisation_name)
        if organisation.persisted?
          finished_picking_organisation = true
        elsif organisation.new_record? && @cli.confirm_new_organisation_needed?
          organisation.save!
          finished_picking_organisation = true
        end
      end

      # connect provider to org
      provider.organisations << organisation

      organisation
    end
```

lib/mcb/provider_cli.rb
```diff
@@ -40,7 +40,7 @@ def ask_region_code
     end
 
     def ask_organisation_name
-      @cli.ask("What organisation should the new provider be added to?  ")
+      @cli.ask("What Organisation should the new provider be added to? (Enter Org Name)  ")
```
timabell: Like the new prompt. I was confused by the question before

timabell:
Like it, caveat Jake's comment.

lib/mcb/provider_editor.rb
```diff
@@ -88,13 +88,22 @@ def try_saving_provider
...
+          organisation = Organisation.find_or_initialize_by(name: org_name_input)
+
+          if organisation.persisted?
+            finished_picking_organisation = true
+          elsif organisation.new_record? && @cli.confirm_new_organisation_needed?
+            organisation.save!
+            finished_picking_organisation = true
+          end
+        else
+          puts "Organisation name cannot be blank."
```
defong: Just wondering should this be in the Organisation model itself as 

`validate :name presence`

richrace: Will to @benilovj suggestion and add this validation. 

So we have a nicer UX and stop accepting any potential bad data.

PR #691, opened 2019-08-08
[1915] Update course start date

davidgisbey:
### Context
Providers need to able to update the start date of their course.

- Adds the ability to update the attribute on the update course endpoint.
- Adds validations to ensure the start date is in the correct cycle

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu:
Should we include the possible `start_date`s in the course response metadata so that the frontend doesn't need to infer them from scratch?

app/serializers/api/v2/serializable_course.rb
```diff
@@ -67,7 +68,8 @@ def enrichment_attribute(name, enrichment_name = name)
 
       meta do
         {
-          edit_options: @object.edit_options.all
+          edit_options: @object.edit_options.all,
+          start_dates: start_dates(@object.recruitment_cycle.year)
```
fofr: This should live in the edit_options object.

markmead:
[[S] Edit 'Course Starts' - (MCBE)](https://trello.com/c/YW4bRhrP/1915-s-edit-course-starts-mcbe)

app/serializers/api/v2/serializable_course.rb
davidgisbey: Done.  Thanks for the feedback.

app/models/concerns/start_date_valid.rb
```diff
@@ -0,0 +1,16 @@
+module StartDateValid
+  extend ActiveSupport::Concern
+  included do
+    def start_date_valid?(recruitment_year, course)
+      if valid_start_date_range(recruitment_year).cover?(course.start_date)
+        true
+      else
+        course.errors.add :start_date, "#{start_date.strftime('%b %Y')} is not in the #{recruitment_cycle.year} cycle"
```
fofr: Should validation just be checking that `start_date.strftime('%b %Y')` is one of the options that we pass to frontend?
Otherwise we have start date logic in two places.

fofr: An existing example:
https://github.com/DFE-Digital/manage-courses-backend/blob/df8ec83c6e1d2e2be63701a620bbd98d9a01842a/app/models/course.rb#L460

davidgisbey: Good point. This logic pre-existed the additional work to expose the list via the metadata. I'll refactor it to check validate against the list sent.

davidgisbey: And done! Thanks Paul.

app/models/edit_course_options.rb
```diff
@@ -36,7 +36,25 @@ def all
...
+     "November #{recruitment_year}",
+     "December #{recruitment_year}",
+     "January #{recruitment_year + 1}",
+     "February #{recruitment_year + 1}",
+     "March #{recruitment_year + 1}",
+     "April #{recruitment_year + 1}",
+     "May #{recruitment_year + 1}",
+     "June #{recruitment_year + 1}",
+     "July #{recruitment_year + 1}"]
+  end
```
misaka: First time I'm seeing this class. It feels like it should be simplified into a concern (separately, of course), thoughts?

spec/lib/mcb/commands/courses/show_spec.rb
```diff
@@ -22,6 +22,7 @@
   let(:course) { create(:course, name: 'P', provider: provider, subjects: [subject1], site_statuses: [site_status1]) }
   let(:rolled_over_course) do
     new_course = course.dup
+    new_course.update(start_date: course.start_date - 1.year)
```
misaka: Ugh, I don't think we should be having to do this to fix this test. We should be able to create a course that is in the next recruitment cycle instead of having to manually roll-over a course. Probably a new card at this point.

misaka: I've just fixed this and pushed up a commit. About the same effort as making a ticket.

misaka:
Is this work part of 1856 or 1915? Looks like it should be 1915 ...

davidgisbey:
Yea it should be 1915. This card got copied for applications open from. I think the original card for rewritten over for applications open from. Hence the card number being wrong.

spec/lib/mcb/commands/courses/show_spec.rb
davidgisbey: Cool. Looks good to me. I'll reflect these changes in the applications open PR too 👍

app/models/course.rb
```diff
@@ -457,4 +462,10 @@ def validate_qualification
   def set_applications_open_from
     self.applications_open_from ||= recruitment_cycle.application_start_date
   end
+
+  def validate_start_date
+    if provider.present?
```
defong: i dont think a course can live without a provider...

app/models/edit_course_options.rb
davidgisbey: Misha and I have agreed this will be completed in a separate PR once this one is merged in.

tvararu:
🚢

app/models/course.rb
defong: silly factory bot issue....

app/models/course.rb
```diff
@@ -149,6 +149,7 @@ def self.entry_requirement_options_without_nil_choice
   validate :validate_enrichment
   validate :validate_course_syncable, on: :sync
   validate :validate_qualification, on: :update
+  validate :validate_start_date, on: :update
```
```suggestion
  validate  :validate_qualification, :validate_start_date, on: :update
```

https://apidock.com/rails/ActiveRecord/Validations/ClassMethods/validate

davidgisbey: Done 👍

PR #678, opened 2019-08-02
[1837] add background jobs for publish

misaka:
### Context

We need to add a background worker to process jobs to sync courses to search-and-compare.

### Changes proposed in this pull request

Add `delayed_job`, configure it for the different environments, and modify Docker to build it so that it can be deployed.

### Guidance to review

This PR is stacked ontop of #643 which was not being merged until we could demonstrate that this whole thing works. That PR should be merged into master once this one is approved, and then this one rebased.

### Checklist

- [ ] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #693, opened 2019-08-08
[1836] Use background process to sync

richrace:
### Context

Add job to publish a single provider's courses to find.

Use ActiveJob to sync

### Guidance to review

I do not have the whole stack running to test this but it is based heavily from https://github.com/DFE-Digital/manage-courses-backend/pull/643

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/providers/sync_courses_with_search_and_compare_spec.rb
+++ b/spec/requests/api/v2/providers/sync_courses_with_search_and_compare_spec.rb
@@ -83,16 +83,10 @@ describe 'Courses API v2', type: :request do
           end
 
           it 'should make the appropriate request' do
-            subject
+            perform_enqueued_jobs do
+              subject
+            end
             expect(stubbed_request).to have_been_requested
-          end
-        end
-
-        context 'when an unsuccessful external call to search has been made' do
-          let(:status) { 451 }
-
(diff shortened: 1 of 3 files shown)


richrace:
[[S] Add job to publish provider's courses to find and run :inline](https://trello.com/c/Fjw7R37k/1836-s-add-job-to-publish-providers-courses-to-find-and-run-inline)

spec/requests/api/v2/providers/sync_courses_with_search_and_compare_spec.rb
```diff
@@ -83,18 +83,12 @@
           end
 
           it 'should make the appropriate request' do
-            subject
+            perform_enqueued_jobs do
+              subject
+            end
```
misaka: #678 adds some extra tests to check that an actual request is made out by the job, what do you think about adding these here?

https://github.com/DFE-Digital/manage-courses-backend/pull/678/files#diff-eed410faa3db2404572332f7ea75e7b2R40-R47

I know that this is potentially mixing in integration testing with request specs, but it'd be pretty convenient to put them here and doesn't feel completely out of place. Also, if we don't do them in this PR we're not likely to add them anywhere else.

PR #704, opened 2019-08-12
[1932] move send from subject to course

richrace:
### Context

In able to easily edit a Course's SEND state. We need to remove the legacy implementation of SEND as a subject.

What we use for SEND for the C# App: https://github.com/DFE-Digital/manage-courses-backend/blob/c976d4ab6da8224db2a35b0e5499410c86627661/app/serializers/search_and_compare/course_serializer.rb#L52 

And the Subjects we pass for the C# App: https://github.com/DFE-Digital/manage-courses-backend/blob/c976d4ab6da8224db2a35b0e5499410c86627661/app/serializers/search_and_compare/course_serializer.rb#L226-L238

~~So I don't think there would be any impact to the C# App.~~

- Add test to API V1 to ensure no regressions
- Add DB Column `is_send` as boolean to Course
- Update tests to reflect this new behaviour 

### Guidance to review

Are microservices effected by this change? 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -24,6 +24,7 @@
 #  discarded_at              :datetime
 #  age_range_in_years        :string
 #  applications_open_from    :date
+#  is_send                   :boolean          default(false)
 #
 
 class Course < ApplicationRecord
@@ -276,10 +277,6 @@ class Course < ApplicationRecord
 
   def level
     Subjects::CourseLevel.new(subjects.map(&:subject_name)).level
-  end
-
-  def is_send?
-    subjects.any?(&:is_send?)
   end
 
   def is_fee_based?
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -24,7 +24,7 @@
 #  discarded_at              :datetime
 #  age_range_in_years        :string
 #  applications_open_from    :date
-#  is_send                   :boolean          default(false)
+#  is_send                   :boolean          default(FALSE)
 #
 
 class Course < ApplicationRecord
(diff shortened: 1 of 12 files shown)


richrace:
[Remove SEND (U3) Subject and add it as attribute to Course](https://trello.com/c/1HMd3Qmk/1932-remove-send-u3-subject-and-add-it-as-attribute-to-course)

fofr:
Really happy to see us beginning the move away from the UCAS model to our own. Nice work :100:

app/controllers/api/v1/courses_controller.rb
```diff
@@ -41,6 +43,17 @@ def build_recruitment_cycle
           year: params[:recruitment_year]
         ) || RecruitmentCycle.current_recruitment_cycle
       end
+
+      # Course now has a `is_send` attribute so we do not need to model `SEND` courses using the
+      # Subject. However, API V1 is still expecting the Subject so we add it back in.
+      def build_send_subjects
```
defong: should this not be in the v1 serializer?

timabell:
I wonder if it might make sense to squash `[1932] Refactor adding SEND Subject from Controller to serializer` into the previous commit

timabell:
@richrace  Can you comment in here on what effect (or lack of) this will have on the not-yet-dead c# in manage-courses-api?

timabell:
Usage in the c# publish-to-find thing https://github.com/DFE-Digital/manage-courses-api/blob/473c046cfd8d1e2ce1e8d046fdf493b35851857c/src/ManageCourses.Domain/Models/Course.cs#L37

timabell:
Matching c# change is good to go https://github.com/DFE-Digital/manage-courses-api/pull/328

PR #706, opened 2019-08-13
[1916] Add Audits to Provider Enrichments

richrace:
### Context

Attribute `accrediting_provider_enrichments` was being added to the audits twice. Once at the top level and once within the `json data`

Do not duplicate attribute in audit

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider_enrichment.rb
+++ b/app/models/provider_enrichment.rb
@@ -23,7 +23,9 @@ class ProviderEnrichment < ApplicationRecord
 
   belongs_to :provider,
              inverse_of: 'enrichments'
-  audited associated_with: :provider
+
+  audited except: :json_data,
+          associated_with: :provider
 
   scope :latest_created_at, -> { order(created_at: :desc) }
   scope :latest_published_at, -> { order(last_published_at: :desc) }
--- a/spec/models/provider_enrichment_spec.rb
+++ b/spec/models/provider_enrichment_spec.rb
@@ -22,13 +22,37 @@ describe ProviderEnrichment, type: :model do
   end
 
   describe 'auditing' do
-    it { should be_audited }
+    it { should be_audited.except(:json_data) }
 
     it 'can link to Provider' do
       provider = create(:provider)
       provider_enrichment = create(:provider_enrichment, provider: provider)
 
       expect(provider_enrichment.audits.last.associated).to eq(provider)
+    end
+
+    context 'when dealing with a `jsonb_accessor` attribute' do
+      let(:accrediting_provider_enrichment) do
+        { "UcasProviderCode" => "ABC", "Description" => "A test value" }
+      end
+      let(:provider_enrichment) { create(:provider_enrichment) }
+
+      before do
+        provider_enrichment.accrediting_provider_enrichments = [accrediting_provider_enrichment]
+        provider_enrichment.save
+      end
+
+      it 'does not audit `json_data` attribute' do
+        expect(provider_enrichment.audits.last.audited_changes).to_not have_key('json_data')
+      end
+
+      it 'does include `accrediting_provider_enrichments` at top level of audited_changes' do
+        changes = provider_enrichment.audits.last.audited_changes
+        expect(changes['accrediting_provider_enrichments']).to include(
+          nil,
+          [include(accrediting_provider_enrichment)]
+        )
+      end
     end
   end
 


richrace:
[[S] Update provider audit trail when changing provider enrichment](https://trello.com/c/BVL0yPPD/1916-s-update-provider-audit-trail-when-changing-provider-enrichment)

spec/models/provider_enrichment_spec.rb
```diff
@@ -22,14 +22,45 @@
...
+              'AccreditingProviderEnrichments' => nil
+            ),
+            hash_including(
+              'AccreditingProviderEnrichments' => [accrediting_provider_enrichment]
+            )
+          ]
+        )
+      end
+
+      it 'does not inlcude `accrediting_provider_enrichments` at top level of audited_changes' do
```
defong:
```suggestion
      it 'does not include `accrediting_provider_enrichments` at top level of audited_changes' do
```

spec/models/provider_enrichment_spec.rb
```diff
@@ -22,14 +22,45 @@
...
       provider = create(:provider)
       provider_enrichment = create(:provider_enrichment, provider: provider)
 
       expect(provider_enrichment.audits.last.associated).to eq(provider)
     end
+
+    context 'when dealing with `accrediting_provider_enrichments`' do
+      let(:accrediting_provider_enrichment) do
+        {
+          "errors" => {}, "Description" => "A test value", "validation_context" => nil
```
defong:
```suggestion
          "UcasProviderCode" => "ABC", "Description" => "A test value"
```
That is more aligned to real life


wondering `AccreditingProviderEnrichment::ArraySerializer.load().dump()`

app/models/provider_enrichment.rb
```diff
@@ -23,7 +23,9 @@ class ProviderEnrichment < ApplicationRecord
 
   belongs_to :provider,
              inverse_of: 'enrichments'
-  audited associated_with: :provider
+
+  audited except: :accrediting_provider_enrichments,
+          associated_with: :provider
```
misaka: Why do we need to exclude `accrediting_provider_enrichments`?

richrace: It gets set twice. Once via the JSON Data and then as an attribute.

https://trello.com/c/BVL0yPPD/1916-s-update-provider-audit-trail-when-changing-provider-enrichment#comment-5d5286285da65f384429af02

misaka: Is this true only for this `jsonb_accessor` attribute? Don't they all do this?

richrace: Ahh. It does like it's any `jsonb_accessor`

misaka: So is it worth excluding `jsonb_accessor` then?

richrace: Yes, looking at that now.

spec/models/provider_enrichment_spec.rb
```diff
@@ -22,14 +22,41 @@
...
+      it 'does not audit `json_data` attribute' do
+        expect(provider_enrichment.audits.last.audited_changes).to_not have_key('json_data')
+      end
+
+      it 'does include `accrediting_provider_enrichments` at top level of audited_changes' do
+        changes = provider_enrichment.audits.last.audited_changes
+        expect(changes['accrediting_provider_enrichments']).to_not eq([
+          nil,
+          accrediting_provider_enrichment
+        ])
```
misaka: Would it be worth making this less brittle?

```
        expect(changes['accrediting_provider_enrichments']).to_not include([
           nil,
           accrediting_provider_enrichment
         ])
```

misaka: Wait, I'm missing something here. The spec says "_does_ include `accrediting_provider_enrichments` ...", how is the `to_not eq([...])` testing that?

richrace: Good catch - sorry about that.

misaka: Harking back to another discussion about `let`, and your comment about `its`, that's one nice thing about using an `its` here (but not saying we should change it now):

```
      its(['accrediting_provider_enrichments']) { should include(nil,[accrediting_provider_enrichment]) }
```

There would be no chance for a miss-match here. :)

spec/models/provider_enrichment_spec.rb
```diff
@@ -22,14 +22,38 @@
...
+
+      it 'does not audit `json_data` attribute' do
+        expect(provider_enrichment.audits.last.audited_changes).to_not have_key('json_data')
+      end
+
+      it 'does include `accrediting_provider_enrichments` at top level of audited_changes' do
+        changes = provider_enrichment.audits.last.audited_changes
+        expect(changes['accrediting_provider_enrichments']).to include(
+          nil,
+          [include(accrediting_provider_enrichment)]
```
misaka: Why is `include` in here? Should it be:

```suggestion
          accrediting_provider_enrichment
```

richrace: `accrediting_provider_enrichments` is an array. And the Data stored to the DB also includes  `errors: {}, validation_context: nil`. So make it less brittle.

```ruby
expect(changes['accrediting_provider_enrichments']).to include(
  nil,
  accrediting_provider_enrichment
)

expected [nil, [{"Description" => "A test value", "UcasProviderCode" => "ABC"}]] to include {"UcasProviderCode" => "ABC", "Description" => "A test value"}
Diff:

 [
   nil,
-   {
-     "UcasProviderCode" => "ABC",
-     "Description" => "A test value"
-   }
+   [{ "Description" => "A test value", "UcasProviderCode" => "ABC" }]
 ]
````

misaka: Cool. Didn't realise the nested `include` did that. Nice one. 👍

PR #707, opened 2019-08-13
[1944] Add provider enrichment details validations

danielburnley:
### Context

C# has validations for the email and telephone numbers for provider enrichments - we need to add these to the ruby application.

image

- Validate email address and telephone formats when they are updated by the user
  - These are not updated on create/all updates as they'd be triggered by updating values that are not on the details page, e.g. About your organisation
  - The error messages have been taken from [The GOV.UK Design System](https://design-system.service.gov.uk/patterns/email-addresses/)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider_enrichment.rb
+++ b/app/models/provider_enrichment.rb
@@ -52,7 +52,10 @@ class ProviderEnrichment < ApplicationRecord
   validates :train_with_us, words_count: { maximum: 250, message: "^Reduce the word count for training with you" }
   validates :train_with_disability, words_count: { maximum: 250, message: "^Reduce the word count for training with disabilities and other needs" }
 
-  validates :email, :website, :telephone,
+  validates :email, email: true, on: :update, if: Proc.new { |enrichment| enrichment.email_changed? }
+  validates :email, email: true, on: :publish
+
+  validates :website, :telephone,
             :address1, :address3, :address4,
             :postcode, :train_with_us, :train_with_disability,
             presence: true, on: :publish
--- a/app/models/provider_enrichment.rb
+++ b/app/models/provider_enrichment.rb
@@ -52,8 +52,10 @@ class ProviderEnrichment < ApplicationRecord
   validates :train_with_us, words_count: { maximum: 250, message: "^Reduce the word count for training with you" }
   validates :train_with_disability, words_count: { maximum: 250, message: "^Reduce the word count for training with disabilities and other needs" }
 
-  validates :email, email: true, on: :update, if: Proc.new { |enrichment| enrichment.email_changed? }
+  validates :email, email: true, on: :update, allow_nil: true
   validates :email, email: true, on: :publish
+
+  validates :telephone, phone: { message: '^Enter a valid telephone number' }, on: :update, allow_nil: true
 
   validates :website, :telephone,
             :address1, :address3, :address4,
--- a/app/validators/email_validator.rb
+++ b/app/validators/email_validator.rb
@@ -1,0 +1,9 @@
+class EmailValidator < ActiveModel::Validator
+  EMAIL_VALIDATION_ERROR_MESSAGE = "^Enter an email address in the correct format, like name@example.com".freeze
+
+  def validate(record)
+    if record.email.nil? || record.email.empty? || !record.email.match?(/@/)
+      record.errors[:email] << EMAIL_VALIDATION_ERROR_MESSAGE
+    end
+  end
+end
--- a/app/validators/email_validator.rb
+++ b/app/validators/email_validator.rb
@@ -1,9 +1,9 @@
-class EmailValidator < ActiveModel::Validator
+class EmailValidator < ActiveModel::EachValidator
   EMAIL_VALIDATION_ERROR_MESSAGE = "^Enter an email address in the correct format, like name@example.com".freeze
 
-  def validate(record)
-    if record.email.nil? || record.email.empty? || !record.email.match?(/@/)
-      record.errors[:email] << EMAIL_VALIDATION_ERROR_MESSAGE
+  def validate_each(record, attribute, value)
+    if value.blank? || !value.match?(/@/)
+      record.errors[attribute] << EMAIL_VALIDATION_ERROR_MESSAGE
     end
   end
 end
--- a/spec/models/provider_enrichment_spec.rb
+++ b/spec/models/provider_enrichment_spec.rb
@@ -103,21 +103,48 @@ describe ProviderEnrichment, type: :model do
 
       describe 'email' do
         it 'validates email is present' do
-          enrichment.email = ""
+          enrichment.email = ''
           enrichment.valid?
 
           expect(enrichment.errors[:email]).to include("^Enter an email address in the correct format, like name@example.com")
         end
 
         it 'validates email contains an @ symbol' do
-          enrichment.email = "meow"
+          enrichment.email = 'meow'
           enrichment.valid?
 
           expect(enrichment.errors[:email]).to include("^Enter an email address in the correct format, like name@example.com")
         end
 
-        it 'does not validate if the email was not updated' do
-          enrichment.website = "cats4lyf.cat"
+        it 'Does not validate the email if it is not present'do
+          enrichment.website = 'cats4lyf.cat'
+
+          expect(enrichment.valid?).to be true
+        end
+      end
+
+      describe 'telephone' do
+        it 'validates telephone is present' do
+          enrichment.telephone = ""
+          enrichment.valid?
+
+          expect(enrichment.errors[:telephone]).to include("^Enter a valid telephone number")
+        end
+
+        it 'Correctly validates valid phone numbers' do
+          enrichment.telephone = "+447 123 123 123"
+          expect(enrichment.valid?).to be true
+        end
+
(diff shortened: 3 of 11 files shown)


spec/requests/api/v2/providers/update_spec.rb
```diff
@@ -202,7 +202,7 @@ def perform_request(provider)
 
       expect(json_response).to have_id(provider.reload.enrichments.first.id.to_s)
       expect(json_response).to have_type('provider_enrichment')
-      expect(json_response).to have_attribute(:email).with_value('email_address')
```
tvararu:
```
** server can't find cats4lyf.cat: NXDOMAIN
```

😿

app/validators/email_validator.rb
```diff
@@ -0,0 +1,9 @@
+class EmailValidator < ActiveModel::Validator
+  EMAIL_VALIDATION_ERROR_MESSAGE = "^Enter an email address in the correct format, like name@example.com".freeze
+
+  def validate(record)
+    if record.email.nil? || record.email.empty? || !record.email.match?(/@/)
```
defong:
```suggestion
    if record.email.blank? || !record.email.match?(/@/)
```

app/validators/email_validator.rb
```diff
@@ -0,0 +1,9 @@
+class EmailValidator < ActiveModel::Validator
+  EMAIL_VALIDATION_ERROR_MESSAGE = "^Enter an email address in the correct format, like name@example.com".freeze
+
+  def validate(record)
```
defong:
```suggestion
  def validate(record, attribute, value)
```

app/validators/email_validator.rb
```diff
@@ -0,0 +1,9 @@
...
+    if record.email.nil? || record.email.empty? || !record.email.match?(/@/)
+      record.errors[:email] << EMAIL_VALIDATION_ERROR_MESSAGE
```
defong:
```suggestion
      record.errors[attribute] << EMAIL_VALIDATION_ERROR_MESSAGE
```

`:email` hard coding that is bad idea

spec/validators/email_validator_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  let(:model) { ExampleModel.new }
+
+  describe 'With nil email address' do
+    before do
+      model.email = nil
+      model.validate(:no_context)
+    end
+
+    it 'Returns invalid' do
+      expect(model.valid?(:no_context)).to be false
```
defong: the default is `:no_context` so why explicit?

danielburnley: The default is `nil`, which causes `new_record?` (built-in for `ActiveRecord`) which isn't a method on the `ExampleModel`, but you can give it any context and it'll skip that so i explicitly named `:no_context`

spec/models/provider_enrichment_spec.rb
```diff
@@ -96,10 +96,81 @@
...
+          expect(enrichment.valid?).to be true
+        end
+      end
+
+      describe 'telephone' do
+        let(:valid_telephone_numbers) do
+          [
+            "+447123 123 123",
+            "+447123123123",
+            "07123123123"
```
defong: extension numbers

brackets...

danielburnley: Added in tests for extension numbers and brackets :tada:

defong:
```sql
SELECT DISTINCT( masked )
FROM   (SELECT Regexp_replace(t, '[0-9]', 'n', 'g') AS masked,
               t
        FROM   (SELECT DISTINCT( json_data ->> 'Telephone' ) AS t
                FROM   provider_enrichment) AS b
        WHERE  t IS NOT NULL
               AND t ~ '[^0-9 ]') AS m
ORDER  BY masked  
```
can we have a bit more test please, there is about 16 patterns

```
"masked"
"(****) *** ****"
"(*****) ******"
"(+**) (*)**** ******"
"+** (*) *** **** ***"
"+**(*)*** ** *****"
"*** **** **** ext ***"
"**** *** ****  ext ***"
"**** *** **** ext ****"
"***** ****** ext ***"
"***** ****** ext. ***" # this one has a dot
"***** ****** ext***"
"*********** ext ***"
"*** **** **** x***"
"**** *** **** x***"
"***** ****** x***"
"*********** x***"

```

spec/requests/api/v2/providers/update_spec.rb
misaka: Why is this not a thing‽‽‽

app/models/provider_enrichment.rb
```diff
@@ -52,7 +52,12 @@ class ProviderEnrichment < ApplicationRecord
   validates :train_with_us, words_count: { maximum: 250, message: "^Reduce the word count for training with you" }
   validates :train_with_disability, words_count: { maximum: 250, message: "^Reduce the word count for training with disabilities and other needs" }
 
-  validates :email, :website, :telephone,
+  validates :email, email: true, on: :update, if: :enrichment_details_updated
```
misaka: Just picking this apart a little, we only validate here if the `email` or `telephone` has been changed, whereas the normal Rails behaviour, iirc, is to always run the validations. Would it be make more sense to check if `email` is present before running this validation, essentially saying that any value here has to be a valid one?

Essentially this comes down to: if we update the email by side-stepping validations (e.g. using `update_attribute`) should we ensure we catch an invalid value here next time we update/validate?

spec/validators/email_validator_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+private
+
+  class ExampleModel
+    include ActiveRecord::Validations
+
+    attr_accessor :email
+
+    validates :email, email: true
+  end
```
misaka: :100: We need more of this kind of thing in our tests!

defong: we do have one more... 

https://github.com/DFE-Digital/manage-courses-backend/blob/e9fba8074abc4527eb89186b7f089e45674abd3d/spec/validators/words_count_validator_spec.rb

app/models/provider_enrichment.rb
danielburnley: The main concern i had was updating any details on the "About your organisation" page would fail whilst you had an invalid email, though checking if they're present might be good enough. The number of people with invalid emails in the enrichment are probably pretty minimal anyway

spec/models/provider_enrichment_spec.rb
danielburnley: :+1: will do

danielburnley:
```
            # "(1234) 123 1234",
            # "1234 123 1234 x123",
            # "1234 123 1234 ext 1234",
            # "1234 123 1234  ext 123",
            # "+44(0)123 12 12345",
```

Looks like these are the formats that are flagged as invalid with this current implementation - thoughts?

app/models/provider_enrichment.rb
danielburnley: Updated to `if: email_present?`/`if: telephone_present?`

spec/models/provider_enrichment_spec.rb
fofr: Does anything in our current data match these? Will they get flagged as invalid?
I prefer to be as permissive as we can

danielburnley: Yeah @tvararu and I have been thinking the same - will look at just doing some basic checking for things like letters (outside of `ext`/`x`/brackets)

app/models/provider_enrichment.rb
```diff
@@ -77,4 +82,12 @@ def set_provider_code
     # Note: provider_code is only here to support c# counterpart, until provide_code is removed from database
     self.provider_code = provider.provider_code if provider_code.blank?
   end
+
+  def email_present?
+    !email.nil?
```
defong:
```suggestion
    email.present?
```

app/models/provider_enrichment.rb
```diff
@@ -77,4 +82,12 @@ def set_provider_code
...
+  end
+
+  def telephone_present?
+    !telephone.nil?
```
defong:
```suggestion
    telephone.present?
```

app/models/provider_enrichment.rb
misaka: We can always check with a dump of production.

app/models/provider_enrichment.rb
misaka: Also: `allow_nil: true` on the validation.

app/models/provider_enrichment.rb
```diff
@@ -52,7 +52,12 @@ class ProviderEnrichment < ApplicationRecord
   validates :train_with_us, words_count: { maximum: 250, message: "^Reduce the word count for training with you" }
   validates :train_with_disability, words_count: { maximum: 250, message: "^Reduce the word count for training with disabilities and other needs" }
 
-  validates :email, :website, :telephone,
+  validates :email, email: true, on: :update, if: :email_present?
```
misaka:
```suggestion
  validates :email, email: true, on: :update, allow_nil: true
```

PR #709, opened 2019-08-13
[1900] update study mode

davidgisbey:
### Context

Users need to be able to update their course to full time or part time

- Allow study mode to be updated
- Update site statuses to the correct vac_status.

### Guidance to review

If updated to full time all site statuses except those with no vacancies should have full time vacancies.

Same with part time except part time vacancies.

There is currently no logic to deal with full time and part time courses. This will be dealt with through support until users have the functionality to select both full time and part time vacancies on each site attached to a course.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -128,7 +128,8 @@ module API
                   :qualification,
                   :age_range_in_years,
                   :start_date,
-                  :applications_open_from)
+                  :applications_open_from,
+                  :study_mode,)
           .permit(
             :about_course,
             :course_length,
@@ -173,6 +174,7 @@ module API
             :age_range_in_years,
             :start_date,
             :applications_open_from,
+            :study_mode,
           )
       end
 
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -3,6 +3,8 @@
 module API
   module V2
     class CoursesController < API::V2::ApplicationController
+      include StudyModeVacancyMapper
+
       before_action :build_recruitment_cycle
       before_action :build_provider
       before_action :build_course, except: :index
@@ -49,9 +51,11 @@ module API
       end
 
       def update
+        study_mode_updated = params.fetch(:course, {})[:study_mode] != @course.study_mode
         update_course
         update_enrichment
         update_sites
+        update_site_statuses if @course.errors.empty? && study_mode_updated
         should_sync = site_ids.present? && @course.recruitment_cycle.current?
         has_synced? if should_sync
 
@@ -95,6 +99,14 @@ module API
         # but we can't actually revert easily from what I can tell because of the
         # remove_site! side effects that occur when it's called.
         @course.errors[:sites] << "^You must choose at least one location" if site_ids.empty?
+      end
+
+      def update_site_statuses
+        @course.site_statuses.each do |site_status|
+          if site_status.vac_status != 'no_vacancies'
+            update_vac_status(@course.study_mode, site_status)
+          end
+        end
       end
 
       def build_provider
@@ -180,6 +192,10 @@ module API
 
       def site_ids
         params.fetch(:course, {})[:sites_ids]
+      end
+
+      def study_mode_params
+        params.fetch(:course, {})[:study_mode]
       end
 
       def has_synced?
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -51,14 +51,12 @@ module API
       end
 
       def update
-        study_mode_updated = params.fetch(:course, {})[:study_mode] != @course.study_mode
         update_course
         update_enrichment
         update_sites
-        update_site_statuses if @course.errors.empty? && study_mode_updated
+        update_site_status_vac_statuses if @course.study_mode_previously_changed?
         should_sync = site_ids.present? && @course.recruitment_cycle.current?
         has_synced? if should_sync
-
 
         if @course.errors.empty? && @course.valid?
           render jsonapi: @course.reload
@@ -101,7 +99,7 @@ module API
         @course.errors[:sites] << "^You must choose at least one location" if site_ids.empty?
       end
 
-      def update_site_statuses
+      def update_site_status_vac_statuses
         @course.site_statuses.each do |site_status|
           if site_status.vac_status != 'no_vacancies'
             update_vac_status(@course.study_mode, site_status)
@@ -192,10 +190,6 @@ module API
 
       def site_ids
         params.fetch(:course, {})[:sites_ids]
-      end
-
-      def study_mode_params
-        params.fetch(:course, {})[:study_mode]
       end
 
       def has_synced?
--- a/spec/requests/api/v2/providers/courses/update_study_mode_spec.rb
+++ b/spec/requests/api/v2/providers/courses/update_study_mode_spec.rb
@@ -1,0 +1,76 @@
+require "rails_helper"
+
+describe 'PATCH /providers/:provider_code/courses/:course_code' do
+  let(:jsonapi_renderer) { JSONAPI::Serializable::Renderer.new }
+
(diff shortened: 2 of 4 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -49,9 +51,11 @@ def publishable
       end
 
       def update
+        study_mode_updated = params.fetch(:course, {})[:study_mode] != @course.study_mode
```
richrace: Does it make more sense to move this check to be a guard statement within https://github.com/DFE-Digital/manage-courses-backend/pull/709/files#diff-7d6b888ba704f4684c5b576a1dd07676R104 

Also, there's duplication, should use this: https://github.com/DFE-Digital/manage-courses-backend/pull/709/files#diff-7d6b888ba704f4684c5b576a1dd07676R197-R199

tvararu:
[[S] Edit 'Full, part time or both?' (MCBE)](https://trello.com/c/dhXGG1TZ/1900-s-edit-full-part-time-or-both-mcbe)

spec/requests/api/v2/providers/courses/update_study_mode_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  let(:permitted_params) do
+    %i[study_mode]
+  end
+
+  before do
+    perform_request(updated_study_mode)
+  end
+
+  context "course has is updated to full time" do
```
tvararu: This probably needs to be reworded `"course has is updated to full time"`

app/controllers/api/v2/courses_controller.rb
davidgisbey: Have used the study_mode_params method. The issue around moving this into the update_site_status method as a  guard clause is that it by that time, the course object will have already been updated. So the course.study_mode will always be the same as the params study mode.

spec/requests/api/v2/providers/courses/update_study_mode_spec.rb
davidgisbey: Good stop. Thanks 👍

richrace:
Here's a suggestion to help with the CodeClimate warning:

```ruby 
  def update
        update_course
        update_enrichment
        update_sites
        update_site_statuses
        ....
  end 

  def update_course
        return unless course_params.values.any?
        return unless @course.course_params_assignable(course_params)

        @course.assign_attributes(course_params)

        @study_mode_changed = @course.study_mode_changed?

        @course.save
  end

  def update_site_statuses
        return if @course.errors.any? # cannot use #valid? as we manually add errors.
        return unless @study_mode_changed
        ....
   end 
```

Thoughts?

app/controllers/api/v2/courses_controller.rb
tvararu: Have had a chat with @davidgisbey IRL, now I understand why the `study_mode_updated` value needs to be cached before the `update_course` method. We've thought about a few alternatives to the current implementation.

First idea is something like this (pseudocode)

```diff
def update
  update_course
  update_enrichment
  update_sites
+ update_site_statuses
  should_sync = site_ids.present? && @course.recruitment_cycle.current?
  has_synced? if should_sync

        
...

def update_course
  return unless course_params.values.any?
  return unless @course.course_params_assignable(course_params)
+ course_params_without_study_mode = course_params
+ @course.assign_attributes(course_params_without_study_mode)
  @course.save
end

...

def update_site_statuses
+ guard_clause_goes_here
+ @course.study_mode = study_mode_params
  @course.site_statuses.each do |site_status|
    if site_status.vac_status != 'no_vacancies'
      update_vac_status(@course.study_mode, site_status)
    end
  end
end
```

Basically, we update the study mode separately from the other attributes as part of updating site statuses.

Second idea:

```diff
def update_site_statuses
+ guard_clause_goes_here
+ @course.study_mode = study_mode_params
  @course.site_statuses.each do |site_status|
    if site_status.vac_status != 'no_vacancies'
      update_vac_status(@course.study_mode, site_status)
    end
  end
end

...

def course_params
  params
    .fetch(:course, {})
      [snip]
      :start_date,
      :applications_open_from,
-     :study_mode,
    )
end
```

Which is removing the `study_mode` from the regular params and treating it separately.

There's also a third idea around checking if the `@course.has_changed` or something similar, must be some methods we can call in ActiveModel.

But to be honest, for my own 2 cents, I think the current implementation is actually fine now that I understand why it is the way it is.

tvararu:
@richrace yeah, that does look good 👍

davidgisbey:
Looks good to me. Making the changes then lets 🚢

app/controllers/api/v2/courses_controller.rb
```diff
@@ -97,6 +101,17 @@ def update_sites
         @course.errors[:sites] << "^You must choose at least one location" if site_ids.empty?
       end
 
+      def update_site_statuses
+        return if @course.errors.any?
+        return unless @study_mode_changed
```
misaka: Minor suggestion: can we just use `@course.study_mode_changed?` here? I think it's literal enough and it seems a bit tricky that this is set in `update_course` and relies on that running before this runs (whereas `@course` is set in a before hook and is a bit more predictable).

richrace: `@course.study_mode_changed?` resets after the `@course.save` call.

misaka: 🤦‍♂

defong: @davidgisbey I knew there was a magic thing you could of used `study_mode_changed?`

richrace: Docs: https://edgeapi.rubyonrails.org/classes/ActiveModel/Dirty.html

richrace: I guess you could do `@course.study_mode_previously_changed?` but it makes me think "when did it previously change?" 🤷‍♂

misaka: That's an interesting one @richrace . Since it's on the `@course` object, it would have to be within the same request cycle, and since this is an update request that could actually make sense.

misaka: Actually, this ties in with a bit of cognitive dissonance I'm experiencing with `update_site_statuses` in general. I feel like in `update` we should be calling:

```
      if @course.study_mode_previously_changed?
        update_site_status_vac_statuses
      end
```

... it just wasn't at all clear to me that the goal of this method is to update the `vac_status` on each `site_status`, if the course's `study_mode` has been changed. I think that should be made clearer.

app/controllers/api/v2/courses_controller.rb
misaka: I had a problem around this too. My idea/suggestion is to rename `update_site_statuses` to `update_site_status_vac_statuses`, and use an if statement in `update` to communicate it's intent more clearly:

```
      if @course.study_mode_previously_changed?
        update_site_status_vac_statuses
      end
```

https://github.com/DFE-Digital/manage-courses-backend/pull/709#discussion_r313833165

defong:
`CoursesController::update` this endpoint is starting to look like a nightmare...
its purpose is to do one of these
- `update_course`
- `update_enrichment`
- `update_sites`
- `update_site_statuses`

then maybe sync it
- `has_synced? if should_sync`

except technically it can do all of it

app/controllers/api/v2/courses_controller.rb
davidgisbey: Have updated it to Misha's suggestion.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -83,6 +88,7 @@ def update_course
         return unless @course.course_params_assignable(course_params)
 
         @course.assign_attributes(course_params)
+        @study_mode_changed = @course.study_mode_changed?
```
richrace: This is now redundant

davidgisbey: So it is. Thanks.

PR #710, opened 2019-08-13
[1880] Edit SEND on course via API

richrace:
### Context

Uses #704 as a base. Need to allow the Front End to set the `is_send` attribute on a Course.

- Add Validation to `Course#is_send` attribute to only allow explicit variables
- Add V2 Course Controller to accept `is_send` param

### Guidance to review

Do we trust the the information sent to the API will always be valid? If so, do we need to be so restrictive about settings `is_send`? 

For example, without the `#is_send=(value)` override then:
```ruby
$ course.is_send? 
=> true
$ course.is_send = 'something'
$ course.is_send?
=> false 
```

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -139,7 +139,8 @@ module API
                   :age_range_in_years,
                   :start_date,
                   :applications_open_from,
-                  :study_mode,)
+                  :study_mode,
+                  :is_send)
           .permit(
             :about_course,
             :course_length,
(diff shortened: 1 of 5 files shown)


richrace:
[[S] Edit SEND (MCFE)](https://trello.com/c/hR6fDSPr/1881-s-edit-send-mcfe)

richrace:
[[S] Edit SEND (MCBE)](https://trello.com/c/BzOJXura/1880-s-edit-send-mcbe)

PR #712, opened 2019-08-14
[1438] Make the accept transition screen idempotent

davidgisbey:
### Context

If a user tries to accept the transition screen while already rolled over, it currently throws a sentry error. This is undesirable behaviour.

- Make the accept transition screen idempotent

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -24,7 +24,9 @@ module API
       end
 
       def accept_rollover_screen
-        @user.accept_rollover_screen!
+        if @user.state == 'transitioned'
+          @user.accept_rollover_screen!
+        end
       end
 
     private
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -18,7 +18,9 @@ module API
       end
 
       def accept_transition_screen
-        @user.accept_transition_screen!
+        if @user.state == 'new'
+          @user.accept_transition_screen!
+        end
       end
 
       def accept_rollover_screen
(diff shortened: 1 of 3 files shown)


spec/requests/api/v2/users/accept_transition_screen_spec.rb
```diff
@@ -60,7 +60,7 @@ def perform_request
       end
 
       it 'returns error' do
-        expect { perform_request }.to raise_error AASM::InvalidTransition
```
benilovj: The spec name no longer matches the spec

davidgisbey:
[Make accept_transition_screen idempotent in mc-be](https://trello.com/c/q3EFc1UD/1438-make-accepttransitionscreen-idempotent-in-mc-be)

app/controllers/api/v2/users_controller.rb
```diff
@@ -18,7 +18,9 @@ def update
       end
 
       def accept_transition_screen
-        @user.accept_transition_screen!
+        if @user.state == 'new'
+          @user.accept_transition_screen!
+        end
       end
 
       def accept_rollover_screen
```
fofr: We should do the same for the rollover screen.

davidgisbey: Cool i'll add that in and edit the card.

spec/requests/api/v2/users/accept_transition_screen_spec.rb
davidgisbey: Thanks i'll change it now.

PR #713, opened 2019-08-14
[1972] Allow non-admin users to create access requests

davidgisbey:
### Context
Currently if a non admin user tries to create an access request it throws an error.

- Tweak the policy to allow any user to create an access request.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/access_request_spec.rb
+++ b/spec/requests/api/v2/access_request_spec.rb
@@ -282,13 +282,15 @@ describe 'Access Request API V2', type: :request do
       it { should have_http_status(:unauthorized) }
     end
 
-    context 'when unauthorized' do
-      let(:unauthorised_user) { create(:user) }
-      let(:payload) { { email: unauthorised_user.email } }
+    context 'authorises non-admin users' do
+      let(:non_admin_user) { create(:user) }
+      let(:payload) { { email: non_admin_user.email } }
 
-      it "should raise an error" do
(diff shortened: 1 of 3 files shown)


spec/requests/api/v2/access_request_spec.rb
```diff
@@ -282,12 +282,12 @@
...
-      let(:unauthorised_user) { create(:user) }
-      let(:payload) { { email: unauthorised_user.email } }
+    context 'allows non admin users to be authorized' do
+      let(:non_admin_user) { create(:user) }
+      let(:payload) { { email: non_admin_user.email } }
 
-      it "should raise an error" do
-        expect { do_post }.to raise_error Pundit::NotAuthorizedError
+      it "should not raise an error" do
+        expect { do_post }.not_to raise_error
```
richrace: Feel like this test is a bit redundant? Everything should *generally* not error. 

Does it make more sense for this to return a `200` response or something more explicit that the expected behaviour has worked.

davidgisbey:
[Users can't create access requests](https://trello.com/c/cFGAXea6/1972-users-cant-create-access-requests)

spec/requests/api/v2/access_request_spec.rb
davidgisbey: Have done 👍

PR #716, opened 2019-08-15
[1856] Set application start and end date to not null

davidgisbey:
### Context
The update applications open logic is reliant on these columns being populated.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/recruitment_cycle.rb
+++ b/app/models/recruitment_cycle.rb
@@ -4,8 +4,8 @@
 #
 #  id                     :bigint           not null, primary key
 #  year                   :string
-#  application_start_date :date
-#  application_end_date   :date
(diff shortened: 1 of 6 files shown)


davidgisbey:
[[S] Edit 'Course Applications Open' - (MCBE)](https://trello.com/c/p2P5U9L2/1856-s-edit-course-applications-open-mcbe)

defong:
you have done the migration to back fill the data right?

davidgisbey:
Yes. That's already been added in the previous migration.

timabell:
> The update applications open logic is reliant on these columns being populated.

imho this should be in the commit message too so you can find it with `git blame`

PR #717, opened 2019-08-15
1982 bad phone

defong:
### Context
Telephone number was not triggering validation on creation of an enrichment from the `enrichment` (`enrichment.save`)
Leading to the enrichment being saved then triggering validation from the provider (`@provider.valid?`) 

On the controller

Making phone calls again 

### Guidance to review
First commit to flush it out
Second commit to clean it up

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider_enrichment.rb
+++ b/app/models/provider_enrichment.rb
@@ -57,7 +57,7 @@ class ProviderEnrichment < ApplicationRecord
   validates :email, email: true, on: :update, allow_nil: true
   validates :email, email: true, on: :publish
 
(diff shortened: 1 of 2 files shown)


defong:
[When first creating a provider enrichment, initially specified attributes get attached to it even if they don't pass validation](https://trello.com/c/Ye6kXdC0/1982-when-first-creating-a-provider-enrichment-initially-specified-attributes-get-attached-to-it-even-if-they-dont-pass-validation)

PR #718, opened 2019-08-15
[1914] Ensure that JSON Data stores/retrieves RegionCode as Integer

richrace:
### Context

`jsonb_accessor` and `ActiveRecord#enums` does not play nice. Rails can use Int/String and be fine. However, we want to store the data as Integers so the C# App can read them.

`jsonb_accessor` stores the enum as a string.

This comment is explaining the order of events. https://trello.com/c/YyfUArP7/1914-investigation-regioncode-being-saved-to-provider-enrichments-as-a-string-causing-manage-api-c-to-fail#comment-5d5669a350c3475785ca1be6

- Include a before save that will always set the `RegionCode` to be an integer

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider_enrichment.rb
+++ b/app/models/provider_enrichment.rb
@@ -17,6 +17,7 @@
 class ProviderEnrichment < ApplicationRecord
   include RegionCode
 
+  before_save :ensure_region_code_is_an_integer_in_json_data
   before_create :set_provider_code
 
   enum status: { draft: 0, published: 1 }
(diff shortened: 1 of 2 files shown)


richrace:
[Investigation: `region_code` being saved to provider enrichments as a string, causing manage-api (c#) to fail](https://trello.com/c/YyfUArP7/1914-investigation-regioncode-being-saved-to-provider-enrichments-as-a-string-causing-manage-api-c-to-fail)

timabell:
Great work @richrace 
If it works, :ship: it! What we have currently is breaking the c#, we can improve on this later.

PR #719, opened 2019-08-15
[1899] Add study mode to course edit options

dankmitchell:
### Context
Study mode

Add study mode to course edit options meta

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/concerns/courses/edit_options/study_mode_options.rb
+++ b/app/models/concerns/courses/edit_options/study_mode_options.rb
@@ -1,0 +1,12 @@
+module Courses
+  module EditOptions
+    module StudyModeOptions
+      extend ActiveSupport::Concern
+      included do
+        def study_mode_options
+          Course.study_modes.keys
+        end
+      end
+    end
+  end
+end
--- a/app/models/concerns/courses/edit_options.rb
+++ b/app/models/concerns/courses/edit_options.rb
@@ -1,11 +1,11 @@
 module Courses
   module EditOptions
     extend ActiveSupport::Concern
-    include AgeRangeOptions
-    include EntryRequirements
-    include QualificationOptions
-    include StartDateOptions
-    include StudyModeOptions
+    include AgeRangeConcern
+    include EntryRequirementsConcern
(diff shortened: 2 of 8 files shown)


dankmitchell:
[Edit 'Full, part time or both?' (MCFE)](https://trello.com/c/4PqXUgSo/1899-edit-full-part-time-or-both-mcfe)

app/models/concerns/courses/edit_options/study_mode_options.rb
```diff
@@ -0,0 +1,12 @@
+module Courses
+  module EditOptions
+    module StudyModeOptions
```
misaka: What do people think about the naming? My feeling is that this should be `StudyModeOptionsConcern` ... that's the standard for Rails classes/modules that aren't models, and even though this is a concern for a model I think that this qualifies.

I know we haven't done this for the existing concerns, or the PageObjects for that matter, I'd be happy to see a commit that's just designed to rename them all instead of doing them piecemeal.

misaka: > My feeling is that this should be `StudyModeOptionsConcern`

Actually, the full path should be something like `Courses::EditOptions::StudyModeConcern` ... no need to repeat `Options` imo.

richrace: 👍 I'd like to see `Concern` suffix being used. 

But I also think the 3 levels of nesting with two `Options` could be improved upon too. 

As how would you do it?  E.g. `module CoursesConcerns::StudyModeOptionsConcern`

I don't like specific concern just for `Edit`

tvararu: I don't think much of it, I think current is fine and the suggested one with the `Concern` suffix is better but I don't have strong feelings for changing this now :)

dankmitchell: Udpated them all 👍

PR #722, opened 2019-08-16
1865 mcb god users

timabell:
### Context

Interim measure - give DfE people access to all orgs for support / research etc.

* Add `--admin` flag - `bin/mcb users grant --admin someone@digital.education.gov.uk`
* Move provider into `-p` flag (because of the way the cli lib works) - `bin/mcb users grant  someone@example.org -P AA1`

### Guidance to review

:ship: ?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/grant_access_wizard.rb
+++ b/lib/mcb/grant_access_wizard.rb
@@ -22,37 +22,37 @@ module MCB
     def setup_user
       return false unless find_or_init_user
 
-      puts MCB::Render::ActiveRecord.user @user
+      puts MCB::Render::ActiveRecord.user @user_to_grant
 
       persist_user_if_new
     end
 
     def find_or_init_user
-      @user = MCB.find_user_by_identifier @id_or_email_or_sign_in_id
-      return @user if @user != nil
+      @user_to_grant = MCB.find_user_by_identifier @id_or_email_or_sign_in_id
+      return @user_to_grant if @user_to_grant != nil
 
       unless @id_or_email_or_sign_in_id.include? '@'
         puts "#{@id_or_email_or_sign_in_id} not found. Specify an email address if you wish to create a user"
         return nil
       end
 
-      @user = User.new(email: @id_or_email_or_sign_in_id)
+      @user_to_grant = User.new(email: @id_or_email_or_sign_in_id)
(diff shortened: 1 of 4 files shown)


timabell:
[[M] add/revoke god users with mcb](https://trello.com/c/oZJmk5fT/1865-m-add-revoke-god-users-with-mcb)

benilovj:
One way you could avoid having an `--admin` flag is to auto-detect `[digital.]education.gov.uk` emails and have an extra step asking the support agent whether they want to deify the new user or not.

misaka:
> One way you could avoid having an `--admin` flag is to auto-detect `[digital.]education.gov.uk` emails and have an extra step asking the support agent whether they want to deify the new user or not.

Going forward, we're likely to change the implementation that an "admin" is someone who has access to all the providers, so it may be un-intuitive that the `provider grant` command is reused for admin access, and really what we want is a command `user admin` or something like that.

misaka:
> what we want is a command `user admin` or something like that

Thinking more about it, it would be nice to eventually have the commands:

```
mcb user admin [list]
mcb user admin add
mcb user admin remove
```

timabell:
This didn't audit at all :-(
Even though the models have `audited`

timabell:
Going to start a new PR for audit as from talking to @richrace it looks like it'll involve a model change.

lib/mcb/grant_access_wizard.rb
```diff
@@ -57,14 +62,42 @@ def persist_user_if_new
...
+    def add_user_to_org
+      puts "\n"
+      puts "You're about to give #{@user} access to organisation #{@organisation.name}. They will manage:"
+      puts MCB::Render::ActiveRecord.providers_table @organisation.providers, name: "Additional Providers"
+      @organisation.users << @user if @cli.agree("Agree?  ")
+    end
+
+    def confirm_and_add_user_to_all_organisations
+      unless @user.admin?
+        puts "Refusing to give non-admin user #{@user} access to all orgs"
```
richrace: `@user` could be renamed to something more real?

Suggestions: `@user_to_update`, `@target_user` or `@user_to_grant`

PR #723, opened 2019-08-18
[1997] add endoint for new course

misaka:

From the ticket:

> ## What
>
> Add "new" endpoint for courses to mcbe that returns default `edit_options` to be
> used when generating the "new" pages on the frontend.

### Context

From the ticket:

> ## Why
> 
> We are keeping the logic behind course states constrained to the backend, so we
> need to be able to query it to find out what outcomes are available, for courses
> as we step the user through the wizard steps.

### Guidance to review

This is required by [1938 Add "new" page for course outcomes](https://trello.com/c/4HGV6r3S/1938-m-add-new-page-for-course-outcomes).

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/concerns/with_qualifications.rb
+++ b/app/models/concerns/with_qualifications.rb
@@ -1,7 +1,7 @@
 module WithQualifications
   extend ActiveSupport::Concern
 
-  included do
+  included do # rubocop: disable Metrics/BlockLength
     # The training programme outcome that originated in the UCAS NetUpdate system.
     #
     # See [UCAS Teacher Training Set-up Guide](https://www.ucas.com/file/115581/download?token=mv-G6P53),
@@ -16,19 +16,19 @@ module WithQualifications
       # Professional: the student will receive a Professional Graduate
       # Certificate of Education (offered at Level 6) or Professional
       # Graduate Diploma in Education (PGDE), with no credits or
-      # modules at postgraduate (master’s) Level 7 on successful
+      # modules at postgraduate (master's) Level 7 on successful
       # completion of the programme.
       professional: "PF",
 
       # Postgraduate: the student will receive a Postgraduate Certificate
       # of Education (PGCE) or other qualification which includes at
       # least one module or some credits at postgraduate
-      # (master’s) Level 7 on successful completion the
+      # (master's) Level 7 on successful completion the
       # programme.
       postgraduate: "PG",
 
       # Both professional and postgraduate: the student has the option of taking at least one
-      # postgraduate (master’s) Level 7 module or obtaining some
+      # postgraduate (master's) Level 7 module or obtaining some
       # postgraduate-level credits as part of the programme.
       professional_postgraduate: "BO",
     }
@@ -61,7 +61,11 @@ module WithQualifications
     end
 
     def qualifications_description
-      qualifications.map(&:upcase).sort.join(" with ")
+      if qualifications
+        qualifications.map(&:upcase).sort.join(" with ")
+      else
+        ''
+      end
     end
 
     def qualification=(value)
--- a/app/services/subject_mapper_service.rb
+++ b/app/services/subject_mapper_service.rb
@@ -92,6 +92,8 @@ class SubjectMapperService
   }.freeze
 
   def self.get_subject_list(course_title, ucas_subjects)
+    return [] if course_title.nil?
+
     ucas_subjects = ucas_subjects.map(&:strip).map(&:downcase)
     level = Subjects::CourseLevel.new(ucas_subjects).level
 
--- a/spec/models/concerns/with_qualifications_spec.rb
+++ b/spec/models/concerns/with_qualifications_spec.rb
@@ -22,6 +22,14 @@ RSpec.describe WithQualifications, type: :model do
     end
   end
 
+  describe '#qualifications_description' do
+    context 'no qualification present' do
+      subject { build(:course, qualification: nil) }
+
+      its(:qualifications_description) { should eq '' }
+    end
+  end
+
   describe "#qualification= and its dependent attribute profpost_flag" do
     subject { build(:course, qualification: :pgce_with_qts) }
 
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -7,11 +7,20 @@ module API
 
       before_action :build_recruitment_cycle
       before_action :build_provider
-      before_action :build_course, except: :index
+      before_action :build_course, except: %i[index new]
 
       deserializable_resource :course,
                               only: %i[update publish publishable],
                               class: API::V2::DeserializableCourse
+
+      def new
+        authorize Course
+
+        @course = @provider.courses.new
+        @course.valid?
+
+        render jsonapi: @course, include: params[:include]
+      end
 
       def index
         authorize @provider, :can_list_courses?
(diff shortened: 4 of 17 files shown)


misaka:
[Add "new" endpoint for courses to mcbe to support creating couses in mcfe](https://trello.com/c/JvPQsXWj/1997-add-new-endpoint-for-courses-to-mcbe-to-support-creating-couses-in-mcfe)

app/models/concerns/with_qualifications.rb
```diff
@@ -16,19 +16,19 @@ module WithQualifications
       # Professional: the student will receive a Professional Graduate
       # Certificate of Education (offered at Level 6) or Professional
       # Graduate Diploma in Education (PGDE), with no credits or
-      # modules at postgraduate (master’s) Level 7 on successful
+      # modules at postgraduate (master's) Level 7 on successful
```
misaka: btw, I'm not being petty with changing the apostrophe to single-quotes, I did it just because this was forcing my editor to stick a utf-8 coding magic comment at the top of the file and I thought may as well keep these clean ASCII for now. This was probably the wrong decision. ¯\_(ツ)_/¯

misaka:
This turned into more work than anticipated. I'm willing to put some time in the evenings to address any comments/questions on this PR, but pls keep in mind I'm trying to be on holiday so I'd greatly appreciate pro-active comments and, as always, encourage additional commits to help get this over the line. :)

app/services/subject_mapper_service.rb
```diff
@@ -92,6 +92,8 @@ class SubjectMapperService
   }.freeze
 
   def self.get_subject_list(course_title, ucas_subjects)
+    return [] if course_title.nil?
```
defong:
```suggestion
    return [] if course_title.blank?
```

spec/models/concerns/with_qualifications_spec.rb
```diff
@@ -22,6 +22,14 @@
     end
   end
 
+  describe '#qualifications_description' do
+    context 'no qualification present' do
+      subject { build(:course, qualification: nil) }
+
+      its(:qualifications_description) { should eq '' }
```
defong: just wondering is should there be a default qualifications?

defong: @fofr

app/services/subject_mapper_service.rb
misaka: I think I was assuming that we only want to guard against uninitialised values, i.e. for a new course, and behave as before for a title that has been set to empty string. Also, it seemed less intrusive to only change the behaviour for nil. No?

spec/models/concerns/with_qualifications_spec.rb
fofr: There's no default. A course must always have a qualification.

app/services/subject_mapper_service.rb
defong: hmmm... coming to think of it a course should have a title regardless either from 
-  user input (for adding new courses) or database (for existing courses)
- or calculated from the subjects list (for adding new courses).

I'm not to sure how this `self.get_subject_list` needs to know that if `title is not set` then `return no subjects`.

maybe scope creep...

original reason for the subject mapper needing the title is because a `English` course may not have a subject associated with it.

spec/models/concerns/with_qualifications_spec.rb
defong: should we enforce that @misaka ?

app/services/subject_mapper_service.rb
misaka: This guard was added purely to allow the API to return a "new" version of the course. We to be able to:

1. not set any attributes, but validate those that are set so that we can start to return meaningful feedback to the user in the "create course" wizard
2. return some of the metadata that's required by the "create course" wizard to populate the steps required to create a course

So we need a way to allow a course that has no title set (i.e. `nil`) to still have this method called on it without blowing up. That could happen here, as in this PR, or in `Course#dfe_subjects` if that's better, I'm not too fussed tbh.  Would that be a better place for it?

spec/models/concerns/with_qualifications_spec.rb
misaka: Similar to `SubjectMapperService.get_subject_list`, the change to `WithQualifications`'s `qualifications_description` was done just to facilitate returning a new course for the "create course" wizard. It's not valid to have a course have no qualifications, but that should be caught by a validation since until we persist to the db, we want to allow invalid course objects to be instantiated so that we can help out the wizard.

app/services/subject_mapper_service.rb
misaka: Just off the back of the comment below for `with_qualification_spec.rb`, I want to re-iterate the same here: It's not valid to have a course have no title, but that should be caught by a validation since until we persist to the db, we want to allow invalid course objects to be instantiated so that we can help out the wizard.

davidgisbey:
image

I think something will needed to be added to the new method which assigns a course_code.

danielburnley:
If we're creating a course code - if the user abandons creation how do we determine whether or not to "free up" that course code again? 

If we've got to have something like that I wonder if allowing "draft" courses is going to have to be the better approach than the current "all in one" thinking - otherwise we'll end up with this slightly odd logic around reserving/releasing course codes :thinking:

tvararu:
@danielburnley after chatting with you and @fofr we've come to the conclusion that the course code is only really *needed* for the course preview, in which it could be substituted for `CODE` or such.  So there should be no need to generate a `CODE` when hitting this.✌️

fofr:
Do you know how we are planning on generating them?

Suggestion: [A-Z][0-9][0-9][0-9]
eg X100 or B456
Avoids rude words. Makes codes easier to remember. Matches pattern of large proportion of existing codes.

davidgisbey:
I think generating a course code will have to be implemented differently to site codes due to the sheer magnitude of possibilities. Site code creates x amount of options and removes one from the possible site codes is used. That's not really feasible for a 4 digit code.

davidgisbey:
Generates one using the above then checks the provider doesn't already have it as a course_code?

tvararu:
@fofr something like this?

```ruby
undesirable_letters = ["O", "I"]
all_letters = ('A'..'Z').to_a
first_letter = (all_letters - undesirable_letters).sample
all_numbers = (0..9).to_a
code = first_letter + all_numbers.sample(3)
```

misaka:
I whole-heartedly endorse not creating a course code at this stage because of the potential issues around reservation, etc. It seems like the course code is something we give to the user as part of the confirmation that the course has been created.

misaka:
Also, is anything still blocking this PR being accepted? @defong you’ve been kind enough to comment on this, are you happy with my responses?

defong:
ummm.... I did approve it....

I guessing this PR is getting the flaks as

- its the first PR for adding a new course
   - ideas and concepts are still in a state of flux
   - what exactly is the state of the course
      - this PR just returns a default course with out persisting it to the database
      - the course current state is just transient  
- we side step a number of things, its catching up on us, such as
   - course code generation/assignments vs reservation
   - subjects/level assignments vs readdress
   - preview vs confirmation
   - overall idea of how to delivery the whole thing 
- the wizardry is still hocus pocus
- they wondering why I was being kind..

I'm happy for this PR to go in to test the water and forcing the above out of the shadows

anyway brief thoughts

| process  	|  notes 	|
|---	|---	|
| start  	|   	where this PR is, nothing is saved|
|no man land   	|ping pong with backend to validate and grab the validation messages, frontend will control/guide the user to fill in items, nothing is saved, just building payload|
|save it| fullfiled, previewable, create is called on the backend or something similar|
|confirm| done|

PR #726, opened 2019-08-19
Example Service Object: Refactor courses copy to new provider

richrace:
### Context

Courses model is huge and should be refactored.

There is no ticket for this as this is mostly for discussion point for possible way to address Fat Models/Fat Controllers.

Make a service object that contains single use case for Copying a course to a new Provider.

Further work should be done to simplify the tests.

### Guidance to review

- Made explicit call to save the Course
- Added transaction so if any part fails; to not leave a course in an odd state
- Made explicit return as the result is used to count.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history


defong:
Looks nice...

Would consider tackling from `copy_to_recruitment_cycle` that point as the is the main entry.

Then the service will be more of a `copy service` via the recruitment cycle as that's where it stems from.

richrace:
I see it as being something a long the lines of:

`services/providers/copy_to_recruitment_cycle_service` thats uses:

`services/courses/copy_to_provider_service`
`services/sites/copy_to_provider_service`

And try to keep each Service small.

PR #730, opened 2019-08-20
[2003] Add migration to copy provider enrichments

danielburnley:
### Context

In order to make providers cycle aware (https://trello.com/c/wXf4ncic/1854-s-make-provider-cycle-aware) we would like to ensure that when they go to the "About your org" pages for the next cycle they are met with data, rather than it being blank. 

- Adds migration which goes over current providers and copies the last enrichment to the current cycle
    - Copying only the last mirrors how courses were rolled over

### Guidance to review

- Check out the branch to your local machine
- Ensure you have production-like data (in terms of quantity)
- Run the migrations

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider_enrichment.rb
+++ b/app/models/provider_enrichment.rb
@@ -20,7 +20,7 @@ class ProviderEnrichment < ApplicationRecord
   before_save :ensure_region_code_is_an_integer_in_json_data
   before_create :set_provider_code
 
-  enum status: { draft: 0, published: 1 }
+  enum status: { draft: 0, published: 1, rolled_over: 2 }
 
   belongs_to :provider,
              inverse_of: 'enrichments'
@@ -34,7 +34,7 @@ class ProviderEnrichment < ApplicationRecord
 
   scope :latest_created_at, -> { order(created_at: :desc) }
   scope :latest_published_at, -> { order(last_published_at: :desc) }
-  scope :draft, -> { where(status: 'draft') }
+  scope :draft, -> { where(status: %w[draft rolled_over]) }
 
   jsonb_accessor :json_data,
                  email: [:string, store_key: 'Email'],
@@ -64,6 +64,10 @@ class ProviderEnrichment < ApplicationRecord
             :address1, :address3, :address4,
             :postcode, :train_with_us, :train_with_disability,
             presence: true, on: :publish
+
+  def draft?
+    status.in? %w[draft rolled_over]
+  end
 
   def has_been_published_before?
     last_published_at.present?
--- a/app/services/provider_enrichments/rollover_enrichment_to_provider_service.rb
+++ b/app/services/provider_enrichments/rollover_enrichment_to_provider_service.rb
@@ -1,0 +1,10 @@
+module ProviderEnrichments
+  class RolloverEnrichmentToProviderService
+    def execute(enrichment:, new_provider:)
+      new_enrichment = enrichment.dup
+      new_enrichment.last_published_at = nil
+      new_enrichment.provider = new_provider
+      new_enrichment.rolled_over!
+    end
+  end
+end
--- a/db/migrate/20190820095253_copy_enrichments_to_next_cycle.rb
+++ b/db/migrate/20190820095253_copy_enrichments_to_next_cycle.rb
@@ -1,0 +1,25 @@
+class CopyEnrichmentsToNextCycle < ActiveRecord::Migration[5.2]
+  def up
+    next_providers = RecruitmentCycle.next_recruitment_cycle.providers.all
+    rollover_service = ProviderEnrichments::RolloverEnrichmentToProviderService.new
+
+    current_providers_with_enrichments.each do |provider|
+      next if provider.enrichments.empty?
+
+      last_enrichment = provider.enrichments.order(created_at: :desc, id: :desc).first.dup
+
+      next_cycle_provider = next_providers.find { |n| n.provider_code == provider.provider_code }
+      rollover_service.execute(enrichment: last_enrichment, new_provider: next_cycle_provider)
+    end
+  end
+
+  def down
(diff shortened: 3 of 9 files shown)


db/migrate/20190820095253_copy_enrichments_to_next_cycle.rb
```diff
@@ -0,0 +1,20 @@
+class CopyEnrichmentsToNextCycle < ActiveRecord::Migration[5.2]
+  def up
+    current_providers = RecruitmentCycle.current_recruitment_cycle.providers.includes(:enrichments).all
+    next_providers = RecruitmentCycle.next_recruitment_cycle.providers.all
+
+    current_providers.each do |p|
+      next if p.enrichments.empty?
```
defong: `current_providers` should we do the filtering beforehand?
ie `current_providers_with_enrichments`?

db/migrate/20190820095253_copy_enrichments_to_next_cycle.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      last_enrichment = p.enrichments.order(created_at: :desc, id: :desc).first.dup
```
defong: `last_enrichment` should this only be the ones that have `status` as `published`? @fofr

db/migrate/20190820095253_copy_enrichments_to_next_cycle.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      next_cycle_provider = next_providers.find { |n| n.provider_code == p.provider_code }
+      next_cycle_provider.enrichments << [last_enrichment]
```
defong: before assigning

`last_published_at` needs to be nil out
`status` must not be `published`

db/migrate/20190820095253_copy_enrichments_to_next_cycle.rb
danielburnley: I took the same logic used within copying courses to providers, which takes the latest enrichment regardless of status:

```ruby
      if (last_enrichment = enrichments.latest_first.first)
        last_enrichment.copy_to_course(new_course)
      end
```

danielburnley:
@defong Providers also have a `content_status` but it's based on the status of the last enrichment:

```ruby
  def content_status
    newest_enrichment = enrichments.latest_created_at.first

    if newest_enrichment.nil?
      :empty
    elsif newest_enrichment.published?
      :published
    elsif newest_enrichment.has_been_published_before?
      :published_with_unpublished_changes
    else
      :draft
    end
  end
```

app/services/provider_enrichments/rollover_enrichment_to_provider_service.rb
```diff
@@ -0,0 +1,11 @@
+module ProviderEnrichments
+  class RolloverEnrichmentToProviderService
+    def execute(enrichment:, new_provider:)
+      new_enrichment = enrichment.dup
+      new_enrichment.last_published_at = nil
+      new_enrichment.provider = new_provider
+      new_enrichment.save!
+      new_enrichment.rolled_over!
```
misaka: I believe `rolled_over!` calls `update!` which calls `save!`. Is it worth changing this to:

```ruby
      new_enrichment = enrichment.dup
      new_enrichment.last_published_at = nil
      new_enrichment.provider = new_provider
      new_enrichment.rolled_over!
```

Just on principle of hitting the DB less?

app/models/provider_enrichment.rb
```diff
@@ -34,7 +34,7 @@ class ProviderEnrichment < ApplicationRecord
 
   scope :latest_created_at, -> { order(created_at: :desc) }
   scope :latest_published_at, -> { order(last_published_at: :desc) }
-  scope :draft, -> { where(status: 'draft') }
+  scope :draft, -> { where(status: 'draft').or(ProviderEnrichment.rolled_over) }
```
dankmitchell: I think this is advised against in rails 6...

misaka: This looks like it should either be:

```suggestion
  scope :draft, -> { where(ProviderEnrichment.draft).or(ProviderEnrichment.rolled_over) }
```

or

```suggestion
  scope :draft, -> { where(status: [:draft, :rolled_over]) }
```

assuming either of these would actually work?

misaka: What part @dankmitchell ?

dankmitchell: `DEPRECATION WARNING: Class level methods will no longer inherit scoping from `draft` in Rails 6.1.`

spec/models/provider_enrichment_spec.rb
```diff
@@ -323,4 +323,18 @@
       end
     end
   end
+
+  describe 'Rollover' do
+    let(:enrichment) { build(:provider_enrichment) }
+
+    before { enrichment.rolled_over! }
+
+    it 'can be rolled over' do
```
defong:
```suggestion
    it 'is status is rolled over' do
```

app/models/provider_enrichment.rb
danielburnley: that latter one seems to work :eyes:

db/migrate/20190820095253_copy_enrichments_to_next_cycle.rb
```diff
@@ -0,0 +1,25 @@
+class CopyEnrichmentsToNextCycle < ActiveRecord::Migration[5.2]
+  def up
+    next_providers = RecruitmentCycle.next_recruitment_cycle.providers.all
+    rollover_service = ProviderEnrichments::RolloverEnrichmentToProviderService.new
+
+    current_providers_with_enrichments.each do |provider|
+      next if provider.enrichments.empty?
```
misaka: Not a blocker since this is a throwaway, but if you end up changing anything else here this check is now redundant because you now do it in `current_providers_with_enrichments`.

PR #733, opened 2019-08-20
[1865] add audits to mcb

richrace:
### Context

Refactor to make adding audits easier 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


richrace:
[[M] add/revoke god users with mcb](https://trello.com/c/oZJmk5fT/1865-m-add-revoke-god-users-with-mcb)

richrace:
I personally disagree with the CodeClimate checks. Chance my view.

lib/mcb/editor/base.rb
```diff
@@ -0,0 +1,31 @@
...
+        check_authorisation
+      end
+
+    protected
+
+      def setup_cli
+        raise NotImplementedError, "Needs to be implemented in Child Class"
+      end
+
+      def check_authorisation
```
defong: Only `CoursesEditor` & `ProviderEditor` actually uses this so maybe

MCB::Editor::Base
MCB::Editor::AuthorisationBase

lib/mcb/editor/base.rb
```diff
@@ -0,0 +1,31 @@
...
+        @requester = requester
+
+        setup_cli
+        check_authorisation
+      end
+
+    protected
+
+      def setup_cli
+        raise NotImplementedError, "Needs to be implemented in Child Class"
```
defong: So ruby no abstract method/class :man_facepalming:

richrace: Rails does (ActiveRecord) - https://api.rubyonrails.org/classes/ActiveRecord/Inheritance/ClassMethods.html

lib/mcb/editor/base.rb
richrace: It's a bit odd that `GrantAccessWizard` and `RevokeAccessWizard` doesn't require Authorisation.

lib/mcb/editor/courses_editor.rb
```diff
@@ -0,0 +1,215 @@
...
+
+      def course
+        if @courses.count != 1
+          raise ArgumentError, "Cannot do this operation when there are multiple courses"
+        end
+
+        @courses.first
+      end
+
+      def check_authorisation
```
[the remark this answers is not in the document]
richrace: Yea, I can move the `protected` methods.

lib/mcb/editor/base.rb
defong: @misaka @timabell this looks odd I guessing the assumption its one of us.

so may be `requester.is_admin` should be fair

I'm not fancying  `check_authorisation` just returns turns`true` as it defeats authorisation via assumption

also see https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1566471414155000

I'm also guessing `AccessRequest` flow is now neglected...

lib/mcb/editor/grant_access_wizard.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    protected
+
+      def setup_cli
+        @cli = HighLine.new
+      end
+
+      def check_authorisation
+        # There is no authorisation for Granting users.
+        true
```
misaka: Let's add an authorisation to the pundit policy.

lib/mcb/editor/base.rb
misaka: I think we should be adding authorisation methods to the Pundit policy.

lib/mcb/editor/revoke_access_wizard.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    protected
+
+      def setup_cli
+        @cli = HighLine.new
+      end
+
+      def check_authorisation
+        # There is no authorisation for Granting users.
+        true
```
misaka: Let's add an authorisation to the pundit policy.

lib/mcb/editor/base.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      def check_authorisation
+        raise NotImplementedError, "Needs to be implemented in Child Class"
+      end
+
+      def audit
+        Audited.audit_class.as_user(@requester) do
+          yield
+        end
+      end
```
misaka: So, nice pattern yall, but why did we need an explicit mechanism like this when we have an existing implicit mechanism?

https://github.com/DFE-Digital/manage-courses-backend/blob/869ce4af455299a126f78015361d15c5404eb3a7/config/initializers/mcb_audit.rb#L7

Not being judgy, I realise there could be a reason this was missed or just didn't work.

richrace: 1. Didn't realise this file (`config/initializers/mcb_audit.rb`) existed
2. Used the same code (`Audited.store[:audited_user] = user`) and in test suite didn't work when testing Audited User
3. This uses the `requester` which is set in the Config so would always Audit.
4. When is `MCB_AUDIT_USER` set?

misaka: > 2. Used the same code (Audited.store[:audited_user] = user) and in the test suite it didn't work when testing the Audited User

We have seen this work so not sure why it wouldn't.

> 3. This uses the requester which is loaded via the Config so would always Audit.

Sorry but not clear on what you're saying here. Don't we always want to audit? Doesn't the config initialiser achieve that?

> 4. When is MCB_AUDIT_USER set?

I should be set automatically by `init_rails` / `rails_runner`:

https://github.com/DFE-Digital/manage-courses-backend/blob/5a3daa0356a80254fc7fbc57dd68a35a5758c7c1/lib/mcb.rb#L35

Just to note, @faissalMT is following up on this, as should be obvious already. :)

richrace: 2. Had a method in `MCB::Editor::Base` that setup the Audited User and it worked on an individual test but not in the suite. So I had low confidence it would work in production.

3/4. Not knowing about the setting the ENV variable and initialiser just meant not clear what's going on. I didn't go hunting down the chain when responding last night.

If I remember correctly when using the MCB tool with @timabell audits were not stored with a User (when we were adding auditing to Grant/Revoke). So we didn't think any code existed that catered for this. Might be worthwhile double checking that current data has audits with the correct User recorded.

richrace: This is probably why Audits were not happening:  https://github.com/DFE-Digital/manage-courses-backend/pull/733/commits/828d25d49e56c75170a4daa92c1d7147133125f1#diff-14af88d7d0f67b03757c0f66837db947L1

PR #736, opened 2019-08-21
[2007] Validate that a course has sites before publishing

tvararu:
### Context

We shouldn't allow publishing courses without at least one site

Add validation with message, update specs

### Guidance to review

:shipit: 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/course/publishable_spec.rb
+++ b/spec/models/course/publishable_spec.rb
@@ -1,6 +1,8 @@
 RSpec.describe Course, type: :model do
   describe '#publishable?' do
     let(:course) { create(:course) }
+    let(:site) { create(:site) }
+    let(:site_status) { create(:site_status, :new, site: site) }
 
     subject { course }
 
@@ -9,7 +11,7 @@ RSpec.describe Course, type: :model do
     context 'with enrichment' do
       let(:enrichment) { build(:course_enrichment, :subsequent_draft, created_at: 1.day.ago) }
       let(:course) {
-        create(:course, enrichments: [enrichment])
+        create(:course, enrichments: [enrichment], site_statuses: [site_status])
       }
 
       its(:publishable?) { should be_truthy }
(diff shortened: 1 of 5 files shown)


tvararu:
[Fix issue with rolled over 2020 courses with 0 `site_statuses`](https://trello.com/c/WiXnwzp8/2007-fix-issue-with-rolled-over-2020-courses-with-0-sitestatuses)

spec/models/course/publishable_spec.rb
```diff
@@ -9,15 +11,33 @@
...
+        end
+
+        it { should_not be_empty }
+      end
+    end
+
+    context 'with no sites' do
+      let(:enrichment) { build(:course_enrichment, :subsequent_draft, created_at: 1.day.ago) }
+      let(:course) {
+        create(:course, enrichments: [enrichment])
```
richrace: Should we explicitly set `site_statuses` as empty here?

```suggestion
        create(:course, site_statuses: [], enrichments: [enrichment])
```

spec/requests/api/v2/providers/courses/publish_spec.rb
```diff
@@ -109,18 +109,24 @@
     describe 'failed validation' do
       let(:json_data) { JSON.parse(subject.body)['errors'] }
 
-      context 'no enrichments' do
+      context 'no enrichments or sites' do
         let(:course) { create(:course, provider: provider) }
```
richrace: And add the similar explicit no site/enrichment here too.

spec/requests/api/v2/providers/courses/publish_spec.rb
```diff
@@ -109,18 +109,24 @@
     describe 'failed validation' do
       let(:json_data) { JSON.parse(subject.body)['errors'] }
 
-      context 'no enrichments' do
+      context 'no enrichments or sites' do
```
defong: can we break this into two tests, please enrichment is painful enough.

Change the original course to fulfil the original tests
then add an explicit one for the no sites.

richrace: I read it as 'no enrichments AND no sites'

tvararu: Yeah, I don't think it's necessary to break this up. The file is already quite long...

defong: cool

PR #737, opened 2019-08-22
[1947] Ensure welcome emails are retried

danielburnley:
### Context

When logging in locally - it was found that if it was the first login it would throw a 500 but still update the email received.

- Update the README to expose the environment variables that will need to be set in order to login for the first time locally
- Don't update the welcome email sent time until after it's sent

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/session_spec.rb
+++ b/spec/requests/api/v2/session_spec.rb
@@ -106,16 +106,16 @@ describe '/api/v2/sessions', type: :request do
         end
       end
 
-      context 'on first login' do
-        let(:user) { create(:user, first_login_date_utc: nil) }
+      context 'When the user has not received a welcome email' do
+        let(:user) { create(:user, welcome_email_date_utc: nil) }
 
         it 'Sends a welcome email to the user' do
           expect(govuk_notify_request).to have_been_made
         end
       end
 
-      context 'on second login' do
-        let(:user) { create(:user, first_login_date_utc: 10.days.ago) }
+      context 'When the user has received a welcome email' do
+        let(:user) { create(:user) }
 
         it 'Does not send a welcome email to the user' do
           expect(govuk_notify_request).not_to have_been_made
--- a/spec/services/send_welcome_email_service_spec.rb
+++ b/spec/services/send_welcome_email_service_spec.rb
@@ -10,7 +10,8 @@ describe SendWelcomeEmailService do
       spy(
         first_name: 'Meowington',
         email: 'meowington@cat.net',
-        first_login_date_utc: nil
+        first_login_date_utc: nil,
+        welcome_email_date_utc: nil
       )
     end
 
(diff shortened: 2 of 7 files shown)


spec/services/send_welcome_email_service_spec.rb
```diff
@@ -51,12 +53,41 @@
...
+      end
+
+      it 'does not send the welcome email' do
+        expect(mailer_spy).not_to have_received(:send_welcome_email)
+      end
     end
 
-    it 'does not send the welcome email' do
-      expect(mailer_spy).not_to have_received(:send_welcome_email)
+    context 'And has not recieved a welcome email' do
```
defong:
```suggestion
    context 'And has not received a welcome email' do
```
i before e only when you can see

spec/services/send_welcome_email_service_spec.rb
```diff
@@ -51,12 +53,41 @@
       expect(current_user_spy).not_to have_received(:update).with(hash_including(first_login_date_utc: Time.now.utc))
     end
 
-    it 'does not update their welcome email date' do
-      expect(current_user_spy).not_to have_received(:update).with(hash_including(welcome_email_date_utc: Time.now.utc))
+    context 'And has recieved a welcome email' do
```
defong:
```suggestion
    context 'And has received a welcome email' do
```

spec/requests/api/v2/session_spec.rb
```diff
@@ -106,16 +106,16 @@
...
 
         it 'Sends a welcome email to the user' do
           expect(govuk_notify_request).to have_been_made
         end
       end
 
-      context 'on second login' do
-        let(:user) { create(:user, first_login_date_utc: 10.days.ago) }
+      context 'When the user has received a welcome email' do
+        let(:user) { create(:user, welcome_email_date_utc: 10.days.ago) }
```
defong:
```suggestion
        let(:user) { create(:user) }
```

you set it in the factory now...

spec/requests/api/v2/session_spec.rb
```diff
@@ -106,16 +106,16 @@
         end
       end
 
-      context 'on first login' do
-        let(:user) { create(:user, first_login_date_utc: nil) }
+      context 'When the user has not received a welcome email' do
+        let(:user) { create(:user, welcome_email_date_utc: nil) }
```
defong:
```suggestion
        let(:user) { create(:user, :first_timer) }
```

maybe add a trait in factory that nils out the `first_login_date_utc` and the `welcome_email_date_utc` fields

danielburnley: Maybe if it becomes more useful outside of these two instances - but I find traits can make it less clear as to what's being tested sometimes

PR #738, opened 2019-08-22
[2054] Refactor Rollover related `copy_to` out of models into Service Objects

richrace:
### Context

Move code outside of Models that don't need to know about it

Introduce service objects with Dependency Inversion

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/services/providers/copy_to_recruitment_cycle_service_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      expect(mocked_copy_site_service).to have_received(:execute).with(site: site, new_provider: new_provider)
+    end
+
+    it 'copies over the courses' do
+      service.execute(new_recruitment_cycle)
+
+      expect(mocked_copy_course_service).to have_received(:execute).with(course: course, new_provider: new_provider)
+    end
+
+    it 'returns a hash of the counts of copied objects' do
```
danielburnley: Should also test the value it returns also :eyes: you can mock that value for the sites and courses in your doubles, ensures that they're wired together as expected

app/services/courses/copy_to_provider_service.rb
```diff
@@ -1,29 +1,31 @@
...
+      @sites_copy_to_course = sites_copy_to_course
+      @enrichments_copy_to_course = enrichments_copy_to_course
     end
 
-    def execute(new_provider)
-      new_course = new_provider.courses.find_by(course_code: @course.course_code)
+    def execute(course:, new_provider:)
+      new_course = new_provider.courses.find_by(course_code: course.course_code)
 
       return nil if new_course.present?
```
danielburnley: These two lines could be refactored into a method that would read a bit nicer (and remove the need to reassign `new_course` further down)

something like:

```ruby
return nil if course_code_already_exists_on_provider?
```

app/services/courses/copy_to_provider_service.rb
```diff
@@ -1,29 +1,31 @@
...
+        new_course = course.dup
         new_course.provider = new_provider
         new_course.save!
 
-        new_course.subjects << @course.subjects
+        new_course.subjects << course.subjects
 
-        last_enrichment = @course.enrichments.latest_first.first
-        last_enrichment.copy_to_course(new_course) if last_enrichment.present?
+        last_enrichment = course.enrichments.latest_first.first
```
danielburnley: These two lines could be moved into:

```ruby
copy_latest_enrichment_to_course(course, new_course)
```

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -0,0 +1,67 @@
+module Providers
+  class CopyToRecruitmentCycleService
+    def initialize(provider:, copy_course_to_provider_service:, copy_site_to_provider_service:)
+      @provider = provider
+      @copy_course_to_provider_service = copy_course_to_provider_service
+      @copy_site_to_provider_service = copy_site_to_provider_service
+    end
+
+    def execute(new_recruitment_cycle)
```
danielburnley: I'd expect the provider to be passed in to here rather than the initializer - then it's clear what provider you're copying to the service

Additionally we may want to copy multiple providers in sequence and having to reinitialise this should be a bit of a faff

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -0,0 +1,67 @@
...
+      providers_count = 0
+      sites_count = 0
+      courses_count = 0
+
+      ActiveRecord::Base.transaction do
+        new_provider = new_recruitment_cycle
+                          .providers
+                          .find_by(provider_code: @provider.provider_code)
+
+        unless new_provider
```
danielburnley: Similar to elsewhere - this check could be made clearer. I _think_ this is ensuring we're not copying something that already exists, but the name `new_provider` has me confused.

Would be better as `unless provider_code_already_exists_on_recruitment_cycle?`

app/services/sites/copy_to_provider_service.rb
```diff
@@ -0,0 +1,16 @@
+module Sites
+  class CopyToProviderService
+    def execute(site:, new_provider:)
+      new_site = new_provider.sites.find_by(code: site.code)
+
+      return nil if new_site.present?
```
danielburnley: `return nil if site_code_already_exists_on_provider?`?

lib/mcb/commands/providers/rollover.rb
```diff
@@ -19,7 +19,15 @@
       counts = nil
       bm = Benchmark.measure do
         Provider.connection.transaction do
-          counts = provider.copy_to_recruitment_cycle(new_recruitment_cycle)
+          service = Providers::CopyToRecruitmentCycleService.new(
+            provider: provider,
+            copy_course_to_provider_service: Courses::CopyToProviderService.new(
```
danielburnley: Might be worth extracting `Courses::CopyToProviderService` out into a variable, otherwise the nesting makes this a little hard to parse :smile:

spec/services/courses/copy_to_provider_service_spec.rb
```diff
@@ -35,34 +42,28 @@
...
 
-      its(:content_status) { should eq :rolled_over }
+  context 'when a published enrichment exists' do
+    let!(:published_enrichment) do
+      create :course_enrichment, :published, course: course
     end
 
-    describe 'the copied enrichment' do
-      subject { new_course.enrichments.first }
+    it 'copies the published enrichment' do
```
danielburnley: Worth also testing that is copies the _latest_ enrichment regardless of status :eyes:

timabell:
[model refactor](https://trello.com/c/e59iHW3L/2054-model-refactor)

timabell:
rebased

PR #739, opened 2019-08-23
[1971] Add bulk publish sync from manage courses to search and c…

defong:
### Context
Bulk pubbing

Added the search and compare api endpoint to sync with for bulk publishing

### Guidance to review
Based of #678 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/jobs/bulk_sync_courses_to_find_job.rb
+++ b/app/jobs/bulk_sync_courses_to_find_job.rb
@@ -1,0 +1,26 @@
+class BulkSyncCoursesToFindJob < ApplicationJob
+  queue_as :find_sync
+
+  def perform
+    syncable_courses = []
+
+    RecruitmentCycle.current_recruitment_cycle.providers
+      .includes(:latest_published_enrichment)
+      .each do |provider|
+        if provider.publishable?
+          syncable_courses.push(*provider.syncable_courses)
+        end
+      end
+
+    if !syncable_courses.empty?
+      request = SearchAndCompareAPIService::Request.new
+
+      unless request.bulk_sync(syncable_courses)
+        raise(RuntimeError.new(
+                "Error #{request.response.status} received syncing courses: " \
+                + courses.join('; ')
+              ))
+      end
+    end
+  end
+end
--- a/app/jobs/bulk_sync_courses_to_find_job.rb
+++ b/app/jobs/bulk_sync_courses_to_find_job.rb
@@ -2,17 +2,9 @@ class BulkSyncCoursesToFindJob < ApplicationJob
   queue_as :find_sync
 
   def perform
-    syncable_courses = []
+    syncable_courses = RecruitmentCycle.syncable_courses
 
-    RecruitmentCycle.current_recruitment_cycle.providers
-      .includes(:latest_published_enrichment)
-      .each do |provider|
-        if provider.publishable?
-          syncable_courses.push(*provider.syncable_courses)
-        end
-      end
-
-    if !syncable_courses.empty?
+    if syncable_courses.present?
       request = SearchAndCompareAPIService::Request.new
 
       unless request.bulk_sync(syncable_courses)
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -180,10 +180,7 @@ module MCB
     def apiv1_opts(opts)
       # the following lines are necessary to make opts work with double **splats and default values
       # See the change introduced in https://github.com/ddfreyne/cri/pull/99 (cri 2.15.8)
-      opts[:url] = opts[:url]
-      opts[:'max-pages'] = opts[:'max-pages']
-      opts[:token] = opts[:token]
-      opts[:all] = opts[:all]
+      opts = expose_opts_defaults_for_splat(opts, :url, :'max-pages', :token, :all)
 
       opts.merge! azure_env_settings_for_opts(**opts)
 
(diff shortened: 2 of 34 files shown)


defong:
[[L] Add bulk publish/sync from manage courses to search-and-compare.](https://trello.com/c/Plh80FJr/1971-l-add-bulk-publish-sync-from-manage-courses-to-search-and-compare)

app/jobs/bulk_sync_courses_to_find_job.rb
```diff
@@ -0,0 +1,26 @@
...
+  def perform
+    syncable_courses = []
+
+    RecruitmentCycle.current_recruitment_cycle.providers
+      .includes(:latest_published_enrichment)
+      .each do |provider|
+        if provider.publishable?
+          syncable_courses.push(*provider.syncable_courses)
+        end
+      end
```
benilovj: I think this would be clearer as something like (untested code):

```ruby
syncable_courses = RecruitmentCycle.current_recruitment_cycle.providers
                                     .includes(:latest_published_enrichment)
                                     .select(&:publishable?)
                                     .flat_map(&: syncable_courses)
```

app/jobs/bulk_sync_courses_to_find_job.rb
```diff
@@ -0,0 +1,26 @@
+class BulkSyncCoursesToFindJob < ApplicationJob
+  queue_as :find_sync
+
+  def perform
```
benilovj: This is relatively subjective, but I'm not a fan about having non-trivial logic in background jobs. I prefer the job being 1-2 lines long and it calling a method on a model or a service. This allows testing the logic separately from testing the job.

app/jobs/bulk_sync_courses_to_find_job.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    RecruitmentCycle.current_recruitment_cycle.providers
+      .includes(:latest_published_enrichment)
+      .each do |provider|
+        if provider.publishable?
+          syncable_courses.push(*provider.syncable_courses)
+        end
+      end
+
+    if !syncable_courses.empty?
```
benilovj:
```suggestion
    if syncable_courses.present?
```
or 
```
    unless syncable_courses.empty?
```

dankmitchell: Memories of the c# world 🤢

defong: Worming our way to get rid of c#...

app/jobs/bulk_sync_courses_to_find_job.rb
defong: Ran out off pint and energy...

defong: moved to recruitment cycle model

app/jobs/bulk_sync_courses_to_find_job.rb
defong: moved to recruitment cycle model
 to get the syncable courses

spec/jobs/bulk_sync_courses_to_find_job_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    syncable_courses
+
+    allow(RecruitmentCycle)
+      .to receive(:syncable_courses).and_return(syncable_courses)
+
+    allow_any_instance_of(SearchAndCompareAPIService::Request)
+      .to receive(:bulk_sync).and_return(true)
+  end
+
+  xit 'queues the expected job' do
```
misaka: Is this disabled for a reason?

misaka:
@defong This can be merged, no? Since it just won't do anything until we have a scheduled job.

defong:
@misaka It was based of #678 as I was kinda expecting the `delayed_job_active_record` stuff would allow for scheduling background job....

misaka:
> It was based of #678 as I was kinda expecting the `delayed_job_active_record` stuff would allow for scheduling background job....

oic, ya I missed that, fair point.

> I'm on the fence of dropping BulkSyncCoursesToFindJob and related items and just retain the other bit of code to

Do you mean just reuse `SyncCoursesToFindJob`? Wouldn't that be a problem since the bulk sync needs to call use a `POST`? I kind of like the cleanliness of it being a different job instead of passing some kind of parameter through to indicate it's a bulk sync.

defong:
I literally mean delete `BulkSyncCoursesToFindJob` class so its no longer dependant on #678 as the actual mechanism to schedule is still unknown. 

It may use 
- sidekiq or something railsy
- may add a endpoint and use azure function of something to hit endpoint to trigger it...

lib/mcb.rb
```diff
@@ -195,6 +195,38 @@ def apiv1_opts(opts)
...
+    #
+    # The opts passed in are examined determine which opts need to be added,
+    # this function essentially just fills in any missing options.
+    #
+    #   opts = apiv2_opts(opts)
+    def apiv2_opts(opts)
+      opts[:url] = opts[:url]
+      opts[:'max-pages'] = opts[:'max-pages']
+      opts[:token] = opts[:token]
+      opts[:all] = opts[:all]
```
faissalMT: The way opts is being used here is confusing, is the intent to ensure these things have been specified or something else?

lib/mcb.rb
```diff
@@ -195,6 +195,38 @@ def apiv1_opts(opts)
...
+
+      opts.merge! azure_env_settings_for_opts(**opts)
+
+      if requesting_remote_connection?(**opts)
+        opts[:url] = MCB::Azure.get_urls(**opts).first
+        opts[:token] = MCB::Azure.get_config(**opts)['AUTHENTICATION_TOKEN']
```
faissalMT: It seems confusing to conditionally overwrite the url and token, I think it'd be better if we assigned one or the other.

misaka:
Recent commit greatly optimises the retrieval and serialisation courses to send to Find by reducing the number of requests made to the DB. This is important to reduce load on the DB which is a shared resource with the backend/API. The timings below illustrates the differences, and if you consider that most of the savings have been in time spent in the DB, this can be quite valuable in reducing impact on backend:

Loading of courses, Old: 286s total (1073s real, take with some salt)
Loading of courses, New: 68s total (92s real)

Serializing of courses, Old: 204s total (238s real)
Serializing of courses, New: 59s total (70s real)

PR #743, opened 2019-08-27
[2001] Update program type

davidgisbey:
### Context

Providers' need to be able to update the study mode of their courses to be within a defined scope.

SCITT's self accredited  courses can either be 
`:pg_teaching_apprenticeship` or
`:scitt_programme`

HEI's self accredited courses can either be
`:pg_teaching_apprenticeship` or 
`:higher_education_programme`

Externally accredited courses can be
`:pg_teaching_apprenticeship` or 
`:school_direct_training_programme` or
`:school_direct_salaried_training_programme`

- Allow the study mode attribute to updated
- Validations 
- Refactor of concerns to model agnostic

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/models/course.rb
```diff
@@ -154,7 +154,7 @@ def self.entry_requirement_options_without_nil_choice
   validate :validate_course_syncable, on: :sync
 
   validate :validate_qualification, :validate_start_date, on: :update
-  validate :validate_qualification, on: :update
+  validate :validate_program_type, on: :update
```
defong: just styling

can we collapse 

`validate :validate_field_1, :validate_field_2, :validate_field_3, on: update`

 all in to one

spec/requests/api/v2/providers/courses/update_program_type_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  let(:permitted_params) do
+    %i[updated_program_type]
+  end
+
+  before do
+    perform_request(updated_program_type)
+  end
+
+  context "course has an updated age range in years" do
```
defong: ?

spec/requests/api/v2/providers/courses/update_program_type_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    it "returns http success" do
+      expect(response).to have_http_status(:success)
+    end
+
+    it "does not change program_type attribute" do
+      expect(course.reload.program_type).to eq(previous_program_type)
+    end
+  end
+
+  context 'with an invalid start date' do
```
defong: ?

davidgisbey:
[Edit 'Course Salaried/Fee Paying/Apprenticeship' - (MCBE)](https://trello.com/c/ONnMSdFO/2001-edit-course-salaried-fee-paying-apprenticeship-mcbe)

davidgisbey:
[Edit 'Course Apprenticeship' - (MCBE)](https://trello.com/c/fVNSGWht/2000-edit-course-apprenticeship-mcbe)

spec/requests/api/v2/providers/courses/update_program_type_spec.rb
davidgisbey: You may well have caught me in a copy paste error 😱

app/models/course.rb
davidgisbey: Have changed it. Thanks

app/models/course.rb
```diff
@@ -371,6 +368,14 @@ def scholarship_amount
     dfe_subjects&.first&.scholarship_amount
   end
 
+  def self_accredited?
+    accrediting_provider_id.blank?
```
```suggestion
       accrediting_provider.blank?
```

app/models/course.rb
```diff
@@ -371,6 +368,14 @@ def scholarship_amount
...
+  end
+
+  def provider_is_a_scitt?
+    provider.scitt == 'Y'
```
```suggestion
    provider.scitt?
```

spec/factories/courses.rb
```diff
@@ -93,6 +93,7 @@
 
     trait :with_accrediting_provider do
       association(:accrediting_provider, factory: :provider)
+      accrediting_provider_id { accrediting_provider.id }
```
defong: don't think that is needed

spec/models/concerns/courses/edit_options/age_range_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      include Courses::EditOptions::AgeRangeConcern
+
+      def level; end
+    end
+
+    klass.new
+  end
+
+  before do
+    allow(example_model).to receive(:level).and_return(level_value)
```
danielburnley: I was surprised to see this here, feels like a weird way to do it. 

I feel an easier way to handle this would be to just make `level` an `attr_accessor` on the model, and in the before block do `example_model.level = level_value`, as this makes it clearer that it's essentially setting `level` on the model rather than stubbing something out

spec/models/concerns/courses/edit_options/start_date_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      def provider; end
+    end
+
+    klass.new
+  end
+
+  let(:provider) { build(:provider) }
+
+  before do
+    allow(example_model).to receive(:provider).and_return(provider)
```
danielburnley: Similar to my other comment - I feel an easier way to do this would be setting the attribute directly

spec/models/concerns/courses/edit_options/program_type_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def provider_is_a_scitt?; end
+    end
+
+    klass.new
+  end
+
+  context 'for a SCITTs self accredited courses' do
+    let(:self_accredited_value) { true }
+    let(:provider_is_a_scitt_value) { 'Y' }
```
danielburnley: Is there a reason you have `Y` here, but `false` down below?

spec/models/concerns/courses/edit_options/program_type_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'rails_helper'
+
+describe Courses::EditOptions::ProgramTypeConcern do
+  let(:example_model) do
+    klass = Class.new do
+      include Courses::EditOptions::ProgramTypeConcern
+
+      def self_accredited?; end
```
danielburnley: If you set attributes directly and implement these methods - I think they'd be clearer to another reader as to what they're for. I find these pretty difficult to parse at first glance to understand why you'd have blank methods here

spec/models/concerns/courses/edit_options/qualifications_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      include Courses::EditOptions::QualificationConcern
+
+      def level; end
+    end
+
+    klass.new
+  end
+
+  before do
+    allow(example_model).to receive(:level).and_return(level_value)
```
danielburnley: Similar to my other comments :eyes:

fofr:
Is the program type a concept we need to continue with? Can it be something that is derived from a model for UCAS API purposes only?

Going forward would we be better served by a model property for "fee paying/salaried/apprenticeship"?

app/models/course.rb
```diff
@@ -152,10 +152,7 @@ def self.entry_requirement_options_without_nil_choice
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_enrichment
   validate :validate_course_syncable, on: :sync
-
-  validate :validate_qualification, :validate_start_date, on: :update
-  validate :validate_qualification, on: :update
-  validate :validate_applications_open_from, on: :update
+  validate :validate_qualification, :validate_start_date, :validate_program_type, :validate_applications_open_from, on: :update
```
misaka: Is this just a tidy-up? If I'm honest, I don't feel like this is an improvement, shoving it onto one line like this. (Clearly, though, the duplicate `validate_qualification` should be removed.)

app/models/course.rb
misaka: Is it even worth having a separate method for this? I feel like `provider.scitt?` is pretty clear about it's intentions.

app/models/course.rb
davidgisbey: Done. Thanks

app/models/course.rb
davidgisbey: Leaving as discussed.

spec/factories/courses.rb
davidgisbey: Removed now the model uses the accrediting provider

spec/models/concerns/courses/edit_options/age_range_spec.rb
misaka: Agreed. Particularly as in this case `example_model` is an anonymous class generated only for the purpose of testing.

app/models/course.rb
```diff
@@ -466,4 +471,11 @@ def validate_applications_open_from
   def valid_date_range
     recruitment_cycle.application_start_date..recruitment_cycle.application_end_date
   end
+
+  def validate_program_type
+    if provider.present? && program_type_options.exclude?(program_type.to_sym)
```
misaka: I think that the guard here, `provider.present?`, should actually be on the `validate` declaration up around line 155 of this file:

```
  validate :validate_program_type, on: :update, if: -> { |c| c.provider.present? }
```

All the more reason to have those `validate` declarations on separate lines.

app/models/course.rb
misaka: Also see [comment below on the implementation of `validate_qualification`](https://github.com/DFE-Digital/manage-courses-backend/pull/743/files#r319417509)

davidgisbey: I also prefer them separately for clarity to be honest, but Sau has been banging on about this haha

defong: I prefer consistency... 
One item per line or multiple item in a line vs mix and match that I saw...

app/models/course.rb
defong: `provider.present?` That was a test quirk... in reality a course will always have a provider...

benilovj:
> Is the program type a concept we need to continue with? Can it be something that is derived from a model for UCAS API purposes only?
> 
> Going forward would we be better served by a model property for "fee paying/salaried/apprenticeship"?

I suspect it may be relevant for the Allocations work and likely Apply, as it's relevant for DfE funding calculations.

app/models/course.rb
defong: was going to ask you about it at someday...

spec/models/concerns/courses/edit_options/qualifications_spec.rb
davidgisbey: Done.

spec/models/concerns/courses/edit_options/start_date_spec.rb
davidgisbey: Done.

spec/models/concerns/courses/edit_options/program_type_spec.rb
davidgisbey: Done.

app/models/provider.rb
```diff
@@ -301,6 +301,11 @@ def accredited_bodies
     end
   end
 
+  def is_it_really_really_a_scitt?
+    #is purposley named poorly to remind us that this needs to be refactored. We should be using the prgram type enum not an attribute
```
defong:
```suggestion
    #is purposely named poorly to remind us that this needs to be refactored. We should be using the program type enum not an attribute
```

app/models/course.rb
davidgisbey: Ok cool. Done!

app/models/course.rb
davidgisbey: Done.

spec/models/concerns/courses/edit_options/age_range_spec.rb
davidgisbey: Has been resolved. Thanks for the feedback

spec/models/concerns/courses/edit_options/program_type_spec.rb
davidgisbey: good spot. Fixed.Thanks

spec/models/concerns/courses/edit_options/age_range_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    klass.new
+  end
+
+  before do
+    example_model.level = level_value
+  end
+
+  context 'for primary' do
+    let(:level_value) { :primary }
+    it 'returns the correct ages range for users to co choose between' do
```
defong:
```suggestion
    it 'returns the correct ages range for users to choose between' do
```

spec/models/concerns/courses/edit_options/age_range_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      expect(example_model.age_range_options).to eq(%w[3_to_7 5_to_11 7_to_11 7_to_14])
+    end
+  end
+
+  context 'for secondary' do
+    let(:level_value) { :secondary }
+    it 'returns the correct age ranges for users to co choose between' do
```
defong:
```suggestion
    it 'returns the correct age ranges for users to choose between' do
```

spec/requests/api/v2/providers/courses/update_program_type_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  context 'with an invaild program type' do
+    let(:updated_program_type) { { program_type: :scitt_programme } }
+    let(:json_data) { JSON.parse(response.body)['errors'] }
+
+    it "returns an error" do
+      expect(response).to have_http_status(:unprocessable_entity)
+      expect(json_data.count).to eq 1
+      expect(response.body).to include("#{updated_program_type[:program_type]} is not valid for an externally accredited course")
```
defong: Can we have one more to test for `"a self accredited" ` error message

spec/factories/courses.rb
```diff
@@ -93,6 +93,7 @@
 
     trait :with_accrediting_provider do
       association(:accrediting_provider, factory: :provider)
+      accrediting_provider_code { accrediting_provider.provider_code }
```
misaka: Why is this needed? `accrediting_provider` is a `belongs_to`, although a fairly complex one, but I just tested this locally and it seems to work as expected without this change.

spec/models/concerns/courses/edit_options/qualifications_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    example_model.level = level_value
+  end
+
+  context 'for a course that’s not further education' do
+    let(:level_value) { :primary }
+
+    it 'returns only QTS options for users to choose between' do
+      expect(example_model.qualification_options).to eq(%w[qts pgce_with_qts pgde_with_qts])
+      example_model.qualification_options.each do |q|
+        expect(q.include?('qts')).to be_truthy
```
misaka:
```suggestion
        expect(q).to include('qts')
```

spec/models/concerns/courses/edit_options/qualifications_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  context 'for a further education course' do
+    let(:level_value) { :further_education }
+
+    it 'returns only QTS options for users to choose between' do
+      expect(example_model.qualification_options).to eq(%w[pgce pgde])
+      example_model.qualification_options.each do |q|
+        expect(q.include?('qts')).to be_falsy
```
misaka:
```suggestion
        expect(q).not_to include('qts')
```

spec/models/concerns/courses/edit_options/qualifications_spec.rb
misaka: Ugh. I just realised that these came out of `spec/models/concerns/courses/edit_options_spec.rb`. Sorry, I didn't catch on when you mentioned that these are being broken out from another spec file. This fix isn't so important then, as they're just being refactored out.

spec/models/concerns/courses/edit_options/qualifications_spec.rb
misaka: Ugh. I just realised that these came out of `spec/models/concerns/courses/edit_options_spec.rb`. Sorry, I didn't catch on when you mentioned that these are being broken out from another spec file. This fix isn't so important then, as they're just being refactored out.

spec/models/course_spec.rb
```diff
@@ -954,4 +954,18 @@
...
+
+  describe '#self_accredited?' do
+    let(:provider) { build(:provider) }
+
+    context 'when self accredited' do
+      subject { create(:course, provider: provider) }
+      its(:self_accredited?) { should be_truthy }
+    end
+
+    context 'when self accredited' do
```
```suggestion
    context 'externally accredited' do
```

misaka:
Hey @davidgisbey, sorry there's a few more comments. Almost there, though. :)

We discussed this on the chat, but just to note down on the PR itself, I think this really should've been more like 3 PRs:

- expose program_type in api v2 provider endpoint
- add update of program_type in api v2 provider endpoint
- refactor edit options into separate concerns

The last one we probably wouldn't have discovered until we started this work, fair enough, but the first two I think are useful distinctions to keep these PRs from getting too big. Sometimes it doesn't work to split the show/update up, but if we can we probably should.

spec/factories/courses.rb
davidgisbey: Yea that was left over a provider_id wasn't being assigned and my previous code was reliant on it. It should have just been removed though not changed to use the provider code. Have now taken in out.

spec/requests/api/v2/providers/courses/update_program_type_spec.rb
davidgisbey: done

spec/models/concerns/courses/edit_options/qualifications_spec.rb
davidgisbey: done.

app/models/provider.rb
```diff
@@ -301,6 +301,11 @@ def accredited_bodies
     end
   end
 
+  def is_it_really_really_a_scitt?
+    #is purposely named poorly to remind us that this needs to be refactored. We should be using the prgram type enum not an attribute
```
timabell: I'm not sure this naming is a good idea, this stuff tends to spread rather than have the desired effect

timabell: If there's further work to do I'd probably card it up and chase it being prioritised, otherwise maybe it isn't that important.

davidgisbey: That's the plan. Misha is currently carding it, but he wanted in named poorly to ensure we prioritise it.

PR #747, opened 2019-08-28
Deleted travis files and updated database.yml

rizzkhan1:
### Context
Cleanup source code and fix inconsistencies with environment files

Delete Travis build pipeline related files
Delete and rename environment settings files


config/settings/development.yml
```diff
@@ -1,2 +1,6 @@
-authentication:
-  secret: secret
+manage_api:
```
dankmitchell: Keep this file as it was, its for local development only

rizzkhan1: done

PR #749, opened 2019-08-28
[1877] Add recruitment cycle to Course.to_s

faissalMT:
### Context
As it stands courses that have been rolled over cannot be distinguished from their initial versions.

Change `Provider.to_s()` to print the provider code, course code, name and recruitment cycle e.g: `[A1/QM1] Introductory Quantum Mechanics (2019/20)`  
  
### Checklist

- [x] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -372,7 +372,7 @@ class Course < ApplicationRecord
   end
 
   def to_s
-    "#{name} (#{course_code})"
+    "#{name} (#{provider.provider_code}/#{course_code}) [#{recruitment_cycle}]"
   end
 
   def next_recruitment_cycle?
--- a/app/models/nctl_organisation.rb
+++ b/app/models/nctl_organisation.rb
@@ -6,8 +6,6 @@
 #  name            :text
 #  nctl_id         :text             not null
 #  organisation_id :integer
-#  urn             :integer
-#  ukprn           :integer
 #
 
 class NCTLOrganisation < ApplicationRecord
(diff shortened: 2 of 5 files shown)


faissalMT:
[[S] Add recruitment year and provider code to `Course#to_s`](https://trello.com/c/usEjgTt1/1877-s-add-recruitment-year-and-provider-code-to-coursetos)

app/models/course.rb
```diff
@@ -372,7 +372,7 @@ def scholarship_amount
   end
 
   def to_s
-    "#{name} (#{course_code})"
+    "[#{course_code}] #{name} (#{recruitment_cycle})"
```
fofr: Can we keep the course code at the end? `"#{name} (#{course_code})"` is the consistent way we display courses everywhere.

Maybe:
`"#{name} (#{course_code}) [#{recruitment_cycle}]"`?

fofr: (Or front load the recruitment cycle)

faissalMT: I find that hard to read, software and code generally tends to place the unique code before details roughly in order of importance (unless it's extremely long, like a hash or memory address), and I've already merged in [a change](https://github.com/DFE-Digital/manage-courses-backend/pull/748) similar to this.

misaka: What about copying the slash as it's in the URL?

```
"#{name} (#{recruitment_cycle}/#{course_code})"
```

faissalMT: The slash is already being used for the provider code, putting the recruitment cycle there would look like (2019/20/QM1)

misaka: Is that bad? That's still easy to parse:

```
course_parts = codes.s

misaka: I only bring this up because I've wanted to try to normalise before, it'd be handy for supplying to `mcb`, for example ... you could provider multiple courses on the cmdline if we had a standard way to serialise their identities like this.

faissalMT: This is for debug, so my concern here is readability. I don't think anything should be trying to parse this.

PR #756, opened 2019-08-29
[2040] Change fields to not editable when published

faissalMT:
### Context
The fields 'SEND', 'applications open' and 'start date' shouldn't be changed after they've been published.  

Wrap `course_params_assignable()` in tests and prevent the changing of `is_send`, `applications_open_from` and `application_start_date` when published.  

### Guidance to review
At the moment this only checks whether `content_status` is published to decide whether to permit the change, it may be the case that we should disable it with other statuses. Also it seems that the error message `enrichments: []` persists regardless of whether the data has any errors, this is a side issue that perhaps warrants further discussion.  

### Checklist

- [x] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -386,23 +386,44 @@ class Course < ApplicationRecord
   # Ideally this would just use the validation, but:
   # https://github.com/rails/rails/issues/13971
   def course_params_assignable(course_params)
-    entry_requirements_assignable(course_params) &&
+    assignable_after_publish(course_params) &&
+      entry_requirements_assignable(course_params) &&
       qualification_assignable(course_params)
   end
 
+  def is_published?
+    content_status == :published
+  end
+  
 private
 
+  def assignable_after_publish(course_params)
+    relevant_params = course_params.slice(:is_send, :applications_open_from, :application_start_date)
+
+    return true if relevant_params.empty? || !is_published?
+
+    relevant_params.each do |field, _value|
+      errors.add(field.to_sym, 'cannot be changed after publish')
+    end
+    false
+  end
+
   def entry_requirements_assignable(course_params)
-    course_params.slice(:maths, :english, :science)
-      .to_h
-      .select { |_subject, value| value && !ENTRY_REQUIREMENT_OPTIONS.key?(value.to_sym) }
-      .map { |subject, _value| errors.add(subject.to_sym, "is invalid") }
-      .empty?
+    relevant_params = course_params.slice(:maths, :english, :science)
+
+    invalid_params = relevant_params.select do |_subject, value|
+      value && !ENTRY_REQUIREMENT_OPTIONS.key?(value.to_sym)
+    end
+    invalid_params.each do |subject, _value|
+      errors.add(subject.to_sym, 'is invalid')
+    end
+
+    invalid_params.empty?
   end
 
   def qualification_assignable(course_params)
     assignable = course_params[:qualification].nil? || Course::qualifications.include?(course_params[:qualification].to_sym)
-    errors.add(:qualification, "is invalid") unless assignable
+    errors.add(:qualification, 'is invalid') unless assignable
 
     assignable
   end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -394,7 +394,7 @@ class Course < ApplicationRecord
   def is_published?
     content_status == :published
   end
-  
+
 private
 
   def assignable_after_publish(course_params)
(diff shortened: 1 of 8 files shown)


faissalMT:
[Change fields to not editable when published](https://trello.com/c/Y79livE0/2040-change-fields-to-not-editable-when-published)

app/models/course.rb
```diff
@@ -382,23 +382,33 @@ def next_recruitment_cycle?
...
   def course_params_assignable(course_params)
-    entry_requirements_assignable(course_params) &&
+    assignable_after_publish(course_params) &&
+      entry_requirements_assignable(course_params) &&
       qualification_assignable(course_params)
   end
 
 private
 
+  def assignable_after_publish(course_params)
```
danielburnley: I find this massive chain of methods pretty hard to follow what's happening, wonder if there's a way to split it up into a few intermediate variables/methods

spec/models/course_spec.rb
```diff
@@ -954,4 +954,71 @@
...
+        expect(course.errors.messages).to eq(enrichments: [])
+      end
+
+      it 'cannot assign invalid qualification' do
+        expect(course.course_params_assignable(qualification: 'invalid')).to eq(false)
+        expect(course.errors.messages).to eq(enrichments: [], qualification: ['is invalid'])
+      end
+    end
+
+    context 'published' do
```
danielburnley: > context published
> context not published

Seems like there's some weird nesting :s

timabell:
I don't think we should merge this till the frontend prevents users hitting it

faissalMT:
https://github.com/DFE-Digital/manage-courses-frontend/pull/570 has already been merged, only thing that remains is having the frontend use the edit options provided here instead of doing the logic itself

app/models/course.rb
```diff
@@ -386,23 +390,39 @@ def next_recruitment_cycle?
...
+    relevant_params = course_params.slice(:is_send, :applications_open_from, :application_start_date)
+
+    invalid_params = relevant_params.select { is_published? }
```
misaka: This pattern threw me ... you're selecting on something unrelated to any of the params being iterated through. I think a guard would be more appropriate, or at least something that more clearly expresses that if the course is published the list of invalid params is empty.

spec/models/course_spec.rb
```diff
@@ -968,4 +968,88 @@
       its(:self_accredited?) { should be_falsey }
     end
   end
+
+  describe '#course_params_assignable' do
+    context 'entry requirement' do
```
```suggestion
    describe 'when setting the entry requirement' do
```

spec/models/course_spec.rb
```diff
@@ -968,4 +968,88 @@
...
+        expect(course.errors.messages).to eq(enrichments: [])
+      end
+
+      it 'cannot be assigned an invalid value' do
+        expect(course.course_params_assignable(maths: 'test')).to eq(false)
+        expect(course.errors.messages).to eq(enrichments: [], maths: ['is invalid'])
+      end
+    end
+
+    context 'qualification' do
```
```suggestion
    describe 'when setting the qualification' do
```

spec/models/course_spec.rb
```diff
@@ -968,4 +968,88 @@
...
+        expect(course.errors.messages).to eq(enrichments: [])
+      end
+
+      it 'cannot assign invalid qualification' do
+        expect(course.course_params_assignable(qualification: 'invalid')).to eq(false)
+        expect(course.errors.messages).to eq(enrichments: [], qualification: ['is invalid'])
+      end
+    end
+
+    context 'published' do
```
```suggestion
    describe 'for publishing' do
```

spec/models/course_spec.rb
```diff
@@ -968,4 +968,88 @@
...
+      let(:user) { create(:user) }
+
+      context 'not published' do
```
```suggestion
      context 'when not published' do
```

spec/models/course_spec.rb
```diff
@@ -968,4 +968,88 @@
...
+          expect(course.errors.messages).to eq(enrichments: [])
+        end
+
+        it 'can assign to applications open from' do
+          expect(course.course_params_assignable(application_start_date: '25/08/2019')).to eq(true)
+          expect(course.errors.messages).to eq(enrichments: [])
+        end
+      end
+
+      context 'course is published' do
```
```suggestion
      context 'when is published' do
```

spec/models/course_spec.rb
misaka: Also, maybe add the setup of course in here (with a draft enrichment)?

PR #758, opened 2019-08-30
[2018] Add accrediting_provider_code to course deserializer

tvararu:
### Context

Editing accredited body.

This lines up the behaviour for this attribute to the way the other attributes work, which also lines up with how the frontend sends them across. Brought the tests into line as well.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu:
[Edit course accredited body (MCFE)](https://trello.com/c/sC2haA98/2018-edit-course-accredited-body-mcfe)

spec/requests/api/v2/providers/courses/update_accrediting_provider_spec.rb
```diff
@@ -0,0 +1,81 @@
+require "rails_helper"
+
+describe 'PATCH /providers/:provider_code/courses/:course_code' do
```
defong: This test for 
response `success`
and database has mad the change

but shouldn't it should also check that the response as the `accrediting_code ` changed as well

like it did in the original spec?

timabell: perhaps, I'm not sure it matters much? We aren't using the returned value

tvararu: Yes; it shouldn't matter because we're not using it.

spec/requests/api/v2/providers/courses/update_accrediting_provider_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect(response).to have_http_status(:success)
+    end
+
+    it "updates the `accrediting_provider_code` attribute to the correct value" do
+      expect(course.reload.accrediting_provider_code).to eq(accrediting_provider_code[:accrediting_provider_code])
+    end
+  end
+
+  context "course has the same SEND value" do
+    context "with values passed into the params" do
```
misaka: I'm not seeing how `course` has "the same SEND value" from the setup below ... has anything changed from the test above? Even the code inside the spec is the same, even though the description is different, what's different?

tvararu: copy pastaaa 🙈

timabell: whoops, it was copy pasted from the SEND test

timabell:
rebased, conflicts fixed

timabell:
Fixed tests, we had a random trailing backtick after issend. Altered the commit because it's not an interesting diff.

tvararu:
@timabell rebased and force pushed to fix flappy test

timabell:
oh, communication fail

defong:
> oh, communication fail

both physical and virtual...

PR #759, opened 2019-08-30
[2044] add created at and changed at timestamp for providers to api v1

defong:
### Context
v1 api providers endpoint
created at and changed at timestamp for providers

Added created at and changed at timestamp for providers to api v1
Amended docs to reflect changes

### Guidance to review
`/api/v1/2019/providers`

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/serializers/provider_serializer.rb
+++ b/app/serializers/provider_serializer.rb
@@ -32,7 +32,7 @@ class ProviderSerializer < ActiveModel::Serializer
 
   attributes :institution_code, :institution_name, :institution_type, :accrediting_provider,
              :address1, :address2, :address3, :address4, :postcode, :region_code, :scheme_member,
-             :utt_application_alerts, :type_of_gt12
+             :utt_application_alerts, :type_of_gt12, :created_at, :changed_at
 
              # application alert recipient has not been added into the enum in the contact model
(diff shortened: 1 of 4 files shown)


defong:
[Add "created_at" & "changed_at" timestamps for providers to API V1](https://trello.com/c/zXJ8mDne/2044-add-created-at-timestamp-for-providers-to-api-v1)

app/serializers/provider_serializer.rb
```diff
@@ -88,6 +88,14 @@ def recruitment_cycle
     object.recruitment_cycle.year
   end
 
+  def created_at
+    object.changed_at.iso8601
```
misaka: Shouldn't this be `created_at`?

misaka: Seems like the tests should ensure that these are two different timestamps to test thoroughly.

Also, did you check to see if it just Does The Right Thing if you don't even bother with these methods?

defong: "Fri, 30 Aug 2019 16:19:35 UTC +00:00"
vs
"2019-08-30T16:26:33Z" (iso)

PR #767, opened 2019-09-03
[2068] Add tests to ensure next cycles courses can be published

davidgisbey:
### Context
Providers need to be able to select publish on courses in their next cycle. 

There are no validations on the backend to stop  next cycles courses being set to published. There is simply no button for users to click on the frontend.

- Add tests to check that publish works for next cycles courses

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


davidgisbey:
[Next cycle publishing (MCBE)](https://trello.com/c/vUIsZkiP/2068-next-cycle-publishing-mcbe)

PR #768, opened 2019-09-03
[2085] Create endpoint for triggering bulk sync

faissalMT:
### Context
We had this functionality in the C# version via WebJobs as Azure allowed us to manually trigger it. We would like to replicate this functionality by creating an endpoint that when sent a message will trigger the jobs.  

Add the /api/v2/sync endpoint that when pinged will queue another sync  

### Checklist

- [x] Attach to Trello card
- [ ] ~~Rebased master~~
- [X] Cleaned commit history

### DIFF
--- a/app/controllers/api/system/application_controller.rb
+++ b/app/controllers/api/system/application_controller.rb
@@ -1,0 +1,13 @@
+module API
+  module System
+    class ApplicationController < ::ApplicationController
+      before_action -> { skip_authorization }
+
+      def authenticate
+        authenticate_or_request_with_http_token do |token|
+          ActiveSupport::SecurityUtils.secure_compare(token, Rails.application.config.system_authentication_token)
+        end
+      end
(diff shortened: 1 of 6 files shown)


faissalMT:
[Create an endpoint for triggering bulk sync jobs](https://trello.com/c/WmDQ2oZi/2085-create-an-endpoint-for-triggering-bulk-sync-jobs)

faissalMT:
My understanding was that API v1 was legacy, rather than about the actor. If your concern is MCB we already have a way to generate v2 tokens:
https://github.com/DFE-Digital/manage-courses-backend/blob/master/lib/mcb.rb#L103

misaka:
I read @defong's suggestion as not to add this to api v1, but to change the authentication mechanism to not be based on a specific user, but some kind of a shared secret, like in api v1. Good suggestion. 👍

PR #772, opened 2019-09-04
[2052] Add initial accredited provider search endpoint

faissalMT:
### Context
We would like to allow users to search accredited providers when adding a course, in order to improve UX so we need a way to query the API for suggestions.  
```
GET /suggest?query=scitt
or
GET /suggest?query=1AA

[ { "providerCode": "1AA", "name": "Big Scitt" } ]
```

This adds the preliminary stages of the provider search endpoint.  

### Guidance to review
Given we are using jsonapi the actual output is somewhat different from the specified output e.g:
```
[
  {
    'id': '1',
    'type': 'provider',
    'attributes': {
      'provider_name': 'Name',
      'provider_code': 'Code'
    }
  }
]
```

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/controllers/api/v2/providers_controller.rb
```diff
@@ -77,6 +77,14 @@ def sync_courses_with_search_and_compare
         head :ok
       end
 
+      def suggest
+        authorize Provider
+        render(
+          jsonapi: Provider.where(id: @current_user.providers),
```
defong: current recruitment cycle?

faissalMT:
[[M] Add endpoint to search for accredited providers](https://trello.com/c/YHINYdfL/2052-m-add-endpoint-to-search-for-accredited-providers)

defong:
can we have a min of 3 chars for `query` please and it starts with alphanumeric value

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'rails_helper'
+
+describe 'GET /suggest' do
+  let(:jsonapi_renderer) { JSONAPI::Serializable::Renderer.new }
+  let(:organisation) { create(:organisation) }
+  let(:provider1) { create(:provider, organisations: [organisation]) }
+  let(:provider2) { create(:provider, organisations: [organisation]) }
+  let(:next_recruitment_cycle) { find_or_create(:recruitment_cycle, :next) }
+  let(:provider3) { create(:provider, provider_name: provider2.provider_name, organisations: [organisation], recruitment_cycle: next_recruitment_cycle) }
+  let(:unauthorized_provider) { create(:provider, organisations: [create(:organisation)]) }
```
defong: what is that one for?

faissalMT: That was for making sure we weren't spitting out providers the user wasn't in

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  before do
+    provider1
+    provider2
+    provider3
+    unauthorized_provider
+  end
+
+  it 'searches for a particular provider for the user for the current recruitment cycle' do
+    get "/api/v2/providers/suggest?query=#{provider2.provider_name}",
```
defong: /api/v2/providers/suggest?query=#{provider2.provider_name[0, n]}

where n is less then 3 fail
where n  + 3 to max succed

spec/requests/api/v2/providers/suggest_provider_spec.rb
defong: @tvararu is that right?

tvararu: I think it should be all providers actually, not just the ones they have access to

tvararu: ^ confirmed by Paul H, it's all the accredited bodies in the system

spec/requests/api/v2/providers/suggest_provider_spec.rb
faissalMT: See https://github.com/DFE-Digital/manage-courses-backend/pull/772/files#diff-117aabed956114d91bb7d529138c0b76R47

defong: That's not the same.

I prefer the test to use values from `provider2.provider_name` if its less then 3 chars it should fail 
`Ge` simply isn't the same as its not the subset of `provider2.provider_name` as we known that `provider2.provider_name` is in there where as `Ge` is questionable

if its 3 to the whole name it should return `provider2`

I would expect given a `provider_name` as `Paul's Uni`

fail
``
`P` 
`Pa`

it failes as the query is less then 3 chars from provider2 provider name

passes
`Pau`
`Paul`
`Paul's` 
`Paul's U` 
`Paul's Un`

This was what was tested
`Paul's Uni`

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'rails_helper'
+
+describe 'GET /suggest' do
+  let(:jsonapi_renderer) { JSONAPI::Serializable::Renderer.new }
+  let(:organisation) { create(:organisation) }
+  let(:provider1) { create(:provider, organisations: [organisation]) }
+  let(:provider2) { create(:provider, organisations: [organisation]) }
+  let(:next_recruitment_cycle) { find_or_create(:recruitment_cycle, :next) }
+  let(:provider3) { create(:provider, provider_name: provider2.provider_name, organisations: [organisation], recruitment_cycle: next_recruitment_cycle) }
```
defong: can we have a test for that as well please

 it 'searches for a particular provider for the user for the next recruitment cycle' do

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        headers: { 'HTTP_AUTHORIZATION' => credentials }
+
+    expect(response.status).to eq(400)
+  end
+
+  it 'returns bad request if query is too short' do
+    get '/api/v2/providers/suggest?query=Ge',
+        headers: { 'HTTP_AUTHORIZATION' => credentials }
+
+    expect(response.status).to eq(400)
```
defong: Does the frontend knows how to handle 400?

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    provider1
+    provider2
+    provider3
+  end
+
+  it 'searches for a particular provider for the user for the current recruitment cycle' do
+    get "/api/v2/providers/suggest?query=#{provider2.provider_name}",
+        headers: { 'HTTP_AUTHORIZATION' => credentials }
+
+    expect(JSON.parse(response.body)['data']).to match_array([
```
defong: also like to see

it 'searches for a non existing provider for the user for the current recruitment cycle' do
    get "/api/v2/providers/suggest?query=#{non existing}",

spec/requests/api/v2/providers/suggest_provider_spec.rb
faissalMT: It won't make a difference whether the provider exists or not, a query should be rejected before it gets to that stage or you might be querying unnecessarily.

spec/requests/api/v2/providers/suggest_provider_spec.rb
faissalMT: Not yet

spec/requests/api/v2/providers/suggest_provider_spec.rb
faissalMT: Done

app/controllers/api/v2/providers_controller.rb
```diff
@@ -77,8 +77,27 @@ def sync_courses_with_search_and_compare
         head :ok
       end
 
+      def suggest
```
dankmitchell: Are we sticking with `suggest`? I think I prefer `search`

app/serializers/api/v2/serializable_provider_suggestion.rb
```diff
@@ -0,0 +1,9 @@
+module API
+  module V2
+    class SerializableProviderSuggestion < JSONAPI::Serializable::Resource
```
dankmitchell: `SerializableProviderSearch` ?

spec/requests/api/v2/providers/suggest_provider_spec.rb
defong: It is a bit pointless, but my justification, is that it is more aligned to real usage

A user will start typing and it will start querying the backend.
Until the user stops typing, each char inputted will cause a http request.

A simple test of 
`get "/api/v2/providers/suggest?query=#{provider2.provider_name}"`
only ever occurs with copy and pasting

So the caller may not have validated input
hence that is why I wanted testing using from 0 to the full provider name.

None of the tests here shows the expectation of a query that has the partial of the full provider name, which is just as important as testing for the full provider name.

endpoints that is expected to be called multiple times in a short space of time warrants a bit more test coverage.

I settle for one partial instead of the whole enumeration to getting to the full provider name

defong: > It won't make a difference whether the provider exists or not

Not sure what you meant?

```
it 'searches for a non existing provider for the user for the current recruitment cycle' do
get "/api/v2/providers/suggest?query=#{non existing}",
```

if you meant that, it  matters as would of hit the database and 0 rows would of returned

faissalMT: Because the database should never be hit if the query is too short.

app/controllers/api/v2/providers_controller.rb
faissalMT: I feel that `suggest` describes the intended usecase better, and regardless it's the terminology I've already used across both PRs and I think it's much more hassle than its worth to change that now unless we decide it's a significant enough hindrance.

dankmitchell:
SHIP IT 🚢

PR #777, opened 2019-09-05
2055 add build new course endpoint

timabell:
### Context

Take 34

#### What

Add an endpoint that the course creation steps can use to build a new course to validate it and retrieve relevant metadata (like the edit options) without persisting it.

Just to clarify, this does not persist any changes, it only validates the attributes being set and calculates new values for the meta values.

#### Why

The course creation steps will need to be able to present the users with the correct options for each step and with any validation errors before it actually saves the course.

* Make build_new return errors or model, and always return edit_options

### Guidance to review

```
im@max:~/repo/dfe/manage/backend(2055-add-build-new-course-endpoint)
$ be rspec spec/requests/api/v2/providers/courses/build_new_spec.rb -fd
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}

GET /providers/:provider_code/courses/build_new
  #build_new
    with no parameters
      returns edit_options and errors but no model
    with enough attributes set in query parameters to make a valid course
      returns the course model and edit_options with no errors
    with an invalid attribute set in query parameters
      returns edit_options and errors but no model
    with an invalid enum attribute set in query parameters
      returns edit_options and errors but no model

Finished in 1.8 seconds (files took 2.19 seconds to load)
4 examples, 0 failures
```

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -18,7 +18,11 @@ module API
 
         @course = @provider.courses.new
 
-        render jsonapi: @course, include: params[:include]
+        if build_new_course(course_params)
+          render jsonapi: @course
+        else
+          render jsonapi_errors: @course.errors, status: :unprocessable_entity
+        end
       end
 
       def index
@@ -105,6 +109,17 @@ module API
         enrichment.assign_attributes(enrichment_params)
         enrichment.status = :draft if enrichment.rolled_over?
         enrichment.save
+      end
+
+      def build_new_course(course_params)
+        # first request has params so return true to indicate success
+        return true unless course_params.values.any?
+
+        # add any enum related validation errors
+        return unless @course.course_params_assignable(course_params)
+
+        @course.assign_attributes(course_params)
+        @course.valid?
       end
 
       def update_course
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -7,13 +7,13 @@ module API
 
       before_action :build_recruitment_cycle
       before_action :build_provider
-      before_action :build_course, except: %i[index new create]
+      before_action :build_course, except: %i[index build_new create]
 
       deserializable_resource :course,
                               only: %i[update publish publishable create],
                               class: API::V2::DeserializableCourse
 
-      def new
+      def build_new
         authorize Course
 
         @course = @provider.courses.new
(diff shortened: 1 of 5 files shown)


timabell:
[Add "courses/build_new" endpoint to support course creation steps](https://trello.com/c/o5qIimQo/2055-add-courses-buildnew-endpoint-to-support-course-creation-steps)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -7,18 +7,22 @@ class CoursesController < API::V2::ApplicationController
...
                               class: API::V2::DeserializableCourse
 
-      def new
+      def build_new
         authorize Course
 
         @course = @provider.courses.new
 
-        render jsonapi: @course, include: params[:include]
+        if build_new_course(course_params)
```
```suggestion
        build_new_course(course_params)
        if @course.valid?
```

think it reads better
build the thing than ask is it valid

rather then build it and ask is it built...

spec/requests/api/v2/providers/courses/build_new_spec.rb
```diff
@@ -0,0 +1,139 @@
...
+  def course_to_jsonapi(course)
+    rendered_course = jsonapi_renderer.render(
+      course,
+      class: {
+        Course: API::V2::SerializableCourse
+      }
+    )
+    # Converting the rendered course to JSON and back normalises any symbols
+    # into strings. Better #deep_symbolize_keys because it also does values.
+    JSON.parse(rendered_course.to_json)
```
defong: maybe 
https://api.rubyonrails.org/v4.2/classes/ActiveSupport/HashWithIndifferentAccess.html

json parsing is normally expensive...

timabell: This is copy pasted from other tests, I don't really understand it

app/controllers/api/v2/courses_controller.rb
```diff
@@ -107,6 +111,18 @@ def update_enrichment
         enrichment.save
       end
 
+      def build_new_course(course_params)
+        # Skip validations when building a new course with no params,
+        # so that frontend can get the initial edit_options:
+        # return true unless course_params.values.any?
```
timabell: needs sorting out

timabell:
I've read the JSON API spec https://jsonapi.org/format/ and it appears to conflict with the requirements of our `build_new` endpoint. Notably:

* Requires id on resources (objects) returned from the server (we aren't persisting anything so there is no id).
* Only supports the four CRUD operations, to the exclusion of all else, so even though `meta` is available there is no request we could attach our generated course to.

I've talked this over briefly with @misaka and we've agreed to ditch json-api for this and create a vanilla json endpoint specifically for course building.

I'll start a new branch to draft this out.

PR #779, opened 2019-09-06
[2091] Remove program type validations

davidgisbey:
### Context

This validations is causing errors for quite a few users. It seems like we may have a fundamental misunderstanding of how accrediting providers may work. SCITTs may in fact be able to be externally accredited by universities.

Our data surrounding SCITTs is also incorrect.

Until the above is resolved we should not have validations on program type to allow users to continue to update their courses.

- Remove program type validations

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


faissalMT:
[Refactor program_type in edit options to hide business logic](https://trello.com/c/HMEGUlpt/2091-refactor-programtype-in-edit-options-to-hide-business-logic)

PR #784, opened 2019-09-10
[2120] Use equal in flapping test for better output

timabell:
### Context

* Currently diff is truncated so can't see where the problem is.
* https://github.com/DFE-Digital/manage-courses-backend/commit/02b61bd63d59a8ee3816ce791f99a00af663d019
* https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=20565&view=ms.vss-test-web.build-test-results-tab

* Use equal in flapping test for better output

### Guidance to review

:ship: if the test still passes

Note that  match_array copes with differences in ordering but I don't
think that matters.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/recruitment_cycle_spec.rb
+++ b/spec/requests/api/v2/recruitment_cycle_spec.rb
@@ -108,7 +108,7 @@ describe '/api/v2/recruitment_cycle', type: :request do
         perform_request
 
         expect(json_response)
-          .to(match_array(
+          .to(eq(
                 'data' => [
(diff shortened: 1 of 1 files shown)


timabell:
[flapping json matcher test in recruitment_cycle_spec.rb](https://trello.com/c/4ZULlCHs/2120-flapping-json-matcher-test-in-recruitmentcyclespecrb)

PR #800, opened 2019-09-13
2128 course level migration

dankmitchell:
### Context
Course subjects > subjects & level

Add new level attribute

### Guidance to review
🚢 

We already have a level attribute on course so the specs don't need updating.

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


db/migrate/20190912131116_add_level_to_course.rb
```diff
@@ -0,0 +1,5 @@
+class AddLevelToCourse < ActiveRecord::Migration[5.2]
+  def change
+    add_column :course, :level, :integer, default: 0
```
misaka: My preference is that we use strings for enums, just because it makes it sssoooo much easier to write SQL by hand. However, there is the obvious downside of ... efficiency. There's two possible ways to address that:

* index the column. This would take a little bit of research, but I'd expect Postgres to introduce efficiencies to an indexed column that would be close to what we'd get by using an integer for an enum (iirc MySQL did this).
* use the native Postgres enum type. This would allow us to refer to enum values by string, but store them as integers. However, it's not clear on what the support for this is in Rails ... last time I tried to do this it meant having to dump the schema as SQL which was painful. I think there may be better support for it now, either built-in or as a gem, but it would take some investigation.

misaka: Some ideas for using the native Postgres support for enums:

- [ar-enum](https://github.com/fnando/ar-enum) - This sounds like it's meant to do exactly what we want, but has very little uptake on it.
- There are [other gems that add extended enum support](https://www.ruby-toolbox.com/categories/Active_Record_Enumerations), some of these are bound to support Postgres native enums.

dankmitchell: Interesting. I was under the assumption that strings are flakier since they could change whereas an integer doesn't...

misaka: Integers can be miss-entered too, if the column isn't an enum in the underlying DB, either way we're relying on AR's enum type to restrict what we can put in there. Really, the ultimate solution is to use AR enums backed with Postgres native enums.

app/models/course.rb
```diff
@@ -66,6 +67,12 @@ class Course < ApplicationRecord
     other: "O",
   }
 
+  enum level: {
+    primary: 'Primary',
+    secondary: 'Secondary',
+    further_education: 'Further education'
+  }, _prefix: :course
```
davidgisbey: Shoukld this be suffix instead of prefix?

dankmitchell: Good spot

db/migrate/20190912131116_add_level_to_course.rb
dankmitchell: I think i've solved this now

dankmitchell:
[Add level field to course table](https://trello.com/c/s5S0Aj8j/2128-add-level-field-to-course-table)

defong:
Rebased and fixed minor issues

db/migrate/20190916155318_clean_up_course_subject_data_post_migration.rb
```diff
@@ -0,0 +1,60 @@
+# rubocop:disable Metrics/BlockLength
+class CleanUpCourseSubjectDataPostMigration < ActiveRecord::Migration[5.2]
+  def change
+    say_with_time 'cleansing subject data' do
+      courses = RecruitmentCycle.second.courses.includes(:subjects)
```
misaka: Please lets not do `.second`, we can't guarantee sorting on this.

misaka: We should hardcode this to the recruitment cycle for "2020" ... that's when we clean up the subjects.

PR #807, opened 2019-09-16
[2129] change validation so start date can be in the  current cycle

davidgisbey:
### Context

Allow users to be able to pick a date in the current cycle. The validation currently only allows the start date to be within the next cycle. 

- Expand the options sent to the front end for users to select from to the current and next cycle

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


davidgisbey:
[Change start_date validation to check the date is in the next or current cycle](https://trello.com/c/5aYHkwR3/2129-change-startdate-validation-to-check-the-date-is-in-the-next-or-current-cycle)

PR #808, opened 2019-09-16
[1858] Allow the modification of course subjects

faissalMT:
### Context
We would like to give users the ability to edit course subjects.

Allow the course subject to be modified with `PATCH` and deserialize it correctly.

### Guidance to review
Only validation left is for MFL. Also is it correct that without ucas subjects the ucas status defaults to secondary.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -21,19 +21,29 @@ module API
 
       def build_new
         authorize @provider
-        course = Course.new(provider: @provider)
-        course.assign_attributes(course_params)
-        course.valid?
+        @course = Course.new(provider: @provider)
+        update_subjects
+        @course.assign_attributes(course_params)
+        @course.valid?
 
+        # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
         json_data = JSONAPI::Serializable::Renderer.new.render(
-          course,
-          class: { Course: API::V2::SerializableCourse },
+          @course,
+          class: {
+            Course: SerializableCourse,
+            Subject: SerializableSubject,
+            PrimarySubject: SerializableSubject,
+            SecondarySubject: SerializableSubject,
+            ModernLanguagesSubject: SerializableSubject,
+            FurtherEducationSubject: SerializableSubject,
+          },
+          include: [:subjects],
         )
 
         json_data[:data][:errors] = []
 
-        course.errors.messages.each do |error_key, _|
-          course.errors.full_messages_for(error_key).each do |error_message|
+        @course.errors.messages.each do |error_key, _|
+          @course.errors.full_messages_for(error_key).each do |error_message|
             json_data[:data][:errors] << {
               "title" => "Invalid #{error_key}",
               "detail" => error_message,
@@ -53,7 +63,20 @@ module API
       end
 
       def show
-        render jsonapi: @course, include: params[:include]
+        # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
+        render jsonapi: @course, include: params[:include], class: {
+          Course: SerializableCourse,
+          SiteStatus: SerializableSiteStatus,
+          Site: SerializableSite,
+          Subject: SerializableSubject,
+          PrimarySubject: SerializableSubject,
+          SecondarySubject: SerializableSubject,
+          ModernLanguagesSubject: SerializableSubject,
+          FurtherEducationSubject: SerializableSubject,
+          Provider: SerializableProvider,
+          ProviderEnrichment: SerializableProviderEnrichment,
+          RecruitmentCycle: SerializableRecruitmentCycle,
+        }
       end
 
       def sync_with_search_and_compare
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -29,14 +29,7 @@ module API
         # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
         json_data = JSONAPI::Serializable::Renderer.new.render(
           @course,
-          class: {
-            Course: SerializableCourse,
-            Subject: SerializableSubject,
-            PrimarySubject: SerializableSubject,
-            SecondarySubject: SerializableSubject,
-            ModernLanguagesSubject: SerializableSubject,
-            FurtherEducationSubject: SerializableSubject,
-          },
+          class: CourseSerializersService.new.execute,
           include: [:subjects],
         )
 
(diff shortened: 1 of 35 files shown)


faissalMT:
[[?] Edit Course 'Subject' (MCBE)](https://trello.com/c/1MOCToMq/1858-edit-course-subject-mcbe)

defong:
>There may be some kind of validation this PR could benefit from but I'm not sure what kind of validation might be needed.

```
Given a course
and it has a level assigned to it
When selecting the subjects
Then the subjects is of that level
```

and the rest

```
Given a primary level course
Then the course can only have one subject
And the subject must be one of the primary type subject
```

```
Given a further education level course
Then the course can only have one subject
And the subject must be a further type subject
and the subject must be `further education` subject
```

```
Given a course that is either
a primary level course or,
a further education level course
Then the course can only have one single subject
```

```
Given a secondary level course
Then the course have one or two subjects
And the subjects must be a secondary type subjects
```

```
Given a secondary level course
And the course is modern foreign language 
Then the course have one or many language subjects
And the subjects must be a language type subjects
```

defong:
Is there something we can do in the course factory to negate the need to set the `level`

ie
`
create :course, :go_figure_level, subjects [primary]
`
or
`
create :course, :with_primary_level_and subject
create :course, :with_secondary_level_and_subject
create :course, :with_secondary_level_and_subjects
create :course, :with_secondary_level_and_modern_language
create :course, :with_further_education_level_and subject
`

i thhink that will removed the need to double checking course creation and needing to know this and that to make it valid

app/controllers/api/v2/courses_controller.rb
```diff
@@ -55,7 +65,20 @@ def index
...
+          SiteStatus: SerializableSiteStatus,
+          Site: SerializableSite,
+          Subject: SerializableSubject,
+          PrimarySubject: SerializableSubject,
+          SecondarySubject: SerializableSubject,
+          ModernLanguagesSubject: SerializableSubject,
+          FurtherEducationSubject: SerializableSubject,
+          Provider: SerializableProvider,
+          ProviderEnrichment: SerializableProviderEnrichment,
+          RecruitmentCycle: SerializableRecruitmentCycle,
```
misaka: This is becoming a real bore. I wonder if we shouldn't have something that automatically registers serializers, for example:

```ruby
module API::V2
  SERIALIZERS = {}

  class SerializableCourse
    # Taking the liberty to assume we can assign to a constant Hash. Is true?
    API::V2::SERIALIZERS[Course] = self
    ...
  end
end
```

And then here:

```ruby
      def show
        render jsonapi: @course, include: params[:include], class: API::V2::SERIALIZERS
      end
```

app/controllers/api/v2/courses_controller.rb
```diff
@@ -195,7 +225,12 @@ def enrichment_params
                   :study_mode,
                   :is_send,
                   :accrediting_provider_code,
-                  :funding_type)
+                  :funding_type,
+                  :name,
+                  :course_code,
+                  :subjects_ids,
+                  :subjects_types,
+                  :level)
```
misaka: What if we had a list of these attribute names in a variable `enrichment_update_attributes`, and used that here and below in `course_params`. Likewise for `course_update_attributes`, and we wouldn't get out of sync like this.

app/controllers/api/v2/courses_controller.rb
faissalMT: Automatically registering serializers might start us on the rabbithole of introducing magic which makes it difficult for us to change things later. Additionally, a given model may have multiple serializers (e.g: ProviderSuggestion vs Provider). Where the hash of classes is repeated we can simply use a service to provide them.

app/controllers/api/v2/courses_controller.rb
faissalMT: What would we do with them? Not all of these parameters are relevant to all functions.

spec/models/course_spec.rb
```diff
@@ -126,6 +126,84 @@
...
+        expect(course.valid?).to be_truthy
+      end
+
+      it "does not validate if the subject is not of that level" do
+        course.subjects = [create(:subject, :mathematics, type: :PrimarySubject)]
+        expect(course.valid?).to be_falsey
+        expect(course.errors[:subjects]).to eq(["must be secondary"])
+      end
+
+      it "does not validate if there are more than 2 subjects" do
```
```
it "validate if there are 2 subjects of secondary" do
expect(course.valid?).to be_truthy

```

app/models/course.rb
```diff
@@ -570,6 +578,44 @@ def has_the_modern_languages_secondary_subject_type?
...
+  def validate_subject_count
+    return if subjects.empty?
+
+    case level
+    when "primary", "further_education"
+      if subjects.count > 1
+        errors.add(:subjects, "has too many subjects")
+      end
+    when "secondary"
+      if subjects.count > 2 && !has_any_modern_language_subject_type?
```
defong: is it out of scope for `modern_language`

`modern_language` + `French` + `Biology`

faissalMT: Asked @fofr

app/controllers/api/v2/courses_controller.rb
```diff
@@ -257,6 +276,10 @@ def funding_type_params
         params.fetch(:course, {})[:funding_type]
       end
 
+      def subject_ids
+        params.fetch(:course, {})[:subjects_ids]
```
defong: why is it noy using permit?

faissalMT: Because we only need one value

app/models/course.rb
faissalMT: @fofr has said that we don't need to restrict this atm

app/controllers/api/v2/courses_controller.rb
```diff
@@ -49,11 +52,12 @@ def index
...
         authorize Course
 
-        render jsonapi: @provider.courses, include: params[:include]
+        render jsonapi: @provider.courses, include: params[:include], class: CourseSerializersService.new.execute
       end
 
       def show
-        render jsonapi: @course, include: params[:include]
+        # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
+        render jsonapi: @course, include: params[:include], class: CourseSerializersService.new.execute
```
defong: @misaka anything fancy we can do on the jsonapi config

https://github.com/DFE-Digital/manage-courses-backend/blob/f4cedf7c693592147600f4131405b21be343eb9d/config/initializers/jsonapi.rb

```
         render jsonapi: @course, include: params[:include], class: CourseSerializersService.new.execute 
```

PR #811, opened 2019-09-17
[2092] Add MCB Command to mark providers as SCITTs

danielburnley:
### Context

We need to use the department's accurate list of SCITT providers (attached) to mark providers as SCITT's in publish. 

We should use the `provider type` field to update these records as SCITTS (DB value `B`). Currently no provider is marked as a SCITT using this field on publish.

- Add an MCB command to mark providers as SCITTS for the current and next recruitment cycle

### Guidance to review

- Run the MCB command `mcb providers mark_providers_as_scitts`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


lib/mcb/commands/providers/mark_providers_as_scitts.rb
```diff
@@ -0,0 +1,62 @@
...
+    "Sutton SCITT", "Teach East", "Teach Kent & Sussex", "Teach SouthEast",
+    "Thamesmead SCITT", "The Basingstoke Alliance SCITT", "The Bedfordshire Schoolsâ Training Partnership", "The Cambridge Partnership",
+    "The Deepings SCITT", "The Grand Union Training Partnership", "The Hampshire LEARN SCITT Partnership", "The Havering Teacher Training Partnership",
+    "The John Taylor SCITT", "The Learning Institute South West", "The National Mathematics and Physics SCITT", "The National Modern Languages SCITT",
+    "The OAKS (Ormiston and Keele SCITT)", "The Pimlico-London SCITT", "The Sheffield SCITT", "The Shire Foundation",
+    "The Solent SCITT", "The South Downs SCITT", "The Tommy Flowers SCITT Milton Keynes", "Three Counties Alliance SCITT",
+    "Titan Partnership Ltd", "TKAT SCITT", "Tudor Grange SCITT", "Two Mile Ash ITT Partnership",
+    "Wandsworth Primary Schoolsâ Consortium", "Wessex Schools Training Partnership", "West Midlands Consortium", "Wildern Partnership",
+    "Yorkshire and Humber Teacher Training", "Yorkshire Wolds Teacher Training"
```
misaka: Just as a general principal, once you've moved to multiline this I'd generally go with one entry per line ... may as well go all the way, this is still really hard to scan, maybe as bad or worse than all-on-one-line.

lib/mcb/commands/providers/mark_providers_as_scitts.rb
```diff
@@ -0,0 +1,62 @@
...
+  ]
+
+  puts 'Updating SCITTS for 2019'
+  MCB.get_recruitment_cycle('recruitment-year': 2019).providers.where(provider_name: scitts).each do |provider|
+    puts "Updating #{provider}"
+    provider.provider_type = "scitt"
+    provider.save!
+  end
```
misaka: Bonus points for printing out the time it took, e.g.:

```
time = Benchmark.measure do
  ...
end
puts "%0.3f seconds" % t.real
```

lib/mcb/commands/providers/mark_providers_as_scitts.rb
```diff
@@ -0,0 +1,62 @@
...
+    "The John Taylor SCITT", "The Learning Institute South West", "The National Mathematics and Physics SCITT", "The National Modern Languages SCITT",
+    "The OAKS (Ormiston and Keele SCITT)", "The Pimlico-London SCITT", "The Sheffield SCITT", "The Shire Foundation",
+    "The Solent SCITT", "The South Downs SCITT", "The Tommy Flowers SCITT Milton Keynes", "Three Counties Alliance SCITT",
+    "Titan Partnership Ltd", "TKAT SCITT", "Tudor Grange SCITT", "Two Mile Ash ITT Partnership",
+    "Wandsworth Primary Schoolsâ Consortium", "Wessex Schools Training Partnership", "West Midlands Consortium", "Wildern Partnership",
+    "Yorkshire and Humber Teacher Training", "Yorkshire Wolds Teacher Training"
+  ]
+
+  puts 'Updating SCITTS for 2019'
+  MCB.get_recruitment_cycle('recruitment-year': 2019).providers.where(provider_name: scitts).each do |provider|
```
misaka: `MCB.get_recruitment_cycle` looks to me like it's a utility method to help parse cmdline args ... so unless you're passing `opts` to it to see what recruitment cycle the user wanted (defaulting to the current one), I would've just gone with a `RecruitmentCycle.each` loop or something. :)

danielburnley: :man_facepalming: I just completely forgot about ActiveRecord when i wrote that apparently

PR #819, opened 2019-09-19
2132 refactor v1 v2 endpoints to use new dfe subjects instead

defong:
### Context
State of flux 

In relation to #800 with regards to subjects matters only.
The previous PR takes the existing knowledge by re-modeling it to remove the need to calculate a given course's `level` and the perceived old `dfe subjects` via the `ucas subjects` assigned to it.
The previously workflow evolves around correctly assigning the required `ucas subjects` in order to create a given course with the desired `level` and desired  old`dfe subjects` via the subject mapper.

The new modeling should now be in the position to address the imbalance.
The course may now be able to set its level in order to set the subjects of that level and apply the appropriative business rules.

This PR only deals with returning subjects outwardly in the correct manner, to facilitate the bear minimum, for the next recruitment cycle.

Removing direct and indirect usage of ucas subjects
Returning appropriate dfe subjects for 
- v1 endpoints
previously it used ucas subjects
- v2 endpoints
previously it used ucas subjects to calculate the old dfe subjects via the subject mapper

### Guidance to review

Items of interest denoted as `TODO` in the last 2 commits, those should be addressed separately to this PR.

subjects stats
`ucas_subject`  is over 80+
`subject mapper` 43
`new dfe subject` 42 (1 name change, 1 removed) + 3 (internal)

#### Beyond scope ####
Subject's consideration of any business logic such as

-  a given course's`level` of `primary`, `secondary`, or `further_education`
-  financial support, ie `bursary`, `early_career_payments`, or  `scholarship`

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -2,7 +2,7 @@ module API
   module V1
     class SubjectsController < API::V1::ApplicationController
       def index
-        @subjects = UCASSubject.all
+        @subjects = Subject.all
         paginate json: @subjects
       end
     end
--- a/app/controllers/api/v1/subjects_controller.rb
+++ b/app/controllers/api/v1/subjects_controller.rb
@@ -2,7 +2,9 @@ module API
   module V1
     class SubjectsController < API::V1::ApplicationController
       def index
-        @subjects = Subject.all
+        @subjects = Subject.where.not(type: "DiscontinuedSubject")
+          .where.not(subject_code: nil)
+
         paginate json: @subjects
       end
     end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -509,7 +509,10 @@ private
     if findable?.blank?
       errors.add :site_statuses, "No findable sites."
     end
-    if dfe_subjects.blank?
+    if subjects
+        .where.not(type: "DiscontinuedSubject")
+        .where.not(subject_code: nil)
+        .blank?
       errors.add :dfe_subjects, "No DfE subject."
     end
   end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -513,7 +513,7 @@ private
         .where.not(type: "DiscontinuedSubject")
         .where.not(subject_code: nil)
         .blank?
-      errors.add :dfe_subjects, "No DfE subject."
+      errors.add :subjects, "No subjects."
     end
   end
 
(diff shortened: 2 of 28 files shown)


spec/requests/api/v1/course_spec.rb
```diff
@@ -28,14 +28,14 @@ def get_course_codes_from_body(body)
     context "without changed_since parameter" do
       before do
         site = FactoryBot.create(:site, code: "-", location_name: "Main Site", provider: provider)
-        subject1 = FactoryBot.find_or_create(:ucas_subject, :secondary)
-        subject2 = FactoryBot.find_or_create(:ucas_subject, :mathematics)
+        subject1 = FactoryBot.find_or_create(:subject, :modern_languages)
+        subject2 = FactoryBot.find_or_create(:subject, :german)
 
         course = FactoryBot.create(:course,
```
defong: That is how a valid ml course should look like

spec/factories/subjects.rb
```diff
@@ -12,29 +12,64 @@
   factory :subject do
```
defong: factory stuff

spec/requests/api/v1/course_spec.rb
faissalMT: There shouldn't be any need to do `FactoryBot.`

defong:
[Refactor v1/v2 endpoints to use new dfe subjects instead](https://trello.com/c/Js1nIA1U/2132-refactor-v1-v2-endpoints-to-use-new-dfe-subjects-instead)

app/controllers/api/v1/subjects_controller.rb
```diff
@@ -2,7 +2,9 @@ module API
   module V1
     class SubjectsController < API::V1::ApplicationController
       def index
-        @subjects = UCASSubject.all
+        @subjects = Subject.where.not(type: "DiscontinuedSubject")
+          .where.not(subject_code: nil)
```
misaka: I think we need a scope for these to `where` clauses, they're used below a whole lot.

app/models/course.rb
```diff
@@ -509,7 +509,10 @@ def validate_course_syncable
     if findable?.blank?
       errors.add :site_statuses, "No findable sites."
     end
-    if dfe_subjects.blank?
+    if subjects
+        .where.not(type: "DiscontinuedSubject")
+        .where.not(subject_code: nil)
+        .blank?
       errors.add :dfe_subjects, "No DfE subject."
```
misaka: I think you missed one.

PR #824, opened 2019-09-20
[2186] Update mcb to use new subjects

faissalMT:
### Context
MCB should be able to use the new subjects rather than the UCAS subjects.  

Use the DfE subjects instead of the UCAS subjects.  

### Guidance to review
I've added `CoursePotentialSubjectService` because this is logic that will likely be used elsewhere but that I don't think belongs in the model since it doesn't pertain directly to courses themselves but rather is an implication of its properties. I am open to feedback, I've tried to maintain a reasonable degree of coupling between it and Rails.

### Checklist

- [X] Attach to Trello card
- [X] Rebased ~~master~~ parent
- [X] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -166,6 +166,7 @@ class Course < ApplicationRecord
   validate :validate_qualification, on: :update
   validate :validate_start_date, on: :update, if: -> { provider.present? }
   validate :validate_applications_open_from, on: :update, if: -> { provider.present? }
+  validate :validate_subjects
 
   after_validation :remove_unnecessary_enrichments_validation_message
 
@@ -534,6 +535,20 @@ private
       errors.add(:applications_open_from, "#{applications_open_from} is not valid for the #{provider.recruitment_cycle.year} cycle. " +
       "A valid date must be between #{recruitment_cycle.application_start_date} and #{recruitment_cycle.application_end_date}")
     end
+  end
+
+  def validate_subjects
+    if has_any_modern_language_subject_type? & !has_the_modern_languages_secondary_subject_type?
+      subjects << SecondarySubject.modern_languages
+    end
+  end
+
+  def has_any_modern_language_subject_type?
+    subjects.any? { |subject| subject.type == "ModernLanguagesSubject" }
+  end
+
+  def has_the_modern_languages_secondary_subject_type?
+    subjects.any? { |subject| subject == SecondarySubject.modern_languages }
   end
 
   def valid_date_range
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -15,4 +15,8 @@ class Subject < ApplicationRecord
   def to_sym
     subject_name.parameterize.underscore.to_sym
   end
+
+  def to_s
+    "#{subject_name} (#{subject_code})"
+  end
 end
--- a/spec/lib/mcb/editor/courses_editor_spec.rb
+++ b/spec/lib/mcb/editor/courses_editor_spec.rb
@@ -13,12 +13,16 @@ describe MCB::Editor::CoursesEditor, :needs_audit_user do
   let(:email) { "user@education.gov.uk" }
   let(:provider) { create(:provider, provider_code: provider_code) }
   let(:accredited_body) { create(:provider, :accredited_body) }
-  let!(:mathematics) { find_or_create(:ucas_subject, :mathematics) }
-  let!(:biology) { find_or_create(:ucas_subject, subject_name: "Biology") }
-  let!(:secondary) { find_or_create(:ucas_subject, :secondary) }
+  let!(:japanese) { find_or_create(:subject, :japanese) }
+  let!(:primary_with_mathematics) { find_or_create(:subject, :primary_with_mathematics) }
+  let!(:biology) { find_or_create(:subject, :biology) }
+  let!(:modern_languages) { find_or_create(:subject, :modern_languages) }
+  let!(:further_education) { find_or_create(:subject, :further_education) }
   let(:current_cycle) { RecruitmentCycle.current_recruitment_cycle }
   let!(:next_cycle) { find_or_create(:recruitment_cycle, year: "2020") }
   let(:is_send) { false }
+  let(:subjects) { [] }
+  let(:level) { "primary" }
   let!(:course) {
     create(:course,
            provider: provider,
@@ -32,8 +36,9 @@ describe MCB::Editor::CoursesEditor, :needs_audit_user do
            qualification: "qts",
            study_mode: "part_time",
            start_date: Date.new(2019, 8, 1),
-           age_range: "secondary",
-           ucas_subjects: [secondary, biology],
+           age_range: "primary",
+           level: level,
+           subjects: subjects,
            applications_open_from: Date.new(2018, 10, 9),
            is_send: is_send)
   }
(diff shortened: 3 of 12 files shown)


faissalMT:
[Update MCB to use new subjects](https://trello.com/c/p9hCjfC8/2186-update-mcb-to-use-new-subjects)

app/models/subject.rb
```diff
@@ -15,4 +15,8 @@ class Subject < ApplicationRecord
   def to_sym
     subject_name.parameterize.underscore.to_sym
   end
+
+  def to_s
+    subject_name
```
misaka: Can we add the code in here like we do with courses and providers?

app/models/ucas_subject.rb
```diff
@@ -2,10 +2,9 @@
 #
 # Table name: ucas_subject
 #
-#  id           :bigint           not null, primary key
-#  type         :text
-#  subject_code :text
+#  id           :integer          not null, primary key
```
misaka: Is this change caused by a Postgres version mismatch? If it's likely to switch back we should probably avoid including it in this PR to cut down on noise.

misaka: Also, why is type being taken out?

spec/lib/mcb/editor/courses_editor_spec.rb
```diff
@@ -496,13 +604,13 @@ def run_new_course_wizard(*input_cmds)
...
+          desired_attributes[:level],
           desired_attributes[:course_code],
           'y', # is SEND confirmation
-          desired_attributes[:recruitment_cycle],
           "y", # confirm creation
           # subject selection
-          "Biology",
-          "Secondary",
+          "2",
+          "4",
```
misaka: Do these numbers map to the subjects "Biology" and "Secondary"? Does that make this test order dependent? Is it worth doing that? Having the names of the subjects seemed nice and readable.

app/services/course_potential_subject_service.rb
```diff
@@ -0,0 +1,24 @@
...
+    @modern_language_subject = modern_language_subject
+    @further_education_subject = further_education_subject
+  end
+
+  def execute(course)
+    case course.level
+    when 'primary'
+      @primary_subject.all
+    when 'secondary'
+      @secondary_subject.all + @modern_language_subject.all
```
defong: i probably break the divide here
if course.subjects has modern language
ModernLanguageSubject .all
else
 @secondary_subject.all

spec/lib/mcb/editor/courses_editor_spec.rb
```diff
@@ -13,12 +13,15 @@ def run_editor(*input_cmds)
...
-  let!(:biology) { find_or_create(:ucas_subject, subject_name: "Biology") }
-  let!(:secondary) { find_or_create(:ucas_subject, :secondary) }
+  let!(:japanese) { find_or_create(:subject, subject_name: 'Japanese', type: :ModernLanguagesSubject) }
+  let!(:mathematics) { find_or_create(:subject, :mathematics, type: :PrimarySubject) }
+  let!(:biology) { find_or_create(:subject, subject_name: 'Biology', type: :SecondarySubject) }
+  let!(:english) { find_or_create(:subject, :english, type: :FurtherEducationSubject) }
   let(:current_cycle) { RecruitmentCycle.current_recruitment_cycle }
   let!(:next_cycle) { find_or_create(:recruitment_cycle, year: '2020') }
   let(:is_send) { false }
+  let(:subjects) { [english, biology] }
```
defong: can we refrain from having tests that makes a course that don't fit in the rules, unless its to test the rules

spec/lib/mcb/editor/courses_editor_spec.rb
```diff
@@ -426,13 +439,107 @@ def run_new_course_wizard(*input_cmds)
...
+          level: 'secondary',
           course_code: '1X2B',
           recruitment_cycle: '2', # the 2nd option should always be the current recruitment cycle
           application_opening_date: "18 October 2018",
           is_send: true
         }
       }
 
+      describe 'selects the correctly leveled subject' do
+        it 'only shows secondary & language subjects if secondary level is selected' do
```
defong: can we double check that this is not the case that is possible in real life

ie
`English` as `SecondarySubject` + `English as a secondary language`
For a `English course`

as far as I am aware 

- 1 or 2 `SecondarySubject`
- 1 `SecondarySubject` with a subject name of `Modern Language` + 1 or more of a `Named Language` from `ModernLanguage`

app/models/ucas_subject.rb
faissalMT: I'm guessing type is being taken out because ucas subjects don't have them, dfe subjects do. Not sure what's going on with the primary key.

app/models/course.rb
```diff
@@ -533,6 +534,20 @@ def validate_applications_open_from
     end
   end
 
+  def validate_subjects
+    if has_a_modern_language? & !has_the_modern_language_subject?
+      subjects << Subject.find_by(subject_name: 'Modern Languages')
```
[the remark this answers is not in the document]
faissalMT: Good idea

spec/services/course_assignable_subject_service_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+  let(:primary_model) { spy }
+  let(:secondary_model) { spy }
+  let(:further_education_model) { spy }
+  let(:modern_language_model) { spy }
+  let(:modern_languages_parent_subject) { spy }
+
+  it "calls all on primary if level is primary" do
+    course = create(:course, level: "primary")
+    service.execute(course)
```
misaka: This test, and the ones below, don't check what gets returned. Would that be useful?

spec/lib/mcb/editor/courses_editor_spec.rb
```diff
@@ -13,12 +13,16 @@ def run_editor(*input_cmds)
   let(:email) { "user@education.gov.uk" }
   let(:provider) { create(:provider, provider_code: provider_code) }
   let(:accredited_body) { create(:provider, :accredited_body) }
-  let!(:mathematics) { find_or_create(:ucas_subject, :mathematics) }
-  let!(:biology) { find_or_create(:ucas_subject, subject_name: "Biology") }
-  let!(:secondary) { find_or_create(:ucas_subject, :secondary) }
+  let!(:japanese) { find_or_create(:subject, subject_name: "Japanese", type: :ModernLanguagesSubject) }
```
misaka: How about traits for all these different types of subjects?

spec/services/course_assignable_subject_service_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    let!(:biology) { create(:subject, subject_name: "Biology", type: :SecondarySubject).becomes(SecondarySubject) }
+    let!(:arabic) { create(:subject, subject_name: "Arabic", type: :ModernLanguagesSubject).becomes(ModernLanguagesSubject) }
+
+    it "returns subjects other than modern language" do
+      course = create(:course, level: "secondary")
+      expect(service.execute(course)).to match_array([biology, arabic])
+    end
+  end
+
+  it "calls all on secondary if level is secondary" do
```
defong: bad name?

faissalMT: Understatement

spec/services/course_assignable_subject_service_spec.rb
faissalMT: Yes, they ensure that we're getting data in the correct way, although I should definitely be testing the return values.

PR #827, opened 2019-09-20
[2200] Upgrade to Rails 6

tvararu:
### Context

We're upgrading to Rails 6 to keep up to date and access the latest features.

This was done by running the following commands:

```bash
bundle update rails
rails app:update
```

I was careful not to override any of our customisations, then I ran the test suite. It was green, so I fixed the deprecation warnings.

### Guidance to review

:heart: :ship:

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu:
[Upgrade to Rails 6](https://trello.com/c/OLbSFlFN/2199-upgrade-to-rails-6)

config/environments/production.rb
```diff
@@ -41,17 +40,61 @@
...
+  # config.action_mailer.raise_delivery_errors = false
+
   # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
   # the I18n.default_locale when a translation cannot be found).
   config.i18n.fallbacks = true
 
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
 
+  # Use default logging formatter so that PID and timestamp are not suppressed.
```
dankmitchell: How do these changes effect those already in this file i.e.https://github.com/DFE-Digital/manage-courses-backend/blob/master/config/environments/production.rb#L35

tvararu: Not sure so I'll just remove this bit 👍

app/models/course_enrichment.rb
```diff
@@ -38,7 +38,7 @@ class CourseEnrichment < ApplicationRecord
   belongs_to :course
 
   scope :latest_first, -> { order(created_at: :desc, id: :desc) }
-  scope :draft, -> { where(status: 'draft').or(CourseEnrichment.rolled_over) }
+  scope :draft, -> { where(status: 'draft').or(CourseEnrichment.unscoped.rolled_over) }
```
misaka: Why this? If we implement course enrichment this will pull them in too, which might be a tricky bug to track down.

tvararu: This is due to https://github.com/rails/rails/pull/35280/files#diff-2ea4047e83f40a075015865a491c82e7R31

My understanding is: Rails 6.0 and under treat the original line as:

```ruby
scope :draft, -> { where(status: 'draft').or(CourseEnrichment.draft.rolled_over) }
                                                              ^^^^^
```

Rails 6.1+ will treat it as:

```ruby
scope :draft, -> { where(status: 'draft').or(CourseEnrichment.unscoped.rolled_over) }
                                                              ^^^^^^^^
```

With that in mind, I think we should actually be adding the `draft` scope in explicitly (instead of `unscoped`) to maintain current behaviour. So I'll update this.

tvararu: Nevermind, that brings in the Rails deprecation warning:

```
DEPRECATION WARNING: Class level methods will no longer inherit scoping from `draft` in Rails 6.1. To continue using the scoped relation, pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `CourseEnrichment.unscoped`. (called from block in <class:CourseEnrichment> at /Users/tvararu/dfe/manage-courses-backend/app/models/course_enrichment.rb:41)
```

It's doing my head in what to do here but all the tests pass with `.unscoped` ?_?

PR #829, opened 2019-09-23
[2207] don't retry when we get 502 from sac api for bulk sync

misaka:
### Context

We get an error from sacui fairly regularly with this process, and as the bulk sync is a pretty heavy job we don't want a bunch of these retrying in the background. And the job just reruns every 2 hours anyway. Until we sort how not to get errors every time it kicks off we should drop failed job runs.

Disable Sidekiq retries for the bulk sync job.

### Guidance to review

:shipit: . o ( This is blocking implementing bulk sync in prod. )

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/jobs/bulk_sync_courses_to_find_job.rb
+++ b/app/jobs/bulk_sync_courses_to_find_job.rb
@@ -1,6 +1,9 @@
 class BulkSyncCoursesToFindJob < ApplicationJob
   queue_as :find_sync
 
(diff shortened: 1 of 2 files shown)


misaka:
[Disable retries on Bulk Sync job](https://trello.com/c/TGvREtFu/2207-disable-retries-on-bulk-sync-job)

PR #853, opened 2019-09-30
[2250] Add validation for subject and level presence

faissalMT:
### Context
There've been some issues with subjects and levels being missing from courses since moving away from UCAS subjects, this should help us better diagnose these issues.

Give a validation message when either subject or level are not present.

### Guidance to review
[This test](https://github.com/DFE-Digital/manage-courses-backend/pull/853/files#diff-2370f7c60a21a83d5258e3f9f253bbb7L119) was deleted because this validation prevents it ever reaching a state where a course cannot be bulk published because it has no subjects.  
I've also refactored the request validation tests a bit while correcting them so that they aren't as fragile.

### Checklist

- [x] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -33,6 +33,10 @@ en:
               blank: "^Complete your course information before publishing"
             sites:
               blank: "^You must pick at least one location for this course"
+            level:
+              blank: "^There is a problem with this course. Contact support to fix it (Error: L)"
+            subjects:
+              blank: "^There is a problem with this course. Contact support to fix it (Error: S)"
         course_enrichment:
           attributes:
             salary_details:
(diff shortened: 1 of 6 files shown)


faissalMT:
[Add validation for level and subjects](https://trello.com/c/lg7GXQ6j/2250-add-validation-for-level-and-subjects)

config/locales/en.yml
```diff
@@ -33,6 +33,10 @@ en:
               blank: "^Complete your course information before publishing"
             sites:
               blank: "^You must pick at least one location for this course"
+            level:
+              blank: "^This course is missing levels, please contact support"
```
dankmitchell:
```suggestion
              blank: "^There is a problem with this course. Contact support to fix it (Error: L)"
```

config/locales/en.yml
```diff
@@ -33,6 +33,10 @@ en:
...
+            subjects:
+              blank: "^This course is missing subjects, please contact support"
```
dankmitchell:
```suggestion
              blank: "^There is a problem with this course. Contact support to fix it (Error: S)"
```

PR #859, opened 2019-10-01
[1809] Add withdraw course functionality

davidgisbey:
### Context
Users need to be able to withdraw a published course. Once published a course should not be able to be deleted/discarded

- Add an endpoint for withdrawing a course

### Guidance to review
- A second PR to add a command to mcb to do this will follow.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -103,6 +103,16 @@ module API
         @course.discard
       end
 
+      def withdraw
+        authorize @course
+        if @course.last_published_at.present?
+          #perhaps this should just be @course.discard . Not 100% sure
+          raise RuntimeError.new("This course has not been published and should be deleted not withdrawn")
+        else
+          @course.withdraw
+        end
+      end
+
       def create
         authorize @provider, :can_create_course?
         return unless course_params.values.any?
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -105,8 +105,7 @@ module API
 
       def withdraw
         authorize @course
-        if @course.last_published_at.present?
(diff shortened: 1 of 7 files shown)


davidgisbey:
[Withdraw Course](https://trello.com/c/pGjgZ0KR/1809-withdraw-course)

app/models/course.rb
```diff
@@ -443,6 +443,16 @@ def funding_type=(funding_type)
     end
   end
 
+  def withdraw
+    if last_published_at.blank?
```
fofr: Should this use `is_published`?

fofr:
Do we need `:withdrawn` in here somewhere:
https://github.com/DFE-Digital/manage-courses-backend/blob/master/app/models/course.rb#L285-L303

app/controllers/api/v2/courses_controller.rb
```diff
@@ -103,6 +103,16 @@ def destroy
         @course.discard
       end
 
+      def withdraw
+        authorize @course
+        if @course.last_published_at.present?
```
dankmitchell:
```suggestion
        if @course.is_published?
```

app/models/course.rb
dankmitchell:
```suggestion
    if is_published?
```

davidgisbey: Done 👍

app/controllers/api/v2/courses_controller.rb
davidgisbey: done

app/models/course.rb
davidgisbey: done

fofr:
After a course is withdrawn, the `is_published` method on Course will give a false positive.

PR #860, opened 2019-10-01
[2253] Add unique constraint to course_subject

timabell:
### Context

Dupe records broke ruby bulk publisher.

* Make that impossible, push error earlier.
* Fix all the tests, which involved making the validation not do magic to languages

### Guidance to review

:ship:

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -554,7 +554,7 @@ private
   end
 
   def has_the_modern_languages_secondary_subject_type?
-    subjects.any? { |subject| subject == SecondarySubject.modern_languages }
+    subjects.any? { |subject| subject&.id == SecondarySubject.modern_languages.id }
   end
 
   def valid_date_range
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -443,6 +443,12 @@ class Course < ApplicationRecord
     end
   end
 
+  def fixup_languages
+    if has_any_modern_language_subject_type? && !has_the_modern_languages_secondary_subject_type?
+      subjects << SecondarySubject.modern_languages
+    end
+  end
+
 private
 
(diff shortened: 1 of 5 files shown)


app/models/course.rb
```diff
@@ -446,6 +446,12 @@ def funding_type=(funding_type)
     end
   end
 
+  def fixup_languages
```
dankmitchell: Is this a one-time method?

timabell: It will be used whenever `mcb courses create` is used to add a language, and in future could be used in the self-service new course flow

faissalMT: Maybe a more descriptive name than `fixup_languages` such as `imply_modern_language` or something?

timabell: `ensure_modern_languages`
?

maybe we should override assignment instead

timabell:
As @faissalMT mentioned IRL this could trip people up in future if they don't know to call the fixup method. Although I think validations is the wrong place for mutation, perhaps overriding the `=`, `<<` operators might be better?

I'd like to get something that

* just works
* obeys rule of least surprise
* is SOLID
* is well tested

timabell:
conflict on schema file fixed; rebased.

timabell:
Do we want to merge this and then iterate? This PR prevents a potential breakage of production bulk publish that's still hanging over us.

app/models/course.rb
dankmitchell: Thanks @faissalMT

timabell:
@dankmitchell renamed, waddyathink? Does it make enough sense to you from cold?

PR #862, opened 2019-10-01
[2267] Fix course count

dankmitchell:
### Context
Course count

- Don't include deleted (discarded) courses in the count

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -162,6 +162,7 @@ class Provider < ApplicationRecord
         LEFT OUTER JOIN (
           SELECT b.provider_id, COUNT(*) courses_count
           FROM course b
+          WHERE b.discarded_at IS NULL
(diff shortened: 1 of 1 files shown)


dankmitchell:
[Course counts on organisation list show the wrong number of courses](https://trello.com/c/9EcGLugq/2267-course-counts-on-organisation-list-show-the-wrong-number-of-courses)

fofr:
Does the count also take into account recruitment cycles? Does it show only the current cycle's total courses?

misaka:
> Does the count also take into account recruitment cycles? Does it show only the current cycle's total courses?

Ya, it works off of `provider_id`, which is unique for a recruitment cycle.

PR #864, opened 2019-10-02
[2258] Only sync published course on course update

fofr:
### Context

When editing these locations for a provider:
https://www.publish-teacher-training-courses.service.gov.uk/organisations/1ZC/2020/courses/38BW/details

After removing a location (Inspiration Trust) on an draft course, it attempted to sync which then errored: https://sentry.io/organizations/dfe-bat/issues/1250607692/

The change was made successfully but the user sees a 500 page.

If a course hasn't been published yet, it's not on Find, and we don't need to sync it.

This wasn't true last cycle, in 2019 a course was on Find if it had been set to running on UCAS – a course's published content was published separately.

From 2020 the only way a course gets set to running is by publishing it. So we can now include this check.

This will prevent errors when trying to sync something in a non-syncable state (eg no findable sites)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -89,7 +89,7 @@ module API
         update_enrichment
         update_sites
         update_site_status_vac_statuses if @course.study_mode_previously_changed?
-        should_sync = site_ids.present? && @course.recruitment_cycle.current? && @course.is_published?
+        should_sync = site_ids.present? && @course.should_sync?
         has_synced? if should_sync
 
         if @course.errors.empty? && @course.valid?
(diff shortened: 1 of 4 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -89,7 +89,7 @@ def update
         update_enrichment
         update_sites
         update_site_status_vac_statuses if @course.study_mode_previously_changed?
-        should_sync = site_ids.present? && @course.recruitment_cycle.current?
+        should_sync = site_ids.present? && @course.recruitment_cycle.current? && @course.is_published?
```
benilovj: should this logic be on the course model? e.g.

```
def syncable?
  recruitment_cycle.current? && is_published?
end
```

fofr:
[Error when removing or adding sites from an unpublished course.](https://trello.com/c/ykxglbsh/2258-error-when-removing-or-adding-sites-from-an-unpublished-course)

PR #867, opened 2019-10-02
Implement public V3 courses API

tvararu:
### Context

Apply (and search-ui soon) will need to query for public course information. One proposal is to have public endpoints living under the `v3` namespace.

Add v3 courses controller and route.

Allows us to run:

```bash
# Look ma, no auth!
curl "http://localhost:3001/api/v3/recruitment_cycles/2020/providers/2CG/courses/2Z3H" | ppjson
{
    "data": {
        "attributes": {
            "about_accrediting_body": null,
            "about_course": null,
            "accrediting_provider": {
                "accrediting_provider": "accredited_body",
...
```

And again, with sparse fields (`?fields[courses]=provider_code,course_code,name`):

```bash
curl -v "http://localhost:3001/api/v3/recruitment_cycles/2019/providers/2KN/courses/39VZ?fields\[courses\]=provider_code,course_code,name" | ppjson
{
    "data": {
        "attributes": {
            "course_code": "39VZ",
            "name": "Mathematics",
            "provider_code": "2KN"
        },
```

### Guidance to review

It's hidden behind a feature flag which is set to `false` in production, because Apply isn't running in prod yet and this allows us to kick the tires on it for a bit before. QA and staging have anonymised information.

TODO:

- [x] Write tests
- [x] Do we need any filtering for published courses, e.g. can this leak draft info if we're not careful?
- [x] ~Is everything in the course serializer fine to expose publicly, or do we need to strip some attributes out? I gave it a look and nothing jumped out at me.~ Will use a feature flag initially
- [x] Should I refactor `build_provider` + `build_course` + `build_recruitment_cycle` at a higher level e.g. `ApplicationController` since they're duplicated everywhere? May as well.
- [x] Add ability to filter for only the attributes we care about via query string params, as the response is quite long (bytes are important!)

### Checklist

- [x] ~Make sure all information from the Trello card is in here~
- [x] ~Attach to Trello card~ https://trello.com/c/bQzgIERR/115-implement-public-courses-endpoint-on-find
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/courses_controller.rb
+++ b/app/controllers/api/v3/courses_controller.rb
@@ -1,0 +1,24 @@
+module API
+  module V3
+    class CoursesController < API::V3::ApplicationController
+      before_action :build_recruitment_cycle
+      before_action :build_provider
+      before_action :build_course
+
+      def show
+        if @course.is_published?
+          render jsonapi: @course, fields: fields_param
+        else
+          raise ActiveRecord::RecordNotFound
+        end
+      end
+
+      def fields_param
+        params.fetch(:fields, {})
+          .permit(:courses)
+          .to_h
+          .map { |k, v| [k, v.split(",").map(&:to_sym)] }
+      end
+    end
+  end
+end
--- a/spec/requests/api/v3/providers/courses/show_spec.rb
+++ b/spec/requests/api/v3/providers/courses/show_spec.rb
@@ -1,0 +1,69 @@
+require "rails_helper"
+
+describe "GET v3/providers/:provider_code/courses/:course_code" do
+  let(:recruitment_cycle) { find_or_create :recruitment_cycle }
+  let(:provider) { create :provider, recruitment_cycle: recruitment_cycle }
+  let(:jsonapi_course) {
+    JSON.parse(
+      JSONAPI::Serializable::Renderer.new.render(
+        course,
+        class: {
+          Course: API::V2::SerializableCourse,
+        },
+      ).to_json,
+    )
+  }
+  let(:jsonapi_response) { JSON.parse(response.body) }
+  let(:route) {
+    "/api/v3/recruitment_cycles/#{recruitment_cycle.year}" \
+    "/providers/#{provider.provider_code}" \
+    "/courses/#{course.course_code}"
+  }
+  let(:course) { create :course, provider: provider, enrichments: enrichments }
+
+  context "with a published course" do
+    let(:enrichments) { [build(:course_enrichment, :published)] }
+
+    it "returns full course information" do
+      get route
+
+      expect(jsonapi_response["data"]).to eq jsonapi_course["data"]
+    end
+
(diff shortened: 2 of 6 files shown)


app/controllers/api/v3/courses_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+      def build_provider
+        @provider = @recruitment_cycle.providers.find_by!(
+          provider_code: params[:provider_code].upcase,
+        )
+      end
+
+      def build_course
+        @course = @provider.courses.find_by!(course_code: params[:code].upcase)
```
defong: Don't think this return only published course

timabell:
@tvararu  What will happen to "apply" if manage-courses-backend goes down / becomes unresponsive etc?

tvararu:
@timabell answered offline, in a nutshell: we're investigating using layers of caching + progressive enhancement to mitigate.

app/controllers/api/v3/courses_controller.rb
```diff
@@ -0,0 +1,33 @@
+module API
+  module V3
+    class CoursesController < API::V3::ApplicationController
+      before_action :build_recruitment_cycle
+      before_action :build_provider
+      before_action :build_course
+
+      def show
+        if @course.is_published?
```
tvararu: @defong we could check if the course is published like this?

spec/requests/api/v3/providers/courses/show_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(jsonapi_response["data"]).to eq jsonapi_course["data"]
+    end
+
+    it "returns sparse course information" do
+      requested_fields = %w[course_code name provider_code].sort
+      get route + "?fields[courses]=#{requested_fields.join(',')}"
+
+      expect(jsonapi_response["data"]["attributes"].keys).to eq requested_fields
+    end
+  end
```
timabell: maybe add a test:
- for a draft course enrichment
- the draft data is not returned

timabell:
Probably best to not merge this till the dust has settled on find 2020 going live.

spec/requests/api/v3/providers/courses/show_spec.rb
tvararu: Done

misaka:
> @tvararu What will happen to "apply" if manage-courses-backend goes down / becomes unresponsive etc?

Not to discourage trying to mitigate for this eventuality, but I think we should consider mcbe to be fairly critical and have some SLAs around it's uptime. I expect that we'll start to work towards moving Find to using it pretty soon, for example.

timabell:
I think we should probably stop confusing version with purpose.

* v1 = for ucas
* v2 = for frontend
* v3 = apply
* all = confusing

I think this will bite us when we actually want to bump the version. This might be the moment to re-consider our url design.

PR #874, opened 2019-10-04
[1809] Withdraw course updates

danielburnley:
### Context

The withdraw course check had it's if condition backwards, this needed to be flipped.

Additionally, we needed to update the `content_status` to `withdrawn` so the frontend could respond appropriately

- Flip the check
- Edit the content status

- [x] Attach to Trello card
- [x]  Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/course/withdraw_spec.rb
+++ b/spec/models/course/withdraw_spec.rb
@@ -12,11 +12,19 @@ describe Course, type: :model do
       course.withdraw
     end
 
-    context "an published course" do
+    context "a published course" do
       let(:enrichment) { build(:course_enrichment, :published) }
 
       it "should not be findable" do
-        expect(course.findable?).to be_falsey
+        expect(course.findable?).to eq(false)
+      end
+
+      it "should not be published" do
+        expect(course.is_published?).to eq(false)
+      end
+
+      it "should not be syncable" do
+        expect(course.syncable?).to eq(false)
       end
 
       it "should have updated the courses site statuses to be suspended and have no vacancies" do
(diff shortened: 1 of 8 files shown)


spec/models/course/withdraw_spec.rb
```diff
@@ -12,7 +12,9 @@
       course.withdraw
     end
 
-    context "an unpublished course" do
+    context "an published course" do
```
fofr: `an published`

fofr:
Can you add a test for `course.is_published?` for the withdrawn content_status.

Is a withdrawn course published?
 - We still need to sync it so we update the sites on Find correctly
 - We probably still want to disable editing of some fields

Does that mean `is_published?` should be renamed?

Would that method be better updated to check for a last published date? 🤔

spec/models/course/withdraw_spec.rb
```diff
@@ -12,9 +12,19 @@
...
+    context "a published course" do
+      let(:enrichment) { build(:course_enrichment, :published) }
+
       it "should not be findable" do
-        expect(course.findable?).to be_falsey
+        expect(course.findable?).to eq(false)
+      end
+
+      it "should not be published" do
+        expect(course.is_published?).to eq(false)
```
fofr: This test says:
"a published course … should not be published"

danielburnley:
When a course is withdrawn it'll be deleted in the next bulk sync - as it's no longer marked as `published` or `syncable` :tada:

PeteWilliams:
[Update prototype to reflect current live journey rejection reasons journeys](https://trello.com/c/1PEHPJlH/2257-update-prototype-to-reflect-current-live-journey-rejection-reasons-journeys)

PR #883, opened 2019-10-07
[2299] Set secondary science requirements to not_set

fofr:
### Context

Similar to https://github.com/DFE-Digital/manage-courses-backend/pull/640

There are no minimum GCSE requirements for science in secondary courses. Most are set to nil, set everything else to nil too (ie not set).

We've already removed the ability to edit these codes in frontend.

Set any non nil secondary course science requirement to nil (eg :not_set – but nil enums behave weirdly, so using nil explicitly)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/db/migrate/20191007103707_fix_secondary_science_course_requirements.rb
+++ b/db/migrate/20191007103707_fix_secondary_science_course_requirements.rb
@@ -2,9 +2,7 @@ class FixSecondaryScienceCourseRequirements < ActiveRecord::Migration[6.0]
   def up
     # There is no requirement for Science for secondary courses
     say_with_time "setting secondary science course requirements to not_set" do
-      Course.select { |course| course.level == "secondary" && !course.science.nil? }.each do |course|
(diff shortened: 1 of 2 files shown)


fofr:
[Fix science requirements for secondary courses](https://trello.com/c/wp3vVlSF/2299-fix-science-requirements-for-secondary-courses)

fofr:
We wanted to do this back in June, but didn’t prioritise the need to contact providers mid-cycle about this fix. Now we are in a new cycle and these settings aren’t displayed in frontend we can make this change.

fofr:
Old card:
 [Add validations to GCSE requirements](https://trello.com/c/OAmzR18z/1804-add-validations-to-gcse-requirements)

fofr:
`Course.update` is failing to update 42 of these courses locally (with recent data dump). Not sure why.

PR #884, opened 2019-10-07
[2295] Disable v1 API for 2019 courses

tvararu:
### Context

We want to ensure that UCAS can't access stale date, so it makes sense to make the old data 404.

Adds a `constraints` option to the route that I've stolen from Frontend, plus checks the query string param.

### Guidance to review

UCAS can request by either specifying the param in the URL or as a query string parameter.

Have tested this locally and it seems to work fine, plus we have good test coverage.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -2,10 +2,25 @@ module API
   module V1
     class ApplicationController < ::ApplicationController
       before_action -> { skip_authorization }
+      before_action :check_recruitment_cycle_is_current_or_next_year
 
       def authenticate
         authenticate_or_request_with_http_token do |token|
           ActiveSupport::SecurityUtils.secure_compare(token, Rails.application.config.authentication_token)
+        end
+      end
+
+      def check_recruitment_cycle_is_current_or_next_year
+        current_year = Settings.current_recruitment_cycle_year
+        next_year = current_year + 1
+
+        requested_year = params[:recruitment_year]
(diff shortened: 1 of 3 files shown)


tvararu:
[deny access to v1 2019 endpoints](https://trello.com/c/EWOyxBBA/2295-deny-access-to-v1-2019-endpoints)

app/controllers/api/v1/application_controller.rb
```diff
@@ -2,12 +2,27 @@ module API
...
           ActiveSupport::SecurityUtils.secure_compare(token, Rails.application.config.authentication_token)
         end
       end
+
+      def check_recruitment_cycle_is_current_year
+        current_year = Settings.current_recruitment_cycle_year
+        next_year = current_year + 1
+
+        requested_year = params[:recruitment_year]
+        is_next_or_current_year = [current_year, next_year]
```
defong:
``` ruby
RecruitmentCycle.current_recruitment_cycle
    RecruitmentCycle.next_recruitment_cycle
```

don't you wanna use one of these?

tvararu: Yes

tvararu: Wow! That's a rabbit hole. It breaks a bunch of different tests because `RecruitmentCycle.next_recruitment_cycle` is nil. Not sure if we should do this right now.

PR #889, opened 2019-10-07
[2305] Dont call update when updating funding type

danielburnley:
### Context

When creating courses - the `funding_type` attribute gets set. This can cause 500s due to an override on `funding_type=` calling update, triggering validations which attempt to validate nil data

Set the value rather than calling update when setting funding type

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -433,20 +433,20 @@ class Course < ApplicationRecord
     case funding_type
     when "salary"
       if !self_accredited?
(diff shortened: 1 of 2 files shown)


danielburnley:
https://trello.com/c/I4jCuUjy/2305-fix-fundingtype-causing-500-errors-during-course-creation

PR #890, opened 2019-10-07
[2240] Add discard to provider and add mcb command

davidgisbey:
### Context

Providers need to be able to be soft deleted.

- Add discard to the provider model
- Add mcb command to discard providers

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/commands/providers/discard.rb
+++ b/lib/mcb/commands/providers/discard.rb
@@ -1,0 +1,14 @@
+summary "Discards a provider"
+usage "discard provider <provider code>"
+param :provider_code, transform: ->(code) { code.upcase }
+
+run do |opts, args, _cmd|
+  MCB.init_rails(opts)
+  current_cycle = RecruitmentCycle.current
+  provider = if opts&.first&.second == current_cycle.year || opts.blank?
+               current_cycle.providers.find_by!(provider_code: args[:provider_code])
+             else
+               RecruitmentCycle.find_by!(year: opts.first.second).providers.find_by!(provider_code: args[:provider_code])
+             end
+  provider.discard
+end
--- a/lib/mcb/commands/providers/discard.rb
+++ b/lib/mcb/commands/providers/discard.rb
@@ -1,14 +1,9 @@
 summary "Discards a provider"
 usage "discard <provider code>"
-param :provider_code, transform: ->(code) { code.upcase }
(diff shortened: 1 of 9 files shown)


davidgisbey:
[Remove ineligible providers](https://trello.com/c/L7whHkua/2240-remove-ineligible-providers)

lib/mcb/commands/providers/discard.rb
```diff
@@ -0,0 +1,14 @@
+summary "Discards a provider"
+usage "discard provider <provider code>"
+param :provider_code, transform: ->(code) { code.upcase }
+
+run do |opts, args, _cmd|
+  MCB.init_rails(opts)
+  current_cycle = RecruitmentCycle.current
+  provider = if opts&.first&.second == current_cycle.year || opts.blank?
```
timabell: This doesn't make sense to me

timabell:
@dankmitchell I think this is fine, if they are discard then they won't appear in v1 any more. If they aren't eligible then that's fine.

lib/mcb/commands/providers/discard.rb
davidgisbey: That's where the recruitment_cycle.year argument turns up from here  
``execute_discard_provider(arguments: [provider.provider_code, "-r", recruitment_cycle.year])``
There's probably a better way to extract it, but it was really late and i was tired haha

spec/factories/providers.rb
```diff
@@ -73,5 +73,9 @@
     trait :next_recruitment_cycle do
       recruitment_cycle { find_or_create :recruitment_cycle, :next }
     end
+
+    trait :discarded do
```
defong: do we actually need this?

``` ruby
let(:discarded) do
(create(:provider)).discard
end
```

davidgisbey:
Looks good to me 🚢 :shipit: 🚢

spec/factories/providers.rb
timabell: yes, to prove it *doesn't* end up in the output

timabell:
I'll improve the opt handling first...

timabell:
more things have come up, not ready yet....

timabell:
rebased and stripped `.DS_Store`

PR #892, opened 2019-10-08
2289 sync courses when saving provider

defong:
### Context
Sync courses when saving a provider

A provider's courses can only be sync

1. the provider is valid
2. it is the current recruitment cycle
3. there is syncable courses

### Guidance to review

spec/requests/api/v2/providers/update_spec.rb
- deals with 1

spec/requests/api/v2/providers/update_with_sync_spec.rb

- deals with 2 and 3

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -35,7 +35,7 @@ module API
         update_ucas_preferences
 
         if @provider.valid?
-          courses_synced?(@provider.syncable_courses) if @recruitment_cycle.current?
+          courses_synced?(@provider.syncable_courses) if @recruitment_cycle.current? && @provider.syncable_courses.present?
           render jsonapi: @provider.reload, include: params[:include]
         else
           render jsonapi_errors: @provider.errors, status: :unprocessable_entity, include: params[:include]
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -35,6 +35,7 @@ module API
         update_ucas_preferences
 
         if @provider.valid?
+          courses_synced?(@provider.syncable_courses) if @recruitment_cycle.current?
           render jsonapi: @provider.reload, include: params[:include]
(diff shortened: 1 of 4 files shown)


defong:
[When saving a provider enrichment also sync it to find at the same time](https://trello.com/c/OT26V8pu/2289-when-saving-a-provider-enrichment-also-sync-it-to-find-at-the-same-time)

app/controllers/api/v2/providers_controller.rb
```diff
@@ -35,6 +35,8 @@ def update
         update_ucas_preferences
 
         if @provider.valid?
+          courses_synced?(@provider.syncable_courses) if @recruitment_cycle.current? && @provider.syncable_courses.present?
```
misaka: Interestingly, we do a `@provider.reload` next line down ... which means that if anything has changed the provider record in the calls to `update_*` above (`update_enrichment`, etc), then the provider record will be updated. I think this is significant because if we expect that provider needs to be reloaded, then `@provider.syncable_courses` might be working off of stale data.

I think we should ensure we do `@provider.reload` before we call `@provider.syncable_courses`, and probably even before we do `@provider.valid?`.

Also, imo this line is too long to have everything on one line like this. Maybe turn it into a multi-line if-statement?

misaka: As per IRL discussion, the `@provider.reload` below is probably miss-placed and should really just be removed. Resolving for now.

spec/requests/api/v2/providers/update_with_sync_spec.rb
```diff
@@ -0,0 +1,193 @@
...
+                data: {
+                  attributes: {},
+                  type: "provider",
+                },
+              },
+            }
+      response
+    end
+
+    include_examples "Unauthenticated, unauthorised, or not accepted T&Cs"
```
misaka: Remove.

PR #902, opened 2019-10-10
[2327] Generate new course title from subjects

danielburnley:
### Context

When creating a course - we need to generate titles from the subjects given to ensure consistent naming.

- Add a new service for generating titles
- Loop in service to `build_new` endpoint

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/services/courses/generate_course_title_service_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+      context "Example one" do
+        let(:subjects) { [Subject.new(subject_name: "Cats")] }
+        it "Returns the subject name" do
+          expect(generated_title).to eq("Cats")
+        end
+
+        include_examples "with SEND"
+      end
+
+      context "Example two" do
```
misaka: Why do we need two examples?

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,54 @@
...
+  private
+
+    def is_modern_lanuage_course?(subjects)
+      subjects[0] == SecondarySubject.modern_languages
+    end
+
+    def modern_language_title(subjects)
+      title = "Modern Languages"
+
+      if subjects.length == 2
```
misaka: Are we testing if there is more than one subject here? Or language?

spec/services/courses/generate_course_title_service_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+        include_examples "with SEND"
+      end
+
+      context "with three or more languages" do
+        let(:subjects) do
+          [
+            modern_languages,
+            Subject.new(subject_name: "Cats"),
+            Subject.new(subject_name: "Dogs"),
+            Subject.new(subject_name: "Ducks"),
```
misaka: This confusing, esp with the service checking if there are 3 "subjects" when generating the title. Shouldn't the languages be in the title?

spec/services/courses/generate_course_title_service_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+          ]
+        end
+
+        it "Returns a name modern language with both languages" do
```
misaka: Adding the the confusion: `Returns a name modern language with both languages`

app/services/courses/generate_course_title_service.rb
misaka: It feels to me like we should be testing for how many languages there are on the course. I'm guessing we don't have an existing scope for that, but it'd be nice to have. Likewise, we could have a scope for the different types of subjects: `primary_subjects`, `secondary_subjects`, `languages`, etc.

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,54 @@
...
+      title
+    end
+
+  private
+
+    def is_modern_lanuage_course?(subjects)
+      subjects[0] == SecondarySubject.modern_languages
+    end
+
+    def modern_language_title(subjects)
```
defong: outcome will look amazing wrong....

https://github.com/DFE-Digital/manage-courses-backend/blob/4a627e7c3e15dac86a87f605cfdf0fb5ef6fd94a/app/services/subject_creator_service.rb#L56

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,54 @@
...
+      title
+    end
+
+    def generated_title(subjects)
+      return "" if subjects.empty?
+
+      if subjects.length == 1
+        subjects[0].to_s
+      else
+        "#{subjects[0]} with #{subjects[1]}"
```
defong: We **MUST** add these tests cases

https://github.com/DFE-Digital/search-and-compare-ui/blob/903e35ea800ccc4cf0b06a18a36e037ab813e229/tests/Unit.Tests/ViewFormatters/CourseExtensions.Tests.cs#L145

its related to 
https://github.com/DFE-Digital/search-and-compare-ui/blob/903e35ea800ccc4cf0b06a18a36e037ab813e229/shared/ViewFormatters/CourseExtensions.cs#L149

if title starts with any of the subject
```
                "Drama",
                "Media Studies",
                "PE",
                "Physical Education",
```

then there is no financial incentives

fyi @davidgisbey what was discussed earlier.

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      title = if course.further_education_course?
+                further_education_title
+              elsif is_modern_lanuage_course?(subjects)
+                modern_language_title(subjects)
+              else
+                generated_title(subjects)
+              end
+
+      title = append_send_info(title) if course.is_send?
```
defong: Physics **with** Psychology **with** Special educational needs and disability

?

spec/services/courses/generate_course_title_service_spec.rb
fofr: Languages should be in the title when there's 3. eg "Modern languages (French, German, Spanish)"
When there's more than 3 that's too many and we default to "Modern languages"

defong: F, G, S

looks alphabetical... is required?

misaka: Also, the test here has 3, but expects none to be in the title. I think this is related to the confusion of testing the length of `subjects` when we mean `languages` in the service.

app/services/courses/generate_course_title_service.rb
danielburnley: Yeah, I was thinking something similar at the time - would be nice to be able to query the languages on the course. I'll give it a look - I wasn't a great fan of it looking at subjects like ^ but wanted to get this in front of people for the general logic anyway

spec/services/courses/generate_course_title_service_spec.rb
danielburnley: I was TDD-ing it out and use examples for triangulation. I usually remove these before committing for this codebase because no where else has it so I wanna stay consistent :eyes: i missed this one though

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,54 @@
...
+      title
+    end
+
+  private
+
+    def is_modern_lanuage_course?(subjects)
+      subjects[0] == SecondarySubject.modern_languages
```
defong: will that be an issue? always first....

misaka: I would worry that, similar to below, `subjects` is all types of "subjects" ... maybe this should be:

```suggestion
      primary_subjects.include? SecondarySubject.modern_languages
```

app/services/courses/generate_course_title_service.rb
misaka: @defong , trying to decypher what you mean exactly, are you talking about the presence of "English as a second or other language" here? Or about the fact that, again, `subjects` could in theory include more than just "Modern Languages" + languages in it?

defong: @misaka 
the output
ie
```ruby
Modern Languages (#{subject_name})
```

throw out

`Modern Languages (Modern languages (other))`
`Modern Languages (English as a second or other language)`

in whichever single, duel, or triple form

danielburnley: I'll chat with @fofr about this tomorrow :eyes:

app/services/courses/generate_course_title_service.rb
danielburnley: Hm - So my concern with this is similar to a comment you had elsewhere

Do we have any consistency in what we deem a courses "Primary" subject and it's "Secondary" subject. I know in Find we had to base it off the title, but I'm concerned the ordering could be inconsistent during creation in the array and it causes issues with bursaries further down the line :thinking:

defong: Find don't care about that as it was never properly define. By ordinal or weight, ot string parsing.

app/services/courses/generate_course_title_service.rb
fofr: > Modern Languages (English as a second or other language)
Could probably shorten to "Modern Languages (English)"

danielburnley: Yeah that makes sense - not the nicest having a bunch of edge case handling but unless we do a database migration to update the name (which would have other visual downsides anyway) it'll have to do for now :thinking:

app/services/courses/generate_course_title_service.rb
fofr: It should be something like "Physics with Pyschology (SEND)"

defong: ummm....

Modern Languages (English) (SEND)

?

defong: @danielburnley confirmed with @fofr its fine 

>Modern Languages (English) (SEND)
Modern Languages (SEND)

with those as the use case should be relatively low and the user still retain the ability to request their own title

thro work like to have specs to explicitly test it as a way of documentation that was considered

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    def modern_language_title(subjects)
+      title = "Modern Languages"
+
+      languages = subjects.select { |s| s.type == "ModernLanguagesSubject" }
+
+      return title if languages.empty? || languages.length >= 4
+
+      languages = languages.map { |language| format_language_name(language) }
```
faissalMT: I prefer to avoid modifying a variable if the meaning changes, `languages` -> `language_names`

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      title = if course.further_education_course?
+                further_education_title
+              elsif is_modern_lanuage_course?(subjects)
+                modern_language_title(subjects)
+              else
+                generated_title(subjects)
+              end
+
+      title = append_send_info(title) if course.is_send?
```
defong: just fussy...

```suggestion
      title+= ' (SEND)' if course.is_send?
```

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+  private
+
+    def is_modern_lanuage_course?(subjects)
+      subjects.any? { |s| s == SecondarySubject.modern_languages }
+    end
+
+    def modern_language_title(subjects)
+      title = "Modern Languages"
```
defong:
```suggestion
      title = SecondarySubject.modern_languages.to_s
```

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,79 @@
...
+    def further_education_title
+      "Further education"
+    end
+
+    def append_send_info(title)
+      title + " (SEND)"
+    end
+
+    def format_subject_name(subject)
+      if subject.subject_name == "Communication and media studies"
```
defong:
```suggestion
      if subject.to_sym :communication_and_media_studies
      # or case insentience
```

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,79 @@
...
+        "Media studies"
+      else
+        subject.to_s
+      end
+    end
+
+    def format_language_name(language)
+      if language.subject_name == "English as a second or other language"
```
defong: same thing case or sym compare

app/services/courses/generate_course_title_service.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    def is_modern_lanuage_course?(subjects)
+      subjects.any? { |s| s == SecondarySubject.modern_languages }
+    end
+
+    def modern_language_title(subjects)
+      title = "Modern Languages"
+
+      languages = subjects.select { |s| s.type == "ModernLanguagesSubject" }
+      languages = languages.reject { |language| language.subject_name == "Modern languages (other)" }
```
defong: sym compare or case insensitive please

spec/services/courses/generate_course_title_service_spec.rb
```diff
@@ -0,0 +1,235 @@
...
+  let(:generated_title) { service.execute(course: course) }
+
+  before { modern_languages }
+
+  shared_examples "with SEND" do
+    context "With SEND" do
+      let(:is_send) { true }
+
+      it "Appends SEND information to the title" do
+        expect(generated_title).to end_with("(SEND)")
```
defong:
```suggestion
        expect(generated_title).to end_with(" (SEND)")
```

spec/requests/api/v2/build_new_course_spec.rb
```diff
@@ -37,6 +42,8 @@
       expect(response).to have_http_status(:ok)
       json_response = parse_response(response)
 
+      course_jsonapi["attributes"]["name"] = "Generated Title"
```
defong: not sure whats this mock is trying to tell me...

``` ruby
# actual service code
    def generated_title(subjects)
      return "" if subjects.empty?
```

give it nothing and mock it so something returns? bad bakery?

danielburnley: I wanted to write a test that just tested "Was this title being generated correctly", but this was the nicest way I could find :man_shrugging: i'll probably change this now i'm rereading this

danielburnley:
Rebased to fix merge conflicts

PR #905, opened 2019-10-11
[2330] Update a site_status’s vacancy status when editing a course study mode in MCB

fofr:
Site statuses have a vacancy status that's dependent on the course's study mode. eg A full time course has full time vacancies.

When updating the course through the UI we were ensuring this was updated. But no such checks were made in MCB.

- Move the method into the course
- Reuse the ensure method in MCB to keep changes consistent and correct

This is how site statuses got out of sync on https://trello.com/c/h4yNl5DH/

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -32,6 +32,7 @@ class Course < ApplicationRecord
   include WithQualifications
   include ChangedAt
   include Courses::EditOptions
+  include StudyModeVacancyMapper
 
   after_initialize :set_defaults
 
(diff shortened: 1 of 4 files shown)


fofr:
[course.publish_sites can fail silently and enrichments still get published](https://trello.com/c/h4yNl5DH/2330-coursepublishsites-can-fail-silently-and-enrichments-still-get-published)

defong:
> Approved for now - though I think this is part of a wider problem with MCB/Logic in controller duplication that might be in other places that we should consider in the future thinking but that doesn't matter for here

@misaka that chat the other day of moving logic out of controller is bearing weight now.

app/models/course.rb
```diff
@@ -442,6 +443,12 @@ def ensure_modern_languages
     end
   end
 
+  def ensure_site_statuses_match_study_mode
+    site_statuses.select(&:with_vacancies?).each do |site_status|
```
misaka: I know this is merged, but just to say there's a scope `with_vacancies` that could've, and I think probably should've, been used here. Just a minor performance issue.

PR #909, opened 2019-10-14
[2237] Remove DfESubject Model for financial incentives

davidgisbey:
### Context

As we are transitioning to pulling financial incentives directly from the database, the DfESubjects model is now redundant and can be removed.

- Remove the DfESubject model and spec
- Change the relevant serialisable course attributes to false. These will change once the solution is put in place.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


davidgisbey:
[Course subjects related tidy up](https://trello.com/c/V22Wtcfy/2237-course-subjects-related-tidy-up)

app/serializers/api/v2/serializable_course.rb
```diff
@@ -47,6 +46,26 @@ def enrichment_attribute(name, enrichment_name = name)
...
+      attribute :has_scholarship_and_bursary? do
+        false
+      end
+
+      attribute :has_early_career_payments? do
+        false
+      end
+
+      attribute :bursary_amount do
+        false
```
defong: i think amount should just be nil

app/serializers/api/v2/serializable_course.rb
```diff
@@ -47,6 +46,26 @@ def enrichment_attribute(name, enrichment_name = name)
...
+      end
+
+      attribute :scholarship_amount do
+        false
```
defong: i think amount should just be nil

PR #910, opened 2019-10-14
Raise an error if trying to publish with invalid site statuses

fofr:
### Context

Avoid publishing bad site statuses/courses in the event that the relationship between a course and it's site statuses has been corrupted somehow.

This happened before and the cause was fixed in #905

This will flag bad data and let us fix the data/causes before Publish/Find get into a broken state.

An error message would swallow the error show the user a message they couldn't recover from.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -163,6 +163,7 @@ class Course < ApplicationRecord
   validates :level, presence: { message: "^There is a problem with this course. Contact support to fix it (Error: L)" }, on: :publish
   validates :subjects, presence: true, on: :publish
   validate :validate_enrichment_publishable, on: :publish
+  validate :validate_site_statuses_publishable, on: :publish
   validate :validate_enrichment
   validate :validate_course_syncable, on: :sync
   validate :validate_qualification, on: :update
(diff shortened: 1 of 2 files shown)


defong:
[course.publish_sites can fail silently and enrichments still get published](https://trello.com/c/h4yNl5DH/2330-coursepublishsites-can-fail-silently-and-enrichments-still-get-published)

misaka:
@fofr This looks about right to me. Is it worth doing an analysis of what courses would stop being published as a result of this?

fofr:
I think this:

```      site_status.errors.full_messages.each do |msg|
        errors.add :base, "Site status: #{msg}"
      end```

Needs to raise an error instead of adding an error message. As it's a symptom of bad data that the user can't recover from.

**Comment by fofr** [comment]
PR updated to raise errors.

File: app/models/course.rb [hunk 1]
```diff
@@ -538,6 +539,14 @@ def validate_enrichment_publishable
     validate_enrichment :publish
   end
 
+  def validate_site_statuses_publishable
+    site_statuses.each do |site_status|
+      unless site_status.valid?
+        raise RuntimeError.new("Site status invalid: #{site_status.errors.full_messages.first}")
```
misaka: Can we add info about which provider and location(s) are invalid? At the moment, pulling the URL out for a request isn't as straightforward as it should be, and that's the only place I can think off, offhand, where we'd have that if we get this error.

PR #911, opened 2019-10-15
[2352] Add potential subjects to edit options

faissalMT:
### Context
As it stands the frontend has no subjects to choose from for subject editing.  

Send potential master subjects via edit options.

### Checklist

- [X] Attach to Trello card
- [x] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/spec/models/concerns/courses/edit_options_spec.rb
+++ b/spec/models/concerns/courses/edit_options_spec.rb
@@ -4,13 +4,28 @@ describe Course, type: :model do
   let(:course) { create(:course, level: "primary", subjects: [subjects]) }
   let(:subjects) { create(:subject, :primary) }
 
-  context "entry_requirements" do
+  describe "subjects" do
+    let(:course) { create(:course, level: "primary", subjects: []) }
+
+    it "returns the subjects the user can choose according to their level" do
+      primary_subject = create(:subject, :primary_with_mathematics)
+      create(:subject, :biology)
+
+      expect(course.potential_subjects).to match_array(
+        [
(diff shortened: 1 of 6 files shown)


faissalMT:
[`edit_options` should provide a list of available non-language subjects to the frontend](https://trello.com/c/9Eo13vkp/2352-editoptions-should-provide-a-list-of-available-non-language-subjects-to-the-frontend)

spec/models/concerns/courses/edit_options_spec.rb
```diff
@@ -4,6 +4,21 @@
   let(:course) { create(:course, level: "primary", subjects: [subjects]) }
   let(:subjects) { create(:subject, :primary) }
 
+  context "subjects" do
```
misaka: I know that this may feel like knit-picking, so apologies if it feels like that, but I think we should be standardising on how we use `describe` and `context`. For the sake of readability, because clearly functionally they are the same. As it turns out, this article reflects my own thinking on how these are used: https://lmws.net/describe-vs-context-in-rspec

Essentially, `describe` is for what functionality is being tested and `context` is the state the system is in for a particular test or set of tests. We should probably document this in the [spec README](https://github.com/DFE-Digital/manage-courses-backend/blob/master/spec/README.md).

To wit, this here feels like it should be a `describe` block, since we're testing the subjects functionality of `edit_options` here.

faissalMT: Fair enough, I just forgot

PR #913, opened 2019-10-15
[2313] optimise sync course

misaka:
### Context

With the recent work on subjects they weren't being preloaded causing the bulk sync job to become an N+1 query. 

Add a method to use the already-loaded course subjects, if present, instead of always refering to the DB.

Also add memoisation of the "Modern Languages" subject to prevent DB hits.

Add creation of subjects in the specs outside of the spec transaction so that the memoised "Modern Languages" object is always the right one. This approach was taken since the subjects are singleton objects and this may present a useful optimisation approach in general to such data.

### Guidance to review

Sorry this turned into something a bit bigger than just the `syncable_subjects`, but it's all part of the boyscout approach.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/factories/subjects.rb
+++ b/spec/factories/subjects.rb
@@ -31,14 +31,8 @@ FactoryBot.define do
 
     trait :primary_with_mathematics do
       subject_name { "Primary with mathematics" }
-      subject_code { "01" }
+      subject_code { "03" }
       type { "PrimarySubject" }
-    end
-
-    trait :primary_with_mathematics do
-      subject_name { "Primary with mathematics" }
-      subject_code { "04" }
-      type { :PrimarySubject }
     end
 
     trait :mathematics do
@@ -118,6 +112,17 @@ FactoryBot.define do
       subject_code { "C1" }
       type { :SecondarySubject }
     end
+
+    trait :modern_languages do
+      subject_name { "Modern Languages" }
+      subject_code { nil }
+      type { :SecondarySubject }
+    end
+  end
+
+  factory :secondary_subject do
+    sequence(:subject_code, &:to_s)
+    subject_name { Faker::ProgrammingLanguage.name }
 
     trait :modern_languages do
       subject_name { "Modern Languages" }
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -348,6 +348,30 @@ describe Course, type: :model do
       end
     end
 
+    describe "#syncable_subjects" do
+      let(:subject) { create :subject, :primary }
+      let(:subject_discontinued) { create :subject, :primary, type: "DiscontinuedSubject" }
+      let(:subject_without_code) { create :subject, :primary, subject_code: nil }
+      let(:course) do
+        create :course,
+               subjects: [subject, subject_discontinued, subject_without_code]
+      end
+
+      it "returns none-discontinued subjects that have a code present" do
+        expect(course.syncable_subjects).to eq [subject]
+      end
+
+      context "with a subjects that has been loaded" do
+        it "does not use where to reload subjects" do
+          allow(course.subjects).to receive(:where)
+
+          expect(course.syncable_subjects).to eq [subject]
+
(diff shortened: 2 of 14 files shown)


misaka:
[Optimise apiv1 courses](https://trello.com/c/qUut4StK/2313-optimise-apiv1-courses)

misaka:
@defong Happy now?? 🌮

spec/models/course_spec.rb
```diff
@@ -348,6 +348,30 @@
       end
     end
 
+    describe "#syncable_subjects" do
+      let(:subject) { create :subject, :primary }
+      let(:subject_discontinued) { create :subject, :primary, type: "DiscontinuedSubject" }
+      let(:subject_without_code) { create :subject, :primary, subject_code: nil }
+      let(:course) do
+        create :course,
+               subjects: [subject, subject_discontinued, subject_without_code]
```
```ruby
create :subject, :humanities
```

can we also add that one to the test as well please

spec/requests/api/v2/build_new_course_spec.rb
```diff
@@ -27,6 +27,7 @@
       class: {
         Course: API::V2::SerializableCourse,
         Subject: API::V2::SerializableSubject,
+        PrimarySubject: API::V2::SerializableSubject,
```
defong: you wanna add the rest....

secondary etc

spec/serializers/api/v2/serializable_course_spec.rb
```diff
@@ -69,6 +69,7 @@
         class: {
           Course: API::V2::SerializableCourse,
           Subject: API::V2::SerializableSubject,
+          PrimarySubject: API::V2::SerializableSubject,
```
defong: you wanna add the rest....

secondary etc

spec/factories/subjects.rb
```diff
@@ -125,4 +119,15 @@
       type { :SecondarySubject }
     end
   end
+
+  factory :secondary_subject do
```
defong: does this collide with the traits above...

misaka: No, because you need the factory name with a trait:

```ruby
create(:subject, :modern_languages)
```

vs

```ruby
create(:secondary_subject, :modern_languages)
```

spec/requests/api/v2/build_new_course_spec.rb
misaka: Not technically needed yet, I don't think? Offhand, could this be part of a larger bit of work to clean up subjects in the specs in general?

spec/serializers/api/v2/serializable_course_spec.rb
misaka: Not technically needed yet, I don't think? Offhand, could this be part of a larger bit of work to clean up subjects in the specs in general?

spec/models/course_spec.rb
misaka: Non capisco, per que?

defong: それが生産中のものであるように

spec/requests/api/v2/build_new_course_spec.rb
defong: that was my offhand comment... i take it as no

spec/serializers/api/v2/serializable_course_spec.rb
defong: that was my offhand comment... i take it as no

spec/factories/subjects.rb
defong: the former would be the same as the latter...

spec/models/course_spec.rb
misaka: Ამის მსგავსად?

misaka: အိုကေ ok ယခု ok?

defong: ydy mae hynny'n anhygoel

PR #917, opened 2019-10-17
Fix course specs failing due to presence of additional subjects

faissalMT:
Master broke due to https://github.com/DFE-Digital/manage-courses-backend/pull/911


misaka:
The other way to have fixed these is to use the `without_subjects: true` tag on these specs, but I think that as these subjects are singleton objects, it probably makes sense to add them to the expectations.

PR #921, opened 2019-10-21
[2362] Ensure sites get assigned to non-persisted courses

danielburnley:
### Context

During course creation, the `build_new` endpoint does not return sites when assigned, this is because the `sites=` method is overridden to change sites based on persisted values.

This is a workaround for that behaviour

- Add tests to existing behaviour
- Add tests for new course
- Change the logic on `sites=` where appropriate

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -377,8 +377,11 @@ class Course < ApplicationRecord
     to_remove = existing_sites - desired_sites
     to_remove.each { |site| remove_site!(site: site) }
 
-    sites.reload if persisted?
-    super(desired_sites) unless persisted?
+    if persisted?
+      sites.reload
+    else
+      super(desired_sites)
+    end
   end
 
   def has_bursary?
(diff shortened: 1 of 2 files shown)


danielburnley:
https://trello.com/c/gM3zwOzF/2362-coursesites-has-been-overriden-and-doesnt-assign-sites-to-new-courses-in-buildnew

app/models/course.rb
```diff
@@ -377,7 +377,8 @@ def sites=(desired_sites)
     to_remove = existing_sites - desired_sites
     to_remove.each { |site| remove_site!(site: site) }
 
-    sites.reload
+    sites.reload if persisted?
```
misaka: How about an `if ... else` here to make it clearer?

danielburnley: Yeah I was thinking that also - i'll change that and merge

PR #925, opened 2019-10-23
2367 add validation for new courses

defong:
### Context
Validation for course

Added validation for the presence of field where possible
Fixed the inferior level from the course factory

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -150,7 +150,7 @@ class Course < ApplicationRecord
   validates :enrichments, presence: true, on: :publish
   validates :is_send, inclusion: { in: [true, false] }
   validates :sites, presence: true, on: :publish
-  validates :level, presence: true, on: :publish
+  validates :level, presence: { message: "^There is a problem with this course. Contact support to fix it (Error: L)" }, on: :publish
   validates :subjects, presence: true, on: :publish
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_enrichment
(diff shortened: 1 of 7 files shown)


defong:
[[S] Add validation for new courses](https://trello.com/c/hIxIMfjp/2367-s-add-validation-for-new-courses)

danielburnley:
Notes from testing frontend:

### Level
- The error given is "There is a problem with this course. Contact support to fix it (Error: L)"
    - Feels oddly worded for new courses

### Age Range in years
- Missing

### Funding type (apprenticeship & fee/salaried)
- Missing

### Entry requirements (science, maths, english)
- Duplicate error messages

```
    Maths can't be blank
    Pick an option for Maths
    English can't be blank
    Pick an option for English
    Science can't be blank
    Pick an option for Science
```

### Applications open from
- Missing

defong:
#### Level
- Fixed via validation message differing based on `:publish` or `:create`
 
#### Age Range in years
- Fixed via validation on `:create`

#### Entry requirements (science, maths, english)
- Fixed by favouring `Pick an option xxx`

#### Not fixing 
##### Funding type (apprenticeship & fee/salaried)
- This is a calculated value
##### Applications open from
- This is used in mcb

danielburnley:
@defong although funding type is is calculated, it does have an `funding_type=` method which sets the `program_type`. Feels like to do validation on this might need something unpicking somewhere :thinking:

defong:
> @defong although funding type is is calculated, it does have an `funding_type=` method which sets the `program_type`. Feels like to do validation on this might need something unpicking somewhere thinking

Yeha, maybe we consider the `funding_type` and `program_type` usage and refactor it to be nicer...

danielburnley:
Feels like refactoring should be it's own task - Wanna start doing the FE task for these validations so having a starting point merged would be good even if it doesn't catch everything (though we should log somewhere what validations remain)

defong:
Carded up the two fields in question
https://trello.com/c/PxA5UD6j/2422-add-validations-for-funding-type-for-when-creating-a-new-course
https://trello.com/c/kY4sH57Y/2423-add-validations-for-applications-open-from-for-when-creating-a-new-course

as they quite more complex then the ones here

PR #926, opened 2019-10-23
[2400] Add available modern languages to edit options

faissalMT:
### Context
As it stands the subject editing functionality cannot add or modify modern languages. In order to be able to modify the list of modern languages the frontend will need to be provided with the modern languages.

In edit options add a subsection for potential modern languages that can be chosen from that is nil for any level that is not secondary, as other levels cannot have modern languages.

### Guidance to review

* argue about naming for a week or so

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


faissalMT:
[[S] Add potential "Modern Language" subjects to edit options](https://trello.com/c/MM7h4rT9/2400-s-add-potential-modern-language-subjects-to-edit-options)

app/models/concerns/courses/edit_options/subjects_concern.rb
```diff
@@ -13,6 +13,15 @@ def potential_subjects
             class: CourseSerializersService.new.execute,
           )[:data]
         end
+
+        def potential_modern_languages
```
defong: shouldn't it be just be ` modern_languages_subject`

faissalMT: I feel like this is more descriptive of its function

app/models/concerns/courses/edit_options.rb
```diff
@@ -26,6 +26,7 @@ def edit_course_options
           show_start_date: show_start_date?,
           show_applications_open: show_applications_open?,
           subjects: potential_subjects,
+          potential_modern_languages: potential_modern_languages,
```
defong: I meant here

app/models/concerns/courses/edit_options/subjects_concern.rb
defong: https://github.com/DFE-Digital/manage-courses-backend/pull/926#discussion_r338504345

app/models/concerns/courses/edit_options.rb
faissalMT: Same as my previous comment

defong: i disagree...you either have it or not have but not potential having it not having it

davidgisbey:
Looks good to me. I think it's fairly clear that potential_modern_languages contains a list of subjects to select from. No point arguing semantics.

app/models/concerns/courses/edit_options.rb
faissalMT: These are intended to be the modern languages a user may be able to select from if any.

defong:
> Looks good to me. I think it's fairly clear that potential_modern_languages contains a list of modern language subjects to select from. No point arguing semantics.

```
subjects: potential_subjects,
potential_modern_languages: potential_modern_languages
```

the exacting line above is called `subject` not `potential_subjects`
so why does `potential_modern_languages` is `potential_modern_languages`?

> These are intended to be the modern languages a user may be able to select from if any.


`semantics` are important
a server gives `exactly` what it can offer not `potential` what it can offer
a server gives `exactly` an empty list not `potential` empty list

the client can select  a `potential` item of their interest

this is client semantic leaking into server

app/models/concerns/courses/edit_options.rb
misaka: I think I agree about making the naming convention standard, and removing `potential` from the key, here. I would go one step further and favour the language `available` over `potential`, but that's less important.

```suggestion
          modern_languages: available_modern_languages,
```

timabell:
> No point arguing semantics

I disagree.

there-are-two-hard-problems-in-computer-science-0-naming-26556167

What about `available_modern_languages` ?

defong:
> @defong
> 
> > in full context it just looks potential ...
> 
> what are you on about? confused



here's the full context annotated fot you @timabell 
```javascript
{
    //meta of something
  "meta": {
     //edit_options of something
    "edit_options":{
       //name of something
      "entry_requirements": [...],
      //name of something
      "qualifications": [...],
      //name of something
      "age_range_in_years": [...],
      //name of something
      "start_dates": [...],
      //name of something
      "study_modes": [...],
      //name of something
      "show_is_send": [...],
      //name of something
      "show_start_date": [...],
      //name of something
      "show_applications_open": [...],
      //name of something
      "subjects": [...],  
      //potential name of something
      "potential_modern_languages":  [...]  
    }
  }
}
```

timabell:
I'm taking over finishing this off

timabell:
rebased

PR #927, opened 2019-10-23
[2364] Assign sites to courses for course creation

danielburnley:
### Context

We need to assign sites to courses in the create/build_new endpoints in order for the frontend to display them, and courses to be created with them

- Assign sites in the course controller
- `distinct` the sites relationship in the course model
  - Due to the way we solved the issue with sites= courses will be created with duplicated site statuses
    To work around that issue for now a distinct was added to the sites has_many to ensure the frontend only receives the sites added, not duplicates

### Checklist

https://trello.com/c/bFxJArKI/2364-assign-sites-to-course-during-creation

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -284,6 +284,13 @@ describe Course, type: :model do
         it "does not persist the course" do
           expect(course).not_to be_persisted
         end
+
+        xcontext "Which is then saved" do
+          it "Should only have two site statuses" do
+            course.save
+            expect(course.site_statuses.count).to eq(2)
+          end
+        end
       end
 
       it "should assign new sites" do
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -24,6 +24,7 @@ module API
         authorize @provider
         @course = Course.new(provider: @provider)
         update_subjects
+        update_sites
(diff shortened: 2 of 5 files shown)


spec/models/course_spec.rb
```diff
@@ -284,6 +284,13 @@
         it "does not persist the course" do
           expect(course).not_to be_persisted
         end
+
+        xcontext "Which is then saved" do
```
misaka: Should this be left pending?

danielburnley: I left it pending for now because I was wondering if it was worth attempting to create some kind of clean up after creation to remove the site_statuses, but after my quick spike I didn't get anywhere so :man_shrugging: 

I can remove it for now, but I do have concerns about potential risks when doing distinct if they're still duplicated. Though I do want to get this out the door :grimacing:

danielburnley: I might spend another 30 minutes on it prior to merging and see

misaka: This taken up tons of cycles already. At the risk of spawning more bugs, maybe we should just get it in and worry about the fallout later. :-/

danielburnley:
Rebased onto latest master - will merge when everything is passed

PR #928, opened 2019-10-23
2396 assign contacts with name

dankmitchell:
### Context
UCAS Contacts

- Ensure all UCAS contacts have a name, telephone number and email

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -21,6 +21,10 @@ class Contact < ApplicationRecord
 
   audited associated_with: :provider
 
+  validates :name, presence: true
+  validates :email, email: true, presence: true
+  validates :telephone, phone: { message: "^Enter a valid telephone number" }, allow_nil: true
+
   enum type: {
     admin: "admin",
          utt: "utt",
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -23,7 +23,7 @@ class Contact < ApplicationRecord
 
   validates :name, presence: true
   validates :email, email: true, presence: true
-  validates :telephone, phone: { message: "^Enter a valid telephone number" }, allow_nil: true
+  validates :telephone, phone: true, allow_nil: true
 
   enum type: {
     admin: "admin",
--- a/db/migrate/20191023195941_assign_provider_name_to_contacts.rb
+++ b/db/migrate/20191023195941_assign_provider_name_to_contacts.rb
@@ -1,0 +1,8 @@
+class AssignProviderNameToContacts < ActiveRecord::Migration[6.0]
+  def change
+    ucas_contacts_without_name = Contact.where(name: nil)
+    ucas_contacts_without_name.each do |contact|
+      contact.update(name: contact.provider.contact_name, telephone: contact.provider.telephone)
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


dankmitchell:
[Ensure all contacts have a "name" value](https://trello.com/c/X2vV1SqL/2396-ensure-all-contacts-have-a-name-value)

app/models/contact.rb
```diff
@@ -21,6 +21,10 @@ class Contact < ApplicationRecord
 
   audited associated_with: :provider
 
+  validates :name, presence: true
+  validates :email, email: true, presence: true
+  validates :telephone, phone: { message: "^Enter a valid telephone number" }, allow_nil: true
```
defong: should message be in yml?

spec/models/contact_spec.rb
```diff
@@ -45,5 +46,56 @@
       contacts.first.save
       expect(provider.reload.changed_at).to be_within(1.second).of Time.now.utc
     end
+
+
+    it { should validate_presence_of(:name) }
```
defong:
```suggestion
    it { should validate_presence_of(:name) }
    it { should validate_presence_of(:emai) }
```

app/models/contact.rb
dankmitchell: Same as provider telephone number

defong:
I think you forgot to up the schema number

db/migrate/20191023195941_assign_provider_name_to_contacts.rb
```diff
@@ -0,0 +1,8 @@
+class AssignProviderNameToContacts < ActiveRecord::Migration[6.0]
+  def change
+    ucas_contacts_without_name = Contact.where(name: nil)
```
misaka: We should probably add `.includes(:provider)` here for performance.

spec/models/contact_spec.rb
```diff
@@ -45,5 +46,56 @@
...
+      end
+
+      it "validates email contains an @ symbol" do
+        contact.email = "bar"
+        contact.valid?
+
+        expect(contact.errors[:email]).to include("^Enter an email address in the correct format, like name@example.com")
+      end
+
+      it "Does not validate the email if it is not present"do
```
misaka: "Does not validate the email if it is not present"? Not sure that fits the test below, or am I missing something?

app/models/contact.rb
misaka: I think @defong is right, we have a bunch of error messages already in locales, and this one probably belongs there.

https://github.com/DFE-Digital/manage-courses-backend/blob/413be4d4ada4407505ae6371614a929f80fbcf3a/config/locales/en.yml#L67-L68

db/migrate/20191023195941_assign_provider_name_to_contacts.rb
dankmitchell: Before: `1.0257s` 
After: `0.7702s`

💯

app/models/contact.rb
dankmitchell: Ah yes, you are right.

spec/models/contact_spec.rb
dankmitchell: Updated

db/migrate/20191023195941_assign_provider_name_to_contacts.rb
misaka: w00t! 😝 Ok, thought it would be a bit slower.

PR #935, opened 2019-10-25
[2416] - Refactor Course request spec

TobyRet:
Small followup PR for https://trello.com/c/5FZxbtXi/2388-fix-ucas-agerange to add a 'primary' and 'secondary' trait to the Course factory. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/factories/courses.rb
```diff
@@ -52,6 +52,16 @@
       )
     }
 
+    trait :primary do
+      level { :primary }
+      age_range_in_years { "3_to_7" }
+    end
+
+    trait :secondary do
+      age_range_in_years { "16_to_18" }
```
fofr: Minor: Can we do 11 to 18, as I think that's most common

fofr: 16 to 18 is an invalid age, courses must span an age range of more than 4 years.

fofr: Screen Shot 2019-10-25 at 15 42 46

TobyRet: Yes of course. Thanks for pointing this out @fofr

defong: feeling like we going to have a whack mole moment....
https://github.com/DFE-Digital/manage-courses-backend/pull/925

spec/factories/courses.rb
```diff
@@ -52,6 +52,16 @@
       )
     }
 
+    trait :primary do
```
defong: I don't think trait is quite right for setting both the `level` and `age_range_in_years`.

Given that there is a `infer_level` trait that assigns the `infer_level?` transient which takes in the the `subjects` that has been assigned to the in order to infer the `level` that the course should be.

ie 
```ruby

create(:course, :infer_level, subjects: ["primary"])

```
maybe `infer_level_and_age_range_in_years` or `infer_age_range_in_years`

if we are going to keep traits as a way of assigning `age_range_in_years` from a pool of available options


```ruby
trait :age_range_in_years_for_secondary do
 age_range_in_years {available_options.sample }
end

trait :age_range_in_years_for_primary do
 age_range_in_years {available_options.sample }
end

trait :age_range_in_years_for_fe do
 age_range_in_years {available_options.sample }
end

```

misaka: I'd suggest we just keep things simple. I don't think it's worth the extra machinery or cognitive load to have fancy, complicated traits that do a load of stuff. I'd prefer traits that are relatively simple short-hand for a common type of course, e.g. `:primary`, and anything else can be handled using traditional mechanisms, i.e. setting attributes.

TobyRet: Yeah I'm not keen on complicated traits / factories either. They should express the domain accurately though.
Is it not right to assume that a 'primary' course would have `level' equal to _primary_ and `age_range_in_years` equal to _3...11_ years?

defong: nah nothing wrong with that

but as long as we aware that this is incompatibility its fair game

```ruby
create(:course, :infer_level, :secondary, subjects: ["primary"])
```

dankmitchell:
[Refactor Course request spec](https://trello.com/c/zvgye3FI/2416-refactor-request-coursespecrb)

misaka:
🚢

spec/factories/courses.rb
defong: @TobyRet 

outcome of battle with @misaka 

trait names `:primary`, `:secondary`, and `:further_education` is poor so i suggest the below without setting the `level`.

```ruby
trait :age_range_in_years_for_secondary do
 age_range_in_years {available_options_for_secondary.sample }
end

trait :age_range_in_years_for_primary do
 age_range_in_years {available_options_for_primary.sample }
end

trait :age_range_in_years_for_further_education do
 age_range_in_years {available_options_for_further_education.sample }
end

```

example 
```ruby
create(:course, :age_range_in_years_for_further_education, level: :further_education)
```

Going follow, the current course factory has limited usages as based on a number of valid values are based on the course `level`

ie a course with a `level` of `further_education` can not have any other `subjects` that is not a `further_education` type subject as well as an `age_range_in_years` that is not appropriate for `further_education`

carded as https://trello.com/c/ZY5CbBG4/2455-course-factory-needs-be-broken-to-down-per-level-as-picking-a-valid-option-for-one-level-can-be-invalid-in-the-other

PR #944, opened 2019-10-30
[2437] Update courses title when subjects are updated

davidgisbey:
### Context

The title courses title(name attribute) is not currently being updated when it's subjects are being updated.

- update the course's name when it's subjects are updated
- refactor the course factory to use the correct name if is has subjects

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -178,6 +178,9 @@ module API
         return if subject_ids.nil?
 
         @course.subjects = Subject.where(id: subject_ids)
+        generate_course_title_service = Courses::GenerateCourseTitleService.new
+        @course.name = generate_course_title_service.execute(course: @course)
+        @course.save
       end
 
       def build_provider
@@ -259,7 +262,8 @@ module API
                   :sites_types,
                   :course_code,
                   :subjects_ids,
-                  :subjects_types)
+                  :subjects_types,
+                  :name)
           .permit(
             :english,
             :maths,
@@ -270,7 +274,6 @@ module API
             :applications_open_from,
             :study_mode,
             :is_send,
-            :name,
             :accrediting_provider_code,
             :funding_type,
             :level,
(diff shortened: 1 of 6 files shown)


davidgisbey:
[Call the course title service when a courses subjects are updated](https://trello.com/c/MhdyziFd/2437-call-the-course-title-service-when-a-courses-subjects-are-updated)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -178,6 +178,9 @@ def update_subjects
         return if subject_ids.nil?
 
         @course.subjects = Subject.where(id: subject_ids)
+        generate_course_title_service = Courses::GenerateCourseTitleService.new
+        @course.name = generate_course_title_service.execute(course: @course)
+        @course.save
```
danielburnley: don't think we should be calling `.save` here, it'll get called elsewhere and this method isn't related to saving :eyes: 
```suggestion
```

spec/factories/courses.rb
```diff
@@ -87,18 +87,28 @@
...
       if evaluator.age.present?
         course.created_at = evaluator.age
         course.updated_at = evaluator.age
         course.changed_at = evaluator.age
       end
 
+      if course.subjects.any?
+        generate_course_title_service = Courses::GenerateCourseTitleService.new
+        course.name = generate_course_title_service.execute(course: course)
+        course.save
```
defong: no need to save, this is factory for create

spec/factories/courses.rb
```diff
@@ -58,7 +58,7 @@
     end
 
     trait :secondary do
-      age_range_in_years { "11_to_18" }
+      age_range_in_years { "16_to_18" }
```
defong: any reason why this was changed?

app/controllers/api/v2/courses_controller.rb
davidgisbey: Done

spec/factories/courses.rb
davidgisbey: done

spec/factories/courses.rb
davidgisbey: It was accidentally changed in a rebase.

PR #945, opened 2019-10-30
2390 add validation for provider when saving

defong:
### Context
The concept of provider enrichments no longer applies, therefore validation needs to be tighten.

Remove the rest of the publish/publishable endpoints
 Amended validation to be tighter

### Guidance to review
:shipit: 

 `train_with_us` and `train_with_disability,` will be validated on update instead of create, due to mcb usage and it is quite lengthy text.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


defong:
[Add validation for provider when saving](https://trello.com/c/CMX569pT/2390-add-validation-for-provider-when-saving)

defong:
checked with @fofr workflow is fine if the user only update Contact details (`email`, `telephone`) or About your organisation (`train_with_us` & `train_with_disability`)  and click on the `Save and publish` button

This is to enforce we capture as much data as possible.

Validation should only show for the fields in their respective page.

PR #947, opened 2019-10-31
[2466] Updated prod subscription in mcb

rizzkhan1:
# Context
Update new prod subscription details in mcb


dankmitchell:
[Update MCB config to CIP subscription](https://trello.com/c/fskwcaHg/2466-update-mcb-config-to-cip-subscription)

PR #951, opened 2019-10-31
[2470] Assign provider to course before other params in creation

danielburnley:
### Context

The course doesn't have the provider set when creating the course with all params at the same time. This calls the `funding_type=` method to be called before the provider is set

Assign the provider to the course first before assigning other attributes

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


danielburnley:
https://trello.com/c/Oxh8sjbS/2470-course-creation-fails-when-program-type-is-set-to-fee

PR #953, opened 2019-11-01
2457  Update Code Climate ID

rizzkhan1:
### Context
Remove CC reporter id from the yaml file and move to the pipeline as a variable. 
Regenerate a new reporter id.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,8 +12,8 @@ variables:
   imageNameBgFindSync: 'manage-courses-bg-find-sync'
   # Support multiple docker-compose files to allow travis and az pipelines to co exist: https://docs.docker.com/compose/extends/#multiple-compose-files
   dockerOverride: 'docker-compose -f docker-compose.yml -f docker-compose.az.yml'
-  cc_Reporter_ID: 'de006d3ac97520a939b8c9b67af12b9e85438b9e9fd4255e78651800b36f48ae'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -12,8 +12,8 @@ variables:
   imageNameBgFindSync: 'manage-courses-bg-find-sync'
   # Support multiple docker-compose files to allow travis and az pipelines to co exist: https://docs.docker.com/compose/extends/#multiple-compose-files
   dockerOverride: 'docker-compose -f docker-compose.yml -f docker-compose.az.yml'
-  cc_Reporter_ID: 'de006d3ac97520a939b8c9b67af12b9e85438b9e9fd4255e78651800b36f48ae'
-  AGENT_JOBSTATUS: $(Agent.JobStatus)
```
dankmitchell: Did you mean to remove this line too?

azure-pipelines.yml
```diff
@@ -12,8 +12,8 @@ variables:
...
+  cc_Reporter_ID: '$(ccReporterID)'
```
dankmitchell: Why is this snake case?

azure-pipelines.yml
rizzkhan1: yes

dankmitchell: Okay

azure-pipelines.yml
rizzkhan1: removed snake case variable

PR #954, opened 2019-11-01
2441 add admin users

dankmitchell:
### Context
Proper admin users aka GOD ACCESS

- Add boolean to existing user model
- Update pundit policy scopes to user new `user.admin?`

### Guidance to review
- Remove all organisations from current user
- Set `admin: true` on current user
- 🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x]  Cleaned commit history

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -23,8 +23,6 @@
 class User < ApplicationRecord
   include AASM
 
-  DFE_EMAIL_PATTERN = '@(digital\.){0,1}education\.gov\.uk$'.freeze
-
   has_many :organisation_users
 
   # dependent destroy because https://stackoverflow.com/questions/34073757/removing-relations-is-not-being-audited-by-audited-gem/34078860#34078860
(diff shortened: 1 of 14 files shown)


dankmitchell:
[Making actual super users](https://trello.com/c/pgJvJwh4/2441-making-actual-super-users)

timabell:
Might be good to make the model invalid when the user is `admin: true` unless the user `email` address matches the old pattern. That would avoid us accidentally making a provider admin.

davidgisbey:
I'm assuming this will be followed up by a migration to set users with @education and @digital emails to admins?

dankmitchell:
> I'm assuming this will be followed up by a migration to set users with @education and @digital emails to admins?

That's one solution, yes @davidgisbey

davidgisbey:
Looks good to me then!

misaka:
Also, not that this would block an approval, but it seems `annotate` isn't running? Fixing that is probably a separate exercise, but maybe for now we could add the `admin` attribute to `course.rb` by hand?

PR #958, opened 2019-11-04
[2415] Ability to soft delete access requests

dankmitchell:
### Context
Being able to remove (discard/decline) Access Requests

Add soft delete to Access Requests

### Guidance to review
```
ar = AccessRequest.find(X)
ar.discard
```

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


dankmitchell:
[Add a decline button to access request support console](https://trello.com/c/hlv3p2nc/2415-add-a-decline-button-to-access-request-support-console)

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -377,4 +377,19 @@
       end
     end
   end
+
+  describe "DELETE #destroy" do
+    context "when an access request is discarded" do
+      let(:access_request) { create(:access_request) }
+
+      before do
+        access_request.discard
```
misaka: This looks like it's a model test, these should be request specs. 😝

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -377,4 +391,21 @@
...
+      access_request
+    end
+
+    before do
+      delete "/api/v2/access_requests/#{first_access_request.id}",
+             headers: { "HTTP_AUTHORIZATION" => credentials }
+    end
+
+    it "should add a discarded_at timestamp" do
+      expect(first_access_request.discarded_at).to be_within(1.second).of(Time.now.utc)
```
misaka: I feel like this is where @danielburnley's comment applies ... it'd be better to test that `access_request` doesn't show up in `AccessRequest.all` or `User.access_requests`, whichever we're more likely to use to get access requests.

dankmitchell: I have a test for this already.

spec/models/access_request_spec.rb
```diff
@@ -107,4 +109,20 @@
     its(:request_date_utc)  { should be_within(1.second).of Time.now.utc }
     its(:status)            { should eq "requested" }
   end
+
+  describe "default scope" do
+    context "discared records should not be returned" do
```
defong:
```suggestion
    context "discarded records should not be returned" do
```

app/policies/access_request_policy.rb
```diff
@@ -15,4 +15,5 @@ def create?
 
   alias_method :index?, :approve?
   alias_method :show?, :approve?
+  alias_method :destroy?, :approve?
```
defong: test?

dankmitchell: This is covered in the `approve` spec so I think we're good. Moving on...

PR #960, opened 2019-11-04
[2476] Refactor content status to service 

davidgisbey:
### Context

The course model is currently over 600 lines and has become a bit of a god class

- refactor content status into a service object

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/content_status_service.rb
+++ b/app/services/courses/content_status_service.rb
@@ -1,8 +1,8 @@
 module Courses
   class ContentStatusService
-    def execute(enrichment, next_cycle_boolean)
+    def execute(enrichment, recruitment_cycle)
       if !enrichment.present?
-        if next_cycle_boolean
+        if recruitment_cycle.next?
           :rolled_over
         else
           :empty
--- a/app/services/courses/content_status_service.rb
+++ b/app/services/courses/content_status_service.rb
@@ -1,0 +1,23 @@
+module Courses
+  class ContentStatusService
+    def execute(enrichment, next_cycle_boolean)
+      if !enrichment.present?
+        if next_cycle_boolean
+          :rolled_over
+        else
+          :empty
+        end
+      elsif enrichment.published?
+        :published
+      elsif enrichment.withdrawn?
+        :withdrawn
+      elsif enrichment.has_been_published_before?
+        :published_with_unpublished_changes
+      elsif enrichment.rolled_over?
+        :rolled_over
+      else
+        :draft
+      end
+    end
+  end
+end
--- a/spec/services/courses/content_status_service_spec.rb
+++ b/spec/services/courses/content_status_service_spec.rb
@@ -1,0 +1,65 @@
+require "rails_helper"
+
+describe Courses::ContentStatusService do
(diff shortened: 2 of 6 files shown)


davidgisbey:
[Look at the course model to see what can be refactored into a service or concern](https://trello.com/c/tYnakJY2/2476-look-at-the-course-model-to-see-what-can-be-refactored-into-a-service-or-concern)

app/services/courses/content_status_service.rb
```diff
@@ -0,0 +1,23 @@
+module Courses
+  class ContentStatusService
+    def execute(enrichment, next_cycle_boolean)
```
fofr:
```suggestion
    def execute(enrichment, next_recruitment_cycle?)
```
Any reason you didn't go with this?

spec/services/courses/content_status_service_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+describe Courses::ContentStatusService do
+  let(:service) { described_class.new }
+  let(:execute_service) { service.execute(enrichment, next_cycle_boolean) }
+  let(:next_cycle_boolean) { false }
+
+  context "when the enrichment parameter is nil" do
+    let(:enrichment) { nil }
+
+    context "and belongs to the next recruitment" do
+      let(:next_cycle_boolean) { true }
```
defong:
```suggestion
      let(:next_cycle_recruitment?) { true }
```

Any reason you didn't go with this?

app/services/courses/content_status_service.rb
davidgisbey: After speaking to Dan he thinks it's better to pass in the cycle to show the dependancy

app/services/courses/content_status_service.rb
```diff
@@ -0,0 +1,23 @@
+module Courses
+  class ContentStatusService
+    def execute(enrichment, recruitment_cycle)
```
misaka: I can't help but wonder if this shouldn't be simply parameterised on `course`. Why should the caller know that the `enrichment` and `recruitment_cycle` are required to calculate the content status? @danielburnley, did you have some thoughts on this?

danielburnley: Yeah David and I did have a chat about it yesterday - This way prevents the service being directly coupled to how enrichments/recruitment cycles are stored on the course, more specifically it prevents us needing to pull in that "Last enrichments" logic into here which feels not particularly nice to have to extend to this service. 

I also prefer things exposing what they depend on specifically rather than giving them a catch-all, at least in the case of only a few arguments required. If we were closer to 5 args needing to be passed in I'd err on the side of the whole course

As an aside, I also prefer named arguments to provide some guidance for future callers and readability where called :eyes:

misaka: Just chatted with @danielburnley IRL, and we agreed to leave it like this, with the intention to clean it up as part of the service containerisation changes. (https://github.com/DFE-Digital/manage-courses-backend/pull/949 et al)

PR #978, opened 2019-11-08
[2509] add semantic logger

misaka:
### Context

Default Rails logging to logstash is messy and painful to use.

Introduce `semantic_logger` which makes request logging to Logstash sane, and also pretties up file logging.

#### Before

12 events are logged for a single API request for a Provider's page. 2 of the events are custom events for authentication (won't show in Prod), and the rest is basically an event per line logged for Rails. And also none of the attributes are broken out to individual fields for searchability:

image

#### After

4 events are logged for a single API request for the provider's endpoint. Same as above, 2 of the events are custom events for authentication (and won't show in Prod), one is for the actual request and another is an extra one that the JSONAPI has spat out to tell us how long it took to generate the JSON. That one should be integrated with the request event, one day.

Also, here you can see that as part of the request event, there are plenty of fields we can use to search or filter on, such as `path`, `method`, or even timing info.

image

### Guidance to review

To test this locally you need to define `SETTINGS__LOGSTASH__HOST` and `SETTINGS__LOGSTASH__PORT` to value you can get from our logit.io account.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,6 +94,9 @@ gem "sidekiq-cron"
 # Offshore logging
 gem "logstash-logger", "~> 0.26.1"
 
+# Semantic Logger makes logs pretty
+gem "rails_semantic_logger"
+
 group :development, :test do
   # add info about db structure to models and other files
   gem "annotate"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -308,6 +308,10 @@ GEM
       nokogiri (>= 1.6)
     rails-html-sanitizer (1.3.0)
       loofah (~> 2.3)
+    rails_semantic_logger (4.4.3)
+      rack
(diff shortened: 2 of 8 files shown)


misaka:
[Add semantic-logger To Logstash Logging MCBE](https://trello.com/c/HbVVG6QE/2509-add-semantic-logger-to-logstash-logging-mcbe)

misaka:
I just noticed that the `user_id` is missing from the logs ... looks like a different thread is running the logstash custom logging stuff, I'm going to have to fix that.

TobyRet:
@misaka - trying to get Logit access so I can test this out locally

PR #987, opened 2019-11-12
2506-Add a config yaml file for pentest env

rizzkhan1:
### Context
Added a pentest.yaml file in config/settings folder to bring app settings setup inline with other environments.


dankmitchell:
Need to add the environment file too before merging this in

PR #994, opened 2019-11-14
[2534] - Don't use simplecov-console gem

TobyRet:
### Context

The `simplecov-console` gem was added recently in order to print results to the console rather than publish results to the 'coverage' folder. However following discussion with team have concluded this is too noisy so am reverting back to the previous method.

- remove `simplecov-console` gem and associated config.
- threshold added for passing Simplecov (75%). Currently the coverage is just over 75% and this is largely influenced by low code coverage in the mcb library.
- reduced noise in console when running specs in parallel

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,7 +5,9 @@ require "simplecov"
 require_relative "config/application"
 
 Rails.application.load_tasks
+Rake::Task["default"].clear 
 
 task lint: ["lint:ruby"]
 task annotate: ["db:annotate"]
-task default: %i[spec annotate lint brakeman]
+task parallel: ["parallel:spec"]
+task default: %i[parallel annotate lint brakeman]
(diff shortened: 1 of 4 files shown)


TobyRet:
[Remove SimpleCov-Console gem](https://trello.com/c/MTpZR32j/2534-remove-simplecov-console-gem)

Rakefile
```diff
@@ -5,7 +5,9 @@ require "simplecov"
 require_relative "config/application"
 
 Rails.application.load_tasks
+Rake::Task["default"].clear 
 
 task lint: ["lint:ruby"]
 task annotate: ["db:annotate"]
-task default: %i[spec annotate lint brakeman]
+task parallel: ["parallel:spec"]
+task default: %i[parallel annotate lint brakeman]
```
danielburnley: Do we want to have a separate explicit PR for moving these tests to now be parallel? :eyes: or at least a commit, feels like a significant change

TobyRet: Yep good idea - will make it a separate commit

PR #995, opened 2019-11-14
[2484] Make AccessRequest respond to include url parameters

faissalMT:
### Context
As it stands the access requests default to always include only the requester, this is inconsistent with dominant convention and prevents us extracting organisation data.

Have the AccessRequest `show` endpoint respond to the `?include` parameter.

### Guidance to review
This is necessary as part of https://github.com/DFE-Digital/manage-courses-frontend/pull/766

### Checklist

- [x] Attach to Trello card
- [X] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/access_requests_controller.rb
+++ b/app/controllers/api/v2/access_requests_controller.rb
@@ -20,7 +20,7 @@ module API
         if @access_request.discarded?
           render jsonapi: nil, status: :not_found
         else
-          render jsonapi: @access_request, include: [:requester]
+          render jsonapi: @access_request, include: params[:include]
         end
       end
 
--- a/spec/requests/api/v2/access_request_spec.rb
+++ b/spec/requests/api/v2/access_request_spec.rb
@@ -107,7 +107,7 @@ describe "Access Request API V2", type: :request do
   describe "GET #show" do
     let(:first_access_request) { create(:access_request) }
     let(:access_requests_show_route) do
-      get "/api/v2/access_requests/#{first_access_request.id}",
+      get "/api/v2/access_requests/#{first_access_request.id}?include=requester",
           headers: { "HTTP_AUTHORIZATION" => credentials }
     end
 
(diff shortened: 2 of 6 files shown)


faissalMT:
[Move AccessRequests functionality](https://trello.com/c/0FsaLfP1/2484-move-accessrequests-functionality)

app/controllers/api/v2/access_requests_controller.rb
```diff
@@ -56,6 +56,8 @@ def build_access_request
       end
 
       def access_request_params
+        #This needs to be modified so requester_email can't be changed by the frontend
+        # or is that not actually an issue?
```
misaka: The `access_request_params` appears to only be used by the `create` action, so I think that's ok.

spec/requests/api/v2/access_request_spec.rb
```diff
@@ -178,6 +178,13 @@
               "state" => first_access_request.requester.state,
               "admin" => first_access_request.requester.admin,
             },
+            "relationships" => {
+              "organisations" => {
+                "meta" => {
+                  "included" => false,
```
misaka: I feel like I'm missing something. The path used in this request adds `include=requester` (see [line 110](https://github.com/DFE-Digital/manage-courses-backend/pull/995/files#diff-153fbe77051b67cc87584ced1f4992cbR110)), which makes sense, but how does `organisation` get included then? Is there an implicit include (which would need to be changed)?

faissalMT: The organisation gets included by passing `?include=requester.organisation`

misaka: That's how it should be included, yes. But this test doesn't do that (see link in above comment). How is it being included?

misaka: Had another look and I see I missed that this is not, in fact, being included, it's just listed as a relationship. Pardon my confusion, looks good.

PR #999, opened 2019-11-15
[2511] - add (more) request ID support to Manage Course Backend

TobyRet:
### Context

When a request comes in, add the value of the X-Request-ID to the logs, and send this along with any background processing jobs, e.g. syncing a course to Find. This is so we can track the requests journey through the system

- Add RequestStore gem to store request ids safely
- Generate a new request ID if there is no request id
- Store the request ID using RequestStore
- Add the request ID to headers when running course sync jobs

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[Add Request ID support to MCBE](https://trello.com/c/orZ6SvYi/2511-add-request-id-support-to-mcbe)

app/controllers/api/v2/application_controller.rb
```diff
@@ -54,7 +55,11 @@ def append_info_to_payload(payload)
             sign_in_id: current_user.sign_in_user_id,
           }
         end
-        payload[:request_id] = request.uuid
+        payload[:request_id] = RequestStore.store[:request_id]
+      end
+
+      def store_request_id
+        RequestStore.store[:request_id] = request.uuid || SecureRandom.uuid
```
misaka: Do we need the `|| SecureRandom.uuid`? So far as I can tell, if there is no request-id one will be generated for us.

TobyRet: I added this based on the acceptance criteria but looks like you are correct judging from this => https://api.rubyonrails.org/classes/ActionDispatch/RequestId.html

TobyRet: Have removed :-)

PR #1005, opened 2019-11-19
[2542] Add descriptive error for missing MCB environment

faissalMT:
### Context
If a user specifies an environment that is not listed in the config file they will receive a cryptic `KeyError`, this is an error that is likely to appear often and as such should be clear in order to avoid wasting time during setup.

Adds a message that clarifies the error to the user.

### Guidance to review
This PR has been written in such a way that it anticipates @davidgisbey's fix for the wrong environment being used, it will direct the user to modify the correct config depending on `RAILS_ENV`.

### Checklist

- [x] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -315,13 +315,25 @@ module MCB
 
     def env_to_azure_map(opts)
       azure_environments_file = File.join("config", "azure_environments.yml")
-      azure_environments = YAML.safe_load(File.read(azure_environments_file))
+      begin
+        azure_environments_file_contents = File.read(azure_environments_file)
+      rescue Errno::ENOENT
+        raise Errno::ENOENT, "Could not find config/azure_environments.yml, consult the MCB section of README.md"
+      end
+
+      azure_environments = YAML.safe_load(azure_environments_file_contents)
+
+      unless azure_environments.key?(opts[:env])
+        raise KeyError, "The environment '#{opts[:env]}' could not be found, have you made sure to add it to your config/azure_environments.yml?"
+      end
+
       azure_environments[opts[:env]]
     end
 
     def load_env_azure_settings(opts)
       if opts.key?(:env)
         env_settings = env_to_azure_map(opts)
+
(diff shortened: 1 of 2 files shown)


faissalMT:
[Add descriptive error for when the details for an environment can't be found](https://trello.com/c/MmnzZ3Fx/2542-add-descriptive-error-for-when-the-details-for-an-environment-cant-be-found)

lib/mcb.rb
```diff
@@ -323,7 +323,11 @@ def env_to_azure_map
 
     def load_env_azure_settings(opts)
       if opts.key?(:env)
-        env_settings = env_to_azure_map.fetch(opts[:env].to_sym)
+        begin
+          env_settings = env_to_azure_map.fetch(opts[:env].to_sym)
+        rescue KeyError
+          raise KeyError, "The environment '#{opts[:env]}' could not be found, have you made sure to add it to your config/settings/#{ENV['RAILS_ENV']}.local.yml?"
+        end
```
timabell: I'm curious why the `if opts.key?(:env)` doesn't make the `rescue` unreachable as I would have expected, do you know?

timabell: (I was originally going to suggest changing the rescue to an if block, then I saw one and was confused)

timabell:
Thanks for doing this :+1:

misaka:
This branch is conflicting probably because of @davidgisbey's changes to how this works. #1006 is going to change it further ... might be worth coordinating with that.

davidgisbey:
All done on my end and merged.

lib/mcb.rb
faissalMT: The `KeyError` arises as a result of `env_to_azure_map.fetch`, the `opts.key?` will only ensure that the options have specified an environment to switch to.

timabell: Ah in which case I'd misunderstood the code as written so I definitely would switch the rescue to an if

PR #1006, opened 2019-11-19
[2517] Refactor mcb.rb to use opts[:env] when loading the settings file

davidgisbey:
### Context
image

Sync to find was not working correctly due to the above method in mcb.rb`

- refactors the above to use opts[:env] to ensure the correct settings are being used for the selected environment.
- adds an azure section to the settings/qa.local.yml, settings/staging.local.yml, settings/production.local.yml files
- updates the README on how to connect to MCB (add .local files for qa, staging and prod.

### Guidance to review
The updated confluence is here.
https://dfedigital.atlassian.net/wiki/spaces/BaT/pages/1182761062/MCB+Configuration
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,15 +194,22 @@ can be useful for people who aren't as familiar with the app, or with certain
 complex operations that just aren't already packaged up in the app.
 
 In order to use the external environment functionality you must set the
-environments in `config/settings/development.local.yml` like so:  
+environments in ```config/azure_environments.yml ```like so:  
 ```
-azure:
-  <environment name>:
+qa:
     webapp: <webapp>
     rgroup: <rgroup>
     subscription: <subscription>
+staging:
+  webapp: <webapp>
+  rgroup: <rgroup>
+  subscription: <subscription>
+production:
+  webapp: <webapp>
+  rgroup: <rgroup>
+  subscription: <subscription>
 ```
-If you are a member of the Find team you may find a filled out config [here](https://dfedigital.atlassian.net/wiki/spaces/BaT/pages/1182761062/MCB+Configuration?atlOrigin=eyJpIjoiMzdkOGJlNGU0NTNhNDYyMGI4NTI3Mzg2ZTVjZGEyMjUiLCJwIjoiYyJ9).
+If you are a member of the Find team you may find a filled out config [here](https://dfedigital.atlassian.net/wiki/spaces/BaT/pages/1182761062/MCB+Configuration?atlOrigin=eyJpIjoiZDg0N2Q2ZTg0NTRiNDQ1MmEwZWQ3M2VhZjMyYjIxNjEiLCJwIjoiYyJ9n).
 
 The script's functionality is accessed using sub-commands with built-in
 documentation. This is the best way to discover it's functionality and the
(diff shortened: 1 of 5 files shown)


davidgisbey:
[Update sync course to find on MCB to use sync service](https://trello.com/c/pfCsAYYP/2517-update-sync-course-to-find-on-mcb-to-use-sync-service)

README.md
```diff
@@ -194,10 +194,10 @@ can be useful for people who aren't as familiar with the app, or with certain
 complex operations that just aren't already packaged up in the app.
 
 In order to use the external environment functionality you must set the
-environments in `config/settings/development.local.yml` like so:  
+environments in `config/settings/development.local.yml`, `config/settings/qa.local.yml`
+`config/settings/staging.local.yml` and `config/settings/production.local.yml` like so:
```
misaka: We should be able to just add these to the file `config/settings.local.yml` instead of requiring them in all the environments.

config/settings/qa.yml
```diff
@@ -6,3 +6,7 @@ bg_jobs:
     cron: "0 */6 * * *"
     class: "BulkSyncCoursesToFindJob"
     queue: find_sync
+azure:
+  webapp: dummy-webapp
+  rgroup: dummy-rgroup
+  subscription: dummy-subscription
```
misaka: If just using `config/settings.local.yml` works, let's put these dummy settings in `config/settligs.yml`?

spec/requests/api/system/force_sync_spec.rb
```diff
@@ -7,6 +7,14 @@
...
 
   before do
+    #The settings need to be reset as it has cached the qa settings (see
+    # db_spec.rb:25)
+    Rails.env = "test"
+    Settings.reload_from_files(
+      Rails.root.join("config", "settings.yml").to_s,
+      Rails.root.join("config", "settings", "#{Rails.env}.yml").to_s,
+      Rails.root.join("config", "environments", "#{Rails.env}.yml").to_s,
+    )
```
misaka: Is this needed because the last test to run is `db_spec.rb`? If so, then this is the wrong way to fix it I think, because:

1. tests could run in a different order, which would mean that a different test could fail (because the Settings are still set to `qa`)
2. other tests could also invoke the `env_to_azure_map` code path, which will break some other tests

I think we the best way to handle this is to stub out the bit that reloads settings in `env_to_azure_map`, it may not be clear how to do that so grab me next time I'm there, or maybe @danielburnley or @TobyRet, to help explain how that might be done.

davidgisbey: I think all the comments have now been addressed. Thanks for the help @misaka !

davidgisbey:
I think all the comments have now been addressed. Thanks for the help @misaka !

misaka:
@davidgisbey Just looking at the test errors, sounds like you might need to stub out `env_to_azure_map`.

davidgisbey:
@misaka have stubbed it out. Just waiting to see if the tests pass then I'll squash the commits and merge.

PR #1015, opened 2019-11-25
[2579] Sort potential subjects by name

davidgisbey:
### Context

Secondary subjects are currently not in alphabetical order.

- Sort secondary subjects before serialising them into the edit options serialiser

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/concerns/courses/edit_options/subjects_concern.rb
+++ b/app/models/concerns/courses/edit_options/subjects_concern.rb
@@ -8,7 +8,7 @@ module Courses
         #
         # https://github.com/DFE-Digital/manage-courses-frontend/blob/master/spec/factories/courses.rb
         def potential_subjects
-          self.assignable_master_subjects
+          self.assignable_master_subjects&.sort_by(&:subject_name)
         end
 
         def available_modern_languages
--- a/spec/models/concerns/courses/edit_options_spec.rb
+++ b/spec/models/concerns/courses/edit_options_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 describe Course, type: :model do
   let(:course) { create(:course, level: "primary", subjects: [subjects]) }
-  let(:subjects) { find_or_create(:primary_subject, :primary) }
+  let(:subjects) { find(:primary_subject, :primary) }
 
   describe "modern languages" do
     context "primary level" do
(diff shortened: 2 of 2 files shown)


davidgisbey:
[Secondary subject page: Sort subjects by name](https://trello.com/c/IFDe6641/2579-secondary-subject-page-sort-subjects-by-name)

spec/models/concerns/courses/edit_options_spec.rb
```diff
@@ -62,6 +63,39 @@
...
+          find_or_create(:secondary_subject, :mathematics),
+          find_or_create(:secondary_subject, :modern_languages),
+          find_or_create(:secondary_subject, :music),
+          find_or_create(:secondary_subject, :philosophy),
+          find_or_create(:secondary_subject, :physical_education),
+          find_or_create(:secondary_subject, :physics),
+          find_or_create(:secondary_subject, :psychology),
+          find_or_create(:secondary_subject, :religious_education),
+          find_or_create(:secondary_subject, :science),
+          find_or_create(:secondary_subject, :social_sciences),
```
timabell: it's surprising to me to see `find_or_create` in the assertion, can we change that to something that would just `find` and would `raise` if it couldn't find it?

timabell: second question, where does this list come from, is it a fragile test because it encodes the whole subject list again?

davidgisbey: It's a sorted list of all the current secondary subjects. So it would break if the subject list changes. That may well actually be desirable. I've refactored it to use find. 

I also realised it was sorting primary subjects too. Not sure if this was desirable behaviour or not so i've changed the logic.

davidgisbey: `find` should work fine due to memoisation of the subjects 👍

app/models/concerns/courses/edit_options/subjects_concern.rb
```diff
@@ -8,7 +8,11 @@ module SubjectsConcern
         #
         # https://github.com/DFE-Digital/manage-courses-frontend/blob/master/spec/factories/courses.rb
         def potential_subjects
-          self.assignable_master_subjects
+          if level == "secondary"
```
faissalMT: When not sort subjects regardless?

app/models/concerns/courses/edit_options/subjects_concern.rb
```diff
@@ -8,7 +8,11 @@ module SubjectsConcern
...
+            self.assignable_master_subjects&.sort_by(&:subject_name)
```
faissalMT: What's the need for the safe navigation operator here?

app/models/concerns/courses/edit_options/subjects_concern.rb
davidgisbey: Yea good shout. Previously had it that way 🤦‍♂

app/models/concerns/courses/edit_options/subjects_concern.rb
davidgisbey: done

PR #1014, opened 2019-11-22
[2519] Add an mcb command for getting accredited courses

timabell:
### Context

We are working on understanding the needs of accredited bodies, in the short-term this will allow support to accurately and repeatably provide csv extracts to accredited bodies.

### Changes proposed in this pull request

* Extend provider model to give access to courses that a provider is the accredited body for.
* Add an mcb command
* Output is one row per course-site

### Not in this PR

* Easier to understand columns (currently exposes our raw model) - caused some confusion re draft / published etc.

### Guidance to review

Give it your worst!

```
mcb providers accredited_course_sites P80
mcb providers accredited_course_sites P80 -E qa
mcb providers accredited_course_sites P80 -f out.csv
```

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #1017, opened 2019-11-26
[2565] - Geocoder setup

TobyRet:
### Context

This PR adds geocoding to the Site and Provider models. When either of these models is created or updated, a background job is triggered to search and store the latititude and longitude for each of their addresses. Note the background jobs are only triggered if address details have changed.

- Stubbing of Geocoder requests
- Creation of geocoding background job
- Triggering of geocoding background job in the Site and Provider model
- Creation of Geocoder config (uses Google for search, adds caching)

### Guidance to review
- Aiming for no more than three rejections :-)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[Add geocoding setup](https://trello.com/c/hDXI6Vse/2565-add-geocoding-setup)

app/jobs/geocode_job.rb
```diff
@@ -0,0 +1,15 @@
+class GeocodeJob < ApplicationJob
```
dankmitchell: Nice!

spec/support/helpers.rb
```diff
@@ -0,0 +1,44 @@
+module Helpers
```
defong: Can we have that by itself
`helper` just seems too generic and probably make this file grow to monster

image

config/settings/development.yml
```diff
@@ -4,3 +4,5 @@ mcbg:
   redis_password:
 system_authentication_token: "Ge32"
 log_level: debug
+google:
```
dankmitchell:
```suggestion
gcp:
```

config/settings/pentest.yml
```diff
@@ -6,3 +6,5 @@ bg_jobs:
     cron: "0 */6 * * *"
     class: "BulkSyncCoursesToFindJob"
     queue: find_sync
+google:
```
dankmitchell:
```suggestion
gcp:
```

defong:
looks like it is suffering from the same fate as #1014 @misaka  @timabell 

image

spec/support/helpers.rb
TobyRet: Yep good idea

TobyRet:
Screenshot 2019-11-29 at 09 57 36

Boo. So I realised that I was only geocoding after updating the address in a Provider and Site. We also want to do this when creating Provider and Site but this is causing stack overflow errors

app/models/provider.rb
```diff
@@ -120,6 +120,17 @@ def current_accredited_courses
...
   before_discard { discard_courses }
 
+  geocoded_by :full_address
+  after_commit -> { GeocodeJob.perform_later("Provider", id) }
+
+  def full_address
+    [address1, address2, address3, address4, postcode].compact.join(", ")
+  end
+
+  def address_changed?
```
dankmitchell: Is this method being used anymore?

TobyRet: Yep it's back in there now

PR #1026, opened 2019-12-02
[2614] Reorder outcome edit options

danielburnley:
### Context

MCFE Should display the options in line with the prototype, with the most common at the top and the least common at the bottom

- Change the ordering to qualifications to be hard coded in the edit options

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/concerns/with_qualifications.rb
+++ b/app/models/concerns/with_qualifications.rb
@@ -39,7 +39,7 @@ module WithQualifications
     # subjects this course was tagged to.
     #
     # Defined here: https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Domain/Models/CourseQualification.cs
-    enum qualification: %i[qts pgce_with_qts pgde_with_qts pgce pgde]
+    enum qualification: %i[pgce_with_qts qts pgde_with_qts pgce pgde]
 
     # This field may seem like an unnecessary overhead when there is already a
     # database-backed `qualification` field. However it's misleading, from the
--- a/spec/models/concerns/courses/edit_options/qualifications_concern_spec.rb
+++ b/spec/models/concerns/courses/edit_options/qualifications_concern_spec.rb
@@ -18,7 +18,7 @@ describe Courses::EditOptions::QualificationConcern do
     let(:level_value) { "primary" }
 
     it "returns only QTS options for users to choose between" do
-      expect(example_model.qualification_options).to eq(%w[qts pgce_with_qts pgde_with_qts])
+      expect(example_model.qualification_options).to match_array(%w[qts pgce_with_qts pgde_with_qts])
       example_model.qualification_options.each do |q|
         expect(q).to include("qts")
       end
--- a/spec/models/concerns/courses/edit_options_spec.rb
+++ b/spec/models/concerns/courses/edit_options_spec.rb
@@ -107,7 +107,7 @@ describe Course, type: :model do
   describe "qualifications" do
     context "for a course that's not further education" do
       it "returns only QTS options for users to choose between" do
-        expect(course.qualification_options).to eq(%w[qts pgce_with_qts pgde_with_qts])
+        expect(course.qualification_options).to match_array(%w[qts pgce_with_qts pgde_with_qts])
         course.qualification_options.each do |q|
           expect(q.include?("qts")).to be_truthy
         end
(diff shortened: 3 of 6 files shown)


danielburnley:
https://trello.com/c/mBHGsPFT/2614-outcome-page-reorder-choices

danielburnley:
I'm not 100% that this solution is the best way to do this, I'm concerned that we could end up with a mismatch in what options are displayed to the user if we ever add one. But that feels unlikely so :man_shrugging: I welcome comments

defong:
Shouldnt this be on frontend?

danielburnley:
@defong We don't want to couple the frontend to the options given by the backend, so we've used edit options to drive the available values on the frontend. Which means the best place to order these as far as i can tell is the backend edit options. 

The frontend is also an option, but would require duplicating all the logic of deciding which options to display which I think is less ideal

spec/models/concerns/courses/edit_options/qualifications_concern_spec.rb
```diff
@@ -18,7 +18,7 @@
     let(:level_value) { "primary" }
 
     it "returns only QTS options for users to choose between" do
-      expect(example_model.qualification_options).to eq(%w[qts pgce_with_qts pgde_with_qts])
+      expect(example_model.qualification_options).to match_array(%w[qts pgce_with_qts pgde_with_qts])
```
defong: The order of laziness, thats just too lazy.

spec/models/concerns/courses/edit_options_spec.rb
```diff
@@ -107,7 +107,7 @@
   describe "qualifications" do
     context "for a course that's not further education" do
       it "returns only QTS options for users to choose between" do
-        expect(course.qualification_options).to eq(%w[qts pgce_with_qts pgde_with_qts])
+        expect(course.qualification_options).to match_array(%w[qts pgce_with_qts pgde_with_qts])
```
defong: The order of laziness, thats just too lazy.

spec/models/concerns/courses/edit_options/qualifications_concern_spec.rb
danielburnley: hah, yeah I didn't wanna have the tests tied to the ordering if possible :man_shrugging: then again we're saying the ordering is important in the backend so i guess they should be

danielburnley:
I didn't even see that line in that file with all the comments :crying_cat_face: But that looks to have worked, nice spot @defong

misaka:
> @defong We don't want to couple the frontend to the options given by the backend, so we've used edit options to drive the available values on the frontend. Which means the best place to order these as far as i can tell is the backend edit options.

I kind of feel like the backend should be responsible for _what_ options are available, and the frontend for the ordering. This would mean the frontend has a list of the ordering, but would check for the existence of each option in `edit_options`.

Does that make sense?

spec/models/concerns/courses/edit_options/qualifications_concern_spec.rb
defong: Its both really subtle and really obscured.
So there is a real need for tests here.

app/models/concerns/with_qualifications.rb
```diff
@@ -39,7 +39,7 @@ module WithQualifications
     # subjects this course was tagged to.
     #
     # Defined here: https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Domain/Models/CourseQualification.cs
-    enum qualification: %i[qts pgce_with_qts pgde_with_qts pgce pgde]
+    enum qualification: %i[pgce_with_qts qts pgde_with_qts pgce pgde]
```
defong: before
image

after
image

You want something like what was describe here  https://github.com/DFE-Digital/manage-courses-backend/pull/1026#pullrequestreview-326432116


Can we leave a comment here as well so someone knows why it was hardcoded like that.

danielburnley:
Yeah given there's now a few of us thinking the frontend should do the ordering not the backend i'm more inclined to find a way to do it there - I wasn't sure if it was just my own gut feeling. Feels like it'll make this backend stay a bit more sane too.

I'll close this, change the trello ticket a bit, and re-do the PR on the frontend. Thanks @defong and @misaka :tada:

PR #1033, opened 2019-12-04
[2631] Add sites to course creation

faissalMT:
### Context
In order for https://github.com/DFE-Digital/manage-courses-frontend/pull/798 to function correctly the course needs to have site information.  

Provide the provider sites when requesting a course object from `build_new()`.

### Guidance to review
We'll want to have an include parameter at some point, but I think this works for now.  
  
This also includes a fix for a test which was breaking if `config/azure_environments.yml` was present on your real filesystem.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/build_new_course_spec.rb
+++ b/spec/requests/api/v2/build_new_course_spec.rb
@@ -11,6 +11,7 @@ describe "/api/v2/build_new_course", type: :request do
            provider_type: "Y"
   end
   let(:provider2) { create(:provider) }
+
   let(:payload) { { email: user.email } }
   let(:token) do
     JWT.encode payload,
@@ -31,8 +32,9 @@ describe "/api/v2/build_new_course", type: :request do
         Subject: API::V2::SerializableSubject,
         PrimarySubject: API::V2::SerializableSubject,
         Provider: API::V2::SerializableProvider,
+        Site: API::V2::SerializableSite,
       },
-      include: %i[subjects sites provider accrediting_provider],
+      include: [:subjects, :sites, :accrediting_provider, :provider, provider: [:sites]],
     ).to_json)
   end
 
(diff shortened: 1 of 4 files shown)


faissalMT:
[Implement back button for further ed providers](https://trello.com/c/39EnNzVs/2631-implement-back-button-for-further-ed-providers)

spec/spec_helper.rb
```diff
@@ -10,6 +10,10 @@
 SimpleCov.start "rails" do
   add_filter "/mcb/"
 end
+SimpleCov.start do
```
defong: Why is this here?

spec/requests/api/v2/build_new_course_spec.rb
```diff
@@ -31,8 +31,9 @@
         Subject: API::V2::SerializableSubject,
         PrimarySubject: API::V2::SerializableSubject,
         Provider: API::V2::SerializableProvider,
+        Site: API::V2::SerializableSite,
       },
-      include: %i[subjects sites provider accrediting_provider],
+      include: [:subjects, :sites, :accrediting_provider, :provider, provider: [:sites]],
```
defong: Is this actually getting tested?

spec/spec_helper.rb
faissalMT: Specs were being included as part of the coverage

spec/requests/api/v2/build_new_course_spec.rb
faissalMT: This is the test, if you remove this change the tests will fail.

defong: i was kinda expecting a test that was like `returns a course with subject relationships`

PR #1034, opened 2019-12-04
[2622] Add V2 organisation endpoint

davidgisbey:
### Context

We need to implement an organisation endpoint for porting over the organisations' support page on Publish.

- Add an organisation endpoint

### Guidance to review

If providers are included it should only return the current cycle's providers. This is due to duplicate providers being returned as providers are not cycle aware.

There will be a follow-up PR which will address the provider index endpoint sorting providers before returning them. This should be done via a query param.  

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


davidgisbey:
[Add V2 organisations endpoint](https://trello.com/c/PCvT62El/2622-add-v2-organisations-endpoint)

config/routes.rb
```diff
@@ -156,6 +157,7 @@
       resources :recruitment_cycles,
                 only: %i[index show],
                 param: :year do
+        resources :organisations, only: :index
```
misaka: Do we actually need / want this endpoint?

davidgisbey: I wasn't 100% sure since orgs aren't cycle aware. Makes it simpler to grab params[:recruitment_cycle] if needed if it does though.

davidgisbey: have removed it 👍

PR #1045, opened 2019-12-06
2647 - Batch geocoding

TobyRet:
### Context
We want to be able to backfill Sites and Providers with geolocation information.

- Handle nil results returned from Geocoder
- Create mcb geocoding command for Sites and Providers. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/jobs/geocode_job.rb
```diff
@@ -4,7 +4,7 @@ class GeocodeJob < ApplicationJob
   def perform(klass, id)
     record = klass.classify.safe_constantize.find(id)
     results = Geocoder.search(record.full_address)
-    if results
+    if results && !results.empty?
```
misaka:
```suggestion
    if results.present?
```

Wouldn't that be the same thing?

lib/tasks/batch_geocode.rake
```diff
@@ -0,0 +1,12 @@
+desc "Batch geocode models"
+namespace :geocoding do
+  task :batch_update, [:class] => :environment do |_task, args|
+    klass = args[:class]
+    model = klass.classify.safe_constantize
+
+    puts "Starting batch geocoding for #{klass}"
+
+    ids = model.where(latitude: nil).or(model.where(longitude: nil)).pluck(:id)
+    ids.each { |id| GeocodeJob.perform_later(klass, id) }
```
misaka: This is just processing _all_ the _model_ records that need to be geocoded ... does the geocoder do it's own batching to limit how much we hit Google?

tbh, I'm a little bit uncomfortable with having this outside of `mcb` ... one of the goals of `mcb` was to contain all these kinds of tasks. It won't be hard to create a new command for this, and even if we don't implement the `-E` option at least it'd still be in `mcb`.

dankmitchell: This is a one-off though so wouldnt a rake task or migration would be more appropriate?

app/jobs/geocode_job.rb
TobyRet: 👍

TobyRet:
[Create custom batch geocoding rake task](https://trello.com/c/SyppX7Hm/2647-create-custom-batch-geocoding-rake-task)

lib/mcb/commands/geocode.rb
```diff
@@ -0,0 +1,37 @@
...
+    raise "You can only geocode Sites and Providers"
+  end
+
+  geocode_record = lambda { |obj|
+    obj.geocode
+    obj.save
+    sleep(opts[:sleep])
+  }
+
+  require "geocoder"
```
defong: Middle of nowhere?

TobyRet: @defong - my understanding (which may be incorrect) is that any external libraries used within MCB should be required just before use (rather than at the top of the file) in order to optimise overall MCB load time. @misaka can you confirm?

lib/mcb/commands/geocode.rb
```diff
@@ -0,0 +1,40 @@
+name "geocode"
```
TobyRet: @misaka - I accept potentially this is going against the grain slightly, i.e. putting the entire geocode command within the commands` directory. This was to prevent code duplication i.e.
`bin/mcb geocode -c "Site"` or `bin/mcb geocode -c "Provider"`

rather than:

`bin/mcb sites geocode` or `bin/mcb providers geocode`

... but let me know if you prefer another way of doing this.

misaka: Ya, I think that if we really wanted to avoid code duplication, it should be a service. But here, the code that would be duplicated is so small, and then this method is inflated a bit to handle the class abstraction, I really wonder if it's worth it.

lib/mcb/commands/geocode.rb
```diff
@@ -0,0 +1,41 @@
...
+  unless %w(provider site).include? opts[:model].downcase
+    raise "You can only geocode Sites and Providers"
+  end
+
+  geocode_record = lambda { |obj|
+    obj.geocode
+    obj.save
+    verbose "Geocoded #{obj.class} Id:#{obj.id} - #{obj}. New lat/long #{obj.latitude},#{obj.latitude} for full_address '#{obj.full_address}'"
+    sleep(opts[:sleep])
+  }
```
misaka: I found this a bit more complicated than it needs to be, I think. Why not just a function?

lib/mcb/commands/geocode.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  model = opts[:model].classify.safe_constantize
+
+  require "geocoder"
+
+  if args.any?
+    args.each do |id|
+      record = model.find(id)
+      geocode_record.call(record)
```
misaka: Should we batch this one too? Just thinking if we ever want to pipe a bunch site and/or provider codes into `xargs`.

lib/mcb/commands/geocode.rb
misaka: Ya, that's right. Although, I think it could go at the top of the `run` method here.

lib/mcb/commands/geocode.rb
TobyRet: 👍

TobyRet:
> course

`mcb site geocode` not `mcb course geocode`

PR #1048, opened 2019-12-10
[2659] - Refactor MCB geocode command

TobyRet:
### Context
Refactoring of the MCB geocoding command

- Define separate MCB geocoding commands for providers and sites
- Extract common geocoding functionality to `lib/mcb`
- Allow geocoding by id(s) to accept `batch_size` command line option

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[Refactor mcb geocoding command](https://trello.com/c/NM3UmBCF/2659-refactor-mcb-geocoding-command)

lib/mcb/commands/providers/geocode.rb
```diff
@@ -0,0 +1,23 @@
...
+
+instance_eval(&MCB.remote_connect_options)
+
+require "geocoder"
+
+run do |opts, args, _cmd|
+  MCB.init_rails(opts)
+
+  if args.any?
+    Provider.where(id: args.to_a).find_each(batch_size: opts[:batch_size]) do |provider|
```
misaka: Should this be `provider_code`? I'd think that that's more convenient probably. Also, it'd be smart to use `recruitment_cycle` here. Have a look at how [the `provider list` commands does this](https://github.com/DFE-Digital/manage-courses-backend/blob/8adac1ac5fad452ed91f59a9120702042a73d722/lib/mcb/commands/providers/list.rb#L9), which relies on a standard options being defined in [the `providers` command](https://github.com/DFE-Digital/manage-courses-backend/blob/8adac1ac5fad452ed91f59a9120702042a73d722/lib/mcb/commands/providers.rb#L5).

spec/lib/mcb/commands/sites_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "spec_helper"
+require "mcb_helper"
+
+describe "mcb sites" do
+  it "configures the database", mcb_cli: true do
+    allow(MCB).to receive(:init_rails)
+
+    with_stubbed_stdout do
+      $mcb.run(%w[providers --webapp=banana])
```
defong: :banana:

lib/mcb/commands/providers/geocode.rb
defong: good call

TobyRet: @misaka - yep use of provider code make sense but I'm not sure I understand the rationale/need for using `recruitment_cycle` in this instance? i.e. under what circumstances would I want geocode Providers in one particular recruitment cycle but not the others?

lib/mcb/commands/sites/geocode.rb
```diff
@@ -0,0 +1,31 @@
...
+summary "Batch geocode Sites"
+usage "geocode [options] [<id>...]"
+option :s, "sleep", "time to sleep between each request", argument: :optional, default: 0.25, transform: method(:Float)
+option :b, "batch_size", "batch size", argument: :optional, default: 100, transform: method(:Integer)
+
+instance_eval(&MCB.remote_connect_options)
+
+require "geocoder"
+
+# TODO: Once Site codes have been changed to three digit values
```
TobyRet: Currently site codes are one character long and we have 17000 sites in prod, hence lots of duplicate codes.

PR #1051, opened 2019-12-11
[2667] - BUGFIX: Update lat and long when Site and Provider address is updated

TobyRet:
### Context
When a Site or Provider address was updated the lat and long was not being updated.

This was because:
- Geolocation was not running under these conditions (we should be looking at what changes _were_ saved to the db)
- Bad test setup that gave a false positive 😿 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/concerns/geolocation.rb
+++ b/app/models/concerns/geolocation.rb
@@ -14,7 +14,11 @@ module Geolocation
     end
 
     def address_changed?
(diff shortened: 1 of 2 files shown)


TobyRet:
[BUG: Site and Provider lat / long not updating when address changes](https://trello.com/c/g9gjrWc7/2667-bug-site-and-provider-lat-long-not-updating-when-address-changes)

PR #1052, opened 2019-12-11
[2655] Allow Further Education courses to be created in new course flow

danielburnley:
### Context

FE courses could not currently be created as they would fail validation based on fields they don't require. To solve this - we need to change validation

- No longer require maths, english, or age range on a FE course
- Set the funding type to "Fee" in build new course/create
- Set the FE Subject on the course in build new course/create

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -154,8 +154,8 @@ class Course < ApplicationRecord
     ENTRY_REQUIREMENT_OPTIONS.reject { |option| option == :not_set }.keys.map(&:to_s)
   end
 
-  validates :maths,   inclusion: { in: entry_requirement_options_without_nil_choice }
-  validates :english, inclusion: { in: entry_requirement_options_without_nil_choice }
+  validates :maths,   inclusion: { in: entry_requirement_options_without_nil_choice }, unless: :further_education_course?
+  validates :english, inclusion: { in: entry_requirement_options_without_nil_choice }, unless: :further_education_course?
   validates :science, inclusion: { in: entry_requirement_options_without_nil_choice }, if: :gcse_science_required?
   validates :enrichments, presence: true, on: :publish
   validates :is_send, inclusion: { in: [true, false] }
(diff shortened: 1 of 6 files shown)


app/models/subjects/further_education_subject.rb
```diff
@@ -13,4 +13,7 @@
 #
 
 class FurtherEducationSubject < Subject
+  def self.find
```
defong: Shouldn't it be in the same vein as 
https://github.com/DFE-Digital/manage-courses-backend/blob/13949f0b130f2213a5b5cfbb3a944c0b21b86d0e/app/models/subjects/secondary_subject.rb#L16-L18

it seems rather odd to do 

``` ruby
subject = FurtherEducationSubject.find
```

danielburnley: Further education is it's own subject - there's only one. I went with this because `FurtherEducation.further_education` looked a bit weird

I thought about doing `Subjects.further_education` on the `Subject` model, IMO `FurtherEducation` should have all the information about `FurtherEducation` subjects, so splitting it into two places seems wrong

misaka: How about `instance`?

misaka: Doing a quick search through the Ruby docs to see if this is a reserved keyword (it's not) turns up some prior art of doing it this way.

danielburnley: :+1: sounds good to me

app/models/course.rb
```diff
@@ -174,7 +174,8 @@ def self.entry_requirement_options_without_nil_choice
   validate :validate_subject_consistency
 
   validates :name, :profpost_flag, :program_type, :qualification, :start_date, :study_mode, presence: true
-  validates :level, :age_range_in_years, presence: true, on: :create
+  validates :age_range_in_years, presence: true, unless: :further_education_course?
```
misaka: Is dropping the `on: :create` intentional here?

misaka: And if so, why was it there and why isn't it on `level` below. Sorry to nit-pick, but I'm just hoping you've thought about this enough to explain the change. :)

danielburnley: Good spot :grimacing: I will change that now

It was explicitly on create before because we have several courses that don't have it, but we want it to be added to all future courses

PR #1053, opened 2019-12-12
[2616] Replace govuk-lint with rubocop-govuk

faissalMT:
### Context
govuk-lint is deprecated software, we should instead prefer rubocop-govuk.

Replace `govuk-lint` with `rubocop-govuk`. This also fixes any immediate linting issues & uses a hacky fix to get around the autoloading bug thing.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [ ] Cleaned commit history


faissalMT:
[Replace govuk-lint with rubocop-govuk](https://trello.com/c/NUkLfQk4/2616-replace-govuk-lint-with-rubocop-govuk)

app/models/provider.rb
```diff
@@ -40,6 +40,9 @@
 #  index_provider_on_recruitment_cycle_id_and_provider_code  (recruitment_cycle_id,provider_code) UNIQUE
 #
 
+# Unpleasant hack to stop autoload error on CI
+require_relative "../services/providers/generate_course_code_service"
```
dankmitchell: Really not sure about this. How is it working now?

app/models/course_enrichment.rb
```diff
@@ -21,7 +21,7 @@
 
 class CourseEnrichment < ApplicationRecord
   include TouchCourse
-  enum status: %i[draft published rolled_over withdrawn]
```
dankmitchell: Does this have any impact on the data at all? I'd rather skip linting on these rules to avoid any risk

app/models/concerns/with_qualifications.rb
```diff
@@ -39,7 +39,7 @@ module WithQualifications
     # subjects this course was tagged to.
     #
     # Defined here: https://github.com/DFE-Digital/manage-courses-api/blob/master/src/ManageCourses.Domain/Models/CourseQualification.cs
-    enum qualification: %i[qts pgce_with_qts pgde_with_qts pgce pgde]
+    enum qualification: { qts: 0, pgce_with_qts: 1, pgde_with_qts: 2, pgce: 3, pgde: 4 }
```
dankmitchell: Does this have any impact on the data at all? I'd rather skip linting on these rules to avoid any risk

misaka: I think this is a good idea, no. This should make it harder to accidentally change the value for any of these enumerated value in the db.

app/models/provider.rb
faissalMT: We don't know why it's broken on CI without this, I went over it with Misha on Friday and didn't want to spend more time on it, we've got a ticket to investigate it: https://trello.com/c/9eWQ9rg3/2686-figure-out-why-this-commit-is-failing-on-azure-but-not-locally

app/models/concerns/with_qualifications.rb
defong: it does make a REAL diff, it infer int via defering it.


https://github.com/DFE-Digital/manage-courses-backend/pull/1026#discussion_r353704193

app/models/provider.rb
misaka: Ya, we started off with replaying the steps that the Azure pipeline followed to recreate this error, but ran out of steam and figured it would be better to kick that can further down the road as tech debt. I'm ok with this hack until we get time to further investigate it.

app/models/concerns/with_qualifications.rb
faissalMT: According to [this](https://vaidehijoshi.github.io/blog/2015/02/03/when-enough-is-enough-how-to-know-when-to-use-enums/) these are equivalent, the only difference is explicit vs implicit. I should point out that this change was for the most part done automatically.

PR #1055, opened 2019-12-12
Change dev psql user to match production db dumps

timabell:
This will remove a step for onboarding new developers when restoring a
sanitized backup of production data.

The production db uses `manage_courses` as a username, and that is encoded
in the production databaes dumps, without this patch you get errors
restoring the production db dumps because the `manage_coures` role does
not exist, meaning you have notice in the scrollback and manually create
the role.

This patch will mean all devs will have to re-run `rake db:setup` and
optionally manually drop the old role.

### Context

Onboarding new dev

* rename dev psql role from `manage_courses_backend` to `manage_courses`

### Guidance to review

* Try it out locally
* :ship:

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -2,8 +2,8 @@ default: &default
   adapter: postgresql
   encoding: unicode
   pool: 5
-  username: <%= ENV['DB_USERNAME'] || 'manage_courses_backend' %>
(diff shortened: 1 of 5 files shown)


jesseyuen:
Good work Tim!

timabell:
we can do better than this by using no user at all

PR #1060, opened 2019-12-16
[2403] Remove content_status and last_published_date from provider

faissalMT:
### Context
Provider enrichments are no longer in use, these fields are leftover from them.

Remove any references to providers' `content_status` or `last_published_date` and correct any tests broken by this.

### Guidance to review
Needs to be checked by @fofr but things seem to still work.  

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -14,7 +14,6 @@
 #  discarded_at                     :datetime
 #  email                            :text
 #  id                               :integer          not null, primary key
-#  last_published_at                :datetime
 #  latitude                         :float
 #  longitude                        :float
 #  postcode                         :text
@@ -33,7 +32,6 @@
 #
 # Indexes
 #
-#  IX_provider_last_published_at                             (last_published_at)
 #  index_provider_on_changed_at                              (changed_at) UNIQUE
 #  index_provider_on_discarded_at                            (discarded_at)
 #  index_provider_on_latitude_and_longitude                  (latitude,longitude)
@@ -219,11 +217,6 @@ class Provider < ApplicationRecord
     ]
 
     attributes.slice(*attribute_names)
-  end
-
-  # NOTE: This can be removed, it should not be in use any more
-  def content_status
-    :published
   end
 
   # This reflects the fact that organisations should actually be a has_one.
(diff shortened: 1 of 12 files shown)


faissalMT:
[Investigate: Remove content_status and last_published_at from provider](https://trello.com/c/2ZS1XMxd/2403-investigate-remove-contentstatus-and-lastpublishedat-from-provider)

app/models/provider.rb
```diff
@@ -198,10 +198,12 @@ def external_contact_info
     attributes.slice(*attribute_names)
   end
 
+=begin
   # NOTE: This can be removed, it should not be in use any more
   def content_status
     :published
   end
+=end
```
misaka: REMOVE PLZ

db/migrate/20191216113352_remove_provider_last_published_at.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveProviderLastPublishedAt < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :provider, :last_published_at
```
misaka: I don't think the annotations ran, I still see [`last_published_at` listed in `provider.rb`](https://github.com/DFE-Digital/manage-courses-backend/pull/1060/files#diff-eeaf9199fba4aed486bc440604e87bf9R17).

spec/requests/api/v1/provider_spec.rb
```diff
@@ -72,7 +72,6 @@ def get_provider_codes_from_body(body)
                region_code: :london,
                accrediting_provider: "Y",
                scheme_member: "Y",
-               last_published_at: DateTime.now.utc,
```
misaka: We don't know what the effect this will have on UCAS. I'm not sure we should remove it.

spec/requests/api/v1/provider_spec.rb
```diff
@@ -129,7 +128,6 @@ def get_provider_codes_from_body(body)
                region_code: :south_west,
                accrediting_provider: "N",
                scheme_member: "N",
-               last_published_at: DateTime.now.utc,
```
misaka: We don't know what the effect this will have on UCAS. I'm not sure we should remove it.

misaka:
We need to answer some questions about changes to the V1 API as a result of removing `last_published_at`, see the Trello ticket.

spec/requests/api/v1/provider_spec.rb
faissalMT: This was only for the test setup.

misaka:
As commented, it looks like `last_published_at` wasn't actually being sent to UCAS. Not sure why it was part of the test setup, though.

PR #1061, opened 2019-12-16
[2664] - Fix display of test results in azure portal

TobyRet:
### Context

Inspired by @timabell 's trip down the rabbit hole - https://github.com/DFE-Digital/manage-courses-frontend/pull/823

When switching over parallel tests, it ended up breaking publishing of test results in Azure. This was because:

* The xml file wasn't copied out of the container so azure DevOps couldn't find it.
* Multiple processes wrote to the same xml file, sometimes making it invalid xml.

* Show test results sooner
* Fix display of test results in azure portal

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/.azure_parallel
+++ b/.azure_parallel
@@ -1,0 +1,5 @@
+--require spec_helper
+--no-profile
+--format documentation
+--format RspecJunitFormatter
+--out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,15 +41,23 @@ steps:
     dockerHubImageName: $(imageName)
     CC_TEST_REPORTER_ID: $(ccReporterID)
 - script: |
+    set -x
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle config --local disable_exec_load true'
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake parallel:setup'
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,,--format progress --format RspecJunitFormatter --out rspec-results.xml --no-profile]"'
+    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,, -O .azure_parallel]"'
+
   displayName: 'Execute tests'
   env:
     DOCKER_OVERRIDE: $(dockerOverride)
     dockerHubUsername: $(dockerHubUsername)
     dockerHubImageName: $(imageName)
     CC_TEST_REPORTER_ID: $(ccReporterID)
+- task: PublishTestResults@2
+  condition: succeededOrFailed()
+  inputs:
+    testRunner: JUnit
+    testResultsFiles: 'rspec-output/*.xml'
+    failedTaskOnFailedTest: true
 - script: |
     $DOCKER_OVERRIDE exec -T web /bin/sh -c "bundle exec rubocop app config db lib spec --format clang"
   displayName: 'Execute linters'
@@ -135,9 +143,3 @@ steps:
   displayName: 'Publish Artifact'
   inputs:
     PathtoPublish: '$(build.artifactstagingdirectory)'
-- task: PublishTestResults@2
-  condition: succeededOrFailed()
-  inputs:
-    testRunner: JUnit
-    testResultsFiles: '*.xml'
-    failedTaskOnFailedTest: true


.rspec_parallel
```diff
@@ -0,0 +1,4 @@
+--no-profile
+--format-progress
+--format RspecJunitFormatter
+--out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml
```
misaka: Is this going to be the default locally? If so, that's going to be a bit of a PITA since it'll generate `RspecJunitFormatter`, no? Is there a way to specify a config file specific to CI?

azure-pipelines.yml
```diff
@@ -41,9 +41,14 @@ steps:
...
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle config --local disable_exec_load true'
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake parallel:setup'
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,,--format progress --format RspecJunitFormatter --out rspec-results.xml --no-profile]"'
+    $DOCKER_OVERRIDE run --name test-run-image -d $(image_name)
+    $DOCKER_OVERRIDE exec test-run-image /bin/sh -c 'bundle config --local disable_exec_load true'
+    $DOCKER_OVERRIDE exec test-run-image /bin/sh -c 'bundle exec rake parallel:setup'
+    $DOCKER_OVERRIDE exec test-run-image /bin/sh -c 'bundle exec rake parallel:spec'
+    $DOCKER_OVERRIDE cp test-run-image:/app/rspec-output .
+    $DOCKER_OVERRIDE stop test-run-image
+    $DOCKER_OVERRIDE rm test-run-image
```
misaka: :100:

.rspec_parallel
TobyRet: I guess one way could be to add the output to `.gitignore`
On previous projects we would have a `CI` environment so that we could create config specifically for this. 
@rizzkhan1 - do you have any ideas?

misaka: What I was worried about is that when we run `rspec` in parallel locally, it'll output in `RspecJunitFormatter` and not in the usual format we're used to. Is that not the case?

Also, ya, we'd need to ignore the output file to make sure it doesn't get checked in.

defong:
[Fix test output on Manage Courses Front backend/frontend](https://trello.com/c/aH6HXjAX/2664-fix-test-output-on-manage-courses-front-backend-frontend)

.rspec_parallel
misaka: Ok, looks like the default `progress` format is enabled in this file, I missed that. However, I'm not sure that configuring it here is the right thing to do. This output is only needed for CI, so why are we enabling it everywhere? Shouldn't we just specify the `--format` options in the Azure Pipeline steps, since this is specific to Azure?

misaka: I say we rename this file so that it doesn't get used by default locally, and also change `--format progress` to `--format documentation`. That would bring us closer to what we had before (it used to cat out the XML file).

.rspec_parallel_azure
```diff
@@ -0,0 +1,4 @@
+--no-profile
+--format progress
```
misaka: Sorry, but can we just tweak this while we're here:

```suggestion
--format documentation
```

azure-pipelines.yml
```diff
@@ -41,15 +41,25 @@ steps:
     dockerHubImageName: $(imageName)
     CC_TEST_REPORTER_ID: $(ccReporterID)
 - script: |
+    set -x
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle config --local disable_exec_load true'
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake parallel:setup'
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,,--format progress --format RspecJunitFormatter --out rspec-results.xml --no-profile]"'
+    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,, -O .rspec_parallel_azure]"'
```
misaka: I feel like we're missing something here. If `rspec` gets an error, it should return an error, but we still need to copy out `rspec-output` to be able to see the errors, right? At the moment we've seen that this build job doesn't fail even if there are errors, so I think what we need to do is:

```
    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,, -O .rspec_parallel_azure]"'
    test_result=$?
    docker cp s_web_1:/app/rspec-output .
    [ "$test_result" -eq "0" ] && true || false
```

misaka: Also how is this working in mcfe ... ?

.foo
```diff
@@ -0,0 +1,4 @@
+--no-profile
```
misaka: FOO ME?? NO! FOO U!!

.rspec
```diff
@@ -1 +1,4 @@
 --require spec_helper
+--format progress
+--format documentation
```
misaka: FTFY

```
--format documentation
--format RspecJunitFormatter
```

azure-pipelines.yml
```diff
@@ -41,15 +41,26 @@ steps:
     dockerHubImageName: $(imageName)
     CC_TEST_REPORTER_ID: $(ccReporterID)
 - script: |
+    set -x
+    set e
```
misaka: I meant `set -e`, but we don't want this now since we need to ensure that the `docker cp` always happens. `set -e` would stop the script from running as soon as `parallel:spec` returned an error.

.foo
TobyRet: 😆

azure-pipelines.yml
TobyRet: Currently it's using a `.rspec_parallel` config file. I'll have a look to see what is currently happening with failing tests.

azure-pipelines.yml
```diff
@@ -41,15 +41,30 @@ steps:
     dockerHubImageName: $(imageName)
     CC_TEST_REPORTER_ID: $(ccReporterID)
 - script: |
+    set -x
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle config --local disable_exec_load true'
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake parallel:setup'
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,,--format progress --format RspecJunitFormatter --out rspec-results.xml --no-profile]"'
+    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'echo "--no-profile
```
TobyRet: Not ideal but at least now this not require changing local config ...

azure-pipelines.yml
```diff
@@ -41,15 +41,30 @@ steps:
...
+    --format documentation
+    --format RspecJunitFormatter
+    --out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml" >> app/.rspec'
+    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake parallel:spec'
+    test_result=$?
+        docker cp s_web_1:/app/rspec-output .
```
defong: `s_web_1` is that intentional?

azure-pipelines.yml
```diff
@@ -41,15 +41,30 @@ steps:
     dockerHubImageName: $(imageName)
     CC_TEST_REPORTER_ID: $(ccReporterID)
 - script: |
+    set -x
```
defong: adding that kinda feels like this whole script block should be in a bash file

azure-pipelines.yml
TobyRet: Yep that is currently the name of the container

azure-pipelines.yml
TobyRet: Yeah possibly - if we were to do this we would then need to move other qualifying script blocks in the azure-pipeline.yml to script blocks to. 

Personally I think it may make it more readable if it's done inline. 
@misaka @rizzkhan1 what do you think?

TobyRet:
£$%^c just seen this. Looking into it
https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=35853&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=b24cf96b-e476-5cd6-97ae-776d6935d22f&l=87

azure-pipelines.yml
misaka: I don't understand. These are essentially shoved into a bash script by the pipeline processor, are you saying we should have a separate bash script for each job here?

misaka: I think we can clean this up by re-adding the `.rspec_parallel` file, maybe as `.rspec_parallel_azure`, that would get rid of 50% of the lines here I think.

azure-pipelines.yml
misaka: Might be worth re-aligning the lines here, I don't think it needs to be indented one more level.

azure-pipelines.yml
```diff
@@ -41,15 +41,26 @@ steps:
...
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle config --local disable_exec_load true'
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake parallel:setup'
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,,--format progress --format RspecJunitFormatter --out rspec-results.xml --no-profile]"'
+    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,, -O .azure_parallel]"'
+    test_result=$?
+        docker cp s_web_1:/app/rspec-output .
```
rizzkhan1: Don't think this is necessary. Output result files should be on the hosted agent  at this point already

.azure_parallel
```diff
@@ -0,0 +1,5 @@
+--require spec_helper
+--no-profile
+--format documentation
+--format RspecJunitFormatter
+--out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml
```
misaka: Give us a newline?

TobyRet: Done 👍

azure-pipelines.yml
misaka: I had my doubts about this, @rizzkhan1, but I think you're right. I did a bit of digging and it looks like that's a result of the [`volume` configuration in `docker-compose.yml`](https://github.com/DFE-Digital/manage-courses-backend/blob/master/docker-compose.yml#L17-L18). The [Docker documentation](http://docs.docker.com/engine/admin/volumes/volumes) appears to support this, at a scan.

misaka: @TobyRet That means we can get rid of all this `test_result` stuff too, assuming it fully works.

azure-pipelines.yml
```diff
@@ -41,15 +41,26 @@ steps:
     dockerHubImageName: $(imageName)
     CC_TEST_REPORTER_ID: $(ccReporterID)
 - script: |
+    set -x
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle config --local disable_exec_load true'
     $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake parallel:setup'
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,,--format progress --format RspecJunitFormatter --out rspec-results.xml --no-profile]"'
+    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec[,, -O .azure_parallel]"'
```
misaka: I noticed that this was originally `SPEC_OPTS="..."` ... changed in [672c20f](https://github.com/DFE-Digital/manage-courses-backend/commit/672c20fd0ef5ccb1a3df0762b81ab86471eab9e3#diff-fec826feae04e51c0d94076385408bdc). Does it make sense to return it to that more readable form? Unless you're just sick of all this ... :)

TobyRet: It appears to be a no-go because `TEST_ENV_NUMBER` needs to be accessed via Rails.env (`ENV`)

i.e. `TEST_ENV_NUMBER` is not defined when the code executed like so ...
```
bundle exec rake parallel:spec SPEC_OPTS="--format RspecJunitFormatter --out rspec-output/rspec-output${TEST_ENV_NUMBER}.xml"
```

Interpolation not poss either

```
 manage-courses-backend 2664-fix-test-output-on-manage-courses-back-end % bundle exec rake "parallel:spec[,, --format RspecJunitFormatter --format documentation --out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml]"
12 processes for 205 specs, ~ 17 specs per process
sh: %=: No such file or directory
sh: %=: No such file or directory
sh: %=: No such file or directory
```

azure-pipelines.yml
TobyRet: Yep looks good to me

azure-pipelines.yml
misaka: Ah. I was just thinking that this is a little bit easy to read:

```suggestion
    $DOCKER_OVERRIDE exec -T web /bin/sh -c 'bundle exec rake "parallel:spec" SPEC_OPTS="-O .azure_parallel"' 
```

No worries, though, this PR has been through enough. :)

TobyRet:
🎉

PR #1062, opened 2019-12-16
[2640] Add validations for course creation

danielburnley:
### Context

The `build_new` endpoint for course creation needs additional validations - this is to enable us to ensure we have correct data being created.

- Add a new validation scope for `:new`
  - The reason for the additional scope was that currently a lot of tests fail when adding some of them to `on: :create` - this could be mitigated in the future but I wanted to minimise the changes in this PR
- Add new validations to courses 

### Guidance to review

- Check out `2640-create-a-course-validation-changes` on MCFE (or master if it is merged)
- Run through the course creation for an accredited body and attempt to submit without any data
  - **Note** the Frontend for age ranges with a custom age range is not handled yet as this needs additional work to remove the validation from the frontend

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -21,12 +21,7 @@ module API
 
       def build_new
         authorize @provider
-        @course = Course.new(provider: @provider)
-        update_subjects
-        update_sites
-        @course.assign_attributes(course_params)
-        update_further_education_fields if @course.level == "further_education"
-        @course.name = @course.generate_name
+        build_new_course
         @course.valid?(:new)
 
         # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
@@ -126,13 +121,9 @@ module API
         authorize @provider, :can_create_course?
         return unless course_params.values.any?
 
+        build_new_course
         course_code = @provider.next_available_course_code
-        @course = Course.new(provider: @provider)
-        @course.assign_attributes(course_params.merge(course_code: course_code))
-        update_subjects
-        update_sites
-        update_further_education_fields if @course.level == "further_education"
-        @course.name = @course.generate_name
+        @course.assign_attributes(course_code: course_code)
 
         if @course.valid?(:new) && @course.save
           render jsonapi: @course.reload
@@ -194,6 +185,15 @@ module API
         @course = @provider.courses.find_by!(course_code: params[:code].upcase)
 
         authorize @course
+      end
+
+      def build_new_course
+        @course = Course.new(provider: @provider)
+        @course.assign_attributes(course_params)
+        update_subjects
+        update_sites
+        update_further_education_fields if @course.level == "further_education"
+        @course.name = @course.generate_name
       end
 
       def enrichment_params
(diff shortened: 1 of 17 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -125,15 +121,11 @@ def create
...
-        @course = Course.new(provider: @provider)
-        @course.assign_attributes(course_params.merge(course_code: course_code))
-        update_subjects
-        update_sites
-        update_further_education_fields if @course.level == "further_education"
-        @course.name = @course.generate_name
+        @course.assign_attributes(course_code: course_code)
 
-        if @course.save
+        if @course.valid?(:new) && @course.save
```
faissalMT: Shouldn't `.save()` also do validation?

app/models/course.rb
```diff
@@ -159,23 +159,29 @@ def self.entry_requirement_options_without_nil_choice
...
+  validate :validate_qualification, on: %i[update new]
   validate :validate_start_date, on: :update, if: -> { provider.present? && start_date.present? }
-  validate :validate_applications_open_from, on: :update, if: -> { provider.present? }
+  validate :validate_applications_open_from, on: %i[update new], if: -> { provider.present? }
   validate :validate_modern_languages
+  validate :validate_languages_count, if: :has_the_modern_languages_secondary_subject_type?
   validate :validate_subject_count
   validate :validate_subject_consistency
 
+  def validate_languages_count
```
faissalMT: Maybe `validate_has_languages`?

app/models/course.rb
```diff
@@ -540,7 +546,11 @@ def validate_course_syncable
   end
 
   def validate_qualification
-    errors.add(:qualification, "^#{qualifications_description} is not valid for a #{level.to_s.humanize.downcase} course") unless qualification_options.include?(qualification)
+    if qualification.blank?
+      errors.add(:qualification, :blank)
+    else
+      errors.add(:qualification, "^#{qualifications_description} is not valid for a #{level.to_s.humanize.downcase} course") unless qualification_options.include?(qualification)
```
```suggestion
      errors.add(:qualification, "^#{qualifications_description} is not valid for a #{level.to_s.humanize.downcase} course") unless qualification.in?(qualification_options)
```

spec/requests/api/v2/build_new_course_spec.rb
```diff
@@ -179,21 +200,21 @@
               },
               {
                 "title" => "Invalid study_mode",
-                "detail" => "Study mode can't be blank",
+                "detail" => "You need to pick an option",
```
faissalMT: Should these not be 'you need to pick a[n] [thing]' like we have with age range and level?

danielburnley: Taken from the content that Paul G wrote - he went with "An option" here

app/controllers/api/v2/courses_controller.rb
danielburnley: It _does_ - but not with the scope `:new`, only with the scope `:create`

PR #1065, opened 2019-12-17
[2697] Include relations that are being lazy loaded to speed up query

faissalMT:
### Context
The organisations index page is extremely slow, to the point where it times out when using prod data.

Reduce the number of queries by ~80%, and speed things up by a factor of ~100.

### Guidance to review
It's still too slow to load the prod data before the server times out.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -14,7 +14,6 @@
 #  discarded_at                     :datetime
 #  email                            :text
 #  id                               :integer          not null, primary key
-#  last_published_at                :datetime
 #  latitude                         :float
 #  longitude                        :float
 #  postcode                         :text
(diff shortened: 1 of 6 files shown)


faissalMT:
[Optimise load time of admin org#index](https://trello.com/c/0H8FRZUT/2697-optimise-load-time-of-admin-orgindex)

app/models/provider.rb
```diff
@@ -14,7 +14,6 @@
 #  discarded_at                     :datetime
 #  email                            :text
 #  id                               :integer          not null, primary key
-#  last_published_at                :datetime
```
misaka: Shouldn't be removing these in this PR.

faissalMT:
Closed because this *solution* has a multitude of issues

PR #1068, opened 2019-12-18
[2702] - Make geocoding more accurate

TobyRet:
### Context
We discovered that there are discrepancies between lat / longs between Publish course previews and Find.

- Providers now have `provider_name` included in geocoding
- Sites now have `location_name` included in geocoding (only if it does not equal "Main site", "Main Site" or "main site")
- This will improve the accuracy of lat / long coordinates
- Unfortunately this now means we have to remove the geocoding concern (due to the fact that Providers have a `provider_name` attribute and Sites have a `location_name` attribute) and move this code back into the Provider and Site models. _Boo. Hiss. Boo_.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[Add 'site_name' and 'provider_name' to geocoding](https://trello.com/c/9BqTmRyR/2702-add-sitename-and-providername-to-geocoding)

app/models/site.rb
```diff
@@ -49,8 +50,35 @@ class Site < ApplicationRecord
                    inclusion: { in: POSSIBLE_CODES, message: "must be A-Z, 0-9 or -" },
                    presence: true
 
+  geocoded_by :full_address
+
   after_commit -> { GeocodeJob.perform_later("Site", id) }, if: :needs_geolocation?
 
+  def needs_geolocation?
+    full_address.present? && (
+    latitude.nil? || longitude.nil? || address_changed?
```
defong: linter issue?

spec/models/provider_spec.rb
```diff
@@ -571,4 +571,81 @@
       it { should_not include last_years_course }
     end
   end
+
+  describe "geolocation" do
```
defong: any chance to move this whole block to its own file
` spec/models/provider/geolocation.rb`

this `provider_spec.rb` is growing

app/models/site.rb
```diff
@@ -49,8 +50,35 @@ class Site < ApplicationRecord
...
+    )
+  end
+
+  def full_address
+    address = [address1, address2, address3, address4, postcode]
+
+    unless location_name == MAIN_SITE
```
defong: image

app/models/site.rb
TobyRet: Don't think so? Rubocop seems ok locally

spec/models/provider_spec.rb
TobyRet: Yeah unfortunately we have had to get rid of the geolocation model concern therefore the geolocation specs had to move back into the model specs

app/models/site.rb
TobyRet: Yep fixed this now

PR #1071, opened 2019-12-18
[2704] Add validation for custom age ranges

davidgisbey:
### Context

We need a validation on the course that checks that custom age ranges are valid for the new and create actions.

It must do three things:
1. Check if the age ranges are 4 or more years apart
2. Check that the from value is valid (so in 3_to_7 that 3 is an integer)
3. Check that the to value is valid (so in 3_to_7 that 7 is an integer)

- Add a validation that is triggered on the new and create course endpoints that does the above.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -642,13 +642,18 @@ private
   end
 
   def validate_custom_age_range
-    age_range_array = age_range_in_years.split("_")
-    if age_range_array.first.to_i.zero?
-      errors.add(:age_range_in_years_from, "#{age_range_in_years} is invalid. A valid from value must be provided.")
-    elsif age_range_array.last.to_i.zero?
-      errors.add(:age_range_in_years_to, "#{age_range_in_years} is invalid. A valid to value must be provided.")
+    age_range_array = age_range_in_years.split("_to_")
+    error_message = "#{age_range_in_years} is invalid. You must enter a valid age range."
+    if age_range_in_years.length >= 9
+      errors.add(:age_range_in_years, error_message)
+    elsif age_range_array.length != 2
+      errors.add(:age_range_in_years, error_message)
+    elsif age_range_array.first.to_i.zero? || age_range_array.first.to_i > 14 || age_range_array.first.to_i < 3
+      errors.add(:age_range_in_years_from, error_message)
+    elsif age_range_array.last.to_i.zero? || age_range_array.last.to_i < 7 || age_range_array.last.to_i > 18
+      errors.add(:age_range_in_years_to, error_message)
     elsif age_range_array.last.to_i - age_range_array.first.to_i < 4
-      errors.add(:age_range_in_years, "#{age_range_in_years} is an invalid age range. A valid age range must cover 4 or more years.")
+      errors.add(:age_range_in_years, "#{age_range_in_years} is invalid. Your age range must cover 4 years.")
     end
   end
 
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -642,18 +642,22 @@ private
   end
 
   def validate_custom_age_range
-    age_range_array = age_range_in_years.split("_to_")
+    # This regex checks that a 1 or 2 digit number is selected for the 'to' and 'from' values.
+    # It also checks that the age range is correctly formatted with _to_ in between these values
+    valid_regex_pattern = /^(\d{1,2})_to_(\d{1,2})$/
+    #The below grabs the 'from' and 'to' ages and puts them into variables if the regex is valid
+    from_age, to_age = valid_regex_pattern.match(age_range_in_years).captures if valid_regex_pattern.match(age_range_in_years)
+
     error_message = "#{age_range_in_years} is invalid. You must enter a valid age range."
-    if age_range_in_years.length >= 9
+
+    if !valid_regex_pattern.match(age_range_in_years)
       errors.add(:age_range_in_years, error_message)
-    elsif age_range_array.length != 2
-      errors.add(:age_range_in_years, error_message)
-    elsif age_range_array.first.to_i.zero? || age_range_array.first.to_i > 14 || age_range_array.first.to_i < 3
+    elsif from_age.to_i < 3 || from_age.to_i > 14
       errors.add(:age_range_in_years_from, error_message)
-    elsif age_range_array.last.to_i.zero? || age_range_array.last.to_i < 7 || age_range_array.last.to_i > 18
+    elsif to_age.to_i < 7 || to_age.to_i > 18
       errors.add(:age_range_in_years_to, error_message)
-    elsif age_range_array.last.to_i - age_range_array.first.to_i < 4
-      errors.add(:age_range_in_years, "#{age_range_in_years} is invalid. Your age range must cover 4 years.")
+    elsif to_age.to_i - from_age.to_i < 4
+      errors.add(:age_range_in_years, "#{age_range_in_years} is invalid. Your age range must cover at least 4 years.")
     end
   end
 
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -172,6 +172,7 @@ class Course < ApplicationRecord
   validate :validate_has_languages, if: :has_the_modern_languages_secondary_subject_type?
   validate :validate_subject_count
   validate :validate_subject_consistency
+  validate :validate_custom_age_range, on: %i[create new], if: -> { age_range_in_years.present? }
 
   validates :name, :profpost_flag, :program_type, :qualification, :start_date, :study_mode, presence: true
   validates :age_range_in_years, presence: true, on: %i[new create], unless: :further_education_course?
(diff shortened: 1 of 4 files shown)


davidgisbey:
[Add validation for custom age range for new courses (MCBE)](https://trello.com/c/6zrT29GB/2704-add-validation-for-custom-age-range-for-new-courses-mcbe)

danielburnley:
It'd be good to confirm with Paul G what the content should be for the error messages so we don't need to do work after this to correct them :tada:

davidgisbey:
Have checked with Paul on the wording of the error messages. He signed off on the updated values.

app/models/course.rb
```diff
@@ -619,6 +620,34 @@ def validate_subject_consistency
     end
   end
 
+  def validate_custom_age_range
+    # This regex checks that a 1 or 2 digit number is selected for the 'to' and 'from' values.
+    # It also checks that the age range is correctly formatted with _to_ in between these values
+    valid_regex_pattern = /^(\d{1,2})_to_(\d{1,2})$/
+    #The below grabs the 'from' and 'to' ages and puts them into variables if the regex is valid
+    from_age, to_age = valid_regex_pattern.match(age_range_in_years).captures if valid_regex_pattern.match(age_range_in_years)
```
```suggestion
    from_age, to_age = valid_regex_pattern.match(age_range_in_years).captures if valid_regex_pattern.match?(age_range_in_years)
```

app/models/course.rb
```diff
@@ -619,6 +620,34 @@ def validate_subject_consistency
     end
   end
 
+  def validate_custom_age_range
+    # This regex checks that a 1 or 2 digit number is selected for the 'to' and 'from' values.
+    # It also checks that the age range is correctly formatted with _to_ in between these values
+    valid_regex_pattern = /^(\d{1,2})_to_(\d{1,2})$/
```
faissalMT: Regex can be hard to parse, consider splitting it into chunks? Like `/^#{two_digit_number}_to_#{two_digit_number}$/`

danielburnley: I'm not a great fan of regexs that require a comment :grimacing: if there isn't a simpler comment at the very least i think this should be extracted out into a method something along the lines of `get_age_range` which returns a hash something like `{from: a, to: b}`

Otherwise when I read this method i'm immediately greeted by a comment and a regex that not everyone will understand - and it takes away from the actual validation logic

app/models/course.rb
davidgisbey: Have fixed this.

app/models/course.rb
davidgisbey: I think it's a pretty straightforward regex. I just have not really used them much, so I mainly left it for people who don't know how they work (we have new juniors starting soon). I really think a regex is correct here due to the fact that our age range in years needs to follow a very specific format. 

Hopefully, I've addressed the rest of your comment though.

davidgisbey: Pretty sure you can't split a Regex like that?

app/services/courses/validate_custom_age_range_service.rb
```diff
@@ -0,0 +1,37 @@
...
+        course.errors.add(:age_range_in_years, error_message)
+      elsif to_age - from_age < 4
+        course.errors.add(:age_range_in_years, "#{age_range_in_years} is invalid. Your age range must cover at least 4 years.")
+      end
+    end
+
+  private
+
+    def get_ages(age_range_in_years, valid_regex_pattern)
+      valid_regex_pattern.match(age_range_in_years)&.named_captures&.transform_values { |v| v.to_i }
```
faissalMT:
```suggestion
      valid_regex_pattern.match(age_range_in_years)&.named_captures&.transform_values { |year| year.to_i }
```

spec/services/courses/validate_custom_age_range_service_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  context "an invalid age range" do
+    let(:error_message) { "is invalid. You must enter a valid age range." }
+
+    context "with an age range of with a gap of less than 4 years" do
+      let(:age_range_in_years) { "5_to_8" }
+      let(:error_message) { "is invalid. Your age range must cover at least 4 years." }
+
+      it "should return an error stating valid age ranges must be 4 years or greater" do
+        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} " + error_message
```
faissalMT:
```suggestion
        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} #{error_message}"
```

spec/services/courses/validate_custom_age_range_service_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+    end
+
+    context "with a from value that does not fall within the valid age range" do
+      let(:age_range_in_years) { "1_to_15" }
+
+      it "should return an error" do
+        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} " + error_message
```
faissalMT:
```suggestion
        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} #{error_message}"
```

spec/services/courses/validate_custom_age_range_service_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+    end
+
+    context "with a to value that does not fall within the valid age range" do
+      let(:age_range_in_years) { "7_to_19" }
+
+      it "should return an error stating valid age ranges must be 4 years or greater" do
+        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} " + error_message
```
faissalMT:
```suggestion
        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} #{error_message}"
```

spec/services/courses/validate_custom_age_range_service_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+    end
+
+    context "with an age range that does not include a valid from age range value" do
+      let(:age_range_in_years) { "to_6" }
+
+      it "should return an error stating that there is an invalid from year" do
+        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} " + error_message
```
faissalMT:
```suggestion
        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} #{error_message}"
```

spec/services/courses/validate_custom_age_range_service_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+    end
+
+    context "with an age range that does not include a valid to age range value" do
+      let(:age_range_in_years) { "2_to" }
+
+      it "should return an error stating that there is an invalid from year" do
+        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} " + error_message
```
faissalMT: "#{age_range_in_years} #{error_message}"

faissalMT:
```suggestion
        expect(course.errors.messages[:age_range_in_years]).to include "#{age_range_in_years} #{error_message}"
```

app/services/courses/validate_custom_age_range_service.rb
```diff
@@ -0,0 +1,37 @@
+module Courses
+  class ValidateCustomAgeRangeService
+    def execute(age_range_in_years, course)
+      error_message = "#{age_range_in_years} is invalid. You must enter a valid age range."
+      valid_regex_pattern = Regexp.new(/^(?<from>\d{1,2})_to_(?<to>\d{1,2})$/)
```
misaka: Low priority, but I'd rename this to `valid_age_range_regex`.

app/services/courses/validate_custom_age_range_service.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      if valid_regex_pattern.match(age_range_in_years)
+        from_age = get_ages(age_range_in_years, valid_regex_pattern)["from"]
+        to_age = get_ages(age_range_in_years, valid_regex_pattern)["to"]
+      end
+
+      if !valid_regex_pattern.match(age_range_in_years)
```
misaka: Running the regex match twice feels like the wrong thing to do here. I think it should either be:

```ruby
      age_range_matches = valid_regex_pattern.match(age_range_in_years)

      if age_range_matches
        from_age = get_ages(age_range_in_years, valid_regex_pattern)["from"]
        to_age = get_ages(age_range_in_years, valid_regex_pattern)["to"]
      end

      # Handle errors separately
      if !age_range_matches
        ...
      end
```

or, because I still find handling the `!age_range_matches` this way, without an `else`, a little consfusing:

```ruby
      if valid_regex_pattern.match(age_range_in_years)
        from_age = get_ages(age_range_in_years, valid_regex_pattern)["from"]
        to_age = get_ages(age_range_in_years, valid_regex_pattern)["to"]

        if from_age_invalid?(from_age)
          course.errors.add(:age_range_in_years, error_message)
        elsif to_age_invalid?(to_age)
          course.errors.add(:age_range_in_years, error_message)
        elsif to_age - from_age < 4
          course.errors.add(:age_range_in_years, "#{age_range_in_years} is invalid. Your age range must cover at least 4 years.")
        end
      else
        course.errors.add(:age_range_in_years, error_message) 
      end
```

davidgisbey: Yea that's way clearer. Thanks for that.

spec/services/courses/validate_custom_age_range_service_spec.rb
davidgisbey: done. Thanks

PR #1072, opened 2019-12-18
Add ADR 1, to add ADRs to this repo

misaka:
### Context

We've been talking about using ADRs for some time. Here we go!

Add ADRs for manage and find to this repo.

### DIFF
--- a/docs/adr/0001-record-architecture-decisions.md
+++ b/docs/adr/0001-record-architecture-decisions.md
@@ -1,0 +1,23 @@
+# 1. Record architecture decisions in manage-courses-backend
+
+Date: 2019-12-18
+
+## Status
+
+Accepted
+
+## Context
+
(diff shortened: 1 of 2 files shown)


docs/adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,23 @@
...
+
+Date: 2019-12-18
+
+## Status
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made for the manage-courses and
```
dankmitchell: I'm really keen to get `manage` changed to `publish` but not a blocker for this

misaka: I like that idea. The idea of merging all the repos into one came up again yesterday, esp in context of this. We may not be doing that yet, but probably worth keeping in mind if we’re looking to rename.

PR #1078, opened 2019-12-19
2645- Add timestamp in container logs

rizzkhan1:
### Context

Add date timestamp in conatiner instances logs


timabell:
Maybe rebase rather than merging master in?

PR #1085, opened 2019-12-23
[2663] fix mcb bulk sync to find

misaka:
### Context

The `mcb apiv2 bulk_sync_to_find` was using the wrong token, because it was added as an `apiv2` command when it's actually in the "system" api.

Move `bulk_sync_to_find` to the right namespace (just `api` instead of `apiv2`) and add support for getting configuration for the system API.

### Guidance to review

The method `system_api_opts` is just a copy of `apiv1_opts`, but it should be refactored to pull out logic that is shared between those methods, and `apiv2_opts`. I'll do that as a separate PR to keep this one smaller.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -192,35 +192,17 @@ module MCB
     # The opts passed in are examined determine which opts need to be added,
     # this function essentially just fills in any missing options.
     #
-    #   opts = system_apiv_opts(opts)
+    #   opts = system_api_opts(opts)
     def system_api_opts(opts)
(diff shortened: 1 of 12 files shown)


misaka:
[`mcb apiv2 bulk_sync_to_find` broken](https://trello.com/c/QTnXRoTd/2663-mcb-apiv2-bulksynctofind-broken)

PR #1086, opened 2019-12-23
[2678] Add timestamps to tables without them

faissalMT:
### Context
A number of tables with being flagged by rubocop as missing timestamps, this adds timestamps to those tables.

Add timestamps to the tables `course_subject` and `subject`

### Guidance to review
These timestamps are nullable in because no data for them currently exists, an alternate solution would be to set the date to the epoch.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course_subject.rb
+++ b/app/models/course_subject.rb
@@ -3,8 +3,10 @@
 # Table name: course_subject
 #
 #  course_id  :integer
+#  created_at :datetime
 #  id         :integer          not null, primary key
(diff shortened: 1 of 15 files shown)


faissalMT:
[[S] Add timestamps to tables that do not have them](https://trello.com/c/92GFlfS5/2678-s-add-timestamps-to-tables-that-do-not-have-them)

PR #1088, opened 2019-12-23
[2677] Remove PE from potential subjects if user is not an admin

davidgisbey:
### Context

PE courses should not be able to be created by non-admin users via the UI.

- Remove PE from the list of potential subjects from the build new course endpoint.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -31,6 +31,12 @@ module API
           include: [:subjects, :sites, :accrediting_provider, :provider, provider: [:sites]],
         )
 
+        if !@current_user.admin?
+          json_data[:data][:meta][:edit_options][:subjects]&.reject! do |subject|
+            subject[:attributes][:subject_name] == "Physical education"
+          end
+        end
+
         json_data[:data][:errors] = []
 
         @course.errors.messages.each do |error_key, _|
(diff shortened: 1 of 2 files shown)


davidgisbey:
[[S] Remove PE from create a course journey](https://trello.com/c/DzJynp7k/2677-s-remove-pe-from-create-a-course-journey)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -30,6 +30,14 @@ def build_new
           class: CourseSerializersService.new.execute,
           include: [:subjects, :sites, :accrediting_provider, :provider, provider: [:sites]],
         )
+        subjects = json_data[:data][:meta][:edit_options][:subjects]
+
+
+        if !@current_user.admin? && subjects_include_pe(subjects)
```
misaka: This extra operation of checking that PE exists in the subjects with `subjects_include_pe`, is a little bit expensive (computing-wise) and I'm not sure it really adds a lot. I'd suggest just taking it out and letting the reject run since it'll be a no-op if it's not there anyway.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -30,6 +30,14 @@ def build_new
...
+          json_data[:data][:meta][:edit_options][:subjects] = subjects.reject do |subject|
```
misaka:
```suggestion
          json_data[:data][:meta][:edit_options][:subjects].reject! do |subject|
```

davidgisbey: Done. Thanks

app/controllers/api/v2/courses_controller.rb
davidgisbey: Done. Thanks

PR #1093, opened 2019-12-30
[2624] Clean up accrediting provider data

davidgisbey:
### Context

Due to rollover 62 providers are marked as accrediting providers when they do not accredit a course in the current cycle.

Since accrediting provider code is the primary key for it's accredited courses, accrediting_provider_id is redundant and should be removed.

### Update

- After discussion, it has been decided that this should be part of the provider edit MCB command which will be run on the necessary providers after the command is merged.

- Edit provider edit MCB command to edit accrediting provider attribute to yes or no
- Drop the accrediting_provider_id column for the course table

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


davidgisbey:
[Clean up providers accredited bodies data](https://trello.com/c/tpbDCpmy/2624-clean-up-providers-accredited-bodies-data)

db/migrate/20191230121237_clean_up_accrediting_provider_data.rb
```diff
@@ -0,0 +1,14 @@
+class CleanUpAccreditingProviderData < ActiveRecord::Migration[6.0]
```
TobyRet: This PR looks good overall but I'm not so sure about using migrations what are effectively 'data changes' rather than 'schema changes'. I think a temporary rake task is the best place for this logic ...

dankmitchell:
Should the accrediting provider code be the primary key? Seems like an old carry over from the c# world

davidgisbey:
I asked there before and Tim gave a reason for it. I think it was so that during rollover the new courses don't have the previous cycles provider id.

db/migrate/20191230121237_clean_up_accrediting_provider_data.rb
davidgisbey: We've previously used migrations for data changes, but i don't have any strong feelings on this. Perhaps @misaka  can weigh in.

misaka: I agree that this doesn't warrant a data migration. I think we only do data migrations when necessary, this one doesn't qualify because it doesn't have to be in lock-step with a schema change or code change.

misaka:
> Should the accrediting provider code be the primary key? Seems like an old carry over from the c# world

iirc, the issue is that when rollover happens, the old provider id would be invalid and we'd have to look it up. Also, technically, it could point to an accredited provider in the wrong recruitment cycle ... etc. All in all, it's a bit messy. I think it's ok for it to be a provider code unless it becomes more of a pain in other ways, e.g. ActiveRecord relationship setup.

misaka:
I'm not entirely sure about this change, tbh. It basically suggests that `accredited_body` reflects whether or not this provider **does provide** qualifications for courses or not, versus whether or not the provider **can provide** qualifications or not (i.e. a `scitt` or a `university`, see `Provider#provider_type=`). So why do we even need it when we can simply derive it from whether any courses list this provider as a qualification provider. Is it an optimisation?

The Trello ticket outlined a couple of options that seem more sensible to me, either `accredited_body` doesn't mean anything and we get rid of it entirely and just derive it's value, or it's value is meaningful and we start to actually use it.

misaka:
Spoke with @davidgisbey and @TobyRet about this. I'll update the ticket with a bigger-picture comment about why we're doing these changes (this is now clearer, it just wasn't documented on the ticket).

As for this change, I still don't think this should be a data migration. We are likely to need to change this on a per-provider level as we go, although we don't yet know how the GUI interface will look so I think we just whack it into `mcb` for now.

@davidgisbey Can we make it work like this in `mcb`:

```
mcb provider edit --accredited-body PROVIDER_CODE1 [PROVIDER_CODE2 ...]
```

This would use the existing code to connect to the DB and make the change directly in it. No need for an API (yet, perhaps). There should also be a way disable this, maybe `--no-accredited-body`? There should be a standard way to disable boolean options in `mcb`.

db/migrate/20191230121237_clean_up_accrediting_provider_data.rb
davidgisbey: Ok. As discussed i've edited the provider edit command 👍

lib/mcb/commands/providers/edit.rb
```diff
@@ -1,11 +1,36 @@
...
+       default: false
+
+
 
 run do |opts, args, _cmd|
   MCB.init_rails(opts)
+  recruitment_cycle = MCB.get_recruitment_cycle(opts)
+  providers = recruitment_cycle.providers
+
+  if opts[:"accrediting-provider"].present?
```
TobyRet: Looks good but [noob question] is there any particular reason why it's not part of the editor wizard?

misaka: > [noob question] is there any particular reason why updating accrediting providers is not part of the editor wizard?

@TobyRet I think we had/have a ticket for doing that, but I haven't been able to find it.

PR #1097, opened 2019-12-31
[2731] Test setup

TobyRet:
### Context

We recently undertook a spike on our test suite to identify opportunities for improve, particularly around decreasing the time it takes to run the tests.

This PR acts on one the recommendations and aims to create a standard test setup process. What this means is that each test will have access to standardised set of test data (with the ability to customise where required). This will remove a lot of the duplicated factories that are instantiated across all the tests and increase overall speed. 

This work is based on similar work undertaken for [Ministry Of Justice Correspondance Tool](https://github.com/ministryofjustice/correspondence_tool_staff/blob/master/spec/support/standard_setup.rb)

### Guidance to review

NB: This PR should be considered a 'first iteration'. Functionality has been added to create a limited set of test data. The TestSetup class has been wired up to two specs. Moving forwards we would need to decide how to migrate all specs over to the TestSetup process (possibly we could add a new PR checklist requirement and migrate as we go?)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,3 +13,7 @@ AllCops:
 
 Rails/HasManyOrHasOneDependent:
   Enabled: false
+
+Style/ClassVars:
+  Exclude:
+    - 'spec/support/test_setup.rb'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,6 +14,3 @@ AllCops:
 Rails/HasManyOrHasOneDependent:
   Enabled: false
 
-Style/ClassVars:
-  Exclude:
-    - 'spec/support/test_setup.rb'
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -40,7 +40,7 @@ FactoryBot.define do
     qualification { :pgce_with_qts }
     with_apprenticeship
 
-    provider
+    provider { find_or_create(:provider) }
 
     study_mode { :full_time }
     maths { :must_have_qualification_at_application_time }
@@ -125,11 +125,6 @@ FactoryBot.define do
       # We've just created a course with this provider's code, so ensure it's
       # up-to-date and has this course loaded.
       course.provider.reload
-
-      # Make sure this Course and and Sites are associated with the same Provider
-      if course.site_statuses.any?
-        course.site_statuses.first.site.provider = course.provider
-      end
     end
 
     trait :infer_level do
@@ -217,14 +212,17 @@ FactoryBot.define do
 
     trait :unpublished_with_primary_maths do
       transient do
-        identifier { "unpublished_with_primary_mathematics" }
+        identifier { "unpublished_with_primary_maths" }
       end
 
-      name { "#{identifier} course" }
+      name { "#{identifier} course name" }
 
       provider { find_or_create :provider, :published_scitt }
-      site_statuses { [build(:site_status, :new)] }
-      enrichments { [build(:course_enrichment, :initial_draft)] }
+      sites { build_list(:site, 1, provider: provider) }
+    end
+
+    trait :draft_enrichment do
+      enrichments { [build(:course_enrichment, :initial_draft, course: nil)] }
     end
   end
 end
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -125,6 +125,11 @@ FactoryBot.define do
       # We've just created a course with this provider's code, so ensure it's
       # up-to-date and has this course loaded.
       course.provider.reload
+
+      # Make sure this Course and and Sites are associated with the same Provider
+      if course.site_statuses.any?
+        course.site_statuses.first.site.provider = course.provider
+      end
     end
 
     trait :infer_level do
@@ -208,6 +213,18 @@ FactoryBot.define do
 
     trait :skip_validate do
       to_create { |instance| instance.save(validate: false) }
+    end
+
+    trait :unpublished_with_primary_maths do
+      transient do
+        identifier { "unpublished_with_primary_mathematics" }
+      end
+
+      name { "#{identifier} course" }
+
+      provider { find_or_create :provider, :published_scitt }
+      site_statuses { [build(:site_status, :new)] }
+      enrichments { [build(:course_enrichment, :initial_draft)] }
     end
   end
 end
--- a/spec/factories/organisations.rb
+++ b/spec/factories/organisations.rb
@@ -17,7 +17,7 @@ FactoryBot.define do
     sequence(:org_id)
 
     trait :with_user do
-      users { [find_or_create(:user)] }
+      users { [create(:user)] }
     end
   end
 end
--- a/spec/factories/organisations.rb
+++ b/spec/factories/organisations.rb
@@ -15,5 +15,9 @@ FactoryBot.define do
   factory :organisation do
     name { "LONDON SCITT" + rand(1000000).to_s }
     sequence(:org_id)
+
+    trait :with_user do
+      users { [find_or_create(:user)] }
+    end
   end
 end
--- a/spec/factories/providers.rb
+++ b/spec/factories/providers.rb
@@ -53,7 +53,7 @@ FactoryBot.define do
     website { Faker::Internet.url }
     accrediting_provider { "N" }
     region_code { "london" }
-    organisations { build_list :organisation, 1 }
+    organisations { [find_or_create(:organisation, :with_user)] }
     association :recruitment_cycle, strategy: :find_or_create
 
     train_with_us { Faker::Lorem.sentence.to_s }
@@ -89,6 +89,14 @@ FactoryBot.define do
 
     trait :previous_recruitment_cycle do
       recruitment_cycle { find_or_create :recruitment_cycle, :previous }
+    end
+
+    trait :published_scitt do
+      transient do
+        identifier { "published_scitt" }
+      end
+
+      provider_name { "#{identifier} provider" }
     end
   end
 end
--- a/spec/factories/providers.rb
+++ b/spec/factories/providers.rb
@@ -96,7 +96,7 @@ FactoryBot.define do
         identifier { "published_scitt" }
       end
 
-      provider_name { "#{identifier} provider" }
+      provider_name { "#{identifier} provider name" }
     end
   end
 end
--- a/spec/factory_specs/course_spec.rb
+++ b/spec/factory_specs/course_spec.rb
@@ -30,4 +30,20 @@ describe "Course factory" do
       expect(CourseEnrichment.where("created_at < ?", 4.days.ago).count).to eq(1)
     end
   end
+
+  context "unpublished_with_primary_maths" do
+    subject { create(:course, :unpublished_with_primary_maths) }
+
+    its(:name) { should eq("unpublished_with_primary_maths course name") }
+
+    it "has the correct number of associations" do
+      expect { subject }.to change { Site.count }.by(1).and change { Provider.count }.by(1)
+    end
+
+    it "has the correct association" do
+      provider_id = subject.provider.id
+
+      expect(subject.sites.first.provider_id == provider_id)
+    end
+  end
 end
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -268,14 +268,10 @@ describe Provider, type: :model do
   end
 
   describe "courses" do
-    let(:provider) { create(:provider, courses: [course]) }
-    let(:course) { build(:course) }
+    let(:course) { TestSetup.unpub_pri_math }
+    let!(:provider) { course.provider }
 
     describe "#courses_count" do
-      before do
-        provider
-      end
-
       it "returns course count using courses.size" do
         allow(provider.courses).to receive(:size).and_return(1)
 
@@ -298,10 +294,8 @@ describe Provider, type: :model do
     end
 
     describe ".include_courses_counts" do
+      let(:courses) { [course] }
       let(:first_provider) { Provider.include_courses_counts.first }
-      before do
-        provider
-      end
 
       it "includes course counts" do
         expect(first_provider.courses_count).to eq(1)
--- a/spec/models/recruitment_cycle_spec.rb
+++ b/spec/models/recruitment_cycle_spec.rb
@@ -13,9 +13,12 @@
 require "rails_helper"
 
 describe RecruitmentCycle, type: :model do
-  subject { find_or_create :recruitment_cycle, year: "2019" }
+  let(:current_cycle) { TestSetup.current_cycle }
+  let(:next_cycle) { TestSetup.next_cycle }
 
-  its(:to_s) { should eq("2019/20") }
+  subject { current_cycle }
+
+  its(:to_s) { should eq("2020/21") }
 
   it "is valid with valid attributes" do
     expect(subject).to be_valid
@@ -44,23 +47,18 @@ describe RecruitmentCycle, type: :model do
   end
 
   describe "current?" do
-    let(:current_cycle) { find_or_create :recruitment_cycle }
-    let(:second_cycle) { find_or_create :recruitment_cycle, :next }
-
     it "should return true when it's the current cycle" do
       expect(current_cycle.current?).to be(true)
     end
 
     it "should return true false it's not the current cycle" do
-      expect(second_cycle.current?).to be(false)
+      expect(next_cycle.current?).to be(false)
     end
   end
 
   context "when there are multiple cycles" do
-    let(:current_cycle) { find_or_create :recruitment_cycle }
-    let!(:second_cycle) { find_or_create :recruitment_cycle, :next }
     let!(:third_cycle) do
-      find_or_create :recruitment_cycle, year: second_cycle.year.to_i + 1
+      find_or_create :recruitment_cycle, year: next_cycle.year.to_i + 1
     end
 
     describe ".current_recruitment_cycle" do
@@ -71,7 +69,7 @@ describe RecruitmentCycle, type: :model do
 
     describe ".next_recruitment_cycle" do
       it "returns the next cycle after the current one" do
-        expect(RecruitmentCycle.next_recruitment_cycle).to eq(second_cycle)
+        expect(RecruitmentCycle.next_recruitment_cycle).to eq(next_cycle)
       end
     end
 
@@ -111,7 +109,7 @@ describe RecruitmentCycle, type: :model do
       context "next_recruitment_cycle" do
         let(:provider) do
           build(:provider, sites: [site],
-                recruitment_cycle: second_cycle)
+                recruitment_cycle: next_cycle)
         end
 
         it "returns the empty" do
@@ -122,21 +120,18 @@ describe RecruitmentCycle, type: :model do
 
     describe "#next" do
       subject { current_cycle }
-      its(:next) { should eq(second_cycle) }
+      its(:next) { should eq(next_cycle) }
 
       it "is nil for the newest cycle" do
         expect(third_cycle.next).to be_nil
       end
 
       it "returns the next cycle along when there is one" do
-        expect(second_cycle.next).to eq(third_cycle)
+        expect(next_cycle.next).to eq(third_cycle)
       end
     end
 
     describe "next?" do
-      let(:current_cycle) { find_or_create :recruitment_cycle }
-      let(:next_cycle) { find_or_create :recruitment_cycle, :next }
-
       it "should return true when it's the next cycle" do
         expect(next_cycle.next?).to be(true)
       end
--- a/spec/requests/api/v2/providers/courses/publishable_spec.rb
+++ b/spec/requests/api/v2/providers/courses/publishable_spec.rb
@@ -1,30 +1,19 @@
 require "rails_helper"
 
 describe "Publishable API v2", type: :request do
-  let(:user)         { create(:user) }
-  let(:organisation) { create(:organisation, users: [user]) }
-  let(:provider)     { create :provider, organisations: [organisation] }
-  let(:payload)      { { email: user.email } }
-  let(:token)        { build_jwt :apiv2, payload: payload }
-  let(:credentials) do
-    ActionController::HttpAuthentication::Token.encode_credentials(token)
-  end
+  let(:course)        { TestSetup.unpub_pri_math }
+  let(:provider)      { course.provider }
+  let(:organisation)  { provider.organisations.first }
+  let(:user)          { provider.users.first }
+  let(:payload)       { { email: user.email } }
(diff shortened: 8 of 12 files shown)


spec/support/test_setup.rb
```diff
@@ -0,0 +1,139 @@
...
+        super
+      end
+    end
+
+    @@recruitment_cycles = {
+      current_cycle: -> { FactoryBot.find_or_create :recruitment_cycle },
+      next_cycle: -> { FactoryBot.find_or_create :recruitment_cycle, :next }
+    }
+
+    @@users = {
```
TobyRet: Would it make sense to have 'creator services' for users, orgs and providers? (similar to the `SubjectCreatorService`). This would cut down on a lot of factory creation

spec/models/recruitment_cycle_spec.rb
```diff
@@ -13,6 +13,10 @@
 require "rails_helper"
 
 describe RecruitmentCycle, type: :model do
+
+  let(:current_cycle) { TestSetup.current_cycle }
+  let(:next_cycle) { TestSetup.next_cycle }
+
   subject { find_or_create :recruitment_cycle, year: "2019" }
```
misaka: Would it make sense, as part of this PR, to change this too? We would eventually want to do that, and since we're changing all the other cycles in this file already, maybe it would (make sense).

spec/requests/api/v2/providers/courses/publishable_spec.rb
```diff
@@ -1,31 +1,27 @@
...
-  let(:token)        { build_jwt :apiv2, payload: payload }
-  let(:credentials) do
-    ActionController::HttpAuthentication::Token.encode_credentials(token)
+  let(:provider) { create :provider, organisations: [organisation] }
+  let(:payload) { { email: user.email } }
+  let(:token) { build_jwt :apiv2, payload: payload }
+  let(:credentials) { ActionController::HttpAuthentication::Token.encode_credentials(token) }
+  let(:setup) do
+    TestSetup.new(with_courses: {
+      course: { provider: provider },
```
misaka: I think the goal should be to avoid having to do this. I think we should change the factories to work with `find_or_create :provider`, which would mean we don't have to provide a provider when creating courses, etc.

spec/factories/courses.rb
```diff
@@ -208,14 +208,16 @@
...
+    trait :unpublished_with_primary_maths do
       transient do
-        identifier { "published_with_primary_mathematics" }
+        identifier { "unpublished_with_primary_mathematics" }
       end
 
       name { "#{identifier} course" }
 
       provider { find_or_create :provider, :published_scitt }
+      site_statuses { [build(:site_status, :new)] }
```
misaka: This will need to be associated with the correct provider. There's a few ways to do this:

1. add functionality to the `site_status` or `site` factory so that if the provider wasn't specified, to introspect the course and use it's provider, perhaps as a separate trait like `:auto_provider`
2. add the course's provider to the attributes here
3. add a trait to the `provider` factory to be able to add a site to it

... ya, this will warrant a little bit more thinking.

TobyRet: Change made here - https://github.com/DFE-Digital/manage-courses-backend/pull/1097/commits/65c368d95828c9c3b04fe6d3ffbf13d46700ec03

spec/support/reference_data.rb
```diff
@@ -25,6 +25,14 @@
   config.before(:all) do
     SubjectCreatorService.new.execute
     SubjectFinancialIncentiveCreatorService.new.execute
```
TobyRet: I guess the SubjectFinancialIncentiveCreatorService and SubjectCreatorService be migrated over to the TestSetup pattern? And this is where sub-classing the TestSetup class might be useful ...

spec/requests/api/v2/providers/courses/publishable_spec.rb
```diff
@@ -69,8 +49,9 @@
     describe "failed validation" do
       let(:json_data) { JSON.parse(subject.body)["errors"] }
 
-      context "no enrichments" do
-        let(:course) { create(:course, provider: provider) }
+      context "no enrichments and location" do
+        let(:course) { TestSetup.unpublished_course_with_no_location_or_enrichment }
```
TobyRet: How 'granular' should we be with regards to course test data?

spec/factories/courses.rb
```diff
@@ -127,6 +127,11 @@
       # We've just created a course with this provider's code, so ensure it's
       # up-to-date and has this course loaded.
       course.provider.reload
+
```
TobyRet: I need to add 'find_or_create' to the provider on line 45

TobyRet:
[Create standard test data setup](https://trello.com/c/AZMRB3wq/2731-create-standard-test-data-setup)

spec/requests/api/v2/providers/courses/publishable_spec.rb
```diff
@@ -43,23 +32,14 @@
     include_examples "Unauthenticated, unauthorised, or not accepted T&Cs"
 
     context "when course and provider is not related" do
-      let(:course) { create(:course) }
+      let(:course)   { create(:course, :unpublished_with_primary_maths) }
+      let(:provider) { create(:provider) }
```
faissalMT: Shouldn't these now be using `TestSetup`

spec/requests/api/v2/providers/courses/publishable_spec.rb
faissalMT: I think we should increase granularity as needed rather than upfront so as to try and have a more uniform set of test courses, but I do think some tests will require high degrees of granularity although I can't think of any off the top of my head.

.rubocop.yml
```diff
@@ -13,3 +13,7 @@ AllCops:
 
 Rails/HasManyOrHasOneDependent:
   Enabled: false
+
+Style/ClassVars:
+  Exclude:
+    - 'spec/support/test_setup.rb'
```
misaka: Is this your preferred way to exclude a specific test for just one file? I'm more used to including it in the file so it's more obvious.

spec/factories/courses.rb
```diff
@@ -125,6 +125,11 @@
...
+      # Make sure this Course and and Sites are associated with the same Provider
+      if course.site_statuses.any?
+        course.site_statuses.first.site.provider = course.provider
```
misaka: Hrm. A side-effect of doing it this way, I think, is that if a test tries to associate a `site` with a `course` that is from another `provider`, say to test a validation or something similar, then this is going to over-write that.

The original comment I left was about making sure that the site created by default had the correct association. I think we should be a bit more cautious about how we do that.

spec/factories/courses.rb
```diff
@@ -209,5 +214,17 @@
     trait :skip_validate do
       to_create { |instance| instance.save(validate: false) }
     end
+
+    trait :unpublished_with_primary_maths do
+      transient do
+        identifier { "unpublished_with_primary_mathematics" }
+      end
+
+      name { "#{identifier} course" }
```
misaka: I realised after I did this that it should actually be named after the attribute, so we can track where it gets used it if shows up elsewhere.

```suggestion
      name { "#{identifier} course name" }
```

spec/factories/organisations.rb
```diff
@@ -15,5 +15,9 @@
   factory :organisation do
     name { "LONDON SCITT" + rand(1000000).to_s }
     sequence(:org_id)
+
+    trait :with_user do
+      users { [find_or_create(:user)] }
```
misaka: Not so sure about the `find_or_create` here (sorry if I suggested it should be this).

Reason being I think it's more common that, by default, each organisation gets one user. If we want one user to be shared across organisations, it's my feeling we should have to do that explicitly.

spec/factories/providers.rb
```diff
@@ -90,5 +90,13 @@
     trait :previous_recruitment_cycle do
       recruitment_cycle { find_or_create :recruitment_cycle, :previous }
     end
+
+    trait :published_scitt do
+      transient do
+        identifier { "published_scitt" }
+      end
+
+      provider_name { "#{identifier} provider" }
```
misaka:
```suggestion
      provider_name { "#{identifier} provider name" }
```

spec/requests/api/v2/providers/courses/publishable_spec.rb
misaka: I think we should try to list out what courses we could reasonably expect to see in the wild. For stuff like failure case testing, maybe a single course that doesn't have any values set might be enough ... trying to cater to each test's requirements with a TestSetup course would be overkill.

spec/support/reference_data.rb
misaka: Well, the thing is that `SubjectCreatorService` and `SubjectFinancialIncentiveCreatorService` are used by the db seeding. I wouldn't want db seeding to use factories, so moving them here would imply duplicating whatever logic is in those services between seeding and factories.

spec/support/reference_data.rb
```diff
@@ -25,6 +25,14 @@
...
+
+    # Creating or finding course test data
+    if Settings.test_setup
+      TestSetup.unpublished_course_with_primary_maths
+      TestSetup.unpublished_course_with_no_location_or_enrichment
+      TestSetup.unpublished_course_with_draft_enrichment
+      TestSetup.unpublished_fee_type_based_course_with_invalid_enrichment
+    end
```
misaka: Isn't this done already by instantiating `TestSetup`?

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
+class TestSetup
+  # Class that instantiates standard set of test data for specs.
+
+  @@recruitment_cycles = {
+    current_cycle: -> { FactoryBot.find_or_create :recruitment_cycle },
```
misaka: Can we get by without using `FactoryBot` here? It seems really noisy and affects reability, imo.

.rubocop.yml
TobyRet: No problem - I will inline the exclusion

spec/factories/organisations.rb
TobyRet: Ok makes sense

spec/support/reference_data.rb
```diff
@@ -25,6 +25,14 @@
   config.before(:all) do
     SubjectCreatorService.new.execute
     SubjectFinancialIncentiveCreatorService.new.execute
+
+    # Creating or finding course test data
+    if Settings.test_setup
```
misaka: Good call using `Settings`

misaka:
I think we still need to work on not creating too many extra records here ... it looks like for every `unpublished_course_with_primary_maths` being created, 4 organisations, 4 users and 2 providers are being created ... and that's for each time the course is being created too.

I tested this by adding a quick example to do some manual testing to `spec/requests/api/v2/providers/courses/publishable_spec.rb`:

```
    it "tests" do
      binding.pry
      nil
    end
```

And then looking at the number of records being created:

```ruby
[1] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Organisation.count
=> 0
[2] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Provider.count
=> 0
[3] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> User.count
=> 0
[4] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Course.count
=> 0
[5] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> TestSetup.unpublished_course_with_primary_maths
=> #<Course:0x00007fe36147b9a8
 id: 1,
 ...>
[6] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Organisation.count
=> 4
[7] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Provider.count
=> 2
[8] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> User.count
=> 4
[9] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Course.count
=> 1
[10] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> TestSetup.unpublished_course_with_primary_maths
=> #<Course:0x00007fe3614d26e0
 id: 2,
 ...>
[11] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Organisation.count
=> 8
[12] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Provider.count
=> 4
[13] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> User.count
=> 8
[14] pry(#<RSpec::ExampleGroups::PublishableAPIV2::POSTPublishable>)> Course.count
=> 2
```

I would expect to only see 1 organisation, 1 user and 1 provider (I don't think we should be creating a course with an accredited body by default) being created, and probably, I think, they should be shared across the courses so that we don't create extra resources by default (we explicitly have to create extra providers if we want courses from different providers to test with).

spec/requests/api/v2/providers/courses/publishable_spec.rb
```diff
@@ -43,23 +32,14 @@
...
-      let!(:course) do
-        create(:course,
-               provider: provider,
-               site_statuses: [site_status],
-               enrichments: [enrichment],
-               age: 17.days.ago,
-               subjects: [primary_with_mathematics])
-      end
+    context "unpublished course with draft enrichment" do
+      let(:course) { TestSetup.unpublished_course_with_draft_enrichment }
```
timabell: Is the intention that we stop using `build`/`create` entirely? Or will still use build/create for interesting edge cases

spec/requests/api/v2/providers/courses/publishable_spec.rb
timabell: I'm also a bit concerned about the proliferation of methods like this for every variation. I think we should set a strong opinion out as to what in what cases `TestSetup` should be extended vs. just building custom things. `unpublished_course_with_no_location_or_enrichment` smells to me a bit of heading down the proliferation path already and maybe we shouldn't change this one?

spec/requests/api/v2/providers/courses/publishable_spec.rb
```diff
@@ -81,15 +62,8 @@
...
-        let(:course) {
-          create(:course, :fee_type_based,
-                 provider: provider,
-                 enrichments: [invalid_enrichment],
-                 site_statuses: [site_status])
-        }
-
         context "invalid enrichment with invalid content lack_presence fields" do
-          let(:invalid_enrichment) { create(:course_enrichment, :without_content) }
+          let(:course) { TestSetup.unpublished_fee_type_based_course_with_invalid_enrichment }
```
timabell: As above, smells of unbounded proliferation coming down the line

spec/requests/api/v2/providers/courses/publishable_spec.rb
misaka: I think that that's a re-stating of the granularity question above, and ya ... we would still use create/build for interesting edge cases, aiming to just cover the majority of the types of courses in `TestSetup` but not all.

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
+class TestSetup
+  # Class that instantiates standard set of test data for specs.
```
timabell: Would be good to mention what problem we are solving by having this

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
...
+      }.merge(attributes))
+    end,
+  }
+
+  # Used when not instantiating a TestSetup object in a before block.
+  # This will create a new fixture, using 'find_or_create' for teams and
+  # users and 'create' for cases. This is used to have a standardised set of
+  # recruitment cycles to work with, but without requiring
+  # pre-instantiation.
+
```
timabell: Is the above the docs for `def self.method_missing`? If so kill the extra newline

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
...
+        enrichments: [FactoryBot.build(:course_enrichment, :without_content)],
+      }.merge(attributes))
+    end,
+  }
+
+  # Used when not instantiating a TestSetup object in a before block.
+  # This will create a new fixture, using 'find_or_create' for teams and
+  # users and 'create' for cases. This is used to have a standardised set of
+  # recruitment cycles to work with, but without requiring
+  # pre-instantiation.
```
timabell: I'm not sure I entirely understand this

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
...
+    if @@recruitment_cycles&.key?(method_name)
+      @@recruitment_cycles[method_name].call
+    elsif @@courses&.key?(method_name)
+      @@courses[method_name].call
+    else
+      super
+    end
+  end
+
+  def self.respond_to_missing?(method_name, include_private = false)
```
timabell: I'm assuming the comment above the above method applies to this too. Might be worth documenting them individually too as they're pretty hard to parse for someone like me.

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
...
+    @@recruitment_cycles&.key?(method_name) ||
+      @@courses&.key?(method_name) ||
+      super
+  end
+
+  attr_reader :courses, :recruitment_cycles
+
+  def initialize(with_courses: nil)
```
timabell: Shouldn't this be at the top of the class?

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
...
+    else
+      course_types = with_courses || @@courses.keys
+      @courses = Hash[
+        @@courses.slice(*course_types).map do |name, kourse|
+          # instantiate course by calling the blocks in @@courses
+          [name, kourse.call]
+        end
+      ]
+    end
+  end
```
timabell: This if/else block is pretty confusing to me. Perhaps a comment, or refactor out into named functions?

spec/support/test_setup.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def method_missing(method_name, *args)
```
timabell: As above; Might be worth documenting `method_missing` and `respond_to_missing` too as I find them both pretty hard to parse.

spec/factories/providers.rb
```diff
@@ -90,5 +90,13 @@
     trait :previous_recruitment_cycle do
       recruitment_cycle { find_or_create :recruitment_cycle, :previous }
     end
+
+    trait :published_scitt do
+      transient do
+        identifier { "published_scitt" }
```
defong: is this a glorified variable?

spec/support/test_setup.rb
```diff
@@ -0,0 +1,95 @@
...
+    end,
+    unpublished_fee_type_based_course_with_invalid_enrichment: ->(attributes = {}) do
+      find_or_create(:course, :fee_type_based, :unpublished_with_primary_maths, {
+        enrichments: [build(:course_enrichment, :without_content)],
+      }.merge(attributes))
+    end,
+  }
+
+  # Used when not instantiating a TestSetup object in a before block.
+  # This will create a new fixture, using 'find_or_create' for teams and
```
defong: lost... what is teams refering to?

spec/factories/providers.rb
TobyRet: Indeed it is

spec/factory_specs/course_spec.rb
```diff
@@ -30,4 +30,23 @@
...
+  context "unpublished_with_primary_maths" do
+  subject! { TestSetup.unpublished_course_with_primary_maths }
+
+    its(:name) { should eq("unpublished_with_primary_maths course name")}
+
+    it "has the correct number of associations" do
+      expect(Site.count).to eq(1)
+      expect(Provider.count).to eq(1)
+      expect(Organisation.count).to eq(1)
+      expect(User.count).to eq(1)
```
misaka: When we add `TestSetup` to the `before :all`, this fail. Probably better to change `subject!` to `subject` and do:

```
      expect { subject }.not_to change { Site.count }
```

... etc

spec/factory_specs/course_spec.rb
```diff
@@ -30,4 +30,23 @@
       expect(CourseEnrichment.where("created_at < ?", 4.days.ago).count).to eq(1)
     end
   end
+
+  context "unpublished_with_primary_maths" do
+  subject! { TestSetup.unpublished_course_with_primary_maths }
```
misaka:
```suggestion
    subject! { TestSetup.unpublished_course_with_primary_maths }
```

spec/factory_specs/course_spec.rb
misaka: Also, I think the factory we're testing here should only create one `Site` and `Provider`, the `provider` factory should only create one `Organisation`, and the `organisation` factory should only create one `User`.

spec/models/provider_spec.rb
```diff
@@ -300,7 +300,7 @@
     describe ".include_courses_counts" do
       let(:first_provider) { Provider.include_courses_counts.first }
       before do
-        provider
+        provider.courses << course
```
misaka: This should be handled in the factory.

spec/requests/api/v2/providers/courses/publishable_spec.rb
misaka: Ya, I agree that this one sounds a bit too specialised.

spec/services/providers/copy_to_recruitment_cycle_service_spec.rb
```diff
@@ -37,6 +36,10 @@
       )
     end
 
+    before do
+      provider.courses << course
+    end
```
misaka: As above somewhere, I think we need to get this into the factory.

spec/support/test_setup.rb
misaka: This feels alright to me. Everything above are class-scoped so this is where the instance-specific stuff starts.

spec/support/test_setup.rb
```diff
@@ -0,0 +1,94 @@
+# rubocop:disable Style/ClassVars
+class TestSetup
```
misaka: iirc, you can just do this:

```
class TestSetup # rubocop:disable Style/ClassVars
```

and it'll only disable those checks for this class.

misaka:
@TobyRet did his presentation of this to the team and we agreed we'd continue to look at this approach. First step is to try to improve the API to not use method missing, but to use something like:

```
@setup = TestSetup.new
@setup.course :unpublished, :primary, :mathematics
```

PR #1099, opened 2020-01-02
[2743] Add v3 /providers/courses endpoint

faissalMT:
### Context
In order to create the results page for find there must be a way to list all courses regardless of course.

Create the path /providers/courses, and allow it to be paginated.

### Guidance to review
https://jsonapi.org/format/#fetching-pagination

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/courses_controller.rb
+++ b/app/controllers/api/v3/courses_controller.rb
@@ -3,14 +3,14 @@ module API
     class CoursesController < API::V3::ApplicationController
       before_action :build_recruitment_cycle
       before_action :build_provider
+      before_action :build_courses
 
       def index
-        @courses = @provider.courses
-        render jsonapi: @courses, include: params[:include]
+        render jsonapi: paginate(@courses), fields: fields_param, include: params[:include], class: CourseSerializersService.new.execute
       end
 
       def show
-        @course = @provider.courses.find_by!(course_code: params[:code].upcase)
+        @course = @courses.find_by!(course_code: params[:code].upcase)
 
         if @course.is_published?
           # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
@@ -22,17 +22,27 @@ module API
 
       def fields_param
         params.fetch(:fields, {})
-          .permit(:courses)
+          .permit(:courses, :providers)
           .to_h
           .map { |k, v| [k, v.split(",").map(&:to_sym)] }
       end
 
     private
 
+      def build_courses
+        @courses = if @provider.present?
+                     @provider.courses
+                   else
+                     @recruitment_cycle.courses
+                   end
+      end
+
       def build_provider
-        @provider = @recruitment_cycle.providers.find_by!(
-          provider_code: params[:provider_code].upcase,
-        )
+        if params[:provider_code].present?
+          @provider = @recruitment_cycle.providers.find_by!(
+            provider_code: params[:provider_code].upcase,
+          )
+        end
       end
     end
   end
--- a/config/initializers/jsonapi.rb
+++ b/config/initializers/jsonapi.rb
@@ -1,5 +1,13 @@
+def page_url(page:)
+  params.then do |new_params|
+    new_params[:page] = {} unless new_params.key? :page
+    new_params[:page][:page] = page
+    url_for(new_params.to_unsafe_h.merge(only_path: true))
+  end
+end
+
 JSONAPI::Rails.configure do |config|
-  # Set a default serializable class mapping.
+# Set a default serializable class mapping.
   config.jsonapi_class = Hash.new { |h, k|
     names = k.to_s.split("::")
     klass = names.pop
@@ -8,8 +16,8 @@ JSONAPI::Rails.configure do |config|
       "API::V2::Serializable#{klass}",
       [*names, klass].join("::"),
     ].lazy
-     .map(&:safe_constantize)
-     .detect(&:present?)
+      .map(&:safe_constantize)
+      .detect(&:present?)
   }
 
   # # Set a default serializable class mapping for errors.
@@ -34,7 +42,21 @@ JSONAPI::Rails.configure do |config|
   # }
   #
   # # Set a default pagination scheme.
-  # config.jsonapi_pagination = ->(_) { nil }
+  config.jsonapi_pagination = ->(collection) do
+    {}.tap do |links|
+      if collection.respond_to?(:next_page) && collection.next_page.present?
+        links[:next] = page_url(page: collection.next_page)
+      end
+
+      if collection.respond_to?(:prev_page) && collection.prev_page.present?
+        links[:prev] = page_url(page: collection.prev_page)
+      end
+
+      if collection.respond_to?(:total_pages)
+        links[:last] = page_url(page: [collection.total_pages - 1, 1].max)
+      end
+    end
+  end
   #
   # # Set a logger.
   # config.logger = Logger.new(STDOUT)
--- a/spec/requests/api/v3/providers/courses/index_spec.rb
+++ b/spec/requests/api/v3/providers/courses/index_spec.rb
@@ -29,12 +29,13 @@ describe "GET v3/providers/:provider_code/courses" do
            applications_open_from: applications_open_from)
   end
 
-  let(:courses_site_status) {
+  let(:courses_site_status) do
     build(:site_status,
           :findable,
           :with_any_vacancy,
           site: create(:site, provider: provider))
-  }
+  end
+
   let(:enrichment)     { build :course_enrichment, :published }
   let(:provider)       { create :provider }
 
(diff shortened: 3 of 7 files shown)


faissalMT:
[Add /providers/courses endpoint for listing all courses to API v3](https://trello.com/c/oy4IpvNY/2743-add-providers-courses-endpoint-for-listing-all-courses-to-api-v3)

app/controllers/api/v3/courses_controller.rb
```diff
@@ -3,14 +3,25 @@ module V3
...
+      before_action :build_courses
 
       def index
-        @courses = @provider.courses
-        render jsonapi: @courses, include: params[:include]
+        links = {}
+        page_index = params.dig(:page, :page)
+        per_page = params.dig(:page, :per_page)
+
+        if per_page.present?
```
misaka: I don't think paging should be optional ... we should default `per_page` to a number from settings (so it can be over-ridden just in case). 100 might be a reasonable start (seems to be the built-in default).

app/controllers/api/v3/courses_controller.rb
```diff
@@ -3,14 +3,25 @@ module V3
...
+          @courses = @courses.page(page_index).per(per_page)
+
+          links[:next] = next_page if next_page.present?
+          links[:prev] = previous_page if previous_page.present?
+        end
```
misaka: Reading the [docs for api-pagination](https://github.com/davidcelis/api-pagination#rails), it looks like it can handle all this header generation for us, which is kind of the point of why we're using it I think. I think all we need to do is:

```
        @courses = paginate @courses
```

faissalMT: api-pagination doesn't adhere to the JSON:API standard for pagination, and so breaks Rails Find.

misaka: Using `paginate` is a nice shortcut to do a lot of the processing that's currently being hand-done. So, if you add this to the action:

```
          @courses = paginate @courses
```

you can then add a method that `jsonapi.rb` uses to auto-generate those links:

```
      def jsonapi_pagination(collection)
        {
          first: url_for(page: 1),
          last: url_for(page: collection.total_pages),
          next: url_for(page: collection.next_page),
          prev: url_for(page: collection.prev_page),
        }
      end
```

(unfortunately it appears this isn't documented, but see the PR jsonapi-rb/jsonapi-rails#35)

With this, you get both the `Link` HTTP header:

```
Link: <http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=1&per_page=2>; rel="first", <http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=1&per_page=2>; rel="prev", <http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=15&per_page=2>; rel="last", <http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=3&per_page=2>; rel="next"
```

And the link section in JSONAPI:

```
    "links": {
        "first": "http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=1",
        "last": "http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=15",
        "next": "http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=3",
        "prev": "http://localhost:3001/api/v3/recruitment_cycles/2020/providers/1LA/courses?page=1"
    }
```

With less code. `jsonapi_pagination` can be put in a parent controller for automatic inclusion, although it'll need a guard to make sure `collection` is paginated.

app/controllers/api/v3/courses_controller.rb
```diff
@@ -29,10 +40,35 @@ def fields_param
...
+      end
+
+      def build_courses
+        @courses = if @provider.present?
+                     @provider.courses
+                   else
+                     @recruitment_cycle.courses
+                   end
+      end
```
misaka: Looking at the server logs this appears to be an N+1 query loading `FinancialIncentive` and `CourseEnrichment` separately.

faissalMT: This is now fixed, FinancialIncentives just needed an include in a service, while course enrichments was fixed by simply excluding unnecessary fields.

spec/requests/api/v3/providers/courses/index_spec.rb
```diff
@@ -51,187 +52,447 @@ def perform_request
...
+                     is_send: true,
+                     site_statuses: [courses_site_status],
+                     enrichments: [enrichment],
+                     maths: :must_have_qualification_at_application_time,
+                     english: :must_have_qualification_at_application_time,
+                     science: :must_have_qualification_at_application_time,
+                     age_range_in_years: "3_to_7",
+                     applications_open_from: applications_open_from)
+            end
+          end
```
misaka: This is going to really slow things down. Can we break this up into contexts:

- we have one test that checks that the default `per_page` is set to 10, only requiring us to create 11 courses
- we have another test that operates with `per_page` set to 2, only requiring us to create 6 courses

spec/requests/api/v3/providers/courses/index_spec.rb
```diff
@@ -51,187 +52,447 @@ def perform_request
...
+              },
+              "relationships" => {
+                "accrediting_provider" => { "meta" => { "included" => false } },
+                "provider" => { "meta" => { "included" => false } },
+                "site_statuses" => { "meta" => { "included" => false } },
+                "sites" => { "meta" => { "included" => false } },
+                "subjects" => { "meta" => { "included" => false } },
+              },
+              "meta" => {
+                "edit_options" => {
```
defong: should these `edit_options` be exposed?

spec/requests/api/v3/providers/courses/index_spec.rb
defong: can we break this into its own file or something?

spec/requests/api/v3/providers/courses/index_spec.rb
faissalMT: They're unnecessary but metadata can't be individually exposed/hidden like other attributes because they're non-standard and I don't want to hardcode it to disable metadata if this is meant to be an open API. I think we should leave it for now.

spec/requests/api/v3/providers/courses/index_spec.rb
faissalMT: This is all regarding the same thing, I don't think it's necessary to move it to a separate file.

spec/requests/api/v3/providers/courses/index_spec.rb
misaka: We'll need a custom serialiser anyway, so while it uses the v2 serialiser sounds like it's not worth the effort to remove it.

config/initializers/jsonapi.rb
```diff
@@ -1,49 +1,71 @@
...
+# # Set default exposures.
+# # A lambda/proc that will be eval'd in the controller context.
+# config.jsonapi_expose = lambda {
+#   { url_helpers: ::Rails.application.routes.url_helpers }
+# }
+#
+# # Set a default pagination scheme.
+  config.jsonapi_pagination = ->(collection) do
+    {}.tap do |links|
+      if collection.methods.include?(:next_page) && collection.next_page.present?
```
misaka:
```suggestion
      if collection.respond_to?(:next_page) && collection.next_page.present?
```

config/initializers/jsonapi.rb
```diff
@@ -1,49 +1,71 @@
...
+        links[:next] = page_url(page: collection.next_page)
+      end
+
+      if collection.methods.include?(:prev_page) && collection.prev_page.present?
```
misaka:
```suggestion
      if collection.respond_to?(:prev_page) && collection.prev_page.present?
```

config/initializers/jsonapi.rb
```diff
@@ -1,49 +1,71 @@
...
+        links[:prev] = page_url(page: collection.prev_page)
+      end
+
+      if collection.methods.include?(:total_pages)
```
misaka:
```suggestion
      if collection.respond_to?(:total_pages)
```

config/initializers/jsonapi.rb
```diff
@@ -1,49 +1,71 @@
...
     h[k] = [
-      "API::V2::Serializable#{klass}",
-      [*names, klass].join("::"),
+    "API::V2::Serializable#{klass}",
+    [*names, klass].join("::"),
     ].lazy
-     .map(&:safe_constantize)
-     .detect(&:present?)
+    .map(&:safe_constantize)
+    .detect(&:present?)
```
misaka: What happened to the indentiation here and the comments below?

PR #1106, opened 2020-01-07
[2732] Block users from creating identical courses

danielburnley:
### Context

As a provider
I need to be blocked from creating an identical course
So that I do not create duplicate courses by mistake

- Add new service that checks if a course a duplicate of an already existing one on the provider
- Prevents creation of a duplicate course

### Guidance to review

- Run through the course creation flow on the frontend
- Create a course on the frontend
- Press back in  your browser
- Attempt to create the same course

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


danielburnley:
https://trello.com/c/xIemKmAm/2732-block-users-from-creating-identical-courses

app/services/courses/validate_unique_course_on_provider_service.rb
```diff
@@ -0,0 +1,21 @@
+module Courses
+  class ValidateUniqueCourseOnProviderService
+    def execute(new_course:)
+      existing_courses = new_course.provider.courses
+      new_course_attritbutes = new_course.attributes.slice(*attributes_to_compare)
```
misaka:
```suggestion
      new_course_attributes = new_course.attributes.slice(*attributes_to_compare)
```

app/services/courses/validate_unique_course_on_provider_service.rb
```diff
@@ -0,0 +1,21 @@
...
+      existing_courses.none? do |existing_course|
+        existing_course_attributes = existing_course.attributes.slice(*attributes_to_compare)
+
+        new_course_attritbutes == existing_course_attributes &&
```
misaka:
```suggestion
        new_course_attributes == existing_course_attributes &&
```

spec/models/course_spec.rb
```diff
@@ -1624,4 +1624,17 @@
       )
     end
   end
+
+  describe "#is_unique_on_provider?" do
+    it "Returns the subjects using the correct service" do
```
misaka: Sorry, I know this is nit-picking, but does it make sense to say it "Returns the subjects" when we're not testing the return value, and doesn't fit with the name of the method? Maybe it could just be "delegates to the correct service".

app/services/courses/validate_unique_course_on_provider_service.rb
```diff
@@ -0,0 +1,21 @@
...
+          new_course.subjects == existing_course.subjects &&
+          new_course.accrediting_provider == existing_course.accrediting_provider
+      end
+    end
+
+  private
+
+    def attributes_to_compare
+      %w[level is_send age_range_in_years qualification program_type study_mode maths english science applications_open_from start_date]
```
defong: shouldn't it be using 

https://guides.rubyonrails.org/active_record_validations.html#uniqueness

?

app/services/courses/validate_unique_course_on_provider_service.rb
```diff
@@ -0,0 +1,21 @@
+module Courses
+  class ValidateUniqueCourseOnProviderService
```
defong: Should this line, given this class is called `ValidateUniqueXXX`

`@course.errors.add(:base, :duplicate)` sit in here so controller is not subjected to change?

misaka: Interesting point. I'd also ask why not use a custom validation within the `Course` model?

danielburnley: I mainly wanted to have it separate as I think we should be more explicit on how we determine whether or not a course is unique. 

This could potentially be moved to a `validates_with` method on the course though now that I think about it (this would pull in Sau's suggestion also)

misaka: Ya, that's what I meant, sorry I wasn't clear. Basically create a custom validator and use `validates_with`, using the existing service-like mechanism.

danielburnley: :+1: sounds good

PR #1109, opened 2020-01-08
mcb - Don't prompt for environment confirmation on qa/staging

timabell:
This has two benefits:

* Easier workflow for throw-away qa/staging data
* Makes it stand out much more when you do actually want production

### Context

Doing a lot of qa/staging stuff while building a new query before
running on prod.

* Don't prompt for environment confirmation on qa/staging

### Guidance to review

Ignore whitespace to see a more meaningful diff

:ship:

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/azure.rb
+++ b/lib/mcb/azure.rb
@@ -88,7 +88,7 @@ module MCB
                                          rgroup: rgroup,
                                          subscription: subscription)
 
-      unless %w(qa staging).include?(app_config["RAILS_ENV"])
+      unless require_confirmation
         print "As a safety measure, please enter the expected RAILS_ENV for #{webapp}: "
         expected_environment = $stdin.readline.chomp
 
@@ -104,5 +104,9 @@ module MCB
 
       app_config["RAILS_ENV"]
     end
+  end
+
+  def require_confirmation
+    %w(qa staging).include?(app_config["RAILS_ENV"])
   end
 end
--- a/lib/mcb/azure.rb
+++ b/lib/mcb/azure.rb
@@ -89,7 +89,6 @@ module MCB
                                          subscription: subscription)
 
       unless %w(qa staging).include?(app_config["RAILS_ENV"])
-        # TODO: only require confirmation on commands that write to the db
         print "As a safety measure, please enter the expected RAILS_ENV for #{webapp}: "
         expected_environment = $stdin.readline.chomp
 
--- a/lib/mcb/azure.rb
+++ b/lib/mcb/azure.rb
@@ -88,13 +88,15 @@ module MCB
                                          rgroup: rgroup,
                                          subscription: subscription)
 
-      # TODO: only require confirmation on commands that write to the db
-      print "As a safety measure, please enter the expected RAILS_ENV for #{webapp}: "
-      expected_environment = $stdin.readline.chomp
+      unless %w(qa staging).include?(app_config["RAILS_ENV"])
+        # TODO: only require confirmation on commands that write to the db
+        print "As a safety measure, please enter the expected RAILS_ENV for #{webapp}: "
+        expected_environment = $stdin.readline.chomp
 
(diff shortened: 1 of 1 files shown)


lib/mcb/azure.rb
```diff
@@ -88,13 +88,15 @@ def self.configure_for_webapp(webapp:, rgroup: nil, subscription: nil, **_opts)
                                          rgroup: rgroup,
                                          subscription: subscription)
 
-      # TODO: only require confirmation on commands that write to the db
-      print "As a safety measure, please enter the expected RAILS_ENV for #{webapp}: "
-      expected_environment = $stdin.readline.chomp
+      unless %w(qa staging).include?(app_config["RAILS_ENV"])
```
faissalMT: We have a handful of properties attributed to various environments that I think would be better placed in their configs (i.e: we'd have a property in `production.yml` like `requires_confirmation: true`) so that we have a single source of truth. I think the idea would allow us to do something like `requires_confirmation(app_config["RAILS_ENV"])`. This might be overkill but I think it's worth at least worth exploring to save us difficulty down the line.

misaka: 👍 I like this idea.

lib/mcb/azure.rb
```diff
@@ -88,13 +88,15 @@ def self.configure_for_webapp(webapp:, rgroup: nil, subscription: nil, **_opts)
...
+        # TODO: only require confirmation on commands that write to the db
```
defong: Is the todo still applies?

timabell: I'd ditch it, I'd say it should go in trello if it needs doing

lib/mcb/azure.rb
timabell: I like this idea, I tried and failed to read Settings from here though :cry:

misaka: Good point. `Settings` won't be able to available until Rails is loaded, but this is called in `rails_runner` to set things up before running Rails. I still think this would be best placed in it's own method though. It'd tidy this method up and it's conceivable that we'd setup some kind of env-specific configuration method in the future.

timabell: Maecenas aliquam nisl sit amet ligula imperdiet dignissim. Donec mattis diam ac magna malesuada, in mollis sapien facilisis. Donec iaculis nunc nisi, vitae sodales sapien lobortis at. Nam vel ultrices arcu, vitae lacinia lectus. Maecenas consectetur placerat nibh congue mattis. In eget justo eget quam venenatis aliquet. Nam nec felis vestibulum, finibus odio mollis, pharetra dui. Morbi nunc nisi, scelerisque vitae enim id, eleifend laoreet mi. Etiam gravida nisi nec nisi sodales, sed efficitur metus tempus. Interdum et malesuada fames ac ante ipsum primis in faucibus. Quisque ut erat ornare magna aliquam tempus non vitae libero. Donec aliquam dui eu elit ultrices, sed egestas elit aliquet. Suspendisse sed placerat erat, eget vulputate mi. Cras eget orci iaculis, blandit purus accumsan, aliquet neque.

Or at least that's what I heard.

timabell: Or in other words I don't really know what you are saying @misaka how about a quick pair when you have time?

timabell:
Rebased & the simple suggestion addressed. Need some help with the other one.

lib/mcb/azure.rb
timabell: Factored out the complexity into a method, agreed to leave complex settings gubbins alone for now. :+1:

dankmitchell:
Can I suggest we drop this if it's causing too much head-scratching...

timabell:
Talked it over with @misaka , did another commit, just waiting for thumbs up now

PR #1110, opened 2020-01-08
1591 return providers accredited by accrediting provider

gpeng:
### Context
We are adding a page that lists all of the providers that deliver courses for a particular accredited provider.

* Add `/api/v2/recruitment_cycles/<year>/providers/<provider_code/training_providers` route
* Add `AccreditedProviderTrainingProvidersController` to serve that URL ^
* Add `Provider#training_providers` method

### Guidance to review

The language is a bit tricky here. We've gone with 'training provider' being a provider that doesn't have it's own accreditation but delivers courses on behalf of an 'accredited provider' (a provider that does. Both of these are represented as a `Provider` object.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -1,0 +1,29 @@
+module API
+  module V2
+    class AccreditedProviderTrainingProvidersController < ApplicationController
+      before_action :build_recruitment_cycle
+      before_action :build_provider
+
+      def index
+        authorize @provider, :can_list_training_providers?
+        providers = @provider.training_providers
+
+        render jsonapi: providers, include: params[:include]
+      end
+
+    private
+
+      def build_recruitment_cycle
+        @recruitment_cycle = RecruitmentCycle.find_by(
+          year: params[:recruitment_cycle_year],
+        ) || RecruitmentCycle.current_recruitment_cycle
+      end
+
+      def build_provider
+        @provider = @recruitment_cycle.providers.find_by!(
+          provider_code: params[:provider_code].upcase,
+        )
+      end
+    end
+  end
+end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -86,7 +86,13 @@ class Provider < ApplicationRecord
            primary_key: :provider_code,
            inverse_of: :accrediting_provider
 
+  # the accredited_providers that this provider is a training_provider for
   has_many :accrediting_providers, -> { distinct }, through: :courses
+
+  # the providers that this provider is an accredited_provider for
+  def training_providers
+    Provider.where(id: current_accredited_courses.map(&:provider_id))
+  end
 
   def current_accredited_courses
     accredited_courses.select { |c| c.provider.recruitment_cycle == recruitment_cycle }
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -40,6 +40,7 @@
 #                                                                        PUT    /api/v2/providers/:provider_code/sites/:id(.:format)                                                                             api/v2/sites#update
 #                                                                        GET    /api/v2/providers/:provider_code/recruitment_cycles(.:format)                                                                    api/v2/recruitment_cycles#index
 #                                                                        POST   /api/v2/providers/:code/sync_courses_with_search_and_compare(.:format)                                                           api/v2/providers#sync_courses_with_search_and_compare
+#                                     api_v2_provider_training_providers GET    /api/v2/providers/:provider_code/training_providers(.:format)                                                                    api/v2/accredited_provider_training_providers#index
 #                                                       api_v2_providers GET    /api/v2/providers(.:format)                                                                                                      api/v2/providers#index
 #                                                                        POST   /api/v2/providers(.:format)                                                                                                      api/v2/providers#create
 #                                                        api_v2_provider GET    /api/v2/providers/:code(.:format)                                                                                                api/v2/providers#show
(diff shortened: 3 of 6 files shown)


gpeng:
[Return providers who have courses accredited by the current provider](https://trello.com/c/CQptgQEK/1591-return-providers-who-have-courses-accredited-by-the-current-provider)

app/models/provider.rb
```diff
@@ -86,8 +86,14 @@ class Provider < ApplicationRecord
            primary_key: :provider_code,
            inverse_of: :accrediting_provider
 
+  # the accredited_providers that this provider is a training_provider for
   has_many :accrediting_providers, -> { distinct }, through: :courses
```
misaka: Is the right name for this `qualification_providers`?

config/routes.rb
```diff
@@ -136,6 +141,11 @@
       get "/recruitment_cycles/:recruitment_cycle_year/providers/suggest", to: "providers#suggest"
 
       resources :organisations, only: :index
+      resources :accredited_body,
```
misaka: Is this resource attached to a controller? And where is it being used?

spec/models/provider_spec.rb
```diff
@@ -347,6 +347,19 @@
     end
   end
 
+  describe "training_providers" do
+    let(:accredited_provider) { create(:provider, accrediting_provider: "Y") }
```
misaka: There's a trait `accredited_body` we can use to set this to `Y`. 😭

spec/models/provider_spec.rb
```diff
@@ -347,6 +347,19 @@
...
+    let(:provider1) { create(:provider, accrediting_provider: "N") }
```
misaka: This appears to be the default for this factory, setting `accrediting_provider` to `N`. If we want this to be explicit, how about a trait to mirror `accredited_body`?

spec/requests/api/v2/providers/accredited_body_get_providers_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+      perform_request
+
+      response
+    end
+
+    context "when unauthorized" do
+      let(:payload) { { email: "foo@bar" } }
+
+      it { should have_http_status(:unauthorized) }
+    end
```
misaka: Would it be worth using the shared example for unauthorised tests?

https://github.com/DFE-Digital/manage-courses-backend/blob/master/spec/support/shared_examples/unauth.rb

config/routes.rb
gpeng: It is not.. Experiment hangover. Removed in 3c2e7a0

spec/models/provider_spec.rb
gpeng: Fixed in 3c2e7a0

spec/models/provider_spec.rb
gpeng: Fixed in 3c2e7a0. I've renamed those too as it's not quite as obvious without that param.

spec/requests/api/v2/providers/accredited_body_get_providers_spec.rb
gpeng: It would as that uncovered a bug in the authorization. Absolute master stroke. Nice one 👍

app/models/provider.rb
gpeng: 🤷‍♂ Who knows? That's existing code. We're pretty settled on accredited_provider and training_provider as the distinction but as we've discussed, plenty of flavours. Paul G suggested training provider which had a certain ring to it so we went with that. I've got a tech debt 'note' to make it consistent. Open to suggestions on where to stick that 😆

misaka: I'm starting to keep my tech debt on my own Trello board, but we might start up a backlog Trello board that the whole team can use. Discussion to come.

The reason I asked here, though, is that I'm thinking what we'll be calling it in the "V3" API, the one that'll be more stable and for public consumption. Thinking about this a little more, I'm not sure we need `qualification_providers`, since we have:

- **accredited_body** - the provider that is a accredited by the DfE to provide qualifications for a course
- **training_provider** - the provider chosen by the accredited_body to provider the training for the course

As for the relationship above, sounds like it should be `accredited_bodies`.

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    private
+
+      def build_recruitment_cycle
+        @recruitment_cycle = RecruitmentCycle.find_by(
+          year: params[:recruitment_cycle_year],
+        ) || RecruitmentCycle.current_recruitment_cycle
+      end
+
+      def build_provider
+        @provider = @recruitment_cycle.providers.find_by!(
```
misaka: Correct me if I'm wrong, but I *think* we need to be using the policy scope here. See https://github.com/DFE-Digital/manage-courses-backend/blob/2731-test-setup-helper/app/controllers/api/v2/providers_controller.rb#L14 for an example.

misaka: Nope. Having thought about this some more, this is ok because we have `authorize @provider, :can_list_training_providers?` in `index`.

config/routes.rb
```diff
@@ -134,8 +136,11 @@
         resources :sites, only: %i[index update show create]
         resources :recruitment_cycles, only: %i[index]
         post :sync_courses_with_search_and_compare, on: :member
+        get :training_providers, to: "accredited_provider_training_providers#index"
       end
 
+      # /providers/<code>/training_providers/<training_provider_code>/courses
+
```
misaka: Does this comment belong here?

gpeng: 🤦‍♂ It does not...

PR #1113, opened 2020-01-09
Remove localhost DB_HOSTNAME config

gpeng:
Turns out this breaks things for other devs so reverting.

### Context

This was added to address an issue I had during on-boarding but is now broken for Linux users. The issue I had seems to have resolved so it may have been unrelated. Possibly moving from Postgres 12 to 9.6 caused an issue for me 🤷‍♂ 

Remove the localhost default value for `DB_HOSTNAME`

### Checklist

- [x] Rebased master
- [x] Cleaned commit history


timabell:
reverts part of #1102

PR #1115, opened 2020-01-10
[2767] mcb - add flag to allow geocode of invalid sites

timabell:
### Context
There are around 800 (out of 16000) Sites that are not currently geocoded. This because these Sites cannot be updated due to validation issues (e.g. they were imported from legacy with invalid attributes e.g. partial postcode or duplicate Site name). 

#### To do

- Update the mcb geocode tasks (Sites and Providers) to skip validation then re-geocode all Sites (and Providers) with nil lat and long on each environment. 

#### Success

* We update the lat and log for invalid sites as a one-off process.
* Command is repeatable in case we need.

### Tech Notes

* This is changing prod data so let's make sure it's covered with tests.

* mcb - add flag to allow geocode of invalid sites

### Guidance to review

Geocode a single site

    bin/mcb sites geocode 11216495 -v -f

Geocode all the sites without lat long

    bin/mcb sites geocode -v -f

See the update help with new `-f` flag

    mcb sites geocode --help

(`-v` for verbose so you can see what it did)

Use rails console to check sites before/after

```ruby
s = Site.where(latitude: nil).sample
```

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/mcb/commands/sites/geocode.rb
+++ b/lib/mcb/commands/sites/geocode.rb
@@ -3,6 +3,7 @@ summary "Batch geocode Sites"
 usage "geocode [options] [<id>...]"
 option :s, "sleep", "time to sleep between each request", argument: :optional, default: 0.25, transform: method(:Float)
 option :b, "batch_size", "batch size", argument: :optional, default: 100, transform: method(:Integer)
+flag :f, "force", "Geocode even if the rest of the site has validation errors"
 
 instance_eval(&MCB.remote_connect_options)
 
@@ -19,13 +20,13 @@ run do |opts, args, _cmd|
     Site
       .where(id: args.to_a)
       .find_each(batch_size: opts[:batch_size]) do |site|
-        MCB.geocode(obj: site, sleep: opts[:sleep])
+        MCB.geocode(obj: site, sleep: opts[:sleep], force: opts[:force])
       end
   else
     Site
       .not_geocoded
       .find_each(batch_size: opts[:batch_size]) do |site|
-        MCB.geocode(obj: site, sleep: opts[:sleep])
+        MCB.geocode(obj: site, sleep: opts[:sleep], force: opts[:force])
       end
   end
 end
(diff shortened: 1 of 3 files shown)


timabell:
[Geocode "invalid" sites](https://trello.com/c/NQ8UBPgu/2767-geocode-invalid-sites)

lib/mcb/commands/sites/geocode.rb
```diff
@@ -3,6 +3,7 @@
 usage "geocode [options] [<id>...]"
 option :s, "sleep", "time to sleep between each request", argument: :optional, default: 0.25, transform: method(:Float)
 option :b, "batch_size", "batch size", argument: :optional, default: 100, transform: method(:Integer)
+flag :f, "force", "Geocode even if the rest of the site has validation errors"
```
TobyRet: _Suggestion_ "Geocode even if validation errors are present"

lib/mcb/commands/sites/geocode.rb
```diff
@@ -19,13 +20,13 @@
     Site
       .where(id: args.to_a)
       .find_each(batch_size: opts[:batch_size]) do |site|
-        MCB.geocode(obj: site, sleep: opts[:sleep])
+        MCB.geocode(obj: site, sleep: opts[:sleep], force: opts[:force])
```
TobyRet: Will the force option be applied to Providers or will this be in a separate PR?

lib/mcb/commands/sites/geocode.rb
timabell: I wanted to highlight that the validation errors where likely in the address, rather than the lat/long we are trying to update. I'm not really sure how to get it across clearly. If I read "Geocode even if validation errors are present" from cold I would probably assume that the errors were something created by the geo-coder rather than pre-existing errors. Thoughts?

lib/mcb/commands/sites/geocode.rb
timabell: That would be a different PR. This trello card was only for sites.

lib/mcb/commands/sites/geocode.rb
TobyRet: The validation errors may be address related but I also found there were some others that were violations of business rules e.g. duplicate names

PR #1117, opened 2020-01-13
[2766] Add new course serializer for v3

dankmitchell:
### Context
Courses V2 serializer exposing the last enrichment for the frontend

- Create a new serializer for V3 that exposes the last "published" enrichment
- Remove metadata as they aren't needed in V3.

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_serializers_service_v3.rb
+++ b/app/services/course_serializers_service_v3.rb
@@ -1,0 +1,43 @@
+class CourseSerializersServiceV3
+  def initialize(
+    course_serializer: API::V3::SerializableCourse,
+    subject_serializer: API::V2::SerializableSubject,
+    primary_subject_serializer: API::V2::SerializableSubject,
+    secondary_subject_serializer: API::V2::SerializableSubject,
+    modern_languages_subject_serializer: API::V2::SerializableSubject,
+    further_education_subject_serializer: API::V2::SerializableSubject,
+    site_status_serializer: API::V2::SerializableSiteStatus,
+    site_serializer: API::V2::SerializableSite,
+    provider_serializer: API::V2::SerializableProvider,
+    provider_enrichment_serializer: API::V2::SerializableProviderEnrichment,
+    recruitment_cycle_serializer: API::V2::SerializableRecruitmentCycle
+  )
+    @course_serializer = course_serializer
+    @subject_serializer = subject_serializer
+    @primary_subject_serializer = primary_subject_serializer
+    @secondary_subject_serializer = secondary_subject_serializer
+    @modern_languages_subject_serializer = modern_languages_subject_serializer
+    @further_education_subject_serializer = further_education_subject_serializer
+    @site_serializer = site_serializer
+    @site_status_serializer = site_status_serializer
+    @provider_serializer = provider_serializer
+    @provider_enrichment_serializer = provider_enrichment_serializer
+    @recruitment_cycle_serializer = recruitment_cycle_serializer
+  end
+
+  def execute
+    {
+      Course: @course_serializer,
+      Subject: @subject_serializer,
+      PrimarySubject: @primary_subject_serializer,
+      SecondarySubject: @secondary_subject_serializer,
+      ModernLanguagesSubject: @modern_languages_subject_serializer,
+      FurtherEducationSubject: @further_education_subject_serializer,
+      SiteStatus: @site_status_serializer,
+      Site: @site_serializer,
+      Provider: @provider_serializer,
+      ProviderEnrichment: @provider_enrichment_serializer,
+      RecruitmentCycle: @recruitment_cycle_serializer,
+    }
+  end
+end
--- a/spec/serializers/api/v3/serializable_course_spec.rb
+++ b/spec/serializers/api/v3/serializable_course_spec.rb
@@ -1,0 +1,217 @@
+require "rails_helper"
+
+describe API::V3::SerializableCourse do
+  let(:jsonapi_renderer) { JSONAPI::Serializable::Renderer.new }
+  let(:draft_enrichment) { build :course_enrichment }
+  let(:published_enrichment) { build :course_enrichment, :published }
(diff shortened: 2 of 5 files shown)


dankmitchell:
[New course pages showing unpublished content](https://trello.com/c/zVFBGl87/2766-new-course-pages-showing-unpublished-content)

spec/serializers/api/v3/serializable_course_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+    end
+
+    it { should have_relationship(:sites) }
+  end
+
+  context "funding_type" do
+    context "fee-paying" do
+      let(:course) { create(:course) }
+
+      it { expect(subject["attributes"]).to include("funding_type" => "apprenticeship") }
```
defong:
```suggestion
      it { expect(subject["attributes"]).to include("funding_type" => "fee") }
```

app/services/course_serializers_service_v3.rb
```diff
@@ -0,0 +1,43 @@
+class CourseSerializersServiceV3
```
defong: just a passing thought, can this be a factory service or something, default is `v2`?

spec/serializers/api/v3/serializable_course_spec.rb
dankmitchell: Solved.  Copy and Paste fail 🤦‍♂

app/services/course_serializers_service_v3.rb
dankmitchell: Hoping @danielburnley might be able to advise on that

misaka: How about including `CourseSerializersServiceV3` and changing what's different? Or use inheritance, I think either is a valid way to do code reuse here.

misaka: Just for the benefit of anyone else having a look, here's the diff between V2 and V3 of the serialiser. I think it's worth trying to include/extend V2 into V3 and just overriding the methods below, but it'll probably take a little bit of fiddling to make sure `enrichment_attribute` and `meta` are over-ridden correctly.

```diff
modified   Users/misaka/Code/DfE/manage-courses-backend/app/serializers/api/v2/serializable_course.rb -> Users/misaka/Code/DfE/manage-courses-backend/app/serializers/api/v3/serializable_course.rb
@@ -1,10 +1,10 @@
 module API
-  module V2
+  module V3
     class SerializableCourse < JSONAPI::Serializable::Resource
       class << self
         def enrichment_attribute(name, enrichment_name = name)
           attribute name do
-            @object.enrichments.last&.__send__(enrichment_name)
+            @object.enrichments.select(&:published?).last&.__send__(enrichment_name)
           end
         end
       end
@@ -61,20 +61,6 @@ module API
       enrichment_attribute :personal_qualities
       enrichment_attribute :required_qualifications
       enrichment_attribute :salary_details
-
-      meta do
-        serializer_service = SerializeCourseService.new
-        edit_options = @object.edit_course_options
-
-        edit_options[:modern_languages] =
-          serializer_service.execute(object: edit_options[:modern_languages])[:serialized][:data]
-        edit_options[:subjects] =
-          serializer_service.execute(object: edit_options[:subjects])[:serialized][:data]
-
-        {
-          edit_options: edit_options,
-        }
-      end
     end
   end
 end
```

dankmitchell: Is there anything else we intend to remove from V3 after your investigation work @misaka ?

gpeng: Does extending V2 here increase the risk of inadvertently breaking V3 with a change to V2? I wouldn't expect that to happen if I was working on an explicitly versioned serializer. I would expect it to be isolated. Maybe split the common functionality into a separate module and include it in both so it's clearer the code is being used in more than one place?

spec/serializers/api/v3/serializable_course_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+      class: {
+        Course: API::V3::SerializableCourse,
+      },
+    ).to_json
+  end
+  let(:parsed_json) { JSON.parse(course_json) }
+
+  subject { parsed_json["data"] }
+
+  it { should have_type("courses") }
```
gpeng: I think `should` is on it's way to being deprecated if it isn't already. `it { is_expected.to` is the new hotness.

app/services/course_serializers_service_v3.rb
misaka: @gpeng That's a good point. This whole thing is a bit of a hack, since we just created the API endpoint and ran with it. I don't think v3 is the shape we want, there's a bunch of extra stuff in there that we don't want to be exposing, so I'm not sure what to do about that aside from what we're already doing ... vetting the attributes and deciding on what to keep/drop/change.

misaka: I think the best case scenario is that we move to a stable API (v3.1 probably at this point), and then deprecate v3 and remove it as soon as Find and Apply aren't using it any more.

Maybe for now the answer is to duplicate, ship and move onto making v3.1

spec/serializers/api/v3/serializable_course_spec.rb
misaka: Ya. mea culpa, I'm just attached to `should`, but we'll probably have to do a bulk switch over to `is_expected`.

dankmitchell: NIce tip @gpeng. I'll change it here but as @misaka says we should do the entire code base  at some stage

app/services/course_serializers_service_v3.rb
dankmitchell: So @misaka. Merge and Ship? :shipit:

PR #1118, opened 2020-01-13
1592 return courses for a provider that the current provider is the accredited body for

gpeng:
### Context

We are adding functionality to allow an accredited body to list training providers and the courses they deliver.

* Add an endpoint to list the courses at:

`/api/v2/providers/<provider_code>/training_providers/<training_provider_code>/courses`

... where `provider_code` is the accredited body awarding the qualification and `training_provider_code` is the training provider delivering the course.

* Add a scope to `CoursePolicy` to handle authorisation.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


gpeng:
[Return courses for a provider that the current provider is the accredited body for](https://trello.com/c/9PMjxNdt/1592-return-courses-for-a-provider-that-the-current-provider-is-the-accredited-body-for)

app/policies/course_policy.rb
```diff
@@ -1,6 +1,25 @@
...
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      if user.admin?
+        scope.all
+      else
+        scope.where(
+          accrediting_provider_code: user.providers.pluck(:provider_code),
```
defong: im not sure is this correct

a user has a provider (its only this scope that should apply)
a provider has many courses
a course maybe accredited by another provider

defong: image

gpeng: That was from a discussion I had with @misaka on Friday. When I was doing it it seemed like we should make sure that no courses were returned that weren't associated, via the training provider, to the accredited body. That is was happens but only as a result of the implementation, it's not being enforced by the authorisation system. I can remove the scope if we agree it's not necessary. Regarding the line 17 where clause... There isn't any way of knowing in the policy which `accredited_body` the user is viewing from so it seemed 'good enough' ™️ to allow them to see courses that any of their providers are the accrediting body for. The relationship is `has_many` in the code rather than 'has a'.

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      attr_reader :recruitment_cycle, :accredited_body, :training_provider
+
+      def build_recruitment_cycle
+        @recruitment_cycle = RecruitmentCycle.find_by(
+          year: params[:recruitment_cycle_year],
+        ) || RecruitmentCycle.current_recruitment_cycle
+      end
+
+      def build_accredited_body
```
defong: this should be build_provider right?

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        @accredited_body = recruitment_cycle.providers.find_by!(
+          provider_code: params[:provider_code].upcase,
+        )
+      end
+
+      def build_training_provider
+        @training_provider = recruitment_cycle.providers.find_by!(
```
defong: there this can be just 

```ruby
@provider.accrediting_providers.find_by :code :training_provider_code
```

app/policies/course_policy.rb
defong: https://github.com/DFE-Digital/manage-courses-backend/blob/3c2e7a0fe1c92550ee62c4e0e6f9d4ebaad3d9b0/app/models/provider.rb#L82

defong: is that what you was looking for?

gpeng: Discussed ^ off-line. Yes, that is what we need but additionally `current` and by provider id.

spec/policies/course_policy_spec.rb
```diff
@@ -25,4 +25,30 @@
...
+
+    context "user not from the accredited body" do
+      let(:organisation) { create(:organisation) }
+
+      it { is_expected.to be_empty }
+    end
+
+    context "an admin" do
+      let(:user) { create(:user, :admin) }
+      it { is_expected.to contain_exactly(course, other_course) }
```
defong: should the other_course be returned?
as the course is not from `accredited_body`
or from the `training_provider`

given its `CoursePolicy::TrainingProviderCourses`

app/policies/course_policy.rb
gpeng: There may be an issue with line 17 in that it doesn't check that accredited body the user is associated with is current for this recruitment cycle. Open to suggestions on that.

spec/policies/course_policy_spec.rb
gpeng: I don't think I can do that in a scope as it only has the user and the input scope to go on. For an admin user we'd have no reference to what the accredited body is, or the training provider. We're reliant on the input scope doing that and this policy scope just ensuring that if the user isn't an admin the courses are filtered appropriately.

app/policies/course_policy.rb
misaka: Ya, I don't know about this yet. Let's have a chat about it.

spec/requests/api/v2/providers/courses/get_for_accredited_body_by_provider_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  let(:training_provider) { create(:provider) }
+
+  let!(:course1) { create(:course, provider: training_provider, accrediting_provider: accredited_body) }
+  let!(:course2) { create(:course, provider: training_provider) }
+
+  def perform_request
+    get request_path, headers: { "HTTP_AUTHORIZATION" => credentials }
+  end
+
+  let(:request_path) { "/api/v2/recruitment_cycles/#{recruitment_cycle.year}/providers/#{accredited_body.provider_code}/training_providers/#{training_provider.provider_code}/courses" }
```
defong:
```suggestion
  let(:request_path) { "/api/v2/recruitment_cycles/#{recruitment_cycle.year}/providers/#{provider.provider_code}/training_providers/#{training_provider.provider_code}/courses" }
```

given the endpoint is for `provider` it reads a bit better to say that a provider has training provider... then jumping to `accredited_body` 

really just naming of things

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
```diff
@@ -0,0 +1,37 @@
+module API
+  module V2
+    class AccreditedBodyTrainingProviderCoursesController < ApplicationController
+      before_action :build_recruitment_cycle, :build_accredited_body, :build_training_provider
+
+      def index
+        authorize accredited_body, :can_list_training_providers?
```
defong:
```suggestion
        authorize provider, :can_list_training_providers?
```

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      attr_reader :recruitment_cycle, :accredited_body, :training_provider
+
+      def build_recruitment_cycle
+        @recruitment_cycle = RecruitmentCycle.find_by(
+          year: params[:recruitment_cycle_year],
+        ) || RecruitmentCycle.current_recruitment_cycle
+      end
+
+      def build_accredited_body
```
defong:
```suggestion
      def build_provider
```

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+        @accredited_body = recruitment_cycle.providers.find_by!(
```
defong:
```suggestion
        @provider = recruitment_cycle.providers.find_by!(
```

gpeng:
@defong & @misaka This has been totally reworked now if you could have another look.

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
gpeng: I think the naming is important in here as the controller is dealing with two different providers so `accredited_body` & `training_provider` makes the distinction. Although they are both instances of `Provider` the distinction is important in this context.

defong: https://github.com/DFE-Digital/teacher-training-api/blob/624e1b77dcca3ef4d8794c687e47aece3347fa19/app/controllers/api/v2/accredited_provider_training_providers_controller.rb#L22-L26

something feels off with naming things

Accredited`Body`TrainingProviderCoursesController
Accredited`Provider`TrainingProvidersController

can we settle on `accredited_provider`

gpeng: I did settle on that in my last PR but since then the decision was made to settle on accredited body. I've got a ticket in my backlog to make it consistent throughout which I'll be pushing for the next sprint. I can change the `...Provider` one in this PR but there are references accredited|accrediting provider all over the place that still need fixing.

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
```diff
@@ -0,0 +1,37 @@
+module API
+  module V2
+    class AccreditedBodyTrainingProviderCoursesController < ApplicationController
+      before_action :build_recruitment_cycle, :build_accredited_body, :build_training_provider
+
+      def index
+        authorize accredited_body, :can_list_training_providers?
+
+        courses = policy_scope(Course).where(provider_id: training_provider.id)
```
misaka: Just an aside, not a blocker, but ... after running into all the slow Apply requests, I think we need to focus a bit more on performance. There are not includes being done here so it worries me a bit, we'll have to think of a generalised approach that can be ported to other endpoints.

app/policies/course_policy.rb
```diff
@@ -1,6 +1,25 @@
...
+      @scope = scope
+    end
+
+    def resolve
+      if user.admin?
+        scope.all
+      else
+        scope
+          .where(provider_id: user.providers.pluck(:id))
+          .or(scope.where(accrediting_provider_code: user.providers.pluck(:provider_code)))
```
misaka: I'm not sure that `scope.where(...)` is correct. My understanding is that you need a plain AREL, and not one that already has conditions on it. For example, Course.where(

Here's a quick test:

```ruby
[6] pry(main)> scitts = Course.scitt_programme
[8] pry(main)> scitts.or(scitts.full_time).to_sql
=> "SELECT \"course\".* FROM \"course\" WHERE \"course\".\"program_type\" = 'SC'"
[9] pry(main)> scitts.or(Course.full_time).to_sql
=> "SELECT \"course\".* FROM \"course\" WHERE (\"course\".\"program_type\" = 'SC' OR \"course\".\"study_mode\" = 'F')"
```

In the code above, `scope` is `Course` so the problem doesn't manifest, but I imagine that if we passed in an AREL that's already been scoped, we may get the wrong results.

spec/policies/course_policy_spec.rb
```diff
@@ -25,4 +25,39 @@
...
+      let(:organisation) { create(:organisation) }
+
+      it { is_expected.to be_empty }
+    end
+
+    context "a user from the provider" do
+      before do
+        accredited_body.organisations.clear
+        training_provider.organisations << organisation
+      end
```
misaka: How about this instead of the pockery-jiggery with the organisations?

```
      let(:user) { training_provider.organisations.first.users.first }
```

spec/requests/api/v2/providers/courses/get_for_accredited_body_by_provider_spec.rb
misaka: Sorry, but I like the added context given by calling it `accredited_body` in this particular case.

spec/requests/api/v2/providers/courses/get_for_accredited_body_by_provider_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      parsed_response = JSON.parse(response.body)["data"]
+      expect(parsed_response.count).to eq(1)
+      expect(parsed_response.first["id"]).to eq(course1.id.to_s)
+    end
+  end
+
+  context "two recruitment cycle" do
+    let(:next_recruitment_cycle) { create(:recruitment_cycle, year: (recruitment_cycle.year.to_i + 1).to_s) }
+    let(:next_training_provider) { create(:provider, recruitment_cycle: next_recruitment_cycle) }
+    let(:next_course) { create(:course, provider: next_training_provider, accrediting_provider: accredited_body) }
```
misaka: `next_recruitment_cycle` should use `find_or_create`, and we have a slight shortcut for a provider in the next recruitment cycle:

```ruby
    let(:next_recruitment_cycle) { find_or_create(:recruitment_cycle, :next) }
    let(:next_training_provider) { create(:provider, :next_recruitment_cycle) }
    let(:next_course) { create(:course, provider: next_training_provider, accrediting_provider: accredited_body) }
```

spec/requests/api/v2/providers/courses/get_for_accredited_body_by_provider_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    let(:request_path) { "/api/v2/recruitment_cycles/#{next_recruitment_cycle.year}/providers/#{next_accredited_body.provider_code}/training_providers/#{next_training_provider.provider_code}/courses" }
+    let(:next_accredited_body) { create(:provider, :accredited_body, organisations: [organisation], recruitment_cycle: next_recruitment_cycle) }
```
misaka: As above, we can use the trait here:

```ruby
    let(:next_accredited_body) { create(:provider, :accredited_body, organisations: [organisation], :next_recruitment_cycle) }
```

spec/requests/api/v2/providers/courses/get_for_accredited_body_by_provider_spec.rb
misaka: btw, how is `next_training_provider` working here ... it's creating it's own organisation, when it should be setting it to `organisation`.

```ruby
    let(:next_training_provider) { create(:provider, :next_recruitment_cycle, organisations: [organisation]) }
```

But then, how is this test working ...

misaka: Ah, @gpeng straightened me out. We don't care that the user be associated with `next_training_provider`, so no need to associate it with `organisation`. :)

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
gpeng: We should benchmark this stuff and get some facts to base that on. I've had a quick look at `SerializableCourse` I don't think we get N+1 anywhere, we just get a few more DB queries. Having 4 or 5 queries might be quicker than trying to include everything here (and of course it might not but it's probably negligible). Might be a good idea to have a maximum response time that we're aiming for on V3.

spec/policies/course_policy_spec.rb
gpeng: I've fixed this but couldn't quite do it the way you suggested. `training_provider.organisations.first.users` is empty so I added a new user and organisation in that context.

gpeng:
@misaka I've made the changes you suggested and left a comment about the performance issue you mentioned. Hopefully this should be good to go now. 🤞

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
misaka: Similar to a suggestion below, I'm voting for `accredited_body` because I think it adds useful context.

spec/requests/api/v2/providers/courses/get_for_accredited_body_by_provider_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      perform_request
+
+      parsed_response = JSON.parse(response.body)["data"]
+      expect(parsed_response.count).to eq(1)
+      expect(parsed_response.first["id"]).to eq(course1.id.to_s)
+    end
+  end
+
+  context "two recruitment cycle" do
+    let(:next_recruitment_cycle) { create(:recruitment_cycle, :next) }
```
misaka: `find_or_create` pwetty pulleeeezzz

spec/requests/api/v2/providers/courses/get_for_accredited_body_by_provider_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    let(:next_training_provider) { create(:provider, recruitment_cycle: next_recruitment_cycle) }
```
misaka: And the trait please?

app/controllers/api/v2/accredited_body_training_provider_courses_controller.rb
misaka: See below comments about `provider` vs `accredited_body`.

PR #1120, opened 2020-01-14
[2801] /subject-areas endpoint

faissalMT:
### Context
In attempting to create the subjects filter on Rails Find we discovered that subject areas as presented in Search and Compare did not exist in the Teacher Training API.

Add the subject areas model, an index endpoint for it, and populate the database with the appropriate subject areas.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/subject_areas_controller.rb
+++ b/app/controllers/api/v3/subject_areas_controller.rb
@@ -1,0 +1,9 @@
+module API
+  module V3
+    class SubjectAreasController < API::V3::ApplicationController
+      def index
+        render jsonapi: SubjectArea.active, fields: fields_param, include: params[:include], class: CourseSerializersService.new.execute[:v3]
+      end
+    end
+  end
+end
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -2,21 +2,24 @@
 #
 # Table name: subject
 #
-#  created_at   :datetime
-#  id           :bigint           not null, primary key
-#  subject_code :text
-#  subject_name :text
-#  type         :text
-#  updated_at   :datetime
+#  created_at      :datetime
+#  id              :bigint           not null, primary key
+#  subject_area_id :bigint
+#  subject_code    :text
+#  subject_name    :text
+#  type            :text
+#  updated_at      :datetime
 #
 # Indexes
 #
-#  index_subject_on_subject_name  (subject_name)
+#  index_subject_on_subject_area_id  (subject_area_id)
+#  index_subject_on_subject_name     (subject_name)
 #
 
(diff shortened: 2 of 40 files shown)


faissalMT:
[Add v3 /subject-areas endpoint](https://trello.com/c/hcDmJwrA/2801-add-v3-subject-areas-endpoint)

config/routes.rb
```diff
@@ -183,6 +184,9 @@
           end
         end
       end
+      resources :subject_areas,
+                only: :index do
```
misaka: Pourquoi un bloc? Débloque!

app/models/subject_area.rb
```diff
@@ -0,0 +1,15 @@
...
+# Table name: subject_area
+#
+#  created_at :datetime         not null
+#  name       :text
+#  typename   :text             not null, primary key
+#  updated_at :datetime         not null
+#
+
+class SubjectArea < ApplicationRecord
+  has_many :subjects, foreign_key: :type, inverse_of: :courses
```
misaka: `inverse_of: courses` ... doesn't look right.

app/models/subject.rb
```diff
@@ -2,21 +2,24 @@
...
 #
-#  index_subject_on_subject_name  (subject_name)
+#  index_subject_on_subject_area_id  (subject_area_id)
+#  index_subject_on_subject_name     (subject_name)
 #
 
 class Subject < ApplicationRecord
   has_many :course_subjects
   has_many :courses, through: :course_subjects
+  belongs_to :subject_area, foreign_key: :type, inverse_of: :subjects
```
misaka: I don't see why this needs an `inverse_of`, and as demonstrated below, why introduce an additional place where something can be done wrong.

```suggestion
  belongs_to :subject_area, foreign_key: :type
```

app/models/subject_area.rb
misaka:
```suggestion
  has_many :subjects, foreign_key: :type
```

app/controllers/api/v3/subject_areas_controller.rb
```diff
@@ -0,0 +1,9 @@
+module API
+  module V3
+    class SubjectAreasController < API::V3::ApplicationController
+      def index
+        render jsonapi: SubjectArea.active, fields: fields_param, include: params[:include], class: CourseSerializersService.new.execute[:v3]
```
misaka: Is the idea we'll just use `include: :subjects` to get the whole chain in one request?

faissalMT: Yes

PR #1122, opened 2020-01-14
[2804] - Centralise geocoding logic

TobyRet:
### Context
MCB's geocoding logic did not restrict searches to the UK region. This meant that geocoding of any invalid Sites (specifically those without an address) would result in coordinates set to somewhere in the U.S (this is the default behaviour).

- Extract geocoding logic to a new service. Geocoding is restricted to the UK (gb).
- Update mcb console and background job to use this new service

Centralising this logic will (hopefully) reduce the chances of bugs resulting from diverging paths in future 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -63,7 +63,7 @@ class Site < ApplicationRecord
   def full_address
     address = [address1, address2, address3, address4, postcode]
 
-    unless location_name.downcase == MAIN_SITE
+    unless location_name && location_name.downcase == MAIN_SITE
       address.unshift(location_name)
     end
 
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -63,7 +63,7 @@ class Site < ApplicationRecord
   def full_address
     address = [address1, address2, address3, address4, postcode]
 
-    unless location_name && location_name.downcase == MAIN_SITE
+    unless location_name.downcase == MAIN_SITE
       address.unshift(location_name)
     end
 
--- a/app/services/geocoder_service.rb
+++ b/app/services/geocoder_service.rb
@@ -1,0 +1,12 @@
+class GeocoderService
+  def self.geocode(obj:, force: false)
+    results = Geocoder.search(obj.full_address, params: { region: "gb" })
+
+    if results.present?
+      result = results.first
+      obj.latitude = result.latitude
+      obj.longitude = result.longitude
+      obj.save(validate: !force)
+    end
+  end
+end
--- a/app/services/geocoder_service.rb
+++ b/app/services/geocoder_service.rb
@@ -6,7 +6,7 @@ class GeocoderService
       result = results.first
       obj.latitude = result.latitude
       obj.longitude = result.longitude
-      obj.save(validate: !force)
+      obj.save!(validate: !force)
     end
   end
 end
--- a/lib/mcb.rb
+++ b/lib/mcb.rb
@@ -478,12 +478,14 @@ module MCB
     end
 
     def geocode(obj:, sleep:, force: false)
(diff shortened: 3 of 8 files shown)


TobyRet:
[MCB geocode commands need to be restricted to UK](https://trello.com/c/rUbLtL6K/2804-mcb-geocode-commands-need-to-be-restricted-to-uk)

TobyRet:
Cache will need to be expired before re-geocoding
https://github.com/alexreisner/geocoder#caching

lib/mcb.rb
```diff
@@ -478,12 +478,22 @@ def launch_repl?(argv)
...
-              "New lat/long #{obj.latitude},#{obj.latitude} for full_address '#{obj.full_address}'"
+      results = Geocoder.search(obj.full_address, params: { region: "gb" })
+
+      if results.present?
+        result = results.first
+        obj.latitude = result.latitude
+        obj.longitude = result.longitude
+        saved = obj.save(validate: !force)
+
+        warn "Updating failed for #{obj.class}:#{obj.id} - #{obj}. Error: #{obj.errors}" unless saved
```
timabell: I think move this into an `else` for the `if` below

lib/mcb.rb
```diff
@@ -478,12 +478,22 @@ def launch_repl?(argv)
...
+        if saved
+          verbose "Geocoded #{obj.class}:#{obj.id} - #{obj}. " \
+                "New lat/long #{obj.latitude},#{obj.latitude} for full_address '#{obj.full_address}'"
```
timabell:
```suggestion
                "New lat/long #{obj.latitude},#{obj.longitude} for full_address '#{obj.full_address}'"
```

spec/services/geocoder_service_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+          to raise_error(ActiveRecord::RecordInvalid)
+    end
+
+    it "geocodes an invalid object if forced" do
+      expect { GeocoderService.geocode(obj: invalid_site, force: true) }.
+          to change { invalid_site.reload.latitude }.from(nil).to(51.4524877).
+              and change { invalid_site.longitude }.from(nil).to(-0.1204749)
+    end
+
+    it "geocodes UK (gb) addresses only" do
```
TobyRet: @defong - I hope you sleep easy tonight 😄

app/models/site.rb
```diff
@@ -63,7 +63,7 @@ def needs_geolocation?
   def full_address
     address = [address1, address2, address3, address4, postcode]
 
-    unless location_name.downcase == MAIN_SITE
+    unless location_name && location_name.downcase == MAIN_SITE
```
defong: why was this changed?

app/services/geocoder_service.rb
```diff
@@ -0,0 +1,12 @@
+class GeocoderService
+  def self.geocode(obj:, force: false)
+    results = Geocoder.search(obj.full_address, params: { region: "gb" })
+
+    if results.present?
+      result = results.first
+      obj.latitude = result.latitude
+      obj.longitude = result.longitude
+      obj.save(validate: !force)
```
defong: should it be the service that save its?

spec/services/geocoder_service_spec.rb
defong: just wanna make sure you snooze like a cuddly :baby:...

app/models/site.rb
TobyRet: Sorry this was an error. Changing back

PR #1125, opened 2020-01-15
[2809] add ElastAlert definition for slow requests

misaka:
### Context

After the recent outage we had, I've taken a look into how to get alerts out of logit.i to let us know that we're getting slow requests. I've created an ElastAlert there that will trigger if the average request time is over 2.5s over a period of 60 seconds.

Add the ElastAlert YAML definition to the repo so we have it. Note that sensitive information has been stripped out of the definition and have to be re-added.

### Guidance to review

A sample of the alert is here: https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1579091158144800

This alert is already deployed.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/etc/ttapi-slow-response-times.yaml
+++ b/etc/ttapi-slow-response-times.yaml
@@ -1,0 +1,47 @@
+# This defines an ElastAlert alert that triggers when we detect a certain number
+# of slow requests, averaged over a period of time.
+#
+# You will need a to copy in the "Kibana Dashboard URL" for the "Publish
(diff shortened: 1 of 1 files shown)


misaka:
[Create slow request alert for TTAPI](https://trello.com/c/SS1E38JK/2809-create-slow-request-alert-for-ttapi)

PR #1135, opened 2020-01-21
2816: Redesign the TestSetup API

TobyRet:
### Context
Further to the work undertaken here https://github.com/DFE-Digital/teacher-training-api/pull/1097 a discussion with devs followed and it was agreed that the TestSetup API should be changed to mirror, more closely, that of FactoryBot.

- Simplification of the TestSetup API and test coverage
- Rather predefining variable names for cached factory instances, user now simply call a new instance method as follows:
`get(factory_type, [traits])`

### Guidance to review
This is the second iteration so as always, keen to get feedback. Hopefully the changes are heading the right direction and the API is now easier to use.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
Will clean up the comments once approved

TobyRet:
[[S] Redesign the TestSetup API](https://trello.com/c/Zlnmu1ku/2816-s-redesign-the-testsetup-api)

spec/factories/courses.rb
```diff
@@ -210,7 +210,7 @@
       to_create { |instance| instance.save(validate: false) }
     end
 
-    trait :unpublished_with_primary_maths do
```
timabell: 1. Did you check there were no usages of this trait before renaming it?
2. Do we need to rename it? It would be a cleaner patch with the rename.

spec/models/provider_spec.rb
```diff
@@ -268,7 +268,8 @@
   end
 
   describe "courses" do
-    let(:course) { TestSetup.unpub_pri_math }
+    let(:test_setup) { TestSetup.new }
+    let(:course) { test_setup.get(:course, :primary, :unpublished) }
```
timabell: I think rename `test_setup` to `test_data`

timabell: :heart: love the new signature :+1:

spec/support/test_setup.rb
```diff
@@ -2,118 +2,54 @@
...
-      @@courses&.key?(method_name) ||
-      super
-  end
-
   attr_reader :courses, :recruitment_cycles
 
-  def initialize(with_courses: nil)
+  def initialize
     @recruitment_cycles = @@recruitment_cycles.transform_values(&:call)
     @courses = @@courses.transform_values(&:call)
```
timabell: I'd think I'd extract a named method for these two lines. `def CreateAndCacheTestRecords`

spec/factories/courses.rb
```diff
@@ -210,7 +210,7 @@
...
+    trait :unpublished do
       transient do
         identifier { "unpublished_with_primary_maths" }
```
defong: whats with primary maths?

spec/support/test_setup.rb
```diff
@@ -2,118 +2,54 @@
...
   }
 
   @@courses = {
-    unpub_pri_math: ->(attributes = {}) do
-      find_or_create(:course, :unpublished_with_primary_maths, attributes)
-    end,
+      [:primary, :unpublished] => ->() do
+        FactoryBot.find_or_create(:course, :primary, :unpublished)
+      end
   }
```
faissalMT: Why not make these methods with switches? It doesn't seem like you're passing around the procs here and it might better reflect what you're trying to do.

spec/support/test_setup.rb
```diff
@@ -2,118 +2,54 @@
...
-      @courses[method_name]
+  def get(name, *traits)
+    case name
+    when :course
+      @courses[traits.sort] || raise(
+          <<~ERR_MSG
+        No predefined course for these traits: #{traits}.
+        Either add it to test_setup.rb or if it's used frequently, just create
+        a FactoryBot factory instance.
+      ERR_MSG
```
timabell: I think I'd DRY the `raise`

```ruby
def UnknownTraits(type_name, traits)
    raise(
          <<~ERR_MSG
        No predefined #{type_name} for these traits: #{traits}.
        Either add it to `test_setup.rb` to take advantage of the caching,
        or create a FactoryBot factory instance directly for one-offs.
      ERR_MSG
end
```

Also iterated on the message

spec/support/test_setup.rb
```diff
@@ -2,118 +2,54 @@
 class TestSetup
```
timabell: I'm thinking we should rename this to `TestDataCache`

spec/support/test_setup.rb
```diff
@@ -2,118 +2,54 @@
...
   # A helper class for instantiating a standard set of factories for testing
```
timabell:
```suggestion
  # A helper class for pre-creating test data with FactoryBot factories
  # and maintaining a cache of these objects
  # to avoid the performance problems of
  # repeatedly asking FactoryBot to create trees of objects.
  # Refs: https://evilmartians.com/chronicles/testprof-2-factory-therapy-for-your-ruby-tests-rspec-minitest
```

spec/test_setup_specs/test_setup_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+describe 'TestSetupSpec' do
+  before(:all) do
+    @test_setup = TestSetup.new
+  end
+
+  it 'finds trait pre-defined in test setup [:primary, :unpublished]' do
+    course = @test_setup.get(:course, :primary, :unpublished)
+
+    expect(course).to eq(Course.first)
```
timabell: Will this survive as we pre-cache more variations of course?

timabell: I think maybe we should just test that it does return a course and that it has the right properties

spec/factories/courses.rb
TobyRet: This was one that I created in iteration 1

spec/models/provider_spec.rb
TobyRet: 👍

spec/support/test_setup.rb
TobyRet: 🤔 mmm ... I prefer to declare instance vars in the initializer if possible as I think alternative creates more in-direction. Happy to hear opinions on this though

spec/support/test_setup.rb
TobyRet: 👍

spec/support/test_setup.rb
TobyRet: Yeah that seems reasonable. Or more more simply 'TestData'? What do others think?

spec/support/test_setup.rb
TobyRet: 👍

spec/test_setup_specs/test_setup_spec.rb
TobyRet: Yeah agree 👍

spec/factories/courses.rb
TobyRet: Fixing ta!

spec/factory_specs/course_spec.rb
```diff
@@ -31,8 +34,8 @@
     end
   end
 
-  context "unpublished_with_primary_maths" do
-    subject { create(:course, :unpublished_with_primary_maths) }
+  context "unpublished" do
+    subject { create(:course, :unpublished) }
```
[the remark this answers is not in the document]
TobyRet: The `:unpublished` trait is also creating sites and associating the same provider with the course and sites. 

We still need to agree what test data is made available through the TestDataCache.

spec/support/test_setup.rb
timabell: Resolved by renaming the factory collections to make it clearer what they are for :+1:

spec/factory_specs/course_spec.rb
```diff
@@ -5,6 +5,9 @@
 
   it { should be_instance_of(Course) }
   it { should be_valid }
+  it "has the correct number of associations" do
```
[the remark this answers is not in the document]
timabell: > shouldn't it be more like this?

no because the default course factory doesn't add sites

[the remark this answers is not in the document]
defong: can we have a test to denote that then, it just look so similar...

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,52 @@
...
+  # It does not matter what order the traits are when passed as arguments.
+
+
+  class << self
+    def get(name, *traits)
+      case name
+      when :course
+        @@courses[traits.sort] || raise_trait_error(:course, traits)
+      when :recruitment_cycle
+        @@recruitment_cycles[traits.sort] || raise_trait_error(:recruitment_cycle, traits)
```
misaka: Don't still need this.

spec/test_setup_specs/test_data_cache_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rspec"
+
+describe "TestSetupSpec" do
```
misaka:
```suggestion
describe TestDataCache do
```

config/settings/test.yml
```diff
@@ -3,3 +3,4 @@ govuk_notify:
   welcome_email_template_id: please_change_me
 system_authentication_token: "Ge32"
 gcp_api_key: please_change_me
+test_setup: false
```
misaka: I don't really feel like this is an app concern ... tests should only ever be run in the `testing` Rails env, for example. I think a bespoke env var would be more appropriate, no?

spec/factories/courses.rb
```diff
@@ -210,12 +210,12 @@
...
 
-    trait :unpublished_with_primary_maths do
+    trait :unpublished do
       transient do
-        identifier { "unpublished_with_primary_maths" }
+        identifier { "unpublished" }
       end
 
-      name { "#{identifier} course name" }
+      name { "#{identifier}_course" }
```
misaka: This is what I was talking about before, btw, about using the string "name" in here so that we can trace how this value is used through the system. I think what was here before is right:

```suggestion
      name { "#{identifier} course name" }
```

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,55 @@
...
+class TestDataCache
+  #  A helper class for pre-creating test data with FactoryBot factories
+  #  and maintaining a cache of these objects to avoid the performance problems
+  #  of repeatedly asking FactoryBot to create trees of objects.
+  #
+  #  Refs: https://evilmartians.com/chronicles/testprof-2-factory-therapy-for-your-ruby-tests-rspec-minitest
+  #
+  # The test data and can be accessed as follows:
+  #
+  # TestSetupCache.get(:course, :primary, :unpublished)
```
misaka:
```suggestion
  # TestDataCache.get(:course, :primary, :unpublished)
```

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,55 @@
...
+  # Note that a course with the requested traits needs to be pre-defined in this class.
+  # It does not matter what order the traits are when passed as arguments.
+
+
+  class << self
+    def get(name, *traits)
+      case name
+      when :course
+        if @@courses == nil
+          # cache not populated, fall back to factory (while we migrate all our tests)
```
misaka: Would it make more sense to test this on a factory-by-factory basis? Something like the following, replacing the `if @@courses == nil` check above.

```ruby
          sorted_traits = traits.sort
          unless @@courses.has_key? sorted_traits
            if @@course_factories.has_key? sorted_traits
              @@courses[sorted_traits] = @@course_factories[sorted_traits].call
            else
              raise_trait_error(:course, traits)
            end
          end
          @@courses[sorted_traits]
```

If you agree, it's probably worth putting most of the above in a separate function.

config/settings/test.yml
TobyRet: 👍

spec/factories/courses.rb
TobyRet: 👍

spec/support/test_data_cache.rb
TobyRet: 👍

spec/support/test_data_cache.rb
TobyRet: I'm don't think this is necessary ...
Unless the caching is switched on then `@@courses` will always be nil. Therefore it probably makes sense to simply return the factory.
What do you think @timabell?

spec/test_setup_specs/test_data_cache_spec.rb
TobyRet: 👍

spec/support/test_data_cache.rb
timabell: What @TobyRet said.

Because the behaviour is:

* The cache can be turned off/on with the env
* If it's on, then the cache is filled in advance so no need to do granular testing
* If it's off, we bypass the cache entirely and fall back to the factories, so again no need to test what is cached

Having said that, this isn't immediately obvious from the code as written, I think maybe we should add a `cachePopulated` field to the class and set that in the setup method and then update this `if` to use it for clarity

TobyRet: ... and / or we can check if the cache is actually turned on? `ENV['TEST_DATA_CACHE']`

timabell: As I mentioned on the phone, that make the behaviour unnecessarily affected by external state.

I'm having a go at implementing this but my first attempt broke something...

timabell: Refactored, all green. Should be clearer now.

timabell:
Just waiting for @misaka 's :+1: now
:pray:

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  class << self
+    def get(name, *traits)
+      case name
+      when :course
+        if @@cache_populated
+          @@courses[traits.sort] || raise_trait_error(:course, traits)
+        else
+          # fall back to factory (while we migrate all our tests)
+          FactoryBot.create(name, *traits)
```
misaka: We should be used the factory as defined in `@@course_factories` so we have a single source of truth for how to create this cached data.

```suggestion
          @@course_factories[traits.sort].call
```

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,62 @@
...
+        end
+      else
+        raise "Unknown model type #{name}: - You need to add this to TestSetup or use a standard FactoryBot factory."
```
misaka: Not critical, but it wouldn't hurt to have the traits here too.

spec/support/test_data_cache.rb
misaka: Looks good to me.

misaka:
@gpeng

> I still think having to create and recreate large trees of data so often is a smell that too much of the code is coupled to the DB and we're going too heavy on high level tests and not writing code that can be more easily tested in isolation with less reliance on high level tests.

Thing is, I'd rather we rely more on high level tests, and we have a setup that's well understood and common to all the tests. I think that ultimately, the value of a unit test approaches zero with time, whereas the higher level tests generally keep their value for the life of the product.

misaka:
@gpeng 

> Also, I've never seen factory behaviour being tested either.

Ya, I know that that's not common, but I've seen factories not behave in the way their intended before, which is why I suggested we add a few.

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  private
+
+    @@cache_populated = false
+    @@courses = nil
+    @@course_factories = {
+        %i[primary unpublished] => -> do
+          FactoryBot.find_or_create(:course, :primary, :unpublished)
+        end,
+    }
```
misaka: Move to the top for visibility, they're private as a function of being `@@` variables.

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,62 @@
...
+      @@cache_populated = false
+    end
+
+  private
+
+    @@cache_populated = false
+    @@courses = nil
+    @@course_factories = {
+        %i[primary unpublished] => -> do
+          FactoryBot.find_or_create(:course, :primary, :unpublished)
```
misaka: It's probably worth sticking with `create` here, since this factory should always create (i.e. the db should be empty).

spec/support/test_data_cache.rb
TobyRet: 👍

spec/support/test_data_cache.rb
TobyRet: 👍

spec/support/test_data_cache.rb
TobyRet: 👍

spec/support/test_data_cache.rb
TobyRet: 👍

spec/support/test_data_cache.rb
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    def get(name, *traits)
+      case name
+      when :course
+        if @@cache_populated
+          @@courses[traits.sort] || raise_trait_error(:course, traits)
+        else
+          # fall back to factory (while we migrate all our tests)
+          @@course_factories[traits.sort].call
```
timabell: What does failure look like if you try and call this with unregistered traits? Would it be easy to figure out what to do to fix it?

timabell: I tried it, it's a nill error :cry: 

How about

```suggestion
          if @@course_factories.key?(traits.sort)
              @@course_factories[traits.sort].call
          else
              raise_model_error(name, traits)
          end
```

timabell:
much better, wonder why the CI failed...

db/seeds.rb
```diff
@@ -3,6 +3,8 @@
 
 CourseSubject.destroy_all
 Course.destroy_all
+FinancialIncentive.destroy_all
+SubjectArea.destroy_all
```
misaka: Should these be here?

app/models/subjects/secondary_subject.rb
```diff
@@ -20,4 +20,8 @@ class SecondarySubject < Subject
   def self.modern_languages
     @modern_languages ||= find_by(subject_name: "Modern Languages")
   end
+
+  def self.clear_modern_languages_cache
+    @modern_languages = nil
+  end
```
misaka: Should this be here?

spec/seed_spec.rb
```diff
@@ -0,0 +1,10 @@
+describe "Seed" do
+  before do
+    SecondarySubject.clear_modern_languages_cache
+  end
+
+  it "seeds without error" do
+    # https://stackoverflow.com/questions/38483820/testing-successful-seed-with-minitest-rspec
+    load Rails.root.join("db/seeds.rb")
+  end
+end
```
misaka: Should this be here?

timabell:
squashed in linter fixes

PR #1144, opened 2020-01-27
[2861] Add search to V3 Provider Index

danielburnley:
### Context

In order to support searching by providers on Rails Find - we need the endpoint in order to search for a specific provider by name or code

Additionally, as this endpoint will be getting hit frequently (potentially once ever key press) it needs to avoid slow performance by accidentally loading more than required, as such sparse field support has been added (https://jsonapi.org/format/#fetching-sparse-fieldsets)

- Add the ability to support a search param at  `v3/.../providers?search=Meow`
- Add the ability to define the fields you wish to return 
  - E.g. `v3/.../providers?fields[providers]=provider_name` for returning only the `provider_name`
  - **Note** This currently defaults to `provider_name`, `provider_code` and `recruitment_cycle_year` to avoid breaking anything that currently depends on the end point

### Guidance to review

- Run the API
- Navigate to localhost:3001/api/v3/recruitment_cycles/2020/providers?search=scitt&fields[providers]=provider_code
- Edit params as you like

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/providers_controller.rb
+++ b/app/controllers/api/v3/providers_controller.rb
@@ -4,6 +4,8 @@ module API
       before_action :build_recruitment_cycle
 
       def index
+        build_fields_for_index
+
         @providers = if params[:search].present?
                        @recruitment_cycle.providers.search_by_code_or_name(params[:search])
                      else
@@ -28,6 +30,22 @@ module API
                                        email website telephone train_with_us
                                        train_with_disability sites
                                        accredited_bodies accredited_body?] }
+      end
+
+    private
+
+      def build_fields_for_index
+        @fields = default_fields_for_index
+
+        return if params[:fields].blank? || params[:fields][:providers].blank?
+
+        @fields[:providers] = params[:fields][:providers].split(",")
+      end
+
+      def default_fields_for_index
+        {
+          providers: %w[provider_name provider_code recruitment_cycle_year],
+        }
       end
     end
   end
--- a/app/controllers/api/v3/providers_controller.rb
+++ b/app/controllers/api/v3/providers_controller.rb
@@ -4,9 +4,13 @@ module API
       before_action :build_recruitment_cycle
 
       def index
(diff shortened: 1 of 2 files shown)


danielburnley:
https://trello.com/c/E5t01cd4/2861-add-provider-search-endpoint-to-v3

app/controllers/api/v3/providers_controller.rb
```diff
@@ -25,6 +31,22 @@ def show
                                        train_with_disability sites
                                        accredited_bodies accredited_body?] }
       end
+
+    private
+
+      def build_fields_for_index
+        @fields = default_fields
+
+        return if params[:fields].blank?
```
misaka: Can we combine the guard below with this one here to reduce the cyclomatic complexity pls?

app/controllers/api/v3/providers_controller.rb
```diff
@@ -25,6 +31,22 @@ def show
...
+
+        @fields[:providers] = params[:fields][:providers].split(",") if params[:fields][:providers].present?
+      end
+
+      def default_fields
```
misaka: Can we match the name of this method: `default_fields_for_index`

spec/requests/api/v3/providers/index_spec.rb
```diff
@@ -113,4 +116,86 @@ def perform_request
...
+        perform_request
+
+        expect(json_response["data"].count).to eq(2)
+        expect(json_response["data"].first).to have_attribute("provider_code").with_value("1AT")
+        expect(json_response["data"].last).to have_attribute("provider_code").with_value("2AT")
+      end
+    end
+
+    context "Seaching for a provider by its provider code" do
+      let(:request_path) { "#{base_provider_path}?search=2AT" }
```
misaka: Do we also need to test searching using lower case too?

danielburnley:
Rebasing master & merging :tada:

PR #1149, opened 2020-01-31
2827 add filtering for courses to api

gpeng:
### Context

Find needs to be able to display a list of currently available courses and filter them by various filters.

* Add scopes to `Course` to enable filtering
* Add a `CourseSearch` service that builds a `Course` scope based on the provided filter parameters.
* Add `/api/v3/recruitment_cycle/<year>/courses` endpoint
* Add indexes to `course_site` table for the enums on `SiteStatus`

### Guidance to review

Filter params are passed in the following format:

`/api/v3/recruitment_cycle/<year>/courses?filter[salary]=true&filter[qualifications]=pgce,pgde`

Out of scope in this PR

* Only returning 'published' courses
* Location filtering

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/course_searches_controller.rb
+++ b/app/controllers/api/v3/course_searches_controller.rb
@@ -1,0 +1,10 @@
+module API
+  module V3
+    class CourseSearchesController < API::V3::ApplicationController
+      def index
+        course_scope = CourseSearchService.call(filter: params[:filter])
+        render jsonapi: paginate(course_scope)
+      end
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -91,6 +91,7 @@
 #                                                                        PUT    /api/v2/access_requests/:id(.:format)                                                                                                    api/v2/access_requests#update
 #                                                                        DELETE /api/v2/access_requests/:id(.:format)                                                                                                    api/v2/access_requests#destroy
 #                                                api_v2_build_new_course GET    /api/v2/build_new_course(.:format)                                                                                                       api/v2/courses#build_new
+#                                                         api_v3_courses GET    /api/v3/courses(.:format)                                                                                                                api/v3/course_searches#index
 #                              api_v3_recruitment_cycle_provider_courses GET    /api/v3/recruitment_cycles/:recruitment_cycle_year/providers/:provider_code/courses(.:format)                                            api/v3/courses#index
 #                               api_v3_recruitment_cycle_provider_course GET    /api/v3/recruitment_cycles/:recruitment_cycle_year/providers/:provider_code/courses/:code(.:format)                                      api/v3/courses#show
 #                                       api_v3_recruitment_cycle_courses GET    /api/v3/recruitment_cycles/:recruitment_cycle_year/providers/courses(.:format)                                                           api/v3/courses#index
@@ -172,6 +173,7 @@ Rails.application.routes.draw do
     end
 
     namespace :v3 do
+      resources :courses, controller: "course_searches", only: :index
       resources :recruitment_cycles,
                 only: :show,
                 param: :year do
--- a/db/migrate/20200131104849_add_site_status_indexes.rb
+++ b/db/migrate/20200131104849_add_site_status_indexes.rb
@@ -1,0 +1,9 @@
+class AddSiteStatusIndexes < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :course_site, :publish, algorithm: :concurrently
+    add_index :course_site, :status, algorithm: :concurrently
+    add_index :course_site, :vac_status, algorithm: :concurrently
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_01_15_141035) do
+ActiveRecord::Schema.define(version: 2020_01_31_104849) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_buffercache"
@@ -119,7 +119,10 @@ ActiveRecord::Schema.define(version: 2020_01_15_141035) do
     t.text "status"
     t.text "vac_status"
     t.index ["course_id"], name: "IX_course_site_course_id"
+    t.index ["publish"], name: "index_course_site_on_publish"
     t.index ["site_id"], name: "IX_course_site_site_id"
+    t.index ["status"], name: "index_course_site_on_status"
+    t.index ["vac_status"], name: "index_course_site_on_vac_status"
   end
 
   create_table "course_subject", id: :serial, force: :cascade do |t|
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -384,6 +384,230 @@ describe Course, type: :model do
     end
   end
 
+  describe "scopes" do
+    describe ".with_recruitment_cycle" do
+      subject { described_class.with_recruitment_cycle(provider.recruitment_cycle.year) }
+      let(:test_course) { create(:course, provider: provider) }
+
+      before { test_course }
+
+      context "course is in recruitment cycle" do
+        let(:provider) { create(:provider) }
+
+        it "is returned" do
+          expect(subject).to contain_exactly(test_course)
+        end
+      end
+
+      context "course is not in recruitment cycle" do
+        let(:provider) { create(:provider, :next_recruitment_cycle) }
+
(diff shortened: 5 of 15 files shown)


gpeng:
[[S] Add filtering for courses to api part 1](https://trello.com/c/T2gCjkut/2827-s-add-filtering-for-courses-to-api-part-1)

app/controllers/api/v3/course_searches_controller.rb
```diff
@@ -0,0 +1,10 @@
+module API
+  module V3
+    class CourseSearchesController < API::V3::ApplicationController
+      def index
+        course_scope = CourseSearch.call(filter: params[:filter], recruitment_cycle_year: params[:recruitment_cycle_year])
```
defong: may be use as default
`Settings.current_recruitment_cycle_year`

config/routes.rb
```diff
@@ -91,9 +91,15 @@
...
+#                                        api_v3_recruitment_cycle_course GET    /api/v3/recruitment_cycles/:recruitment_cycle_year/courses/:id(.:format)                                                                 api/v3/course_searches#show
+#                                                                        PATCH  /api/v3/recruitment_cycles/:recruitment_cycle_year/courses/:id(.:format)                                                                 api/v3/course_searches#update
+#                                                                        PUT    /api/v3/recruitment_cycles/:recruitment_cycle_year/courses/:id(.:format)                                                                 api/v3/course_searches#update
+#                                                                        DELETE /api/v3/recruitment_cycles/:recruitment_cycle_year/courses/:id(.:format)                                                                 api/v3/course_searches#destroy
```
defong: apart from `GET|POST` I don't think there is need for the other verbs...

config/routes.rb
```diff
@@ -91,9 +91,15 @@
 #                                                                        PUT    /api/v2/access_requests/:id(.:format)                                                                                                    api/v2/access_requests#update
 #                                                                        DELETE /api/v2/access_requests/:id(.:format)                                                                                                    api/v2/access_requests#destroy
 #                                                api_v2_build_new_course GET    /api/v2/build_new_course(.:format)                                                                                                       api/v2/courses#build_new
+#                                       api_v3_recruitment_cycle_courses GET    /api/v3/recruitment_cycles/:recruitment_cycle_year/courses(.:format)                                                                     api/v3/course_searches#index
```
defong: I think there we should expose `/course` and use that as the default for frontend

`recruitment_cycle_year` is nice but I don't think we have an actual need to expose it.

db/migrate/20200131104849_add_site_status_indexes.rb
```diff
@@ -0,0 +1,9 @@
+class AddSiteStatusIndexes < ActiveRecord::Migration[6.0]
+  disable_ddl_transaction!
+
+  def change
+    add_index :course_site, :publish, algorithm: :concurrently
```
defong: :+1: 

should annotate add to these bit as well?
``` ruby
# Indexes
#
#  IX_course_site_course_id  (course_id)
#  IX_course_site_site_id    (site_id)
#  xyz


class SiteStatus < ApplicationRecord
```

db/schema.rb
```diff
@@ -119,7 +119,10 @@
     t.text "status"
     t.text "vac_status"
     t.index ["course_id"], name: "IX_course_site_course_id"
+    t.index ["publish"], name: "index_course_site_on_publish"
```
defong: before we forget 

just maybe

https://www.postgresql.org/docs/9.6/indexes-multicolumn.html

```
  scope :findable, -> { status_running.published_on_ucas }


```
@timabell fyi

app/services/course_search.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  class << self
+    def call(**args)
+      new(args).call
+    end
+  end
+
+  def call
+    scope = Course
```
defong: Just wondering does scope need to be ordered to be more performant?

ie

```ruby
faster = course
      .with_recruitment_cycle(recruitment_cycle_year)
      .findable

# vs 

slower = course
      .findable
      .with_recruitment_cycle(recruitment_cycle_year)

```

config/routes.rb
```diff
@@ -175,6 +181,7 @@
       resources :recruitment_cycles,
                 only: :show,
                 param: :year do
+        resources :courses, controller: "course_searches"
```
misaka: For providers, the search was incorporated as part of the [existing `CourseController`](https://github.com/DFE-Digital/teacher-training-api/pull/1144/files#diff-63584eeba1d49b4829082079274d0447R9), I like this because we should support searches, etc, on any `Course` index.

Can we just add a `courses` resource here that uses the existing `CourseController` and add search support to it?

app/services/course_search.rb
misaka: No, these should be functionally equivalent.

Don't forget these are `ActiveRecord::Relation` objects, which means that a) the accumulate the parameters and only execute them once the results are requested, and b) we can examine what the SQL looks like:

```sql
> Course.findable.with_recruitment_cycle('2020').to_sql
SELECT "course".* FROM "course"
       INNER JOIN "course_site" ON "course_site"."course_id" = "course"."id"
       INNER JOIN "provider" ON "provider"."id" = "course"."provider_id"
       INNER JOIN "recruitment_cycle" ON "recruitment_cycle"."id" = "provider"."recruitment_cycle_id"
       WHERE "course_site"."status" = 'R'
             AND "course_site"."publish" = 'Y'
             AND "recruitment_cycle"."year" = '2020'


> Course.with_recruitment_cycle('2020').findable.to_sql
SELECT "course".* FROM "course"
       INNER JOIN "provider" ON "provider"."id" = "course"."provider_id"
       INNER JOIN "recruitment_cycle" ON "recruitment_cycle"."id" = "provider"."recruitment_cycle_id"
       INNER JOIN "course_site" ON "course_site"."course_id" = "course"."id"
       WHERE "recruitment_cycle"."year" = '2020'
             AND "course_site"."status" = 'R'
             AND "course_site"."publish" = 'Y'

```

config/routes.rb
misaka: Ya, I think that that makes sense too. `find` doesn't need to know what the current cycle is.

spec/models/course_spec.rb
```diff
@@ -384,6 +384,230 @@
...
+        expect(subject).to contain_exactly(course_school_direct_salaried_training_programme)
+      end
+    end
+
+    describe ".with_qualifications" do
+      let(:course_qts) { create(:course, :resulting_in_qts) }
+      let(:course_pgce_with_qts) { create(:course, :resulting_in_pgce_with_qts) }
+      let(:course_pgde_with_qts) { create(:course, :resulting_in_pgde_with_qts) }
+      let(:course_pgce) { create(:course, :resulting_in_pgce) }
+      let(:course_pgde) { create(:course, :resulting_in_pgde) }
```
misaka: What do you think about adding these, and others, to `TestDataCache`? It'd be nice to start to populate that with commonly-used course types, instead of having to back-fill more and more of these.

db/schema.rb
gpeng: Discussed offline. The individual indices will be fine here with this table size but this is a potential optimisation in the future.

config/routes.rb
gpeng: I've updated this after offline discussion. This is now `/api/V3/courses` and only routes `index`.

gpeng:
We discussed adding search to the existing courses controller offline and have decided not to do this because:

* We don't currently have a user need for search on that endpoint
* it would require adding provider filtering to the `CourseSearch` class which is additional complexity we don't need yet
* The `CourseSearchesController` class will be easy to incorporate into `CoursesController` in the future if we need to without breaking the API. (It would also be difficult to remove search from `CoursesController` if we added it and realised we didn't have a user need for it after all.

spec/models/course_spec.rb
gpeng: I've added the `TestDataCache` as discussed offline. Turns out, there were only a small number that worked with it as most in this PR require parameters. I've left it in for now but I'm not sure the overhead is worth it in the unit tests as in this PR. I wonder if we'd be best benchmarking the specs to identify the slowest one and then trying `TestDataCache` on that and getting some data on the performance and seeing the development overhead to add it. We could also try some other approaches on the same file and compare before we spread it too far through the app.

app/services/course_search.rb
```diff
@@ -0,0 +1,50 @@
+class CourseSearch
```
misaka: Our other services use the suffix `Service`, can we bring this one in line? Can be in another PR.

PR #1142, opened 2020-01-24
2827 add filtering for courses to api part 1

defong:
### Context
Scope

### Changes proposed in this pull request
Added scopes for the course model

### Guidance to review

Nice cut off point

Will have a following PR for exposing it on the api endpoint

its going to shaped up something like this in the controller

```
@course.only_with_salary(params)
	.only_with_vacancies(params)
	.with_study_modes(params)
	.with_qualifications(params)
```

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #1155, opened 2020-02-04
2855 add filtering for courses to api part 2

gpeng:
### Context

The courses endpoint should only return published courses. In this context a 'published' course is one that has its most recent associated `CourseEnrichment` in `published` state.

* Add `Course.published` scope. This delegates the 'what is published' stuff to `CourseEnrichment`. I think that's the right place for it but it's tricky to get a chainable scope back from `CourseEnrichment.current_published`. I tried variations on SQL and Arel but it either didn't work or was super slow. This seems to work ok but the `group` calls need to be 'creatively' namespaced.
* Add the `published` scope into `CourseSearch`

### Guidance to review

This is on top of #1142 which is still in review and likely to change. The CI failures aren't directly related to this work. I'm looking at those separately.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_search.rb
+++ b/app/services/course_search.rb
@@ -13,6 +13,7 @@ class CourseSearch
   def call
     scope = Course
       .findable
+      .published
       .with_recruitment_cycle(recruitment_cycle_year)
 
     scope = scope.with_salary if funding_filter_salary?
--- a/app/services/course_search.rb
+++ b/app/services/course_search.rb
@@ -1,0 +1,50 @@
+class CourseSearch
+  def initialize(filter:, recruitment_cycle_year: Settings.current_recruitment_cycle_year)
+    @filter = filter || {}
+    @recruitment_cycle_year = recruitment_cycle_year
+  end
+
+  class << self
+    def call(**args)
+      new(args).call
+    end
+  end
+
+  def call
+    scope = Course
+      .findable
+      .with_recruitment_cycle(recruitment_cycle_year)
+
+    scope = scope.with_salary if funding_filter_salary?
+    scope = scope.with_qualifications(qualifications) if qualifications.any?
+    scope = scope.with_vacancies if has_vacancies?
+    scope = scope.with_study_modes(study_types) if study_types.any?
+    scope
+  end
+
+  private_class_method :new
+
+private
+
+  attr_reader :filter, :recruitment_cycle_year
+
+  def funding_filter_salary?
+    filter[:funding] == "salary"
+  end
+
+  def qualifications
+    return [] if filter[:qualification].blank?
+
+    filter[:qualification].split(",")
+  end
+
+  def has_vacancies?
+    filter[:has_vacancies].to_s.downcase == "true"
+  end
+
+  def study_types
+    return [] if filter[:study_type].blank?
+
+    filter[:study_type].split(",")
+  end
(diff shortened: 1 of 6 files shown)


gpeng:
[[S] Add filtering for courses to api part 2](https://trello.com/c/U8vXDXS7/2855-s-add-filtering-for-courses-to-api-part-2)

app/services/course_search.rb
```diff
@@ -13,6 +13,7 @@ def call(**args)
   def call
     scope = Course
       .findable
+      .published
```
defong:
```ruby
[13] pry(main)> cfpwrc = Course.findable .published.with_recruitment_cycle("2020").size
=> {[42513, 12928759]=>1}
[14] pry(main)> cfpwrc = Course.findable .published.size
=> {[42513, 12928759]=>1}
[15] pry(main)> cfpwrc = Course.findable.size
=> 44968
[16] pry(main)> cfpwrc = Course.findable.with_recruitment_cycle("2020").size
=> 23810
[17] pry(main)> cfpwrc = Course.published.size
=> {[42513, 12928759]=>1}
[18] pry(main)> cfpwrc = Course.all.size
=> 37202
[19] pry(main)> cfpwrc = Course.findable.distinct.size
=> 25003
```

:thinking:

gpeng: Do some of those look wrong?

misaka:
```
> Course.findable .published.with_recruitment_cycle("2020").length
=> 1
> Course.findable.published.with_recruitment_cycle("2020").length
=> 1
```

Why only one course returned by those scopes added together?

defong: root cause
`published`

misaka: This is why:

```
> Course.published.length
=> 1
```

Although I'm still not clear on why yet.

```
SELECT "course".* FROM "course"
       INNER JOIN "course_enrichment" ON "course_enrichment"."course_id" = "course"."id"
       WHERE (course_enrichment.created_at = (
              SELECT MAX(course_enrichment.created_at)
                     FROM course_enrichment))
              AND "course_enrichment"."status" = 1
       GROUP BY course_enrichment.course_id, course_enrichment.id
```

misaka: Actually, this looks suspicious:

```
course_enrichment.created_at = (
  SELECT MAX(course_enrichment.created_at) FROM course_enrichment)
)
```

gpeng: Tricky one. I think it's something up with the grouping. I've been struggling to get it to work and be able to chain the scopes. The grouping on `course_enrichment` should be in the subquery but it's getting added at the end when the scopes are chained. I'll try some other things.

gpeng:
I've updated this now. We didn't need to get the 'latest' `CourseEnrichment` so the complexity in there has been removed. the `Course.published` scope now returns courses that have a `published` `CourseEnrichment`. Although we don't explicitly exclude `withdrawn` courses in the `published` scope, the `findable` scope excludes these as for a withdrawn course enrichment is also marked `suspended` and `findable` requires the enrichment to be in `running` state. This is obviously not ideal but until we rationalise the `Course` state this should provider workable filtering for Find.

PR #1162, opened 2020-02-06
2927 - Add provider suggestions endpoint to V3

TobyRet:
### Context

Before users can have provider suggestions displayed on new Find we need to make this data available via the v3 API. Up until this point provider suggestions were only available via v2.

Adds `/provider-suggestions` end point to V3

### Guidance to review
@misaka and I discussed the following points:
- _Could provider suggestion logic could be moved to the index action on the ProvidersController?_ After investigating this further this did not seem to be a viable option as we would end up creating complications by incorporating parameters into the /providers endpoint particularly around error handling. On that basis the provider suggestion logic was moved to an index action on a new controller - ProviderSuggestionsController. If people are not happy with this solution then I can drop the last commit and have a 'suggestions' action on the ProvidersController (which mirrors what happens in V2).

- V2 and V3 will have similar provider suggestion logic. _Can this be extracted to a shared service?_ This has not happened as yet. Unlike V2, V3 does not have any form of authorisation. Once this logic was removed in V3 there seemed to be little left that was worth sharing. Open to suggestions though ...

### Testing locally
```
teacher-training-api 2927-provider-suggestions-endpoint-v3 % curl localhost:3001/api/v3/provider-suggestions?query=acme
{"data":[{"id":"3","type":"provider","attributes":{"provider_name":"ACME SCITT 0","provider_code":"A0"}},{"id":"1","type":"provider","attributes":{"provider_name":"Acme SCITT","provider_code":"A01"}},{"id":"4","type":"provider","attributes":{"provider_name":"ACME SCITT 1","provider_code":"A1"}},{"id":"5","type":"provider","attributes":{"provider_name":"ACME SCITT 2","provider_code":"A2"}},{"id":"6","type":"provider","attributes":{"provider_name":"ACME SCITT 3","provider_code":"A3"}}],"jsonapi":{"version":"1.0"}}%
```

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/provider_suggestions_controller.rb
+++ b/app/controllers/api/v3/provider_suggestions_controller.rb
@@ -1,0 +1,27 @@
+module API
+  module V3
+    class ProviderSuggestionsController < API::V3::ApplicationController
+      before_action :build_recruitment_cycle
+
+      def index
+        return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
+        return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
+
+        found_providers = @recruitment_cycle.providers
+                              .search_by_code_or_name(params[:query])
+                              .limit(5)
+
+        render(
+          jsonapi: found_providers,
+          class: { Provider: SerializableProviderSuggestion },
+        )
+      end
+
+    private
+
+      def begins_with_alphanumeric(string)
+        string.match?(/^[[:alnum:]].*$/)
+      end
+    end
+  end
+end
--- a/app/controllers/api/v3/provider_suggestions_controller.rb
+++ b/app/controllers/api/v3/provider_suggestions_controller.rb
@@ -9,7 +9,7 @@ module API
 
         found_providers = @recruitment_cycle.providers
                               .search_by_code_or_name(params[:query])
-                              .limit(5)
+                              .limit(10)
 
         render(
           jsonapi: found_providers,
--- a/app/controllers/api/v3/provider_suggestions_controller.rb
+++ b/app/controllers/api/v3/provider_suggestions_controller.rb
@@ -13,7 +13,7 @@ module API
 
         render(
           jsonapi: found_providers,
-          class: { Provider: SerializableProviderSuggestion },
+          class: { Provider: SerializableProvider },
         )
       end
 
--- a/app/serializers/api/v3/serializable_provider_suggestion.rb
+++ b/app/serializers/api/v3/serializable_provider_suggestion.rb
@@ -1,9 +1,0 @@
-module API
-  module V3
-    class SerializableProviderSuggestion < JSONAPI::Serializable::Resource
-      type "provider"
-
-      attributes :provider_name, :provider_code
-    end
-  end
-end
--- a/app/serializers/api/v3/serializable_provider_suggestion.rb
+++ b/app/serializers/api/v3/serializable_provider_suggestion.rb
@@ -1,0 +1,9 @@
+module API
+  module V3
+    class SerializableProviderSuggestion < JSONAPI::Serializable::Resource
+      type "provider"
+
+      attributes :provider_name, :provider_code
+    end
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -190,7 +190,7 @@ Rails.application.routes.draw do
           end
         end
       end
-      get "providers/suggest", to: "providers#suggest"
+      get "provider-suggestions", to: "provider_suggestions#index"
 
       resources :subject_areas, only: :index
     end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -98,6 +98,7 @@
 #                                     api_v3_recruitment_cycle_providers GET    /api/v3/recruitment_cycles/:recruitment_cycle_year/providers(.:format)                                                                   api/v3/providers#index
 #                                      api_v3_recruitment_cycle_provider GET    /api/v3/recruitment_cycles/:recruitment_cycle_year/providers/:code(.:format)                                                             api/v3/providers#show
 #                                               api_v3_recruitment_cycle GET    /api/v3/recruitment_cycles/:year(.:format)                                                                                               api/v3/recruitment_cycles#show
+#                                            api_v3_provider_suggestions GET    /api/v3/provider-suggestions(.:format)                                                                                                   api/v3/provider_suggestions#index
 #                                                   api_v3_subject_areas GET    /api/v3/subject_areas(.:format)                                                                                                          api/v3/subject_areas#index
 #                                                        api_system_sync POST   /api/system/sync(.:format)                                                                                                               api/system/force_sync#sync
 #                                                              error_500 GET    /error_500(.:format)                                                                                                                     error#error_500
@@ -189,6 +190,7 @@ Rails.application.routes.draw do
           end
         end
       end
+      get "providers/suggest", to: "providers#suggest"
 
       resources :subject_areas, only: :index
     end
--- a/spec/requests/api/v2/providers/suggest_provider_spec.rb
+++ b/spec/requests/api/v2/providers/suggest_provider_spec.rb
@@ -10,61 +10,68 @@ describe "GET /suggest" do
   let(:credentials) do
     ActionController::HttpAuthentication::Token.encode_credentials(token)
   end
+  let(:provider) { create(:provider, provider_name: "PROVIDER 1", organisations: [organisation]) }
+  let(:provider2)  { create(:provider, provider_name: "PROVIDER 2", organisations: [organisation]) }
 
-  it "searches for a particular provider for the user in the current recruitment cycle" do
-    provider = create(:provider, provider_name: "PROVIDER 1", organisations: [organisation])
-    create(:provider, provider_name: "PROVIDER 2", organisations: [organisation])
+  context "current recruitment cycle" do
+    before do
+      provider
+      provider2
+    end
 
-    get "/api/v2/providers/suggest?query=#{provider.provider_name}",
-        headers: { "HTTP_AUTHORIZATION" => credentials }
+    it "searches for a particular provider" do
+      get "/api/v2/providers/suggest?query=#{provider.provider_name}",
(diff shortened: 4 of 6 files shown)


TobyRet:
[Add provider suggestions endpoint to V3](https://trello.com/c/k20xykl9/2927-add-provider-suggestions-endpoint-to-v3)

config/routes.rb
```diff
@@ -187,6 +188,7 @@
           end
         end
       end
+      get "provider_suggestions", to: "provider_suggestions#index"
```
dankmitchell: Should the URL be hyphenated?

app/controllers/api/v3/provider_suggestions_controller.rb
```diff
@@ -0,0 +1,27 @@
+module API
+  module V3
+    class ProviderSuggestionsController < API::V3::ApplicationController
+      before_action :build_recruitment_cycle
+
+      def index
```
TobyRet: Potentially scope to have a shared action made available via a concern so that V2 and V3 could access? V2 is slightly different however in that it does authentication ...

app/controllers/api/v3/provider_suggestions_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+        return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
+        return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
+
+        found_providers = @recruitment_cycle.providers
+                              .search_by_code_or_name(params[:query])
+                              .limit(5)
```
dankmitchell: We have authority to increase this number to`10`

config/routes.rb
TobyRet: Ah yes - will make the change

app/controllers/api/v3/provider_suggestions_controller.rb
TobyRet: 👍

app/controllers/api/v3/provider_suggestions_controller.rb
gpeng: I don't think this bit of duplication is too bad. We're changing the number of results returned and the v2 is the suggest action.

spec/requests/api/v3/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  let(:jsonapi_renderer) { JSONAPI::Serializable::Renderer.new }
+  let(:organisation) { create(:organisation) }
+  let(:provider1) { create(:provider, provider_name: "PROVIDER 1", organisations: [organisation]) }
+  let(:provider2) { create(:provider, provider_name: "PROVIDER 2", organisations: [organisation]) }
+  let(:next_recruitment_cycle) { find_or_create(:recruitment_cycle, :next) }
+  let(:provider3) { create(:provider, provider_name: provider2.provider_name, organisations: [organisation], recruitment_cycle: next_recruitment_cycle) }
+  let(:provider4) { create(:provider, organisations: [organisation], recruitment_cycle: next_recruitment_cycle) }
+
+  before do
+    provider1
```
gpeng: This is a small thing in this file but adds up over the test suite. We should try and limit the amount of resources we create in the tests.

spec/requests/api/v3/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    provider2
+    provider3
+    provider4
+  end
+
+  it "searches for a particular provider for the user in the current recruitment cycle" do
+    get "/api/v3/provider-suggestions?query=#{provider2.provider_name}"
```
gpeng: This could be in a context and then only needs `provider2`. You could argue that it needs another one that isn't returned to make it a better test I suppose.

spec/requests/api/v3/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        "attributes" => {
+          "provider_code" => provider2.provider_code,
+          "provider_name" => provider2.provider_name,
+        },
+      },
+    ])
+  end
+
+  it "searches for a provider that is not in the current recruitment cycle" do
+    get "/api/v3/provider-suggestions?query=#{provider4.provider_name}"
```
gpeng: This one only needs provider 4. I won't comment on the others. I'm sure you get the picture.

spec/requests/api/v3/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+        "attributes" => {
+          "provider_code" => provider2.provider_code,
+          "provider_name" => provider2.provider_name,
+        },
+      },
+    ])
+  end
+
+  it "limits responses to a maximum of 5 items" do
+    20.times do
```
gpeng: Sorry... lied about the comments. This could just create 6 rather than 20.

spec/requests/api/v3/providers/suggest_provider_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      create(:provider, provider_name: "provider X", organisations: [organisation])
+    end
+
+    get "/api/v3/provider-suggestions?query=provider"
+
+    expect(JSON.parse(response.body)["data"].length).to eq(10)
```
gpeng: Why 10? We should check that there are 5 providers shouldn't we?

spec/requests/api/v3/providers/suggest_provider_spec.rb
gpeng: Oh... just saw your new commit. This only needs 11 and the title still says 5 :)

spec/requests/api/v3/providers/suggest_provider_spec.rb
TobyRet: I did not update the test description sorry. Dan has asked for 10 suggestions to be returned

app/serializers/api/v3/serializable_provider_suggestion.rb
```diff
@@ -0,0 +1,9 @@
+module API
+  module V3
+    class SerializableProviderSuggestion < JSONAPI::Serializable::Resource
+      type "provider"
+
+      attributes :provider_name, :provider_code
```
misaka: Do we need One More Serializer (:tm:) here? We should be able to just use `SerializableProvider` with sparse fields?

spec/requests/api/v3/providers/suggest_provider_spec.rb
TobyRet: @gpeng - yep good call. I've refactored the spec

spec/requests/api/v3/providers/suggest_provider_spec.rb
misaka: I would agree that we need another provider in there to demonstrate that it doesn't get returned. It's cases like this that was driving my proposal for standard test data setup.

app/serializers/api/v3/serializable_provider_suggestion.rb
TobyRet: Yeah sure - I think as long as there are not plans to add many more attributes?

spec/requests/api/v3/providers/suggest_provider_spec.rb
TobyRet: 👍

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -3,49 +3,46 @@
...
   let(:credentials) do
     ActionController::HttpAuthentication::Token.encode_credentials(token)
   end
 
-  before do
-    provider1
-    provider2
-    provider3
-    provider4
-  end
```
misaka: Why are we removing these? It looks like we're removing the tests that certain providers are _not_ in the results, is that intentional?

app/controllers/api/v3/provider_suggestions_controller.rb
misaka: I would still be tempted to use a service ... I'm not an advocate for services-for-absolutely-everything, but search does seem like a good spot to use a service, even if the logic is minimal. We can address this later though.

app/serializers/api/v3/serializable_provider_suggestion.rb
misaka: If we need to make a new serialiser, we always can ... until then I think we should try to stick to one serializer per resource / API version combination.

spec/requests/api/v2/providers/suggest_provider_spec.rb
TobyRet: Yep - I think it's about getting the balance right. 
We don't need to create these providers for every spec, so now we are are just creating when required

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -3,49 +3,47 @@
...
-
   it "searches for a particular provider for the user in the current recruitment cycle" do
```
gpeng: I think these should be contexts e.g.

```ruby
context "when the provider is in the current recruitment cycle" do
  let(:provider_in_cycle) { ... }
  let(:provider_out_of_cycle) { ... }

  it "is returned" do
    # the test
  end
end
```

spec/requests/api/v2/providers/suggest_provider_spec.rb
```diff
@@ -3,74 +3,78 @@
...
-    provider4
-  end
-
-  it "searches for a particular provider for the user in the current recruitment cycle" do
-    get "/api/v2/providers/suggest?query=#{provider2.provider_name}",
-        headers: { "HTTP_AUTHORIZATION" => credentials }
+  context "current recruitment cycle" do
+    it "searches for a particular provider" do
+      provider = create(:provider, provider_name: "PROVIDER 1", organisations: [organisation])
+      create(:provider, provider_name: "PROVIDER 2", organisations: [organisation])
```
misaka: I totally missed, after the first few glances, that we are in fact creating two providers here. What if we gave this a name, like `_provider_missing_from_results`?

PR #1166, opened 2020-02-11
2946 rationalise the courses endpoint

gpeng:
### Context

We need to filter courses on Find and we're using the JSON API Client gem. The existing URL structure and distinction between `CoursesController` and `CourseSearchesController` was difficult to work with. Also, the requirement for `provider_code` and `recruitment_cycle_year` in Find complicated everything.

We decided to respond to:

/courses
/recruitment_cycles/:year/courses
/recruitment_cycles/:year/providers/:code/courses

and to use `CoursesController#index` for this.

* Respond on urls
  - /api/v3/courses
  - /api/recuitment_cycles/:year/courses
  - /api/recuitment_cycles/:year/providers/:code/courses
* Add a scope parameter to `CourseSearchService`
* Enable filtering on all endpoints using the `CourseSearchService`
* Remove /api/v3/recruitment_cycles/:year/providers/courses
* Remove `CourseSearch` (renamed and committed by mistake previously)
* Remove `CourseSearchesController` functionality has been added to
`CoursesController`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/services/course_search_service.rb
```diff
@@ -11,9 +11,7 @@ def call(**args)
   end
 
   def call
-    scope = Course
-      .findable
-      .with_recruitment_cycle(recruitment_cycle_year)
+    scope = course_scope.findable
```
misaka: I feel like `findable` should actually be passed in by the caller, since different controllers could potentially not care about `findable`, and since we have the mechanism available let's use it?

misaka:
I like how this is shaping up. :100:

dankmitchell:
[Rationalise the /courses endpoint](https://trello.com/c/zAsqZl5V/2946-rationalise-the-courses-endpoint)

app/services/course_search_service.rb
gpeng: I wondered about that. Could we move it when we need to add another caller that doesn't want just findable courses rather than now when the only use we have is findable only?

defong: fair call on YAGNI

gpeng:
> minor question
> 
> > /api/recuitment_cycles/:year/providers/:code/courses

I thought there was in the [original specs](https://github.com/DFE-Digital/teacher-training-api/blob/c8c94ffdb19b48b2f2a747742717426de0140f9f/spec/requests/api/v3/providers/courses/index_spec.rb) but it seems there isn't a test that it excludes other providers. Bear with and I'll add one.

app/services/course_search_service.rb
misaka: So long as we don't forget about it at a later time and introduce a subtle bug. :)

defong: true, but it will be along the line with introduce new feature and introduce a bug that did not expect the course search to only return `findable`

misaka:
:shipit:

PR #1167, opened 2020-02-11
[2938] 1d add order to coursesubject

misaka:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/models/course.rb
```diff
@@ -97,7 +97,7 @@ class Course < ApplicationRecord
              optional: true
 
   has_many :course_subjects
-  has_many :subjects, through: :course_subjects
+  has_many :subjects, -> { order "course_subject.priority" }, through: :course_subjects
```
misaka: I'd actually move the ordering into `course_subjects` and add `before_add` hook to automatically order any assigned `CourseSubject` objects accordingly:

```ruby
  has_many :course_subjects,
           -> { order :priority },
           inverse_of: "course",
           before_add: :set_course_subject_position

  def set_course_subject_position(course_subject)
    return if course_subject.priority.present?

    course_subject.priority = if course_subjects.any?
                                course_subjects.last.priority + 1
                              else
                                0
                              end
  end
```

misaka: This means that we would can add subjects to a course and get ordering by default:

```ruby
> c = Course.new
=> #<Course:0x00007fcb9c85b398 ...>
> english = Subject.find_by(subject_name: "English")
=> #<SecondarySubject:0x00007fcb9a5bab18 id: 21, ...>
> maths = Subject.find_by(subject_name: "Mathematics")
=> #<SecondarySubject:0x00007fcb9d4a4208 id: 25, ...>
> c.subjects = [maths, english]
=> [#<SecondarySubject:0x00007fcb9a6107c0 id: 25, ...>,
 #<SecondarySubject:0x00007fcb9a4f9cd8 id: 21, ...>]
> c.course_subjects
=> [#<CourseSubject:0x00007fcb9d1d6d00 id: nil, course_id: nil, subject_id: 25, created_at: nil, updated_at: nil, priority: 0>,
 #<CourseSubject:0x00007fcb9d234d88 id: nil, course_id: nil, subject_id: 21, created_at: nil, updated_at: nil, priority: 1>]
```

This even appears to work when a course already has a subject, the priorities end up starting at 0:

```ruby
[8] pry(main)> c = rc.courses.sample
=> #<Course:0x00007fc5dea20bd8
 id: 12924670,
 ...>
[9] pry(main)> c.subjects
=> [#<SecondarySubject:0x00007fc5c2148388 id: 28, type: "SecondarySubject", subject_code: "C6", subject_name: "Physical education", created_at: nil, updated_at: nil, subject_area_id: nil>]
[10] pry(main)> c.course_subjects
=> [#<CourseSubject:0x00007fc5c2218e70 id: 25900, course_id: 12924670, subject_id: 28, created_at: nil, updated_at: nil, priority: nil>]
[11] pry(main)> c.course_subjects.first.priority = 1
=> 1
[12] pry(main)> c.course_subjects.first.save
=> true
[13] pry(main)> c.course_subjects
=> [#<CourseSubject:0x00007fc5c2218e70 id: 25900, course_id: 12924670, subject_id: 28, created_at: nil, updated_at: Tue, 11 Feb 2020 19:11:08 UTC +00:00, priority: 1>]
[14] pry(main)> c.subjects = [maths, english]
=> [#<SecondarySubject:0x00007fc5de700e30 id: 25, type: "SecondarySubject", subject_code: "G1", subject_name: "Mathematics", created_at: nil, updated_at: nil, subject_area_id: nil>,
 #<SecondarySubject:0x00007fc5da160ba0 id: 21, type: "SecondarySubject", subject_code: "Q3", subject_name: "English", created_at: nil, updated_at: nil, subject_area_id: nil>]
[15] pry(main)> c.course_subjects
=> [#<CourseSubject:0x00007fc5de90b0e0 id: 40785, course_id: 12924670, subject_id: 25, created_at: Tue, 11 Feb 2020 19:11:48 UTC +00:00, updated_at: Tue, 11 Feb 2020 19:11:48 UTC +00:00, priority: 0>,
 #<CourseSubject:0x00007fc5de372048 id: 40786, course_id: 12924670, subject_id: 21, created_at: Tue, 11 Feb 2020 19:11:48 UTC +00:00, updated_at: Tue, 11 Feb 2020 19:11:48 UTC +00:00, priority: 1>]
```

app/controllers/api/v2/courses_controller.rb
```diff
@@ -168,6 +169,11 @@ def update_subjects
         return if subject_ids.nil?
 
         @course.subjects = Subject.where(id: subject_ids)
+
+        subject_ids.map(&:to_i).each_with_index do |id, index|
+          @course.course_subjects.select { |cs| cs.subject_id == id }.first.priority = index
+        end
+
```
misaka: I would make this a bit more concise, and just add the course subjects directly, since we have to expose that part already.

```ruby
        @course.course_subjects = subject_ids.map.with_index do |subject_id, idx|
          CourseSubject.create(priority: idx, subject_id: subject_id)
        end
```

app/serializers/api/v2/serializable_course.rb
```diff
@@ -47,7 +47,11 @@ def enrichment_attribute(name, enrichment_name = name)
 
       has_many :site_statuses
       has_many :sites
-      has_many :subjects
+      has_many :subjects do
+        data do
+          @object.course_subjects.sort_by(&:priority).map(&:subject)
+        end
+      end
```
misaka: I would've expected the `subjects` association to already be ordered correctly, but even better, if we just order `course_subjects` as I suggest above we're definitely covered and don't need to do this.

I tested this locally, btw, and it appears to work.

app/services/courses/generate_course_name_service.rb
```diff
@@ -1,7 +1,7 @@
 module Courses
   class GenerateCourseNameService
     def execute(course:)
-      subjects = course.subjects
+      subjects = course.course_subjects.sort_by(&:priority).map(&:subject)
```
misaka: Also not needed if we just order the `course_subjects` association.

app/models/course.rb
misaka: btw, one reason I think it's more appropriate to order `course_association` is because the `priority` column exists there.

db/schema.rb
```diff
@@ -130,6 +130,8 @@
     t.integer "subject_id"
     t.datetime "created_at", precision: 6
     t.datetime "updated_at", precision: 6
+    t.boolean "main", default: false
+    t.integer "priority"
```
misaka: I use the name you have here, `priority`, in my comments, but looking around I found [`acts_as_list`](https://github.com/brendon/acts_as_list) which proposes the name `position`. Unfortunately I don't think that that gem will quite work (I don't think it'll work with our model where `position` lies on the join table, not on the item that's being joined), but I think that that name is better than `priority` or `order`.

misaka:
All in all, I think we can make this work, and if we move the ordering into the `course_subjects` association, and use a `before_add` hook, we'll be able to avoid any inconsistencies (as you mentioned in Slack).

You made a point about the importance of ordering not being apparent in the API interface. I'm not sure how important that is, we could describe this as being an "ordered_list" in the documentation, maybe that would help? We could, in theory, expose the `priority`/`position` parameter, that probably wouldn't be a terrible idea.

app/controllers/api/v2/courses_controller.rb
danielburnley: We don't need this for with the `:before_add` it seems :eyes:

danielburnley: Though they do have to be added in this way to keep the ordering (Subject.where is ordered by ID)

misaka: I thought it might be useful here just to re-enforce the ordering, and I think it's possible the hook could trigger a SQL request. But those issues aside, ya, we may not need/want it here.

danielburnley:
Closed - was implemented in #1176

PR #1172, opened 2020-02-13
[2958] Add ability to sort by provider name on v3 course endpoint

faissalMT:
# Context 
Users on Search and Compare can currently sort alphabetically by provider name, in order to provide this functionality for Find we must create an endpoint that supports this behaviour in accordance with JSON API's standards

Add an endpoint that supports sorting

### Guidance to review
https://jsonapi.org/format/#fetching-sorting

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -1,7 +1,8 @@
 class CourseSearchService
-  def initialize(filter:, course_scope: Course)
+  def initialize(filter:, sort: nil, course_scope: Course)
     @filter = filter || {}
     @course_scope = course_scope
+    @sort = sort
   end
 
   class << self
@@ -13,6 +14,9 @@ class CourseSearchService
   def call
     scope = course_scope.findable
 
+    scope = scope.by_provider_name_ascending if sort_by_provider_ascending?
+    scope = scope.by_provider_name_descending if sort_by_provider_descending?
+
     scope = scope.with_salary if funding_filter_salary?
     scope = scope.with_qualifications(qualifications) if qualifications.any?
     scope = scope.with_vacancies if has_vacancies?
@@ -23,6 +27,14 @@ class CourseSearchService
   private_class_method :new
 
 private
+
+  def sort_by_provider_ascending?
+    @sort == "provider.provider_name"
+  end
+
+  def sort_by_provider_descending?
+    @sort == "-provider.provider_name"
+  end
 
   attr_reader :filter, :course_scope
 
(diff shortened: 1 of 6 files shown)


app/services/course_search_service.rb
```diff
@@ -24,6 +28,14 @@ def call
 
 private
 
+  def should_sort_by_provider_ascending?
```
gpeng: I don't think we need the `should_` prefix on these two methods. Also, could they go under the `attr_reader`? Might just be me but I think we usually have the `attr_reader` at the top of the private bit. You could also add a reader for `:sort`.

spec/models/course_spec.rb
```diff
@@ -74,6 +74,62 @@
     it { should have_many(:financial_incentives) }
   end
 
+  fdescribe "#by_provider_name_ascending" do
```
gpeng: there's an `f` left in 'ere.

spec/requests/api/v3/courses/index_spec.rb
```diff
@@ -35,6 +35,60 @@
   let(:findable_status) { build(:site_status, :findable) }
   let(:published_enrichment) { build(:course_enrichment, :published) }
 
+  context "ordering" do
```
gpeng: This should probably be `describe`

spec/requests/api/v3/courses/index_spec.rb
```diff
@@ -35,6 +35,60 @@
...
+      let(:request_path) { "/api/v3/courses?include=provider&sort=provider.provider_name" }
+
+      it "returns an ordered list" do
+        get request_path
+
+        json_response = JSON.parse(response.body)
+        course_hashes = json_response["data"]
+        expect(course_hashes.first["id"]).to eq(course_a.id.to_s)
+        expect(course_hashes.second["id"]).to eq(course_b.id.to_s)
+        #Course.includes(:provider).order("provider.provider_name")
```
gpeng: Stray comment here.

spec/models/course_spec.rb
```diff
@@ -74,6 +74,62 @@
...
+             enrichments: [build(:course_enrichment, :published)])
+    end
+
+    let(:provider_b) { create(:provider, provider_name: "Provider B") }
+    let(:course_b) do
+      create(
+        :course,
+        name: "Course A",
+        provider: provider_b,
+        site_statuses: [build(:site_status, :findable)],
```
gpeng: We don't need the `site_statuses` and `enrichments` for these two tests.

PR #1174, opened 2020-02-15
[2949] Fix api v3 latest published enrichment

dankmitchell:
### Context
The latest published enrichment isn't always correct because of the order 

Use `max_by` instead of `last` as per - https://github.com/DFE-Digital/teacher-training-api/blob/master/app/serializers/search_and_compare/course_serializer.rb#L152:L156

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/serializers/api/v3/serializable_course.rb
+++ b/app/serializers/api/v3/serializable_course.rb
@@ -4,7 +4,10 @@ module API
       class << self
         def enrichment_attribute(name, enrichment_name = name)
           attribute name do
-            @object.enrichments.select(&:published?).last&.__send__(enrichment_name)
+            @object.enrichments
+              .select(&:published?)
+              .max_by { |e| [e.created_at, e.id] }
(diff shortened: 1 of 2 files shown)


dankmitchell:
[BUG: API v3 is not getting the latest published enrichment](https://trello.com/c/BwpmXlZo/2949-bug-api-v3-is-not-getting-the-latest-published-enrichment)

timabell:
Shouldn't this have a diff on a test too?

dankmitchell:
@timabell this PR just locks down the ordering of the "published" enrichments; the tests were already mocked in the correct order so wasn't sure on the value of adding more tests around this.

misaka:
@dankmitchell I made some changes that I think better reflect course enrichments' life-cycle. Can you double-check?

PR #1176, opened 2020-02-17
[2909] Change course to support ordering of subjects

danielburnley:
### Context

Secondary courses can have multiple subjects that have to be stored in the correct order to ensure that bursaries are calculated correctly.

To solve this, we are adding a `position` value on the `course_subject` in order to keep the original order they are given.

- Add an integer field `:position`  to `CourseSubject`
- Assign `position` when course subjects are added to a secondary course
- Assign `position` based on the order of the params given when updating/building a new course

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/db/migrate/20200212161248_add_position_to_course_subject.rb
+++ b/db/migrate/20200212161248_add_position_to_course_subject.rb
@@ -1,0 +1,5 @@
+class AddPositionToCourseSubject < ActiveRecord::Migration[6.0]
+  def change
+    add_column :course_subject, :position, :integer
+  end
+end
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -104,6 +104,7 @@ module API
 
         if @course.errors.empty? && @course.valid?
           @course.save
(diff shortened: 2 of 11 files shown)


dankmitchell:
[Change course to support ordering of subjects.](https://trello.com/c/ZZx3jCcC/2909-change-course-to-support-ordering-of-subjects)

db/migrate/20200212161248_add_position_to_course_subject.rb
```diff
@@ -0,0 +1,5 @@
+class AddPositionToCourseSubject < ActiveRecord::Migration[6.0]
+  def change
+    add_column :course_subject, :position, :integer
```
dankmitchell: Do we need a default of `0`

misaka: Don't think so. The subjects that are in there now don't currently have any position assigned to them, so I think it's ok for them to be `null`. Once we backfill values, maybe then it'll make sense to default to 0.

danielburnley:
Rebasing + Merging :tada:

PR #1180, opened 2020-02-18
[2986] Spike location searching

faissalMT:
### Context
This is a spike to illuminate avenues we might have to go down when properly implementing the sorting & search by locations

 - Remove the geocoder gem
 - Add the GeoKit gem
 - Replace some uses of the geocoder Gem

### Guidance to review
Doesn't look like there's a way to have a Redis cache for GeoKit? How big of a problem is this
The various GeoKit scopes for ordering by distance and only including things within a distance work fine (unless the distance is absurd, in which case it will silently fail, so you might want to note that if you're doing ad-hoc testing)
This will not be merged

### Checklist

- [x] Attach to Trello card

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,7 +87,7 @@ gem "discard"
 gem "govuk_notify_rails"
 
 # Geocoding for Ruby
-gem "geocoder"
+#gem "geocoder"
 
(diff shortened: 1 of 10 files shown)


danielburnley:
Completed in #1201

PR #1184, opened 2020-02-19
[2993] Add support for sorting by course name as well as provider name

faissalMT:
### Context
https://github.com/DFE-Digital/find-teacher-training/pull/106 highlighted that we need the ability to sort by both course name and provider name, as this is what C# does.  

Add canonical sorting.

### Guidance to review
https://github.com/DFE-Digital/search-and-compare-api/blob/3aa6be82dcf5548892580e900ea1e37decf63a0f/src/api/Controllers/CoursesController.cs#L226

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -157,8 +157,21 @@ class Course < ApplicationRecord
     end
   end
 
-  scope :by_provider_name_ascending, -> { includes(:provider).order("provider.provider_name asc") }
-  scope :by_provider_name_descending, -> { includes(:provider).order("provider.provider_name desc") }
+  scope :by_name_ascending, -> do
+    order(name: :asc)
+  end
+
+  scope :by_name_descending, -> do
+    order(name: :desc)
+  end
+
+  scope :ascending_canonical_order, -> do
+    joins(:provider).merge(Provider.by_name_ascending).order("name asc")
+  end
+
+  scope :descending_canonical_order, -> do
+    joins(:provider).merge(Provider.by_name_descending).order("name desc")
+  end
 
   scope :changed_since, ->(timestamp) do
     if timestamp.present?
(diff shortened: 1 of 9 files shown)


app/models/course.rb
```diff
@@ -157,8 +157,8 @@ def new_draft_attributes
     end
   end
 
-  scope :by_provider_name_ascending, -> { includes(:provider).order("provider.provider_name asc") }
-  scope :by_provider_name_descending, -> { includes(:provider).order("provider.provider_name desc") }
+  scope :ascending_canonical_order, -> { includes(:provider).order("provider.provider_name asc").order("name asc") }
```
dankmitchell:
```suggestion
  scope :ascending_canonical_order, -> { includes(:provider).order("provider.provider_name asc", "name asc") }
```

app/models/course.rb
```diff
@@ -157,8 +157,8 @@ def new_draft_attributes
...
+  scope :descending_canonical_order, -> { includes(:provider).order("provider.provider_name desc").order("name desc") }
```
dankmitchell:
```suggestion
  scope :descending_canonical_order, -> { includes(:provider).order("provider.provider_name desc, name desc") }
```

app/models/course.rb
```diff
@@ -157,8 +157,8 @@ def new_draft_attributes
     end
   end
 
-  scope :by_provider_name_ascending, -> { includes(:provider).order("provider.provider_name asc") }
-  scope :by_provider_name_descending, -> { includes(:provider).order("provider.provider_name desc") }
+  scope :ascending_canonical_order, -> { includes(:provider).order("provider.provider_name asc, name asc") }
```
gpeng: Ideally these scopes would reference a scope on `Provider` to avoid `Course` being coupled to the `Provider` DB schema. Also, they should use `joins` not `includes` as we don't need the `Provider` fields to be returned with the query.

PR #1186, opened 2020-02-20
[2980] - Add subject filtering to the V3 API

TobyRet:
### Context
Allow courses to be filtered by subject(s) (V3 API)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -186,7 +186,7 @@ class Course < ApplicationRecord
     where(study_mode: Array(study_modes) << "full_time_or_part_time")
   end
   scope :with_subjects, ->(subject_codes) do
-    includes(:subjects).where(subject: { subject_code: subject_codes })
+    joins(:subjects).merge(Subject.with_subject_codes(subject_codes))
   end
 
   scope :with_qualifications, ->(qualifications) do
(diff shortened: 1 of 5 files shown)


TobyRet:
[[M] Add subject filtering to the API](https://trello.com/c/m1ezJY78/2980-m-add-subject-filtering-to-the-api)

app/models/course.rb
```diff
@@ -185,6 +185,9 @@ def new_draft_attributes
   scope :with_study_modes, ->(study_modes) do
     where(study_mode: Array(study_modes) << "full_time_or_part_time")
   end
+  scope :with_subjects, ->(subject_codes) do
+    includes(:subjects).where(subject: { subject_code: subject_codes })
```
gpeng: Can we move the subject bit into a scope on `Subject`? That would keep the 'what is a subject code' concern in `Subject`. I think this should also be `joins` rather than `includes`.

TobyRet: Yup no prob

TobyRet: https://github.com/DFE-Digital/teacher-training-api/pull/1186/commits/717c08cd04b479687dc6e2a04a226afb94f7f345

PR #1188, opened 2020-02-20
2932 m add provider filtering to the api

SamJamCul:
### Context

[Check this trello card](https://trello.com/c/RBaGAf8C/2932-m-add-provider-filtering-to-the-api) for details - we're adding the ability to filter courses by provider name. 

Add scoping to providers to filter by exact provider name matches
Add scoping to courses to filter by provider names
Add method to CourseSearchService to filter by provider names

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -206,6 +206,10 @@ class Course < ApplicationRecord
     where(qualification: qualifications)
   end
 
+  scope :with_provider_name, ->(provider_name) do
+    joins(:provider).merge(Provider.where(provider_name: provider_name))
+  end
+
   def self.entry_requirement_options_without_nil_choice
     ENTRY_REQUIREMENT_OPTIONS.reject { |option| option == :not_set }.keys.map(&:to_s)
   end
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -22,6 +22,7 @@ class CourseSearchService
     scope = scope.with_vacancies if has_vacancies?
     scope = scope.with_study_modes(study_types) if study_types.any?
     scope = scope.with_subjects(subject_codes) if subject_codes.any?
+    scope = scope.with_provider_name(provider_name) if provider_name.present?
     scope
   end
 
@@ -63,5 +64,11 @@ private
     return [] if filter[:subjects].blank?
 
     filter[:subjects].split(",")
+  end
+
+  def provider_name
+    return [] if filter[:"provider.provider_name"].blank?
+
+    filter[:"provider.provider_name"]
   end
 end
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -853,6 +853,30 @@ describe Course, type: :model do
         end
       end
     end
+
+    describe ".with_provider_name" do
+      let(:provider) { create(:provider, provider_name: "ACME") }
+      let(:provider2) { create(:provider, provider_name: "DAVE") }
+      let(:course) { create(:course, provider: provider) }
+      let(:provider_name) { "ACME" }
+
+      before do
+        provider
+        provider2
+        course
+        provider_name
+      end
+
+      context "with an existing provider name" do
+        subject { described_class.with_provider_name(provider_name) }
+        it { is_expected.to contain_exactly(course) }
+      end
+
+      context "with a provider name with no courses" do
+        subject { described_class.with_provider_name("DAVE") }
+        it { is_expected.to be_empty }
+      end
+    end
   end
 
   describe "changed_at" do
--- a/spec/requests/api/v3/courses/index_spec.rb
+++ b/spec/requests/api/v3/courses/index_spec.rb
@@ -257,7 +257,7 @@ describe "GET v3/courses" do
     end
   end
 
-  context "subjects filter" do
+  describe "subjects filter" do
     let(:course_with_A1_subject) do
       create(:course,
              enrichments: [published_enrichment],
@@ -307,6 +307,44 @@ describe "GET v3/courses" do
 
       before do
         course_with_A1_subject
+      end
+
+      it "is not returned" do
+        get request_path
+        json_response = JSON.parse(response.body)
+        course_hashes = json_response["data"]
+        expect(course_hashes.count).to eq(0)
+      end
+    end
+  end
+
+  describe "provider filter" do
+    context "with a provider specified" do
+      let(:request_path) { "/api/v3/courses?filter[provider.provider_name]=#{filtered_provider_course.provider.provider_name}" }
+      let(:provider_filtered_by) { create(:provider) }
+      let(:filtered_provider_course) { create(:course, provider: provider_filtered_by, site_statuses: [findable_status], enrichments: [published_enrichment]) }
+
+      before do
+        provider_filtered_by
+        filtered_provider_course
+      end
+
+      it "is returned" do
+        get request_path
+        json_response = JSON.parse(response.body)
+        course_hashes = json_response["data"]
+        expect(course_hashes.count).to eq(1)
+      end
+    end
+
+    context "with a different provider specified" do
+      let(:request_path) { "/api/v3/courses?filter[provider.provider_name]=a+different+provider" }
+      let(:provider_excluded) { create(:provider) }
+      let(:excluded_provider_course) { create(:course, provider: provider_excluded, site_statuses: [build(:site_status, :findable)], enrichments: [published_enrichment]) }
+
+      before do
+        provider_excluded
+        excluded_provider_course
       end
 
       it "is not returned" do


SamJamCul:
[[M] Add provider filtering to the API](https://trello.com/c/RBaGAf8C/2932-m-add-provider-filtering-to-the-api)

app/controllers/api/v3/courses_controller.rb
```diff
@@ -7,7 +7,7 @@ class CoursesController < API::V3::ApplicationController
 
       def index
         course_search = CourseSearchService.call(filter: params[:filter], sort: params[:sort], course_scope: @courses)
-
+
```
danielburnley: :eyes:

spec/models/course_spec.rb
```diff
@@ -819,6 +819,28 @@
         end
       end
     end
+
+    describe ".with_provider_name" do
```
danielburnley:
```suggestion
    describe "#with_provider_name" do
```

spec/models/provider_spec.rb
```diff
@@ -169,6 +169,16 @@
     it { should_not include(provider2) }
   end
 
+  describe ".with_full_name" do
```
danielburnley:
```suggestion
  describe "#with_full_name" do
```

app/models/course.rb
```diff
@@ -190,6 +190,10 @@ def new_draft_attributes
     where(qualification: qualifications)
   end
 
+  scope :with_provider_name, ->(provider_name) do
+    where(provider_id: Provider.with_full_name(provider_name).select(:id))
```
```suggestion
    joins(:providers).merge(Provider.with_full_name(provider_name))
```

app/controllers/api/v3/courses_controller.rb
SamJamCul: Sorted

spec/models/course_spec.rb
```diff
@@ -819,6 +819,28 @@
...
+        course
+        provider_name
+      end
+
+      context "with ACME provider name" do
+        subject { described_class.with_provider_name(provider_name) }
+        it { is_expected.to contain_exactly(course) }
+      end
+
+      context "with Dave provider name" do
```
faissalMT:
```suggestion
      context "with the name of a provider that doesn't exist" do
```

spec/models/course_spec.rb
```diff
@@ -819,6 +819,28 @@
...
+      let(:course) { create(:course, provider: provider) }
+      let(:provider_name) { "ACME" }
+
+      before do
+        provider
+        course
+        provider_name
+      end
+
+      context "with ACME provider name" do
```
faissalMT:
```suggestion
      context "with the name of a provider that exists" do
```

spec/models/course_spec.rb
faissalMT: Should note that this convention also appears elsewhere in the file, outside of this commits changes

spec/requests/api/v3/courses/index_spec.rb
```diff
@@ -257,6 +257,45 @@
...
+        json_response = JSON.parse(response.body)
+        course_hashes = json_response["data"]
+        expect(course_hashes.count).to eq(1)
+      end
+    end
+
+    context "with a different provider specified" do
+      let(:request_path) { "/api/v3/courses?filter[provider.provider_name]=a+different+provider" }
+      let(:provider_excluded) { create(:provider) }
+      let(:excluded_provider_course) { create(:course, provider: provider_excluded, site_statuses: [build(:site_status, :findable)], enrichments: [published_enrichment]) }
```
faissalMT: Did you end up seeing if you could get rid of these attributes?

app/controllers/api/v3/courses_controller.rb
SamJamCul: Not quite sorted, should this be one or two lines in betwee?

spec/models/course_spec.rb
SamJamCul: Sorry, deleted my old comment by mistake -  The other tests for similar scope have a ".", should those change as well? This site I randomly found thinks as it's a class method, rather than an instance method, it should be a "." http://www.betterspecs.org/

spec/models/course_spec.rb
SamJamCul: This would be a separate test, as I'm not explicitly testing for a provider that doesn't exist - there isn't going to be a scenario where the api is looking for that, as users are only able to pick from a valid list of providers. I can rewrite the test to include a provider that doesn't have any courses though?

spec/requests/api/v3/courses/index_spec.rb
SamJamCul: I did manage to get rid of some, but the two left here are required for a course to show up in a search. The scoping will remove them otherwise

spec/models/course_spec.rb
SamJamCul: I've now modified the test so that it's checking an existing provider with no courses, and an existing provider with courses. This more accurately reflects the actual situations in which it'll be invoked!

app/models/course.rb
SamJamCul: Can I ask what this change is for?

spec/models/course_spec.rb
gpeng: I'd usually go with `.` for a class method and `#` for an instance method.

app/models/course.rb
gpeng:
```
[5] pry(main)> Course.where(provider_id: Provider.with_full_name("dave").select(:id)).to_sql
=> "SELECT \"course\".* FROM \"course\" WHERE \"course\".\"provider_id\" IN (SELECT \"provider\".\"id\" FROM \"provider\" WHERE \"provider\".\"provider_name\" = 'dave')"
[6] pry(main)> Course.joins(:provider).merge(Provider.with_full_name("dave")).to_sql
=> "SELECT \"course\".* FROM \"course\" INNER JOIN \"provider\" ON \"provider\".\"id\" = \"course\".\"provider_id\" WHERE \"provider\".\"provider_name\" = 'dave'"
```

I think @faissalMT 's suggestion comes out with the cleaner (and likely faster) SQL here. I can't remember why I didn't use `merge` on some of the earlier scopes. I think it was something to do with it not chaining properly when we join them together for the course search. I don't think that's an issue with `merge` though, it was probably to do with what was being merged.

faissalMT: @SamJamCul It doesn't make a huge difference here since you're (almost always) just picking out one item, but as Graham said doing things this way tends to be faster and more straightforward.

defong: definitely better sql there

app/services/course_search_service.rb
```diff
@@ -57,4 +58,10 @@ def study_types
 
     filter[:study_type].split(",")
   end
+
+  def provider_name
+    return [] if filter[:"provider.provider_name"].blank?
+
+    filter[:"provider.provider_name"].split(",")
```
defong: Are we expecting multiple provider names? 

``` ruby
[1] pry(main)> filter = {"provider.provider_name": "a,b,c"}
=> {:"provider.provider_name"=>"a,b,c"}
[2] pry(main)> filter[:"provider.provider_name"].split(",")    
=> ["a", "b", "c"]
```

SamJamCul: There's no expectation of it, but it was the easiest way of turning the data into an array! If you know a better way I'd be happy to replace it

spec/models/provider_spec.rb
misaka: As with `with_provider_name`, `with_full_name` is a class method so `.` is appropriate.

app/services/course_search_service.rb
misaka: There are a small number of providers that have commas in their names:

```
> RecruitmentCycle.current.providers.where("provider_name LIKE '%,%'").count
=> 41
```

Are they being escaped properly on the client side?

If all you want is conversion to an Array, then `Array()` works:

```
> Array("foo")
=> ["foo"]
> Array(["foo"])
=> ["foo"]
```

app/models/provider.rb
```diff
@@ -110,6 +110,7 @@ def current_accredited_courses
   scope :search_by_code_or_name, ->(search_term) {
     where("provider_name ILIKE ? OR provider_code ILIKE ?", "%#{search_term}%", "%#{search_term}%")
   }
+  scope :with_full_name, ->(provider_name) { where(provider_name: provider_name) }
```
misaka: For the sake of discussion, is there a benefit over just using `where(provider_name: provider_name)`? As I see it:

Pros:
- a few characters less
- abstracts away how we find the provider name

Cons:
- one more line of code in the model
- ~20 lines of specs

Are the Pros outweighing the Cons here?

app/models/course.rb
TobyRet: Some stackoverflow discussion - https://stackoverflow.com/questions/50490009/rails-activerecord-where-vs-merge

app/models/provider.rb
TobyRet: If `with_full_name` is part of a chain of scopes then perhaps stylistically, nice to have at same level of abstraction otherwise not sure if there are many advantages

app/services/course_search_service.rb
SamJamCul: Cool, I'll change this then

app/models/course.rb
SamJamCul: Changed!

PR #1190, opened 2020-02-21
[3018] Update subject db schema

gpeng:
###Context

There were some production issues on 20/2/2020 after we deployed some subject updates. The updates don't look to have caused the issues but while investigating I spotted a couple of minor things in the DB.

* Remove unused subject area columns
* Add index to subject type as it is the foreign key for subject area
* Add index to subject subject_code as we use it in the API query
* Add index to subject area typename as it is the primary key

### Guidance to review

`SubjectArea` uses `typename` as the primary key. The fields on subject that have been removed are all `nil` on production.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/db/migrate/20200221150255_add_subject_indexes.rb
+++ b/db/migrate/20200221150255_add_subject_indexes.rb
@@ -1,0 +1,6 @@
+class AddSubjectIndexes < ActiveRecord::Migration[6.0]
+  def change
+    add_index :subject, :type
+    add_index :subject, :subject_code
+  end
+end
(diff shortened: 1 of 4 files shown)


gpeng:
[Unregistered subject and timeout errors in subject filtering](https://trello.com/c/7CgDwVtO/3018-unregistered-subject-and-timeout-errors-in-subject-filtering)

TobyRet:
> tbh, I still wonder about removing the unused `created_at` and `updated_at`, it doesn't seem they're in the way and if we do add a new subject one day it could be useful to understand what's going on. ¯\_(ツ)_/¯ But happy to give it a pass if you feel they're just noise.

I'm curious about the this as well @gpeng - seems like it would be useful meta info to know when a record was updated (especially as the type could possible change over time). If we do go ahead and remove do we then need to be consistent and do the same to other tables?

gpeng:
@TobyRet / @misaka I'm fine with leaving them and would generally just add them to pretty much everything but they're nil. If we're going to leave them we should populate some values in there as calling code shouldn't have to worry about `nil`s in those columns.

PR #1194, opened 2020-02-24
[3003] Add mailer for course updates

SamJamCul:
Co-authored-by: defong <de_fong2k@yahoo.com>

### Context

[See this card for more details](https://trello.com/c/Xnoghg3q/3003-s-notify-accredited-body-users-that-something-has-changed-on-a-course)

There is also a Notify template that needs to be finalised, which this will directly link to. There's a template ID that will need to be included in the live environments as a result. 

Adds a mailer to mailer set that can be used to invoke an API call to notify.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/mailers/course_update_email_mailer.rb
+++ b/app/mailers/course_update_email_mailer.rb
@@ -1,0 +1,29 @@
+class CourseUpdateEmailMailer < GovukNotifyRails::Mailer
+  def course_update_email(course, attribute_changed, user)
+    set_template(Settings.govuk_notify.course_update_email_template_id)
+
+    set_personalisation(
+      provider_name: course.provider.provider_name,
+      course_name: course.name,
+      course_code: course.course_code,
+      attribute_changed: attribute_changed,
+      attribute_change_datetime: format_update_datetime(course.updated_at),
+      course_url: create_course_url(course),
+    )
+
+    mail(to: user.email)
+  end
+
+private
+
+  def create_course_url(course)
+    "#{Settings.publish_url}" \
+      "/organisations/#{course.provider.provider_code}" \
+      "/#{course.provider.recruitment_cycle.year}" \
+      "/courses/#{course.course_code}"
+  end
+
+  def format_update_datetime(datetime)
+    datetime.strftime("%-k.%M on %-e %B %Y")
+  end
+end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -17,6 +17,8 @@ current_recruitment_cycle_year: 2020
 govuk_notify:
   api_key: please_change_me
   welcome_email_template_id: please_change_me
+  course_update_email_template_id: please_change_me
+publish_url: http://localhost:3000
 mcbg:
   redis_password: <%= SecureRandom.base64 %>
 system_authentication_token: <%= SecureRandom.base64 %>
--- a/spec/mailers/course_update_email_mailer_spec.rb
+++ b/spec/mailers/course_update_email_mailer_spec.rb
@@ -1,0 +1,50 @@
+require "rails_helper"
+
+describe CourseUpdateEmailMailer, type: :mailer do
+  let(:course) { create(:course, :with_accrediting_provider, updated_at: DateTime.new(2001, 2, 3, 4, 5, 6)) }
+  let(:user) { create(:user) }
+  let(:mail) { described_class.course_update_email(course, "name", user) }
+
+  before do
+    course
+    mail
+  end
+
+  context "sending an email to a user" do
+    it "sends an email with the correct template" do
+      expect(mail.govuk_notify_template).to eq(Settings.govuk_notify.course_update_email_template_id)
+    end
+
+    it "sends an email to the correct email address" do
+      expect(mail.to).to eq([user.email])
+    end
+
+    it "includes the provider name in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:provider_name]).to eq(course.provider.provider_name)
+    end
+
+    it "includes the course name in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:course_name]).to eq(course.name)
+    end
+
+    it "includes the course code in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:course_code]).to eq(course.course_code)
+    end
+
+    it "includes the updated detail in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:attribute_changed]).to eq("name")
+    end
+
+    it "includes the datetime for the detail update in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:attribute_change_datetime]).to eq("4.05 on 3 February 2001")
+    end
+
+    it "includes the URL for the course in the personalisation" do
+      url = "#{Settings.publish_url}" \
+        "/organisations/#{course.provider.provider_code}" \
+        "/#{course.provider.recruitment_cycle.year}" \
+        "/courses/#{course.course_code}"
+      expect(mail.govuk_notify_personalisation[:course_url]).to eq(url)
+    end
+  end
+end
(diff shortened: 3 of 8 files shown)


SamJamCul:
[[S] Notify accredited body users that something has changed on a course](https://trello.com/c/Xnoghg3q/3003-s-notify-accredited-body-users-that-something-has-changed-on-a-course)

app/mailers/course_update_email_mailer.rb
```diff
@@ -0,0 +1,29 @@
...
+      course_url: create_course_url(course)
+    )
+
+    mail(to: course.accrediting_provider.email)
+  end
+
+  private
+
+  def create_course_url(course)
+    "https://www.publish-teacher-training-courses.service.gov.uk" \
```
TobyRet: Can the URL be put into config for consistency?

app/mailers/course_update_email_mailer.rb
```diff
@@ -0,0 +1,29 @@
+class CourseUpdateEmailMailer < GovukNotifyRails::Mailer
+  def send_course_update_email(course, attribute_changed)
```
gpeng: This should just be `course_update_email`. This method just creates it, it will be sent with `course_update_email(...).deliver`

app/models/course.rb
```diff
@@ -11,14 +11,14 @@
 #  discarded_at              :datetime
 #  english                   :integer
 #  id                        :integer          not null, primary key
-#  is_send                   :boolean          default(FALSE)
+#  is_send                   :boolean          default("false")
```
TobyRet: Not sure how these annotation updates relate to the commit. Would a separate commit be better?

app/mailers/course_update_email_mailer.rb
gpeng: The base URL should come from settings as it will differ in different environments.

spec/mailers/course_update_email_mailer_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    it "Includes the course code in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:course_code]).to eq(course.course_code)
+    end
+
+    it "Includes the updated detail in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:attribute_changed]).to eq("name")
+    end
+
+    it "Includes the datetime for the detail update in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:attribute_change_datetime]).to eq(course.updated_at.strftime("%k.%M on %e %B %y"))
```
gpeng: Can we hard code the expected updated date so it is easier to confirm the formatting.

spec/mailers/course_update_email_mailer_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:course) { create(:course, :with_accrediting_provider) }
+  let(:mail) { described_class.send_course_update_email(course, "name") }
+
+  before do
+    course
+    mail
+  end
+
+  context "Sending an email to a user" do
+    it "Sends an email with the correct template" do
```
TobyRet: Kind of nit-picky but think the convention is to down-case the first character of the string (`context` & `it`)

spec/mailers/course_update_email_mailer_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+describe CourseUpdateEmailMailer, type: :mailer do
+  let(:course) { create(:course, :with_accrediting_provider) }
+  let(:mail) { described_class.send_course_update_email(course, "name") }
+
+  before do
+    course
+    mail
+  end
+
+  context "Sending an email to a user" do
```
gpeng: We don't usually capitalise the context, it etc strings

app/models/course.rb
SamJamCul: I think the linter keeps catching this or something, I'll remove it from the commits.

spec/mailers/course_update_email_mailer_spec.rb
SamJamCul: I might need a hand doing some kind of time seed thing, do you know how that would work? I know you can use timecop to freeze it, but that wouldn't mean I could put a hard string in there.

config/settings.yml
```diff
@@ -17,6 +17,8 @@ current_recruitment_cycle_year: 2020
 govuk_notify:
   api_key: please_change_me
   welcome_email_template_id: please_change_me
+  course_update_email_template_id: please_change_me
+  course_base_url: please_change_me
```
defong: Can you stick local host for find please

It more like find_url as it's the base for generating course url. 

Then we can move it to its own and not under govuk_notify

app/mailers/course_update_email_mailer.rb
```diff
@@ -0,0 +1,29 @@
...
+private
+
+  def create_course_url(course)
+    "#{Settings.govuk_notify.course_base_url}" \
+      "/organisations/#{course.provider.provider_code}" \
+      "/#{course.provider.recruitment_cycle.year}" \
+      "/courses/#{course.course_code}"
+  end
+
+  def parse_update_datetime(datetime)
```
defong: strftime - format date and time

So why parse?

spec/mailers/course_update_email_mailer_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    mail
+  end
+
+  context "sending an email to a user" do
+    it "sends an email with the correct template" do
+      expect(mail.govuk_notify_template).to eq(Settings.govuk_notify.course_update_email_template_id)
+    end
+
+    it "sends an email to the correct email address" do
+      expect(mail.to).to eq([course.accrediting_provider.email])
```
defong: I'm not sure if that is the correct email address to use? Should it not be the users that are associated with that provider?

app/mailers/course_update_email_mailer.rb
SamJamCul: good point!

spec/mailers/course_update_email_mailer_spec.rb
SamJamCul: It's not clear from the ticket who the email is meant to be sent to - it just says "As an accredited body, I need to receive a notification". I think there's a bit of policy decision here to be made! I'll bring it up at standup

config/settings.yml
SamJamCul: Is it find or publish url? It's looking at the course page, but from publish rather than find.

app/mailers/course_update_email_mailer.rb
```diff
@@ -0,0 +1,29 @@
+class CourseUpdateEmailMailer < GovukNotifyRails::Mailer
+  def course_update_email(course, attribute_changed)
```
danielburnley: @SamJamCul I think this should take an argument of the user that is emailed - this will later be determined from the UserNotifications table (or whatever we call it) - I wouldn't expect the mailer to be doing the logic of deciding who to email

SamJamCul: Cool, can do.

danielburnley: To expand on this:

The mailer itself should be responsible simply for sending out the email, in this case (given it's notify) that also includes adding the correct personalisations

I would expect the subsequent service that gets created to handle the logic of determining: 

- Who gets emailed
- The course to email about
- And what attribute is changed

As users will be able to turn it off and on by themselves

spec/mailers/course_update_email_mailer_spec.rb
SamJamCul: I did what Dan suggested and adding a user argument, which we can then draw the correct email from. Don't need to worry about it for now!

dankmitchell:
This looks great. Nice work @defong and @SamJamCul 💯

config/settings.yml
defong: opps pub thats what i meant

PR #1195, opened 2020-02-24
[3016] Add send_courses filtering to the V3 

AgaDufrat:
We need to allow to filter `send_courses` as it is [used by Find](https://github.com/DFE-Digital/find-teacher-training/blob/master/app/view_objects/results_view.rb#L108) 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -210,6 +210,10 @@ class Course < ApplicationRecord
     joins(:provider).merge(Provider.where(provider_name: provider_name))
   end
 
+  scope :with_send, -> do
+    where(is_send: true)
+  end
+
   def self.entry_requirement_options_without_nil_choice
     ENTRY_REQUIREMENT_OPTIONS.reject { |option| option == :not_set }.keys.map(&:to_s)
   end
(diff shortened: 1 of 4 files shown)


AgaDufrat:
[[M] Add `send_courses` filtering to the API](https://trello.com/c/wm28L7cp/3016-add-sendcourses-filtering-to-the-api)

app/models/course.rb
```diff
@@ -210,6 +210,10 @@ def new_draft_attributes
     joins(:provider).merge(Provider.where(provider_name: provider_name))
   end
 
+  scope :with_send, -> do
```
TobyRet: Might be worth considering a recent related discussion regarding scopes
https://github.com/DFE-Digital/teacher-training-api/pull/1188#discussion_r382725562

AgaDufrat: Should I remove the scope and do 
`scope = Course.where(is_send: true) if send_courses_filter?` in the service? 🤔

PR #1199, opened 2020-02-25
[3047] Correctly show if site_status has vacancies

danielburnley:
### Context

New Find was displaying `Yes` for all sites with vacancies. This was due to
incorrectly calculating the value on the model when serialising.

- Change `has_vacancies?` to correctly return true/false

### Guidance to review

- Set a course to have a site with no vacancies
- Look at the course on find
- The site should show `No` for vacancies

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -100,7 +100,7 @@ class SiteStatus < ApplicationRecord
   end
 
   def has_vacancies?
-    SiteStatus.findable.with_vacancies.any?
+    with_vacancies?
   end
 
 private
--- a/spec/requests/api/v2/providers/courses_spec.rb
+++ b/spec/requests/api/v2/providers/courses_spec.rb
@@ -27,7 +27,7 @@ describe "Courses API v2", type: :request do
            study_mode: :full_time,
            subjects: subjects,
            is_send: true,
-           site_statuses: [courses_site_status],
+           site_statuses: courses_site_statuses,
            enrichments: [enrichment],
            maths: :must_have_qualification_at_application_time,
            english: :must_have_qualification_at_application_time,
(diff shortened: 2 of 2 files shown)


danielburnley:
https://trello.com/c/gVgG1oFZ/3047-no-vacancies-not-appearing-correctly-on-find

timabell:
should there be a change to the tests for this?

app/models/site_status.rb
```diff
@@ -100,7 +100,7 @@ def self.default_vac_status_given(study_mode:)
   end
 
   def has_vacancies?
-    SiteStatus.findable.with_vacancies.any?
+    with_vacancies?
```
gpeng: It would be faster to keep the `any?` on here and we're expecting it to return a boolean.

```
2020-02-26 09:42:07.047078 D [56616:70222132461600] (0.407ms) ActiveRecord -- SiteStatus Exists? -- { :sql => "SELECT 1 AS one FROM \"course_site\" WHERE \"course_site\".\"vac_status\" != $1 LIMIT $2", :binds => { :vac_status => "", :limit => 1 } }
2020-02-26 09:42:24.551091 D [56616:70222132461600] (42.8ms) ActiveRecord -- SiteStatus Load -- { :sql => "SELECT \"course_site\".* FROM \"course_site\" WHERE \"course_site\".\"vac_status\" != $1", :binds => { :vac_status => "" } }
```

The difference probably wouldn't be as dramatic here as it's a smaller result set that would be returned but every little bit helps 😄

danielburnley: `with_vacancies?` returns a boolean already so i'm not sure I can use `.any?` :thinking:

danielburnley: The old method was incorrect in that it operated on a result set rather than a single instance

gpeng: Ah yes. So it does. Misread it 👍

danielburnley:
Rebasing and merging :tada:

PR #1201, opened 2020-02-26
[2835] Add location filtering

faissalMT:
### Context

To support location filtering on Find the we need to add it to the API

- Change to geokit-rails
- Add the filter (latitude+longitude+radius) to the v3 course index API

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


dankmitchell:
[[L] location filter: Add filtering to api](https://trello.com/c/UkVHbumo/2835-l-location-filter-add-filtering-to-api)

config/initializers/geokit_config.rb
```diff
@@ -0,0 +1,100 @@
+# These defaults are used in Geokit::Mappable.distance_to and acts_as_mappable
+Geokit::default_units = :miles # others :kms, :nms, :meters
```
TobyRet: @faissalMT - is the Redis cache enabled for this as was for the `geocoder` gem?

TobyRet: Just spoken to @misaka about this and perhaps not so important to have the caching enabled now that the geocoding backfilling has been done

app/models/course.rb
```diff
@@ -157,6 +157,10 @@ def new_draft_attributes
     end
   end
 
+  scope :within, ->(range, origin:) do
+    joins(:sites).merge(Site.within(range, origin: origin))
+  end
```
misaka: I thought I was able to do `acts_as_mappable` here.

PR #1202, opened 2020-02-26
[3030] Add health check endpoint to test DB,Redis & Sidekiq

aliuk2012:
### Context
We need an endpoint that we can use to test that the API's DB, Redis Instance and Sidekiq is all working. We will use this to alert us when there are issues with the platform.

As a member of the team
I would like to know when the app is having an outage
So that I can respond as soon as possible

Add healthcheck endpoints to our services that report the status of any dependencies the app has, i.e. DB connection, Redis connection, API connection (using /ping), etc.

### Success

* We have an endpoint that checks that the API can connect to:
  * DB
  * Redis
  * sidekiq
* The status is emitted as JSON.

Returns the following JSON:

```json
{
  "checks": {
    "database": true,
    "redis": true,
    "sidekiq": true,
    "sidekiq_queue": true
  }
}
```
### Guidance to review
run `bundle exec rails s` and then in the browser go to `localhost:3001/healthcheck` to see the results. Try stopping your db,redis and sidekiq and test the results.
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


aliuk2012:
[Add /healthcheck endpoint to ttapi](https://trello.com/c/XNOMhtNQ/3030-add-healthcheck-endpoint-to-ttapi)

aliuk2012:
[Add /healthcheck endpoint to ttapi](https://trello.com/c/XNOMhtNQ/3030-add-healthcheck-endpoint-to-ttapi)

aliuk2012:
[Add /healthcheck endpoint to ttapi](https://trello.com/c/XNOMhtNQ/3030-add-healthcheck-endpoint-to-ttapi)

aliuk2012:
[Add /healthcheck endpoint to ttapi](https://trello.com/c/XNOMhtNQ/3030-add-healthcheck-endpoint-to-ttapi)

aliuk2012:
[Add /healthcheck endpoint to ttapi](https://trello.com/c/XNOMhtNQ/3030-add-healthcheck-endpoint-to-ttapi)

TobyRet:
@aliuk2012 - are there plans to set up a cron job to ping the health check endpoint? What will we be using to monitor success / failure?

aliuk2012:
I'm not 100% sure, I'll leave @misaka to answer those very valid questions.

timabell:
I think I like `/ping` as an endpoint, it's easy to remember

timabell:
I'd squash all those commits into one.

aliuk2012:
I was going to repurpose the existing `/ping` endpoint but decided against it for two reasons
- I wasn't sure where we were using it and making changes might break things
- `/ping` could be a very cheap way just to check a connection can be made without querying activerecord/sidekiq and redis.

aliuk2012:
I'll squash them once this PR is ready to be merged

timabell:
Works for me :heart: 
Good job!

```
$ curl localhost:3001/healthcheck
{"checks":{"database":true,"redis":false,"sidekiq":false,"sidekiq_queue":false}}%                                                                                                
```

app/controllers/heartbeat_controller.rb
```diff
@@ -0,0 +1,51 @@
+require "sidekiq/api"
+
+class HeartbeatController < ActionController::API
+  def ping
```
timabell: do we need this as well as `healthcheck`? seems redundant

timabell:
I think we should kill the old ping-pong unless we know for sure it's needed. Clean code and all that. We can still tell the difference between dependency failure (`502`) and complete service failure (`500` or connection failure) with the new endpoint.

timabell:
* front-door spams the hell out of `/ping` so wouldn't want to make that expensive
* swap-slot uses ping too, would be good for that to switch to the advanced healthcheck in this PR

:ship: it

app/controllers/heartbeat_controller.rb
misaka: I say we keep it for stuff like Frontdoor that pings several times a second.

app/controllers/heartbeat_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def healthcheck
+    checks = {
+      database: database_alive?,
+      redis: redis_alive?,
+      sidekiq: sidekiq_alive?,
+      sidekiq_queue: sidekiq_queue_healthy?,
+    }
+
+    status = :bad_gateway unless checks.values.all?
```
misaka: This could be clearer that we'll end up with a success code.

PR #1205, opened 2020-02-26
[3055] All modern languages are now returned in edit options

faissalMT:
Co-authored-by: asmega <phil.lee@digital.education.gov.uk>

### Context
The way we currently handle modern languages means that frontends can't always tell which subjects are modern language or regular subjects, and cannot tell whether or not a given subject they have is the modern language subject. This makes it difficult to create a wizard that allows for the modification of modern language subjects in an existing course, or to allow us to decide whether or not a given subject should be replaced by a new value when modifying a subject during course creation.

Always send modern languages via `edit_options` and send details regarding the modern language subject.

### Guidance to review
This change will break the the modern language pages, as they will infer that they need to change the modern languages from the fact that modern languages are being sent back, rather than being `nil`, hence we'll want to hold back this card until the frontends are updated.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/requests/api/v3/providers/courses/index_spec.rb
```diff
@@ -233,7 +233,121 @@ def perform_request
...
+                        "subject_code" => "24",
+                        "bursary_amount" => "26000",
+                        "early_career_payments" => "2000",
+                        "scholarship" => nil,
+                        "subject_knowledge_enhancement_course_available" => true,
+                      },
+                    },
+                  ],
+                  "modern_languages_subject" => {
+             "id" => 33,
```
defong: linter?

PR #1211, opened 2020-02-28
[3025] Add User notifications table

danielburnley:
### Context

In order to control which users have notifications on for which provider - we need to add a join
table between the `User` and `Provider`.

- Add a join table between user and provider

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/db/migrate/20200227105316_add_user_notifications_table.rb
+++ b/db/migrate/20200227105316_add_user_notifications_table.rb
@@ -1,0 +1,14 @@
+class AddUserNotificationsTable < ActiveRecord::Migration[6.0]
+  def change
+    create_table :user_notification do |t|
+      t.integer :user_id, null: false
+      t.string :provider_code, null: false
+      t.boolean :course_update, default: false
+
+      t.timestamps
+    end
+
+    add_foreign_key :user_notification, :user
+    add_index :user_notification, :provider_code
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_02_21_151215) do
+ActiveRecord::Schema.define(version: 2020_02_27_105316) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_buffercache"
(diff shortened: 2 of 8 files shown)


db/schema.rb
```diff
@@ -303,6 +303,14 @@
     t.index ["email"], name: "IX_user_email", unique: true
   end
 
+  create_table "user_notification", force: :cascade do |t|
+    t.integer "user_id"
```
defong: can be nullable?

db/schema.rb
```diff
@@ -303,6 +303,14 @@
...
+    t.string "provider_code"
```
defong: can be nullable?

db/schema.rb
faissalMT:
```suggestion
    t.integer "user_id", null: false
```

db/schema.rb
faissalMT:
```suggestion
    t.integer "provider_code", null: false
```

spec/models/user_notification_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      let(:provider_one) { create(:provider) }
+      let(:provider_two) { create(:provider) }
+
+      before do
+        UserNotification.new(user_id: user.id, provider_code: provider_one.provider_code).save!
+        UserNotification.new(user_id: user.id, provider_code: provider_two.provider_code).save!
+      end
+
+      it "Returns only user notifications the first provider" do
+        notifications = UserNotification.for_provider(provider_one.provider_code).all
```
defong:
```suggestion
        notifications = described_class.for_provider(provider_one.provider_code)
```
maybe use `described_class`

I don't think there is a need for `.all`

I'm not sure  of `for_provider`

```ruby 
UserNotification.find_by(provider_code: provider_one.provider_code)
```

In fact do we actually need it....
```ruby
provider.user_notifications # give lists of users...
course.provider.user_notifications # give lists of users...
course.accrediting_provider.user_notifications # give lists of users that belongs to accrediting_provider...
```

`for_provider` I leave that for your consideration

db/migrate/20200227105316_add_user_notifications_table.rb
```diff
@@ -0,0 +1,13 @@
+class AddUserNotificationsTable < ActiveRecord::Migration[6.0]
+  def change
+    create_table :user_notification do |t|
+      t.integer :user_id, null: false
+      t.string :provider_code, null: false
```
defong: maybe index  something similar to 

https://github.com/DFE-Digital/teacher-training-api/blob/f4cedf7c693592147600f4131405b21be343eb9d/db/migrate/20190711074433_change_course_accrediting_provider_id_to_code.rb#L8

spec/models/user_notification_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+#  updated_at    :datetime         not null
+#  user_id       :integer          not null
+#
+describe UserNotification, type: :model do
+  describe "associations" do
+    let(:organisation) { create(:organisation, providers: [provider]) }
+    let(:user) { create(:user, organisations: [organisation]) }
+    let(:provider) { create(:provider) }
+
+    subject { UserNotification.new(user_id: user.id, provider_code: provider.provider_code) }
```
defong:
```suggestion
    subject { described_class.new(user_id: user.id, provider_code: provider.provider_code) }
```

maybe use `described_class`

spec/models/user_notification_spec.rb
danielburnley: Described class could work for sure - will give it a poke

Though I do think you're right here, the trello task implied that methods would be needed, but didnt specify. I think I could scrap this and do the follow up card and see if any methods are required after doing that

db/migrate/20200227105316_add_user_notifications_table.rb
danielburnley: :eyes:

PR #1212, opened 2020-02-28
[3061] Can return all courses for sitemap

asmega:
### Context

- https://trello.com/c/zOB0CYyM/3061-fix-sitemap
- Sitemap currently only able to generate 10 courses due to pagination
- This is the backend fix to remove pagination from courses if correct parameters are sent

- If correct params are given to courses endpoint we remove pagination
and return all courses
- So that it can be used by Find to generate sitemap
- Switch v3 courses controller to use v3 courses serialiser
- Courses controller now always includes provider association
- This is to prevent n+1 queries
- jsonapi :include is not used as causes many additional SQL queries
- Test slices out `updated_at` as cannot use `Timecop` as multiple
providers cannot be created at the same time due to uniqueness index

### Guidance to review

- Visit http://localhost:3001/api/v3/recruitment_cycles/2020/courses?fields[courses]=course_code,provider_code,updated_at
- Should return all records without pagination
- Should only return enough data to generate sitemap client side

### Checklist

- [X] Attach to Trello card
- [x] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/courses_controller.rb
+++ b/app/controllers/api/v3/courses_controller.rb
@@ -6,9 +6,21 @@ module API
       before_action :build_courses
 
       def index
-        course_search = CourseSearchService.call(filter: params[:filter], sort: params[:sort], course_scope: @courses)
+        course_search = CourseSearchService.call(filter: params[:filter],
+                                                 sort: params[:sort],
+                                                 course_scope: @courses)
 
-        render jsonapi: paginate(course_search), fields: fields_param, include: params[:include], meta: { count: course_search.count }, class: CourseSerializersService.new.execute
+        results = if disable_pagination?
+                    course_search
+                  else
+                    paginate(course_search)
+                  end
+
+        render jsonapi: results,
+          fields: fields_param,
+          include: params[:include],
+          meta: { count: results.count },
+          class: CourseSerializersServiceV3.new.execute
       end
 
       def show
@@ -24,11 +36,21 @@ module API
 
     private
 
+      def disable_pagination?
+        if params[:fields] && params[:fields][:courses]
+          (params[:fields][:courses].split(",") & fields_required_to_disable_pagination).size == 3
+        end
+      end
+
+      def fields_required_to_disable_pagination
+        %w[course_code provider_code changed_at]
+      end
+
       def build_courses
         @courses = if @provider.present?
-                     @provider.courses
+                     @provider.courses.includes(:provider)
                    else
-                     @recruitment_cycle.courses
+                     @recruitment_cycle.courses.includes(:provider)
                    end
       end
 
--- a/app/serializers/api/v3/serializable_course.rb
+++ b/app/serializers/api/v3/serializable_course.rb
@@ -19,7 +19,7 @@ module API
                  :content_status, :ucas_status, :funding_type,
                  :level, :is_send?, :english, :maths, :science, :gcse_subjects_required,
                  :age_range_in_years, :accrediting_provider,
-                 :accrediting_provider_code, :level
+                 :accrediting_provider_code, :level, :changed_at
 
       attribute :start_date do
         @object.start_date.strftime("%B %Y") if @object.start_date
--- a/Gemfile
+++ b/Gemfile
@@ -104,7 +104,7 @@ gem "request_store"
 gem "dry-container"
 
 # For geocoding and geographic logic (e.g: filtering sites by ranges)
-gem 'geokit-rails'
+gem "geokit-rails"
 
 group :development, :test do
   # add info about db structure to models and other files
--- a/spec/requests/api/v3/courses/index_spec.rb
+++ b/spec/requests/api/v3/courses/index_spec.rb
@@ -512,4 +512,41 @@ describe "GET v3/courses" do
       expect(headers["Total"]).to be_present
     end
   end
+
+  describe "when requesting course_code, provider_code, changed_at fields for sitemap" do
+    around do |example|
+      max_per_page = Kaminari.config.max_per_page
+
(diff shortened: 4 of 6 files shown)


app/serializers/api/v3/serializable_course.rb
```diff
@@ -19,7 +19,7 @@ def enrichment_attribute(name, enrichment_name = name)
                  :content_status, :ucas_status, :funding_type,
                  :level, :is_send?, :english, :maths, :science, :gcse_subjects_required,
                  :age_range_in_years, :accrediting_provider,
-                 :accrediting_provider_code, :level
+                 :accrediting_provider_code, :level, :updated_at
```
dankmitchell: I think we need ‘changed_at’ instead as that looks at all associations too

dankmitchell:
[Fix sitemap (backend)](https://trello.com/c/zOB0CYyM/3061-fix-sitemap)

app/controllers/api/v3/courses_controller.rb
```diff
@@ -24,11 +36,19 @@ def show
 
     private
 
+      # Used for sitemap on Find
+      # When these fields are specified return all results for sitemap
```
danielburnley: Could this comment be removed by improving the readability of the code?

app/controllers/api/v3/courses_controller.rb
```diff
@@ -24,11 +36,19 @@ def show
...
+      def disable_pagination?
+        if params[:fields] && params[:fields][:courses]
+          (params[:fields][:courses].split(",") & %w[course_code provider_code updated_at]).size == 3
+        end
+      end
```
danielburnley: Disabling pagination via a specific combination of `fields` feels... odd to me? I'd expect something more explict and less surprising but i'm not sure what it could be. 

At the very least I think this could read better if the check was something along the lines of:

```ruby
params[:fields][:courses].split(",") == fields_required_To_disable_pagination
``` 

Or something similarly explict

misaka: I think I agree. I had in mind an approach where the user could disable pagination / increase the page size over a threshold, and if so requested it would then ensure that the fields requested are appropriate (any of `%w[course_code provider_code updated_at]`). If any other fields are requested, it would raise a 4XX error, perhaps with info about which fields can/can't be requested.

app/controllers/api/v3/courses_controller.rb
```diff
@@ -24,11 +36,19 @@ def show
...
+
       def build_courses
         @courses = if @provider.present?
                      @provider.courses
```
misaka: Does this automatically include providers?

asmega: It does not. I have now added this

app/controllers/api/v3/courses_controller.rb
asmega: done

app/serializers/api/v3/serializable_course.rb
asmega: I wasn't aware of this. It has now been changed :)

app/controllers/api/v3/courses_controller.rb
asmega: I've have made @danielburnley suggestion of extracting to a method. Although `==` cannot be used as ordering is not guaranteed.

As for the more general note of how we disable/enable pagination and flow I think this needs to be done at a higher level overall otherwise we'll get shotgun snippets of code like this everywhere. That I think is outside the scope of this ticket

misaka: Why is this outside the scope of this ticket? This ticket is about that very thing, being able to disable pagination, so we should implement it in a fitting way, unless it's a big enough piece of work that warrants being pulled out.

asmega: If the client wants to disable pagination they should be able to add a flag/param for this which we should respect. If the client wants fewer or custom fields they should ask for that too.

If we feel like we're giving out too much info (fields-wise) I think that means that the data should be on another endpoint which the client would have to make a request to.

If we're concerned about who does this then ie which clients. may this should be done at an authorisation level eg. only these sets of api keys can disable pagination as a feature

If we're concerned about the amount of data sent because it's the whole dataset and therefore performance issues then I think pagination should not be disabled and the client will have to simply paginate through the dataset

misaka: I'm not sure I understand why this means that we should implement this as a magical "we'll disable pagination if you specify the right fields" approach.

My point is that the client should be in charge of defining the request it's making, and it shouldn't get _ALL_ the results back if it happens to decide that it only needs `course_code`, `provider_code` and `updated_at`. That would probably be surprising, and being a public API we shouldn't make assumptions about how it's used, and try to be as unsurprising as possible.

asmega: Ok, I see. I can do `?paginate=false` defaults to true and only if these fields are specified then disable pagination. My point is that the feature of disabling pagination should that only be applicable for this endpoint? Otherwise we'll have other areas where we'll end up having to do the same thing

misaka: Ya, so that's a good point that this mechanism should be more widely defined. I'd be happy if we hacked this in for now with `per_page=65535` or something like that. At least it'll be less surprising.

asmega:
I'm closing this in favour of https://github.com/DFE-Digital/teacher-training-api/pull/1220 to solve the pagination problem first

asmega:
https://github.com/DFE-Digital/teacher-training-api/pull/1224 now replaces this PR

PR #1217, opened 2020-03-03
[3069] Sort courses by distance

asmega:
### Context

- https://trello.com/c/8wGCf2P3/3069-m-sort-courses-by-distance-to-sites
- Backend can now sort results by distance from origin

- Able to sort results by distance from origin
- Does not implement parameter checking

### Guidance to review

- Given you have a dataset with geocoded sites
- via rails console `Course.limit(100).by_distance(origin: [54.9713392, -1.6112336])`
- or other origin. this should return results sorted by distance from origin with closest first
- the should be using the geocoded data from `Site`

### Checklist

- [X] Attach to Trello card
- [x] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/spec/services/course_search_service_spec.rb
+++ b/spec/services/course_search_service_spec.rb
@@ -44,6 +44,18 @@ describe CourseSearchService do
         end
       end
 
+      context "by distance" do
+        let(:sort) { "distance" }
+        let(:filter) do
+          { latitude: 54.9713392, longitude: -1.6112336 }
+        end
+
+        it "orders in descending order" do
+          expect(findable_scope).to receive(:by_distance).with(origin: [54.9713392, -1.6112336]).and_return(expected_scope)
+          expect(subject).to eq(expected_scope)
+        end
+      end
+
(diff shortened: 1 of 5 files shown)


spec/services/course_search_service_spec.rb
```diff
@@ -44,6 +44,18 @@
...
+        let(:filter) do
+          { latitude: 54.9713392, longitude: -1.6112336 }
+        end
+
+        it "orders in descending order" do
+          expect(findable_scope).to receive(:by_distance).with(origin: [54.9713392, -1.6112336]).and_return(expected_scope)
+          expect(subject).to eq(expected_scope)
+        end
+      end
+
```
misaka: Worth also adding a section to the request spec for ordering?

https://github.com/DFE-Digital/teacher-training-api/blob/2890-add-openapi-spec-for-courses/spec/requests/api/v3/courses/index_spec.rb#L93

asmega: yep, now added

PR #1220, opened 2020-03-03
[3061] Able to disable pagination for v3

asmega:
### Context

- https://trello.com/c/zOB0CYyM/3061-fix-sitemap-backend
- We firstly need to solve the problem of being able to disable pagination before we even begin to fix the sitemap
- This PR allows v3 endpoints to disable pagination if desired
- There will be a subsequent PR to fix the sitemap

- If requested pagination is over the default number we check if this is allowed
- This check is done by implementing `allowed_to_disable_pagination?`
- If not implemented or returns false then we fallback to the default

### Guidance to review

- Existing endpoints should still be paginated to 100
- Implement `allowed_to_disable_pagination?` on a controller to return true
- if `params[:page][:per_page]` provided pagination should be respected
- There will be a subsequent PR to fix the sitemap

### Checklist

- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/application_controller.rb
+++ b/app/controllers/api/v3/application_controller.rb
@@ -3,11 +3,29 @@ module API
     class ApplicationController < ActionController::API
       rescue_from ActiveRecord::RecordNotFound, with: :jsonapi_404
 
+      before_action :check_disable_pagination
+
       def jsonapi_404
         render jsonapi: nil, status: :not_found
       end
 
     private
+
+      def check_disable_pagination
+        return unless params[:page]
+
+        if params[:page][:per_page].to_i > Kaminari.config.default_per_page
+          return if allowed_to_disable_pagination?
+
+          params[:page][:per_page] = Kaminari.config.default_per_page
+        end
+      end
+
+      # Override if you want to allow an endpoint to disable pagaination
+      # e.g. by checking params
+      def allowed_to_disable_pagination?
+        false
+      end
 
       def build_recruitment_cycle
         @recruitment_cycle = RecruitmentCycle.find_by(
--- a/config/initializers/kaminari_config.rb
+++ b/config/initializers/kaminari_config.rb
@@ -2,7 +2,7 @@
 
 Kaminari.configure do |config|
   config.default_per_page = 100
-  config.max_per_page = 100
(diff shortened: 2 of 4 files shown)


app/controllers/api/v3/application_controller.rb
```diff
@@ -3,12 +3,22 @@ module V3
...
 
     private
 
+      def check_disable_pagination
+        if params[:page] && params[:page][:per_page].to_i > Kaminari.config.default_per_page
+          unless respond_to?(:allowed_to_disable_pagination?) && allowed_to_disable_pagination?
+            params[:page][:per_page] = Kaminari.config.default_per_page
+          end
+        end
+      end
```
misaka: How about using this as a guard:

```
return unless respond_to?(:allowed_to_disable_pagination?)
```

And then you can take it out of the unless on line 16, reducing it's Cognitive Complexity. :)

misaka: Actually, could also have the guard:

```
return unless params[:page]
```

asmega: i can do the second guard you suggested. however not the first guard as if that fails we need to fallback to the default

asmega: This has now been changed. `ApplicationController` implements `allowed_to_disable_pagination ?` and returns false. Subclasses can override accordingly. The guards have now been put in place

config/initializers/kaminari_config.rb
```diff
@@ -2,7 +2,7 @@
 
 Kaminari.configure do |config|
   config.default_per_page = 100
-  config.max_per_page = 100
```
misaka: Is 100,000 a bit big?

config/initializers/kaminari_config.rb
```diff
@@ -2,7 +2,7 @@
...
+  config.max_per_page = 100_000
```
misaka: Is 100,000 a bit big?

misaka: I was mixed up, thought this was being used in the way `default_per_page` is used. All good!

PR #1227, opened 2020-03-09
[3094] Added prevention for suggesting providers with no courses

defong:
### Context
prevention for suggesting providers with no courses

Added prevention for suggesting providers with no courses

### Guidance to review
Provider that are either the accredited body or the provider of a findable course will only be shown

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/provider_suggestions_controller.rb
+++ b/app/controllers/api/v3/provider_suggestions_controller.rb
@@ -7,12 +7,8 @@ module API
         return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
         return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
-        providers_ids = @recruitment_cycle.courses.findable.distinct.pluck(:provider_id)
-        accrediting_provider_codes = @recruitment_cycle.courses.findable.distinct.pluck(:accrediting_provider_code).compact
-
         found_providers = @recruitment_cycle.providers
-                              .where(provider_code: accrediting_provider_codes)
-                              .or(@recruitment_cycle.providers.where(id: providers_ids))
+                              .with_findable_courses
                               .search_by_code_or_name(params[:query])
                               .limit(10)
 
--- a/app/controllers/api/v3/provider_suggestions_controller.rb
+++ b/app/controllers/api/v3/provider_suggestions_controller.rb
@@ -7,7 +7,12 @@ module API
         return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
         return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
+        providers_ids = @recruitment_cycle.courses.findable.distinct.pluck(:provider_id)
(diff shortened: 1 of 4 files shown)


defong:
[Location search is currently showing providers without any courses](https://trello.com/c/RxZ8RwZm/3094-location-search-is-currently-showing-providers-without-any-courses)

spec/requests/api/v3/providers/suggest_provider_spec.rb
```diff
@@ -68,7 +70,8 @@
 
   it "limits responses to a maximum of 10 items" do
     11.times do
-      create(:provider, provider_name: "provider X")
+      courses = [build(:course, site_statuses: [build(:site_status, :findable)])]
```
faissalMT: `[build(:site_status, :findable)]` can probably be DRYed a bit

app/controllers/api/v3/provider_suggestions_controller.rb
```diff
@@ -7,7 +7,12 @@ def index
         return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
         return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
+        providers_ids = @recruitment_cycle.courses.findable.distinct.pluck(:provider_id)
```
gpeng: This logic shouldn't be here in the controller. Can we move it to a scope on `Provider`? Maybe `with_courses` or `with_active_courses` or something like that. Then the logic for `active_courses` should be on `Course` and the `Provider` scope would merge that in. They could both be easily tested in a bit more isolation.

app/controllers/api/v3/provider_suggestions_controller.rb
```diff
@@ -7,7 +7,12 @@ def index
...
+        accrediting_provider_codes = @recruitment_cycle.courses.findable.distinct.pluck(:accrediting_provider_code).compact
+
         found_providers = @recruitment_cycle.providers
+                              .where(provider_code: accrediting_provider_codes)
```
gpeng: ... and we could just add the provider scope in here.

spec/requests/api/v3/providers/suggest_provider_spec.rb
defong: not too sure what you mean...

site_status, comes from site, which is needed for provider

spec/models/provider_spec.rb
```diff
@@ -588,6 +588,47 @@
...
+        findable_course_with_accrediting_provider
+        non_findable_course
+        non_findable_course_with_accrediting_provider
+      end
+
+      subject {
+        described_class.with_findable_courses
+      }
+
+      it "should return only findable courses' provider and/or accrediting provider" do
```
gpeng: You could split this into contexts and it would be easier to read. 

e.g.
``` 
context "when the provider is the accredited body for a course"  do
  it "is returned"
end

context "when the course is delivered by the provider" do
  it "is returned"
end

context "when the course is unrelated to the provider" do
  it "is not returned"
end
```

PR #1228, opened 2020-03-09
Update spec readme to let-free style

timabell:
### Context

* `let` blocks introduce additional indirection
* The human brain [only has room for around 7 pieces of information in working memory](https://en.wikipedia.org/wiki/Working_memory#Capacity)
* The Apply team have been successful with this approach, and reports are that it is better to work with.
* @timabell spent Friday afternoon exceeding mental-stack on a test partly because of the amount of indirection.

This PR would begin the move towards less indirection in our specs and therefore less you have to keep in your head at once to reason about and modify the tests.

The PR intended to be a discussion point for the team.

If merged we should then write new specs in this style, and clean up previous specs when we have to work on them.

* Set `let`-free specs to be the new code style for our tests
* Also pulls in the given-when-then style to see if we feel like going whole-hog.

### Guidance to review

* discuss
* comment
* approve if you approve

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


timabell:
It seems to me from what I've seen/read that `let` causes extra mental overhead understanding code and as such should only be used where it's really necessary rather than as a default.

It did occur to me that we might be using let to allow re-use of data generated by FactoryBot, but I'm not clear on the boundaries between tests, so maybe it doesn't actually help.

spec/README.md
```diff
@@ -1,3 +1,102 @@
+# Testing styleguide
+
+We use ["Futurelearn style" acceptance tests](https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec). This style was originally from [the "apply" team's spec docs](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/ba9125e9aa5f212c18b63c7c3d49f8785486df83/docs/testing-styleguide.md)
+
+This style guide is newer than the codebase, so you will see a lot of `let` statements around. Use the boy-scout rule and update the style as you touch things.
+
+## Rules
+
+1. Use a single scenario per file. This prevents the files from becoming too large. Separate logical blocks of steps with newlines.
+2. Use instance variable to carry state between steps. Don't use `let` or `before` blocks.
```
misaka: I'm not entirely clear on why this is one rule, I think it should be two. e.g.:

2. Create your data in the `given_` block instead of using `let` statements.
3. Use instance variables to carry data and state between steps.

spec/README.md
```diff
@@ -1,3 +1,102 @@
+# Testing styleguide
+
+We use ["Futurelearn style" acceptance tests](https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec). This style was originally from [the "apply" team's spec docs](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/blob/ba9125e9aa5f212c18b63c7c3d49f8785486df83/docs/testing-styleguide.md)
```
misaka: I think it should be clear that this is talking about feature tests, or are we proposing that unit and request tests follow these rules too?

spec/README.md
```diff
@@ -17,27 +114,23 @@ To create a course, its provider and any other objects needed to create a valid
 course and save it to the database (obeying nullability and foreign key rules):
 
 ```
-let(:course)  { create :course }
+def when_a_course_exists  { create :course }
```
misaka: This would need to be assigned to an instance variable. Also, I think we should distinguish between feature tests and unit/request specs.

misaka: Same comment applies to the relevant changes below.

timabell:
I'm not really sure, perhaps we should get some input from the apply folks as to what worked for them best.

timabell:
I think this needs further consideration. Closing for now.

timabell:
@misaka I hadn't really considered the differences but I think you're right that maybe it should only be for feature tests having had a play with it on a real test.

PR #1229, opened 2020-03-09
[3086] De-duplicate search results

asmega:
### Context

- https://trello.com/c/c8h5V8zx/3086-qa-order-by-distance-not-always-correct
- This removes duplicates from search results
- Duplicates mainly due to courses having multiple sites

- when sorting by distance from origin deal with courses with multiple
sites
- only return a course once even if it has multiple sites
- use the closest site when sorting by distance
- only use site that are new or running
- now select `distinct` courses for search results
- meta count scopes to `course.id`
- add tests to cover duplicate results bug

### Guidance to review

- http://localhost:3002/results?fulltime=False&hasvacancies=True&l=1&lat=50.82253000000001&lng=-0.137163&loc=Brighton%2C+UK&lq=Brighton%2C+UK&parttime=False&qualifications=QtsOnly%2CPgdePgceWithQts%2COther&rad=20&senCourses=false&sortby=2&subjects=31
- There should be no duplicate courses
- Courses should be sorted by distance

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/courses_controller.rb
+++ b/app/controllers/api/v3/courses_controller.rb
@@ -11,7 +11,7 @@ module API
         render jsonapi: paginate(course_search),
           fields: fields_param,
           include: params[:include],
-          meta: { count: course_search.count },
+          meta: { count: course_search.count("course.id") },
           class: CourseSerializersServiceV3.new.execute
       end
 
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -14,9 +14,17 @@ class CourseSearchService
   def call
     scope = course_scope.findable
 
-    scope = scope.ascending_canonical_order if sort_by_provider_ascending?
-    scope = scope.descending_canonical_order if sort_by_provider_descending?
-    scope = scope.by_distance(origin: origin) if sort_by_distance?
+    if sort_by_provider_ascending?
+      scope = scope.ascending_canonical_order
+      scope = scope.select("provider.provider_name", "name")
+    elsif sort_by_provider_descending?
+      scope = scope.descending_canonical_order
+      scope = scope.select("provider.provider_name", "name")
+    elsif sort_by_distance?
+      scope = scope.joins(sites_with_distance_from_origin)
+      scope = scope.select("course.*, distance")
+      scope = scope.order(:distance)
+    end
 
     scope = scope.with_salary if funding_filter_salary?
     scope = scope.with_qualifications(qualifications) if qualifications.any?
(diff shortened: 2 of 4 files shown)


TobyRet:
Ran locally and working great 👍

app/controllers/api/v3/courses_controller.rb
```diff
@@ -11,7 +11,7 @@ def index
         render jsonapi: paginate(course_search),
           fields: fields_param,
           include: params[:include],
-          meta: { count: course_search.count },
+          meta: { count: course_search.count("course.id") },
```
TobyRet: What's the reasoning behind this change?

spec/requests/api/v3/courses/index_spec.rb
```diff
@@ -155,6 +155,57 @@
...
+
+          json_response = JSON.parse(response.body)
+          course_ids << json_response.dig("data", 0, "id")
+
+          get page_3
+
+          json_response = JSON.parse(response.body)
+          course_ids << json_response.dig("data", 0, "id")
+
+          expect(course_ids.compact.size).to eql(2)
```
defong: why `compact`?

spec/requests/api/v3/courses/index_spec.rb
```diff
@@ -155,6 +155,57 @@
...
+
+        let(:page_1) { "/api/v3/courses?sort=distance&latitude=0&longitude=0&page[page]=1" }
+        let(:page_2) { "/api/v3/courses?sort=distance&latitude=0&longitude=0&page[page]=2" }
+        let(:page_3) { "/api/v3/courses?sort=distance&latitude=0&longitude=0&page[page]=3" }
+
+        # jsonapi removes duplicates on the same page
+        # therefore this test needs to span multiple pages
+        # in order for the test to be valid
+
+        it "does not return duplicates" do
```
defong: can we have a explicit test

`          expect(course_ids).to eq ([1,2])

spec/requests/api/v3/courses/index_spec.rb
asmega: when the bug was present there were three pages. so the array had three items. it should only return 2 items. when the bug was fixed it returned `[1,2,nil]` as there is now nothing on the 3rd page. we have to compact to remove the extra result

spec/requests/api/v3/courses/index_spec.rb
asmega: yes, i will change that now

asmega: done

app/services/course_search_service.rb
```diff
@@ -14,9 +14,21 @@ def call(**args)
   def call
     scope = course_scope.findable
 
-    scope = scope.ascending_canonical_order if sort_by_provider_ascending?
-    scope = scope.descending_canonical_order if sort_by_provider_descending?
-    scope = scope.by_distance(origin: origin) if sort_by_distance?
+    if sort_by_provider_ascending?
+      scope = scope.ascending_canonical_order
+      scope = scope.select("provider.provider_name", "name")
+    end
```
misaka: Can the results be sorted by more than one field at a time? If not, which I think is the case, is it worth being clearer about this in the code:

```ruby
if sort_by_provider_ascending?
  ...
elsif sort_by_provider_descending?
  ...
elsif sort_by_distance?
  ...
end
```

asmega: done

PR #1230, opened 2020-03-10
DRAFT [3097] ignore sites with no address in location searches

gpeng:
**DRAFT - NOT READY FOR REVIEW**

### Context

Some of the site records in the DB contain inadequate address data. We have decided to continue to omit courses at these sites from search results until we can fix the data.

Add `Course.with_locatable_site` (which delegates to similar scopes on `SiteStatus` and `Site`) and use it to filter `CourseSearchService` results.

### Guidance to review

I think this code will remove courses without a valid site from all search results. The card only specifies it being removed from location searches and distance ordering. I need to check this before review.

I've added this as a totally separate set of scopes. I think it could also have been incorporated into `.findable` which would have resulted in less changes but I decided it was better to make it explicit as `findable` is already a bit 'mysterious' and adding to it would only make things worse. This code should also only be a stop gap while we clean the data, after which, this PR could be reverted.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -56,6 +56,8 @@ class Site < ApplicationRecord
 
   after_commit -> { GeocodeJob.perform_later("Site", id) }, if: :needs_geolocation?
 
+  scope :with_locatable_address, -> { where.not(address1: "").or(Site.unscoped.where.not(postcode: "")) }
+
   def needs_geolocation?
     full_address.present? && (
     latitude.nil? || longitude.nil? || address_changed?
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -223,6 +223,10 @@ class Course < ApplicationRecord
     where(is_send: true)
   end
 
+  scope :with_locatable_site, -> do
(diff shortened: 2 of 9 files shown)


app/models/site.rb
```diff
@@ -56,6 +56,8 @@ class Site < ApplicationRecord
 
   after_commit -> { GeocodeJob.perform_later("Site", id) }, if: :needs_geolocation?
 
+  scope :with_locatable_address, -> { where.not(address1: "").or(Site.unscoped.where.not(postcode: "")) }
```
defong:
```suggestion
  scope :with_locatable_address, -> { where.not(address1: "", postcode: "")) }
```

Can we keep it tight
https://github.com/DFE-Digital/teacher-training-api/blob/f41b6f576ac1225a4c00fb617a5e46670ad2e998/app/models/site.rb#L44-L47

gpeng: I think that would do an `AND` and we need an `OR` here. I'm closing this PR for now anyway for a rejig.

PR #1231, opened 2020-03-10
[3087] Tag background jobs correctly when logging

faissalMT:
### Context
We had a large number of errors appearing on Logit but we weren't aware of where they were coming from.

Add job id, queue name and job name when logging.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/jobs/bulk_sync_courses_to_find_job.rb
+++ b/app/jobs/bulk_sync_courses_to_find_job.rb
@@ -5,6 +5,9 @@ class BulkSyncCoursesToFindJob < ApplicationJob
   end
 
   def perform
+    RequestStore.store[:job_id] = provider_job_id
+    RequestStore.store[:job_queue] = queue_name
+
     syncable_courses = RecruitmentCycle.syncable_courses
 
     if syncable_courses.present?
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -35,12 +35,6 @@ Rails.application.configure do
   config.log_tags = [:request_id] # Prepend all log lines with the following tags.
   config.log_level = Settings.log_level
 
-  if Settings.logstash.host && Settings.logstash.port
-    config.logger = LogStashLogger.new(Settings.logstash.to_h)
-  else
-    config.logger = ActiveSupport::Logger.new(STDOUT)
-    config.logger.warn("logstash not configured, falling back to standard Rails logging")
-  end
   config.active_record.logger = nil # Don't log SQL in production
 
   # Use a different cache store in production.
--- a/config/initializers/semantic_logger.rb
+++ b/config/initializers/semantic_logger.rb
@@ -1,18 +1,66 @@
-if Settings.logstash.host && Settings.logstash.port
-  logstash_formatter = Proc.new do |event|
-    # The value here appears to break logging to logstash / elasticsearch
-    event["duration"] = event["duration_ms"]
-    event["duration_ms"] = nil
+class LogStashFormatter < SemanticLogger::Formatters::Raw
+  def format_json_message_context
+    if hash[:message].present?
+      context = JSON.parse(hash[:message])["context"]
+      hash[:sidekiq_job_context] = hash[:message]
+      hash[:message] = context
+    end
+  rescue JSON::ParserError
+    nil
+  end
 
-    # For some reason logstash / elasticsearch drops events where the payload
-    # is a hash. These are more conveniently accessed at the top level of the
-    # event, anyway, so we move it there.
+  def format_exception
+    exception_message = hash.dig(:exception, :message)
+    if exception_message.present?
+      hash[:message] = "Exception occured: #{exception_message}"
+    end
+  end
+
+  def format_backtrace
+    if hash[:message]&.start_with?("/")
+      message_lines = hash[:message].split("\n")
+      if message_lines.all? { |line| line.start_with?("/") }
+        hash[:backtrace] = hash[:message]
+        hash[:message] = "Exception occured: #{message_lines.first}"
+      end
+    end
+  end
+
+  # The value here appears to break logging to logstash / elasticsearch
+  def format_duration
+    hash[:duration] = hash[:duration_ms]
+    hash[:duration_ms] = nil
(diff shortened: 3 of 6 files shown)


app/jobs/bulk_sync_courses_to_find_job.rb
```diff
@@ -5,6 +5,9 @@ class SearchAndCompareRequestError < StandardError
   end
 
   def perform
+    RequestStore.store[:job_id] = provider_job_id
```
TobyRet: I get that RequestStore is storing the `job_id` and `queue_name`. I'm guessing that the `customize_event` you defined in `production.rb` passes on these values. Is the event only sent if values exist?

faissalMT: No, if they don't exist they'll just be nil.

misaka: Kibana has a way to filter events based on the non-existence of a field. Sounds like it might be better to _not_ include this if there is no data for it (in `production.rb`).

config/environments/production.rb
```diff
@@ -1,3 +1,11 @@
+def logstash_settings
+  logstash_custom_event = ->(event) do
+    event["job_id"] = RequestStore.store[:job_id]
+    event["job_queue"] = RequestStore.store[:job_queue]
+  end
+  Settings.logstash.to_h.merge(customize_event: logstash_custom_event)
+end
```
misaka: Would we not want this data if `qa` or `staging`?

misaka:
[[1d] Tag background jobs correctly for logit.](https://trello.com/c/5eW3TNWt/3087-1d-tag-background-jobs-correctly-for-logit)

misaka:
@faissalMT did you have a look at how Semantic Logger manages to do one-line logging with Rails, and whether there's a way to do it with Sidekiq?

faissalMT:
If Azure isn't just using the `docker-compose` then it will need to have `RAILS_APPLICATION_NAME` set for each container. For the `docker-compose` I've added the relevant values.

@misaka I had a brief look around but I couldn't find anything about 'one line logging' or similar terms

misaka:
> Rails
>
> The rails_semantic_logger gem replaces the default Rails logger with Semantic Logger. It also reduces Rails logging output in production to almost a single line for every Controller-Action call.

https://rocketjob.github.io/semantic_logger/rails.html

gpeng:
> @gpeng If Azure isn't just using the `docker-compose` then it will need to have `RAILS_APPLICATION_NAME` set for each container. For the `docker-compose` I've added the relevant values.

I don't think it does but I can't see where the separate containers for jobs run in Azure. Maybe @rizzkhan1 / @saliceti can advise?

config/initializers/semantic_logger.rb
```diff
@@ -14,5 +14,5 @@
   end
 
   log_stash = LogStashLogger.new(Settings.logstash.to_h.merge(customize_event: logstash_formatter))
-  SemanticLogger.add_appender(logger: log_stash, level: :info, formatter: :json)
+  SemanticLogger.add_appender(logger: log_stash, level: :info, formatter: :one_line)
```
misaka: I'm not sure this is the change we want, have you tested this? You can test by setting the appropriate env vars to log to Logit. But ... I expect that we need `:json` here, to interface with logstash.

What I'm trying to say, and I'm sorry I haven't been clear, is when a request comes into Rails, `semantic-logger` takes all the log events that normally get emitted, and turn them into a single event with all the information. To borrow from the [Semantic Logger Rails page](https://rocketjob.github.io/semantic_logger/rails.html) I mentioned above, this means that this:

Normal Rails Logs

Gets turned into this:

Semantic Logger Rails Logs

And likewise, instead of sending a bunch of events to Logit, it just sends one. What I wanted us to spend a little bit of times on is to see if we can replicate this for Sidekiq so that we send one log event with just the info we need to Logit for each job that gets processed. It's a little pie-in-the-sky, but I figured that with the prior-art in Semantic Logger we just might be able to figure this out.

rizzkhan1:
> > @gpeng If Azure isn't just using the `docker-compose` then it will need to have `RAILS_APPLICATION_NAME` set for each container. For the `docker-compose` I've added the relevant values.
> 
> I don't think it does but I can't see where the separate containers for jobs run in Azure. Maybe @rizzkhan1 / @saliceti can advise?

I think env value `RAILS_APPLICATION_NAME` set for each container in docker-compose file method should work fine. Alternatively we could set that as an environmental variable on the infrastructure - latter would require changes in azure\template.json file

config/initializers/semantic_logger.rb
```diff
@@ -1,18 +1,65 @@
...
-    event["duration"] = event["duration_ms"]
-    event["duration_ms"] = nil
+class LogStashFormatter < SemanticLogger::Formatters::Raw
+  def format_json_message_context
+    if hash[:message].present?
+      context = JSON.parse(hash[:message])["context"]
+      hash[:sidekiq_job_context] = hash[:message]
+      hash[:message] = context
+    end
+  rescue JSON::ParserError
```
misaka: Is this the right thing to do? I _think_ this executes in a separate thread and if this exception is raised it won't actually interfere with Sidekiq.

faissalMT: The intent here is to continue as normal if the data in the message isn't valid JSON, rather than deciding based on whether or not it begins with a given string.

misaka: My concern is just that we'll swallow up this error and never see it. OTOH, if we raise an error in this thread, I expect it'll log something to stdout and keep trying to send stuff to logstash. Will approve anyway as this is mostly conjecture.

misaka:
Do we need to add the setting `SETTINGS__APPLICATION` to the pipeline?

config/initializers/semantic_logger.rb
faissalMT: No the error will still be sent to logstash, it just won't be nicely formatted

PR #1238, opened 2020-03-11
3076 m send notification when a course has been created

SamJamCul:
### Context

Please see this card for full details https://trello.com/c/KbHLq4Pq/3076-m-send-notification-when-a-course-has-been-created

This pr creates a new mailer for creating a course - if a user has opted in for notifications, they'll get an email via Notify letting them know a course that they accredit has been created. 

You can find the Notify template here, you'll need an account https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534/templates/c4944115-6e73-4b30-9bc2-bf784c0e9aaa

* Created a mailer for the new email
* Updated the user_notifications table to include a toggle for create course emails
* Added a hook to the course model to trigger the email on creation

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


SamJamCul:
[[M] Send notification when a course has been created](https://trello.com/c/KbHLq4Pq/3076-m-send-notification-when-a-course-has-been-created)

app/models/course.rb
```diff
@@ -267,6 +269,17 @@ def notify_accredited_body?
     (saved_changes.keys & update_notification_attributes).any?
   end
 
+  def send_create_notification_to_accredited_body
+    users = UserNotification.where(provider_code: accrediting_provider_code, create_course: true).map(&:user)
```
gpeng: We should make this a scope on `UserNotification` rather than coupling to the DB fields from here. e.g. `UserNotification.users_requesting_create_course_notification_for(provider: <code>)` or sumfink.

app/models/course.rb
```diff
@@ -256,6 +256,8 @@ def self.entry_requirement_options_without_nil_choice
 
   after_update :send_notification_to_accredited_body, if: :notify_accredited_body?
 
+  after_create :send_create_notification_to_accredited_body, if: :notify_accredited_body?
```
gpeng: I don't think we should send these emails from callbacks here for a few reasons:

* the model should be concerned with persistence and data not email
* the effect of this is unexpected `Course.create` sends an email. Every time we want to create a course in tests we have to be aware of (and possibly stub) unintended consequences
* we've just added another 100 lines to the ~2.5k lines of spec for this class

I don't think this should block this PR but if everyone agrees I'll add another ticket to refactor this

spec/models/course_spec.rb
```diff
@@ -2135,6 +2135,119 @@
...
+    end
+
+    let(:mail_spy) { spy }
+    let(:mailer_spy) { spy(create_course_email: mail_spy) }
+
+    before do
+      stub_const("CreateCourseEmailMailer", mailer_spy)
+    end
+
+    context "A self-accredited course" do
```
gpeng: we generally do these all in lower case

spec/models/course_spec.rb
```diff
@@ -2135,6 +2135,119 @@
...
+          it "Does not send a notification" do
+            course
+
+            expect(mailer_spy).not_to have_received(:create_course_email)
+          end
+        end
+      end
+      context "With multiple users with notifications enabled" do
+        before do
+          UserNotification.new(user: user_one, provider_code: accrediting_provider.provider_code, create_course: true).save!
```
gpeng: We should create a factory for these. That said, if we add the scope we can stub it here rather than have to involve the database (as long as this is covered in a higher level integration test somewhere)

app/models/course.rb
```diff
@@ -268,9 +270,19 @@ def notify_accredited_body?
     (saved_changes.keys & update_notification_attributes).any?
   end
 
-  def send_notification_to_accredited_body
-    users = UserNotification.where(provider_code: accrediting_provider_code, course_update: true).map(&:user)
+  def send_create_notification_to_accredited_body
+    users = User.joins(:user_notifications).merge(UserNotification.course_create_notification_requests(accrediting_provider_code))
```
gpeng: This looks a bit hairy but it avoids the `n+1` of using `map`. We haven't used query objects in the app as yet but this might be a good candidate for one. I don't think we should do it now though, just a consideration for when we refactor this.

spec/models/course_spec.rb
```diff
@@ -2135,6 +2135,119 @@
...
+
+    before do
+      stub_const("CourseCreateEmailMailer", mailer_spy)
+    end
+
+    context "a self-accredited course" do
+      let(:course) { create(:course, :self_accredited, provider: accrediting_provider) }
+
+      before do
+        UserNotification.new(user: user_one, provider_code: accrediting_provider.provider_code, course_create: true).save!
```
gpeng: We can use the factory here now (and in the rest of this file)

SamJamCul: Whoops, my find and replace was incomplete. Done now!

app/controllers/api/v2/courses_controller.rb
```diff
@@ -75,6 +75,7 @@ def publish
           @course.publish_sites
           @course.publish_enrichment(@current_user)
           @course.reload
+          Courses::PublishCourseService.new.execute(course: @course)
```
gpeng: The second `Course` is redundant here. Could we use `Course::PublishService` or `CourseService::Publish`? Also we should decide a consistent pattern for this sort of service. I prefer using a static `.call` as in [app/services/course_search_service.rb](app/services/course_search_service.rb) as it makes the `.new` method private. We can move all of the boiler plate in that service into a module to share.

gpeng: Can we add a card to move all of the other things into this service too once this is sufficiently ready for user research?

spec/services/courses/publish_course_service_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+
+          context "when the course does not appear on find" do
+            let(:site_status) { create(:site_status, :unpublished) }
+
+            it "does not send a notification" do
+              expect(mailer_spy).not_to have_received(:course_create_email)
+            end
+          end
+        end
+        context "with multiple users with notifications enabled" do
```
gpeng: Missing a line break here

app/controllers/api/v2/courses_controller.rb
misaka: > Also we should decide a consistent pattern for this sort of service. I prefer using a static `.call`

:100:

spec/factories/user_notifications.rb
```diff
@@ -0,0 +1,20 @@
...
+#  provider_code :string           not null
+#  updated_at    :datetime         not null
+#  user_id       :integer          not null
+#
+# Indexes
+#
+#  index_user_notification_on_provider_code  (provider_code)
+#
+FactoryBot.define do
+  factory :user_notification do
```
defong: ain't this missing association/relationship for `user`/`provider`?

PR #1240, opened 2020-03-11
[3118] Ignore ungeocoded sites in distance search

asmega:
# Context

- https://trello.com/c/fYgK7vnQ/3118-sort-by-distance-only-uses-sites-that-have-been-geocoded
- We are not excluding un-geocoded sites from distance sorting
- These sites are then deemed be at the origin of the search query as a result

- When sorting by distance we exclude any site that has not been geocoded
- These are sites that do not have latitude and longitude present

### Guidance to review

- visit http://localhost:3002/results?fulltime=False&hasvacancies=True&l=1&lat=53.9588365&lng=-1.0794777&loc=York+YO1%2C+UK&lq=York+YO1%2C+UK&parttime=False&qualifications=QtsOnly%2CPgdePgceWithQts%2COther&rad=20&senCourses=False&sortby=2&subjects=31%2C32
- Results should be ordered correctly

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -52,16 +52,19 @@ private
     # only want new and running sites
     new_and_running_sites = sites_with_status.where(site_status[:status].in(%w[new_status running]))
 
+    # we only want sites that have been geocoded
+    geocoded_new_and_running_sites = new_and_running_sites.where(sites[:latitude].not_eq(nil).and(sites[:longitude].not_eq(nil)))
+
     # only sites that have a locatable address
     # there are some sites with no address1 or postcode that cannot be
     # accurately geocoded. We don't want to return these as the closest site.
     # This should be removed once the data is fixed
-    new_and_running_sites = new_and_running_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
+    locatable_new_and_running_sites = geocoded_new_and_running_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
 
     # select course_id and nearest site with shortest distance from origin
     # as courses may have multiple sites
     # this will remove duplicates by aggregating on course_id
-    courses_with_nearest_site = new_and_running_sites.project(:course_id, Arel.sql("MIN#{Site.distance_sql(OpenStruct.new(lat: origin[0], lng: origin[1]))} as distance")).group(:course_id)
+    courses_with_nearest_site = locatable_new_and_running_sites.project(:course_id, Arel.sql("MIN#{Site.distance_sql(OpenStruct.new(lat: origin[0], lng: origin[1]))} as distance")).group(:course_id)
 
     # form a temporary table with results
     distance_table = Arel::Nodes::TableAlias.new(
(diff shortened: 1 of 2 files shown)


app/services/course_search_service.rb
```diff
@@ -52,6 +52,9 @@ def sites_with_distance_from_origin
     # only want new and running sites
     new_and_running_sites = sites_with_status.where(site_status[:status].in(%w[new_status running]))
 
+    # we only want sites that have been geocoded
+    new_and_running_sites = new_and_running_sites.where(sites[:latitude].not_eq(nil).and(sites[:longitude].not_eq(nil)))
```
timabell: The meaning of this variable name doesn't seem right with the additional `where`.

spec/requests/api/v3/courses/index_spec.rb
```diff
@@ -227,6 +227,25 @@
           expect(course_hashes.second["id"]).to eq(course_a.id.to_s)
         end
       end
+
+      context "when a course has a site that has not been geocoded" do
```
timabell: Can we push this test further down? I think a test that tests the service + db would be sufficient and faster.

timabell:
Use the github attach button in trello to get a proper two-way link between the card and PR.

timabell:
Good work, test + fix. Just some code structure comments from me.

spec/requests/api/v3/courses/index_spec.rb
asmega: I can do that next time. It would only be slightly faster as serialisation is not performed. I don't feel pushing this particular working test down at this moment in time is worthwhile.

app/services/course_search_service.rb
asmega: I have now renamed some local variables

timabell:
[Sort by distance only uses sites that have been geocoded](https://trello.com/c/fYgK7vnQ/3118-sort-by-distance-only-uses-sites-that-have-been-geocoded)

app/services/course_search_service.rb
```diff
@@ -52,16 +52,19 @@ def sites_with_distance_from_origin
...
 
+    # we only want sites that have been geocoded
+    geocoded_new_and_running_sites = new_and_running_sites.where(sites[:latitude].not_eq(nil).and(sites[:longitude].not_eq(nil)))
+
     # only sites that have a locatable address
     # there are some sites with no address1 or postcode that cannot be
     # accurately geocoded. We don't want to return these as the closest site.
     # This should be removed once the data is fixed
-    new_and_running_sites = new_and_running_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
+    locatable_new_and_running_sites = geocoded_new_and_running_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
```
timabell: nice one, much clearer

spec/requests/api/v3/courses/index_spec.rb
timabell: This :point_right: https://martinfowler.com/articles/practical-test-pyramid.html#TheTestPyramid

But I won't hold up the PR if you don't want to do it.

PR #1243, opened 2020-03-12
[3120] - Accept non-alphanumeric provider suggestion queries

TobyRet:
### Context
Please see https://github.com/DFE-Digital/find-teacher-training/pull/197 for further context. This PR should be merged first. 

The ProviderSuggestions controller should process encoded queries and not return a Bad request status (400) by default. If a provider search still returns no results then the API will return an empty array and 200 status.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/provider_suggestions_controller.rb
+++ b/app/controllers/api/v3/provider_suggestions_controller.rb
@@ -5,7 +5,6 @@ module API
 
       def index
         return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
-        return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
         found_providers = @recruitment_cycle.providers
                               .with_findable_courses
--- a/spec/requests/api/v3/providers/suggest_provider_spec.rb
+++ b/spec/requests/api/v3/providers/suggest_provider_spec.rb
@@ -79,13 +79,13 @@ describe "GET /provider-suggestions" do
     expect(JSON.parse(response.body)["data"].length).to eq(10)
   end
 
-  it "returns bad request if query is empty" do
+  it "returns status: bad request (400) if query is empty" do
     get "/api/v3/provider-suggestions"
 
     expect(response.status).to eq(400)
   end
 
-  it "returns bad request if query is too short" do
+  it "returns status: bad request (400) if query is too short" do
     provider
 
     get "/api/v3/provider-suggestions?query=#{provider.provider_name[0, 2]}"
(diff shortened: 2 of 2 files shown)


spec/requests/api/v3/providers/suggest_provider_spec.rb
```diff
@@ -79,23 +79,17 @@
...
     get "/api/v3/provider-suggestions?query=#{provider.provider_name[0, 2]}"
 
     expect(response.status).to eq(400)
   end
-
-  it "returns bad request if start of query is not alphanumeric" do
-    get "/api/v3/provider-suggestions?query=%22%22%22%22"
-
-    expect(response.status).to eq(400)
-  end
```
timabell: can you invert this test instead of deleting it?

"should not return bad request if start of query is not alphanumeric"

app/controllers/api/v3/provider_suggestions_controller.rb
```diff
@@ -5,7 +5,6 @@ class ProviderSuggestionsController < API::V3::ApplicationController
 
       def index
         return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
-        return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
```
timabell: As per my other comment there's no test showing the new behaviour. We need a test to prevent future regressions.

TobyRet:
[Provider-suggestions query throws `URI::InvalidURIError`](https://trello.com/c/Ou8oRtwC/3120-provider-suggestions-query-throws-uriinvalidurierror)

spec/requests/api/v3/providers/suggest_provider_spec.rb
TobyRet: I don't see the point in testing default behaviour?

app/controllers/api/v3/provider_suggestions_controller.rb
TobyRet: There is no 'new' behaviour being added. Rather I am removing behaviour that was added in error. We are just reverting back to the default behaviour, hence why no test required in my opinion. What do other's think?

timabell: Personally I'd add a test to ensure that we won't get a regression in future for *any* reason including rails and library (gem) upgrades. The test would be living documentation that we expect this behaviour to work. I can see that it's debatable, but that's what I'd do.

spec/requests/api/v3/providers/suggest_provider_spec.rb
timabell: see below

PR #1245, opened 2020-03-12
[3116] Add previous value and current value to change email notification

gpeng:
### Context

We want to send an email notification when a course is updated. The email will contain the name of the attribute that has been updated, the original value and the updated value. This PR adds the original and updated value.

Note: The API currently only supports one attribute being updated per update. This has been implemented because the Publish app only supports updating one attribute at a time (I think). The API does support updating more than one attribute at a time so we shouldn't assume that a caller won't do that. I've not addressed that here. I've added a card to refactor this code too to move the notification and email concerns out of the Course model. [Trello](https://trello.com/c/vItr7tVZ/340-move-notification-sending-out-of-course-callbacks)

* Update the mailer to receive original and updated field values
* Update `Course` to send the values
* Move formatting of the attribute name into the mailer as it's really a 'view' concern. The mailer is at least a bit closer to the view. We can't do it 'in' the view as that's in Notify.
* Add `CourseAttributeFormatterService` to handle the formatting of the values (they are `under_scored_values` in the DB

### Guidance to review

The template for this is in Notify and it hasn't been updated as yet so it's a bit tricky to review the functionality.

### Implemented formatting

Age: `10_to_14` -> "10 to 14"
Study mode: `part_time` -> "part time"
Qualification: `pgce_with_qts` -> "PGCE with QTS"
Entry requirements: `not_required` -> "not required"

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/mailers/course_update_email_mailer.rb
+++ b/app/mailers/course_update_email_mailer.rb
@@ -16,8 +16,8 @@ class CourseUpdateEmailMailer < GovukNotifyRails::Mailer
       attribute_changed: I18n.t("course.update_email.#{attribute_name}"),
       attribute_change_datetime: format_update_datetime(course.updated_at),
       course_url: create_course_url(course),
-      original_value: original_value,
-      updated_value: updated_value,
+      original_value: formatter.call(name: attribute_name, value: original_value),
+      updated_value: formatter.call(name: attribute_name, value: updated_value),
     )
 
     mail(to: recipient.email)
@@ -33,5 +33,9 @@ private
 
   def format_update_datetime(datetime)
     datetime.strftime("%-k:%M%P on %-e %B %Y")
+  end
+
+  def formatter
+    CourseAttributeFormatterService
   end
 end
--- a/app/mailers/course_update_email_mailer.rb
+++ b/app/mailers/course_update_email_mailer.rb
@@ -13,7 +13,7 @@ class CourseUpdateEmailMailer < GovukNotifyRails::Mailer
       provider_name: course.provider.provider_name,
       course_name: course.name,
       course_code: course.course_code,
-      attribute_changed: attribute_name,
+      attribute_changed: I18n.t("course.update_email.#{attribute_name}"),
       attribute_change_datetime: format_update_datetime(course.updated_at),
       course_url: create_course_url(course),
       original_value: original_value,
--- a/app/mailers/course_update_email_mailer.rb
+++ b/app/mailers/course_update_email_mailer.rb
@@ -1,17 +1,26 @@
 class CourseUpdateEmailMailer < GovukNotifyRails::Mailer
-  def course_update_email(course, attribute_changed, user)
+  def course_update_email(
+    course:,
+    attribute_name:,
+    original_value:,
+    updated_value:,
+    recipient:
+  )
+
     set_template(Settings.govuk_notify.course_update_email_template_id)
 
     set_personalisation(
       provider_name: course.provider.provider_name,
       course_name: course.name,
       course_code: course.course_code,
-      attribute_changed: attribute_changed,
+      attribute_changed: attribute_name,
       attribute_change_datetime: format_update_datetime(course.updated_at),
       course_url: create_course_url(course),
+      original_value: original_value,
+      updated_value: updated_value,
     )
 
-    mail(to: user.email)
+    mail(to: recipient.email)
   end
 
 private
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -275,9 +275,11 @@ class Course < ApplicationRecord
 
     users.each do |user|
       CourseUpdateEmailMailer.course_update_email(
-        self,
-        I18n.t("course.update_email.#{updated_attribute}"),
-        user,
+        course: self,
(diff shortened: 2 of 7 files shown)


gpeng:
[[M] Add [previous value] and [current value] to change email notification](https://trello.com/c/Wd8c512b/3116-m-add-previous-value-and-current-value-to-change-email-notification)

app/models/course.rb
```diff
@@ -276,7 +276,9 @@ def send_notification_to_accredited_body
     users.each do |user|
       CourseUpdateEmailMailer.course_update_email(
         self,
-        I18n.t("course.update_email.#{updated_attribute}"),
+        updated_attribute,
+        saved_changes[updated_attribute].first,
+        saved_changes[updated_attribute].second,
         user,
```
TobyRet: I'm a bit confused by names of the args being passed into `course_update_email` and how these map to the method signature - `saved_changes[updated_attribute].first`, `saved_changes[updated_attribute].second`

app/mailers/course_update_email_mailer.rb
```diff
@@ -1,14 +1,16 @@
...
 
     set_personalisation(
       provider_name: course.provider.provider_name,
       course_name: course.name,
       course_code: course.course_code,
-      attribute_changed: attribute_changed,
+      attribute_changed: I18n.t("course.update_email.#{updated_attribute_name}"),
       attribute_change_datetime: format_update_datetime(course.updated_at),
       course_url: create_course_url(course),
+      original_value: formatter.call(name: updated_attribute_name, value: original_value),
```
TobyRet: I'm not massively familiar with Govuk Notify - am I right in thinking there is a template somewhere that needs to be updated to use`original_value` ?

fofr:
How does an example email look with this?

app/mailers/course_update_email_mailer.rb
aliuk2012: `set_personalisation` sets up the values for the placeholders in govuk notify template. I think what makes it hard to visualize is that the template is not stored in this repo but on govuk notify service itself.

app/models/course.rb
aliuk2012: I agree, would it maybe make it easier to read if `#course_update_email` had named params. It also has the benefit of not having to know which order the params need to be

app/mailers/course_update_email_mailer.rb
gpeng: Yes, there's a template. I haven't updated it as yet. It currently looks like this

Screenshot 2020-03-13 at 17 41 27

app/models/course.rb
gpeng: I've updated it to use keyword arguments so hopefully that should be a bit clearer.

gpeng:
> How does an example email look with this?

I haven't updated the template as yet. The intention is to update the second paragraph with:

```
((provider_name)) changed the ((attribute_changed)) of ((course_name)) (((course_code))) from ((original_value)) to ((updated_value)). The change was made at ((attribute_change_datetime)).
```

benilovj:
You might want to consider moving the email contents into the app and just have a layout in Notify, Apply does this and it allows previewing the emails in-app.

aliuk2012:
> You might want to consider moving the email contents into the app and just have a layout in Notify, Apply does this and it allows previewing the emails in-app.

@benilovj, thanks for sharing that interesting approach and something I've not thought about before but I do think that introduces a few negatives.

1. More data being sent over the wire to Notify for each email sent out instead of just placeholder values.
2. By having all the copy in rails this will block non-developers on the team from being able to make quick simple copy changes without involving a developer.

app/services/course_attribute_formatter_service.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  class << self
+    def call(**args)
+      new(args).call
+    end
+  end
+
+  def call
+    return value.tr("_", " ") if age_range?
```
aliuk2012: Could we not call `strip_underscores` instead?

gpeng: We could even call `age_range_value` which is what I meant to do 🤦‍♂ Bear with...

benilovj:
> > You might want to consider moving the email contents into the app and just have a layout in Notify, Apply does this and it allows previewing the emails in-app.
> 
> @benilovj, thanks for sharing that interesting approach and something I've not thought about before but I do think that introduces a few negatives.
> 
> 1. More data being sent over the wire to Notify for each email sent out instead of just placeholder values.

Not sure how big an issue this.

> 2. By having all the copy in rails this will block non-developers on the team from being able to make quick simple copy changes without involving a developer.

I understand this concern, but we haven't really seen this in practice. All 6 of our content and interaction designers are running the app locally and making changes through pull requests on GitHub.

> 3. You cant preview it like you can in the UI without actually sending an email.

This is how Apply's Rails mailer previews look (these are actually available not just locally, but on the QA environment as well):

image

and here's how it looks for real:

image

which looks pretty similar to me.

gpeng:
This would also mean we could preview template changes without changing the production template, or maintaining two templates in Notify with the complications and risk that brings. I think that's worth considering as we add more notifications.

PR #1255, opened 2020-03-17
[3102] - Background mail notifications

TobyRet:
### Context
* We have a new Sidekiq background container running, similar to geocoding.

* Mail is sent out via a background task.
* When you call `.deliver_later` on the mailers it will [automatically use an ActionMailer::DeliveryJob.](https://edgeapi.rubyonrails.org/classes/ActionMailer/MessageDelivery.html#method-i-deliver_later)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/send_welcome_email_service.rb
+++ b/app/services/send_welcome_email_service.rb
@@ -1,12 +1,16 @@
 class SendWelcomeEmailService
-  def initialize(mailer:)
-    @mailer = mailer
+  class << self
+    def call(current_user:)
+      new.call(current_user: current_user)
+    end
   end
 
-  def execute(current_user:)
+  def call(current_user:)
     return if current_user.welcome_email_date_utc
 
-    @mailer.send_welcome_email(first_name: current_user.first_name, email: current_user.email).deliver_now
+    WelcomeEmailMailer.
+      send_welcome_email(first_name: current_user.first_name, email: current_user.email).
+      deliver_now
 
     current_user.update(
       welcome_email_date_utc: Time.now.utc,
--- a/app/services/send_welcome_email_service.rb
+++ b/app/services/send_welcome_email_service.rb
@@ -10,7 +10,7 @@ class SendWelcomeEmailService
 
     WelcomeEmailMailer.
       send_welcome_email(first_name: current_user.first_name, email: current_user.email).
-      deliver_now
+      deliver_later(queue: "mailer")
 
     current_user.update(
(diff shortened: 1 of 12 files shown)


TobyRet:
[[M] Background mail notifications](https://trello.com/c/jb4y3epg/3102-m-background-mail-notifications)

app/services/send_welcome_email_service.rb
```diff
@@ -1,12 +1,16 @@
...
   end
 
-  def execute(current_user:)
+  def call(current_user:)
     return if current_user.welcome_email_date_utc
 
-    @mailer.send_welcome_email(first_name: current_user.first_name, email: current_user.email).deliver_now
+    WelcomeEmailMailer.
+      send_welcome_email(first_name: current_user.first_name, email: current_user.email).
+      deliver_later(queue: "mailer")
```
faissalMT: This might seem like a nitpick but please put the dot operator at the beginning of your lines. It might be easy to misread these as being calls to local methods.

faissalMT: This might seem like a nitpick but please put the dot operator at the beginning of your lines. It might be easy to misread these as being calls to local methods.

TobyRet: I've always put the dot at the end - https://github.com/rubocop-hq/ruby-style-guide#multi-line-method-chains

Happy to change if that's the team convention.

What do other's think?

dankmitchell: I prefer a leading dot. I'm surprised rubocop doesn't have an rule 🤔

TobyRet: No prob - will update. Perhaps there is a way to get Rubocop to check for this ...

TobyRet: Looks like govuk rubocop have disabled the dot check - https://github.com/alphagov/rubocop-govuk/search?q=dot&unscoped_q=dot

TobyRet: @dankmitchell @faissalMT - given the above can I suggest that this PR is merged then we add the Rubocop dot check (and any fixes) in a subsequent PR? I wonder if we need a team vote on this also. Personally I don't care if we do a leading or trailing dot but I think this may be a contentious issue for others ... Maybe that's why GovUK disabled the check in the first place

PR #1269, opened 2020-03-23
[3072] - GeocodingService bug fix

TobyRet:
### Context
Some legacy sites and providers have missing address details, which are considered 'invalid'. When we batch geocode we override the validation checks and attempt to geocode their address with what little information we have. 

A problem arises when we attempt to geocode Sites and Providers with no address details whatsoever (i.e. `address1`, `address2`, `address3`, `address4`, `postcode` are all blank).

Batch geocoding is done in a single process. On Publish, when a site or provider is updated, this is done in a separate process via a back ground job. 

_Currently_, where a Site or Provider **has no address** whatsoever:

1. The mcb geocode command geocodes the object via the GeocoderService.
2. The Geocoder API returns an unsuccessful response. The GeocoderService still attempts to update the record regardless.
3. Upon save, an after db commit hook is called to checked if the object requires geocoding. At this point because the object's full address is considered to be present (ie `Site#full_address` returns `", , , ,"`) and both lat and long is `nil`, the geocoding job request is created.  

Introduce new guard clauses so that geocoding only happens when it's supposed to.

1. Return `nil` for `#full_address` if a Site / Provider has no address whatsoever. 
2. Only call the Google Geocoding API if the object has a `full_address`
3. Only update the object of the Geocoding API returns a success response.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -152,7 +152,11 @@ class Provider < ApplicationRecord
   end
 
   def full_address
-    [provider_name, address1, address2, address3, address4, postcode].compact.join(", ")
+    address = [provider_name, address1, address2, address3, address4, postcode]
+
+    return "" if address.all?(&:blank?)
+
+    address.compact.join(", ")
   end
 
   def address_changed?
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -69,6 +69,8 @@ class Site < ApplicationRecord
       address.unshift(location_name)
     end
 
+    return "" if address.all?(&:blank?)
+
     address.compact.join(", ")
   end
 
--- a/app/services/geocoder_service.rb
+++ b/app/services/geocoder_service.rb
@@ -1,8 +1,10 @@
 class GeocoderService
   def self.geocode(obj:, force: false)
+    return if obj.full_address.blank?
+
     result = Geokit::Geocoders::GoogleGeocoder.geocode(obj.full_address, bias: "gb")
 
-    if result.present?
+    if result&.success == true
       obj.latitude = result.latitude
       obj.longitude = result.longitude
 
--- a/spec/services/geocoder_service_spec.rb
+++ b/spec/services/geocoder_service_spec.rb
@@ -1,6 +1,15 @@
 describe GeocoderService do
   describe "#geocode" do
-    let(:valid_site) { create(:site, region_code: nil) }
+    let(:valid_site) {
+      create(:site,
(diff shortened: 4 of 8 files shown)


app/models/provider.rb
```diff
@@ -152,7 +152,11 @@ def needs_geolocation?
   end
 
   def full_address
-    [provider_name, address1, address2, address3, address4, postcode].compact.join(", ")
+    address = [provider_name, address1, address2, address3, address4, postcode]
+
+    return nil if address.all?(&:blank?)
```
TobyRet: NullObject pattern candidate?

app/models/site.rb
```diff
@@ -69,6 +69,8 @@ def full_address
       address.unshift(location_name)
     end
 
+    return nil if address.all?(&:blank?)
```
TobyRet: NullObject pattern candidate?

TobyRet:
[Sort out geocoding / places Google API blockage](https://trello.com/c/9eH56bLn/3072-sort-out-geocoding-places-google-api-blockage)

spec/services/geocoder_service_spec.rb
```diff
@@ -1,36 +1,69 @@
...
-    it "does not geocode an invalid object by default" do
-      expect { GeocoderService.geocode(obj: invalid_site) }.
-          to raise_error(ActiveRecord::RecordInvalid)
-    end
+    context "a valid object" do
+      it "geocodes a valid object" do
+        expect { GeocoderService.geocode(obj: valid_site) }.
+          to change { valid_site.reload.latitude }.from(nil).to(50.8312522).
+            and change { valid_site.longitude }.from(nil).to(-1.3792036).
+              and change { valid_site.region_code }.from(nil).to("south_east")
```
gpeng: Indenting is a bit messed up here

app/models/provider.rb
gpeng: I think because we would be returning a string if there was an address the null object would sort of be an empty string. The caller would then check `full_address.blank?` for null-ish-ness.

app/services/geocoder_service.rb
```diff
@@ -1,8 +1,10 @@
 class GeocoderService
   def self.geocode(obj:, force: false)
+    return unless obj.full_address
```
gpeng: This might be more explicity as `return if obj.full_address.blank?`

PR #1272, opened 2020-03-24
[3161] govuk tech docs

asmega:
### Context

- This adds govuk tech docs which is able to generate a static site to contain documentation
- It is able to consume the generated OpenAPI spec to generate more humanly digestible documentation

- Previous generated documentation has now been removed from the rails app
- This adds gov uk tech docs https://github.com/alphagov/tech-docs-gem
- Tech docs generates a static site to /public via docker build process
- Updated README on how to generate static docs
- Tech docs consumes OpenAPI spec to generate docs
- We are temporarily using branch version of Tech docs to render `anyOf` references in the OpenAPI spec
- Fixed intermittent failing test by replacing with fuzzy assertion

### Guidance to review

- visit http://localhost:3001/ should see documentation that is correct
- should be able to update and generate new docs based on info added to README 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


asmega:
[Improve the OpenAPI Spec Docs](https://trello.com/c/J9ihJV1R/3161-improve-the-openapi-spec-docs)

docs/Gemfile
```diff
@@ -0,0 +1,12 @@
+# If you do not have OpenSSL installed, change
+# the following line to use "http://"
+source "https://rubygems.org"
+
+# For faster file watcher updates on Windows:
+gem "wdm", "~> 0.1.0", platforms: %s{mswin mingw}
+
+# Windows does not come with time zone data
+gem "tzinfo-data", platforms: %i{mswin mingw jruby}
```
misaka: Do we need to support Windows?

asmega: These are the defaults from gov uk tech docs. happy to remove if we don't need to support Windows

asmega: I have now removed these

config/routes.rb
```diff
@@ -113,26 +105,118 @@
 #
 # Routes for OpenApi::Rswag::Api::Engine:
 
+#                                                                 Prefix Verb   URI Pattern                                                                                                                              Controller#Action
```
misaka: The routes are being duplicated here for some reason.

asmega: I have regenerated. Looks like I messed up a rebase and didn't notice the duplicates

docs/config/tech-docs.yml
```diff
@@ -0,0 +1,38 @@
+# Host to use for canonical URL generation (without trailing slash)
+host: docs.teacher-training-courses.service.gov.uk
```
misaka: Should we use `api2.publish-teacher-training-courses.service.gov.uk` until we have a new domain setup?

asmega: done

public/index.html
```diff
@@ -0,0 +1,322 @@
+<!doctype html>
```
misaka: Is it worth having these generated static files here? It could be handy, but then it also risks having old / stale files in here. Might be better to be forced to generate these as part of the build pipeline.

spec/requests/api/v3/providers/courses/show_spec.rb
```diff
@@ -97,6 +97,9 @@
       changed_at = json_response["data"]["attributes"].delete("changed_at")
       expect(Time.zone.parse(changed_at)).to be_within(10).of(Time.zone.now)
 
+      last_published_at = json_response["data"]["attributes"].delete("last_published_at")
+      expect(Time.zone.parse(last_published_at)).to be_within(10).of(enrichment.last_published_timestamp_utc)
```
misaka: Does this belong here?

misaka: Ah, I see this has it's own commit. 👍

docs/Gemfile
```diff
@@ -0,0 +1,6 @@
+# If you do not have OpenSSL installed, change
+# the following line to use "http://"
+source "https://rubygems.org"
+
+# Include the tech docs gem
+gem "govuk_tech_docs", git: "https://github.com/asmega/tech-docs-gem", branch: "any-of-api-renderer"
```
aliuk2012: could we not move this fork to dfe repo? Otherwise, if @asmega moves on or decides to later remove this repo it will break our docs.

public/index.html
asmega: I can try making it part of the docker image build process

docs/Gemfile
asmega: Agreed. I have made the fork here: https://github.com/DFE-Digital/tech-docs-template I will add my branch there and point to it

asmega: sorry wrong repo should be https://github.com/DFE-Digital/tech-docs-gem

public/index.html
asmega: I have deleted the generated files from `/public` these are now added as part of the docker build process

docs/config/tech-docs.yml
```diff
@@ -0,0 +1,38 @@
...
+# headings.
+max_toc_heading_level: 6
+
+# Prevent robots from indexing (e.g. whilst in development)
+prevent_indexing: false
+
+show_contribution_banner: true
+github_repo: DFE-Digital/teacher-training-api
+
+api_path: https://api2.publish-teacher-training-courses.service.gov.uk/api-docs/v1/swagger.json
```
misaka: This is grabbing the API spec from the running server, we should probably use `../swagger/v1/swagger.json`.

docs/config/tech-docs.yml
```diff
@@ -0,0 +1,38 @@
+# Host to use for canonical URL generation (without trailing slash)
+host: api2.publish-teacher-training-courses.service.gov.uk
+
+# Header-related options
+show_govuk_logo: true
+service_name: Teacher Training Courses
```
misaka:
```suggestion
service_name: Find Teacher Training
```

I tried using "Teacher Training Courses" in a previous PR, but people had strong feelings it should follow the existing service name.

misaka: (I do think that this is potentially it's own service, but maybe we should tackle that later)

docs/config/tech-docs.yml
misaka: Incidentally, we should make more effort to make it clear that this isn't the existing `v1`, and I'm not sure that the filename `swagger` is really descriptive. How about we use the directory `swagger/public-v1/api_spec.json` or something like that?

docs/source/release-notes.html.md.erb
```diff
@@ -0,0 +1,10 @@
+---
+title: Release notes
+weight: 2
+---
+
+# Release Notes
+
+## v1.0 - 13 March 2020
```
misaka: How can we make this clear that this is the public v1? Or do we care since people shouldn't see the private one ever?

README.md
```diff
@@ -181,6 +181,12 @@ Use the following command to generate OpenAPI specification:
 bundle exec rake rswag:specs:swaggerize
 ```
 
+We use [Tech Docs](https://github.com/alphagov/tech-docs-gem) to build documentation. To update documentation, the relevant files can be bound in `/docs`. The following command will then build the static site
+
+```sh
+cd docs && bundle install && bundle exec middleman build --build-dir=../public
```
misaka: I have a feeling we should add the generated files to `.gitignore`, otherwise they'll end up getting checked in at some point.

docs/source/release-notes.html.md.erb
asmega: I don't think it's needed if people never see the difference. We can add that distinction if we need to document the private api here

docs/config/tech-docs.yml
asmega: ok, i've made the suggested changes

README.md
asmega: we won't need to do this any more as it's done by the docker build. I will update the docs so people use middleman directly

asmega: docs have now been updated so we shouldn't need to do the gitignore

PR #1280, opened 2020-03-26
[3156] Refactor notifications

faissalMT:
### Context
Course notifications were implemented in the course model, which already handles far too much.

Move the course notifications into a service and refactor getting notifiable users into a scope.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/controllers/api/v2/courses_controller.rb
```diff
@@ -143,6 +144,8 @@ def update_course
 
         @course.assign_attributes(course_params)
         @course.save
+
+        Courses::UpdateNotificationService.new.call(course: @course)
```
asmega: Would this line be better after line 98 instead?

faissalMT: Since the service needs to be called after a given save I'd like to keep it near the actual call to `.save()`

asmega: sure, i guess out of scope of this PR this controller action could do with a tidy up

PR #1287, opened 2020-03-30
[2893] Add Location to API docs

asmega:
### Context

- https://trello.com/c/fdFc3kNR/2893-create-openapi-spec-for-locations
- Add Location to API documentation as it is does not exists

- Add documentation

### Guidance to review

- run docs, see https://github.com/DFE-Digital/teacher-training-api#documentation
- visit http://localhost:3001
- see added Location documentation, compared to https://api2.qa.publish-teacher-training-courses.service.gov.uk/#about

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/swagger/public_v1/template.yml
+++ b/swagger/public_v1/template.yml
@@ -372,6 +372,64 @@ components:
         version:
           type: string
           example: "1.0"
+    LocationAttributes:
+      type: object
+      properties:
+        code:
+          type: string
+          example: "W"
+        name:
+          type: string
+          description: "The name of the location"
+          example: "Henry Maynard Primary School"
+        street_address_1:
+          type: string
+          description: "Building or street line one"
+          example: "The Rose Building"
+        street_address_2:
+          type: string
+          description: "Building or street line two"
+          example: "Maynard Road"
+        city:
+          type: string
+          description: "Town or city"
+          example: "London"
+        county:
+          type: string
+          description: "County"
+          example: "London"
+        postcode:
+          type: string
+          description: "The postcode of the location"
+          example: "E17 9JE"
+        region_code:
+          type: string
+          example: "london"
+        recruitment_cycle_year:
+          type: string
+          example: "2020"
+    LocationResource:
+      type: object
+      properties:
+        id:
+          type: string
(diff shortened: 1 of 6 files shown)


asmega:
[Create OpenAPI spec for locations](https://trello.com/c/fdFc3kNR/2893-create-openapi-spec-for-locations)

swagger/public_v1/template.yml
```diff
@@ -372,6 +372,47 @@ components:
...
+          example: "Maynard Road"
+        address2:
+          type: string
+          example: ""
+        address3:
+          type: string
+          example: ""
+        address4:
+          type: string
+          example: "London"
```
misaka: For all these address properties, we use the wording "Building and street" for `address1` and `address2`, and then "Town or city" for `address3` and "County" for for `address4`. In the provider attributes I've tried to use language that's closer to that ... I think we should do that here.

See [the provider attributes in my PR](https://github.com/DFE-Digital/teacher-training-api/pull/1278/files#diff-9d39c599c09a32142f10c55337b8a56eR78) and [@gpeng's comments](https://github.com/DFE-Digital/teacher-training-api/pull/1278#discussion_r400029777).

swagger/public_v1/template.yml
```diff
@@ -372,6 +372,47 @@ components:
         version:
           type: string
           example: "1.0"
+    LocationAttributes:
+      type: object
+      properties:
+        code:
+          type: string
+          example: "W"
+        location_name:
```
misaka: I've been trying to trim these down to just `name` with providers and courses, shall we do that here for consistency?

swagger/public_v1/template.yml
asmega: I've made these changes and it should now match your PR

swagger/public_v1/template.yml
asmega: agreed, done

misaka:
Can we add requests to this? Nothing fancy, but something to build on, and we should place this at the root of the API: `/api/public/v1/locations` is where I think it should go.

PR #1290, opened 2020-03-31
[3195] Add the commit sha to logging and sentry

danielburnley:
### Context

We want to see the version of the code that is deployed in both logging and when we get errors reported

- Add the `COMMIT_SHA` to both Sentry and Logstash

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/initializers/logstash.rb
+++ b/config/initializers/logstash.rb
@@ -1,6 +1,8 @@
 LogStashLogger.configure do |config|
+  commit_sha = File.read(Rails.root.join("COMMIT_SHA")).strip
   config.customize_event do |event|
     event["application"] = Settings.application
     event["environment"] = Rails.env
+    event["release"] = commit_sha
   end
 end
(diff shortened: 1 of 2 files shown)


config/initializers/logstash.rb
```diff
@@ -1,6 +1,8 @@
 LogStashLogger.configure do |config|
+  commit_sha = File.open("#{__dir__}/../../COMMIT_SHA") { |f| f.read.strip }
```
misaka:
```suggestion
  commit_sha = Rails.root.join("COMMIT_SHA") { |f| f.read.strip }
```

config/initializers/raven.rb
```diff
@@ -7,4 +7,7 @@
     ["ActiveRecord::RecordNotFound"]
 
   config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
+
+  commit_sha = File.open("#{__dir__}/../../COMMIT_SHA") { |f| f.read.strip }
```
misaka:
```suggestion
  commit_sha = File.read(Rails.root.join("COMMIT_SHA")).strip
```

PR #1291, opened 2020-03-31
[3210] Add v3 subjects endpoint

faissalMT:
### Context  
Find currently contains this hack to get all subjects.
https://github.com/DFE-Digital/find-teacher-training/blob/master/app/view_objects/results_view.rb#L364

Have v3 provide a proper way to provide all subjects

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v3/subjects/index_spec.rb
+++ b/spec/requests/api/v3/subjects/index_spec.rb
@@ -843,5 +843,56 @@ describe "GET v3 /subjects" do
         "version" => "1.0",
       })
     end
+
+    context "when sorting by name" do
+      let(:request_path) { "/api/v3/subjects?fields[subjects]=subject_name&sort=subject_name" }
+      it "sorts the subject names in ascending order" do
+        expect(json_response["data"].map { |subject| subject["attributes"]["subject_name"] }).to eq([
+          "Art and design",
+          "Biology",
+          "Business studies",
+          "Chemistry",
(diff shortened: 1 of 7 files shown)


spec/requests/api/v3/subjects/index_spec.rb
```diff
@@ -0,0 +1,847 @@
...
+  let(:request_path) { "/api/v3/subjects" }
+  let(:json_response) { JSON.parse(response.body) }
+
+  before do
+    get request_path
+  end
+
+  it "returns the correct data" do
+    expect(json_response).to eq("data" => [
+        {
```
danielburnley: Maybe it's worth moving this out to a file - this is a pretty lengthy response and not _overly_ useful to see the whole thing of in the test IMO

spec/requests/api/v3/subjects/index_spec.rb
```diff
@@ -0,0 +1,847 @@
...
+      "jsonapi" => {
+      "version" => "1.0",
+    })
+  end
+
+  context "when specifying particular fields" do
+    let(:request_path) { "/api/v3/subjects?fields[subjects]=subject_name" }
+
+    it "returns the correct data" do
+      expect(json_response).to eq("data" => [
```
danielburnley: Similar comment to above - though we're not expecting to change these files much

PR #1294, opened 2020-04-02
3033 setup benchmarking performance test

gpeng:
### Context

We have performance issues on some endpoints. As part of the work to address this we want to run performance regression tests as part of the staging deploy pipeline. The thresholds in theses tests will be reduced as we improve performance.

The tests are intended to warn (but not block) during deployment. They haven't been added to the pipeline yet as this will need to be merged before they can be added.

Add spec/performance/pre_deploy_spec.rb that runs multiple requests against specified endpoints and compares their average response against a predefined threshold.

### Guidance to review

The tests can be run locally as per the docs.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/performance/pre_deploy_spec.rb
+++ b/spec/performance/pre_deploy_spec.rb
@@ -1,0 +1,38 @@
+require "net/http"
+require "rspec-benchmark"
+
+TEST_SAMPLE_COUNT = 5
+
+RSpec.configure do |config|
+  config.include RSpec::Benchmark::Matchers
+end
+
+def get(path)
+  uri = URI(ENV.fetch("CUSTOM_HOST_NAME", "http://localhost:3001"))
+  uri.path = path
+  Net::HTTP.get(uri)
+end
+
+recruitment_cycle = ENV.fetch("RECRUITMENT_CYCLE", "2020")
+provider_code = ENV.fetch("PROVIDER_CODE", "U80")
+course_code = ENV.fetch("COURSE_CODE", "2P3K")
+
+tests = {
+  "/api/v3/recruitment_cycles/#{recruitment_cycle}/providers/#{provider_code}" => 300,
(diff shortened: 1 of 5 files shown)


gpeng:
[Setup benchmarking / performance test](https://trello.com/c/lyS6VGel/3033-setup-benchmarking-performance-test)

spec/performance/pre_deploy_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+def get(path)
+  uri = URI(ENV.fetch("CUSTOM_HOST_NAME", "http://localhost:3001"))
+  uri.path = path
+  Net::HTTP.get(uri)
+end
+
+tests = {
+  "/api/v3/recruitment_cycles/2020/providers/U80" => 300,
+  "/api/v3/recruitment_cycles/2020/providers/U80/courses/2P3K" => 300,
```
defong, line 18: maybe 
these 2 should be configurable so we can have a separated azure tasks additional capabilities

gpeng: Not sure what you mean. Configurable in what way? We're pretty much aiming for a proof of concept here so as we can get it set up and then iterate on them along the way.

defong:
```
CUSTOM_HOST_NAME=https://api2.publish-teacher-training-courses.service.gov.uk /
PROVIDER_CODE=U80 /
COURSE_CODE=2P3K /
bundle exec rspec spec/performance/pre_deploy_spec.rb
```

Something like so...

gpeng: Right... I see what you mean. I think the plan is to run these on the integration test data we add once that's in there. I'll add the other env variables so as it's more flexible. Good idea.

PR #1306, opened 2020-04-08
[3253] - Filter training providers by subject and funding type

TobyRet:
### Context
We want to filter training providers by what kind of courses they provide, for example, list providers that have fee-funded, physical education courses

Adds `funding_type` and `subject` filters to the accredited body training providers endpoint 

Example request path:

```
      "/api/v2/recruitment_cycles/#{recruitment_cycle.year}/providers/#{accredited_provider.provider_code}/training_providers?filter[funding_type]=fee,salary&[subjects]=C6,A2"

```

### Out of scope
I suggest that _negative filters_ are tackled in another PR. Possibly we may end up handling client side.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -6,7 +6,17 @@ module API
 
       def index
         authorize @provider, :can_list_training_providers?
-        providers = @provider.training_providers
+        providers = if params[:filter]
+                      course_scope = Course.where(provider: training_providers)
+
+                      eligible_training_provider_ids = CourseSearchService
+                                                         .call(filter: params[:filter], course_scope: course_scope)
+                                                         .pluck(:provider_id)
+
+                      training_providers.where(id: eligible_training_provider_ids)
+                    else
+                      training_providers
+                    end
 
         render jsonapi: providers, include: params[:include]
       end
@@ -23,6 +33,10 @@ module API
         @provider = @recruitment_cycle.providers.find_by!(
           provider_code: params[:provider_code].upcase,
         )
+      end
+
+      def training_providers
+        @provider.training_providers
       end
     end
   end
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -16,7 +16,7 @@ module API
                                                          .call(filter: params[:filter], course_scope: course_scope)
                                                          .pluck(:provider_id)
 
-                      training_providers.where(id: eligible_training_provider_ids)
+                      training_providers.where(id: eligible_training_provider_ids).order(:provider_name)
                     else
                       training_providers
                     end
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -7,7 +7,10 @@ module API
       def index
         authorize @provider, :can_list_training_providers?
         providers = if params[:filter]
-                      course_scope = Course.where(provider: training_providers)
+                      course_scope = Course.where(
+                        provider: training_providers,
+                        accrediting_provider_code: @provider.provider_code,
+                      )
 
                       eligible_training_provider_ids = CourseSearchService
                                                          .call(filter: params[:filter], course_scope: course_scope)
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -239,19 +239,21 @@ class Course < ApplicationRecord
     program_types = []
 
     if funding_types.include?("salary")
-      program_types.push :school_direct_salaried_training_programme
+      program_types << :school_direct_salaried_training_programme
     end
 
     if funding_types.include?("apprenticeship")
-      program_types.push :pg_teaching_apprenticeship
+      program_types << :pg_teaching_apprenticeship
     end
 
     if funding_types.include?("fee")
-      program_types.push(
-        :higher_education_programme,
-        :scitt_programme,
-        :school_direct_training_programme,
-      )
+      %i[
+        higher_education_programme
+        scitt_programme
+        school_direct_training_programme
+      ].each do |program_type|
+        program_types << program_type
+      end
     end
 
     where(program_type: program_types)
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -250,7 +250,7 @@ class Course < ApplicationRecord
       program_types.push(
         :higher_education_programme,
         :scitt_programme,
-        :school_direct_training_programme
+        :school_direct_training_programme,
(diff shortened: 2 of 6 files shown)


app/controllers/api/v2/accredited_provider_training_providers_controller.rb
```diff
@@ -6,7 +6,21 @@ class AccreditedProviderTrainingProvidersController < API::V2::ApplicationContro
 
       def index
         authorize @provider, :can_list_training_providers?
-        providers = @provider.training_providers
+        if params[:filter]
```
TobyRet, line 9: Not overly satisfied with this section of code. 
In order to use the CourseSearchService in all it's glory we need to pass in an ActiveRecordRelation containing courses ...

app/models/course.rb
```diff
@@ -235,6 +235,28 @@ def new_draft_attributes
     where(is_send: true)
   end
 
+  scope :with_funding_types, ->(funding_types) do
```
TobyRet, line 238: Mapping from nice to not so nice

TobyRet:
[2a. [L] Add support to providers endpoint to filter by subject (e.g. PE)](https://trello.com/c/jImmdH7D/3253-2a-l-add-support-to-providers-endpoint-to-filter-by-subject-eg-pe)

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
asmega: does `Course.where(provider: provider.training_providers)` give you the same result? it should save the ruby `map`

asmega: personal preference for `providers = if params[:filter]` and delete the line above. but don't know if rubocop allows it?

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
```diff
@@ -6,7 +6,21 @@ class AccreditedProviderTrainingProvidersController < API::V2::ApplicationContro
...
+            tp.courses.pluck(:id)
+          end.flatten
+
+          course_scope = Course.where(id: course_ids)
+
+          eligible_training_provider_ids = CourseSearchService
+                                             .call(filter: params[:filter], course_scope: course_scope)
+                                             .pluck(:provider_id)
+
+          providers = @training_providers.select { |provider| eligible_training_provider_ids.include?(provider.id) }
```
asmega, line 20: is this a better option?
```
providers = @training_providers.where(id: eligible_training_provider_ids)
```

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
```diff
@@ -24,6 +38,14 @@ def build_provider
           provider_code: params[:provider_code].upcase,
         )
       end
+
+      def training_providers
+        @training_providers ||= @provider.training_providers.includes(:courses)
+      end
+
+      def training_provider_ids
```
asmega, line 46: is this new method used any where?

app/models/course.rb
```diff
@@ -235,6 +235,28 @@ def new_draft_attributes
...
+    program_types = []
+
+    if funding_types.include?("salary")
+      program_types.push :school_direct_salaried_training_programme
```
asmega, line 242: personal preference to use shovel `<<` instead of `push` so
```
program_types << :school_direct_salaried_training_programme
```

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
TobyRet: Ah of course. Good spot. That will clean up the code nicely

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
TobyRet: Not so sure about this one? By using 'select' we save another db call ...

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
TobyRet: This is gone now :-)

app/models/course.rb
TobyRet: Sure no prob

spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+    let(:non_pe_course) do
+      create(:course,
+             level: :secondary,
+             provider: delivering_provider2,
+             subjects: [biology],
+             site_statuses: [build(:site_status, :findable, site: build(:site))],
+             accrediting_provider: accredited_provider)
+    end
+
+    let(:delivering_provider1) { create(:provider) }
```
AgaDufrat, line 49: We don't seem to have an official name for this kind of provider but we started calling them `training_providers` ([small discussion here](https://ukgovernmentdfe.slack.com/archives/CAHAJFR0X/p1586179199434100?thread_ts=1586179143.434000&cid=CAHAJFR0X)). It's used across the api and Publish. I was wondering if we could rename them to keep things consistent.

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
```diff
@@ -6,7 +6,17 @@ class AccreditedProviderTrainingProvidersController < API::V2::ApplicationContro
...
+                      eligible_training_provider_ids = CourseSearchService
+                                                         .call(filter: params[:filter], course_scope: course_scope)
+                                                         .pluck(:provider_id)
+
+                      training_providers.where(id: eligible_training_provider_ids)
+                    else
+                      training_providers
+                    end
 
         render jsonapi: providers, include: params[:include]
```
asmega, line 21: do we need to consider pagination in the api? if not then not, then theres no issue. if yes then we want the result set to be ordered so we can paginate through them

spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
TobyRet: Thanks @AgaDufrat - I will make the change

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
TobyRet: @AgaDufrat - do you need pagination for your ticket? If need I'd be inclined to deploy this as is and do pagination as a separate ticket (if req'd)

AgaDufrat: My ticket doesn't say anything about the. The highest possible number of training providers we can have on the pages that use this is 36. More of a product question.

PR #1308, opened 2020-04-09
[3246] paginate v2 providers endpoint

asmega:
### Context

- https://trello.com/c/xM3j6Ohd/3246-paginate-providers-in-publish-and-api
- Server side pagination has not been implemented of v2 providers endpoint
- This is consumed by publish and is slow
- We implement server side pagination to improve performance

- add pagination to v2 providers endpoint
- hard coded to 10 providers per page
- adds meta count to response

### Guidance to review

```
curl -H "Authorization: Bearer bats" "http://localhost:3001/api/v2/recruitment_cycles/2020/providers" | ruby -e "require 'json'; puts JSON.parse(gets)['data'].size"
```

- the above should return 10

```
curl -H "Authorization: Bearer bats" "http://localhost:3001/api/v2/recruitment_cycles/2020/providers?page[page]=1" |  ruby -e "require 'json'; puts JSON.parse(gets)['data']"
```

- should return page 1

```
curl -H "Authorization: Bearer bats" "http://localhost:3001/api/v2/recruitment_cycles/2020/providers?page[page]=2" |  ruby -e "require 'json'; puts JSON.parse(gets)['data']"
```

- should return page 2

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -11,13 +11,19 @@ module API
 
       def index
         authorize Provider
+
         providers = policy_scope(@recruitment_cycle.providers)
                       .include_courses_counts
                       .includes(:recruitment_cycle)
+                      .by_name_ascending
+
         providers = providers.where(id: @user.providers) if @user.present?
 
-        render jsonapi: providers.by_name_ascending,
-               fields: { providers: %i[provider_code provider_name courses
+        render jsonapi: paginate(providers, per_page: 10),
+               meta: { count: providers.count(:provider_code) },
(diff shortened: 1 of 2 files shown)


asmega:
[Agree pagination pattern and refactor Find pagination](https://trello.com/c/xM3j6Ohd/3246-agree-pagination-pattern-and-refactor-find-pagination)

dankmitchell:
This does hide the pagination on PUBLISH since we're only returning  x10  records now so that'll need to be fixed cc. @aliuk2012

PR #1310, opened 2020-04-15
3284 bug autocomplete not reporting all providers

aliuk2012:
- A new normalised column added to the provider table. Removes any non-alphanumeric characters and downcases the provider name before storing it in the new column
- New `before_save` callback to calculate and update the new column
- provider scope changed to normalize query string and use the new normalized column

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


aliuk2012:
[Bug Autocomplete not reporting all providers](https://trello.com/c/aCTUwu9e/3284-bug-autocomplete-not-reporting-all-providers)

app/models/provider.rb
```diff
@@ -310,4 +315,8 @@ def services
       )
     end
   end
+
+  def set_provider_name_search
+    self.provider_name_search = self.provider_name.downcase.gsub(/[^0-9a-z]/i, "") if self.provider_name
```
TobyRet, line 320: Regex could be extracted (repeated on line 116)

aliuk2012: Not sure I follow you. This method is to normalise the data that we are going to store. Line 116 is normalising the query string the user inputted.

db/migrate/20200415150049_set_provider_name_search_on_provider.rb
```diff
@@ -0,0 +1,12 @@
+class SetProviderNameSearchOnProvider < ActiveRecord::Migration[6.0]
+  def up
```
TobyRet, line 2: I can see the case for using a migration to add / update columns (as above) but not sure it should be used to populate data - I would use a rake task for this

db/migrate/20200415150049_set_provider_name_search_on_provider.rb
```diff
@@ -0,0 +1,12 @@
...
+    say_with_time "Adding normalised version of provider_name to provider_name_search" do
```
TobyRet, line 3: As this is a once-off may be more appropriate to do via a temporary rake task
Keen to hear what others think ...

app/models/provider.rb
TobyRet: `/[^0-9a-z]/i` could be extracted to a constant?

db/migrate/20200415150049_set_provider_name_search_on_provider.rb
defong: Migration serves the purpose of snapshoting both
- structural changes 
- data manipulation


`20200415144710`_add_provider_name_search_to_provider.rb (structural changes)
`20200415150049`_set_provider_name_search_on_provider.rb (data manipulation)

when this is removed `20200415150049` and replace with a temporary rake task...
who will know
- when to execute it
- why you execute
- why database is broken
- why did we keep/remove temporary rake task

db/schema.rb
```diff
@@ -220,6 +220,7 @@
     t.jsonb "accrediting_provider_enrichments"
     t.float "latitude"
     t.float "longitude"
+    t.string "provider_name_search", null: false
```
defong, line 223: since this is `not null` why is its dependent provider_name `nullable`?

spec/factories/providers.rb
```diff
@@ -41,6 +42,7 @@
 FactoryBot.define do
   factory :provider do
     provider_name { "ACME SCITT" + rand(1000000).to_s }
+    provider_name_search { self.provider_name.downcase.gsub(/[^0-9a-z]/i, "") }
```
defong, line 45: does the before hook not kicks in?

db/schema.rb
aliuk2012: Good question, I just assumed the provider_name would not be nullable

app/models/provider.rb
```diff
@@ -112,7 +113,9 @@ def current_accredited_courses
   end
 
   scope :search_by_code_or_name, ->(search_term) {
-    where("provider_name ILIKE ? OR provider_code ILIKE ?", "%#{search_term}%", "%#{search_term}%")
+    modified_search_term = "%#{search_term.downcase.gsub(/[^0-9a-z]/i, '')}%"
```
misaka, line 116: Let's use `URI.decode`

spec/factories/providers.rb
aliuk2012: It wasn't needed. 👍

db/migrate/20200415150049_set_provider_name_search_on_provider.rb
TobyRet: I don't think rails migrations for data migration is the norm in Rail land @defong ...

From: https://www.ombulabs.com/blog/rails/data-migrations/three-useful-data-migrations-patterns-in-rails.html

**Pros**
- It's quick because you can ship a data migration in a few minutes
- You use Rails core logic that makes sure you don't run a migration more than once

**Cons**
- Migrations fall out of date very quickly and you don't have a quick way to find out
- You're not following Rails conventions: Files in db/migrate should migrate database structure, not database records

From: https://thoughtbot.com/blog/data-migrations-in-rails

"_Rails migrations should only be used for schema changes and not for actual data changes in the database._"

app/models/provider.rb
aliuk2012: I don't think we need this as a constant as its no longer in the same file.

db/migrate/20200415150049_set_provider_name_search_on_provider.rb
defong: That decision of not actual data changes in the migration was already defined by 20190227124613_initialize_provider_changed_at.rb

asmega: Personally I'm fine with data migrations in rails migrations. A few things to note are:
- split it up from the structural change which is done
- down migration, probably not needed here as previous migration will remove the column
- if some one new starts the project they should be able to run all existing migrations without an error. This usually happens if there is existing data a migration needs is missing and causes a failure. Doesn't look like it will happen in this case.

app/controllers/api/v2/providers_controller.rb
```diff
@@ -50,12 +50,13 @@ def update
 
       def suggest
         authorize Provider
-
         return render(status: :bad_request) if params[:query].nil? || params[:query].length < 3
         return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
+        normalise_query = CGI.unescape(params[:query]).downcase.gsub(/[^0-9a-z]/i, "")
```
misaka, line 56: I'm surprised we have to do this, considering https://guides.rubyonrails.org/action_controller_overview.html#hash-and-array-parameters says that Rails already does this:

> The actual URL in this example will be encoded as "/clients?ids%5b%5d=1&ids%5b%5d=2&ids%5b%5d=3" as the "[" and "]" characters are not allowed in URLs. Most of the time you don't have to worry about this because the browser will encode it for you, and Rails will decode it automatically, but if you ever find yourself having to send those requests to the server manually you should keep this in mind.

Have we checked that unescaping is necessary?

misaka: Also, I think that the normalisation should be happening in the `Provider` model, while the parameter, if any, should happen here in the controller. Sorry if that wasn't clear with what I said earlier, but since the normalisation column is in the model I think it belongs there, no?

app/controllers/api/v3/providers_controller.rb
```diff
@@ -7,7 +7,8 @@ def index
         build_fields_for_index
 
         @providers = if params[:search].present?
-                       @recruitment_cycle.providers.search_by_code_or_name(params[:search])
+                       normalise_query = CGI.unescape(params[:search]).downcase.gsub(/[^0-9a-z]/i, "")
```
misaka, line 10: Similar as above about whether we need `CGI.unescape` at all, and separation of responsibilities.`

app/models/provider.rb
misaka: I think @TobyRet is right. I'd go further, there should be a single method to do normalisation, and if possible used for both generating data for this column, and normalising the query when used to search.

db/migrate/20200415150049_set_provider_name_search_on_provider.rb
misaka: If we don't use data migrations, then I think we need to ensure our code is compatible with both versions of the data since there will be a time lag between when the migration runs and when the data will be updated. I don't mind moving towards that, but tbh I feel like it's more work to have code compatible with multiple versions in general.

app/controllers/api/v2/providers_controller.rb
aliuk2012: I'm not sure what you mean here? I had it in the Provider model scope but we discussed this yesterday and thought we wanted to move this out to the controller?

As for unescaping, it is necessary if the request being made has been escaped, it basically decodes it to the original state so we can normalize it ourselves

app/models/provider.rb
gpeng: +1 for extracting this. I'd make it a service which could be used here and in the controllers etc. We could unit test it then and get rid of all of the more expensive request specs that are really just testing that this works.

app/controllers/api/v2/providers_controller.rb
misaka: What was suggested is that unescaping should be moved here, not the normalisation.

misaka: re: escaping, so what you're saying is that we're escaping the escaping? i.e. we're doubling up on the escaping?

app/models/provider.rb
aliuk2012: Extract out into another PR #1324

app/controllers/api/v2/providers_controller.rb
aliuk2012: #1324 splits this out into its own service if you look at the specs, you will see we have still used `CGI.unescape`. You are correct that Rack will be unescaping it for us but I still think it's worth leaving in here should this service be used for whatever reason and the query is not unescaped.

aliuk2012:
#1324, #1326, #1328 replicates most of this and addresses the issues raised in this PR

PR #1311, opened 2020-04-16
Test faster Dockerfile

asmega:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


misaka:
@asmega This looks like the right approach to me.

asmega:
this code change has no impact on the build time

PR #1313, opened 2020-04-16
[3172] Fix dupe site_statuses on course create

asmega:
### Context

- https://trello.com/c/vElWfWcT/3172-prevent-duplicate-site-statuses-being-created-on-course-creation
- duplicate site_statuses were being generated on course creation

- this would occur for unpersisted records
- as hand creating site_statuses
- then calling super which is default activerecord
- default activerecord creates the join on has_many through
- hence dupe joins
- now the handcrafting only happens for persisted records
- non persisted uses default rails behaviour

### Guidance to review

- checkout branch
- take a note of `SiteStatus.count`
- in publisher create a course
- after creation of course `SiteStatus.count` should increase by the number of sites you selected and not double as it was previously

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -448,28 +448,16 @@ class Course < ApplicationRecord
     end
   end
 
-  def add_site!(site:)
-    is_course_new = ucas_status == :new # persist this before we change anything
-    site_status = site_statuses.find_or_initialize_by(site: site)
-    site_status.start! unless is_course_new
-    site_status.save! if persisted?
-  end
-
-  def remove_site!(site:)
-    site_status = site_statuses.find_by!(site: site)
-    ucas_status == :new ? site_status.destroy! : site_status.suspend!
-  end
-
   def sites=(desired_sites)
     existing_sites = sites
 
-    to_add = desired_sites - existing_sites
-    to_add.each { |site| add_site!(site: site) }
-
-    to_remove = existing_sites - desired_sites
-    to_remove.each { |site| remove_site!(site: site) }
-
     if persisted?
+      to_add = desired_sites - existing_sites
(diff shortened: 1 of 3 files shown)


asmega:
[Prevent duplicate site statuses being created on course creation](https://trello.com/c/vElWfWcT/3172-prevent-duplicate-site-statuses-being-created-on-course-creation)

dankmitchell:
Looks good to me. @misaka can you review?

spec/models/course_spec.rb
```diff
@@ -1678,31 +1679,10 @@
...
 
       it "sets the site to new when a new site is added" do
-        expect { subject.add_site!(site: new_site) }.to change { subject.reload.site_statuses.size }.
-          from(0).to(1)
+        expect { subject.sites = [new_site] }.to change { subject.reload.site_statuses.size }.from(0).to(1)
         expect(new_site_status.status).to eq("new_status")
       end
     end
-
-    context "for mixed courses with new and running locations" do
```
asmega, line 1687: I've removed this as i do not believe it is possible to have a mixture of new and running `site_statuses`

misaka: 👍

app/models/course.rb
```diff
@@ -553,6 +541,18 @@ def assignable_subjects
 
 private
 
+  def add_site!(site:)
+    is_course_new = ucas_status == :new
```
misaka, line 545: This temporary variable is probably not adding anything, and also a bit odd as `remove_site!` doesn't bother with it. Happy to worry about this later if you prefer.

PR #1314, opened 2020-04-17
[3290] Vacancies only consider findable courses

asmega:
### Context

- https://trello.com/c/5CqPh7eV/3290-bug-course-still-showing-on-find-even-though-no-vacancies
- Search would return courses that had no vacancies despite user searching for courses with vacancies

- When combining scopes #findable and #with_vacancies
- These would be combined with OR instead of AND
- This would introduce bugs where a course could be return that should
not be because it was unfindable with vacancies AND findable with no
vacancies
- This fixes scopes so activerecord can combine these scopes correctly

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -202,8 +202,8 @@ class Course < ApplicationRecord
   end
 
   scope :with_recruitment_cycle, ->(year) { joins(provider: :recruitment_cycle).where(recruitment_cycle: { year: year }) }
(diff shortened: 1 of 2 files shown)


asmega:
[Bug: course still showing on Find even though no vacancies](https://trello.com/c/5CqPh7eV/3290-bug-course-still-showing-on-find-even-though-no-vacancies)

PR #1318, opened 2020-04-20
Add node js versions to "docs"

aliuk2012:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


defong:
just wondering should it not be in root?

aliuk2012:
@defong good point, I hadn't thought of that but I think it's better to place it here because it may seem as though API itself relies on specific versions of node whereas it's only needed if you want to locally run the api docs

misaka:
Ya, I agree that it's more appropriate here. There may be a better way to organise all this, but for now I think it better if our docs and api are in sync.

PR #1321, opened 2020-04-20
3257 4 pre a add crud allocationrequest to the api

gpeng:
### Context

We are adding a re-request allocation feature to Publish. An API endpoint is required to allow the user to create an `Allocation` record that represents a request to run a PE course again.

This endpoint will be used to:

* re-request another allocation: accredited_body_id and provider_id are supplied. (not yet implemented: The number of places is inferred from the previous cycle allocation)

* notify us that no allocation is required: provider_id` and `number_of_places=0`

* request a new allocation: `provider_id` and `number_of_places=<number_required>`

The card initially specified create and update but we don't have a use case for update currently. All of the above use cases will create a new record.

### Discuss

This PR currently creates the `Allocation` record with `accredited_body_id` and `provider_id`. Using the IDs will allow us to ensure that only one `Allocation` record per pair, per recruitment cycle is created as the IDs are unique per recruitment cycle. It is going to make it difficult to retrieve a previous year's record though. If we use the provider codes then we need another field to disambiguate across cycles but it will be trivial to retrieve the previous year.

Possible options:

* Leave as is and do the more complex stuff when retrieving the previous year
* store the ids and the codes
* store the codes and the recruitment cycle year or id

Add `POST /api/v2/providers/<provider_code>/allocations`
Add `Allocation`
Add `AllocationPolicy`
Add `SerializableAllocation`

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history


gpeng:
[[M] 4 pre-a. Add C(R)U(D) Allocation to the API](https://trello.com/c/NH1J0gZk/3257-m-4-pre-a-add-crud-allocation-to-the-api)

gpeng:
[[M] 4 pre-a. Add C(R)U(D) Allocation to the API](https://trello.com/c/NH1J0gZk/3257-m-4-pre-a-add-crud-allocation-to-the-api)

gpeng:
I've just realised that this doesn't follow the URL scheme that we agreed in https://trello.com/c/V5Zg3kRc/3277-m-add-listing-to-allocations-endpoint. I'll update...

app/controllers/api/v2/allocations_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        else
+          render jsonapi_errors: @allocation.errors, status: :unprocessable_entity
+        end
+      end
+
+    private
+
+      def accredited_body
+        @accredited_body ||= begin
+                               accredited_body_code = params[:provider_code]
```
TobyRet, line 18: If possible can we start using the term `training_provider(s)` as opposed to `provider` where appropriate 
See https://ukgovernmentdfe.slack.com/archives/CAHAJFR0X/p1586179199434100?thread_ts=1586179143.434000&cid=CAHAJFR0X

TobyRet: I know it's possible as we are stuck with the routes and this is a more general point for wider discussion, but as an example this route that was added made my head do a couple of somersaults before I realised what was going on 

```
/api/v2/providers/:provider_code/allocations(.:format)      
```

Looking at the above it's not clear whether this is referring to the the Accrediting Body or Training Provider (yet appreciate these are both types of provider) - this knowledge is implicit, would be good if it were more explicit ?

spec/requests/api/v2/allocations_spec.rb
```diff
@@ -0,0 +1,64 @@
+require "rails_helper"
+
+RSpec.describe "/api/v2/providers/<accredited_body_code>/allocations", type: :request do
+  describe "POST" do
+    context "with valid parameters" do
+      it "returns 201" do
```
TobyRet, line 6: Really think these bdd-styley tests are great. Way more readable 👍

app/controllers/api/v2/allocations_controller.rb
```diff
@@ -0,0 +1,33 @@
+module API
+  module V2
+    class AllocationsController < API::V2::ApplicationController
+      def create
+        authorize @allocation = Allocation.new(allocation_params.merge(accredited_body_id: accredited_body.id))
```
asmega, line 5: i think we should get this in for now, but a discussion on whether or not this end point should be idempotent may come into play when we look at when users change their request

config/routes.rb
```diff
@@ -24,6 +24,7 @@
 #                                               api_v2 GET    /api/v2/recruitment_cycles/:recruitment_cycle_year/providers/suggest(.:format)                                                           api/v2/providers#suggest
 #                                 api_v2_organisations GET    /api/v2/organisations(.:format)                                                                                                          api/v2/organisations#index
 #                   api_v2_provider_recruitment_cycles GET    /api/v2/providers/:provider_code/recruitment_cycles(.:format)                                                                            api/v2/recruitment_cycles#index
+#                          api_v2_provider_allocations POST   /api/v2/providers/:provider_code/allocations(.:format)                                                                                   api/v2/allocations#create
```
TobyRet, line 27: As an example referring to my earlier appoint above ... (NB this is more of a general comment for discussion)
Looking at the new route that has been generated  its not immediately obvious was the relationship is between the allocation and the  Accrediting Body or Training Provider  - this knowledge is implicit, would be easier to get one's head around if it were more explicit in the code somehow?

app/controllers/api/v2/allocations_controller.rb
gpeng: Yeah, I agree. I think we should consider using `accredited_body`, `training_provider` in the URLs too. I did discuss this a bit with @misaka when we were coming up with the URLs. Let's revisit as I think it's a good idea.

app/controllers/api/v2/allocations_controller.rb
gpeng: I was thinking to just add a uniqueness validation to prevent duplicates being created once we settle on what a 'duplicate' is with the recruitment cycle stuff and then just add an explicit update action when we're doing the updates. Happy to discuss that along the way though.

config/routes.rb
gpeng: Yep, agreed as above.

app/controllers/api/v2/allocations_controller.rb
gpeng: We've discussed this further offline. If we change these URLs then we're really need to start considering `AccreditedBody` and `TraningProvider` as separate resources with the client classes etc in the frontend apps or we'd have to start making the existing resources quite messy to generate these URLs. So... I think we should move forward with this for now pending a bit of further discussion (cc/ @misaka)

PR #1322, opened 2020-04-21
3277 m add listing to allocations endpoint

gpeng:
### Context

Add an endpoint to the allocations API to allow allocations for an accredited body to be retreived.

Add `GET /api/v2/providers/:accredited_body_code/allocations`
Update the `AllocationPolicy` with a scope and allow admins

### Guidance to review

This is on top of #1321. There is discussion in the description there about the relationship of allocations with the recruitment cycle. This PR carries on relating through the provider ids (which are recruitment cycle specific). I wonder if it would be useful to associate `Allocation` with recruitment cycle directly? 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/allocations_spec.rb
+++ b/spec/requests/api/v2/allocations_spec.rb
@@ -23,6 +23,18 @@ RSpec.describe "/api/v2/providers/<accredited_body_code>/allocations", type: :re
     end
   end
 
+  describe "GET" do
+    it "returns the allocations for the accredited body from the current recruitment cycle" do
+      given_an_accredited_body_exists
+      given_the_accredited_body_has_a_training_provider
+      given_the_accredited_body_has_allocations_for_the_current_recruitment_cycle
+      given_the_accredited_body_has_allocations_from_the_previous_recruitment_cycle
+      given_i_am_an_authenticated_user_from_the_accredited_body
+      when_i_get_the_allocations_index_endpoint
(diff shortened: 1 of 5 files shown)


gpeng:
[[M] Add listing to allocations endpoint](https://trello.com/c/V5Zg3kRc/3277-m-add-listing-to-allocations-endpoint)

spec/requests/api/v2/allocations_spec.rb
```diff
@@ -50,6 +72,10 @@ def when_invalid_parameters_are_posted_to_the_allocations_endpoint
     post "/api/v2/providers/#{@accredited_body.provider_code}/allocations", params: { allocation: { provider_id: invalid_provider_id } }, headers: { "HTTP_AUTHORIZATION" => @credentials }
   end
 
+  def when_i_get_the_allocations_index_endpoint
+    get "/api/v2/providers/#{@accredited_body.provider_code}/allocations", headers: { "HTTP_AUTHORIZATION" => @credentials }
```
defong, line 76: Just wondering why not use the railsy path?

gpeng: I think if you hard code the URL then you're testing the routes are resolving as you expect. If you use the path helpers then you could have a mistake in the route and the test still pass. So, I'd go with hard coding it at this top level of testing.

PR #1326, opened 2020-04-23
3284 Normalize Provider name for improving search results

aliuk2012:
### Context
#1310 will be split into three smaller PR and deployed individually. This is the two of three.

Updates the data so that we can use the new column for improved provider search results 
- Adds a new column `provider_name_search` to Provider table
- Migration to populate the new column
- Adds a "not null" constraint to the new column
- Adds a "not null" constraint toDB and model validation to Provider
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/models/provider.rb
```diff
@@ -310,4 +314,8 @@ def services
       )
     end
   end
+
+  def set_provider_name_search
+    self.provider_name_search = QueryNormalizerService.call(query: self.provider_name)
```
defong, line 319: does it need a token test... ie, mock was called etc?

spec/models/provider_spec.rb
```diff
@@ -390,6 +400,19 @@
     end
   end
 
+  describe "before_save" do
+    subject { create :provider, provider_name: "Location 1 (Removed spaces and special characters)"}
```
defong, line 404: linter

app/models/provider.rb
gpeng: I agree... I think we might be better with a mocked test here as the `QueryNormalizerService` is tested already. We just need to make sure it is used.

gpeng:
Closing in favour of https://github.com/DFE-Digital/teacher-training-api/pull/1332

PR #1332, opened 2020-04-24
Add pg_search to Provider

gpeng:
### Context

The provider suggestions and search are being confounded by the inclusion of unicode characters, incorrect numbers of spaces etc. This introduces a more effective search that can handle that.

image

^ note sexy apostrophe and extra spaces in the search term being correctly matched 🕺 

* Bundle pg_search gem
* Switch search_by_code_or_name to use a combined scope on
`provider_code` and `provider_name`
 
### Guidance to review

Run find locally against this version of the API. Update a provider that has findable courses to 
have a name including a standard apostrophe. Enter the name but substitute a unicode apostrophe. The provider suggestion should now work.

This will need thorough product review on QA when it is merged as we may alter the expected search results. This introduces some fuzziness and partial-ness (😬) to the search.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -171,16 +171,6 @@ describe Provider, type: :model do
     end
   end
 
-  describe ".search_by_code_or_name" do
-    let(:provider1) { create(:provider, provider_name: "Zork") }
-    let(:provider2) { create(:provider, provider_name: "Acme") }
-
-    subject { Provider.search_by_code_or_name("zork") }
-
-    it { should include(provider1) }
-    it { should_not include(provider2) }
-  end
-
   describe "#update_changed_at" do
     let(:provider) { create(:provider, changed_at: 1.hour.ago) }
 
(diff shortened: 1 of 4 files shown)


gpeng:
[Add pg_search to fix unicode search errors](https://trello.com/c/4jx26WmY/3311-add-pgsearch-to-fix-unicode-search-errors)

spec/models/provider_spec.rb
```diff
@@ -731,4 +721,46 @@
...
+      it { is_expected.to contain_exactly(matching_provider) }
+    end
+
+    context "with non matching case name" do
+      let(:search_term) { "dave's searches" }
+      it { is_expected.to contain_exactly(matching_provider) }
+    end
+
+    context "with partial search term" do
+      let(:search_term) { "dave" }
```
defong, line 762: Just out of curiosity search terms as dive or ave?

gpeng: It seems to only match complete words at the moment but we can expand that in config.

So with a provider name Boys' School

matches:

* school boys
* boys school
* boys’   school

doesn't match:

* bys school

gpeng:
The original implementation was breaking a partial search test. I've updated now to add partial support. 
image

spec/models/provider_spec.rb
gpeng: This isn't true now... see below

PR #1336, opened 2020-04-28
[3216] Remove vnet config from template

saliceti:
### Context
When we deploy, the route table added by CIP is removed. It is required to use the Internet Access Service (IAS). This is because our Azure template also configures vnet and subnet, but without route table.

Use a simpler network template which does not try to recreate the network config, but only references the existing.

### Guidance to review
It was deployed here: https://dev.azure.com/dfe-ssp/Become-A-Teacher/_releaseProgress?_a=release-environment-logs&releaseId=5843&environmentId=23304
And the route table has not been removed: https://portal.azure.com/#@platform.education.gov.uk/resource/subscriptions/f013afaf-9685-4a48-91ae-155e77aa42fc/resourceGroups/s121d01-core/providers/Microsoft.Network/virtualNetworks/s121d01-core-vn-01/subnets

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -847,7 +847,7 @@
       "properties": {
         "mode": "Incremental",
         "templateLink": {
-          "uri": "[concat(variables('deploymentUrlBase'), 'container-instance-network-profile-subnet-delg.json')]",
+          "uri": "[concat(variables('deploymentUrlBase'), 'container-instance-network-profile.json')]",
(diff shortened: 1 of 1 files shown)


saliceti:
[Spike Azure static IP](https://trello.com/c/Yguk9SY5/3216-spike-azure-static-ip)

PR #1337, opened 2020-04-28
[3317] - Add V2 API user notifications create endpoint

TobyRet:
### Context

In order to allow users to opt in and opt out of receiving course update notifications we need to add some new end points to V2.

- Endpoint added for creating a user notification - `POST /api/v2/users/:user_id/notifications`

- Accepted parameters:
    - user_id (str)
    - course_create (bool)
    - course_update (bool)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[[S] API - Create Notifications endpoint](https://trello.com/c/hbzZH5SL/3317-s-api-create-notifications-endpoint)

app/controllers/api/v2/notifications_controller.rb
```diff
@@ -0,0 +1,42 @@
+module API
+  module V2
+    class NotificationsController < API::V2::ApplicationController
+      before_action :build_provider
+      deserializable_resource :user_notification,
+                              class: API::V2::DeserializableNotification
+      def create
```
TobyRet, line 7: It was suggested in the ticket that we go for a single endpoint to create and update notifications but inclined to thing we should go for a separate update endpoint unless people strongly disagree?

aliuk2012: I agree with you about a separate update endpoint but does it have to be another PR?

app/models/user_notification.rb
```diff
@@ -23,6 +23,8 @@ class UserNotification < ApplicationRecord
              primary_key: :provider_code,
              inverse_of: :user_notifications
 
+  validates :course_create, :course_update, presence: true
```
aliuk2012, line 26: ♥️

spec/requests/api/v2/users/notifications_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    post "/api/v2/users/#{@user.id}/notifications",
+         params: params,
+         headers: { "HTTP_AUTHORIZATION" => @credentials }
+  end
+
+  def then_a_new_opt_in_notification_is_created
+    expect(response).to have_http_status(:created)
+    parsed_response = JSON.parse(response.body)
```
defong, line 110: passing thought

```ruby 
def parsed_response
  JSON.parse(response.body)
end
```

spec/requests/api/v2/users/notifications_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+      },
+    }
+
+    post "/api/v2/users/#{@user.id}/notifications",
+         params: params,
+         headers: { "HTTP_AUTHORIZATION" => @credentials }
+  end
+
+  def when_valid_parameters_are_posted_opting_out_of_notifications
+    params = {
```
defong, line 91: passing thought

```ruby
    params = generate_some_params (course_create, course_update)


def generate_some_params (course_create, course_update)

 {
      "_jsonapi" => {
        "data" => {
          "type" => "user_notifications",
          "attributes" => {
            "course_create" => "",
            "course_update" => ""
          },
        },
      },
    }
end
```

config/routes.rb
```diff
@@ -133,6 +133,7 @@
 
     namespace :v2 do
       resources :users, only: %i[update show] do
+        resources :notifications, only: %i[create]
```
defong, line 136: forgot to annotate, hanging build

app/controllers/api/v2/notifications_controller.rb
gpeng: +1 for a separate endpoint. Mostly because having a single endpoint for update and create isn't a pattern we have so it will be more consistent with the rest of the API. I'm easy on whether you do it here or another PR.

gpeng:
Didn't we decide we were not going to update in the create and that we were going to create one `Notification` per accredited body rather than one per provider that has the accredited body as its accredited body?

SamJamCul:
> 
> 
> Didn't we decide we were not going to update in the create and that we were going to create one `Notification` per accredited body rather than one per provider that has the accredited body as its accredited body?

I'm not sure I understand, this change will create a user_notification row for every accredited body that the current user is associated with. At the moment we're doing this as a master switch, so when the endpoint gets hit, it turns on all possible notifications for that user!

app/controllers/api/v2/notifications_controller.rb
```diff
@@ -0,0 +1,31 @@
+module API
+  module V2
+    class NotificationsController < API::V2::ApplicationController
+      deserializable_resource :user_notification,
+                              class: API::V2::DeserializableNotification
+      def create
+        @current_user.providers.where(accrediting_provider: :accredited_body).map do |provider|
```
misaka, line 7:
```suggestion
        @current_user.providers.accredited_body.map do |provider|
```

PR #1338, opened 2020-04-29
[3273] Forbid access for discarded users

dankmitchell:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


dankmitchell:
@misaka I weighed that up but felt like the policy was the correct place to decide on access... happy to talk about this.

misaka:
Chatted with @dankmitchell, I think we've agreed it would be good to have this check in both places, in authentication to prevent access to any pages when logging-in, and in the policies to prevent access to viewing resources.

PR #1340, opened 2020-04-29
3321 m Allocation api changes

gpeng:
### Context

`Allocation` needs to store a `request_type` so as we can easily distinguish between new allocations and those that are either re-requests of a previous year or the user notifying us that they don't want an allocation this year having previously had one.

Add `Allocation#request_type` enum.

Values:

* `initial` a new request. The ticket specified `new` as the value here but that is reserved.
* `repeat` a re-request, called repeat to avoid the redundant `-request`
* `declined` no allocation requested this year

Add some support for this to the API endpoint. The client is currently sending `number_of_places` for all types so we need to change that before we can complete this work.

Not included in this PR but still to do:

* Retrieve the number_of_places value from the previous year for `repeat` types
* Validate that only `initial` requests can specify a number of places
* Probably move the hacky callback from `Allocation`. The logic differs between update and create and would probably be better handled in a service. We'll inevitably add other functionality around create/update e.g audit, notification etc too so a service will be the right thing to do ™️ there. 
* Remove the number of places testing in the request spec except for initial requests

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


gpeng:
[[M] Allocation API changes](https://trello.com/c/fJVD9GoI/3321-m-allocation-api-changes)

gpeng:
@TobyRet yeah. The terminology is a bit tricky. I think this is consistent with what @AgaDufrat has though. That action you linked is to make a `renew_request` and so the form passes `request_type: "renew"`. What do you think? Were you suggesting we go `request_type: 'renew_request'`? I think we had that from tech breakdance initially but the `_request` suffix seemed redundant.

PR #1341, opened 2020-04-30
[3282] Import PE allocations

asmega:
### Context

- https://trello.com/c/DYxjNwkG/3282-4c-add-last-years-allocation-data-to-database
- Import PE allocations for 2020/2021
- This only imports into the database and does not contain the work wire up the data

- Imports 2020/2021 PE allocations from CSV
- There are 7 allocations where provider does not offer a PE course

### Guidance to review

- `bundle exec rails c`

```ruby
service = AllocationImporterService.new(path_to_csv: Rails.root.join('config/data/itt_allocations_2020_to_2021/physical_education.csv'))
service.execute
```

- should create allocations, check db

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


asmega:
[[S] 4c. Add last year's allocation data to database](https://trello.com/c/DYxjNwkG/3282-4c-add-last-years-allocation-data-to-database)

app/services/allocation_importer_service.rb
```diff
@@ -0,0 +1,36 @@
...
+  attr_reader :path_to_csv
+
+  def initialize(path_to_csv:)
+    @path_to_csv = path_to_csv
+  end
+
+  def execute
+    rows.each do |row|
+      training_provider = Provider.where(recruitment_cycle_id: 2).find_by(provider_code: row["training_provider_code"])
+      accredited_body_provider = Provider.where(recruitment_cycle_id: 2).find_by(provider_code: row["accredited_body_provider_code"])
```
misaka, line 13: This is probably ok in this case, but generally might be safer not to rely on the recruitment cycle always having id `2`:

```ruby
recruitment_cycle = RecruitmentCycle.current
training_provider = recruitment_cycle.providers.find_by(provider_code: row["training_provider_code"])
accredited_body_provider = recruitment_cycle.providers.find_by(provider_code: row["accredited_body_provider_code"])
```

misaka:
We can't run this on the deployed environments, are we going to integrate this with `mcb`?

app/services/allocation_importer_service.rb
asmega: yea i want to be explicit here as we're importing for a specific year. i don't really see it being reused otherwise we could inject the concerning recruitment_cycle. But i don't see that worth doing and see this as throw away code

asmega:
> We can't run this on the deployed environments, are we going to integrate this with `mcb`?

I was going to run via `be ./bin/mcb console`. Do you me to write an extra command? I don't really see it being used except this once

app/services/allocation_importer_service.rb
```diff
@@ -0,0 +1,36 @@
+require "csv"
+
+class AllocationImporterService
```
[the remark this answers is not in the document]
asmega: I wasn't planning on an MCB command. I have no idea how that works and will need to read up first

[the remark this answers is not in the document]
misaka: This is a one-off, so as @asmega says it's suited to just be used with `mcb console`.

config/data/itt_allocations_2020_to_2021/physical_education.csv
```diff
@@ -0,0 +1,390 @@
+allocation,training_provider_code,accredited_body_provider_code
```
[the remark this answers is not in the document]
misaka: I wondered the same. I probably wouldn't have, but it is public data so I don't think the _shouldn't_.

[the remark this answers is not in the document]
asmega: we can delete later, it's only needed for this one-off

app/services/allocation_importer_service.rb
misaka: How are we not importing for a specific year? Isn't the data for the 2020-2021 recruitment cycle?

asmega: > Isn't the data for the 2020-2021 recruitment cycle?

Yes

Sorry, I don't think I follow. `recruitment_cycle_id: 2` is the latest and correct recruitment cycle?

misaka: It is, I just wasn't clear about the comment being that we're not importing for a specific year. I don't think this is blocking.

PR #1342, opened 2020-04-30
[3318] - V2 API User Notifications index endpoint

TobyRet:
### Context
In order to allow users to opt in and opt out of receiving course update notifications we need to add some new end points to V2.

- Adds index endpoint for user notifications

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/notifications_controller.rb
+++ b/app/controllers/api/v2/notifications_controller.rb
@@ -3,6 +3,15 @@ module API
     class NotificationsController < API::V2::ApplicationController
       deserializable_resource :user_notification,
                               class: API::V2::DeserializableNotification
+
+      def index
+        authorize @notifications = @current_user.user_notifications
+
+        render jsonapi: @notifications,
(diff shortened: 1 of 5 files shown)


TobyRet:
[API - Get Notifications endpoint](https://trello.com/c/37rxrPGI/3318-api-get-notifications-endpoint)

app/controllers/api/v2/notifications_controller.rb
```diff
@@ -0,0 +1,40 @@
+module API
+  module V2
+    class NotificationsController < API::V2::ApplicationController
+      deserializable_resource :user_notification,
+                              class: API::V2::DeserializableNotification
+
+      def index
+        authorize @notifications = UserNotification.where(user_id: params[:user_id])
```
defong, line 8: given that there is `@current_user`

is there a real need to pass in the `user_id` as that kinda suggusts that they can be managed by anyone else, with out any bounded checks

TobyRet: Yep I've changed

PR #1344, opened 2020-04-30
3308 Refactor training provider search

asmega:
### Context

- https://trello.com/c/XSJooptJ/3308-m-6a-api-add-negated-subject-filtering-support-to-providers-endpoint
- Original PR name `3308 negate subject provider search`
- ~~Enables searching of training providers that do not offer specified subjects~~
- The original task is no longer required and this PR only contains a refactor

- Refactor `AccreditedProviderTrainingProvidersController` with internal service class
- Results now ordered by `provider_name`

### Guidance to review

- only change is now results are always sorted `provider_name`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -4,22 +4,45 @@ module API
       before_action :build_recruitment_cycle
       before_action :build_provider
 
+      class TrainingProviderSearch
+        attr_reader :provider, :params
+
+        def initialize(provider:, params:)
+          @provider = provider
+          @params = params
+        end
+
+        def call
+          scope = training_providers.order(:provider_name)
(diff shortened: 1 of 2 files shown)


asmega:
[[M] 6a. (API) Add negated subject filtering support to providers endpoint](https://trello.com/c/XSJooptJ/3308-m-6a-api-add-negated-subject-filtering-support-to-providers-endpoint)

AgaDufrat:
I don't think this will work in a way we want 😞 
We want it to return training_providers that don't offer a specific subject (say PE). Currently a training provider will not be returned if they *only* offer that negated subject. However, if they have some other ones eg. biology and PE, biology will be returned by CourseSearchService and this provider's id will end up in `eligible_training_provider_ids`.

PR #1346, opened 2020-05-01
[3295] - Update Course Create & Update Mailer

TobyRet:
### Context
As an... accredited body.
I need to... receive a notification when a PUBLISHED course I accredit has a basic detail change or a new course is published
So that... I can update my SRS (Student Record System) or contact the provider about the change.

Updates to course create and update mailer personalisation:

- remove the time period (AM/PM)
- add course description
- add course funding type

A further update will be required to the GOV.UK Notify templates to interpolate the new fields

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[Turn on accredited body notifications (basic detail + new course)](https://trello.com/c/BPnNcbBW/3295-turn-on-accredited-body-notifications-basic-detail-new-course)

app/mailers/course_create_email_mailer.rb
```diff
@@ -22,6 +24,6 @@ def create_course_url(course)
   end
 
   def format_create_datetime(datetime)
-    datetime.strftime("%-k:%M%P on %-e %B %Y")
+    datetime.strftime("%-k:%M on %-e %B %Y")
```
aliuk2012, line 27: I remember working on a similar issue https://github.com/DFE-Digital/teacher-training-api/pull/1237.

TBH, I think we should just have a standard global date/time format so that we do not have to think about date/time format every time we need to playback date/time to users.

aliuk2012: Actually looking back at that PR and the ticket I was the one that changed this formating from what it was and this was after a discussion between Paul G, Sam C and myself. I might still have the slack conversation, I'll see if I can find it.

AgaDufrat:
More of a product/content (cc: @izzwat) question but I'm wondering why we are removing `am/pm`. I remember that I always had to format time in that way in GDS. This style is also mentioned in https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style

aliuk2012:
https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times

TobyRet:
> More of a product/content (cc: @izzwat) question but I'm wondering why we are removing `am/pm`. I remember that I always had to format time in that way in GDS. This style is also mentioned in https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style

Currently it's using the 24 hour format, hence you do not require 'AM/PM'
I was not aware of any previous discussions on the format that @aliuk2012 mentions - will ask Izzy for clarification

spec/mailers/course_create_email_mailer_spec.rb
```diff
@@ -31,6 +31,14 @@
...
+      expect(mail.govuk_notify_personalisation[:course_description]).to eq(course.description)
+    end
+
+    it "includes the course funding type in the personalisation" do
+      expect(mail.govuk_notify_personalisation[:course_funding_type]).to eq(course.funding_type)
+    end
+
     it "includes the datetime for the creation in the personalisation" do
       expect(mail.govuk_notify_personalisation[:create_course_datetime]).to eq("4:05am on 3 February 2001")
     end
```
aliuk2012, line 44: Could you add a test for the afternoon time format? just so we don't end up questioning this again

TobyRet: Yep - now added to `time_format_spec.rb`

spec/lib/time_format_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "spec_helper"
+
+describe TimeFormat do
```
TobyRet, line 3: The specs were never running up until now

app/mailers/course_create_email_mailer.rb
TobyRet: Formatting fixed

PR #1347, opened 2020-05-01
[3345] add magic link tokens to users

misaka:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/sessions_controller.rb
+++ b/app/controllers/api/v2/sessions_controller.rb
@@ -2,7 +2,7 @@ module API
   module V2
     class SessionsController < API::V2::ApplicationController
       include ValidateJsonapiType
-      skip_before_action :check_terms_accepted, only: :create
+      skip_before_action :check_terms_accepted, only: %i[create create_by_magic]
 
       deserializable_resource :session
 
@@ -13,6 +13,31 @@ module API
           create_params.merge(
             last_login_date_utc: Time.now.utc,
           ),
+        )
+
+        record_first_login
+        send_welcome_email
+
+        render jsonapi: @current_user
+      end
+
+      def create_by_magic
+        skip_authorization
+
+        params.require(:magic_link_token)
+        if params[:magic_link_token] != @current_user.magic_link_token
+          return render status: :forbidden
+        end
+
+        magic_link_token_age = Time.zone.now - @current_user.magic_link_token_sent_at
+        if magic_link_token_age > Settings.magic_link.max_token_age.seconds
+          return render status: :bad_request
+        end
+
+        @current_user.update(
+          last_login_date_utc: Time.now.utc,
+          magic_link_token: nil,
+          magic_link_token_sent_at: nil,
         )
 
         record_first_login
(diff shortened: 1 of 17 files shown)


app/controllers/api/v2/sessions_controller.rb
```diff
@@ -21,6 +21,31 @@ def create
         render jsonapi: @current_user
       end
 
+      def create_by_magic
```
gpeng, line 24: It would be good to extract this into a service and/or `MagicLinkToken` class. Seems like a lot of steps and also some important logic that it would be good to have unit test coverage of.

spec/requests/api/v2/sessions/create_by_magic_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      user.update(first_login_date_utc: nil)
+
+      perform_request
+
+      user.reload
+      expect(user.first_login_date_utc).to be_within(4.seconds).of Time.now.utc
+    end
+
+    it "sends a welcome email" do
+      allow(SendWelcomeEmailService).to receive(:call)
```
gpeng, line 54: We should check `ActionMailer::Base.deliveries` here so as we test our full stack rather than stubbing our service.

spec/requests/api/v2/users/generate_and_send_magic_link_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  context "when unauthenticated" do
+    let(:payload) { { email: "foo@bar" } }
+
+    before do
+      perform_request
+    end
+
+    subject { response }
+
+    xit { should have_http_status(:ok) }
```
gpeng, line 34: These are being skipped

misaka: That's intentional, for now. There's a ticket in Trello to finish this work but I can't login to Trello at the moment.

spec/requests/api/v2/sessions/create_by_magic_spec.rb
misaka: Ya, I figured out how to do this for another test, forgot to back-port. Done.

app/controllers/api/v2/sessions_controller.rb
misaka: Another case of expediency / laziness. Can we do that as part of another ticket? Like the one about not returning an error if the user doesn't exist, which will touch on this bit of code.

gpeng: 👍 https://trello.com/c/vHEoXZQK/3359-m-add-magic-link-journey-for-non-existent-users

spec/requests/api/v2/users/generate_and_send_magic_link_spec.rb
gpeng: 👍

defong:
[Add support for magic link tokens to users](https://trello.com/c/pYR2Vjh2/3345-add-support-for-magic-link-tokens-to-users)

PR #1350, opened 2020-05-05
Fix missing diagram updates

aliuk2012:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/models/provider.rb
```diff
@@ -18,7 +18,8 @@
 #  longitude                        :float
 #  postcode                         :text
 #  provider_code                    :text
-#  provider_name                    :text
+#  provider_name                    :text             not null
+#  provider_name_search             :string           not null
```
aliuk2012, line 22: This line may have been added in error, just running annotate again

aliuk2012: Yip, local issue.

PR #1351, opened 2020-05-05
[3234] Expose `provider_id` in v2 Allocation

asmega:
### Context

- https://trello.com/c/DsJjDa09/3234-s-6b-pick-a-provider-page-searching-for-an-organisation-that-i-dont-already-accredit-no-js-autofill

- So clients can marry up allocations with their training providers

### Guidance to review

- create an allocation
- hit endpoint to grab all allocations
- json should include correct `provider_id`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/serializers/api/v2/serializable_allocation.rb
+++ b/app/serializers/api/v2/serializable_allocation.rb
@@ -2,7 +2,7 @@ module API
   module V2
     class SerializableAllocation < JSONAPI::Serializable::Resource
       type "allocations"
-      attributes :number_of_places, :request_type
+      attributes :number_of_places, :request_type, :provider_id
 
       belongs_to :accredited_body
       belongs_to :provider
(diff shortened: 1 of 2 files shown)


asmega:
[[S] 6b. "Pick a provider" page - searching for an organisation that I don't already accredit (no JS autofill)](https://trello.com/c/DsJjDa09/3234-s-6b-pick-a-provider-page-searching-for-an-organisation-that-i-dont-already-accredit-no-js-autofill)

app/serializers/api/v2/serializable_allocation.rb
```diff
@@ -2,7 +2,7 @@ module API
   module V2
     class SerializableAllocation < JSONAPI::Serializable::Resource
       type "allocations"
-      attributes :number_of_places, :request_type
+      attributes :number_of_places, :request_type, :provider_id
```
TobyRet, line 5: @asmega - is there any particular reason why you are exposing the `provider_id` here rather than accessing `allocation.provider.id`? The attributes exposed here should only be those that are not accessible via the associations.

asmega: yep totally. I'm closing this PR as its not needed

PR #1354, opened 2020-05-11
[3364] User getting 403 errors when viewing their training providers on the courses as an accredited body page

gpeng:
### Context

When an accredited body user tries to list courses for a provider that they are not directly associated with but that does deliver courses on behalf of the accredited body they are getting a 403 response in publish.

Publish currently calls the `ProvidersController#show` endpoint and includes the courses and the `ProviderPolicy#show?` policy does not allow this. Updating that policy will have undesirable side effects and will require complex coupling of with `Course`. So, we're going to switch `Publish` to use the `CourseController#index` endpoint to retrieve these courses. Unfortunately that also uses the `ProviderPolicy#show?` (via an alias) when the courses are scoped by provider. It also disregards any other filter params in that case.

* Don't use `ProviderPolicy` in the `CoursesController#index`
* Rely on the `CoursePolicy::Scope` to restrict access appropriately
* Remove the `can_list_courses?` policy from `ProviderPolicy'

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -55,14 +55,12 @@ module API
 
       def index
         authorize Course
-        if @provider.present?
-          authorize @provider, :can_list_courses?
-          scope = @provider.courses
-        else
-          scope = policy_scope(Course).kept
-          scope = scope.with_recruitment_cycle(@recruitment_cycle.year)
-          scope = scope.with_accredited_bodies(accredited_bodies) if accredited_bodies.present?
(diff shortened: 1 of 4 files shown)


gpeng:
[User getting 403 errors when viewing their training providers on the "courses as an accredited body" page](https://trello.com/c/OdnDymOC/3364-user-getting-403-errors-when-viewing-their-training-providers-on-the-courses-as-an-accredited-body-page)

PR #1358, opened 2020-05-13
[3299] Migrate allocation with codes

asmega:
### Context

- https://trello.com/c/rRYKmNry/3299-m-fix-last-years-allocation-on-data
- Imported allocation is not quite in the right state and need to be fixed
- Add extra info to perform migration for this and for future data

- Allocations now contain extra info: codes on top of ids and recruitment cycle reference
- Creating an allocation endpoint now persists this extra info
- Extract Allocation creation to a service
- Implement RecruitmentCycle#previous for convenience
- Add script to be run to migration imported allocations into correct state
- Script changes and add some data to recruitment cycle 1 so migration can occur 

### Guidance to review

- Creating allocation should continue to work, no api change. additional data should be set
- `AllocationMigrationService.new.execute` should migrate data into correct state with no errors

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


asmega:
[[M] Fix last years allocation on data](https://trello.com/c/rRYKmNry/3299-m-fix-last-years-allocation-on-data)

app/services/allocation_migration_service.rb
```diff
@@ -0,0 +1,45 @@
...
+      accredited_body = allocation.accredited_body
+      provider = allocation.provider
+
+      allocation.accredited_body_code = accredited_body.provider_code
+      allocation.provider_code = provider.provider_code
+
+      previous_accredited_body = previous_recruitment_cycle.providers.find_by!(provider_code: accredited_body.provider_code)
+
+      if previous_accredited_body.accrediting_provider == "not_an_accredited_body"
+        previous_accredited_body.update(accrediting_provider: "accredited_body")
```
defong, line 15: https://github.com/DFE-Digital/teacher-training-api/blob/10782c7cdf1cdd24c8fc2bf717b05a2027fb334f/app/models/provider.rb#L240-L247

what is the `provider_type`?

asmega: the ones being change in the migration are all `scitt`s

PR #1361, opened 2020-05-15
[3383] use previous allocation

asmega:
### Context

- https://trello.com/c/tZVaEdzD/3383-use-last-year-allocations
- Now that we have imported last years allocations we can start using them instead of dummy data

- Delete allocation import code as this has been done and no longer needed
- Added missing annotations
- Find and use number of places from last previous allocation

### Guidance to review

- Use with https://github.com/DFE-Digital/publish-teacher-training/pull/1099
- Request a repeat allocation
- This allocation should have same number_of_places as previous allocation and can be seen on the confirmation page

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


asmega:
[Use last year allocations](https://trello.com/c/tZVaEdzD/3383-use-last-year-allocations)

app/models/allocation.rb
```diff
@@ -19,13 +22,23 @@
...
-  validate :accredited_body_is_an_accredited_body
   validates :number_of_places, numericality: true
 
+  validate :accredited_body_is_an_accredited_body
+
   enum request_type: { initial: 0, repeat: 1, declined: 2 }
 
+  def previous
+    Allocation.find_by(
+      provider_code: provider_code,
```
gpeng, line 36: Can we add indexes for these three columns?

asmega: done

PR #1364, opened 2020-05-18
[3360] - Prioritise accredited body's own courses in search results

TobyRet:
### Context
- When users search for an accredited body who delivers its own courses _and_ has partners who delivers courses on it's behalf, then the accredited body's courses should be prioritised in the results .

- Scope added to prioritise university courses in results where appropriate.
- There appears to be a modest 30ms decrease in `db_runtime` (avg) as a result of the changes

warwick

### Guidance to review
- Do a provider search for 'University of Warwick' - you can see that its courses are now returned at the beginning of the results, other courses are returned thereafter alphabetically by provider name.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -148,6 +148,10 @@ class Course < ApplicationRecord
     joins(:provider).merge(Provider.by_name_descending).order("name desc")
   end
 
+  scope :accredited_body_order, ->(provider_name) do
+    joins(:provider).merge(Provider.by_provider_name(provider_name))
+  end
+
   scope :changed_since, ->(timestamp) do
     if timestamp.present?
       changed_at_since(timestamp)
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -13,19 +13,6 @@ class CourseSearchService
 
   def call
     scope = course_scope.findable
-
-    if sort_by_provider_ascending?
-      scope = scope.ascending_canonical_order
-      scope = scope.select("provider.provider_name", "name")
-    elsif sort_by_provider_descending?
-      scope = scope.descending_canonical_order
-      scope = scope.select("provider.provider_name", "name")
-    elsif sort_by_distance?
-      scope = scope.joins(sites_with_distance_from_origin)
-      scope = scope.select("course.*, distance")
-      scope = scope.order(:distance)
-    end
-
     scope = scope.with_salary if funding_filter_salary?
     scope = scope.with_qualifications(qualifications) if qualifications.any?
     scope = scope.with_vacancies if has_vacancies?
@@ -36,7 +23,28 @@ class CourseSearchService
     scope = scope.within(filter[:radius], origin: origin) if locations_filter?
     scope = scope.with_funding_types(funding_types) if funding_types.any?
 
-    scope.distinct
+    # The 'where' scope will remove duplicates
+    # An outer query is required in the event the provider name is present.
+    # This prevents 'PG::InvalidColumnReference: ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list'
+    outer_scope = Course.where(id: scope.select(:id))
+
+    if provider_name.present?
+      outer_scope = outer_scope
+                      .accredited_body_order(provider_name)
+                      .ascending_canonical_order
+    elsif sort_by_provider_ascending?
+      outer_scope = outer_scope.ascending_canonical_order
+      outer_scope = outer_scope.select("provider.provider_name", "course.*")
+    elsif sort_by_provider_descending?
+      outer_scope = outer_scope.descending_canonical_order
+      outer_scope = outer_scope.select("provider.provider_name", "course.*")
+    elsif sort_by_distance?
+      outer_scope = outer_scope.joins(sites_with_distance_from_origin)
+      outer_scope = outer_scope.select("course.*, distance")
+      outer_scope = outer_scope.order(:distance)
+    end
+
+    outer_scope
   end
 
   private_class_method :new
--- a/spec/services/course_search_service_spec.rb
+++ b/spec/services/course_search_service_spec.rb
@@ -7,8 +7,10 @@ describe CourseSearchService do
       let(:filter) { {} }
 
       it "defaults to Course" do
-        expect(Course).to receive_message_chain(:findable, :distinct).and_return(findable_scope)
-        expect(subject).to eq(findable_scope)
+        expect(Course).to receive_message_chain(:findable).and_return(select_scope)
+        expect(select_scope).to receive(:select).and_return(inner_query_scope)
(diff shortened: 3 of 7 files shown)


app/services/course_search_service.rb
```diff
@@ -26,17 +41,9 @@ def call
...
-    scope = scope.with_subjects(subject_codes) if subject_codes.any?
-    scope = scope.with_provider_name(provider_name) if provider_name.present?
-    scope = scope.with_send if send_courses_filter?
-    scope = scope.within(filter[:radius], origin: origin) if locations_filter?
-    scope = scope.with_funding_types(funding_types) if funding_types.any?
-
-    scope.distinct
+    scope
+    #  For some reason the api breaks if the 'scope.distinct' remains :-()
+    # scope.distinct
```
TobyRet, line 46: ??
```
Cause: PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: ..., $9, $10, $11, $12, $13, $14, $15, $16) ORDER BY CASE WHEN ...
```

TobyRet: https://stackoverflow.com/questions/12693089/pgerror-select-distinct-order-by-expressions-must-appear-in-select-list

defong: its needed because of 

https://github.com/DFE-Digital/teacher-training-api/blob/826b00d4a6331cb6a36a7dfd0ed8332c702b0705/app/services/course_search_service.rb#L15

app/models/course.rb
```diff
@@ -183,6 +183,10 @@ def new_draft_attributes
     joins(:provider).merge(Provider.by_name_descending).order("name desc")
   end
 
+  scope :accrediting_body_order, ->(provider_name) do
```
defong, line 186: `ing`  vs `id`

app/services/course_search_service.rb
defong: fyi 

either approach it like
https://github.com/DFE-Digital/search-and-compare-api/blob/3aa6be82dcf5548892580e900ea1e37decf63a0f/src/api/DatabaseAccess/CourseDbContext.cs#L121-L127

in either case sub querying it will be easier

ie
```sql

select * from course where id in (select id from course where filter)
```

and move the ordering/distinct etc  to the outer query

misaka: @TobyRet If the Stackoverflow Q you linked-to is correct, maybe you need to do:

```ruby
    scope = scope.distinct
```

before all the ordering.

TobyRet: No unfortunately this did not work. An outer query had to be created instead.

spec/services/course_search_service_spec.rb
```diff
@@ -7,19 +7,23 @@
       let(:filter) { {} }
 
       it "defaults to Course" do
-        expect(Course).to receive_message_chain(:findable, :distinct).and_return(findable_scope)
-        expect(subject).to eq(findable_scope)
+        expect(Course).to receive_message_chain(:findable).and_return(course_ids_scope)
```
TobyRet, line 10: As we are not longer calling `distinct` there are lots of changes in this spec. I'm not a huge fan of RSpec 'shared examples' but perhaps there is a case for using them here given the repetition.

TobyRet:
[[M] List university's own courses before partner courses](https://trello.com/c/emlW2vMN/3360-m-list-universitys-own-courses-before-partner-courses)

app/services/course_search_service.rb
```diff
@@ -36,7 +23,34 @@ def call
     scope = scope.within(filter[:radius], origin: origin) if locations_filter?
     scope = scope.with_funding_types(funding_types) if funding_types.any?
 
-    scope.distinct
+    # The 'where' scope will remove duplicates
+    # An outer query is required in the event the provider name is present.
+    # This prevents 'PG::InvalidColumnReference: ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list'
+    outer_scope = Course.where(id: scope.pluck(:id))
```
misaka, line 29: Would it worth using `scope.select(:id)` here instead of `pluck`? I'm thinking Postgresql might be able to optimise the sub-query if we keep it all in one query.

```suggestion
    outer_scope = Course.where(id: scope.select(:id))
```

app/services/course_search_service.rb
```diff
@@ -36,7 +23,34 @@ def call
...
+    if sort_by_provider_descending?
+      outer_scope = outer_scope.descending_canonical_order
+      outer_scope = outer_scope.select("provider.provider_name", "course.*")
+    end
+
+    if sort_by_distance?
+      outer_scope = outer_scope.joins(sites_with_distance_from_origin)
+      outer_scope = outer_scope.select("course.*, distance")
+      outer_scope = outer_scope.order(:distance)
+    end
```
misaka, line 51: Why not use `elsif` as before? These should mutually exclusive so it'd be nice to make that clear.

app/services/course_search_service.rb
TobyRet: Done

fofr:
I expected an explicit test somewhere to say "when searching by provider who is an accredited body, it shows that provider first"

app/services/course_search_service.rb
misaka: Just for future reference, it looks like `pluck` can be faster than `select` for small (around 30 or less) result-sets. I tested a few scenarios:

**Searching for all Universtity of Warwick (W20) courses on Find**
courses returned: `16`
average select db time increase: `+95.26ms`
average select total time increase: `+157.80ms`

**Searching for all University College London (U80) courses on Find**
courses returned: `28`
average select db request time increase: `+49.28ms`
average select total request time increase: `-30.3ms`

**Searching for all Primary courses on Find**
courses returned: `1,580`
average select db time increase: `-103.7ms`
average select total time increase: `-400.40ms`

(lower times mean `select` is faster than `pluck`)

I think the gains from speeding up larger results is probably worth more than the losses from slowing down smaller ones, but we'd probably need to do a proper study of the distribution of results we end up returning is, and the speed-up/slow-down we'd get, would be needed.

TobyRet:
> I expected an explicit test somewhere to say "when searching by provider who is an accredited body, it shows that provider first"

Apologies the descriptions were request spec were rubbish. Have updated. Hopefully that's clearer now ...
https://github.com/DFE-Digital/teacher-training-api/pull/1364/files#diff-230985fb50f1d3b01a7b55d41484ee88R519

spec/services/course_search_service_spec.rb
defong: you probably could of left it as as it `scope` and the `outer_scope` does the ordering etc now...

PR #1365, opened 2020-05-18
3368 top level stats endpoint publish

defong:
### Context
vital pub stats

Add pub status

### Guidance to review
http://localhost:3001/reporting

```json
{
  "providers": { ... },
  "courses": { ... },
 "publish": {
    "users": {
      "total": {
        "all": 3022,
        "active_users": 1852,
        "non_active_users": 1170
      },
      "recent_active_users": 468
    },
    "providers": {
      "total": {
        "all": 1939,
        "providers_with_non_active_users": 1314,
        "providers_with_recent_active_users": 625
      },
      "with_1_recent_active_users": 539,
      "with_2_recent_active_users": 65,
      "with_3_recent_active_users": 9,
      "with_4_recent_active_users": 1,
      "with_more_than_5_recent_active_users": 11
    },
    "courses": {
      "total_updated_recently": 1604,
      "updated_non_findable_recently": 35,
      "updated_findable_recently": 1569,
      "updated_open_courses_recently": 671,
      "updated_closed_courses_recently": 898,
      "created_recently": 68
    }
  }
}
```
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  class << self
+    def call(recruitment_cycle_scope:)
+      new(recruitment_cycle_scope: recruitment_cycle_scope).call
+    end
+  end
+
+  def call
+    days_ago = 30.days.ago
```
misaka, line 14: Code Climate is complaining that this method is over 25 lines long. I was on the fence about mentioning it until I saw the specs for the scopes below ... I'd break these values out into methods and separate the logic and tests out from the presentation here.

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def active_users_last_30_days_count
+    @active_users_last_30_days_count ||= active_users_last_30_days.count
+  end
+
+  def provider_count
+    @provider_count ||= @providers.count
+  end
+
+  def user_activites
```
misaka, line 68: How about calling this `active_user_breakdown`?

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,116 @@
...
+    {
+      total: {
```
misaka, line 70: Don't need the `total` here do we?

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,116 @@
...
+        all: user_count,
+        active_users: active_users_count,
+        non_active_users: user_count - active_users_count,
+        active_users_last_30_days: active_users_last_30_days_count,
+      },
+    }
+  end
+
+  def provider_activites
```
misaka, line 79: `provider_breakdown`

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,116 @@
...
+    @provider_count ||= @providers.count
+  end
+
+  def user_activites
+    {
+      total: {
+        all: user_count,
+        active_users: active_users_count,
+        non_active_users: user_count - active_users_count,
+        active_users_last_30_days: active_users_last_30_days_count,
```
misaka, line 74: Move this last one out of `total`?

app/models/provider.rb
```diff
@@ -120,6 +120,10 @@ def current_accredited_courses
       .or(self.where(provider_code: Course.findable.select(:accrediting_provider_code)))
   end
 
+  scope :with_users, ->(users) do
+    joins(:users).merge(users)
+  end
```
misaka, line 125: I think that this scope does a `join`, which potentially ends up returning multiple identical providers, is surprising. I'd leave it at `merge(users)`, or maybe just skip the scope altogether since even having `merge(users)` here is slightly odd as it requires an AREL `users`.

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def providers_with_active_users_distinct_count
+    @providers_with_active_users_distinct_count ||= providers_with_active_users.distinct.count
+  end
+
+  def grouped_providers_with_x_active_users
+    @grouped_providers_with_x_active_users ||= providers_with_active_users.group(:id)
+      .order(count_id: :desc)
+      .count(:id)
+      .group_by(&:second)
```
misaka, line 49:
```ruby
def recent_active_user_count_by_provider
  @recent_active_user_count_by_provider ||= active_users_last_30_days
    .joins(:provider)          # Results include a user entry for _each_ matching provider
    .merge(@provider)          # Limit our scope to the current recruitment Cycle
    .group("provider.id")
    .count                     # Count the users for each provider
end

def grouped_providers_with_x_active_users
  @grouped_providers_with_x_active_users ||= recent_active_user_count_by_provider
    .group_by(&:second)        # Group the results by the number of users they have
    .transform_values(&:count) # Count the results
end
```

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,116 @@
...
+  def active_users_last_30_days
+    @active_users_last_30_days ||= active_users.last_login_since(days_ago)
+  end
+
+  def providers_with_active_users
+    @providers_with_active_users ||= @providers.with_users(active_users_last_30_days)
+  end
+
+  def providers_with_active_users_distinct_count
+    @providers_with_active_users_distinct_count ||= providers_with_active_users.distinct.count
```
misaka, line 42:
```ruby
def providers_with_active_users_distinct_count
  @providers_with_active_users_distinct_count ||= @providers
    .joins(:users)
    .merge(active_users_last_30_days)
    .distinct
    .count
end

Comment (misaka) [reply app/services/publish_reporting_service.rb:42]: Can we get rid of the `distinct` here?

File: spec/services/publish_reporting_service_spec.rb [hunk 9]
```diff
@@ -0,0 +1,120 @@
...
+        updated_closed_courses_last_30_days: 25,
+        created_last_30_days: 10,
+      },
+    }
+  end
+
+  let(:recruitment_cycle_scope) do
+    instance_double(RecruitmentCycle)
+  end
+  let(:courses_scope) { class_double(Course) }
```
misaka, line 40:
```suggestion
  let(:courses_scope) { Course::ActiveRecord_Relation }
```

defong: `uninitialized constant Course::ActiveRecord_Relation`

app/services/publish_reporting_service.rb
defong: > Can we get rid of the `distinct` here?

no

app/models/provider.rb
defong: removed inlined altogether

app/services/publish_reporting_service.rb
defong: filped

spec/services/publish_reporting_service_spec.rb
misaka: Remember we were debating just using `Course.all`? Just use that.

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  def call
+    {
+      users: user_breakdown,
+      providers: provider_breakdown,
+      courses: course_breakdown,
+    }
+  end
+
+  private_class_method :new
```
misaka, line 21: Why be strict about this? It's not like `self.call` does anything magical.

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,123 @@
...
+        all: provider_count,
+        providers_with_non_active_users: (provider_count - providers_with_recent_active_users_distinct_count),
+        providers_with_recent_active_users: providers_with_recent_active_users_distinct_count,
+      },
+
+      with_1_recent_active_users: grouped_providers_with_x_active_users[1] || 0,
+      with_2_recent_active_users: grouped_providers_with_x_active_users[2] || 0,
+      with_3_recent_active_users: grouped_providers_with_x_active_users[3] || 0,
+      with_4_recent_active_users: grouped_providers_with_x_active_users[4] || 0,
+      with_more_than_5_recent_active_users: grouped_providers_with_x_active_users.keys.excluding(4, 3, 2, 1).sum { |k| grouped_providers_with_x_active_users[k] || 0 },
```
misaka, line 98: nnnggghhh ... can we un-inline plz

app/services/publish_reporting_service.rb
```diff
@@ -0,0 +1,123 @@
...
+    }
+  end
+
+  def course_breakdown
+    courses_changed_at_since = @courses.changed_at_since(days_ago)
```
misaka, line 103:
```suggestion
    courses_recently_changed = @courses.changed_at_since(days_ago)
```

defong:
https://gist.github.com/defong/94b397bfd0f1bbfc30871c6d8cd16ffa

app/services/publish_reporting_service.rb
defong: so why not strict? why do you want to hold on to an instance?

PR #1366, opened 2020-05-18
[3416] change healthcheck to return service unavailable

misaka:
### Context

Meant to fix the Sidekiq healthcheck, but can't currently find evidence of it being broken.

Change `502 bad gateway` to `503 service unavailable`.

### Guidance to review

Another PR will be made if we can find anything to fix in the Sidekiq healthcheck.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


misaka:
[sidekiq healthcheck is returning false](https://trello.com/c/0KiyP0hW/3416-sidekiq-healthcheck-is-returning-false)

PR #1369, opened 2020-05-19
[3393] Update api notifications endpoint

gpeng:
### Context

We are adding a feature to publish to allow a user to enable/disable email notifications for courses that they are associated with (currently by being a member of an accredited body). 

The user can only enable/disable all notifications and we are not providing granularity in the preferences currently.

Add an API endpoint to allow a user to set and retrieve their preferences.

`GET /user_notification_preferences/:user_id` to retrieve
`PUT /user_notification_preferences/:user_id` to create/update

There is a previous implementation of this functionality but we were exposing too much complexity in the response which causes a lot of complexity in the client app. This is a second attempt with the intention of handling that complexity in the API.

The previous implementation will be removed in a separate PR. (https://github.com/DFE-Digital/teacher-training-api/pull/1370)

The API returns a simple response with only two fields:

* `enabled`
* `updated_at` - this is either the most recent updated at from the user's `UserNotification` records or nil if they haven't set their preferences before. We need to differentiate between these states in the frontend to provide alternative UI and this seemed like the simplest approach.

The `UserNotificationPreferences` object is not stored in the DB directly so is keyed by the user's ID which we expose as `id`.

Preferences are stored as `UserNotification` records with one record per accredited body. This is a bit of a hangover from the user research and is more complex than we require currently. We've chosen to work with that data model for now but to abstract the complexity away. 

When a user updates their preference we clear the previous records and recreate which gives an opportunity for the list to be kept in sync with the users provider associations should they have changed.

### Guidance to review

Publish has not been updated to use this endpoint yet so it's not possibly to product review until that PR is opened. We have tested with a prototype version and all seems to be well in the world.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


gpeng:
[Update API /notifications endpoint](https://trello.com/c/GBpxivpg/3393-update-api-notifications-endpoint)

aliuk2012:
❤️ Good idea 👏 
> When a user updates their preference we clear the previous records and recreate which gives an opportunity for the list to be kept in sync with the users provider associations should they have changed.

app/models/user_notification_preferences.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    user_notifications.maximum(:updated_at).iso8601
+  end
+
+  def update(enable_notifications:)
+    UserNotification.transaction do
+      rollback_on_error do
+        UserNotification.where(user_id: id).destroy_all
+
+        user.providers.accredited_body.map do |provider|
```
TobyRet, line 26: Probably not for this PR but the `providers.accredited_body` is a little confusing. Can this return more than one accrediting body?

TobyRet: I see you are destroying and creating rather than updating in the pure sense. Was there a particular reason for this?

app/controllers/api/v2/user_notification_preferences_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      deserializable_resource :user_notification_preferences,
+                              class: API::V2::DeserializableUserNotificationPreferences
+
+      def show
+        authorize user_notification_preferences = UserNotificationPreferences.new(user_id: params[:id])
+
+        render jsonapi: user_notification_preferences
+      end
+
+      def update
```
TobyRet, line 13: I agree with the point in the PR description about trying to remove complexity from the front end. This version does make things simpler, but has been agreed that granular notification updates won't be a thing on the near horizon?

app/models/user_notification_preferences.rb
gpeng: Yeah.. it is but it's an existing scope. We'll have a look how much it would take to change. There's also another issue with this code as we get duplicates in some cases if the user is related to more than one provider associated with the same accredited body. We'll update 👍

app/controllers/api/v2/user_notification_preferences_controller.rb
gpeng: I don't think it has been agreed and may well likely happen. I think we should go with the requirements we know though and I don't think it will be too difficult to change to support granular permissions. The `UserNotificationPreferences` resource will still be singular but have collections of create and update permissions or something like that. I think in many ways it will be easier to handle adding further granularity with this approach e.g. if we were to add notifications that a user has been created or something totally unrelated to courses. That may end up a totally different set of tables etc on the backend so would require a co-ordinating class to represent the total preferences.

app/models/user_notification_preferences.rb
gpeng: Turns out that's a scope from an enum. So an alternative would be `user.providers.where(accrediting_provider: "accredited_body")` which is to my eye even worse 😆 We've left as is for now but added a `distinct` to weed out the duplicates. Hopefully that also makes it a bit more obvious that it's not singular.

PR #1372, opened 2020-05-19
[3231] Add update service to allocations

SamJamCul:
### Context

https://trello.com/c/mC7AqJvX/3231-s-11-change-my-allocations-request

* Add update service to set number of places appropriately for allocation

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/services/allocations/update.rb
+++ b/app/services/allocations/update.rb
@@ -1,0 +1,36 @@
+module Allocations
+  class Update
+    attr_accessor :object, :params
+
+    def initialize(allocation, params)
+      @object = allocation
+      @params = params
+
+      set_number_of_places
+      set_request_type
+
+      object
+    end
+
+    def execute
+      object.save
+    end
+
+  private
+
+    def set_number_of_places
+      object.number_of_places = case params[:request_type]
+                                when "declined"
+                                  0
+                                when "repeat"
+                                  object.previous&.number_of_places || 0
+                                when "initial"
+                                  params[:number_of_places]
+                                end
+    end
+
+    def set_request_type
+      object.request_type = params[:request_type]
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


SamJamCul:
[[S] 11. Change my allocations request](https://trello.com/c/mC7AqJvX/3231-s-11-change-my-allocations-request)

app/services/allocations/update.rb
```diff
@@ -0,0 +1,36 @@
...
+  private
+
+    def set_number_of_places
+      object.number_of_places = case params[:request_type]
+                                when "declined"
+                                  0
+                                when "repeat"
+                                  object.previous&.number_of_places || 0
+                                else
+                                  params[:number_of_places]
```
defong, line 28: so if its not

`declined` or `repeat` what is is it?

https://github.com/DFE-Digital/teacher-training-api/blob/3dbee3b7ef813cd7006927e9c7a75315d2b0d787/spec/factories/allocations.rb#L7-L12

Feels head scratching it is missing a trait as well as a test for the other `request_type`

app/services/allocations/update.rb
```diff
@@ -0,0 +1,36 @@
...
+      @params = params
+
+      set_number_of_places
+      set_request_type
+
+      object
+    end
+
+    def execute
+      object.save
```
defong, line 16: why isn't this return value tested?

app/services/allocations/update.rb
SamJamCul: The third option for request_type is `initial` but we're going to be addressing this in another PR. Updating an initial request to set a different number of places is going to require some further logic here, and we aren't quite sure yet what it'll look like.

app/services/allocations/update.rb
SamJamCul: it it now, thanks

PR #1374, opened 2020-05-20
[3422] Provider search can now include self

asmega:
### Context

- https://trello.com/c/Hwg5ZBqT/3422-accredited-bodies-should-be-able-to-request-pe-allocations
- For allocations we need to return accrediting providers if they also have a PR fee funded course

- on accredited provider search if the accredited provider has a course
themselves that meets the criteria the accredited provider is returned
in the results
- previously only training providers would be considered

### Guidance to review

- Navigate to a provider with their own fee funded PE course eg `2AT`
- go to the index allocations page
- should see yet to request repeat allocation

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
+++ b/spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
@@ -67,7 +67,7 @@ describe "AccreditedBody API v2", type: :request do
 
       let(:filters) { "" }
 
-      it "is returned" do
+      it "returns the accredited provider in the results" do
         get request_path, headers: { "HTTP_AUTHORIZATION" => credentials }
         json_response = JSON.parse(response.body)
         provider_hashes = json_response["data"]
--- a/spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
+++ b/spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
@@ -60,7 +60,7 @@ describe "AccreditedBody API v2", type: :request do
       get request_path, headers: { "HTTP_AUTHORIZATION" => credentials }
     end
 
-    context "when accredited_provider has a course" do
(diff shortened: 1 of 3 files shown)


asmega:
[Accredited bodies should be able to request PE allocations](https://trello.com/c/Hwg5ZBqT/3422-accredited-bodies-should-be-able-to-request-pe-allocations)

TobyRet:
[[M] List university's own courses before partner courses](https://trello.com/c/emlW2vMN/3360-m-list-universitys-own-courses-before-partner-courses)

spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
```diff
@@ -51,6 +60,21 @@ def perform_request
...
     end
 
+    context "when accredited_provider has a course" do
+      before do
+        accredited_provider_course
+      end
+
+      let(:filters) { "" }
+
+      it "is returned" do
```
misaka, line 70: I know this is nit-picky, but I feel like you and I understand this, but it won't be obvious at all to others. Also, `it` in this context is the response from `GET /providers`. Could we make this a little more explicit:

```suggestion
      it "returns the accredited provider in the results" do
```

spec/requests/api/v2/providers/accredited_body_filter_providers_spec.rb
```diff
@@ -51,6 +60,21 @@ def perform_request
       get request_path, headers: { "HTTP_AUTHORIZATION" => credentials }
     end
 
+    context "when accredited_provider has a course" do
```
misaka, line 63: Similar to the comment about the `it "is returned"`, could we just be a little more clear here:

```suggestion
    context "when accredited_provider has a self-accredited course" do
```

PR #1377, opened 2020-05-22
[3429] Import missed allocations

asmega:
### Context

- https://trello.com/c/x2El42cz/3429-import-additional-allocations
- There were missed allocations that were submitted after the deadline and were not in the published spreadsheet

- Reintroduce allocation import script
- Add missed allocation data for import
- Introduce uniqueness index to prevent duplicate allocations from being created

### Guidance to review

- Test against most recent db dump
- Run the following in a rails console
- `service = AllocationImporterService.new(path_to_csv: Rails.root.join('config/data/itt_allocations_2020_to_2021/missed_physical_education.csv')); service.execute`
- New or updated allocations should be present

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/allocation_importer_service.rb
+++ b/app/services/allocation_importer_service.rb
@@ -1,0 +1,46 @@
+require "csv"
+
+class AllocationImporterService
+  attr_reader :path_to_csv
+
+  def initialize(path_to_csv:)
+    @path_to_csv = path_to_csv
+  end
+
+  def execute
+    rows.each do |row|
+      training_provider = Provider.where(recruitment_cycle_id: 1).find_by(provider_code: row["provider_code"])
+      accredited_body_provider = Provider.where(recruitment_cycle_id: 1).find_by(provider_code: row["accredited_body_code"])
+
+      if training_provider.blank?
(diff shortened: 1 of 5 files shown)


asmega:
[[S] Import additional allocations](https://trello.com/c/x2El42cz/3429-import-additional-allocations)

app/services/allocation_importer_service.rb
```diff
@@ -0,0 +1,46 @@
+require "csv"
+
+class AllocationImporterService
+  attr_reader :path_to_csv
+
+  def initialize(path_to_csv:)
+    @path_to_csv = path_to_csv
+  end
+
+  def execute
```
misaka, line 10: Aren't we migrating to using `call`? @gpeng was going to start enforcing this I thought. 😛

asmega: I prefer `call`, README says to use `execute` so maybe we should change the README

gpeng: We are/were but haven't as yet so I let it slip. I've seen this @asmega flavour elsewhere so it's currently as consistent as anything else 😆

PR #1378, opened 2020-05-22
[3399] - Allow provider searches with queries containing 2 characters

TobyRet:
### Context
On Publish and Find we now want to allow users to run provider searches containing 2 characters. 
NB: Hold fire on merging until front ends are updated

- Changes the provider search query minimum character limit to  1
- Adds missing test coverage

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/providers_controller.rb
+++ b/app/controllers/api/v3/providers_controller.rb
@@ -7,6 +7,8 @@ module API
         build_fields_for_index
 
         @providers = if params[:search].present?
+                       return render(status: :bad_request) if params[:search].length < 2
+
                        @recruitment_cycle.providers.search_by_code_or_name(params[:search])
                      else
                        @recruitment_cycle.providers
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -51,7 +51,7 @@ module API
       def suggest
         authorize Provider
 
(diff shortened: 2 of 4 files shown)


TobyRet:
[[S] - Enable searching for provider with only 2 character query](https://trello.com/c/i9LJS3Jn/3399-s-enable-searching-for-provider-with-only-2-character-query)

app/controllers/api/v3/providers_controller.rb
```diff
@@ -7,7 +7,9 @@ def index
         build_fields_for_index
 
         @providers = if params[:search].present?
-                       @recruitment_cycle.providers.search_by_code_or_name(params[:search])
+                       return render(status: :bad_request) if params[:search].length < 2
+
+                       @recruitment_cycle.providers.search_by_code_or_name(params[:search]).limit(5)
```
asmega, line 12: why was a limit of 5 added?

TobyRet: Ah - that should not be there. I'll remove.

misaka:
I was testing this out in the review app, and was still getting a 500 when I tried searching with 2 characters (see [Sentry PUBLISH-TEACHER-TRAINING-74](https://sentry.io/organizations/dfe-bat/issues/1692602858/?environment=qa&project=1407453&query=is%3Aunresolved)).

I guess this just because the review app uses the qa api, which still needs 3+ chars?

PR #1379, opened 2020-05-26
[3416] fix queues checking

misaka:
### Context

Made a mistake in how queues were processed. 😢 Goes to show that this test is a bit tricky and brittle ... my issue locally was being masked by the fact that I have jobs in the queue that we're no longer processing.

Fix this. Hopefully. Also, add some docs.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,4 +294,8 @@ Find (Publish Teacher Training & Teacher Training API) build and release process
 
 ### <a name="releases-triggering"></a>Deploying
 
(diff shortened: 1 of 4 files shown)


misaka:
[sidekiq healthcheck is returning false](https://trello.com/c/0KiyP0hW/3416-sidekiq-healthcheck-is-returning-false)

PR #1381, opened 2020-05-27
3452 Duplicate usernotification records are being created

gpeng:
### Context

When updating users notification preferences we found that the `User#providers.accredited_body` scope was sometimes returning duplicates. We added `distinct` to filter them out but in some cases (which I haven't been able to recreate in a test but see locally), the duplicates still occurred.

Update the query to retrieve the user's accredited body list by adding `pluck` which seems to make `distinct` actually be distinct.

### Guidance to review

Difficult to product review other than checking the notification preferences still work.

### Checklist
 
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


gpeng:
[Duplicate UserNotification records are being created](https://trello.com/c/9AbDW65V/3452-duplicate-usernotification-records-are-being-created)

app/models/user_notification_preferences.rb
```diff
@@ -23,12 +23,12 @@ def update(enable_notifications:)
       rollback_on_error do
         UserNotification.where(user_id: id).destroy_all
 
-        user.providers.accredited_body.distinct.map do |provider|
+        user.providers.accredited_body.distinct.pluck(:provider_code).each do |provider_code|
```
defong, line 26:
``` ruby
user.providers.accredited_body.where(recruitment_cycle: ....current)
```

gpeng: Ah yes... good spot. I've updated 👍

app/models/user_notification_preferences.rb
```diff
@@ -43,6 +43,10 @@ def update(enable_notifications:)
 
 private
 
+  def user_accredited_body_codes
+    user.providers.accredited_body.in_current_cycle.distinct.pluck(:provider_code)
```
defong, line 47:
```suggestion
    user.providers.accredited_body.in_current_cycle.pluck(:provider_code)
```

https://github.com/DFE-Digital/teacher-training-api/blob/e45a0005585e96d53dd3de3c7c60f9ec1b0e2bb7/db/schema.rb#L243

gpeng: I think we still need the distinct as the user even though the provider record has that constraint, it's possible for the user to be associated with the same provider through more than one organisation so this association can return duplicates.

defong: fair call :+1:

PR #1387, opened 2020-05-29
3468 swagger clean up follow up

defong:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


config/routes.rb
```diff
@@ -5,7 +5,14 @@
   get :sha, controller: :heartbeat
   get :reporting, controller: :reporting
 
+  # NOTE: this has questionable existance.
```
defong, line 8: do we actually need this?

config/routes.rb
```diff
@@ -5,7 +5,14 @@
...
+  # Why do we need ui?
+  # When we have that https://github.com/alphagov/tech-docs-gem
   mount OpenApi::Rswag::Ui::Engine => "/api-docs"
+  # NOTE: this has questionable existance.
+  # Why do we need api?
+  # When it seems to be a static file server...
+  # That only works after `bundle exec rake rswag:specs:swaggerize`
```
defong, line 15: if we dont actual use the ui is there away to 

`bundle exec rake rswag:specs:swaggerize`
which uses really just generates `swagger/public_v1/api_spec.json`
and then server is via `/api-docs/public_v1/api_spec.json`

config/routes.rb
asmega: I don't think this is needed any more. originally we had an open api spec but no way of presenting it, this was then added to fit that space. However it needed to be gov branded so was not fit for purpose. This is now presented thru the tech docs gem

asmega:
the placeholder code is needed as i believe some tests depend on it. you can try deleting and see if the tests still pass

config/routes.rb
asmega: the command needs to be fired at some point to generate the new open api spec if there have been any changes. this could be made as part of the build process

defong:
[Swagger clean up](https://trello.com/c/mvSP3p8a/3468-swagger-clean-up)

PR #1388, opened 2020-05-29
[3440] Allocations can now be filtered

asmega:
### Context

- https://trello.com/c/dtbw2raY/3440-stop-filtering-out-providers-from-provider-search-initial-allocation-request
- The frontend needs to to determine what to show the user depending on whether or not an allocation exists
- This change modifies the allocation endpoints so the results can be filtered

- Can now pass training_provider_code to find allocations only
associated with the specified training provider
- allocations#show handles `include` to return associations

### Guidance to review

```ruby
Allocation.where(recruitment_cycle_year: recruitment_cycle.year)
  .where(provider_code: provider.provider_code, training_provider_code: training_provider[:provider_code])
   .first
```

- Using the above code in publish given an allocation exists for the training provider it should be returned otherwise should return nil

```
 Allocation.include(:provider).find(params[:id]).first
```

- This should now return the provider association

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/allocations_controller.rb
+++ b/app/controllers/api/v2/allocations_controller.rb
@@ -7,7 +7,13 @@ module API
       def index
         authorize Allocation
 
-        render jsonapi: policy_scope(Allocation.where(accredited_body_id: accredited_body.id)),
+        scope = Allocation.where(accredited_body_id: accredited_body.id)
+
+        if params[:filter] && params[:filter][:training_provider_code]
+          scope = scope.where(provider_code: params[:filter][:training_provider_code])
+        end
+
+        render jsonapi: policy_scope(scope),
                include: params[:include],
                status: :ok
       end
@@ -15,9 +21,7 @@ module API
       def show
         authorize @allocation = Allocation.find(params[:id])
 
-        render jsonapi: @allocation,
-          include: params[:include],
-          status: :ok
+        render jsonapi: @allocation, include: params[:include], status: :ok
       end
 
       def create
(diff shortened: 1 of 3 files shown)


asmega:
[Stop filtering out providers from provider search - initial allocation request](https://trello.com/c/dtbw2raY/3440-stop-filtering-out-providers-from-provider-search-initial-allocation-request)

app/controllers/api/v2/allocations_controller.rb
```diff
@@ -7,15 +7,21 @@ class AllocationsController < API::V2::ApplicationController
       def index
         authorize Allocation
 
-        render jsonapi: policy_scope(Allocation.where(accredited_body_id: accredited_body.id)),
+        scope = Allocation.where(accredited_body_id: accredited_body.id)
+
+        if params[:filter] && params[:filter][:training_provider_code]
+          scope = scope.where(provider_code: params[:filter][:training_provider_code])
+        end
```
TobyRet, line 14: I'm finding this a little confusing ...

In the PR guidance it states the following 
```
Allocation.where(recruitment_cycle_year: recruitment_cycle.year)
  .where(provider_code: provider.provider_code, training_provider_code: training_provider[:provider_code])
   .first
```

But in the if statement we are passing `provider_code: [training_provider_code]` ?

I may be missing something here ...

asmega: no. because in publish the `provider_code` is the accredited body and not the training provider

TobyRet: Ah right  I see now ... So confusing. I guess this publish / api inconsistency will addressed when we get around to this - https://trello.com/c/VCOC2FlP/36-make-accreditedprovider-etc-naming-consistent 🤞

PR #1389, opened 2020-05-29
3459 Course updated emails not being sent

gpeng:
### Context

Course updated emails don't appear to be being sent when a course is updated. Turns out we were checking notifiable users using the provider code rather than the accredited body code on the course so what was actually happening is that there would have been emails sent in some cases, just not to the correct users. This wasn't clear as we don't have many subscriptions and may explain why we didn't see this in user research when the same code was used.

* Update the `UserNotificationService` to use the correct provider code when querying the users to be notified
* Remove the `User.notifiable_users` as the `User` model has enough to worry about. Move that logic into the service as it seems like the service that should be concerned with this relationship. This also makes it consistent with the `PublishService`. 

### Guidance to review

This fixes the bug that is blocking the feature being launched. There is a bit of tidy up needed of this code to get it consistent and sort all of the naming inconsistencies etc. There's a card for that so be gentle https://trello.com/c/zxFwsm2N/3458-rename-create-to-publish-throughout-the-notifications-code

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,11 +19,6 @@ class User < ApplicationRecord
   scope :admins, -> { where(admin: true) }
   scope :non_admins, -> { where.not(admin: true) }
   scope :active, -> { where.not(accept_terms_date_utc: nil) }
-  scope :notifiable_users, ->(provider_code) do
-    joins(:user_notifications).merge(
-      UserNotification.course_update_notification_requests(provider_code),
-    )
-  end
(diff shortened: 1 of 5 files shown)


gpeng:
[Course updated emails not being sent](https://trello.com/c/oU51JVMB/3459-course-updated-emails-not-being-sent)

PR #1390, opened 2020-05-30
XXXX cache api docs for builds

misaka:
### Context

Teacher training API builds are slow.

image

Make builds faster by building branch-specific images and ensuring we use them in the docker cache.

A few more details about what's changed:

- We build an image for the `middleman` stage separately and push it up to docker-hub.
- We pull down the `middleman` image before each build and try to cache from it it.
- Don't tag either images, `api` or `middleman`, as `latest` anymore, but instead:
- Tag `api` and `middleman` images with the branch name and push/pull them to dockerhub so we're more likely to be able to reuse them next build.

### Guidance to review

Makes builds faster.

image

Oh, and ya, I'll squash this mess back together.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,16 @@ FROM ruby:2.6.5-alpine AS middleman
 
 RUN apk add --update --no-cache npm git build-base
 
-ADD . .
+COPY docs/Gemfile docs/Gemfile.lock /
 
-RUN cd docs && bundle install --jobs=4 && bundle exec middleman build --build-dir=../public && cd -
+RUN bundle install --jobs=4
+
+COPY docs /docs
+COPY public /public
+COPY swagger /swagger
+
+WORKDIR docs
+RUN bundle exec middleman build --build-dir=../public
 
 ###
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,11 +13,13 @@ variables:
   imageNameBgMailer: 'teacher-training-bg-mailer'
   dockerOverride: 'docker-compose -f docker-compose.yml'
 
+
 steps:
 - script: |
     set -x
     GIT_SHORT_SHA=$(echo $(Build.SourceVersion) | cut -c 1-7)
     docker_path=$(dockerHubUsername)/$(imageName)
+    docker_middleman_path=$(dockerHubUsername)/$(imageName)-middleman
     docker_bg_geocode_path=$(dockerHubUsername)/$(imageNameBgGeoCode)
     docker_bg_mailer_path=$(dockerHubUsername)/$(imageNameBgMailer)
     set +x # We confuse VSTS if we trace these commands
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -2,9 +2,16 @@ FROM ruby:2.6.5-alpine AS middleman
...
 
-RUN cd docs && bundle install --jobs=4 && bundle exec middleman build --build-dir=../public && cd -
+RUN bundle install --jobs=4
+
+COPY docs /docs
+COPY public /public
+COPY swagger /swagger
+
+WORKDIR docs
+RUN bundle exec middleman build --build-dir=../public && cd -
```
saliceti, line 14: What's the poiunt of `cd -`?

misaka: No point, it was just left over from before. Will trim

defong: can we stick `bundle exec rake rswag:specs:swaggerize` somwhere where here?

Dockerfile
```diff
@@ -2,9 +2,16 @@ FROM ruby:2.6.5-alpine AS middleman
...
 
-ADD . .
+COPY docs/Gemfile docs/Gemfile.lock /
 
-RUN cd docs && bundle install --jobs=4 && bundle exec middleman build --build-dir=../public && cd -
+RUN bundle install --jobs=4
+
+COPY docs /docs
+COPY public /public
+COPY swagger /swagger
```
defong, line 11: not sure what the `COPY` in here is for really...

azure-pipelines.yml
```diff
@@ -30,22 +32,59 @@ steps:
 
 - script: |
     set -x
-    docker pull $(docker_path):latest || true
-  displayName: "Pull latest docker image to cache"
+    docker pull $(docker_path):$GIT_BRANCH || true
+    docker pull $(docker_path)-middleman:$GIT_BRANCH || true
+  displayName: "Pull any existing images for this branch to cache"
+  # When this branch is first created, there won't be any images to cache, even though master
```
saliceti, line 38: You could pull master always, and pull from the branch conditionnally ex: `[ $GIT_BRANCH != master ] && ...`

misaka: I think the tricky bit is in the `docker-compose.yml`, it needs to `cache_from` the correct image (I ran into a problem when specifying multiple `cache_from` so that wouldn't work). We could drive that from a env var, maybe, but I think it may be worth it to get what's here in first.

Dockerfile
misaka: The `swagger` folder has the OpenAPI spec, which gets used during the `middleman` run, so this `COPY` ensures it's there.

misaka: tbh, I wonder if it's worth the effort. We might be able to tidy this up, but the entire goal of this stage is to generate a `public` directory that has a rendered version of the API docs, which then gets used in the final build.

This used to be `ADD . .`, but since we only use the three directories above I figured it'd be a quick win to pick them out instead of copying the entire app directory.

docker-compose.yml
```diff
@@ -9,19 +9,28 @@ services:
      - dbdata:/var/lib/postgresql/data
     environment:
     - POSTGRES_PASSWORD=developmentpassword
+  middleman:
+    build:
+      context: .
+      target: middleman
+      cache_from:
+        - ${dockerHubUsername:-dfedigital}/teacher-training-api-middleman:${GIT_BRANCH:-latest}
+    image: ${dockerHubUsername:-dfedigital}/teacher-training-api-middleman:${GIT_BRANCH:-latest}
```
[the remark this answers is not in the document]
misaka: Ya, I think I was assuming that the same agent would be running all the steps, so it wouldn't need to have pushed the `master` image (note that we don't pull as part of the "execute tests" step). However, I'm not sure there's a reason _not_ to push the `master`.

I don't think `latest` would ever show up here when running on CI, this would purely be for convenience when running locally so devs can get it working without having to specifying the env var `GIT_BRANCH`. This may be more surprising than convenient, however, time will tell and we can change this at any time.

azure-pipelines.yml
```diff
@@ -30,22 +32,59 @@ steps:
...
+
+- task: Docker@1
+  displayName: Docker Hub login
+  inputs:
+    command: "login"
+    containerregistrytype: Container Registry
+    dockerRegistryEndpoint: DfE Docker Hub
+- task: Docker@1
+  displayName: Push image $(dockerHubUsername)/$(imageName):$(Build.SourceBranchName)
+  condition: and(succeeded(), ne(variables['Build.SourceBranch'], 'refs/heads/master'))
```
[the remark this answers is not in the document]
misaka: Ya, I think you're right and `master` should be pushed up too. 👍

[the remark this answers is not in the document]
misaka: Oh, wait, I see now. `master` only gets pushed if the tests run without errors. Along with the `$(Build.BuildNumber)` image. See line 144 of this file. The point of pushing the branches up early, before the tests run, is that it could potentially save time on a subsequent build.

Thinking about that, though, I think that that last assumption is risky and wrong: we should probably never push the image before the tests run. First of all, no matter the branch we never want to have an image with failing code pushed up, and secondly I'm not sure we'd be saving a lot of time with caching a broken image anyway.

Dockerfile
misaka: The generated file `swagger/public_v1/api_spec.json` is checked into the repo, so this would mean that we could potentially have an openapi spec that's not up-to-date in the repo, but generate a different set of docs.

Sounds to me like what we outta do is run `bundle exec rake rswag:specs:swaggerize` and fail the build if it turns out `swagger/public_v1/api_spec.json` isn't up-to-date. Sounds like that belongs in another PR.

azure-pipelines.yml
```diff
@@ -13,11 +13,13 @@ variables:
   imageNameBgMailer: 'teacher-training-bg-mailer'
   dockerOverride: 'docker-compose -f docker-compose.yml'
 
+
 steps:
 - script: |
     set -x
     GIT_SHORT_SHA=$(echo $(Build.SourceVersion) | cut -c 1-7)
     docker_path=$(dockerHubUsername)/$(imageName)
+    docker_middleman_path=$(dockerHubUsername)/$(imageName)-middleman
```
defong, line 22: calling it `middleman`

seems weird...
may `doc-gen` or something along those lines to reflect that that docker file is do....

run  middleman to generate static documentation

PR #1392, opened 2020-06-01
[3426] Add delete method to allocations

SamJamCul:
### Context

https://trello.com/c/T6kURpH0/3426-m-change-initial-allocation-requests

* Add a delete endpoint for allocations
* Modify update method to allow for number of places change without request type
* Modify allocation model to use include feature

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


SamJamCul:
[[M] Change "initial" allocation requests.](https://trello.com/c/T6kURpH0/3426-m-change-initial-allocation-requests)

app/controllers/api/v2/allocations_controller.rb
```diff
@@ -42,6 +44,15 @@ def update
         end
       end
 
+      def destroy
```
TobyRet, line 47: 'delete' or 'destroy'? Your PR title says 'delete'

spec/requests/api/v2/allocations_spec.rb
```diff
@@ -58,6 +58,29 @@
     end
   end
 
+  describe "DELETE" do
```
TobyRet, line 61: Same here?

app/controllers/api/v2/allocations_controller.rb
defong: https://stackoverflow.com/questions/22757450/difference-between-destroy-and-delete

spec/requests/api/v2/allocations_spec.rb
```diff
@@ -58,6 +58,29 @@
...
+    context "with an existing allocation" do
```
defong, line 62: feels wide, is there a reason not restricting or tests covering
- [ ] user association
- [ ] recruitment cycle

spec/requests/api/v2/allocations_spec.rb
SamJamCul: I'm not quite sure what the right way to phrase it is - it's a DELETE method in the API call, but it's a destroy method in the allocations controller!

spec/requests/api/v2/allocations_spec.rb
SamJamCul: For the delete method we haven't put any scoping on recruitment cycle - it only locates the record by id. 

For the user association, I've added a test where the user isn't from that accredited body!

defong: can add the scope for recruitment cycle then please... ther eis no reason to allow for deletion fo something out of bounds

app/controllers/api/v2/allocations_controller.rb
```diff
@@ -42,8 +44,26 @@ def update
...
+        authorize @allocation = Allocation.find(params[:id])
+
+        if not_current_recruitment_cycle?
+          return render jsonapi_errors: ["not current recruitment cycle"], status: :unprocessable_entity
+        end
+
+        if @allocation.delete
```
defong, line 54: maybe

``` ruby
  def safe_delete(recruitment_cycle:)
    if recruitment_cycle == self.recruitment_cycle
      self.delete
    else
      errors.add(:safe_delete, "not meet")
    end
  end
```

SamJamCul: Done!

PR #1393, opened 2020-06-01
[3417] don't validate courses when rolling over

misaka:
This may seem counter-intuitive, but it seems that our courses aren't all valid
at the current time, for example the "age_range_in_years" is nil on a bunch of
courses, but is validated as being present on new/create. We should fix this,
but for the time being we just need to duplicate the broken-ness.

This change also adds a little extra info when an error is raised to know which
course code was being processed at the time of the error.

### Context

When testing rollover locally, we got the error that a course was invalid pretty early on in the process.

Be messy and don't validate a course before saving. It seems that last year's rollover also didn't validate courses, and that the `save!` was added as part of the refactoring later in PR #726.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/services/providers/copy_to_recruitment_cycle_service.rb
+++ b/app/services/providers/copy_to_recruitment_cycle_service.rb
@@ -47,7 +47,7 @@ module Providers
       provider.courses.each do |course|
         new_course = @copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
         courses_count += 1 if new_course.present?
-      rescue
+      rescue Exception # rubocop: disable Lint/RescueException
         logger.fatal "error trying to copy course #{course.course_code}" if logger
         raise
       end
--- a/app/services/providers/copy_to_recruitment_cycle_service.rb
+++ b/app/services/providers/copy_to_recruitment_cycle_service.rb
@@ -5,7 +5,7 @@ module Providers
     def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:, logger: nil)
       @copy_course_to_provider_service = copy_course_to_provider_service
       @copy_site_to_provider_service = copy_site_to_provider_service
-      @logger = logger
+      @logger = logger || Logger.new("/dev/null")
     end
 
     def execute(provider:, new_recruitment_cycle:)
--- a/app/services/providers/copy_to_recruitment_cycle_service.rb
+++ b/app/services/providers/copy_to_recruitment_cycle_service.rb
@@ -1,8 +1,11 @@
 module Providers
   class CopyToRecruitmentCycleService
-    def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:)
+    attr :logger
+
+    def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:, logger: nil)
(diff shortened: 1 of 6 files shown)


misaka:
[Does rollover via MCB still work?](https://trello.com/c/zQGvFhTD/3417-does-rollover-via-mcb-still-work)

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -1,8 +1,11 @@
 module Providers
   class CopyToRecruitmentCycleService
-    def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:)
+    attr :logger
+
+    def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:, logger: nil)
       @copy_course_to_provider_service = copy_course_to_provider_service
       @copy_site_to_provider_service = copy_site_to_provider_service
+      @logger = logger
```
asmega, line 8:
```suggest
      @logger = logger || Logger.new("/dev/null")
```

asmega: You can use null object and remove the `if` statements

defong: will that work on a windows?

spec/mcb_helper.rb
```diff
@@ -1,3 +1,5 @@
+require 'spec_helper'
```
defong, line 1:
```suggestion
require "spec_helper"
```

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -44,6 +47,9 @@ def copy_courses_to_new_provider(provider, new_provider)
       provider.courses.each do |course|
         new_course = @copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
         courses_count += 1 if new_course.present?
+      rescue
```
defong, line 50: what are you rescuing?

> Avoid rescuing without specifying an error class.

app/services/providers/copy_to_recruitment_cycle_service.rb
misaka: > will that work on a windows?

It should running under mingw or WSL. :)

app/services/providers/copy_to_recruitment_cycle_service.rb
misaka: Ya, good point, I was just being sloppy. ⏩ ⏩

misaka: Just to be clear, anything, it's really just a bail-out clause.

defong: it got nothing to do with me being picky... its failing linter

https://dfe-ssp.visualstudio.com/Become-A-Teacher/_build/results?buildId=68147&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=59a85588-b0ba-5043-24c4-d9e29d89c6f6&l=65

PR #1394, opened 2020-06-01
[3466] Add extra endpoint to show a provider

asmega:
### Context

- https://trello.com/c/QpRrdwxL/3466-pre-launch-task-make-accredited-bodies-able-to-see-allocation-pages
- We don't want accredited bodies to see training providers under normal circumstances but sometimes we do. This adds another endpoint to allow for this
- This is needed for the allocation work so accredited bodies can access any training provider
- There is also another endpoint for suggestions so allocations autocomplete can search for any provider

- Additional endpoint for provider lookup
- This additional endpoint is identical to the show endpoint
- It uses a different policy
- The policy allows anyone to see any provider
- The additional endpoint was added instead of the policy of the #show
endpoint modified so on the client side a accredited body could not see
a training providers details under "normal" use cases which is desired
- This endpoint was added for the allocations work as it is possible for
an allocation to be associated with any provider

#### Suggestion endpoint

- Add endpoint for any provider suggestion
- This additional endpoint is identical to the suggest endpoint
- It uses a different policy
- The policy allows anyone to suggest any provider for the current cycle
- This endpoint was added for the allocations work so it is possible for
an allocation find any provider for autocomplete purposes

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -3,7 +3,7 @@ module API
     class ProvidersController < API::V2::ApplicationController
       before_action :get_user, if: -> { params[:user_id].present? }
       before_action :build_recruitment_cycle
-      before_action :build_provider, except: %i[index suggest]
+      before_action :build_provider, except: %i[index suggest suggest_any]
 
       deserializable_resource :provider,
                               only: :update,
@@ -63,6 +63,23 @@ module API
         return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
         found_providers = policy_scope(@recruitment_cycle.providers)
+          .search_by_code_or_name(params[:query])
+          .limit(5)
+
+        render(
+          jsonapi: found_providers,
+          class: { Provider: SerializableProviderSuggestion },
+        )
+      end
+
+      # Suggest any provider from the current recruitment cycle
+      def suggest_any
+        authorize Provider
+
+        return render(status: :bad_request) if params[:query].nil? || params[:query].length < 2
+        return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
+
+        found_providers = @recruitment_cycle.providers
           .search_by_code_or_name(params[:query])
           .limit(5)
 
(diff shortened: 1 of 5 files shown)


asmega:
[PRE-LAUNCH TASK: Make accredited bodies able to see allocation pages](https://trello.com/c/QpRrdwxL/3466-pre-launch-task-make-accredited-bodies-able-to-see-allocation-pages)

app/controllers/api/v2/providers_controller.rb
```diff
@@ -33,6 +33,14 @@ def show
         render jsonapi: @provider, include: params[:include]
       end
 
+      # GET /providers/:code/show_from_accredited_body
+      # Allows accredited bodies to see a provider they accredit
```
defong, line 37:
```suggestion
      # Allows an accredited body to see a training provider they accredit
```

app/policies/provider_policy.rb
```diff
@@ -44,4 +48,16 @@ def suggest?
   alias_method :update?, :show?
   alias_method :build_new?, :show?
   alias_method :can_list_training_providers?, :show?
+
+private
+
+  def accredited_bodies
+    provider.accredited_bodies.map do |p|
```
defong, line 55: it not this one right?

https://github.com/DFE-Digital/teacher-training-api/blob/d7211042f9f2d82601f4e2bdfa82aad4f7b0a13d/app/models/provider.rb#L223-L232

``` ruby
bath = Provider.where(provider_code: "B20", recruitment_cycle: RecruitmentCycle.current_recruitment_cycle).first
bath.training_providers

# or
nsta = Provider.where(provider_code: "16Z", recruitment_cycle: RecruitmentCycle.current_recruitment_cycle).first
nsta.accrediting_providers

```

asmega: Thats returns some random hash. Can't we work with objects? So why couldn't that method return a collection of `Provider` objects?

app/controllers/api/v2/providers_controller.rb
asmega: I have now changed to use your suggestion

app/policies/provider_policy.rb
defong: Oh doesn't x.accrediting_providers or x.training_providers one of them should work

asmega: PR has changed this code has now been deleted

PR #1398, opened 2020-06-04
[3494] Add auditing to allocations

dankmitchell:
🚢

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -3,6 +3,8 @@ class Allocation < ApplicationRecord
   belongs_to :accredited_body, class_name: "Provider"
   belongs_to :recruitment_cycle
 
+  audited associated_with: :provider
+
   validates :provider, :accredited_body, presence: true
   validates :number_of_places, numericality: true
 
--- a/spec/models/allocation_spec.rb
+++ b/spec/models/allocation_spec.rb
@@ -6,6 +6,10 @@ RSpec.describe Allocation do
       subject.valid?
     end
 
+    describe "auditing" do
+      it { should be_audited.associated_with(:provider) }
+    end
+
     it "requires accredited_body" do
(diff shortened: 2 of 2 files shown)


dankmitchell:
[Add auditing to allocations](https://trello.com/c/mZG5eeNK/3494-add-auditing-to-allocations)

app/models/allocation.rb
```diff
@@ -3,6 +3,8 @@ class Allocation < ApplicationRecord
   belongs_to :accredited_body, class_name: "Provider"
   belongs_to :recruitment_cycle
 
+  audited
```
misaka, line 6: Shouldn't we associate this with the provider?

misaka:
```suggestion
  audited associated_with: :provider
```

spec/models/allocation_spec.rb
```diff
@@ -6,6 +6,10 @@
       subject.valid?
     end
 
+    describe "auditing" do
+      it { should be_audited }
```
misaka, line 10:
```suggestion
      it { should be_audited.associated_with(:provider) }
```

app/models/allocation.rb
asmega: I checked earlier
```
=> #<Audited::Audit:0x00007fea288ac6d0
 id: 259515,
 auditable_id: 436,
 auditable_type: "Allocation",
 associated_id: nil,
 associated_type: nil,
 user_id: 8916,
 user_type: "User",
 username: nil,
 action: "update",
 audited_changes: {"request_type"=>[1, 2], "number_of_places"=>[28, 0]},
 version: 2,
 comment: nil,
 remote_address: "::1",
 request_uuid: "3c498ddc-5048-4214-a819-28cebe7370fd",
 created_at: Thu, 04 Jun 2020 10:07:41 UTC +00:00>
```
The user and allocation is sufficient as that associates with everything that we need

PR #1399, opened 2020-06-04
[3456] Update user notifications when accredited body changes

AgaDufrat:
### Context
A user's notification preferences are stored with one record (`UserNotification`) per accredited body. If the user is associated with a new accredited body and they have existing preferences stored we need to add a record for the new accredited body with the same preferences.

The opposite process also needs to take place if they are removed from an accredited body.

Update user notifications when user is added or removed to an accredited body.
When user is granted access to a new organisation, they will receive notifications for the new organisation, if they opted in.   
When user's access to an organisation is revoked, they will no longer receive notifications for the organisation that was removed, if they opted in. 

Adding/removing access is done via MCB.

_Note:_
It is also possible to add a user to organisations in Publish `/access-requests/new` by copying another user's permission. However, this feature creates a new user so there is no need to update the notifications since the user has never opted-in.
https://github.com/DFE-Digital/teacher-training-api/blob/01eb5f002cb1125af0f669d774f41dbf9ba1d22f/app/models/access_request.rb#L17

### Guidance to review
1. Turn on notifications in Publish http://localhost:3000/notifications

2. Give yourself access to a new provider eg
`bundle exec bin/mcb users grant name.surname@digital.education.gov.uk -p B20`

- In the API console
`u = User.find_by(email: <email>)`
  - Check if the new organisations is added `u.organisations`
  - Check if the UserNotification was created `UserNotification.where(user_id: u.id)`

3. Revoke your access to that organisation
`bundle exec bin/mcb users revoke name.surname@digital.education.gov.uk --provider_code B20` and check in the console

4. Delete notifications `UserNotification.destroy_all` repeat 2. and 3. but the UserNotification should not be created

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/user_associations_service/create.rb
+++ b/app/services/user_associations_service/create.rb
@@ -1,0 +1,38 @@
+module UserAssociationsService
+  class Create
+    attr_reader :organisation, :user
+
+    class << self
+      def call(**args)
+        new(args).call
+      end
+    end
+
+    def initialize(organisation:, user:)
+      @organisation = organisation
+      @user = user
+    end
+
+    def call
+      add_user_to_organistation
+      update_user_notification_preferences
+    end
+
+    private_class_method :new
+
+  private
+
+    def add_user_to_organistation
+      organisation.add_user(user)
+    end
+
+    def update_user_notification_preferences
+      user_notification_preferences = UserNotificationPreferences.new(user_id: user.id)
+      return if user_notification_preferences.updated_at.nil?
+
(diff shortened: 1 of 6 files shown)


AgaDufrat:
[[L] Update user's notification preferences if their accredited body associations change](https://trello.com/c/iBQkf3Wy/3456-l-update-users-notification-preferences-if-their-accredited-body-associations-change)

PR #1400, opened 2020-06-04
[3341] - Optimise courses as an accredited body query

TobyRet:
### Context
“Courses as accredited body” page is extremely slow, 20-40sec Time to First Byte. This is because the API endpoint returning the list of training providers or course counts are extremely slow.

This is a spike to see what improvements could be made. This is one option for feedback purposes.

### Changes proposed 
- Return the courses as an accredited body count as meta
- NB: this code has not been refactored. No tests yet
- Publish `ProvidersController` would just then need a minor tweak to access the count via the meta object

```
def index
    page = (params[:page] || 1).to_i
    per_page = 10

    @providers = providers.page(page)

     # No longer making a separate Courses API call
    @pagy = Pagy.new(count: @providers.meta.count, page: page, items: per_page)

    @providers_view = ProvidersView.new(providers: providers)

    render "providers/no_providers", status: :forbidden if @providers.empty?
    redirect_to provider_path(@providers.first.provider_code) if @providers.size == 1
  end
```
### Load times
Running locally, the follow show Publish load times

#### Before the change

```
003] (10.195s) ProvidersController -- Completed #training_providers -- {:controller=>"ProvidersController", :action=>"training_providers", :params=>{"provider_code"=>"K30", "year"=>"2020"}, :format=>"HTML", :method=>"GET", :path=>"/organisations/K30/2020/training-providers", :status=>200, :view_runtime=>240.8, :user=>{:id=>"10866", :sign_in_user_id=>"52fd0689-809e-469b-9f7b-28741f03949f"}, :request_id=>"f0e139ae-8efe-40ef-a269-c33a5c281945", :status_message=>"OK"}
```

#### After the change
```
003] (827.4ms) ProvidersController -- Completed #training_providers -- {:controller=>"ProvidersController", :action=>"training_providers", :params=>{"provider_code"=>"K30", "year"=>"2020"}, :format=>"HTML", :method=>"GET", :path=>"/organisations/K30/2020/training-providers", :status=>200, :view_runtime=>193.04, :user=>{:id=>"10866", :sign_in_user_id=>"52fd0689-809e-469b-9f7b-28741f03949f"}, :request_id=>"58a2bce5-0972-445d-a48d-1bba95ee0761", :status_message=>"OK"}
```

### Run it
To see it in action, run this branch against this Publish branch - https://github.com/DFE-Digital/publish-teacher-training/tree/3341-optimise-courses-as-accredited-body

Visit `localhost:3000/organisations/M80/2020/training-providers`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -50,7 +50,21 @@ module API
 
         providers = TrainingProviderSearch.new(provider: @provider, params: params).call
 
-        render jsonapi: providers, include: params[:include]
+        course_counts = {}
+
+        providers.map do |provider|
+          courses_count = provider
+                            .courses
+                            .where(accrediting_provider_code: @provider.provider_code,)
+                            .count
+          course_counts[provider.provider_code] = courses_count
+        end
+
+        render jsonapi: providers,
+               include: params[:include],
+               meta: {
+                 course_counts: course_counts
+               }
       end
 
     private
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -48,22 +48,20 @@ module API
       def index
         authorize @provider, :can_list_training_providers?
 
-        providers = TrainingProviderSearch.new(provider: @provider, params: params).call
+        providers = TrainingProviderSearch.new(provider: @provider, params: params)
+                                          .call
+                                          .include_accredited_courses_counts(@provider.provider_code)
 
-        course_counts = {}
+        accredited_courses_counts = {}
 
-        providers.map do |provider|
-          courses_count = provider
-                            .courses
-                            .where(accrediting_provider_code: @provider.provider_code,)
-                            .count
-          course_counts[provider.provider_code] = courses_count
+        providers.each do |p|
+          accredited_courses_counts[p.provider_code] = p.accredited_courses_count
         end
 
         render jsonapi: providers,
                include: params[:include],
                meta: {
-                 course_counts: course_counts
+                 accredited_courses_counts: accredited_courses_counts,
                }
       end
 
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -182,11 +182,11 @@ class Provider < ApplicationRecord
   end
 
   def courses_count
-    self.respond_to?("included_courses_count") ? included_courses_count : courses.size
+    self.has_attribute?("included_courses_count") ? included_courses_count : courses.size
   end
 
   def accredited_courses_count
-    self.respond_to?("included_accredited_courses_count") ? included_accredited_courses_count : 0
+    self.has_attribute?("included_accredited_courses_count") ? included_accredited_courses_count : 0
   end
 
   def update_changed_at(timestamp: Time.now.utc)
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -167,8 +167,26 @@ class Provider < ApplicationRecord
     ).select("provider.*, COALESCE(a.courses_count, 0) AS included_courses_count")
   end
 
+  def self.include_accredited_courses_counts(provider_code)
+    joins(
+      %{
+        LEFT OUTER JOIN (
+          SELECT b.provider_id, COUNT(*) courses_count
+          FROM course b
+          WHERE b.discarded_at IS NULL
+          AND b.accrediting_provider_code = #{ActiveRecord::Base.connection.quote(provider_code)}
+          GROUP BY b.provider_id
+        ) a ON a.provider_id = provider.id
+      },
+      ).select("provider.*, COALESCE(a.courses_count, 0) AS included_accredited_courses_count")
+  end
+
   def courses_count
     self.respond_to?("included_courses_count") ? included_courses_count : courses.size
+  end
+
+  def accredited_courses_count
+    self.respond_to?("included_accredited_courses_count") ? included_accredited_courses_count : 0
   end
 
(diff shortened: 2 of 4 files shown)


misaka:
I see. Ya, this makes sense. Currently the requests made to API from Publish are:

```
5.697ms GET /api/v2/recruitment_cycles/2020
17.7ms  GET /api/v2/recruitment_cycles/2020/providers/Y50
82.3ms  GET /api/v2/recruitment_cycles/2020/providers/Y50/training_providers?recruitment_cycle_year=2020
2.457s  GET /api/v2/recruitment_cycles/2020/courses?filter%5Baccrediting_provider_code%5D=Y50
```

So ya, by returning the counts as part of the `training_providers` request you'd be saving a bunch of time.

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
```diff
@@ -50,7 +50,21 @@ def index
...
-        render jsonapi: providers, include: params[:include]
+        course_counts = {}
+
+        providers.map do |provider|
+          courses_count = provider
+                            .courses
+                            .where(accrediting_provider_code: @provider.provider_code,)
+                            .count
+          course_counts[provider.provider_code] = courses_count
+        end
```
misaka, line 61: There is also an [existing method `include_course_counts`](https://github.com/DFE-Digital/teacher-training-api/blob/master/app/models/provider.rb#L157-L168) that optimises this:

```ruby
  # This is used by the providers index; it is a replacement for `.includes(:courses)`,
  # but it only fetches the counts for the associated courses. By not fetching all the
  # course objects for 1000+ providers, the db query runs much faster, and the view spends
  # less time rendering because there's less data to comb through.
  def self.include_courses_counts
    joins(
      %{
        LEFT OUTER JOIN (
          SELECT b.provider_id, COUNT(*) courses_count
          FROM course b
          WHERE b.discarded_at IS NULL
          GROUP BY b.provider_id
        ) a ON a.provider_id = provider.id
      },
    ).select("provider.*, COALESCE(a.courses_count, 0) AS included_courses_count")
  end
```

Used in [API::V2::ProvidersController#index](https://github.com/DFE-Digital/teacher-training-api/blob/master/app/controllers/api/v2/providers_controller.rb#L16).

I think this would change the shape of the response, iirc it adds the course count to the meta for each individual provider, but it'd be nice to use an existing pattern.

TobyRet: I'm uncertain whether given the granularity of the end point whether the accredited_courses_count meta should be added to each individual Provider resource or this should be added at the root of the payload as described on line 64

app/models/provider.rb
```diff
@@ -167,10 +167,28 @@ def self.include_courses_counts
     ).select("provider.*, COALESCE(a.courses_count, 0) AS included_courses_count")
   end
 
+  def self.include_accredited_courses_counts(provider_code)
```
TobyRet, line 170: A new statement had to be created because we have need to inject the `accredited_body_code`

TobyRet:
[Speed up “Courses as accredited body” page](https://trello.com/c/vYkAmVjl/3341-speed-up-courses-as-accredited-body-page)

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
TobyRet: Leaving the new meta object in the root as discussed

app/controllers/api/v2/accredited_provider_training_providers_controller.rb
```diff
@@ -48,9 +48,21 @@ def training_providers
...
 
-        providers = TrainingProviderSearch.new(provider: @provider, params: params).call
+        providers = TrainingProviderSearch.new(provider: @provider, params: params)
+                                          .call
+                                          .include_accredited_courses_counts(@provider.provider_code)
 
-        render jsonapi: providers, include: params[:include]
+        accredited_courses_counts = {}
+
+        providers.map do |p|
```
defong, line 57: either change `map`  to `each`

```ruby
        providers.each do |p|
```
or actually use `map` to return the array

```ruby
accredited_courses_counts = providers.map do |p|
```

app/models/provider.rb
```diff
@@ -167,10 +167,28 @@ def self.include_courses_counts
...
+      },
+      ).select("provider.*, COALESCE(a.courses_count, 0) AS included_accredited_courses_count")
+  end
+
   def courses_count
     self.respond_to?("included_courses_count") ? included_courses_count : courses.size
   end
 
+  def accredited_courses_count
+    self.respond_to?("included_accredited_courses_count") ? included_accredited_courses_count : courses.size
```
defong, line 189: if it responds to `included_accredited_courses_count`
then return `included_accredited_courses_count` else
return `courses.size`

course.size is not the same as `accredited`_courses_count returning 0 should be fine

app/models/provider.rb
```diff
@@ -167,10 +167,28 @@ def self.include_courses_counts
...
+      },
+      ).select("provider.*, COALESCE(a.courses_count, 0) AS included_accredited_courses_count")
+  end
+
   def courses_count
     self.respond_to?("included_courses_count") ? included_courses_count : courses.size
   end
 
+  def accredited_courses_count
+    self.respond_to?("included_accredited_courses_count") ? included_accredited_courses_count : 0
```
misaka, line 189: It took me a little thinking to understand what this was all about. I know this is just a copy of the above, but it'd be more readable like this:

```suggestion
    has_attribute?(:included_accredited_courses_count) ? included_accredited_courses_count : 0
```

TobyRet: 👍

PR #1404, opened 2020-06-05
[3287] Add 0 allocation request validations

AgaDufrat:
### Context
We shouldn't accept `0` allocation requests for initial requests.  "Repeat"
request should still allow "0" places because this is where we don't have date
for the training provider.

Validate number_of_places is not zero if the request type is "initial" 

### Guidance to review
- try to create new allocations with the following params
provider_code: "B20", provider_id: 15052, **request_type: "initial", number_of_places: 0**

Note: when I tried adding `presence: true` Publish blows up

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -9,6 +9,7 @@ class Allocation < ApplicationRecord
   validates :number_of_places, numericality: true
 
   validate :accredited_body_is_an_accredited_body
+  validate :non_zero_initial_request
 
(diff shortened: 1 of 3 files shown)


AgaDufrat:
[[S] Add 0 allocation request validations](https://trello.com/c/BMy4XcQU/3287-s-add-0-allocation-request-validations)

PR #1412, opened 2020-06-09
[3517] add bg jobs dockerfile

misaka:
### Context

We need to run background jobs. We also need a generic job runner as it's turning into too much overhead having one container per task (mailer, geosync, etc).

Add a generic background job runner image that's a copy of the existing ones, but with a generic name. A future PR will scrap the existing ones and configure this one to run all the queues.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/bg-jobs/Dockerfile
+++ b/bg-jobs/Dockerfile
@@ -1,0 +1,7 @@
+ARG  DOCKER_REPOSITORY
+ARG  TEACHER_TRAINING_API
+ARG  CODE_VERSION=${GIT_BRANCH:-latest}
+
+FROM ${DOCKER_REPOSITORY}/${TEACHER_TRAINING_API}:${CODE_VERSION}
+
+CMD bundle exec rails db:migrate && bundle exec sidekiq -q save_statistic
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
      - dbdata:/var/lib/postgresql/data
     environment:
     - POSTGRES_PASSWORD=developmentpassword
+
   middleman:
     build:
       context: .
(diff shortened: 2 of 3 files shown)


misaka:
[Create docker image for stats job](https://trello.com/c/OqlC5omC/3517-create-docker-image-for-stats-job)

docker-compose.yml
```diff
@@ -76,3 +79,21 @@ services:
...
+    volumes:
+      - .:/app
+    depends_on:
+      - web
+      - db
+    environment:
+      - DB_HOSTNAME=db
+      - DB_USERNAME=postgres
+      - DB_PASSWORD=developmentpassword
+      - SETTINGS__APPLICATION=teacher-training-api-bg-mailer
```
defong, line 99:
```suggestion
      - SETTINGS__APPLICATION=teacher-training-api-bg-jobs
```

bg-jobs/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ARG  DOCKER_REPOSITORY
+ARG  TEACHER_TRAINING_API
+ARG  CODE_VERSION=${GIT_BRANCH:-latest}
+
+FROM ${DOCKER_REPOSITORY}/${TEACHER_TRAINING_API}:${CODE_VERSION}
+
+CMD bundle exec rails db:migrate && bundle exec sidekiq -q save_statistic
```
saliceti, line 7: Can we create a generic container which subscribes to all the queues?
Or just have one queue?

misaka: I'd propose we do that work as a separate PR, once we have this one up and running and stable. We can then have a single PR to retire the other two containers and allow this one to process all the queues.

docker-compose.yml
misaka: 🤦

bg-jobs/Dockerfile
saliceti: OK makes sense too

PR #1413, opened 2020-06-09
[3502] Fix branch name so its sent to code climate

aliuk2012:
### Guidance to review
I've checked the request made to code climate and as you can see the branch name is no longer `HEAD` and you can see the git commit assigned too.
image

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,7 +97,7 @@ steps:
     GIT_BRANCH: $(Build.SourceBranchName)
 - script: |
     set -x
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c "bundle exec rake cc:report"
+    $DOCKER_OVERRIDE exec -T -e GIT_BRANCH=$GIT_BRANCH web rails cc:report
   displayName: 'Execute Code Climate reporter'
   env:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -97,7 +97,7 @@ steps:
     GIT_BRANCH: $(Build.SourceBranchName)
 - script: |
     set -x
-    $DOCKER_OVERRIDE exec -T web /bin/sh -c "bundle exec rake cc:report"
+    $DOCKER_OVERRIDE exec -T web /bin/sh -c "export GIT_BRANCH=${GIT_BRANCH} && bundle exec rake cc:report"
```
misaka, line 100:
```suggestion
    $DOCKER_OVERRIDE exec -T -e GIT_BRANCH=$GIT_BRANCH web rake cc:report
```

I'm just testing this, but:

- the `-e` flag is meant to do just this
- `exec` is documented as taking the args: `COMMAND [ARGS ...]`
- rake doesn't need `bundle exec` I'm pretty sure anymore (I'm still testing)

PR #1414, opened 2020-06-09
[3330] Added forbidden for discarded user

defong:
### Context
A discarded user

`do not pass go do not collect 200`

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


defong:
[[M] Prevent access to publish for discarded users](https://trello.com/c/kdBZwQr8/3330-prevent-access-to-publish-for-discarded-users)

app/controllers/api/v2/application_controller.rb
```diff
@@ -7,6 +7,7 @@ class ApplicationController < ::ApplicationController
 
       before_action :store_request_id
       before_action :check_terms_accepted
+      before_action :check_user_is_kept
```
misaka, line 10: Should this run before the other actions here? Otherwise it sounds like we might tell them they need to accept terms and conditions when in reality they have no access at all.

defong: done

PR #1418, opened 2020-06-12
[3532] Add default kept scope to user mode

SamJamCul:
### Context
https://trello.com/c/Z9uH4OZB/3532-only-show-kept-users-on-orgusers-list

Add default scope to users to prevent showing discarded 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -34,6 +34,18 @@ describe Provider, type: :model do
     end
   end
 
+  describe "users" do
+    let(:discarded_user) { create(:user, :discarded) }
+
+    before do
+      provider.organisation.users << discarded_user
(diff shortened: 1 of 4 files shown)


SamJamCul:
[Only show "kept" users on org#users list](https://trello.com/c/Z9uH4OZB/3532-only-show-kept-users-on-orgusers-list)

spec/models/provider_spec.rb
```diff
@@ -34,6 +34,18 @@
     end
   end
 
+  fdescribe "users" do
```
asmega, line 37:
```suggestion
  describe "users" do
```

PR #1421, opened 2020-06-15
[3336] - Accredited Body New Feature Interruption Screen

TobyRet:
### Context
On Publish a new interruption screen will be shown to users who are associated with an accredited body. 

In order to determine whether the screen should be shown the V2 users#show endpoint now contains a meta object confirming whether the user is associated with an accredited body.

### Out of scope
- Another PR will be raised to remove the AASM library and state machine. 
- Custom user endpoints for triggering state will also be removed in a subsequent PR (as all state updated now happen through the update action)

### Guidance to review
This PR relies on this Publish PR-  https://github.com/DFE-Digital/publish-teacher-training/pull/1178 - please see this for instructions on on how to test changes 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -39,6 +39,7 @@ class User < ApplicationRecord
     state :new, initial: true
     state :transitioned
     state :rolled_over
+    state :seen_accredited_body_features
 
     event :accept_transition_screen do
       transitions from: :new, to: :transitioned
@@ -56,6 +57,14 @@ class User < ApplicationRecord
   # accepts array or single organisation
   def remove_access_to(organisations_to_remove)
     self.organisations = self.organisations - Array(organisations_to_remove)
+  end
+
+  def associated_with_accredited_body?
+    providers
+      .in_current_cycle
+      .accredited_body
+      .count
+      .positive?
   end
 
 private
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -60,6 +60,7 @@ module API
             :sign_in_user_id,
             :welcome_email_date_utc,
             :invite_date_utc,
+            :state,
             :accept_terms_date_utc,
           )
       end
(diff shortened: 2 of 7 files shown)


TobyRet:
[[M] Support interruption screen](https://trello.com/c/8BkCq4GO/3336-m-support-interruption-screen)

TobyRet:
[[M] Support interruption screen](https://trello.com/c/8BkCq4GO/3336-m-support-interruption-screen)

app/models/user.rb
```diff
@@ -58,6 +59,10 @@ def remove_access_to(organisations_to_remove)
     self.organisations = self.organisations - Array(organisations_to_remove)
   end
 
+  def associated_with_accredited_body?
+    providers.accredited_body.count != 0
```
defong, line 63:
```suggestion
    providers.accredited_body.count.positive?
```

app/models/user.rb
```diff
@@ -39,6 +39,7 @@ class User < ApplicationRecord
     state :new, initial: true
     state :transitioned
     state :rolled_over
+    state :seen_accredited_body_features
```
misaka, line 42: Just to check, @TobyRet, are we leaving this here until we get the state machine functioning in Publish, and then rip it out as a separate piece of work?

app/models/user.rb
```diff
@@ -58,6 +59,14 @@ def remove_access_to(organisations_to_remove)
     self.organisations = self.organisations - Array(organisations_to_remove)
   end
 
+  def associated_with_accredited_body?
+    providers
+      .where(recruitment_cycle: RecruitmentCycle.current)
```
defong, line 64:
```suggestion
      .in_current_cycle
```
I found Graham's beautiful code
https://github.com/DFE-Digital/teacher-training-api/blob/a9e98c21cc7e214f383a96c8055da5ed533f6523/app/models/provider.rb#L91

app/models/user.rb
TobyRet: Yep that is correct

app/models/user.rb
TobyRet: 👍

PR #1424, opened 2020-06-15
[3531] Amended admin users to be less involved

defong:
### Context
Admin user are omnipotent

Amended admin users to be less involved

### Guidance to review
Admin users are basically omnipotent and serves a different purpose.

If an admin user is associated to organisation then that is removed.
-  this removes the user associations provider

If an admin user is associated user notification then that is removed.
- see the above
- if remained then there is discrepancies with expectations as business rules are applied

An admin user is not the same as a normal user, so will not have the exacting functionality.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


defong:
[Remove all DfE users / admins from providers](https://trello.com/c/6fsFitek/3531-remove-all-dfe-users-admins-from-providers)

PR #1425, opened 2020-06-15
[3500] Add provider suggestions filtering

asmega:
### Context

- https://trello.com/c/OO8AqZZe/3500-course-creation-accredited-body-search
- Provider suggestions can not be filtered to only show accredited bodies which is needed by publish for autocomplete purposes (or non-js equivalent)

- Provider suggestions can now be filtered to only suggest providers that are accredited bodies

### Guidance to review

- with valid api key `curl http://localhost:3001/api/v2/providers/suggest_any?filter%5Baccredited_body%5D=true&query=london`
- should only return accredited bodies

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -79,17 +79,23 @@ module API
         return render(status: :bad_request) if params[:query].nil? || params[:query].length < 2
         return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
-        found_providers = @recruitment_cycle.providers
-          .search_by_code_or_name(params[:query])
-          .limit(5)
+        scope = @recruitment_cycle.providers
+                                  .search_by_code_or_name(params[:query])
+                                  .limit(5)
+
+        scope = scope.accredited_body if only_accredited_body_filter?
 
         render(
-          jsonapi: found_providers,
+          jsonapi: scope,
           class: { Provider: SerializableProviderSuggestion },
         )
       end
 
     private
+
+      def only_accredited_body_filter?
+        params.dig(:filter, :only_accredited_body) == "true"
+      end
 
       def begins_with_alphanumeric(string)
         string.match?(/^[[:alnum:]].*$/)
--- a/spec/controllers/api/v2/providers_controller_spec.rb
+++ b/spec/controllers/api/v2/providers_controller_spec.rb
@@ -31,9 +31,26 @@ RSpec.describe API::V2::ProvidersController do
       let(:course) { create(:course) }
       let!(:other_provider) { course.provider }
 
-      it "returns the specified provider" do
+      it "returns unassociated providers" do
         get :suggest_any, params: { query: other_provider.provider_code }
         expect(JSON.parse(response.body).dig("data").map { |p| p["id"] }).to eql([other_provider.id.to_s])
+      end
+    end
+
+    context "with a filter" do
+      let(:user) { provider.users.first }
+      let!(:provider) { create(:provider) }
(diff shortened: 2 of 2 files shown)


asmega:
[Course creation accredited body search](https://trello.com/c/OO8AqZZe/3500-course-creation-accredited-body-search)

app/models/provider.rb
```diff
@@ -25,6 +25,8 @@ class Provider < ApplicationRecord
     not_an_accredited_body: "N",
   }
 
+  scope :are_accredited_bodies, -> { where(accrediting_provider: "Y") }
```
defong, line 28: not needed

app/controllers/api/v2/providers_controller.rb
```diff
@@ -79,18 +79,24 @@ def suggest_any
...
-          jsonapi: found_providers,
+          jsonapi: scope,
           class: { Provider: SerializableProviderSuggestion },
         )
       end
 
     private
 
+      def accredited_body_filter?
+        params[:filter] && params[:filter][:accredited_body] == "true"
```
defong, line 97:
```suggestion
        params.dig(:filter, :only_accredited_body) == "true"
```

`accredited_body` I'm not sure of the naming maybe `only_accredited_body` since it a tristate


>     should only return accredited bodies

spec/controllers/api/v2/providers_controller_spec.rb
```diff
@@ -31,10 +31,22 @@
...
     end
+
+    context "with a filter" do
+      let(:user) { provider.users.first }
+      let!(:provider) { create(:provider) }
+      let!(:in_scope_provider) { create(:provider, :accredited_body, provider_name: "in provider") }
+      let!(:out_scope_provider) { create(:provider, provider_name: "out provider") }
+
+      it "applies filter to results" do
+        get :suggest_any, params: { query: "provider", filter: { accredited_body: "true" } }
```
defong, line 47: where is the false test?

app/controllers/api/v2/providers_controller.rb
```diff
@@ -79,18 +79,24 @@ def suggest_any
...
         return render(status: :bad_request) unless begins_with_alphanumeric(params[:query])
 
-        found_providers = @recruitment_cycle.providers
-          .search_by_code_or_name(params[:query])
-          .limit(5)
+        scope = @recruitment_cycle.providers
+                                  .search_by_code_or_name(params[:query])
+                                  .limit(5)
+
+        scope = scope.are_accredited_bodies if accredited_body_filter?
```
defong, line 86:
```suggestion
        scope = scope.accredited_body if accredited_body_filter?
```

https://github.com/DFE-Digital/teacher-training-api/blob/fdb35d6428d7ad39ad570fe39cba2086ebde8318/app/models/provider.rb#L23-L26

asmega: nice, i wasn't aware it made a scope too

spec/controllers/api/v2/providers_controller_spec.rb
asmega: I can add a false test. It is kind of implied here https://github.com/DFE-Digital/teacher-training-api/pull/1425/files#diff-deffb905a7d19149f046988a181c4ae1R34 when the value is not provided

asmega:
`suggest` vs `suggest_any` do different things so I assume both need to exist. I'd have to check if publish is using `suggest` before I can delete. That would be outside the scope of this ticket though

PR #1427, opened 2020-06-17
[3656] - Cleanup - remove user state machine

TobyRet:
### Context
- Following the recent work to implement a user state machine on Publish (https://github.com/DFE-Digital/publish-teacher-training/pull/1178) we now no longer need this on the API side

- This PR removes the user state machine along with two V2 user endpoints that are no longer required as all state updates now happen through a single endpoint (`api/V2/user", POST`)

### Guidance to review
Do not merge until https://github.com/DFE-Digital/publish-teacher-training/pull/1178 is merged

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/services/access_request_approval_service.rb
```diff
@@ -12,6 +12,7 @@ def call
       user.first_name      = @access_request.first_name
       user.last_name       = @access_request.last_name
       user.invite_date_utc = Time.now.utc
+      user.state = "new"
```
TobyRet, line 15: Perhaps a better alternative may be to add a default value to the sate column in the user table

db/seeds.rb
```diff
@@ -280,6 +280,7 @@ def create_standard_provider_and_courses_for_cycle(recruitment_cycle, superuser)
     email: Faker::Internet.unique.email,
     first_name: Faker::Name.first_name,
     last_name: Faker::Name.last_name,
+    state: "new"
```
TobyRet, line 283: https://github.com/DFE-Digital/teacher-training-api/pull/1427/files#r441441126

spec/factories/users.rb
```diff
@@ -7,6 +7,7 @@
     welcome_email_date_utc { Faker::Time.backward(days: 1).utc }
     accept_terms_date_utc { Faker::Time.backward(days: 1).utc }
     sign_in_user_id { SecureRandom.uuid }
+    state { "rolled_over" }
```
TobyRet, line 10: https://github.com/DFE-Digital/teacher-training-api/pull/1427/files#r441441126

app/services/access_request_approval_service.rb
gpeng: Or we could make the state machine in the client recognise `nil` as the initial 'new' state. Doing it here or in the database means we have the logic of 'state' in the client and the API or the DB. Not sure that really matters all that much though 🤷

misaka: `nil` makes sense to me, if `aasm` is happy with it.

TobyRet: @misaka @gpeng been mulling over this a bit more and I think the best course of action would be to leave as is (i.e. if a new user is created via mcb or an access request, we manually set the initial state to 'new') or model `aasm`'s behaviour by setting a default for the state column. 

Leaving it as `nil` will create extra complications. Currently in prod we have around 1500 users with state == `new` and none with state == `nil`. This is because the `aasm` on the API sets the default state to `new`. I'm concerned that having new users with one of two states may cause problems.

TobyRet: @gpeng @misaka Looks like this may be possible
https://stackoverflow.com/questions/46988394/dynamically-set-initial-state-with-aasm-gem

TobyRet: https://github.com/DFE-Digital/publish-teacher-training/pull/1178/commits/755cc687bf82c5204e09490e03b29e9cf7eae426

TobyRet: Oh actually, looks like the state column already has a non null constraint (set by AASM). We cannot create users with `state = nil`. Do we want to remove the constraint in light of the above?

TobyRet:
[[M] Support interruption screen](https://trello.com/c/8BkCq4GO/3336-m-support-interruption-screen)

app/services/access_request_approval_service.rb
TobyRet: Not null constraint has now been removed via a db migration

PR #1428, opened 2020-06-17
[3564] Use correct docker image when building bg job images

gpeng:
### Context

The save statistics job is currently failing. Turns out all of the background job runner instances are stuck on code that is a couple of weeks old as they haven't been building correctly.

Pass `GIT_BRANCH` as a build arg to docker and use that rather than `latest` to identify the correct code to build. Also set defaults for the other args. I don't think these are really necessary but we decided to leave them for now to keep the changes down to what is required to fix the issue.

I've cleared the queued jobs and retries from all environments to avoid muddying the statistical waters when this runs properly.

### Guidance to review

I've deployed this branch to QA (Release-1505) and checked the code is running correctly. To test you could either deploy and shell onto the instance and check that this SHA is being built or do the same locally by building the images. Or give the code a once over and take my word for it 🤣 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


docker-compose.yml
```diff
@@ -46,6 +46,7 @@ services:
       args:
         - DOCKER_REPOSITORY=${dockerHubUsername:-dfedigital}
```
defong, line 47: don't think you need that now...
or any of the args looks defaulted in dockerfile

gpeng: Yeah... I don't think so either but we decided to leave it for now and just focus on what was broken.

PR #1431, opened 2020-06-17
[3563] Add endpoint to show training provider

asmega:
### Context

- https://trello.com/c/6S5IQpl1/3563-accredited-bodies-cant-see-the-courses-they-accredit
- The Providers#show endpoint is only accessible to users that are associated to the provider via an organisation
- Therefore it is not suitable for displaying of training provider which should be accessible to accredited bodies 

- Add new endpoint
- This endpoint allows accredited body users to see a specific training
provider
- This endpoint is also accessible to admins
- This prevents ProvidersController#show from being overloaded
- ProvidersController#show is used for only showing direct providers the
user has access to

### Guidance to review

- paired with https://github.com/DFE-Digital/publish-teacher-training/pull/1195
- follow review guidance at https://github.com/DFE-Digital/publish-teacher-training/pull/1195

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -65,6 +65,14 @@ module API
                }
       end
 
+      def show
+        training_provider = @provider.training_providers.find_by(provider_code: params[:training_provider_code])
+
+        authorize training_provider, :can_show_training_provider?
+
+        render jsonapi: training_provider
+      end
+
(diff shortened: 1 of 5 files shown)


asmega:
[Accredited bodies can't see the courses they accredit](https://trello.com/c/6S5IQpl1/3563-accredited-bodies-cant-see-the-courses-they-accredit)

PR #1433, opened 2020-06-18
[3574] Hardcode allocation year to 2020

asmega:
### Context

- https://trello.com/c/KHA3DiHM/3574-hardcode-allocation-year-for-rollover
- Allocations will be present over different cycles and rollover
- However we want allocations to be created in correct cycle

- Hardcode year to `2020`
- This should be controlled by us and not the client
- This will have to be changed once a year, every year

### Guidance to review

- Change the current recruitment cycle
- Create a new allocation
- Allocation should be created for the year `2020`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -1,4 +1,6 @@
 class Allocation < ApplicationRecord
+  ALLOCATION_CYCLE_YEAR = Settings.allocation_cycle_year.to_s.freeze
+
   belongs_to :provider
   belongs_to :accredited_body, class_name: "Provider"
   belongs_to :recruitment_cycle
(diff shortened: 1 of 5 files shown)


asmega:
[Hardcode allocation year for rollover and beyond](https://trello.com/c/KHA3DiHM/3574-hardcode-allocation-year-for-rollover-and-beyond)

app/models/allocation.rb
```diff
@@ -1,4 +1,6 @@
 class Allocation < ApplicationRecord
+  CURRENT_CYCLE_YEAR = "2020".freeze
```
aliuk2012, line 2: is it worth having this as a setting? eg `ALLOCATION_CYCLE_YEAR` so we know its different from the usual current/next cycle?

aliuk2012: Maybe near here https://github.com/DFE-Digital/teacher-training-api/blob/master/config/settings.yml#L6? so its all controlled from one place

asmega: I've made 2 changes:
- Firstly I've renamed the constant so there's no confusion between it and the recruitment cycle year
- This is now configurable via the settings stuff

fofr:
I don't like this.

The allocations requests might be made in the current cycle, but their outcomes will be attached to the wrong cycle.

I also don't like the risk of not updating this.

Can't we tie them to the next cycle by creating an empty next cycle to attach them to? cc @misaka

PR #1356, opened 2020-05-12
Prototype: Expand area covered by University locations

fofr:
Do not merge. Draft PR as part of https://trello.com/c/tZvkZ4Hu/3310-locations-prototype

Make each university location behave as a 5 mile circle, rather than a single point.

PR #1435, opened 2020-06-18
3538 l treat university locations as an area rather than a single point in api

defong:
### Context
based upon #1356 

Added a `boosted_distance` column where by it can be used as for the ordering
This is to improve university course ranking via distance searches
Added an `expand_university` filter flag when it is not set or set as `false` then it will retain current/old behaviour order via `distance`, when set to true then it will use the new `boosted_distance` for ordering.

~~Added provider type to serialiser~~ move to here https://github.com/DFE-Digital/teacher-training-api/pull/1440

### Guidance to review
Best to review commit by commit
Best to compare before and after https://github.com/DFE-Digital/find-teacher-training/pull/353

### Benchmark

command
```sh
time curl "http://localhost:3002/results?fulltime=false&hasvacancies=true&l=1&lat=54.9827086&lng=-1.6572078&loc=Newcastle+upon+Tyne+NE4+9YH%2C+UK&lq=Newcastle+upon+Tyne+NE4+9YH%2C+UK&parttime=false&qualifications%5B%5D=QtsOnly&qualifications%5B%5D=PgdePgceWithQts&qualifications%5B%5D=Other&rad=20&senCourses=false&sortby=2&subjects%5B%5D=31"
```

find (9c72d5fb8791db05c04db275acb6ddaba8be82be)

api

master (88058ede663f006559c23976878ac90f1a82ec37)

curl   0.01s user 0.02s system 0% cpu 2.626 total
curl   0.01s user 0.02s system 0% cpu 2.816 total
curl   0.01s user 0.02s system 0% cpu 2.687 total

3538-l-treat-university-locations-as-an-area-rather-than-a-single-point-in-api (82ccd36b139600fe682b7a6104798aeb0ef39545)

curl   0.01s user 0.03s system 1% cpu 2.453 total
curl   0.01s user 0.02s system 0% cpu 2.654 total
curl   0.01s user 0.02s system 1% cpu 2.486 total

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -40,8 +40,14 @@ class CourseSearchService
       outer_scope = outer_scope.select("provider.provider_name", "course.*")
     elsif sort_by_distance?
       outer_scope = outer_scope.joins(courses_with_distance_from_origin)
-      outer_scope = outer_scope.select("course.*, distance")
-      outer_scope = outer_scope.order(:distance)
+      boosted_distance_column = "
+      (CASE
+        WHEN provider.provider_type = 'O' THEN (distance - 10)
+        ELSE distance
+      END) as boosted_distance"
+      select_criteria = "course.*, distance, #{boosted_distance_column}"
+      outer_scope = outer_scope.joins(:provider).select(select_criteria)
+      outer_scope = outer_scope.order(:boosted_distance)
     end
 
     outer_scope
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -39,7 +39,7 @@ class CourseSearchService
       outer_scope = outer_scope.descending_canonical_order
       outer_scope = outer_scope.select("provider.provider_name", "course.*")
     elsif sort_by_distance?
-      outer_scope = outer_scope.joins(sites_with_distance_from_origin)
+      outer_scope = outer_scope.joins(courses_with_distance_from_origin)
       outer_scope = outer_scope.select("course.*, distance")
       outer_scope = outer_scope.order(:distance)
     end
@@ -51,44 +51,56 @@ class CourseSearchService
 
 private
 
-  def sites_with_distance_from_origin
+  def locatable_sites
     site_status = SiteStatus.arel_table
     sites = Site.arel_table
 
-    # Create virtual table with sites and site statuses
-    sites_with_status = site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
-
-    # only want new and running sites
-    running_and_published_sites = sites_with_status.where(site_status[:status].eq(SiteStatus.statuses[:running]).and(site_status[:publish].eq(SiteStatus.publishes[:published])))
+    # Only running and published site statuses
+    running_and_published_criteria = site_status[:status].eq(SiteStatus.statuses[:running]).and(site_status[:publish].eq(SiteStatus.publishes[:published]))
 
     # we only want sites that have been geocoded
-    geocoded_running_and_published_sites = running_and_published_sites.where(sites[:latitude].not_eq(nil).and(sites[:longitude].not_eq(nil)))
+    has_been_geocoded_criteria = sites[:latitude].not_eq(nil).and(sites[:longitude].not_eq(nil))
 
     # only sites that have a locatable address
     # there are some sites with no address1 or postcode that cannot be
     # accurately geocoded. We don't want to return these as the closest site.
     # This should be removed once the data is fixed
-    locatable_running_and_published_sites = geocoded_running_and_published_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
+    locatable_address_criteria = sites[:address1].not_eq("").or(sites[:postcode].not_eq(""))
 
+    # Create virtual table with sites and site statuses
+    locatable_sites = site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
+
+    locatable_sites = locatable_sites.where(running_and_published_criteria)
+    locatable_sites = locatable_sites.where(has_been_geocoded_criteria)
+    locatable_sites = locatable_sites.where(locatable_address_criteria)
+
+    locatable_sites
+  end
+
+  def course_id_with_lowest_locatable_distance
     # select course_id and nearest site with shortest distance from origin
     # as courses may have multiple sites
     # this will remove duplicates by aggregating on course_id
-    courses_with_nearest_site = locatable_running_and_published_sites.project(:course_id, Arel.sql("MIN#{Site.distance_sql(OpenStruct.new(lat: origin[0], lng: origin[1]))} as distance")).group(:course_id)
+    locatable_sites.project(:course_id, Arel.sql("MIN#{Site.distance_sql(OpenStruct.new(lat: origin[0], lng: origin[1]))} as distance")).group(:course_id)
+  end
 
+  def distance_table
     # form a temporary table with results
     distance_table = Arel::Nodes::TableAlias.new(
       Arel.sql(
-        format("(%s)", courses_with_nearest_site.to_sql),
+        format("(%s)", course_id_with_lowest_locatable_distance.to_sql),
       ), "distances"
     )
 
+    distance_table
+  end
+
+  def courses_with_distance_from_origin
     # grab courses table and join with the above result set
     # so distances from origin are now available
     # we can then sort by distance from the given origin
     courses_table = Course.arel_table
-    courses_table.join(distance_table).on(
-      courses_table[:id].eq(distance_table[:course_id]),
-    ).join_sources
+    courses_table.join(distance_table).on(courses_table[:id].eq(distance_table[:course_id])).join_sources
   end
 
   def locations_filter?
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -91,7 +91,9 @@ private
     # select course_id and nearest site with shortest distance from origin
     # as courses may have multiple sites
     # this will remove duplicates by aggregating on course_id
-    locatable_sites.project(:course_id, Arel.sql("MIN#{Site.distance_sql(OpenStruct.new(lat: origin[0], lng: origin[1]))} as distance")).group(:course_id)
+    origin_lat_long = OpenStruct.new(lat: origin[0], lng: origin[1])
+    lowest_locatable_distance = Arel.sql("MIN#{Site.distance_sql(origin_lat_long)} as distance")
+    locatable_sites.project(:course_id, lowest_locatable_distance).group(:course_id)
   end
 
   def distance_table
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -59,21 +59,21 @@ private
     sites_with_status = site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
 
     # only want new and running sites
-    new_and_running_sites = sites_with_status.where(site_status[:status].in(%w[new_status running]))
+    running_and_published_sites = sites_with_status.where(site_status[:status].eq(SiteStatus.statuses[:running]).and(site_status[:publish].eq(SiteStatus.publishes[:published])))
 
     # we only want sites that have been geocoded
-    geocoded_new_and_running_sites = new_and_running_sites.where(sites[:latitude].not_eq(nil).and(sites[:longitude].not_eq(nil)))
+    geocoded_running_and_published_sites = running_and_published_sites.where(sites[:latitude].not_eq(nil).and(sites[:longitude].not_eq(nil)))
 
     # only sites that have a locatable address
     # there are some sites with no address1 or postcode that cannot be
     # accurately geocoded. We don't want to return these as the closest site.
     # This should be removed once the data is fixed
-    locatable_new_and_running_sites = geocoded_new_and_running_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
+    locatable_running_and_published_sites = geocoded_running_and_published_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
 
     # select course_id and nearest site with shortest distance from origin
     # as courses may have multiple sites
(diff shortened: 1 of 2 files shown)


defong:
[[L] Treat University locations as an area rather than a single point in API](https://trello.com/c/7WXy7DzQ/3538-l-treat-university-locations-as-an-area-rather-than-a-single-point-in-api)

app/services/course_search_service.rb
```diff
@@ -39,9 +39,15 @@ def call
       outer_scope = outer_scope.descending_canonical_order
       outer_scope = outer_scope.select("provider.provider_name", "course.*")
     elsif sort_by_distance?
-      outer_scope = outer_scope.joins(sites_with_distance_from_origin)
-      outer_scope = outer_scope.select("course.*, distance")
-      outer_scope = outer_scope.order(:distance)
+      outer_scope = outer_scope.joins(courses_with_distance_from_origin)
+      boosted_distance_column = "
+      (CASE
+        WHEN provider.provider_type = 'O' THEN (distance - 10)
```
gpeng, line 45: This might be a bit clearer if we put these values in variables e.g.

```
university_provider_type = '0'
university_location_area_radius = 10
```

app/services/course_search_service.rb
```diff
@@ -39,9 +39,15 @@ def call
       outer_scope = outer_scope.descending_canonical_order
       outer_scope = outer_scope.select("provider.provider_name", "course.*")
     elsif sort_by_distance?
-      outer_scope = outer_scope.joins(sites_with_distance_from_origin)
-      outer_scope = outer_scope.select("course.*, distance")
-      outer_scope = outer_scope.order(:distance)
+      outer_scope = outer_scope.joins(courses_with_distance_from_origin)
+      boosted_distance_column = "
```
gpeng, line 43: Would `distance_with_university_area_adjustment` or something like that be clearer. Bit of a mouthful admittedly 😬

app/services/course_search_service.rb
```diff
@@ -51,44 +57,56 @@ def call
...
+    locatable_address_criteria = sites[:address1].not_eq("").or(sites[:postcode].not_eq(""))
+
+    # Create virtual table with sites and site statuses
+    locatable_sites = site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
+
+    locatable_sites = locatable_sites.where(running_and_published_criteria)
+    locatable_sites = locatable_sites.where(has_been_geocoded_criteria)
+    locatable_sites = locatable_sites.where(locatable_address_criteria)
 
+    locatable_sites
```
gpeng, line 83: We don't need this line. 81 will return `locatable_sites`

app/services/course_search_service.rb
```diff
@@ -51,44 +57,56 @@ def call
...
+  def distance_table
     # form a temporary table with results
     distance_table = Arel::Nodes::TableAlias.new(
       Arel.sql(
-        format("(%s)", courses_with_nearest_site.to_sql),
+        format("(%s)", course_id_with_lowest_locatable_distance.to_sql),
       ), "distances"
     )
 
+    distance_table
```
gpeng, line 101: Again, we don't need this or 95 really, just return the table directly.

spec/requests/api/v3/providers/courses/index_spec.rb
```diff
@@ -87,6 +87,7 @@ def perform_request
                 "is_send?" => true,
                 "level" => "primary",
                 "applications_open_from" => provider.courses[0].applications_open_from.strftime("%Y-%m-%d"),
+                "provider_type" => nil,
```
gpeng, line 90: There are a couple of providers with `provider_type == nil` but what does that mean? Should we make sure there's a value in the tests?

spec/services/course_search_service_spec.rb
```diff
@@ -65,9 +66,18 @@
         it "orders in descending order" do
           expect(findable_scope).to receive(:select).and_return(inner_query_scope)
           expect(Course).to receive(:where).and_return(findable_scope)
-          expect(findable_scope).to receive(:joins).and_return(select_scope)
-          expect(select_scope).to receive(:select).and_return(order_scope)
-          expect(order_scope).to receive(:order).and_return(expected_scope)
+          expect(findable_scope).to receive(:joins).and_return(joins_provider_scope)
+          expect(joins_provider_scope).to receive(:joins).with(:provider).and_return(select_scope)
+          boosted_distance_column = "
```
gpeng, line 71: I wonder if this mocking approach has got a bit out of hand in here now. We're really deep in the implementation now rather than the behaviour. That ship has sailed a bit as there are a lot of existing tests in this pattern.

spec/services/course_search_service_spec.rb
```diff
@@ -424,4 +434,64 @@
...
+        Course.all
+      end
+
+      subject do
+        described_class.call(filter: null_island, sort: "distance", course_scope: scope)
+      end
+
+      it "returns correctly" do
+        expect(subject.count(:id)).to eq(2)
+        expect(subject.first.boosted_distance).to eq(-10)
```
gpeng, line 488: Do we want to be allowing negative values for this field. Shouldn't it return 0 if the calculated value is negative?

fofr: If you have many that are within 10 miles, if we make them all 0 won't we lose our ranking ability?

spec/services/course_search_service_spec.rb
defong: This mocking approach is forcing me to enforce chock points, implementation changes then this test must **break**, dare touch the chock points then be prepared to be ran over. 

It is becoming a little bit of a valueless approach as it has obfuscated the intent, given how much domain knowledge is required to comprehend it, the tests just demonstrates vagueness throughout.

Which was also why I wrote the full blown unit test as it is getting elbow deep. 
So we can have more **fun** picking holes like https://github.com/DFE-Digital/teacher-training-api/pull/1435#discussion_r442320179

spec/services/course_search_service_spec.rb
defong: How I see it 
1. the `boosted_distance` is for the `ranking`
2. the `distance` is the actual `calculated` geometry
3. a `perceived distance` is needed for the perception/display to user

ie
`boosted_distance` is a negative value, `perceived distance` is therefore `0`
`boosted_distance` is a positive value, `perceived distance` is therefore `boosted_distance`

spec/requests/api/v3/providers/courses/index_spec.rb
defong: yes need to look into that...

fofr:
@defong Are you able to deploy this branch to the research environment so I can see how the rankings behave?

defong:
> @defong Are you able to deploy this branch to the research environment so I can see how the rankings behave?

@fofr It should be all there now

app/services/course_search_service.rb
```diff
@@ -51,44 +52,64 @@ def call
 
 private
 
-  def sites_with_distance_from_origin
+  def distance_with_university_area_adjustment
+    university_provider_type = Provider.provider_types[:university]
+    university_location_area_radius = 10
+    <<~EOSQL.gsub(/^[\s\t]*/, "").gsub(/[\s\t]*\n/, " ").strip
```
misaka, line 58: Very confusing. How about:

```suggestion
    <<~EOSQL.gsub(/\s+/m, " ").strip
```

or without the regexp:

```suggestion
    <<~EOSQL.tr_s(" \t\n", " ").strip
```

app/services/course_search_service.rb
```diff
@@ -51,44 +52,64 @@ def call
...
     # This should be removed once the data is fixed
-    locatable_new_and_running_sites = geocoded_new_and_running_sites.where(sites[:address1].not_eq("").or(sites[:postcode].not_eq("")))
+    locatable_address_criteria = sites[:address1].not_eq("").or(sites[:postcode].not_eq(""))
+
+    # Create virtual table with sites and site statuses
+    locatable_sites = site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
+
+    locatable_sites = locatable_sites.where(running_and_published_criteria)
+    locatable_sites = locatable_sites.where(has_been_geocoded_criteria)
+    locatable_sites.where(locatable_address_criteria)
```
misaka, line 88: Not to be petty, but could we make this a little bit more Ruby-like:

```ruby
    locatable_sites = site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
                        .where(running_and_published_criteria)
                        .where(has_been_geocoded_criteria)
                        .where(locatable_address_criteria)
```

Or even, since we've already given the name of the result in the method:

```ruby
    site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
      .where(running_and_published_criteria)
      .where(has_been_geocoded_criteria)
      .where(locatable_address_criteria)
```

app/services/course_search_service.rb
```diff
@@ -51,44 +52,64 @@ def call
...
     site_status = SiteStatus.arel_table
     sites = Site.arel_table
 
-    # Create virtual table with sites and site statuses
-    sites_with_status = site_status.join(sites).on(site_status[:site_id].eq(sites[:id]))
-
-    # only want new and running sites
-    new_and_running_sites = sites_with_status.where(site_status[:status].in(%w[new_status running]))
+    # Only running and published site statuses
+    running_and_published_criteria = site_status[:status].eq(SiteStatus.statuses[:running]).and(site_status[:publish].eq(SiteStatus.publishes[:published]))
```
misaka, line 72: I really like how you're breaking down the query into chunks. Would it not be too much if we went just a tiny but further with this one?

```ruby
     running_criteria = site_status[:status].eq(SiteStatus.statuses[:running])
     published_criteria = site_status[:publish].eq(SiteStatus.publishes[:published])
```

And of course you'd need one more `.where(published_criteria)` below.

app/services/course_search_service.rb
```diff
@@ -51,44 +52,64 @@ def call
...
+    locatable_sites = locatable_sites.where(has_been_geocoded_criteria)
+    locatable_sites.where(locatable_address_criteria)
+  end
 
+  def course_id_with_lowest_locatable_distance
     # select course_id and nearest site with shortest distance from origin
     # as courses may have multiple sites
     # this will remove duplicates by aggregating on course_id
-    courses_with_nearest_site = locatable_new_and_running_sites.project(:course_id, Arel.sql("MIN#{Site.distance_sql(OpenStruct.new(lat: origin[0], lng: origin[1]))} as distance")).group(:course_id)
+    locatable_sites.project(:course_id, Arel.sql("MIN#{Site.distance_sql(OpenStruct.new(lat: origin[0], lng: origin[1]))} as distance")).group(:course_id)
```
misaka, line 95: Similar to some of the changes above, can we also break this down to make it a little more readable? Maybe with:

```ruby
    origin_lat_long = OpenStruct.new(lat: origin[0], lng: origin[1])
    site_distance_sql = Site.distance_sql(origin_lat_long)
```

app/services/course_search_service.rb
defong: I'm not sure.

What I really want to enshrined is

https://github.com/DFE-Digital/teacher-training-api/blob/a97b30d65a5e2f357e696d0a5f406e6d1b144fb2/app/models/site_status.rb#L59

Which is a meaningful chunk.. not sure about calling it `findable_criteria` there is already too much unfindable things around.

The given logic got to go through this 
- [findable](https://github.com/DFE-Digital/teacher-training-api/blob/b8c807c3e2f7eb5d2b3438f3ea65be54f9d8a2fb/app/services/course_search_service.rb#L15)

which is this
- [findable](https://github.com/DFE-Digital/teacher-training-api/blob/fdb35d6428d7ad39ad570fe39cba2086ebde8318/app/models/course.rb#L182)

which is really back to this
- [findable](https://github.com/DFE-Digital/teacher-training-api/blob/a97b30d65a5e2f357e696d0a5f406e6d1b144fb2/app/models/site_status.rb#L59)

then its
- [delimited](https://github.com/DFE-Digital/teacher-training-api/blob/b8c807c3e2f7eb5d2b3438f3ea65be54f9d8a2fb/app/services/course_search_service.rb#L29)

before it hits `findable_criteria` and then filtered

app/services/course_search_service.rb
defong:
```suggestion
    origin_lat_long = OpenStruct.new(lat: origin[0], lng: origin[1])
    lowest_locatable_distance = Arel.sql("MIN#{Site.distance_sql(origin_lat_long)} as distance")
    locatable_sites.project(:course_id, lowest_locatable_distance).group(:course_id)
```

May that instead?

app/services/course_search_service.rb
defong: sure I just borrowed the existing style...

spec/requests/api/v3/providers/courses/index_spec.rb
defong: https://github.com/DFE-Digital/teacher-training-api/pull/1440

defong:
Added `do not merge` label until `Find teacher training` as it needs to accommodate the changes

defong:
http://localhost:3002/results?fulltime=false&hasvacancies=true&l=1&lat=52.62640949999999&lng=1.2796911&loc=Norwich+NR2%2C+UK&lq=Norwich+NR2%2C+UK&parttime=false&qualifications[]=QtsOnly&qualifications[]=PgdePgceWithQts&qualifications[]=Other&rad=50&senCourses=false&sortby=2&subjects[]=31

vs https://www.find-postgraduate-teacher-training.service.gov.uk/results?fulltime=false&hasvacancies=true&l=1&lat=52.62640949999999&lng=1.2796911&loc=Norwich+NR2%2C+UK&lq=Norwich+NR2%2C+UK&parttime=false&qualifications[]=QtsOnly&qualifications[]=PgdePgceWithQts&qualifications[]=Other&rad=50&senCourses=false&sortby=2&subjects[]=31


image

defong:
Good to go solo or duel

PR #1439, opened 2020-06-19
[3558] - add 'course vacancies full' notification

TobyRet:
### Context
When a user indicates that a course is full (has no vacancies) a notification will now be sent to notification subscribers.

- Adds 'course vacancies full' emailer
-  Adds test coverage for 'course withdrawal' notification

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/site_statuses_controller.rb
+++ b/app/controllers/api/v2/site_statuses_controller.rb
@@ -7,6 +7,10 @@ module API
         site_status = authorize SiteStatus.find(params[:id])
         site_status.update site_status_params
 
+        if site_status.no_vacancies?
+          NotificationService::CourseVacanciesFilled.call(course: site_status.course)
+        end
+
         render jsonapi: site_status
       end
 
--- a/app/controllers/api/v2/site_statuses_controller.rb
+++ b/app/controllers/api/v2/site_statuses_controller.rb
@@ -5,11 +5,14 @@ module API
 
       def update
         site_status = authorize SiteStatus.find(params[:id])
-        site_status.update site_status_params
 
-        if site_status.no_vacancies?
+        site_status.assign_attributes(site_status_params)
+
+        if site_status.vacancies_filled?
           NotificationService::CourseVacanciesFilled.call(course: site_status.course)
         end
+
+        site_status.save!
 
         render jsonapi: site_status
       end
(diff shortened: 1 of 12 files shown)


app/controllers/api/v2/site_statuses_controller.rb
```diff
@@ -7,11 +7,19 @@ def update
         site_status = authorize SiteStatus.find(params[:id])
         site_status.update site_status_params
 
+        if no_vacancies?
```
asmega, line 10:
```suggestion
        if site_status.no_vacancies?
```
Does this work? If so you can delete private method

TobyRet: Good spot. I've made the change

asmega:
[Turn on other accredited body notifications for "No Vacancies"](https://trello.com/c/IuV3gCaj/3558-turn-on-other-accredited-body-notifications-for-no-vacancies)

app/controllers/api/v2/site_statuses_controller.rb
```diff
@@ -7,6 +7,10 @@ def update
         site_status = authorize SiteStatus.find(params[:id])
         site_status.update site_status_params
 
+        if site_status.no_vacancies?
```
asmega, line 10: do you need to check if it has changed? otherwise for example if there is a course that already have no vacancies and the `status` or `publish` field get changes this would also send out the notification

app/controllers/api/v2/site_statuses_controller.rb
```diff
@@ -7,6 +7,10 @@ def update
         site_status = authorize SiteStatus.find(params[:id])
         site_status.update site_status_params
 
+        if site_status.vacancies_filled?
```
asmega, line 10: not what I had in mind I was thinking of https://api.rubyonrails.org/classes/ActiveModel/Dirty.html
```suggestion
        if site_status.vac_status_changed?(to: "no_vacancies")
```
if inline or probably better belongs in the site_status model

you probably do want to keep the `running` guard which you added

TobyRet: https://github.com/DFE-Digital/teacher-training-api/pull/1439/files#diff-1d30bc581333157c3ffe153a1c241349R88

PR #1443, opened 2020-06-22
[3557] Remove course url from withdrawn notification

SamJamCul:
### Context
https://trello.com/c/uuwGgy2X/3557-turn-on-accredited-body-notifications-for-when-course-is-withdrawn

Turned up in product review, the course url doesn't work for the accredited body seeing as the course is withdraw, so removing it from the notification

Remove the course url stuff from withdrawl notification

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/mailers/course_withdraw_email_mailer.rb
+++ b/app/mailers/course_withdraw_email_mailer.rb
@@ -9,17 +9,8 @@ class CourseWithdrawEmailMailer < GovukNotifyRails::Mailer
       course_name: course.name,
       course_code: course.course_code,
       withdraw_course_datetime: gov_uk_format(datetime),
-      course_url: create_course_url(course),
     )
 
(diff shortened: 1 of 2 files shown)


SamJamCul:
[Turn on accredited body notifications for when course is withdrawn](https://trello.com/c/uuwGgy2X/3557-turn-on-accredited-body-notifications-for-when-course-is-withdrawn)

SamJamCul:
> 
> 
> This is fine. We'll need to remove it from the template in Notify as well if we haven't already. Are we going to make this work in future?

Removed from the notify template - Because the course won't be on Find, we'll need to decide if accredited bodies should be given a link to it on Publish, which is slightly unusual and not something we've done before. Definitely an option though!

PR #1447, opened 2020-06-23
Add db:create to Dockerfile

gpeng:
DO NOT MERGE - this is just to create a test docker image.

Temporary commit to check if rails DB user has create permissions and
stop the container being destroyed for long enough for further testing.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


gpeng:
Dockerhub push is failing. Maybe case sensitivity issue with the branch name.

PR #1453, opened 2020-06-24
3575 paginate organisations index

gpeng:
### Context

The organisations support page in publish loads all organisations and is consequently very slow and ties up the API for too long per request. We are going to add pagination to the page.

* Add pagination to `/api/v2/organisations`
* Add ordering by name as this was previously being done in the client app but can't be now due to pagination

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


gpeng:
[[M] Paginate org#support page](https://trello.com/c/VvZx5lWw/3575-m-paginate-orgsupport-page)

app/controllers/api/v2/organisations_controller.rb
```diff
@@ -2,16 +2,24 @@ module API
   module V2
     class OrganisationsController < API::V2::ApplicationController
       def index
+        per_page = params[:page][:per_page] || 10
```
asmega, line 5: move to a private method?

app/controllers/api/v2/organisations_controller.rb
```diff
@@ -2,16 +2,24 @@ module API
...
         if params_includes_provider?
           current_recruitment_cycle = RecruitmentCycle.current
           @organisations = @organisations.includes(:providers, :users, :nctl_organisations)
                              .where(provider: { recruitment_cycle_id: current_recruitment_cycle.id })
+
         end
 
-        render jsonapi: @organisations, include: params[:include],
-          fields: { providers: %i[provider_code provider_name], users: %i[email sign_in_user_id first_name last_name] }
+        render jsonapi: paginate(@organisations.order(:name), per_page: per_page),
```
misaka, line 16: I notice that we have an N+1 query in here ...

```
2020-06-24 21:10:19.163050 D [96597:puma threadpool 002] (0.821ms) ActiveRecord -- { :sql => "SELECT COUNT(*) FROM \"course\" WHERE \"course\".\"provider_id\" = $1 AND \"course\".\"discarded_at\" IS NULL", :binds => { :provider_id => 13996 } }
2020-06-24 21:10:19.164485 D [96597:puma threadpool 002] (0.414ms) ActiveRecord -- { :sql => "SELECT COUNT(*) FROM \"course\" WHERE \"course\".\"provider_id\" = $1 AND \"course\".\"discarded_at\" IS NULL", :binds => { :provider_id => 13329 } }
2020-06-24 21:10:19.165423 D [96597:puma threadpool 002] (0.307ms) ActiveRecord -- { :sql => "SELECT COUNT(*) FROM \"course\" WHERE \"course\".\"provider_id\" = $1 AND \"course\".\"discarded_at\" IS NULL", :binds => { :provider_id => 14190 } }
...
```

one for each provider. Paging is helping out, but there's also [`Provider.include_course_count`](https://github.com/DFE-Digital/teacher-training-api/blob/master/app/models/provider.rb#L157) which should optimise this, if it'll work in this query. Is it worth adding it in while we're making changes here?

gpeng: That's coming from the `meta` block in the `SerializableProvider`. I think we need to get rid of that which is probably beyond the scope of this PR. Did you have a way of hooking in `Provider.include_course_count`?

app/controllers/api/v2/organisations_controller.rb
gpeng: Good shout. I've done that, and fixed the failure it was also causing. 👍

PR #1237, opened 2020-03-11
[3080] Format date time to our standard format

aliuk2012:
### Changes proposed in this pull request
Update time format from 14:54 to 2:54pm in the course changes notifications emails.

### Checklist

- [X] Make sure all information from the Trello card is in here
- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history
- [X] Tested by running locally

PR #1461, opened 2020-07-01
[3573] Admins can update provider_name

asmega:
### Context

- https://trello.com/c/KXiJdzYT/3573-m-make-provider-name-editable
- Admin users of publish want to be able to update provider names for convenience

- Thru strong params allow `provider_name` for admins
- This is done thru a convention available via pundit

#### Non admins

- Use publish with https://github.com/DFE-Digital/publish-teacher-training/pull/1237
- Login a non-admin
- Navigate to provider contact page to update their details
- Modify the DOM so it sends a new value for `provider[provider_name]`
- Submit
- Provider name should not be update

#### Admins

- Use publish with https://github.com/DFE-Digital/publish-teacher-training/pull/1237
- Login an admin
- Navigate to provider contact page to update their details
- Submit
- Provider name should name should now be updated

#### Code climate

- I don't see these low level changes as being beneficial
- The update endpoint seems to be updating multiple resources, either this endpoint should be split to solve this problem
- Or the resources should be combined so there is no duplication

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -166,6 +166,7 @@ module API
           .except(
             :train_with_us,
             :train_with_disability,
+            :provider_name,
             :email,
             :telephone,
             :website,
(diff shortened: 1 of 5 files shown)


asmega:
[[M] Make provider name editable](https://trello.com/c/KXiJdzYT/3573-m-make-provider-name-editable)

PR #1241, opened 2020-03-11
[3111] Cleanup time formats

faissalMT:
### Context
There are a number of date formats used across TTAPI and it's possible that these could begin diverging in implementation as we proceed with adding notifications.

### Changes proposed in this pull request
DRY up all time date formatting using a module.

### Guidance to review
This provides a place for the date format needed by notifications to live.

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #1463, opened 2020-07-02
[3572] Course titles can be modified by admins

asmega:
### Context

- https://trello.com/c/bKhHARtf/3572-m-make-course-title-editable
- This is the api side on allowing admins to modify course title

- Permitted course parameters moved to policy and depending on whether admin/non-admin returns different fields for mass assignment
- Updating of course now name available in the API and only available to admins
- Added `SerializableCourseWithoutName` to handle existing test cases which simply serialised an entire course and threw it at the api. This can no longer be done for non-admins hence the serialiser was added to remove the `name` field

### Guidance to review

- Using publish https://github.com/DFE-Digital/publish-teacher-training/pull/1241
- Login as admin
- Find a course
- Change the title of the course
- Course title should now be changed
- Attempting the same API request as a non-admin but in correct organisation should not update the course title

### Code climate

- Issues are relating to long lists of fields. Personally I think this is fine for now

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/build_new_course_spec.rb
+++ b/spec/requests/api/v2/build_new_course_spec.rb
@@ -42,7 +42,6 @@ describe "/api/v2/build_new_course", type: :request do
   context "with subjects" do
     let(:params) do
       { course: {
-        name: "Foo Bar Course",
         maths: "must_have_qualification_at_application_time",
         english: "must_have_qualification_at_application_time",
         subjects_ids: subjects.map(&:id),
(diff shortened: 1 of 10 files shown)


asmega:
[[M] Make course title editable](https://trello.com/c/bKhHARtf/3572-m-make-course-title-editable)

spec/requests/api/v2/build_new_course_spec.rb
```diff
@@ -42,7 +42,6 @@
   context "with subjects" do
     let(:params) do
       { course: {
-        name: "Foo Bar Course",
```
asmega, line 45: Under normal circumstance the client is not allowed to send `name` unless they are an admin. The previous code use to accept it then ignore it.

PR #1469, opened 2020-07-09
[3660] Accepting terms returns user json

asmega:
### Context

- https://trello.com/c/sDVFcGD0/3660-update-session-when-user-accepts-terms
- After the user accepts the terms the client is not able to keep in sync with the API

- Accepting terms return json of the updated use
- This is so publish can keep in sync with the API

### Guidance to review

- Hit the accept terms endpoint for a user
- Should now return json of the user

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -20,7 +20,12 @@ module API
 
       def accept_terms
         @user.accept_terms_date_utc = Time.zone.now
-        @user.save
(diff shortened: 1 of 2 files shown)


asmega:
[update session when user accepts terms](https://trello.com/c/sDVFcGD0/3660-update-session-when-user-accepts-terms)

PR #1470, opened 2020-07-10
3661 magic link user not accepted terms

avinhurry:
### Trello Ticket

https://trello.com/c/MmpNfxk3/3661-s-magic-link-user-who-has-not-accepted-terms

### Context/Problem

If magic link is turned on and the user attempts on via magic link and they have not accepted terms (i.e. `accept_terms_date_utc` is nil) then receive 500 on Pubish (403 Foribidden from API). 

This is because the API `UserController` action `generate_and_send_magic_link` is only allowed to be called if the user has accepted terms.  https://github.com/DFE-Digital/teacher-training-api/blob/master/app/controllers/api/v2/users_controller.rb#L6

### To replicate
- enable magic link locally
- update your user in the db to have `accept_terms_date_utc` to nil
- login via magic link
- watch the fire

### To fix
- Update the UserController `check_terms_accepted` skip before action to include `generate_and_send_magic_link`
- Write a spec

It prevents users from getting an error when using magic links if they have not accepted the terms and conditions.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


avinhurry:
[[S] - Magic link: User who has not accepted terms](https://trello.com/c/MmpNfxk3/3661-s-magic-link-user-who-has-not-accepted-terms)

spec/requests/api/v2/users/generate_and_send_magic_link_spec.rb
```diff
@@ -57,4 +57,12 @@ def perform_request
       expect(system_user.magic_link_token).not_to be_nil
     end
   end
+
+  context "when user has not accepted terms and conditions" do
+    let(:user) { create(:user, :inactive) }
+    it "returns the status OK" do
+      perform_request
+      expect(response).to have_http_status(:no_content)
```
defong, line 65: that is not `OK`

avinhurry: @defong fixed

PR #1473, opened 2020-07-13
[3359] - Send 'unrecognised email' notification

TobyRet:
### Context

If a user attempts to sign in via magic link but their email address is unknown, then a notification is sent to that email address saying that the are an unknown user.

### Guidance to review
- Turn magic link on locally on Publish via `development.local.yml'
```
features:
  signin_intercept: true
  signin_by_email: true
  dfe_signin: false
```
- Go to `/signin`
- Sign in via magic link using a made up address (e.g. foo@bar.com)
- API logs should show that an UnrecognisedEmail notification was sent

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/users_controller.rb
+++ b/app/controllers/api/v2/users_controller.rb
@@ -4,6 +4,8 @@ module API
       before_action :build_user, except: :generate_and_send_magic_link
       deserializable_resource :user, only: :update
       skip_before_action :check_terms_accepted, only: %i[accept_terms generate_and_send_magic_link]
+      skip_before_action :authenticate, only: :generate_and_send_magic_link
+      skip_before_action :check_user_is_kept, only: :generate_and_send_magic_link
 
       def show
         render jsonapi: @user,
@@ -31,10 +33,28 @@ module API
       def generate_and_send_magic_link
         skip_authorization
 
-        GenerateAndSendMagicLinkService.call(user: current_user)
+        user = User.where(email: email_from_token).first
+
+        if user
+          GenerateAndSendMagicLinkService.call(user: user)
+        else
+          NotificationService::UnrecognisedEmail.call(email: email_from_token)
+        end
       end
 
     private
+
+      def email_from_token
+        @email_from_token ||= authenticate_or_request_with_http_token do |token|
+          decoded_token = JWT.decode(
+            token,
+            Settings.authentication.secret,
+            Settings.authentication.algorithm,
+            )
+          (decoded_token_payload, _algorithm) = decoded_token
+          decoded_token_payload["email"]
+        end
+      end
 
       def build_user
         @user = authorize User.find(params[:id])
--- a/app/mailers/email_unrecognised_mailer.rb
+++ b/app/mailers/email_unrecognised_mailer.rb
@@ -1,0 +1,7 @@
+class EmailUnrecognisedMailer < GovukNotifyRails::Mailer
+  def email_unrecognised(email)
+    set_template(Settings.govuk_notify.email_unrecognised_template_id)
+
+    mail(to: email)
+  end
+end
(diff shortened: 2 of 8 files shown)


app/controllers/api/v2/users_controller.rb
```diff
@@ -26,11 +28,35 @@ def accept_terms
...
-        GenerateAndSendMagicLinkService.call(user: current_user)
+        # If user is not found current_user returns "HTTP Token: Access denied.\n"
+        current_user = authenticate_or_request_with_http_token do |token|
+          AuthenticationService.new(logger: Rails.logger).execute(token)
+        end
+
+        if current_user.is_a?(User)
+          GenerateAndSendMagicLinkService.call(user: current_user)
+        else
+          send_email_address_unrecognised_email
```
defong, line 39: does this present another  issue of either
* confirming the non existence of email
* generating emails to unknown recipients

app/controllers/api/v2/users_controller.rb
```diff
@@ -26,11 +28,35 @@ def accept_terms
       def generate_and_send_magic_link
         skip_authorization
 
-        GenerateAndSendMagicLinkService.call(user: current_user)
+        # If user is not found current_user returns "HTTP Token: Access denied.\n"
+        current_user = authenticate_or_request_with_http_token do |token|
+          AuthenticationService.new(logger: Rails.logger).execute(token)
```
defong, line 33: can we split this service into smaller chunks?

one to actual do the jwt related stuff
the other the authenticate bit fetching from db etc

then we can get rid of this  `current_user.is_a?(User)`

defong: `generate_and_send_magic_link`

should this method even exist whereby its `skip_before_action :authenticate`
yet do the exact same thing with `authenticate_or_request_with_http_token` which is in here `authenticate`

defong: This actually reads really confusing

we have a `generate_and_send_magic_link` that grabs a `current_user` via `AuthenticationService` by bypassing any of the `authenticate`/`authorization` related 

check if the `current_user` is an actual `user` then do the actual `generate_and_send_magic_link` otherwise do `send_email_address_unrecognised_email`

app/controllers/api/v2/users_controller.rb
TobyRet: I'm not sure I understand sorry ... 
Are you asking whether we should actually be sending emails to unknown recepients?

app/controllers/api/v2/users_controller.rb
TobyRet: I've cleaned it up a bit. Hopefully it reads better now. There is an AuthenticationService that would be idea for moving the JWT decoding logic to however, it's not fit purpose in this instance. I'm not sure if there is value creating a separate service?

TobyRet:
[[M] Add magic link journey for non-existent users](https://trello.com/c/vHEoXZQK/3359-m-add-magic-link-journey-for-non-existent-users)

app/controllers/api/v2/users_controller.rb
defong: > Are you asking whether we should actually be sending emails to unknown recepients?
Yes as its confirming non existence as well as sending email to unknown recipients

app/controllers/api/v2/users_controller.rb
```diff
@@ -31,11 +33,29 @@ def accept_terms
...
+        if user
+          GenerateAndSendMagicLinkService.call(user: user)
+        else
+          NotificationService::UnrecognisedEmail.call(email: email_from_token)
+        end
       end
 
     private
 
+      def email_from_token
```
defong, line 47: lacks validation

TobyRet: Not sure I understand the change you want here 🤔 
Are you saying the email address needs to be validated?

TobyRet: Will add email validation to Publish in separate PR @defong

PR #1474, opened 2020-07-16
[3653] Publish course enrichments validations

asmega:
### Context

- https://trello.com/c/97e3BxBu/3653-missing-broken-error-summaries-when-trying-to-publish-a-course-in-rollover
- When publish a course that does not have an enrichment the error message are not very user friendly

- If a course does not have an enrichment we now present back errors as
if the course did have an enrichment
- This gives more meaningful errors to the user
- Rather than the enrichment is missing

image

### Guidance to review

- Find or create a course without an enrichment
- Attempt to publish it
- Should see errors as seen in attached screenshot
- All errors should link to relevant error or page at least with the exception for age range which will be fixed in another change

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


asmega:
[Missing/broken error summaries when trying to Publish a course in rollover](https://trello.com/c/97e3BxBu/3653-missing-broken-error-summaries-when-trying-to-publish-a-course-in-rollover)

PR #1478, opened 2020-07-21
[3679] - Subject change notification

TobyRet:
### Context
When a user updates a courses subjects we want to trigger a new 'Subjects updated' notification

- 'Subjects updated' notification is triggered when the V2 API receives a request to update a course's subjects
- Any number of subjects can be updated

### 

### Guidance to review
- To test locally you will need to be subscribed to notifications and have a Notify key
- Go to a course that you are subscribed to get course updates for. Edit the courses subjects via 'basic details` tab. Once the update is submitted a subjects updated notification should fire.
- See a real example of the notification on Notify - https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534/notification/1b579e0d-fd6b-4cf9-9806-321cd716c8f1

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -111,11 +111,13 @@ module API
 
           @course.course_subjects.each(&:save)
 
-          NotificationService::CourseSubjectsUpdated.call(
-            course: @course,
-            previous_subject_names: @previous_subject_names,
-            previous_course_name: @previous_course_name,
-          ) if subject_ids
+          if subject_ids
+            NotificationService::CourseSubjectsUpdated.call(
+              course: @course,
+              previous_subject_names: @previous_subject_names,
+              previous_course_name: @previous_course_name,
+            )
+          end
 
           render jsonapi: @course.reload
         else
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -188,6 +188,8 @@ module API
         if request_has_duplicate_subject_ids?
           @course.errors.add(:subjects, :duplicate)
         else
+          previous_subject_names = subject_names_for(@course)
+          previous_course_name = @course.name
           @course.subjects = []
 
           @course.subjects = Subject.find(subject_ids)
@@ -196,8 +198,23 @@ module API
             @course.course_subjects.select { |cs| cs.subject_id == id.to_i }.first.position = index
           end
 
-          @course.name = @course.generate_name
+          updated_subject_names = subject_names_for(@course)
+          updated_course_name = @course.generate_name
+
+          @course.name = updated_course_name
+
+          NotificationService::CourseSubjectsUpdated.call(
+            course: @course,
+            previous_subject_names: previous_subject_names,
+            updated_subject_names: updated_subject_names,
+            previous_course_name: previous_course_name,
+            updated_course_name: updated_course_name,
+          )
         end
+      end
+
+      def subject_names_for(course)
+        course.subjects.map(&:subject_name)
       end
 
       def build_provider
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -111,6 +111,12 @@ module API
 
           @course.course_subjects.each(&:save)
 
+          NotificationService::CourseSubjectsUpdated.call(
+            course: @course,
+            previous_subject_names: @previous_subject_names,
+            previous_course_name: @previous_course_name,
+          ) if subject_ids
+
           render jsonapi: @course.reload
         else
           render jsonapi_errors: @course.errors, status: :unprocessable_entity
@@ -188,28 +194,17 @@ module API
         if request_has_duplicate_subject_ids?
           @course.errors.add(:subjects, :duplicate)
         else
-          previous_subject_names = subject_names_for(@course)
-          previous_course_name = @course.name
-          @course.subjects = []
+          @previous_subject_names = subject_names_for(@course)
+          @previous_course_name = @course.name
 
+          @course.subjects = []
           @course.subjects = Subject.find(subject_ids)
 
           subject_ids.each_with_index do |id, index|
             @course.course_subjects.select { |cs| cs.subject_id == id.to_i }.first.position = index
           end
 
-          updated_subject_names = subject_names_for(@course)
-          updated_course_name = @course.generate_name
-
-          @course.name = updated_course_name
-
-          NotificationService::CourseSubjectsUpdated.call(
-            course: @course,
-            previous_subject_names: previous_subject_names,
-            updated_subject_names: updated_subject_names,
-            previous_course_name: previous_course_name,
-            updated_course_name: updated_course_name,
-          )
+          @course.name = @course.generate_name
         end
       end
 
--- a/app/mailers/course_subjects_updated_email_mailer.rb
+++ b/app/mailers/course_subjects_updated_email_mailer.rb
@@ -4,11 +4,10 @@ class CourseSubjectsUpdatedEmailMailer < GovukNotifyRails::Mailer
   def course_subjects_updated_email(
     course:,
     previous_subject_names:,
-    updated_subject_names:,
     previous_course_name:,
-    updated_course_name:,
     recipient:
   )
+
     set_template(Settings.govuk_notify.course_subjects_updated_email_template_id)
 
     set_personalisation(
@@ -17,9 +16,9 @@ class CourseSubjectsUpdatedEmailMailer < GovukNotifyRails::Mailer
       subject_change_datetime: gov_uk_format(course.updated_at),
       course_url: create_course_url(course),
       previous_subjects: format(previous_subject_names),
-      updated_subjects: format(updated_subject_names),
+      updated_subjects: format(course.subjects.map(&:subject_name)),
       previous_course_name: previous_course_name,
-      updated_course_name: updated_course_name,
+      updated_course_name: course.name,
     )
 
     mail(to: recipient.email)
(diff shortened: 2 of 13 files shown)


app/controllers/api/v2/courses_controller.rb
```diff
@@ -102,6 +102,12 @@ def update
           @course.save
           @course.course_subjects.each(&:save)
 
+          NotificationService::CourseSubjectsUpdated.call(
+            course: @course,
+            previous_subject_names: @previous_subject_names,
```
TobyRet, line 107: Yep. Instance vars. WTF you may ask?
In the notification we need to show the user what the course subjects have changed from and to. Same deal with the course title. When a course's subjects change the course title changes dynamically to reflect the new subjects. We cannot use Rails Dirty API to (easily) to check if course associations have changed and how.

app/controllers/api/v2/courses_controller.rb
```diff
@@ -176,8 +182,10 @@ def update_subjects
         if request_has_duplicate_subject_ids?
           @course.errors.add(:subjects, :duplicate)
         else
-          @course.subjects = []
+          @previous_subject_names = subject_names_for(@course)
```
TobyRet, line 185: Also worth noting that the `update_subjects` method is being called via the `create` action causing further complications / nightmares

app/mailers/course_subjects_updated_email_mailer.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+private
+
+  def format(subject_names)
+    if subject_names.length == 1
+      subject_names.first
+    else
+      subject_names.join(", ")
+    end
```
asmega, line 34: suggestion```
      subject_names.join(", ")
```
Does this not work? Why do you need to treat a collection of 1 different to a collection greater than 1?

File: app/services/notification_service/course_subjects_updated.rb [hunk 4]
```diff
@@ -0,0 +1,42 @@
...
+      @course = course
+      @previous_subject_names = previous_subject_names
+      @previous_course_name = previous_course_name
+    end
+
+    def call
+      return false unless notify_accredited_body?
+      return false unless course.in_current_cycle?
+
+      users = User.joins(:user_notifications).merge(UserNotification.course_update_notification_requests(course.accredited_body_code))
```
asmega, line 19: you can extract this out to a private method instead of using a local variable

app/mailers/course_subjects_updated_email_mailer.rb
TobyRet: Over-site my my part!

app/services/notification_service/course_subjects_updated.rb
TobyRet: Could do but what in way would doing so improve things? I would _extract method_ if:

- doing so made the code more readable (we already have a users variable and it already seems clear that we are pulling users from the db?)
- There was more than one place in the class that was using the same logic. In this case there isn't.

defong:
[[S] Basic detail notification: Subject change](https://trello.com/c/RXZ1AVEO/3679-s-basic-detail-notification-subject-change)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -102,6 +102,14 @@ def update
           @course.save
           @course.course_subjects.each(&:save)
 
+          if subject_ids
+            NotificationService::CourseSubjectsUpdated.call(
+              course: @course,
+              previous_subject_names: @previous_subject_names,
+              previous_course_name: @previous_course_name,
```
aliuk2012, line 109: Not strictly related to this line of code but why do you need to use the previous course name?

aliuk2012: Might be worth checking how @avinhurry & I did course title changed notification https://github.com/DFE-Digital/teacher-training-api/pull/1481. Could this approach perhaps be a better on instead of having another notification template, azure changes etc.

TobyRet: As discussed - in the controller the courses subjects are being re-assigned (determined by what ever subject ids are passed in params). See https://github.com/DFE-Digital/teacher-training-api/pull/1478/files#diff-7d6b888ba704f4684c5b576a1dd07676R190

When updating subjects on a course the name is automatically updated to reflect the change in subjects. This is why the previous course name is being captured in an instance var and passed through to the service and mailer.

app/mailers/course_subjects_updated_email_mailer.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    set_template(Settings.govuk_notify.course_subjects_updated_email_template_id)
+
+    set_personalisation(
+      provider_name: course.provider.provider_name,
+      course_code: course.course_code,
+      subject_change_datetime: gov_uk_format(course.updated_at),
+      course_url: create_course_url(course),
+      previous_subjects: format(previous_subject_names),
+      updated_subjects: format(course.subjects.map(&:subject_name)),
```
misaka, line 19: Makes you wish there was a `course.subjects.subject_names` or something. :)

config/settings/test.yml
```diff
@@ -5,6 +5,7 @@ govuk_notify:
   course_publish_email_template_id: please_change_me
   course_withdraw_email_template_id: please_change_me
   course_vacancies_filled_email_template_id: please_change_me
+  course_subjects_updated_email_template_id: please_change_me
```
misaka, line 8: @aliuk2012 recently demonstrated that there's no need to keep these template IDs secret. Can we start to just put them into the config here? And we'll have a larger ticket later to change the remaining ones?

aliuk2012: They should be in the global `settings.yml` file so they apply across all environments

azure/template.json
```diff
@@ -199,6 +199,13 @@
         "description": "The GOV.UK Notify template id for the magic signin link"
       }
     },
+    "govukNotifyCourseSubjectsUpdatedEmailTemplateId": {
+      "type": "string",
+      "defaultValue": "",
+      "metadata": {
+        "description": "The GOV.UK Notify template id for the course subjects updated email"
+      }
+    },
```
misaka, line 208: We should be able to get rid of these once we add these to the settings.

config/settings.yml
```diff
@@ -13,6 +13,7 @@ govuk_notify:
   magic_link_email_template_id: please_change_me
   course_withdraw_email_template_id: please_change_me
   course_vacancies_filled_email_template_id: please_change_me
+  course_subjects_updated_email_template_id: please_change_me
```
misaka, line 16: The comment on `config/settings/test.yml` should be here, can we put the real templated IDs in here pls.

config/settings/test.yml
misaka: Do we need these in the test env still?

app/services/notification_service/course_subjects_updated.rb
```diff
@@ -0,0 +1,44 @@
...
+        ).deliver_later(queue: "mailer")
+      end
+    end
+
+  private
+
+    attr_reader :course, :previous_subject_names, :previous_course_name
+
+    def users
+      @users ||= User.joins(:user_notifications).merge(UserNotification.course_update_notification_requests(course.accredited_body_code))
```
misaka, line 34: Would it make sense to move this into `User`? Normally it makes sense for a service to coordinate between multiple models, but in this case the `UserNotification` is kind of an extension to `User`.

TobyRet: Good idea @misaka 
Done -> https://github.com/DFE-Digital/teacher-training-api/pull/1478/commits/3b48730bd9a2f20d10ae0f1d5374ac8016d1a2ed

spec/services/notification_service/course_subjects_updated_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+        it "does not send a notification" do
+          expect(CourseSubjectsUpdatedEmailMailer).to_not receive(:course_subjects_updated_email)
+          expect(course.recruitment_cycle).to_not eql(RecruitmentCycle.current)
+          described_class.call(
+            course: course,
+            previous_subject_names: previous_subject_names,
+            previous_course_name: previous_course_name,
+          )
+        end
+      end
```
misaka, line 85: I'm finding these tests really hard to read. Now that I've been staring at them for a bit I feel like I can kind of understand what's being tested, but I can't help but wonder if it wouldn't be better organised if we used the pattern where we do call the service in a before block, and then do the setup and testing separately. Something like:

```ruby
    describe "#call" do
      # all the let blocks

      before do
        setup_notifications

        described_class.call(
          course: course,
          previous_subject_names: subject_names,
          previous_course_name: course_name,
        )
      end

      describe CourseSubjectsUpdatedEmailMailer do
        before do
          allow(CourseSubjectsUpdatedEmailMailer).to receive(:course_subjects_updated_email)
        end

        context "with a course that is in the current cycle" do
          it { is_expected.to receive(:course_subjects_updated_email) }
        end

        context "with a course that is not in the current cycle" do
          let(:provider) { create(:provider, :next_recruitment_cycle) }
          let(:course) { create(:course, accredited_body_code: accredited_body.provider_code, provider: provider) } 

          it { is_expected.to_not receive(:course_subjects_updated_email) }
        end
      end

      # and the rest of the specs modified similarly
```

misaka: Alternatively, it seems that this code is used each time to trigger the work needed for the test:

```ruby
described_class.call(
  course: course,
  previous_subject_names: previous_subject_names,
  previous_course_name: previous_course_name,
)
```

I find it's hard to parse that while also keeping all the context in my head ... so maybe turn that into a method that makes is clear that it's the same call throughout:

```ruby
def call_service
  described_class.call(
    course: course,
    previous_subject_names: previous_subject_names,
    previous_course_name: previous_course_name,
  )
end


it "does not send a notification" do
  expect(CourseSubjectsUpdatedEmailMailer).to_not receive(:course_subjects_updated_email)
  expect(course.recruitment_cycle).to_not eql(RecruitmentCycle.current)
  call_service
end 
```

TobyRet: Yep cool - have made the changes

PR #1482, opened 2020-07-23
[3684] Store request id from header for V3 api

asmega:
### Context

- https://trello.com/c/2jQCohxd/3684-add-requestid-to-logging-for-find
- V3 api was not taking the `X-Request-Id` header from the request. Therefore a new `request_id` was being generated and this would not correlate with request ids from Find

- V3 requests store request_id from header

### Guidance to review

- Checkout this PR
- `curl -H "X-Request-Id: foo" "http://localhost:3001/api/v3/subjects?fields%5Bsubjects%5D=subject_name%2Csubject_code&sort=subject_name"`
- add `binding.pry` after app/controllers/api/v3/application_controller.rb:44
- `request.uuid` should be `foo`
- alternatively if you want to check the logs you will have to enable the tagged logger for the dev environment

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/application_controller.rb
+++ b/app/controllers/api/v3/application_controller.rb
@@ -3,6 +3,7 @@ module API
     class ApplicationController < ActionController::API
       rescue_from ActiveRecord::RecordNotFound, with: :jsonapi_404
 
+      before_action :store_request_id
       before_action :check_disable_pagination
 
       def jsonapi_404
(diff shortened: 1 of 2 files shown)


asmega:
[Add `request_id` to logging for Find](https://trello.com/c/2jQCohxd/3684-add-requestid-to-logging-for-find)

PR #1483, opened 2020-07-24
[3678] - Course vacancies open (all locations)

TobyRet:
### Context
Currently, subscribed users are only being notified if a course has no vacancies (all locations).

See associated Publish PR - https://github.com/DFE-Digital/publish-teacher-training/pull/1269

- When a user updates course locations to from closed to open we want to trigger the vacancies updated notification. Specifically, the notification is triggered when a course has a single location and it is updated from closed to open, or when the course has multiple locations and all locations are changed to open.

### Examples you can look at
- Vacancies open notification - https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534/notification/c9a7b3a6-5c58-4287-9448-55a93c9f5436
- Vacancies closed notification - https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534/notification/6b894262-57a0-40ba-87d3-807ba677ca56

### Guidance to review
- A new template has been created to handle the vacancies closed / open scenarios. This template uses optional logic so may be worth considering if this is still readable or if separate templates are better. See https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534/templates/3ae884e9-8495-44cf-9928-907b89a9f356
- If we do go with the above template then the old `vacancies filled` template can be deleted - https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534/templates/6cc1d88e-a130-4112-8237-2c9a4e0087a0. A separate PR will be raised to remove the template id vars from Azure.

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


TobyRet:
[Vacancies open - (all open)](https://trello.com/c/FXm1vJhZ/3678-vacancies-open-all-open)

app/controllers/api/v2/courses_controller.rb
```diff
@@ -133,9 +133,12 @@ def create
         create_new_course
       end
 
-      def send_vacancies_filled_notification
+      def send_vacancies_updated_notification
```
asmega, line 136: why is this endpoint needed? Should publish not just update the course and the api works out if a notification should be sent? Why does the client itself need to know state and then decide if it should send a notification or not by hitting this endpoint?

TobyRet: I'm going to get @defong and @SamJamCul to weigh in on this as they may have some more historical context. 

Whether a course actually has vacancies or not is actually determined by the SiteStatus resource - https://github.com/DFE-Digital/publish-teacher-training/pull/1242/files#diff-30720a018bf9d1c8a2a6274262bfb656R66. 

One option would be to create an endpoint for modifying a collection of SiteStatuses, else we need to check if a course has vacancies every time a single site status is updated.

asmega: So one option I considered is making this a callback so when a `SiteStatus` is update it pings the course to see if it should send the notification. This would keep things API side but i think this will have race conditions depending on the order the site statuses get updated.

So could at the very least not keep the logic api side. eg
```ruby
vacancies_filled: params[:course][:vacancies_filled]
```

Why does the client (publish) need to tell the api if the vacancies of a course has been filled? The API should know this surely and determine this and not the client? 

It looks like the data has been shipped to publish, publish does this determination, then sends it back to the API.

asmega: But yes I think the best solution would be to have an endpoint that can update a collection of resources in a transaction as you mentioned

misaka: Ya, we've tried this before, and it's been a world of hurt. JSONAPI doesn't support this, you have to update resources individually. You _can_ update what the collection of resources that are associated to another resource, but that doesn't update the resources in the collection themselves, that just updates the association.

I suggest that we stick with updating resources individually and triggering a notification from publish, I don't think that that's too terrible given the limitations we're working with, but I do think that this adds ammunition to the argument that we should just merge publish and api, which I think suddenly makes more sense now that we know better where Find is going, etc.

PR #1487, opened 2020-07-29
[3752] Use `api` instead `api2`

aliuk2012:
### Context
In the api docs, we referred to the endpoint as `api2` instead we should
be referring to `api`.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -33,7 +33,7 @@ Authorization: Bearer your_api_key
 ## Example
 
 ```shell
-curl "https://api2.publish-teacher-training-courses.service.gov.uk/api/v1/2019/subjects" -H "Authorization: Bearer your_api_key"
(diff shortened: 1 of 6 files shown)


aliuk2012:
[Don't link `https://api2.publish-teacher-training-courses.service.gov.uk/api/public/{version}` as it's not a valid link. We can probably fix this link so that it does work.](https://trello.com/c/WuNzsiIS/3752-dont-link-https-api2publish-teacher-training-coursesservicegovuk-api-public-version-as-its-not-a-valid-link-we-can-probably-fix)

aliuk2012:
@misaka I was wondering about the dns changes. As we haven't published these docs could we still merge this for the time being?

aliuk2012:
Blocked until DNS has been done

PR #1489, opened 2020-07-29
[3753] Responses & Responses eg's produces same id

aliuk2012:
### Context
We were using malformed HTML where the "Responses" section had the same
id as "Responses Example". This meant that the left side nav was highlighting
both entries because their URL was the same.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


aliuk2012:
[Fix `Responses` and `Response Examples` being highlighted together in LHS nav.](https://trello.com/c/Z55Y3vrC/3753-fix-responses-and-response-examples-being-highlighted-together-in-lhs-nav)

PR #1490, opened 2020-07-30
3682 basic detail notification ucas apply

JonF1:
### Context
https://trello.com/c/vDDbpmtB/3682-m-basic-detail-notification-ucas-apply
Multiple emails are now sent. One per attribute being changed. 

- Added a loop to course update attribute changes
- Added locale entry requirement value changes

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/services/course_attribute_formatter_service_spec.rb
+++ b/spec/services/course_attribute_formatter_service_spec.rb
@@ -1,4 +1,4 @@
-require_relative "../../app/services/course_attribute_formatter_service"
+require "rails_helper"
 
 RSpec.describe CourseAttributeFormatterService do
   subject { CourseAttributeFormatterService.call(name: name, value: value) }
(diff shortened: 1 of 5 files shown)


JonF1:
[[M] Basic detail notification: UCAS Apply](https://trello.com/c/vDDbpmtB/3682-m-basic-detail-notification-ucas-apply)

spec/services/course_attribute_formatter_service_spec.rb
```diff
@@ -1,4 +1,5 @@
 require_relative "../../app/services/course_attribute_formatter_service"
```
asmega, line 1: You probably don't need this line anymore since you added the one below it

JonF1: We've removed this

aliuk2012:
What happens when all three attributes are updated? I thought we proved on Wednesday that only the first attribute email gets sent

PR #1493, opened 2020-07-31
[3738] Prettify HTTP method and path headings

aliuk2012:
### Context
I made the HTTP method uppercase and styled as a caption because unlike
Apply's API our URL are longer and so they would automatically wrap to a
new line.

## Before 

image

## After

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/docs/lib/govuk_tech_docs/open_api/renderer.rb
+++ b/docs/lib/govuk_tech_docs/open_api/renderer.rb
@@ -41,7 +41,7 @@ module GovukTechDocs
       def path(text)
         path = @document.paths[text]
         id = text.parameterize
-        operations = operations(path, id)
+        operations = operations(text: text, path: path, path_id: id)
         @template_path.result(binding)
       end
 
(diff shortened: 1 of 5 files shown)


avinhurry:
[Fix issues in tech docs](https://trello.com/c/T21mf8h7/3738-fix-issues-in-tech-docs)

docs/lib/govuk_tech_docs/open_api/renderer.rb
```diff
@@ -122,11 +122,12 @@ def schemas_from_schema(schema)
         schemas
       end
 
-      def operations(path, path_id)
+      def operations(text, path, path_id)
```
TobyRet, line 125: Is it potentially worth passing these args in as a hash now given two of the params are strings?

darokel: or keyword args?

PR #1494, opened 2020-07-31
[3716] - CourseAttributeFormatterService - return 'unknown' when value is nil

TobyRet:
### Context
- A course may have certain attributes (e.g.like 'age_range_in_years) that are nil. In this scenario the `CourseAttributeFormatterService` breaks and the course update notification fails to send

- Populate the notify template with a value regardless ('unknown') so that it does not break.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_attribute_formatter_service.rb
+++ b/app/services/course_attribute_formatter_service.rb
@@ -58,6 +58,10 @@ private
   end
 
   def strip_underscores
-    value.tr("_", " ")
+    if value
+      value.tr("_", " ")
+    else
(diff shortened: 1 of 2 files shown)


TobyRet:
[[S] No email sent if previous age range was not set.](https://trello.com/c/jngA616d/3716-s-no-email-sent-if-previous-age-range-was-not-set)

asmega:
Needs rebasing :)

PR #1504, opened 2020-08-06
3736 - implemented the endpoint for the provider

avinhurry:

Added documentation for the null provider 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/application_controller.rb
+++ b/app/controllers/api/public/v1/application_controller.rb
@@ -2,6 +2,11 @@ module API
   module Public
     module V1
       class ApplicationController < ActionController::API
+        rescue_from ActiveRecord::RecordNotFound, with: :jsonapi_404
+
+        def jsonapi_404
+          render jsonapi: nil, status: :not_found
+        end
       end
     end
   end
--- a/app/controllers/api/public/v1/providers_controller.rb
+++ b/app/controllers/api/public/v1/providers_controller.rb
@@ -9,7 +9,7 @@ module API
                       else
                         providers.by_name_descending
                       end
-          render jsonapi: providers, class: { Provider: API::Public::V1::SerializableProvider }, fields: { providers: provider_fields }
+          render jsonapi: providers, class: { Provider: API::Public::V1::SerializableProvider }, fields: fields
         end
 
         def show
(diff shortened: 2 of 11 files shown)


avinhurry:
[[S] Implement API endpoint - /api/public/v1/recruitment_cycles/:year/providers/:provider_code](https://trello.com/c/zI3kyVcI/3736-s-implement-api-endpoint-api-public-v1-recruitmentcycles-year-providers-providercode)

asmega:
Just a note your `include` won't work because theres is only one serializer specified, we will have to swap this out once we implement more. It also won't work because haven't specified `belongs_to` on the provider serializer

app/controllers/api/public/v1/providers_controller.rb
```diff
@@ -13,19 +13,16 @@ def index
...
+          code = params.fetch(:code, params[:provider_code])
+          provider = recruitment_cycle.providers
+                                        .find_by!(
+                                          provider_code: code.upcase,
+                                        )
+
+          render jsonapi: provider,
+                 class: { Provider: API::Public::V1::SerializableProvider },
+                 include: params[:include],
+                 fields: { providers: provider_fields }
```
asmega, line 25: not something made in this PR but `provider_fields` should return nil if nothing is specified. The default behaviour seems to serialize everything defined in the serializer if it is nil, therefore there is no need to have a default provider fields passed in.

app/controllers/api/public/v1/application_controller.rb
```diff
@@ -2,6 +2,11 @@ module API
   module Public
     module V1
       class ApplicationController < ActionController::API
+        rescue_from ActiveRecord::RecordNotFound, with: :jsonapi_404
```
darokel, line 5: I think we need a standardised way of dealing with these types of errors. What about something like:

```
    def render_json_error(code: nil, message:, status: :internal_server_error)
      render json: { code: code, error: message }.compact, status: status
    end
```

app/controllers/api/public/v1/providers_controller.rb
avinhurry: Done, thanks.

app/controllers/api/public/v1/application_controller.rb
avinhurry: I've added a card for this to my ideas bored. We can chat about it

PR #1505, opened 2020-08-07
[3747] Document courses endpoint by year

darokel:
### Context

- https://trello.com/c/Zth5QQPE/3747-document-api-endpoint-api-public-v1-recruitmentcycles-year-courses
- Document `/api/public/v1/recruitment_cycles/:year/courses`

- Document the public courses endpoint scoped by year
- Move the existing courses endpoint scoped under a `providers` module
- Update `api_spec.json` to reflect current state of yaml

### Guidance to review

- Run the middleman docs, see the new endpoint documentation for courses by year
- Discuss whether it's worth making the endpoint snake_case rather than kebab-case to be consistent with other endpoints

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


darokel:
[Document API endpoint - /api/public/v1/recruitment_cycles/:year/courses](https://trello.com/c/Zth5QQPE/3747-document-api-endpoint-api-public-v1-recruitmentcycles-year-courses)

asmega:
So If we do it this way it means we'll have to manage 2 courses controller depending on the level of nesting

darokel:
> So If we do it this way it means we'll have to maintain 2 courses controllers. One for year and one for provider

I think that's fine as they will potentially start to deal with different concerns. Otherwise we'd have to mix in logic to handle different constraints (year, provider etc) in one controller/action.

spec/api/courses_spec.rb
```diff
@@ -46,50 +40,13 @@
                 example: { page: 2, per_page: 10 },
                 description: "Pagination options to navigate through the collection."
 
-      response "200", "The collection of courses." do
+      response "200", "The collection of courses for the specified year." do
```
defong, line 43:
```suggestion
      response "200", "The collection of courses." do
```
based on 
https://api2.qa.publish-teacher-training-courses.service.gov.uk/api-reference.html#recruitment_cycles-year-providers-get

spec/api/courses_spec.rb
```diff
@@ -1,23 +1,17 @@
 require "swagger_helper"
 
 describe "API" do
-  path "/recruitment_cycles/{year}/providers/{provider_code}/courses" do
-    get "Returns the courses for the specified provider." do
-      operationId :public_api_v1_provider_courses
-      tags "course"
+  path "/recruitment_cycles/{year}/courses" do
+    get "Returns the courses for the specified year." do
```
defong, line 5:
```suggestion
    get "Returns the courses for the specified recruitment cycle." do
```

based on 
https://api2.qa.publish-teacher-training-courses.service.gov.uk/api-reference.html#recruitment_cycles-year-providers-get

PR #1509, opened 2020-08-10
Fixes issue with rspec parallel xml outputs.

aliuk2012:
Our currently implementation seems not to be working anyway. To simplify
our process we are now using `.rspec_parallel` as a single config file
Azure should be able to use this output to report test coverage in azure.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


azure-pipelines.yml
```diff
@@ -1,207 +0,0 @@
-trigger:
```
asmega, line 1: i'm pretty sure you can't delete this file

aliuk2012: 🤦  I was not meant to delete this. Thanks for spotting it

.rspec_parallel
```diff
@@ -1,3 +1,5 @@
 --require rails_helper
 --format progress
 --format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
+--no-profile --format RspecJunitFormatter --out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml
```
asmega, line 4: i'm not sure the `--no-profile` will work here? regardless it doesn't seem to cause an issue

PR #1511, opened 2020-08-11
[3781] Add ADR to remove before_action variables

asmega:
### Context

- https://trello.com/c/xKBtll4v/3781-adr-for-removal-of-beforeaction-instance-variable-loading

- Raise ADR to swap `before_action` loading of instance variables with lazy loaded memoised private methods

### Guidance to review

- Read ADR

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/docs/adr/0004-before-action-variables.md
+++ b/docs/adr/0004-before-action-variables.md
@@ -1,0 +1,91 @@
+# 4. Before action variables
+
+Date: 2020-08-11
+
+## Status
+
+Accepted
+
+## Context
+
+This code base currently has a lot of `before_action` in controllers. For
+example:
+
+- https://github.com/DFE-Digital/teacher-training-api/blob/6e18d28cafe3d4595a15eb6670cc6950ddf692b8/app/controllers/api/v2/courses_controller.rb#L6
+
+These `before_action` calls typically load data into an instance variable
+which are then used later.
+
+However this can become unwieldy when the controller has multiple actions.
+Some actions require one set of variables whilst another action may not and
+there are resulting overlaps between the variables in use.
+
+Therefore to manage the actions will typically have `only` or `except` options
+set to control these conditions.
+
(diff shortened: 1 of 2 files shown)


asmega:
[ADR for removal of before_action instance variable loading](https://trello.com/c/xKBtll4v/3781-adr-for-removal-of-beforeaction-instance-variable-loading)

docs/adr/0004-before-action-variables.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### Pros
+
+- No work to change existing code.
+
+#### Cons
+
+- Persistence of overhead of understanding variables loaded in controllers.
+
+### 1. Use lazy loaded memoised private methods
```
misaka, line 49: `2.` ?

misaka:
@asmega Worth an addition to the README at the same time?

docs/adr/0004-before-action-variables.md
asmega: whoops, yea i did fix this but forgot to push

docs/adr/0004-before-action-variables.md
```diff
@@ -0,0 +1,91 @@
...
+```ruby
+def index
+  render jsonapi: courses
+end
+
+private
+
+def courses
+  @courses ||= Course.all
+end
```
tijmenb, line 67: This has the side effect of setting the `@courses` instance variable, which can be used in the views. 

A common pattern to complement the above is to prefix the instance variable with a underscore to signify it's private to the controller.

```rb
def courses
  @_courses ||= Course.all
end
```

asmega: It should be ok here as there is no real "view" layer as this is an API. But yes for a "standard" rails app that is probably the way to go

PR #1516, opened 2020-08-12
[3737] Change CourseSearchService base scope

asmega:
### Context

- https://trello.com/c/Igxd218k/3737-m-implement-api-endpoint-api-v3-recruitmentcycles-year-providers-providercode-courses
- The CourseSearchService is currently constrained to `findable` courses only
- We want to open up this constraint so the class can also be used to search for non-findable courses too
- This will be required for the public api

- This allows for greater flexbility and does not limit the class to
only return findable courses
- The onus is on the caller to pass correct base scope
- This means this class can now be used to search non-findable courses
too

### Guidance to review

- v3 api should remain the same. therefore find search functionality should remain the same
- should be able to call CourseSearchService class with a base scope and should return non findable courses too

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
+++ b/app/controllers/api/v2/accredited_provider_training_providers_controller.rb
@@ -28,7 +28,7 @@ module API
           provider_courses = Course.where(provider: [provider.id])
           training_provider_courses = Course.where(provider: training_providers).where(accredited_body_code: provider.provider_code)
 
-          provider_courses.or(training_provider_courses)
+          provider_courses.or(training_provider_courses).findable
         end
 
         def eligible_training_provider_ids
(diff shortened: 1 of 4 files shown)


asmega:
[[M] Implement API endpoint - /api/v3/recruitment_cycles/:year/providers/:provider_code/courses](https://trello.com/c/Igxd218k/3737-m-implement-api-endpoint-api-v3-recruitmentcycles-year-providers-providercode-courses)

asmega:
[Remove findable constraint from CourseSearchService](https://trello.com/c/XhIbOKE9/3793-remove-findable-constraint-from-coursesearchservice)

PR #1517, opened 2020-08-12
[3737] Start public/v1 provider courses endpoint

asmega:
### Context

- https://trello.com/b/fXA6ioZN/find-publish-sprint-board
- This starts the implementation of public/v1 api endpoint for provider courses#index
- This PR has missing features such as sorting and filtering and will be added in future PRs

- Implement basic controller endpoint for provider courses

### Guidance to review

- Sorting is out of scope of this PR
- Filtering is out of scope of this PR
- Included associations is out of scope of this PR
- Fields returned is out of scope of this PR
- Hit the API endpoint, should returns the providers courses

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/serializers/api/public/v1/serializer_service.rb
+++ b/app/serializers/api/public/v1/serializer_service.rb
@@ -1,0 +1,14 @@
+module API
+  module Public
+    module V1
+      class SerializerService
+        def call
+          {
+            Course: API::Public::V1::SerializableCourse,
+            Provider: API::Public::V1::SerializableProvider,
+          }
+        end
+      end
+    end
+  end
+end
--- a/spec/controllers/api/public/v1/providers/courses_controller_spec.rb
+++ b/spec/controllers/api/public/v1/providers/courses_controller_spec.rb
@@ -1,0 +1,64 @@
+require "rails_helper"
+
+RSpec.describe API::Public::V1::Providers::CoursesController do
+  let(:provider) { create(:provider) }
+  let(:recruitment_cycle) { provider.recruitment_cycle }
+
+  describe "#index" do
+    context "when there are no courses" do
+      before do
+        get :index, params: {
+          recruitment_cycle_year: recruitment_cycle.year,
+          provider_code: provider.provider_code,
+        }
+      end
+
+      it "returns empty array of data" do
+        expect(JSON.parse(response.body)["data"]).to eql([])
+      end
+    end
+
+    context "when there are courses" do
+      before do
+        create(:course, provider: provider)
+        create(:course, provider: provider)
+
+        get :index, params: {
+          recruitment_cycle_year: "2020",
+          provider_code: provider.provider_code,
+        }
+      end
+
+      it "returns correct number of courses" do
+        expect(JSON.parse(response.body)["data"].size).to eql(2)
+      end
+    end
+
+    describe "pagination" do
+      let(:courses) do
+        array = []
+
+        7.times { |n| array << build(:course, id: n + 1) }
+
+        array
+      end
+
+      before do
+        allow(controller).to receive(:courses).and_return(courses)
+
+        get :index, params: {
+          recruitment_cycle_year: "2020",
+          provider_code: "ABC",
+          page: {
+            page: 2,
+            per_page: 5,
+          },
+        }
+      end
+
+      it "can pagingate to page 2" do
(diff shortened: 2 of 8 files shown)


asmega:
[[M] Implement API endpoint - /api/v3/recruitment_cycles/:year/providers/:provider_code/courses](https://trello.com/c/Igxd218k/3737-m-implement-api-endpoint-api-v3-recruitmentcycles-year-providers-providercode-courses)

app/serializers/api/public/v1/serializer_service.rb
```diff
@@ -0,0 +1,14 @@
+module API
+  module Public
+    module V1
+      class SerializerService
+        def execute
+          {
+            Course: API::Public::V1::SerializableCourse,
+            Provider: API::Public::V1::SerializableProvider,
```
misaka, line 8: Just to note, we don't serialize the provider in this PR, so we don't need this. We will need to add `recruitment_cycle` and `locations`, however, when we add support for inclusion.

Optional: Remove `Provider` here and re-add it when we add inclusion.

asmega: It will get re-added soon, so i do see the benefit of removing it here. I don't see any negative side effects of this remaining

asmega:
> * Are we planning to put includes, ordering and filtering in here? Or separate tickets? I vote we get this in and make separate tickets.

From the meeting we said we'd break this down. I'll keep one ticket but raise multiple PRs

> * I've noticed recently we have `/api/v3/recruitment_cycles/:year/courses`, I forget if we need that here or not, but I would guess we might. In which case, we should have a way of retrieving courses if there's no `provider_id`.

Theres another controller for this, we can point to this controller if needed or do another implementation. Regardless that is outside the scope of this change and we'll deal with that when that endpoint is implemented

app/serializers/api/public/v1/serializer_service.rb
```diff
@@ -0,0 +1,14 @@
+module API
+  module Public
+    module V1
+      class SerializerService
+        def execute
```
defong, line 5: so `call` vs `execute`?

see https://github.com/DFE-Digital/teacher-training-api/pull/1521

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -0,0 +1,57 @@
+require "rails_helper"
+
+RSpec.describe API::Public::V1::Providers::CoursesController do
+  let(:provider) { create(:provider) }
+  let(:recruitment_cycle) { provider.recruitment_cycle }
+
+  describe "#index" do
+    context "when there are no courses" do
+      it "returns empty array of data" do
+        get :index, params: {
```
darokel, line 10: can we put the `get` in a before block that way we setup the tests in a an arrange/act/assert structure. that way the it blocks are just asserting on what has been arranged/performed?

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    context "when there are courses" do
+      before do
+        create(:course, provider: provider)
+        create(:course, provider: provider)
+      end
+
+      it "returns correct number of courses" do
+        get :index, params: { recruitment_cycle_year: "2020", provider_code: provider.provider_code }
```
darokel, line 25: same for this eg https://github.com/DFE-Digital/teacher-training-api/pull/1517/files#r470030625

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        array
+      end
+
+      before do
+        allow(controller).to receive(:courses).and_return(courses)
+      end
+
+      it "can pagingate to page 2" do
+        get :index, params: {
```
darokel, line 44: and this?

app/serializers/api/public/v1/serializer_service.rb
asmega: done

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
asmega: cool, i've changed all these

PR #1519, opened 2020-08-12
[3788] Move GOV.UK Notify template id out of azure

aliuk2012:
-https://trello.com/c/sZ4woLkJ/3788-remove-govuk-notify-template-ids-into-repository-settings 

There is no reason for us to hide our template ID and use Azure to set
them. We can leave them in our source code so long as we still key the
api key secure then a potential attacker will not be able to use that
template.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -510,34 +510,6 @@
                 "value": "[parameters('settingsAuthenticationSecret')]"
               },
               {
-                "name": "SETTINGS__GOVUK_NOTIFY__API_KEY",
-                "value": "[parameters('govukNotifyApiKey')]"
-              },
-              {
-                "name": "SETTINGS__GOVUK_NOTIFY__WELCOME_EMAIL_TEMPLATE_ID",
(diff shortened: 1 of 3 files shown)


azure/template.json
```diff
@@ -509,34 +509,6 @@
                 "name": "SETTINGS__AUTHENTICATION__SECRET",
                 "value": "[parameters('settingsAuthenticationSecret')]"
               },
-              {
```
dankmitchell, line 512: I think this one needs to remain

aliuk2012: From the trello ticket:
>Additionally, `SETTINGS__GOVUK_NOTIFY__API_KEY` is only required in the `environmentVariables` section as it's only used in the background mailer process. It can be removed from the `appServiceAppSettings` section.

dankmitchell:
[Remove GOVUK Notify template IDs into repository settings](https://trello.com/c/sZ4woLkJ/3788-find-and-move-non-secret-settings-into-repository-settings)

PR #1521, opened 2020-08-13
add ADR "Service Entry Points"

misaka:
Had a conversation in Slack about this, again, which served as a reminder to create this ADR and update the services REAMDE.

https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1597310908061300

### DIFF
--- a/app/services/README.md
+++ b/app/services/README.md
@@ -5,8 +5,8 @@ a single place that instantiates all the services related to a model
 
 ## Conventions
 
-- Use the `execute` method as the interface on the service
-- Use named arguments to call `execute` to explicitly state what the arguments going in are
+- Use the `call` method as the interface on the service
+- Use named arguments to call `call` to explicitly state what the arguments going in are
 - Services with logic related to a single instance of a model specifically are executed on the model
 - Services related to multiple models or collections of a model are executed outside of the model (e.g. 
 services for rollover)
@@ -30,7 +30,7 @@ end
 
 ```ruby
 def do_the_thing
-  services[:service_name].execute(foo: "bar")
+  services[:service_name].call(foo: "bar")
 ```
 
 ### Testing the service is delegated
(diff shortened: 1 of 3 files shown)


app/services/README.md
```diff
@@ -5,8 +5,8 @@ a single place that instantiates all the services related to a model
 
 ## Conventions
 
-- Use the `execute` method as the interface on the service
-- Use named arguments to call `execute` to explicitly state what the arguments going in are
+- Use the `call` method as the interface on the service
+- Use named arguments to call `call` to explicitly state what the arguments going in are
```
gpeng, line 9: Is 'keyword arguments' the correct term here?

misaka: I think it depends on the service, I'm not sure we're dictating whether they are keyword arguments are regular arguments here.

docs/adr/0005-service-entry-points.md
```diff
@@ -0,0 +1,36 @@
...
+
+### 2. Use "call" as an entry point
+
+Some of our service have been using `call`, and in informal polling this has
+been popular among some devs. This also mimics existing callable Ruby objects,
+which may be a good thing.
+
+## Decision
+
+Use option 2, `call`. Additionally, provider both the instance method `#call` and a class
```
gpeng, line 30: type 'provider' -> 'provide'

gpeng (changes requested):

misaka:
> Should we standardise on using this [`master`/app/services/service_pattern.rb](https://github.com/DFE-Digital/teacher-training-api/blob/master/app/services/service_pattern.rb) (and Publish [DFE-Digital/publish-teacher-training@`master`/app/services/service_pattern.rb](https://github.com/DFE-Digital/publish-teacher-training/blob/master/app/services/service_pattern.rb))? It just clears up some of the 'boiler plate' and hides the `.new` loosely enforcing the pattern.

Ya, it's a good point. We could just have an ADR about using that module, but then we might not document our choice to use `call`. Maybe it's ok to have two ADRs, I can make a new one for using the service.

docs/adr/0005-service-entry-points.md
misaka: Fixed and squashed.

gpeng (approved):

PR #1522, opened 2020-08-13
[3790] api courses endpoint include

SamJamCul:
### Context
https://trello.com/c/iIyuBytF/3790-add-inclusion-to-public-api-courses-endpoints

* Add belongs_to relationships for course endpoint for accredited bodies, providers, and recruitment cycles

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/providers/courses_controller.rb
+++ b/app/controllers/api/public/v1/providers/courses_controller.rb
@@ -5,6 +5,7 @@ module API
         class CoursesController < API::Public::V1::ApplicationController
           def index
             render jsonapi: paginate(courses),
+              include: include_param,
               class: API::Public::V1::SerializerService.new.call
           end
 
@@ -38,6 +39,10 @@ module API
 
           def recruitment_cycle
             @recruitment_cycle ||= RecruitmentCycle.find_by(year: params[:recruitment_cycle_year])
+          end
+
+          def include_param
+            params.fetch(:include, "")
           end
         end
       end
--- a/spec/controllers/api/public/v1/providers/courses_controller_spec.rb
+++ b/spec/controllers/api/public/v1/providers/courses_controller_spec.rb
@@ -60,5 +60,42 @@ RSpec.describe API::Public::V1::Providers::CoursesController do
         expect(JSON.parse(response.body)["data"].size).to eql(2)
       end
     end
+
+    describe "include" do
+      let!(:course) { create(:course, :with_accrediting_provider, provider: provider) }
+
+      context "when includes specified" do
+        before do
+          get :index, params: {
+            recruitment_cycle_year: provider.recruitment_cycle.year,
+            provider_code: provider.provider_code,
+            include: "provider,accredited_body",
+          }
+        end
+
+        it "returns the provider and accrediting body connected to the course" do
+          expect(JSON.parse(response.body)["data"][0]["relationships"].keys).to include("provider")
+          expect(JSON.parse(response.body)["data"][0]["relationships"].keys).to include("accredited_body")
+          expect(JSON.parse(response.body)["included"][0]["id"]).to eql(course.accrediting_provider.id.to_s)
+          expect(JSON.parse(response.body)["included"][0]["type"]).to eql("providers")
+          expect(JSON.parse(response.body)["included"][1]["id"]).to eql(provider.id.to_s)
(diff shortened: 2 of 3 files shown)


SamJamCul:
[Add inclusion to public API courses endpoint(s)](https://trello.com/c/iIyuBytF/3790-add-inclusion-to-public-api-courses-endpoints)

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -60,5 +60,37 @@
...
     end
+
+    describe "include" do
+      let(:course) { create(:course, :with_accrediting_provider, provider: provider) }
+      before do
+        course
+      end
+
+      it "returns the provider and accrediting body connected to the course" do
+        get :index, params: {
```
darokel, line 71: Could we put this `get` request in the before block so it's in the arrange/act/assert structure? So the it block is just doing the assertion

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -60,5 +60,37 @@
...
+        expect(JSON.parse(response.body)["data"][0]["relationships"].keys).to include("provider")
+        expect(JSON.parse(response.body)["data"][0]["relationships"].keys).to include("accredited_body")
+        expect(JSON.parse(response.body)["included"][0]["id"]).to eql(course.accrediting_provider.id.to_s)
+        expect(JSON.parse(response.body)["included"][0]["type"]).to eql("providers")
+        expect(JSON.parse(response.body)["included"][1]["id"]).to eql(provider.id.to_s)
+        expect(JSON.parse(response.body)["included"][1]["type"]).to eql("providers")
+      end
+
+      it "doesn't include subjects as they aren't serialized" do
+        get :index, params: {
```
darokel, line 86: same for this

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -60,5 +60,37 @@
         expect(JSON.parse(response.body)["data"].size).to eql(2)
       end
     end
+
+    describe "include" do
+      let(:course) { create(:course, :with_accrediting_provider, provider: provider) }
+      before do
```
darokel, line 66: Can we add a space between the let and before block

app/controllers/api/public/v1/providers/courses_controller.rb
```diff
@@ -39,6 +40,10 @@ def provider
           def recruitment_cycle
             @recruitment_cycle ||= RecruitmentCycle.find_by(year: params[:recruitment_cycle_year])
           end
+
+          def include_param
```
defong, line 44: do we need to do something like 

https://github.com/DFE-Digital/teacher-training-api/blob/5748d7c3b8e9de80e701bf3f869899d8cd818833/app/controllers/api/public/v1/providers/locations_controller.rb#L33-L37

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -60,5 +60,37 @@
...
+      end
+
+      it "doesn't include subjects as they aren't serialized" do
+        get :index, params: {
+          recruitment_cycle_year: provider.recruitment_cycle.year,
+          provider_code: provider.provider_code,
+          include: "subjects",
+        }
+
+        expect(JSON.parse(response.body)["data"][0]["relationships"].keys).to_not include("subjects")
```
defong, line 92: @darokel maybe we should do some slice and dice so this PR can benefit

https://github.com/DFE-Digital/teacher-training-api/blob/5748d7c3b8e9de80e701bf3f869899d8cd818833/spec/support/request_helper.rb#L1-L5

darokel: yeah good shout

app/controllers/api/public/v1/providers/courses_controller.rb
asmega: I don't see why this is needed. The serialiser already specifies which associations can be serialised. Adding `PERMITTED_INCLUSIONS` would simply do the same work twice

PR #1528, opened 2020-08-14
Clean up duplicate JSON.parse calls into re-usable helper

darokel:
### Context

- I noticed a lot of duplicated calls to `JSON.parse` for parsing the response body in test files

- Standardise helper method for parsing JSON responses in test files to avoid duplicating calls. 
- Instead of `JSON.parse` you would just do `json_response` which would also memoize the results.

### Guidance to review

- See the new helper method

### DIFF
--- a/spec/support/request_helpers.rb
+++ b/spec/support/request_helpers.rb
@@ -1,0 +1,5 @@
+module RequestHelpers
+  def json_response
+    @json_response ||= JSON.parse(response.body)
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -24,4 +24,7 @@ Style/HashTransformKeys:
   Enabled: true
 Style/HashTransformValues:
   Enabled: true
+Naming/MemoizedInstanceVariableName:
+  EnforcedStyleForLeadingUnderscores: optional
+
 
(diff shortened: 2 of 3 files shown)


spec/support/request_helpers.rb
```diff
@@ -0,0 +1,5 @@
+module RequestHelpers
+  def json_response
+    @json_response ||= JSON.parse(response.body)
```
misaka, line 3: May be more prudent to use `@_json_response` for a memoised value here.

darokel: @misaka Currently this raises a rubocop violation https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/MemoizedInstanceVariableName. We would need to change `EnforcedStyleForLeadingUnderscores: disallowed (default)` to something like `required` or `optional`, thoughts?

misaka: As discussed, I'd be happy with `optional` here. I personally prefer the underscore here, but I understand that it's not always required.

darokel: done https://github.com/DFE-Digital/teacher-training-api/pull/1528/commits/991061d39e4caf7deaf073f8e1f29c929a2a089a

PR #1530, opened 2020-08-17
[3791] Add filtering to courses public/v1

asmega:
### Context

- https://trello.com/c/cis5KaLn/3791-add-filtering-to-public-api-courses-endpoint
- Add filtering to public/v1 provider courses#index endpoint

- Filtering uses existing `CourseSearchService`
- Rename in api `funding` to `funding_type` for compatibility to use `CourseSearchService` otherwise aliasing will be required to support both options
- For docs extract `Filter` schema to another file

### Guidance to review

- Depends on https://github.com/DFE-Digital/teacher-training-api/pull/1531 otherwise incorrect result set will be returned when `has_vacancies` filter is applied
- Run `curl -X GET  --header "Content-Type: application/json" "localhost:3001/api/public/v1/recruitment_cycles/2020/providers/2HA/courses?filter\[has_vacancies\]=true"` should return only courses with vacancies
- Run `curl -X GET  --header "Content-Type: application/json" "localhost:3001/api/public/v1/recruitment_cycles/2020/providers/2HA/courses"` should return only courses with and without vacancies

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


SamJamCul:
[[S] Add filtering to public API courses endpoint](https://trello.com/c/cis5KaLn/3791-s-add-filtering-to-public-api-courses-endpoint)

misaka (approved):

darokel (approved):

defong (approved):

PR #1531, opened 2020-08-17
[3811] Add findable to with_vacancies scope for courses

SamJamCul:
### Context
https://trello.com/c/ejcChJPq/3811-change-with-has-vacancies-to-only-include-findable-courses

* Add findable scoping to the with_vacancies scope for courses

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -496,7 +496,7 @@ class Course < ApplicationRecord
   end
 
   def ensure_site_statuses_match_study_mode
-    site_statuses.select(&:with_vacancies?).each do |site_status|
+    site_statuses.not_no_vacancies.each do |site_status|
       update_vac_status(study_mode, site_status)
     end
   end
--- a/app/models/site_status.rb
+++ b/app/models/site_status.rb
@@ -62,9 +62,10 @@ class SiteStatus < ApplicationRecord
     status_running? && published_on_ucas?
   end
 
-  scope :with_vacancies, -> { where.not(vac_status: :no_vacancies) }
+  scope :with_vacancies, -> { where.not(vac_status: :no_vacancies).findable }
+
   def with_vacancies?
-    !no_vacancies?
+    !no_vacancies? && findable?
   end
 
   def self.default_vac_status_given(study_mode:)
--- a/app/services/provider_reporting_service.rb
+++ b/app/services/provider_reporting_service.rb
@@ -3,7 +3,7 @@ class ProviderReportingService
     @providers = providers_scope.distinct
 
     @training_providers = @providers.where(id: Course.findable.select(:provider_id))
-    @open_providers = @providers.where(id: Course.findable.with_vacancies.select(:provider_id))
+    @open_providers = @providers.where(id: Course.with_vacancies.select(:provider_id))
 
     @closed_providers = @training_providers.where.not(id: @open_providers)
   end
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -693,7 +693,7 @@ describe Course, type: :model do
       context "course has vacancies" do
         let(:site_statuses) do
           [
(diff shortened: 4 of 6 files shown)


SamJamCul:
[Change with_/has_ vacancies to only include findable courses](https://trello.com/c/ejcChJPq/3811-change-with-has-vacancies-to-only-include-findable-courses)

SamJamCul:
> Looks good. Can we also change `SiteStatus#with_vacancies` to keep things consistent, though?

Done, also changed the SiteStatus `with_vacancies?` as well

app/models/course.rb
```diff
@@ -183,7 +183,7 @@ def new_draft_attributes
 
   scope :with_recruitment_cycle, ->(year) { joins(provider: :recruitment_cycle).where(recruitment_cycle: { year: year }) }
   scope :findable, -> { joins(:site_statuses).merge(SiteStatus.findable) }
-  scope :with_vacancies, -> { joins(:site_statuses).merge(SiteStatus.with_vacancies) }
+  scope :with_vacancies, -> { joins(:site_statuses).merge(SiteStatus.findable.with_vacancies) }
```
defong, line 186: was this meant to be changed?

given 

```ruby
# app/models/site_status.rb 
scope :with_vacancies, -> { where.not(vac_status: :no_vacancies).findable }
```

SamJamCul: You're right, it shouldn't need to be changed seeing as site status already covers it

misaka:
Just noticed https://github.com/DFE-Digital/teacher-training-api/blob/3811-change-with-has-vacancies-to-only-include-findable-course/app/models/course.rb#L498-L502

```ruby
  def ensure_site_statuses_match_study_mode
    site_statuses.select(&:with_vacancies?).each do |site_status|
      update_vac_status(study_mode, site_status)
    end
  end
```

We've changed how `with_vacancies?` behaves, have we thought through whether that will be an issue?

misaka:
Is it worth changing this bit too? It sounds like having `findable` applied twice will be a null-op, but it could also be confusing to the reader.

https://github.com/DFE-Digital/teacher-training-api/blob/master/app/services/publish_reporting_service.rb#L109-L110

```ruby
    findable_courses = courses_changed_recently.findable.distinct
    open_courses = findable_courses.with_vacancies
```

SamJamCul:
> Is it worth changing this bit too? It sounds like having `findable` applied twice will be a null-op, but it could also be confusing to the reader.
> 
> https://github.com/DFE-Digital/teacher-training-api/blob/master/app/services/publish_reporting_service.rb#L109-L110
> 
> ```ruby
>     findable_courses = courses_changed_recently.findable.distinct
>     open_courses = findable_courses.with_vacancies
> ```

For this one we did spot it, but decided against changing it, as the `findable_courses` bit is used elsewhere, as well as having the `distinct` on the end. It felt like it would be more hassle adding `findable` to the other uses rather than just having this redundant one here.

SamJamCul:
> Just noticed https://github.com/DFE-Digital/teacher-training-api/blob/master/app/models/course.rb#L498-L502
> 
> ```ruby
>   def ensure_site_statuses_match_study_mode
>     site_statuses.select(&:with_vacancies?).each do |site_status|
>       update_vac_status(study_mode, site_status)
>     end
>   end
> ```
> 
> We've changed how `with_vacancies?` behaves, have we thought through whether that will be an issue?

Fixed this, we're now checking it slightly differently, by looking directly at each site statuses vac_status. This bypasses the `with_vacancies?` findable problem, keeping the original functionality intact!

misaka (approved):
Good work guys, this was tricky, but is one step closer to Perfect Datas!!! 💯

PR #1532, opened 2020-08-17
modified azure/template.json accommodate multiple custom-domain (api\api2)

felixtheflex:
### Context

The change is to accommodate both api and ap2 custom domain requires for all environment.

Ammending arm template to support multiple custom domains

### Checklist

- [x ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -35,6 +35,12 @@
         "description": "The custom hostname to add to the app service."
       }
     },
+    "customDomains": {
+      "type": "array",
+      "metadata": {
+        "description": "Custom domains specified in an array format:-[\"api.staging.publish-teacher-training-courses.service.gov.uk\", \"api2.staging.publish-teacher-training-courses.service.gov.uk\"]"
+      }
+    },
     "certificateName": {
       "type": "string",
       "defaultValue": "",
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -35,6 +35,12 @@
         "description": "The custom hostname to add to the app service."
       }
     },
+    "customDomains": {
+      "type": "array",
+      "metadata": {
+        "description": "The custom hostname to add to the app service."
```
saliceti, line 41: Should it describe custom domains instead?

azure/template.json
```diff
@@ -501,8 +554,8 @@
               }
             ]
           },
-          "customHostName": {
-            "value": "[parameters('customHostName')]"
+          "customDomains": {
+            "value": "[reference('custom-domains').outputs.customDomains.value]"
           },
           "certificateThumbprint": {
             "value": "[if(greater(length(parameters('customHostname')), 0), reference('app-service-certificate', '2018-11-01').outputs.certificateThumbprint.value, '')]"
```
vigneshmsft, line 561: this can be removed as well

azure/template.json
misaka: Actually, `api` and `api2` are the hostname segment, no? Should this be `fqdn`?

felixtheflex: > 
> 
> Should it describe custom domains instead?

Descriptions now corrected......

PR #1534, opened 2020-08-18
[3789] - Create endpoint for courses based on recruitment cycle

darokel:
### Context

- https://trello.com/c/VD20nCHv/3789-implement-api-endpoint-api-public-v1-recruitmentcycles-year-courses

- Add `recruitment_cycle/:recruitment_cycle_year/courses` endpoint
- Update v1/public documentation for this endpoint

### Guidance to review

- Fire off a request for a given recruitment_cycle year to get a list of that cycle's courses. Here is an example request you can use (providing you have the sanitised data available):

`curl --location --request GET 'localhost:3001/api/public/v1/recruitment_cycles/2020/courses?include=provider,recruitment_cycle&page[page]=1&page[per_page]=10' \
--header 'Content-Type: application/json'`

- Assert the list of courses are returned correctly for that year. You can compare with the results from a rails console
- Assert you can pass paginate the results

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/courses_controller.rb
+++ b/app/controllers/api/public/v1/courses_controller.rb
@@ -3,23 +3,21 @@ module API
     module V1
       class CoursesController < API::Public::V1::ApplicationController
         def index
-          render json: {
-            data: [
-              {
-                id: "124",
-                type: "Course",
-                attributes: {
-                  code: "3GTY",
-                  provider_code: "066",
-                  age_minimum: 10,
-                  age_maximum: 14,
-                },
-              },
-            ],
-            jsonapi: {
-              version: "1.0",
-            },
-          }
+          render jsonapi: paginate(courses), include: include_param, class: API::Public::V1::SerializerService.call
+        end
+
+      private
+
+        def courses
+          @courses ||= recruitment_cycle&.courses
+        end
+
+        def recruitment_cycle
+          @recruitment_cycle ||= RecruitmentCycle.find_by(year: params[:recruitment_cycle_year])
+        end
+
+        def include_param
+          params.fetch(:include, "")
         end
       end
     end
--- a/app/controllers/api/public/v1/courses_controller.rb
+++ b/app/controllers/api/public/v1/courses_controller.rb
@@ -9,11 +9,11 @@ module API
       private
 
         def courses
-          @courses ||= recruitment_cycle&.courses
+          @courses ||= recruitment_cycle.courses
         end
 
         def recruitment_cycle
-          @recruitment_cycle ||= RecruitmentCycle.find_by(year: params[:recruitment_cycle_year])
+          @recruitment_cycle ||= RecruitmentCycle.find_by!(year: params[:recruitment_cycle_year])
         end
 
         def include_param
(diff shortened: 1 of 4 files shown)


darokel:
[Implement API endpoint - /api/public/v1/recruitment_cycles/:year/courses](https://trello.com/c/VD20nCHv/3789-implement-api-endpoint-api-public-v1-recruitmentcycles-year-courses)

spec/docs/courses_spec.rb
```diff
@@ -39,9 +39,19 @@
...
+                required: false,
+                description: "The associated data for this resource.",
+                schema: {
+                  enum: %w[accredited_body provider recruitment_cycle],
+                },
+                example: "recruitment_cycle,provider"
 
       response "200", "The collection of courses." do
         let(:year) { "2020" }
+        let(:include) { "provider" }
```
asmega, line 55: can you create one course for this please. at the moment it passes because there are no courses

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -3,23 +3,21 @@ module Public
...
+        end
+
+      private
+
+        def courses
+          @courses ||= recruitment_cycle&.courses
+        end
+
+        def recruitment_cycle
+          @recruitment_cycle ||= RecruitmentCycle.find_by(year: params[:recruitment_cycle_year])
```
asmega, line 16:
```suggestion
          @recruitment_cycle ||= RecruitmentCycle.find_by!(year: params[:recruitment_cycle_year])
```
This way we throw an error which will return a 404. it also means you can remove the guard on `courses`

spec/docs/courses_spec.rb
darokel: good spot and done https://github.com/DFE-Digital/teacher-training-api/pull/1534/files#diff-7add9c00508140e920eeff037644b94cR57

asmega (approved):

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -3,23 +3,21 @@ module Public
...
+        end
+
+      private
+
+        def courses
+          @courses ||= recruitment_cycle.courses
+        end
+
+        def recruitment_cycle
+          @recruitment_cycle ||= RecruitmentCycle.find_by!(year: params[:recruitment_cycle_year])
```
misaka, line 16: I just realised something ... we have two routes that get all the courses for a recruitment cycle in the v3 API:

```
/api/v3/recruitment_cycles/:recruitment_year/courses
/api/v3/courses
```

The later one defaults to the current recruitment year and is what Find uses to retrieve courses, this allows the API to decide which courses to show, when necessary. We're going to want to support that here, could a separate PR but just want to bring that up here.

darokel: Yeah could we do that in a separate ticket/PR, i'd like to avoid adding stuff to tickets once they've started 👍

misaka: Works for me.

misaka: https://trello.com/c/bqY8jqRL/3826-add-api-endpoint-api-public-v1-courses

misaka (approved):

PR #1537, opened 2020-08-19
[3818] Placeholder for confirmed allocations

asmega:
### Context

- https://trello.com/c/sAOGBGQF/3818-backend-allocation-for-final-numbers
- Updates allocations endpoints to include confirmed allocation numbers

- Exposes confirmed_number_of_places as part of the allocation endpoints
- This commit must be rolled out before the publish changes

### Guidance to review

-  Test in unison with https://github.com/DFE-Digital/publish-teacher-training/pull/1318
- Find a provider with allocations eg `B20`
- Modify one of their allocations in the rails console and set `confirmed_number_of_places` to `123`
- View the allocation via publisher web interface. Should see `123`
- The other unset allocation should be blank i.e. the app does not throw an error

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


asmega:
[Backend allocation for final numbers](https://trello.com/c/sAOGBGQF/3818-backend-allocation-for-final-numbers)

db/migrate/20200819084304_add_confirmed_number_of_places_to_allocations.rb
```diff
@@ -0,0 +1,5 @@
+class AddConfirmedNumberOfPlacesToAllocations < ActiveRecord::Migration[6.0]
+  def change
+    add_column :allocation, :confirmed_number_of_places, :integer, null: true
```
darokel, line 3: Don't we want to default this to 0? Otherwise we'd have to check for `nil` all the time if using this attribute in calculations.

asmega: I think we want nil/null here. So we know the the data hasn't been populated. This is a placeholder and will be populated with data once it's been handed to us

darokel: FYI this was discussed offline

darokel (approved):

misaka (approved):

defong (approved):

PR #1539, opened 2020-08-20
[3805] - Implement sorting and filtering for public courses endpoint

darokel:
### Context

- https://trello.com/c/u8s9har0/3805-s-linked-to-3789-add-sorting-to-api-public-v1-recruitmentcycles-year-courses
https://trello.com/c/t0UwzFDt/3806-s-linked-to-3789-add-filtering-to-api-public-v1-recruitmentcycles-year-courses
- Dependent on this PR https://github.com/DFE-Digital/teacher-training-api/pull/1534

- Implements sorting/filtering via the CourseSearchService in `/api/public/v1/:recruitment_cycle_year/courses`

### Guidance to review

-  Fire off the example request below and add your sorting and filtering options in the request.
- Assert the courses returned match provided sorting/filtering params

CURL:

```
curl --location --request GET 'localhost:3001/api/public/v1/recruitment_cycles/2020/courses?sort=name,provider.provider_name&filter[funding_type]=salary' \
```

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/controllers/api/public/v1/courses_controller_spec.rb
+++ b/spec/controllers/api/public/v1/courses_controller_spec.rb
@@ -20,14 +20,18 @@ RSpec.describe API::Public::V1::CoursesController do
     context "when there are courses" do
       before do
         provider.courses << build_list(:course, 2, provider: provider)
-
-        get :index, params: {
-          recruitment_cycle_year: "2020",
-        }
       end
 
-      it "returns correct number of courses" do
-        expect(json_response["data"].size).to eql(2)
+      context "default response" do
+        before do
+          get :index, params: {
+            recruitment_cycle_year: "2020",
+          }
+        end
+
+        it "returns correct number of courses" do
+          expect(json_response["data"].size).to eql(2)
+        end
       end
 
       context "with pagination" do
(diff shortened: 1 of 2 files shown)


darokel:
[[S] [Linked to 3789] - Add filtering to /api/public/v1/recruitment_cycles/:year/courses](https://trello.com/c/t0UwzFDt/3806-s-linked-to-3789-add-filtering-to-api-public-v1-recruitmentcycles-year-courses)

darokel:
[[S] [Linked to 3789] Add sorting to /api/public/v1/recruitment_cycles/:year/courses](https://trello.com/c/u8s9har0/3805-s-linked-to-3789-add-sorting-to-api-public-v1-recruitmentcycles-year-courses)

asmega (approved):

spec/controllers/api/public/v1/courses_controller_spec.rb
```diff
@@ -70,6 +74,46 @@
           expect(provider_id).to eq(provider.id)
         end
       end
+
+      context "with sorting" do
```
misaka, line 78: Just noticed that `with includes` above tests how the JSON API renderer works, it's response, etc. This section only tests it calls `CourseSearchService` the right way. Minor point, could be worth aligning.

misaka (approved):

spec/controllers/api/public/v1/courses_controller_spec.rb
darokel: Yeah this was intentional. The `CourseSearchService` is already tested quite extensively. I didn't see the value in duplicating those here in the controller. I am relying on that search class to do its job but asserting it gets the right params to perform that task.

I suppose we could lead with that same assumption for the include scenario but I want to sanity check the include params are being whitelisted properly by the JSON API library we use and that it returns the correct data back.

PR #1541, opened 2020-08-20
[3809] - Remove `recruitment_cycle_year` from the public api's providers endpoint

darokel:
### Context

Including this data forces the endpoint to include `recruitment_cycles` into the SQL query. Instead of doing this we should allow the client to include the `recruitment_cycle` if they need to get the year.

- https://trello.com/c/uoPwkZ9m/3809-remove-recruitmentcycleyear-from-the-public-apis-providers-endpoint

- Remove from API v1 providers endpoint documentation.
- Remove from API v1 providers serialiser.
- Remove from API v1 providers endpoint tests.
- Refactor existing API v1 providers tests to conform with agreed approach. I've moved Sau's hard work into `spec/controllers/api/public/v1/providers_controller_spec.rb`
- Added pagination and include support to existing public providers endpoint as it looks like these weren't already there

### Guidance to review

- Fire up the public api docs locally and assert the the recruitment_cycle_year has been removed form the Provider's OpenAPI schema
- Use the curl request below to assert the `recruitment_cycle_year` is not included in the json response, unless explicitly set in the `include` query string

CURL:

```
curl --location --request GET 'localhost:3001/api/public/v1/recruitment_cycles/2020/providers?include=recruitment_cycle&page[page]=1&page[per_page]=10' \
``

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


darokel:
[Remove `recruitment_cycle_year` from the public api's providers endpoint](https://trello.com/c/uoPwkZ9m/3809-remove-recruitmentcycleyear-from-the-public-apis-providers-endpoint)

PR #1545, opened 2020-08-21
[3820] remove PE from edit subjects list for non admins

avinhurry:
### Context

Removes the PE subject from subject list for non admins as providers require permission to recruit for this course. It is not visible as a subject on the add new course flow.

Remove the PE subject from subject list for non admins.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/courses_controller.rb
+++ b/app/controllers/api/v2/courses_controller.rb
@@ -65,8 +65,19 @@ module API
       end
 
       def show
-        # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
-        render jsonapi: @course, include: params[:include], class: CourseSerializersService.new.execute
+        json_data = JSONAPI::Serializable::Renderer.new.render(
+          @course,
+          class: CourseSerializersService.new.execute,
+          include: params[:include],
+        )
+
+        unless @current_user.admin?
+          json_data[:data][:meta][:edit_options][:subjects]&.reject! do |subject|
+            subject[:attributes][:subject_name] == "Physical education"
+          end
+        end
+
+        render json: json_data
       end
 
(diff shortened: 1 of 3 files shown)


asmega (approved):
I think this is fine for now since it works. I think we should move this kind of logic to at least to the serialiser in the future. A controller shouldn't have to detect a user type then reach deep down into a data structure to remove something very specific. Fundamentally it feels a bit wrong

app/controllers/api/v2/courses_controller.rb
```diff
@@ -65,8 +65,19 @@ def index
...
       def show
-        # https://github.com/jsonapi-rb/jsonapi-rails/issues/113
-        render jsonapi: @course, include: params[:include], class: CourseSerializersService.new.execute
+        json_data = JSONAPI::Serializable::Renderer.new.render(
+          @course,
+          class: CourseSerializersService.new.execute,
+          include: params[:include],
+        )
+
+        unless @current_user.admin?
```
darokel, line 74: this feels like it should be handled in the serializer. You can expose the user information to the serializer via `expose:` and deal with the logic in there?

darokel: ie https://github.com/DFE-Digital/teacher-training-api/blob/master/app/controllers/api/public/v1/providers/courses/locations_controller.rb#L10

dankmitchell (approved):

app/controllers/api/v2/courses_controller.rb
darokel: discussed doing this alternatively offline, but potentially in a future PR

PR #1548, opened 2020-08-24
[54] - Implement api/public/v1/courses

darokel:
### Context

- https://trello.com/c/bqY8jqRL/54-add-api-endpoint-api-public-v1-courses

- This is the actual courses endpoint that is used by Find and Apply, it defaults to the current recruitment cycle when retrieving courses, and is nearly identical to `/api/public/v1/recruitment_cycles/:year/courses`

### Guidance to review

- Implement `api/public/v1/courses`
- Use curl command below to assert courses for current cycle are returned when no cycle is provided in the request

Curl:

```
curl --location --request GET 'localhost:3001/api/public/v1/courses?include=provider,recruitment_cycle&page[page]=1&page[per_page]=3' \
```
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/courses_controller.rb
+++ b/app/controllers/api/public/v1/courses_controller.rb
@@ -19,7 +19,9 @@ module API
         end
 
         def recruitment_cycle
-          @recruitment_cycle ||= RecruitmentCycle.find_by!(year: params[:recruitment_cycle_year])
+          @recruitment_cycle = RecruitmentCycle.find_by(
+            year: params[:recruitment_cycle_year],
+          ) || RecruitmentCycle.current_recruitment_cycle
         end
 
         def include_param
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -114,9 +114,11 @@ Rails.application.routes.draw do
           end
         end
 
-        get "provider_suggestions", to: "provider_suggestions#index"
         resources :subjects, only: %i[index]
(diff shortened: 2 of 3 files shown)


darokel:
[Add API endpoint - /api/public/v1/courses](https://trello.com/c/bqY8jqRL/54-add-api-endpoint-api-public-v1-courses)

defong (approved):

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -19,7 +19,9 @@ def courses
         end
 
         def recruitment_cycle
-          @recruitment_cycle ||= RecruitmentCycle.find_by!(year: params[:recruitment_cycle_year])
+          @recruitment_cycle = RecruitmentCycle.find_by(
```
asmega, line 22:
```suggestion
          @recruitment_cycle ||= RecruitmentCycle.find_by(
```
can we put the memoization back in? just it incase it gets called again it doesn't have to make another API request

asmega (approved):

app/controllers/api/public/v1/courses_controller.rb
darokel: yup! copy and paste fail 🤦‍♂️

PR #1557, opened 2020-09-01
[32] Changes application start and end dates in db

JonF1:
# Context
* https://trello.com/c/d79wSVBk/32-check-that-the-when-will-applications-open-error-message-displays-the-correct-dates

The validation for ensuring users enter a valid date when adding a new course was wrong for the 2020 to 2021 recruitment cycle. It was using the same day/months as the 2019 to 2020 cycle when they needed to be changed. 

The last entry in the application start and end date columns in the recruitment cycle table was changed using a migration to  the correct day/months as per ticket.

### Guidance to review
Add a new course to next year's recruitment cycle and trigger the validation 

### Checklist

- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history


JonF1:
[[S] check that the 'when will applications open' error message displays the correct dates.](https://trello.com/c/d79wSVBk/32-check-that-the-when-will-applications-open-error-message-displays-the-correct-dates)

db/migrate/20200828125910_change_application_start_and_end_date_for_recruitment_cycles.rb
```diff
@@ -0,0 +1,19 @@
+class ChangeApplicationStartAndEndDateForRecruitmentCycles < ActiveRecord::Migration[6.0]
+  def up
+    recruitment_cycle = RecruitmentCycle.find_by(year: "2021")
+    return if recruitment_cycle.nil?
+
+    recruitment_cycle.application_start_date = "13/10/2020"
```
asmega, line 6: I'd be slightly uncomfortable running this or any date in this format. It does work locally but your system probably knows you're in the London timezone and may assume British dates. The server will most likely be in UTC and might not parse this string as you intended.

I'd probably opt for a date object so there is no uncertainty:
```suggestion
    recruitment_cycle.application_start_date = Date.new(2020, 10, 13)
```

JonF1: Thanks Phil, I've updated

asmega (approved):

defong (approved):

PR #1564, opened 2020-09-08
use single container instance for background jobs

vigneshmsft:

configure single container instance to listen to multiple queues.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -117,10 +117,10 @@
         "description": "Connection string for Sentry monitoring for backend app."
       }
     },
-    "sentryDSNContainers": {
-      "type": "array",
-      "metadata": {
-        "description": "Connection string for Sentry monitoring for background geocoding and other container instances."
+    "sentryDSNBGJobs":{
+      "type" : "string",
+      "metadata":{
+        "description" : "Sentry DSN for monitoring the background bgJobs container."
       }
     },
     "settingsAuthenticationSecret": {
@@ -164,19 +164,19 @@
       }
     },
     "containerInstanceName": {
-      "type": "array",
+      "type": "string",
       "metadata": {
-        "description": "The name of the Azure Container Instances resource. This will be its identifier in Azure and can be different from the image name."
+        "description": "The name of the Azure Container Instance resource. This will be its identifier in Azure and can be different from the image name."
       }
     },
     "containerName": {
-      "type": "array",
+      "type": "string",
       "metadata": {
         "description": "The name or role of the container to identify individual containers in multi-container setup."
       }
     },
     "bgNameTag": {
-      "type": "array",
+      "type": "string",
       "metadata": {
         "description": "List of app names used to tag the container logs."
       }
@@ -228,12 +228,6 @@
       "type": "string",
       "metadata": {
         "description": "The email address to be used for psql security alerts."
-      }
-    },
-    "backgroundContainerImageReference": {
-      "type": "array",
-      "metadata": {
-        "description": "The container image to pull from the server. Should be in image:tag format."
       }
     },
     "vnetProfileName": {
@@ -327,6 +321,7 @@
     "appServiceName": "[concat(variables('resourceNamePrefix'), '-as')]",
     "appServicePlanName": "[concat(variables('resourceNamePrefix'), '-asp')]",
     "appServiceRuntimeStack": "[concat('DOCKER|', parameters('dockerHubUsername'), '/', parameters('containerImageReference'))]",
+    "bgJobsContainerImageReference": "[concat(parameters('dockerHubUsername'), '/', parameters('containerImageReference'))]",
     "databaseResourceNamePrefix": "[toLower(concat(parameters('resourceEnvironmentName'),'-', parameters('databaseServiceName')))]",
     "databaseServerName": "[concat(variables('databaseResourceNamePrefix'), '-psql')]",
     "databaseUsername": "[concat(parameters('databaseUsername'), '@', variables('databaseServerName'))]",
@@ -454,7 +449,7 @@
               },
               "certificateThumprint": {
                 "value": "[reference('app-service-certificate', '2018-11-01').outputs.certificateThumbprint.value]"
-               }   
+               }
           }
       }
     },
@@ -895,13 +890,9 @@
       ]
     },
     {
-      "name": "[concat('container-instances', copyIndex())]",
+      "name": "bg-jobs-container-instances",
       "type": "Microsoft.Resources/deployments",
       "apiVersion": "2017-05-10",
-      "copy":{
-        "name": "container-instances-loop",
-        "count": "[length(parameters('containerName'))]"
-      },
       "properties": {
         "mode": "Incremental",
         "templateLink": {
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -41,16 +41,16 @@ services:
       - AGENT_JOBSTATUS=${AGENT_JOBSTATUS}
       - GIT_BRANCH=${GIT_BRANCH}
 
-  bg-geocode:
+  bg:
     build:
```
saliceti, line 45: Can we remove the build here? This uses the same image as `web`

docker-compose.yml
```diff
@@ -41,16 +41,16 @@ services:
...
         - DOCKER_REPOSITORY=${dockerHubUsername:-dfedigital}
         - TEACHER_TRAINING_API=${dockerHubImageName:-teacher-training-api}
         - CODE_VERSION=${GIT_BRANCH:-latest}
-      context: bg-geocode
+      context: .
       cache_from:
-        - ${dockerHubUsername:-dfedigital}/teacher-training-bg-geocode:${GIT_BRANCH:-latest}
-    image: ${dockerHubUsername:-dfedigital}/teacher-training-bg-geocode:${GIT_BRANCH:-latest}
+        - ${dockerHubUsername:-dfedigital}/teacher-training-api:${GIT_BRANCH:-latest}
+    image: ${dockerHubUsername:-dfedigital}/teacher-training-api:${GIT_BRANCH:-latest}
```
saliceti, line 53: This now uses the generic image, so the command should be overriden here, shoudn't it?

docker-compose.yml
felixtheflex: done....

docker-compose.yml
felixtheflex: done...

saliceti (approved):
OK

azure/template.json
```diff
@@ -117,10 +117,10 @@
...
     },
-    "sentryDSNContainers": {
-      "type": "array",
-      "metadata": {
-        "description": "Connection string for Sentry monitoring for background geocoding and other container instances."
+    "sentryDSNSideKiq":{
+      "type" : "string",
+      "metadata":{
+        "description" : "Sentry DSN for monitoring the background sidekiq container."
       }
```
misaka, line 124: Should this be `sentryDSNBGJobs` maybe? i.e. `bg jobs` instead of `sidekiq`?

azure/template.json
```diff
@@ -327,6 +321,7 @@
     "appServiceName": "[concat(variables('resourceNamePrefix'), '-as')]",
     "appServicePlanName": "[concat(variables('resourceNamePrefix'), '-asp')]",
     "appServiceRuntimeStack": "[concat('DOCKER|', parameters('dockerHubUsername'), '/', parameters('containerImageReference'))]",
+    "backgroundContainerImageReference": "[concat(parameters('dockerHubUsername'), '/', parameters('containerImageReference'))]",
```
misaka, line 324: Whether we chose `sidekiq` or `bg jobs`, perhaps we should align the name here with parameter name on line 120.

azure/template.json
misaka: Ya, I've just noticed that elsewhere in this file, we refer to a "background" container. Feels like that name should be aligned with the above parameter.

docker-compose.yml
```diff
@@ -33,73 +37,28 @@ services:
...
-        - TEACHER_TRAINING_API=${dockerHubImageName:-teacher-training-api}
-        - CODE_VERSION=${GIT_BRANCH:-latest}
-      context: bg-jobs
-      cache_from:
-        - ${dockerHubUsername:-dfedigital}/teacher-training-bg-jobs:${GIT_BRANCH:-latest}
-    image: ${dockerHubUsername:-dfedigital}/teacher-training-bg-jobs:${GIT_BRANCH:-latest}
-    volumes:
+  bg:
+    image: ${dockerHubUsername:-dfedigital}/teacher-training-api:${GIT_BRANCH:-latest}
+    command: bundle exec sidekiq -c 5 -C config/sidekiq.yml
```
misaka, line 52: Just to note that the Dockerfiles for the bg containers were running `rails db:migrate` before `sidekiq`. I think it's ok here in docker-compose not to do that, but we should consider doing the `rails db:migrate` when we run these containers in the deployed environments.

misaka (changes requested):
Just think we need to resolve the naming of the sentry DSN param for the background container, it should probably be changed to `background` or `bg jobs`.

docker-compose.yml
saliceti: In docker compose there is a dependency between web and bgjobs so migrate will always have run when bg starts.

In deployed environments, it may be dangerous to add migrate to bgjobs as it may end up running in parallel webapp//bgjobs. We could simply add a similar dependency in terraform so the webapp is always updated first and runs migrate. So bgjobs wouldn't have to run it.

azure/template.json
felixtheflex: done....

azure/template.json
felixtheflex: done..

azure/template.json
```diff
@@ -895,13 +890,9 @@
       ]
     },
     {
-      "name": "[concat('container-instances', copyIndex())]",
+      "name": "background-container-instances",
```
saliceti, line 893: "background-container-instances" -> "bg-jobs-container-instances" ?

azure/template.json
```diff
@@ -1018,7 +1012,7 @@
       "type": "Microsoft.Resources/deployments",
       "apiVersion": "2017-05-10",
       "dependsOn": [
-        "container-instances-loop"
+        "background-container-instances"
```
saliceti, line 1015: "background-container-instances" -> "bg-jobs-container-instances" ?

azure/template.json
felixtheflex: done...

azure/template.json
felixtheflex: done..

docker-compose.yml
misaka: [Rails migrations are safe to run concurrently](https://github.com/rails/rails/pull/22122). The problem you describe already happens in environments that have multiple web servers that can be spun-up at the same time. Putting `db:migrate` into the startup of bg-jobs just means that that process should always be safe to run (it won't error out saying that the db is the wrong version).

That said, using a dependency in terraform is even better, happy to go with that. :)

saliceti (approved):
OK

misaka (approved):

docker-compose.yml
```diff
@@ -33,73 +37,28 @@ services:
...
-      args:
-        - DOCKER_REPOSITORY=${dockerHubUsername:-dfedigital}
-        - TEACHER_TRAINING_API=${dockerHubImageName:-teacher-training-api}
-        - CODE_VERSION=${GIT_BRANCH:-latest}
-      context: bg-jobs
-      cache_from:
-        - ${dockerHubUsername:-dfedigital}/teacher-training-bg-jobs:${GIT_BRANCH:-latest}
-    image: ${dockerHubUsername:-dfedigital}/teacher-training-bg-jobs:${GIT_BRANCH:-latest}
-    volumes:
+  bg:
```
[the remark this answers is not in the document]
felixtheflex: done...

PR #1566, opened 2020-09-10
[0101] - Bulk updated applications_open_from to '13/10/20'

avinhurry:
### Context

Providers are struggling to publish courses due to the application_open_from date being in the past.  When they update the course it throws a validation error (which is not surfaced on the front end).

Updated applications_open_from date for the affected 17390 courses.

### Guidance to review

Restore backup database from production and run both API and Publish.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


avinhurry:
[Wrong date stated for Application open date for courses rolled over on Publish](https://trello.com/c/0ewZlfww/101-wrong-date-stated-for-application-open-date-for-courses-rolled-over-on-publish)

db/migrate/20200910143249_fix_course_application_open_from_dates.rb
```diff
@@ -0,0 +1,14 @@
+class FixCourseApplicationOpenFromDates < ActiveRecord::Migration[6.0]
+  def up
+    recruitment_cycle = RecruitmentCycle.find_by(year: "2021")
+    return if recruitment_cycle.nil?
+
+    courses = recruitment_cycle.courses.where.not(:applications_open_from => recruitment_cycle.application_start_date..recruitment_cycle.application_end_date)
```
darokel, line 6:
```suggestion
    courses = recruitment_cycle.courses.where.not(applications_open_from: recruitment_cycle.application_start_date..recruitment_cycle.application_end_date)
```

db/migrate/20200910143249_fix_course_application_open_from_dates.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    courses.update_all applications_open_from: recruitment_cycle.application_start_date
```
darokel, line 8:
```suggestion
courses.update_all(applications_open_from: recruitment_cycle.application_start_date)
```

db/migrate/20200910143249_fix_course_application_open_from_dates.rb
```diff
@@ -0,0 +1,14 @@
+class FixCourseApplicationOpenFromDates < ActiveRecord::Migration[6.0]
+  def up
+    recruitment_cycle = RecruitmentCycle.find_by(year: "2021")
+    return if recruitment_cycle.nil?
+
+    courses = recruitment_cycle.courses.where.not(applications_open_from: recruitment_cycle.application_start_date..recruitment_cycle.application_end_date)
```
gpeng, line 6: Can we split this line up a bit...

```suggestion
   valid_start_date_range = recruitment_cycle.application_start_date..recruitment_cycle.application_end_date
   courses = recruitment_cycle.courses.where.not(applications_open_from: valid_start_date_range)
```

db/migrate/20200910143249_fix_course_application_open_from_dates.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    courses.update_all(applications_open_from: recruitment_cycle.application_start_date)
+  end
+
+  def down
+    # cannot go back as this will be invalid
```
gpeng, line 12: We can make this reversible if we want to be more cautious. We could add a new column e.g. `old_applications_from_value` and copy the existing values to it before the update. Then drop the column in a subsequent migration in a future PR once we're happy this one didn't cause issues. Whether we do that depends on how happy we are with the risk. Might be overkill here 🤷

gpeng (commented):
I'm still a bit against using ActiveRecord in migrations as the domain objects change and as migrations don't break tests so they get forgotten and don't work in the future. I'm also not too keen on updating the data here but if we do maybe we should use SQL which is a more appropriate language in this layer (the DB layer essentially). 

Maybe using something like [after_party](https://github.com/theSteveMitchell/after_party) might be something to look at?

db/migrate/20200910143249_fix_course_application_open_from_dates.rb
gpeng: Discussed offline. We're going to proceed as is with this and review carefully in staging.

darokel (approved):
Assuming that this has been discussed offline with @gpeng  👍

gpeng (approved):

PR #1567, opened 2020-09-11
[89] Handle Pagy::OverflowError

darokel:
### Context

- https://trello.com/c/ZtNKrEqA/89-m-ignore-suppress-pagy-errors

- Handle `Pagy::OverflowError`
- Introduce a generic helper method to deal with error responses in json

### Guidance to review

Fire the following curl command locally (assuming you have enough courses in your db):

```
curl --location --request GET 'localhost:3001/api/v3/recruitment_cycles/2020/courses?filter[has_vacancies]=true&filter[provider.provider_name]=The%20University%20of%20Warwick&include=site_statuses.site,provider,subjects&page[page]=5&page[per_page]=10&sort=provider.provider_name,name' \
```

Assert you receive the following response:

```json
{
    "errors": [
        {
            "status": 400,
            "title": "BAD REQUEST",
            "detail": "The requested page param was out of range and invalid for this request."
        }
    ]
}

```

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


darokel:
[[M] ignore/suppress pagy errors](https://trello.com/c/ZtNKrEqA/89-m-ignore-suppress-pagy-errors)

gpeng (changes requested):
The concern and rescue_from stuff looks great. We need to return the error object with `errors` as a top level field and the errors in an array as per [the JSON:API style](https://jsonapi.org/format/#error-objects).

app/controllers/concerns/response_error_handler.rb
```diff
@@ -0,0 +1,9 @@
+module ResponseErrorHandler
+  extend ActiveSupport::Concern
+
```
defong, line 3: just wondering 

should this snippet sit here? 
```ruby
        rescue_from Pagy::OverflowError do |_exception|
          render_json_error(code: 400, message: I18n.t("exceptions.pagy.overflow"), status: :bad_request)
        end
```

spec/requests/api/v3/courses/index_spec.rb
```diff
@@ -191,9 +191,9 @@
...
           course_ids << json_response.dig("data", 0, "id")
 
-          expect {
-            get page_3
-          }.to raise_error(Pagy::OverflowError)
+          get page_3
+
+          expect(response).to have_http_status(:bad_request)
 
           expect(course_ids.compact).to eql([course_b.id.to_s, course_a.id.to_s])
```
defong, line 198: are we still expected to return course ids after bad request?

darokel: This was a failing spec that seemed to assert the `Pagy::Overflow` error would be raised in this scenario. I'm not entirely sure of the context but it looks like it's dealing with course_ids for the previous successful requests (not the third). Perhaps this assertion for `Pagy::Oveflow` is redundant here as I've already added a spec for it in the new public api for courses.

app/controllers/concerns/response_error_handler.rb
darokel: I wondered that too. I'm not too fussed but I did think that putting it into the module might be a better place for it (and others).

darokel: Actually, I think we should handle these rescues in the relevant controllers to avoid polluting this module with exceptions which might not make sense for other controllers.

spec/requests/api/v3/courses/index_spec.rb
darokel: Update: I've reverted this scenario back to it's original assertion based on this pr https://github.com/DFE-Digital/teacher-training-api/commit/475bf2781629d3b04247a8ae32fed36de50f7488#diff-230985fb50f1d3b01a7b55d41484ee88R201

app/controllers/api/v2/application_controller.rb
```diff
@@ -2,11 +2,16 @@ module API
...
     class ApplicationController < ::ApplicationController
       include Pagy::Backend
+      include ResponseErrorHandler
 
       attr_reader :current_user
 
       rescue_from ActiveRecord::RecordNotFound, with: :jsonapi_404
 
+      rescue_from Pagy::OverflowError do |_exception|
+        render_json_error(status: 400, message: I18n.t("pagy.overflow"))
```
defong, line 12: lacks coverage for this bit

defong (approved):
More or less all there, just a minor tidy up

app/controllers/concerns/response_error_handler.rb
gpeng: We should move that block Sau quoted into a separate, pagy overflow handling concern and include it.

gpeng:
The error format is looking good now. What do you think about DRY-ing up that block by moving it into a separate concern?

darokel:
> The error format is looking good now. What do you think about DRY-ing up that block by moving it into a separate concern?

Done https://github.com/DFE-Digital/teacher-training-api/pull/1567/commits/c69f196da46288ba0e00063854acff2ba7dff141

gpeng (approved):

PR #1569, opened 2020-09-14
[88] Age range validation error message no longer display underscores

avinhurry:
### Context

The age range validation error message was incorrectly displaying underscores. 

The underscores no longer appear.

image

### Guidance to review

- Login to Publish 
- Create a new course 
- On the age range page select 'another age range'
- Enter an invalid range such as 0 to 50
- Click submit 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/requests/api/v2/providers/courses/create/age_range_in_years_spec.rb
+++ b/spec/requests/api/v2/providers/courses/create/age_range_in_years_spec.rb
@@ -81,41 +81,46 @@ RSpec.describe "POST /providers/:provider_code/courses/:course_code" do
 
     context "with a from value that does not fall within the valid age range" do
       let(:age_range_in_years) { "1_to_15" }
+      let(:age_range_in_years_formatted) { "1 to 15" }
+
 
       it "should return an error stating valid age ranges must be 4 years or greater" do
         expect(response).to have_http_status(:unprocessable_entity)
         expect(json_data.count).to eq 1
-        expect(response.body).to include "#{age_range_in_years} #{error_message}"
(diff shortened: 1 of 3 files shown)


avinhurry:
[[S] Age range validation error message displays underscores](https://trello.com/c/LRC74uOZ/88-s-age-range-validation-error-message-displays-underscores)

conorlyster (approved):
Product review approved

spec/requests/api/v2/providers/courses/create/age_range_in_years_spec.rb
```diff
@@ -85,7 +85,7 @@
       it "should return an error stating valid age ranges must be 4 years or greater" do
         expect(response).to have_http_status(:unprocessable_entity)
         expect(json_data.count).to eq 1
-        expect(response.body).to include "#{age_range_in_years} #{error_message}"
+        expect(response.body).to include "#{age_range_in_years.to_s.tr('_', ' ')} #{error_message}"
```
gpeng, line 88: What do you think about putting this into another `let` rather than doing the splitting in the expectation. Might be easier to read. e.g. `let(:expected_validation_age_range) { "1 to 15" }`. I prefer it when you can clearly see what you're expecting to happen rather than mentally executing the code. It is also exactly the code in the implementation so if there was an error in that code you wouldn't really see it in the test. e.g. if that returned ' ' for some reason it would return it in the code and the test so it would pass but wouldn't be what we actually wanted.

gpeng (changes requested):

gpeng (approved):

PR #1576, opened 2020-09-18
[130] service status dashboard rollover data

JonF1:
### Context
* https://trello.com/c/3C81CAYu/130-service-status-dashboard-rollover-data

Rollover data is needed for the new rollover tab on the performance dashboard as per the prototype. As a result, data needs to be added to the reporting endpoint on the API.

At the /reporting endpoint
Screenshot 2020-09-18 at 17 33 28

### Guidance to review
* https://publish-dashboard-prototype.herokuapp.com/with-tables#rollover  

N.B There's been a copy change to new courses 'published' instead of created.

### Checklist

- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/services/rollover_reporting_service.rb
+++ b/app/services/rollover_reporting_service.rb
@@ -1,0 +1,69 @@
+class RolloverReportingService
+  def initialize
+    @rollover_scope = RecruitmentCycle.current_recruitment_cycle
+  end
+
+  class << self
+    def call
+      new.call
+    end
+  end
+
+  def call
+    return {} if @rollover_scope.next.blank?
+
+    {
+      total: {
+          published_courses: published_courses_count,
+          new_courses_published: new_courses_published_count,
+          deleted_courses: deleted_courses_count,
+          existing_courses_in_draft: existing_courses_in_draft_count,
+          existing_courses_in_review: existing_courses_in_review,
+      },
+    }
+  end
+
+  private_class_method :new
+
+private
+
+  def next_findable
+    @rollover_scope.next.courses.findable
+  end
+
+  def published_courses
+    next_findable
+  end
+
+  def published_courses_count
+    published_courses.distinct.count
+  end
+
+  def new_courses_published
+    next_findable.distinct.created_at_since(RecruitmentCycle.next.created_at + 1.day)
+  end
+
+  def new_courses_published_count
+    new_courses_published.count
+  end
+
+  def deleted_courses
+    Course.discarded.where(provider: Provider.where(recruitment_cycle: RecruitmentCycle.next))
+  end
+
+  def deleted_courses_count
+    deleted_courses.count
+  end
+
+  def existing_courses_in_draft
+    @rollover_scope.next.courses.changed_since(RecruitmentCycle.next.created_at + 1.day).where.not(id: published_courses)
+  end
+
+  def existing_courses_in_draft_count
+    existing_courses_in_draft.distinct.count
+  end
+
+  def existing_courses_in_review
+    @rollover_scope.next.courses.where.not(id: published_courses).where.not(id: deleted_courses).where.not(id: existing_courses_in_draft).count
+  end
+end
--- a/app/services/statistic_service.rb
+++ b/app/services/statistic_service.rb
@@ -5,6 +5,7 @@ class StatisticService
       courses: CourseReportingService.call(courses_scope: recruitment_cycle.courses),
       publish: PublishReportingService.call(recruitment_cycle_scope: recruitment_cycle),
       allocations: AllocationReportingService.call(recruitment_cycle_scope: recruitment_cycle),
+      rollover: RolloverReportingService.call,
     }
   end
 
--- a/spec/services/rollover_reporting_service_spec.rb
+++ b/spec/services/rollover_reporting_service_spec.rb
@@ -1,0 +1,82 @@
+require "rails_helper"
+
+describe RolloverReportingService do
+  let(:published_running_site) { create(:site_status, :published, :running) }
+  let(:provider) { create(:provider, :next_recruitment_cycle) }
+  let(:published_course) { create(:course, provider: provider, site_statuses: [published_running_site]) }
+  let(:new_published_course) { create(:course, age: provider.recruitment_cycle.created_at + 28.hours, provider: provider, site_statuses: [published_running_site]) }
+  let(:deleted_course) { create(:course, :deleted, provider: provider) }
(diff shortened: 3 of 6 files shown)


JonF1:
[[S] Service Status Dashboard - rollover data](https://trello.com/c/3C81CAYu/130-s-service-status-dashboard-rollover-data)

app/services/statistic_service.rb
```diff
@@ -5,6 +5,7 @@ def self.reporting(recruitment_cycle:)
       courses: CourseReportingService.call(courses_scope: recruitment_cycle.courses),
       publish: PublishReportingService.call(recruitment_cycle_scope: recruitment_cycle),
       allocations: AllocationReportingService.call(recruitment_cycle_scope: recruitment_cycle),
+      rollover: (RolloverReportingService.call if RecruitmentCycle.next_recruitment_cycle.present?),
```
gpeng, line 8: We should make the call here regardless of whether the next cycle is present or not and have the service decide what to return in that case. It should probably return an empty object `{}` rather than nil.

gpeng (changes requested):

app/services/rollover_reporting_service.rb
```diff
@@ -0,0 +1,67 @@
...
+  class << self
+    def call
+      new.call
+    end
+  end
+
+  def call
+    {
+      total: {
+          published_courses: published_courses_count,
```
darokel, line 15: Could we sort out the tab spaces here?

darokel (changes requested):
I think we could we re-organise the `rollover_reporting_service_spec.rb` spec into an 'Arrange/Act/Assert' structure so that it's easier to visually process what's being tested. Perhaps move the fixture creation into before blocks and the `it` blocks simply do the assertions. You may need to use `context` to further group the scenarios based on the fixtures

app/services/statistic_service.rb
JonF1: Thanks Graham!

app/services/rollover_reporting_service.rb
JonF1: Fixed

JonF1:
> I think we could we re-organise the `rollover_reporting_service_spec.rb` spec into an 'Arrange/Act/Assert' structure so that it's easier to visually process what's being tested. Perhaps move the fixture creation into before blocks and the `it` blocks simply do the assertions. You may need to use `context` to further group the scenarios based on the fixtures. Give me a shout if you want to pair on this

Thanks for your help on this David

app/services/rollover_reporting_service.rb
```diff
@@ -0,0 +1,72 @@
...
+        published_courses: published_courses_count,
+        new_courses_published: new_courses_published_count,
+        deleted_courses: deleted_courses_count,
+        existing_courses_in_draft: existing_courses_in_draft_count,
+        existing_courses_in_review: existing_courses_in_review,
+      },
+    }
+    else
+      {}
+    end
```
defong, line 25: this don't look like its tests....

also is there an actual need to return anything?

app/services/rollover_reporting_service.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+
```
defong, line 28:
```suggestion
```

can be removed?

spec/services/rollover_reporting_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  describe ".call" do
+    subject(:result) { described_class.call }
+
+    context "number of published courses after rollover" do
+      before do
+        provider
+        published_course
+      end
+
+      it "published courses" do
```
darokel, line 21:
```suggestion
it "returns only published courses" do
```

spec/services/rollover_reporting_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+    end
+
+    context "number of new published courses after rollover" do
+      before do
+        provider
+        new_published_course
+      end
+
+      it "newly published courses" do
```
darokel, line 32:
```suggestion
it "returns the correct newly published courses" do
```

spec/services/rollover_reporting_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+    end
+
+    context "number of deleted courses after rollover" do
+      before do
+        provider
+        deleted_course
+      end
+
+      it "deleted courses" do
```
darokel, line 43:
```suggestion
it "returns the correct deleted courses" do
```

spec/services/rollover_reporting_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    context "number of courses after rollover that are in draft" do
+      before do
+        provider
+        course_in_draft
+        second_course_in_draft
+        new_published_course
+      end
+
+      it "draft" do
```
darokel, line 57:
```suggestion
it "returns the correct draft courses" do
```

spec/services/rollover_reporting_service_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    context "number of courses after rollover that are in review" do
+      before do
+        provider
+        course_in_draft
+        new_published_course
+        deleted_course
+        published_course
+      end
+
+      it "in review" do
```
darokel, line 71:
```suggestion
it "returns the correct courses in review" do
```

darokel (approved):
Nearly there, left a few suggestions around wording but overall 👍

app/services/rollover_reporting_service.rb
JonF1: fixed, cheers

SamJamCul (approved):
Looks good!

app/services/rollover_reporting_service.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  class << self
+    def call
+      new.call
+    end
+  end
+
+  def call
+    if @rollover_scope.next.present?
```
gpeng, line 13: This should be a guard clause

```suggestion
    return {} unless @rollover_scope.next.present?
    
```

gpeng (changes requested):

app/services/rollover_reporting_service.rb
JonF1: thanks

gpeng (approved):

PR #1579, opened 2020-09-25
[167] Implement update endpoint for contacts

darokel:
### Context

- https://trello.com/c/NLKqADvh/167-add-update-endpoint-for-ucas-contacts-in-api-v2

- Add a new endpoint to update individual contacts for a provider

### Guidance to review

- Load the rails server locally and fire a request to update a contact: `PUT 'localhost:3001/api/v2/contacts/:id`
- Assert the contact is updated.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/contacts_controller.rb
+++ b/app/controllers/api/v2/contacts_controller.rb
@@ -1,0 +1,32 @@
+module API
+  module V2
+    class ContactsController < API::V2::ApplicationController
+      deserializable_resource :contact, class: API::V2::DeserializableContact
+
+      def update
+        authorize contact
+
+        if contact.update(contact_params)
+          render jsonapi: contact, status: :ok
+        else
+          render jsonapi_errors: contact.errors, status: :unprocessable_entity
+        end
+      end
+
+    private
+
+      def contact
+        @contact ||= Contact.find(params[:id])
+      end
+
+      def contact_params
+        params.require(:contact)
+          .permit(
+            :name,
+            :email,
+            :telephone,
+          )
+      end
+    end
+  end
+end
--- a/app/serializers/api/v2/serializable_provider.rb
+++ b/app/serializers/api/v2/serializable_provider.rb
@@ -85,6 +85,7 @@ module API
 
       has_many :sites
       has_many :users
+      has_many :contacts
 
       has_many :courses do
         meta do
--- a/spec/serializers/api/v2/serializable_contact_spec.rb
+++ b/spec/serializers/api/v2/serializable_contact_spec.rb
@@ -1,0 +1,19 @@
+require "rails_helper"
+
+describe API::V2::SerializableContact do
+  let(:contact) { build(:contact) }
+  let(:resource) { described_class.new(object: contact) }
+
+  it "sets type to contacts" do
+    expect(resource.jsonapi_type).to eq(:contacts)
+  end
+
+  subject { JSON.parse(resource.as_jsonapi.to_json) }
+
+  it { should have_type "contacts" }
+
+  it { should have_attribute(:name).with_value(contact.name) }
+  it { should have_attribute(:email).with_value(contact.email) }
+  it { should have_attribute(:telephone).with_value(contact.telephone) }
+  it { should have_attribute(:type).with_value(contact.type) }
+end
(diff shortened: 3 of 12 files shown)


darokel:
[Add update endpoint for ucas contacts in API v2.](https://trello.com/c/NLKqADvh/167-add-update-endpoint-for-ucas-contacts-in-api-v2)

defong (changes requested):
failing tests

darokel:
> failing tests

Pushed up some fixes

spec/serializers/api/v2/serializable_contact_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  let(:contact) { build(:contact) }
+  let(:resource) { described_class.new(object: contact) }
+
+  it "sets type to contacts" do
+    expect(resource.jsonapi_type).to eq(:contacts)
+  end
+
+  subject { JSON.parse(resource.as_jsonapi.to_json) }
+
+  it { should have_type "contacts" }
```
gpeng, line 13: We should (no pun intended) use `is_expected.to` rather than `should` here.

gpeng (changes requested):
This looks good for this ticket ie. 'update' but I think we at least need `show` which we have a use case for in the functionality that we're implementing. We shouldn't be getting the whole provider (which is one of our slowest endpoints) to edit a contact. We also don't have create/delete which we need a product discussion about. Where do the contacts get created initially?

app/serializers/api/v2/serializable_provider.rb
```diff
@@ -85,6 +85,7 @@ class SerializableProvider < JSONAPI::Serializable::Resource
 
       has_many :sites
       has_many :users
+      has_many :contacts
```
defong, line 88: This looks wrong

https://github.com/DFE-Digital/teacher-training-api/blob/471e479c7cf7e84de5c2dd37263cf70a19b06ba1/app/serializers/api/v2/serializable_provider.rb#L50-L68


https://github.com/DFE-Digital/teacher-training-api/blob/a97b30d65a5e2f357e696d0a5f406e6d1b144fb2/app/models/contact.rb#L14-L21

It looks like its `has_one` x 5 times
https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/has_one

defong (changes requested):

app/serializers/api/v2/serializable_provider.rb
darokel: We're steering away from this approach so it's more agnostic and not have to define an attribute for every contact.

gpeng: I'm going to add another ticket to remove those so it's not part of this PR.

spec/serializers/api/v2/serializable_contact_spec.rb
darokel: done https://github.com/DFE-Digital/teacher-training-api/pull/1579/commits/d1a210cdbf10e1cca1e2c95897167a09da4a88ac

gpeng:
Discussed offline. This PR is update only. Show will be added in https://trello.com/c/X1EkaWrR/184-add-api-v2-contacts-id-show-endpoint. Create and destroy will be scoped for future work as will removing the now redundant individual contact elements on the provider response.

gpeng (approved):

app/controllers/api/v2/contacts_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    class ContactsController < API::V2::ApplicationController
+      deserializable_resource :contact, class: API::V2::DeserializableContact
+
+      def update
+        authorize contact
+
+        if contact.update(contact_params)
+          render jsonapi: contact, status: :ok
+        else
+          render jsonapi_errors: contact.errors, status: :unprocessable_entity
```
defong, line 12: Can we have the unhappy path tests please.

darokel: done https://github.com/DFE-Digital/teacher-training-api/pull/1579/commits/996cb2eaadcae1d8aa7cb9fa0755d5682656c18b

defong (approved):

PR #1584, opened 2020-09-30
[202] Add permission_given field to Contact

gpeng:
### Context

We are adding functionality to update contacts on Publish. Explicit permission is required when a user updates a contact.

### Changes

Add `permission_given` field to the `Contact` model. This allows us to validate and return errors to the client.

**Slightly contentious bit:** We haven't migrated data for existing contact records as permission is required whenever the record is saved so it will be updated then. We could include an update in the migration but the existing records didn't get explicit permission in the same way as they were manually added.


defong:
[Add permission field to contact update](https://trello.com/c/i5B4KMXI/202-add-permission-field-to-contact-update)

defong:
It should be fine to default existing contact's `permission_given` to `true`, as by virtue of manual intervention, permission would of been granted/implied.

So `permission_given` does that mean that if saved it will always be true?

gpeng:
> It should be fine to default existing contact's `permission_given` to `true`, as by virtue of manual intervention, permission would of been granted/implied.
> 
> So `permission_given` does that mean that if saved it will always be true?

Yeah... it almost makes it redundant but having the field makes error handling etc much easier.

darokel:
I think we want to default the boolean db value to false unless permission is explicitly given which we then set to true? @gpeng @defong

gpeng:
> I think we want to default the boolean db value to false unless permission is explicitly given which we then set to true? @gpeng @defong 

Yeah, I wasn't really sure on that. It can only ever be true as we require that in the validations before we create but it's a bit tricky with the existing data. I can default it to false and then update the existing records after it's deployed 🤷

defong:
> > I think we want to default the boolean db value to false unless permission is explicitly given which we then set to true? @gpeng @defong
> 
> Yeah, I wasn't really sure on that. It can only ever be true as we require that in the validations before we create but it's a bit tricky with the existing data. I can default it to false and then update the existing records after it's deployed shrug

If is in the database, right now, then it should be `true`, as permission would of inherently been given.

gpeng:
@defong & @darokel I've added `false` as a default and set existing `Contact#permission_given` to `true` as part of the migration now if you could give this another look please.

darokel (approved):

defong (approved):

dankmitchell (approved):

PR #1590, opened 2020-10-06
[239] Fix rollover reporting information

aliuk2012:
## Context
There was an error in Publish performance dashboard where the page failed
to load because the next recruitment cycle is not created until we do
rollover again.

We decided that this should really be handled the api and instead of
attempting to count data that doesn't exist we decided that the guard
would instead return a default data structure with zero out figures.

We tested this locally and this worked with publish because the rollover
tab is hidden from the user until the next cycle is created and we
are in rollover state.

Co-authored-by: SamJamCul <samuel.culley@digital.education.gov.uk>
Co-authored-by: Jonathan Filar <58793682+JonF1@users.noreply.github.com>

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/services/rollover_reporting_service.rb
+++ b/app/services/rollover_reporting_service.rb
@@ -10,17 +10,25 @@ class RolloverReportingService
   end
 
   def call
-    return {} if @rollover_scope.next.blank?
-
-    {
-      total: {
+    if @rollover_scope.next.blank?
+      { total: {
+          published_courses: 0,
+          new_courses_published: 0,
+          deleted_courses: 0,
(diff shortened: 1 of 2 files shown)


aliuk2012:
[[S] Don't display rollover info in dashboard](https://trello.com/c/gX88KzVB/239-dont-display-rollover-info-in-dashboard)

defong (approved):

darokel (approved):

dankmitchell (approved):

PR #1598, opened 2020-10-09
[236] Upgrade to Ruby 2.7.2

gpeng:
### Context

API is currently using Ruby 2.6.5

* Upgrade Ruby to 2.7.2
* Upgrade Bundler to 2.1.4
* Update the docs bundle to get the container to build

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


gpeng:
[Bump ruby version](https://trello.com/c/GdzjA1UU/236-bump-ruby-version)

.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.6.5
+2.7.2
```
aliuk2012, line 1: is this needed because we have `.tool-versions` and we are using asdf on registers

gpeng: Probably not. I figured it was painless enough to keep around in case people are using rbenv.

aliuk2012 (approved):

defong (approved):

PR #1606, opened 2020-10-13
Add nested Arm template for App service plan metric

felixtheflex:
### Context

Add nested arm template for App Service Plan metric alert. This allows alert to be sent, when certain threshold is reached e.g. CPU utilisation reaches over 80%

Add nested arm template for App Service Plan metric alert

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -319,7 +319,15 @@
       "metadata": {
       "description": "Configure Number of database server CPU"
       }
-    }
+    },
+  "actionGroupName": {
+      "type": "string",
+      "defaultValue": "",
+      "minLength": 1,
+      "metadata": {
+          "description": "Name of ActionGroup to send notification"
+      }
+  }
   },
   "variables": {
     "deploymentUrlBase": "https://raw.githubusercontent.com/DFE-Digital/bat-platform-building-blocks/master/templates/",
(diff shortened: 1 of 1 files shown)


azure/template.json
```diff
@@ -319,6 +319,38 @@
       "metadata": {
       "description": "Configure Number of database server CPU"
       }
+    },
+    "subscription": {
```
vigneshmsft, line 323: use a more descriptive name for this and `resourceGroup` parameters?
prefix appServicePlan or `monitoring` or a description of what it refers to

felixtheflex:
[Monitor high CPU / high memory usage](https://trello.com/c/wbn59PXr/279-monitor-high-cpu-high-memory-usage)

azure/template.json
felixtheflex: done... also, i have decided to remove the resourceInstace parameter, as this is nothing but the appservicename.... So i passed the variable called "appServicePlanName" instead to the parameter to the parameter of the arm template...

vigneshmsft: thanks..also do you need to pass the `subscriptionId` ? 
can you use `subscription().id` or `subscription().subscriptionId` or use the `subscriptionResourceId()` function inside the linked template?

felixtheflex: good idea..!

azure/template.json
```diff
@@ -319,6 +319,22 @@
...
+    },
+  "actionGroupName": {
+      "type": "string",
+      "defaultValue": "",
+      "minLength": 1,
+      "metadata": {
+          "description": "Name of ActionGroup to send notification"
+      }
+  },
+  "appServicePlanResourceGroup": {
```
saliceti, line 331: Do we need a new parameter?
Can't we use resourcegroup() funstion like here: https://github.com/DFE-Digital/teacher-training-api/blob/62dc4e4f40e18a3f3dafcc13b2436169f9dd4407/azure/template.json#L494 ?

vigneshmsft (approved):

azure/template.json
felixtheflex: done....

saliceti (approved):

PR #1609, opened 2020-10-15
[252] Improve v3/providers/<id> performance.

gpeng:
### Context

We had an incident recently and one of the contributing factors was the poor performance of the API endpoints, particularly the course and provider endpoints.

This is part one of the remedial work to address that. 

Improve performance of the `providers#show` endpoint. Most of the performance issues on this endpoint are caused by N+1 queries. We support deep nesting of results in the json-api format. The client can supply their own `include` parameter to specify the results they require but in reality there are only two different requests for this. We weren't including the requested children in the database query. Supporting this dynamically would be quite complex as the database structure doesn't map 1-1 with the response structure that we provide. So... in this PR we are hard coding the includes. 

It's difficult to fully gauge the improvement until we get to production but equivalent requests to QA look to be > 2 times faster. There is some performance penalty at the database end which may adversely effect providers with fewer courses but these weren't causing us problems anyway. We've also doubled the CPU resource on the DB in production so there will likely still be a net gain.

There are still some N+1s in this request, most notably the `accrediting_provider` call from the course serializer. This scope is parameterised and can't easily be included without restructuring the data.

We have moved away from allowing this deep nesting in public/v1 and should look to moving clients to that.

* Add includes for the nested tables
* Remove some N+1 issues in `Course` methods that were using the `latest_enrichment` scope
* Remove unused ucas contacts methods in the provider serializer.
* Enable Skylight in QA to support investigations without affecting production

### Guidance to review

This endpoint is used from Apply. To test performance on QA I SSH-ed onto QA Find on PAAS and used  `Provider.where(recruitment_cycle_id: 3).find(<get a slow provider_code from kibana>)`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


darokel (approved):
👍

gpeng:
[Investigate: Go through Skylight and understand our slow requests](https://trello.com/c/xaLC39wP/252-investigate-go-through-skylight-and-understand-our-slow-requests)

defong (approved):
awesome

PR #1610, opened 2020-10-15
284 add query includes on the v3 providersindex endpoint

SamJamCul:
### Context
https://trello.com/c/wlaJ1u1D/284-add-query-includes-on-the-v3-providersindex-endpoint

* Include recruitment cycle when using v3 providers_index, helps with searching through large numbers of results. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/providers_controller.rb
+++ b/app/controllers/api/v3/providers_controller.rb
@@ -4,15 +4,13 @@ module API
       before_action :build_recruitment_cycle
 
       def index
+        if params[:search].present?
+          return render(status: :bad_request) if params[:search].length < 2
+        end
+
         build_fields_for_index
-
-        @providers = if params[:search].present?
-                       return render(status: :bad_request) if params[:search].length < 2
-
-                       @recruitment_cycle.providers.search_by_code_or_name(params[:search])
-                     else
-                       @recruitment_cycle.providers
-                     end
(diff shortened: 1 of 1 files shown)


SamJamCul:
[API Performance: Add query includes on the v3 `providers#index` endpoint](https://trello.com/c/wlaJ1u1D/284-api-performance-add-query-includes-on-the-v3-providersindex-endpoint)

defong (approved):
maybe a bit hard to read

line 7 vs line 13

app/controllers/api/v3/providers_controller.rb
```diff
@@ -4,15 +4,13 @@ class ProvidersController < API::V3::ApplicationController
       before_action :build_recruitment_cycle
 
       def index
-        build_fields_for_index
-
-        @providers = if params[:search].present?
-                       return render(status: :bad_request) if params[:search].length < 2
+        if params[:search].present?
```
darokel, line 7: I think you could just shorten this down to `if params[:search]&.length < 2` and remove the additional check on line 8.

darokel (approved):

app/controllers/api/v3/providers_controller.rb
gpeng: I thought that too but that evaluates to `if nil < 2` and errors if `params[:search]` isn't present.

PR #1617, opened 2020-10-19
[252] Add includes for course index

gpeng:
### Context

Part of work to improve performance of the V3 endpoints.

Include relationships in the course query in `v3::CoursesController#index`

The CourseSearchService handles this query. This commit adds includes to the outer query scope within the service. This reduces the N+1 issues but doesn't relate to the includes parameters accepted at the controller. This endpoint is used by Find so we probably don't need that kind of flexibility.

The performance improvement we'll get in production is difficult to gauge as we don't have parity on the QA environment but locally there is a significant improvement and query definitely looks better in Skylight. I'll get some better figures from Kibana once it's deployed.

Before:
Local: 2020-10-19 16:44:33.995788 I [27181:puma threadpool 002] (1.447s) API::V3::CoursesController -- Completed #index

image

After: 
2020-10-19 17:08:01.947578 I [27181:puma threadpool 003] (677.6ms) API::V3::CoursesController -- Completed #index

image

### Guidance to review

Find should still work. I've run this branch on QA and all seems good.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -1,5 +1,9 @@
 class CourseSearchService
-  def initialize(filter:, sort: nil, course_scope: Course)
+  def initialize(
+    filter:,
+    sort: nil,
+    course_scope: Course
+  )
     @filter = filter || {}
     @course_scope = course_scope
     @sort = Set.new(sort&.split(","))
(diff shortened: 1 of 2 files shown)


app/services/course_search_service.rb
```diff
@@ -26,7 +30,12 @@ def call
     # The 'where' scope will remove duplicates
     # An outer query is required in the event the provider name is present.
     # This prevents 'PG::InvalidColumnReference: ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list'
-    outer_scope = Course.where(id: scope.select(:id))
+    outer_scope = Course.includes(
+      :enrichments,
+      :subjects,
```
defong, line 35:
```suggestion
      subjects: [:financial_incentive],
```

probaby this on as well

defong (approved):

app/services/course_search_service.rb
gpeng: Good spot. I didn't see N+1 from that as I didn't select the salary filter when I was testing 🤦 I've updated 👍

darokel (approved):
👍 ⚡

PR #1622, opened 2020-10-26
[342] revert duration/duration_ms formatting for logstash

misaka:
Turns out this approach was not compatible with the standard way of logging and
can result in breaking the indexing for the 'duration' field, breaking
aggregation and other operations reliant on the 'duration' field. The correct
approach is to leave 'duration_ms', fix any discrepencies in the logit filters
to ensure that it's indexed as a number, and then use 'duration_ms' for
aggregation.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/initializers/semantic_logger.rb
+++ b/config/initializers/semantic_logger.rb
@@ -26,12 +26,6 @@ class LogStashFormatter < SemanticLogger::Formatters::Raw
     end
   end
 
-  # The value here appears to break logging to logstash / elasticsearch
-  def format_duration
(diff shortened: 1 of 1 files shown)


misaka:
[Revert duration formatting for logstash](https://trello.com/c/7NjgHEx6/342-revert-duration-formatting-for-logstash)

saliceti (approved):

gpeng (approved):

PR #1627, opened 2020-10-27
[203] Remove 'mailer' sidekiq queue

gpeng:
### Context

We were using this queue for mailer jobs but it isn't default so we changed it in a83a0a8.

This removes the old queue which should now be empty.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -3,5 +3,4 @@
   - default
   - geocoding
   - save_statistic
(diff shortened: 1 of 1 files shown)


darokel (approved):

vigneshmsft (approved):
LGTM

PR #1635, opened 2020-11-05
[372] Fix course enrichment bug

jhenderson:
### Context
When users update a course, the changes would sometimes not show up in preview mode. This was because `API::V2::SerializableCourse.enrichment_attribute` didn't explicitly request for the latest course enrichment record (ordered by created_at). Using `last()` on its own doesn't guarantee the order we want.

With regards to performance, since the same query is executed multiple times within the same action, ActiveRecord will cache the query `@object.enrichments.most_recent&.first&.public_send(enrichment_name)`, helping us to avoid the N + 1 problem.

Use an ordered scope instead of calling `last()` to fetch the latest course enrichment record.

Introducing this fix also caused other tests to fail as they depended on `API::V2::SerializableCourse.enrichment_attribute` using the `last()` approach.

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -94,8 +94,7 @@ class Course < ApplicationRecord
            source: :subject,
            class_name: "ModernLanguagesSubject"
 
-  has_many :enrichments,
-           class_name: "CourseEnrichment" do
+  has_many :enrichments, class_name: "CourseEnrichment" do
     def find_or_initialize_draft
       # This is a ruby search as opposed to an AR search, because calling `draft`
       # will return a new instance of a CourseEnrichment object which is different
@@ -107,7 +106,7 @@ class Course < ApplicationRecord
     end
 
     def new_draft_attributes
-      latest_published_enrichment = latest_first.published.first
+      latest_published_enrichment = most_recent.published.first
 
       if latest_published_enrichment.present?
         latest_published_enrichment_attributes = latest_published_enrichment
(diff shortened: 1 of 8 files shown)


jhenderson:
[Course enrichment bug](https://trello.com/c/Z2YZ8OpV/372-course-enrichment-bug)

config/settings/development.yml
```diff
@@ -5,3 +5,4 @@ mcbg:
 system_authentication_token: "Ge32"
 log_level: debug
 gcp_api_key: please_change_me
+developer_auth: false
```
defong, line 8: no need for this change...

you can over override with `development.local.yml`

defong (approved):
Looks find to me.

@gpeng you ok with the scope approached?
https://github.com/DFE-Digital/teacher-training-api/blob/d9d2395d53ff0aa70fafaa54252b03f458b30332/app/models/course.rb#L569-L571

config/settings/development.yml
jhenderson: Good spot. I wasn't intending to add this file.

jhenderson:
@defong @gpeng If we use `max_by`, which is a Ruby method on `Enumerable`, then we'll end up loading all the ActiveRecord objects into memory before calling `max_by` - we should try to avoid that.

dankmitchell (approved):

PR #1643, opened 2020-11-11
Upgrade to Terraform 0.13 and modules

vigneshmsft:
### Context

Upgrade to terraform configuration to 0.13

Modules for PaaS and StatusCake
Upgrade to terraform 0.13

next steps: remove `SETTINGS_` variables and use YAML config file for non-secrets (in repo)  and secrets (in github secrets)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


felixtheflex (approved):

terraform/modules/paas/resources.tf
```diff
@@ -0,0 +1,61 @@
...
+  }
+  routes {
+    route = cloudfoundry_route.web_app_cloudapps_digital_route.id
+  }
+}
+
+resource cloudfoundry_app worker_app {
+  name              = local.worker_app_name
+  space             = data.cloudfoundry_space.space.id
+  health_check_type = "port"
```
saliceti, line 27: I think `process` is more adapted for a background job

terraform/modules/paas/resources.tf
```diff
@@ -0,0 +1,61 @@
...
+  domain   = data.cloudfoundry_domain.local.id
+  space    = data.cloudfoundry_space.space.id
+  hostname = local.web_app_name
+}
+
+resource cloudfoundry_service_instance postgres {
+  name         = local.postgres_service_name
+  space        = data.cloudfoundry_space.space.id
+  service_plan = data.cloudfoundry_service.postgres.service_plans["small-11"]
+  json_params  = "{\"enable_extensions\": [\"pg_buffercache\",\"pg_stat_statements\", \"plpgsql\", \"btree_gin\", \"btree_gist\"]}"
```
saliceti, line 53: Can we confirm these extensions are required?
 @gpeng @misaka maybe?

terraform/modules/paas/resources.tf
```diff
@@ -0,0 +1,61 @@
...
+}
+
+resource cloudfoundry_service_instance redis {
+  name         = local.redis_service_name
+  space        = data.cloudfoundry_space.space.id
+  service_plan = data.cloudfoundry_service.redis.service_plans["tiny-5_x"]
+  json_params  = "{\"maxmemory_policy\": \"allkeys-lfu\" }"
```
saliceti, line 60: Can we confirm this customisation is required?
@gpeng @misaka maybe?
This is for [caching](https://redis.io/topics/lru-cache) but I believe we don't use redis for caching?

terraform/modules/paas/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+
+variable docker_image {}
+
+variable app_environment {}
+
+variable app_environment_variables { type = map }
+
+locals {
+  web_app_name          = "teacher-training-api-${var.app_environment}"
+  worker_app_name       = "teacher-training-api-bgwkr-${var.app_environment}"
```
saliceti, line 19: "worker" and "postgres" would be more consistent with TV and GiT

terraform/terraform.tf
```diff
@@ -0,0 +1,50 @@
...
+      version = "1.0.0"
+    }
+  }
+  backend "azurerm" {
+  }
+}
+
+provider cloudfoundry {
+  api_url      = local.cf_api_url
+  user         = var.cf_user != "" ? var.cf_user : null
```
saliceti, line 23: I found we can simplify by setting the default valut to `null` in variable section. See:
- https://github.com/DFE-Digital/teacher-vacancy-service/blob/terraform-monitoring-0-13-4/terraform/monitoring/variables.tf#L1
- https://github.com/DFE-Digital/teacher-vacancy-service/blob/terraform-monitoring-0-13-4/terraform/monitoring/provider.tf#L4

terraform/workspace_variables/production.tfvars
```diff
@@ -1,6 +1,20 @@
-alerts =  {
-ttapi = {
-    website_name = "prod-teacher-training-api"
+#PaaS
+cf_space                  = "bat-prod"
+paas_app_environment      = "prod"
+paas_web_app_instances    = 1
```
saliceti, line 4: Instances in prod (web and worker) should be at least 2

terraform/modules/paas/resources.tf
vigneshmsft: @saliceti this is the current list
|       Name        | Version |   Schema   |                        Description                        |
--------------------|---------|------------|-----------------------------------------------------------|
| btree_gin          | 1.0     | public     | support for indexing common datatypes in GIN |
| btree_gist         | 1.2     | public     | support for indexing common datatypes in GiST |
| pg_buffercache     | 1.2     | public     | examine the shared buffer cache |
| pg_stat_statements | 1.4     | public     | track execution statistics of all SQL statements executed |
| plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language |

terraform/modules/paas/main.tf
```diff
@@ -0,0 +1,62 @@
...
+
+resource cloudfoundry_app worker_app {
+  name              = local.worker_app_name
+  space             = data.cloudfoundry_space.space.id
+  health_check_type = "process"
+  instances         = var.worker_app_instances
+  memory            = var.worker_app_memory
+  docker_image      = var.docker_image
+  command           = local.worker_app_start_command
+  strategy          = "blue-green-v2"
```
saliceti, line 32: It's actually fine to use default strategy for workers and it's faster

terraform/modules/paas/resources.tf
gpeng: We're definitely using the btree stuff. I think we have a use case for `uuid-ossp` by the look of it. There's a card in this sprint to start generating UUIDs on records and it may be useful.

terraform/modules/paas/resources.tf
gpeng: We don't use redis for cacheing currently. Only for storing async jobs for sidekiq. We might want to look at using it (or memcached) though just not at the moment.

terraform/modules/paas/resources.tf
saliceti: Thanks. I would remove plpgsql at least. We don't use stored procedures and hopefully we never will.

vigneshmsft: removed `plpgsql` extension

defong: `uuid-ossp` can we add that in please, 

There is dev work will involves GUID/UUID that will throw PaaS back into the dark ages.


```
select uuid_generate_v4() 
```

Without it we will be forced to generate uuid in `ruby` land and we know how rubbish that is.

vigneshmsft: appears `uuid-ossp` exists by default, the above query also returns `UUID`

saliceti (approved):

PR #1669, opened 2020-11-30
Update the provider show endpoint to use the V3 ProviderSerializer

davidgisbey:
### Context

At the moment the CourseSerializersServiceV3 uses the V2 ProviderSerializer. It should really be using the V3 ProviderSerializer. This PR uses the v3 serializer and updates it to include all the requisite attributes for Find and Apply.

- Updates the CourseSerializersServiceV3 to use the V3 ProviderSerializer
- Updates the attrs used in the V3 ProviderSerializer
- Updates the provider show and V3 Provider Serializer tests
- Return a limited number of provider attrs on the provider suggestions endpoint

### Guidance to review

Is this everything that needs to be done?

I've tested find locally and it all seems to be working as expected.
 
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/provider_suggestions_controller.rb
+++ b/app/controllers/api/v3/provider_suggestions_controller.rb
@@ -14,6 +14,8 @@ module API
         render(
           jsonapi: found_providers,
           class: { Provider: SerializableProvider },
+          fields: { providers: %i[provider_code provider_name provider_type
+                                  latitude longitude recruitment_cycle_year] },
         )
       end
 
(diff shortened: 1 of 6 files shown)


dankmitchell:
Do we need to update the docs to match @davidgisbey ?

davidgisbey:
We've actually moved over the public API this morning on Apply and are planning on moving Find over this sprint so I'm just going to close this since it's going to be deprecated shortly anyway.

PR #1675, opened 2020-12-04
[600] Added throttling

defong:
### Context
throttling

Added throttling

### Guidance to review

Run
```bash
rails dev:cache 
```
Then spank the server

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -112,6 +112,9 @@ gem "pg_search"
 # End-user application performance monitoring
 gem "skylight"
 
+# Rack middleware for blocking & throttling abusive requests
+gem "rack-attack"
+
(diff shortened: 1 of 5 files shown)


gpeng (commented):
We should set the limits and enable in production. Can you work out a reasonable requests per minute limit by checking logit?

defong:
> We should set the limits and enable in production. Can you work out a reasonable requests per minute limit by checking logit?

There is the wider issue of caching, do we need an external store, or each to their own.

defong:
This is a **NO** go.

PR #1690, opened 2020-12-14
[625] Fixed pagination for public v1 endpoints

defong:
### Context
Pagination for public v1 endpoints
Adhere to https://jsonapi.org/format/#fetching-pagination

### Guidance to review

http://localhost:3001/api/public/v1/recruitment_cycles/2021/providers

image

http://localhost:3001/api/public/v1/recruitment_cycles/2021/providers/T92/courses
image

http://localhost:3001/api/public/v1/recruitment_cycles/2021/courses?page=180
image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/application_controller.rb
+++ b/app/controllers/api/public/v1/application_controller.rb
@@ -11,12 +11,32 @@ module API
           render jsonapi: nil, status: :not_found
         end
 
+        def jsonapi_pagination(collection)
+          collection.present? && pagy_scope.present? ? pagination_links : {}
+        end
+
       private
 
-        def paginate(scope)
-          _pagy, paginated_records = pagy(scope, items: per_page, page: page)
+      # child must define pagy_scope method for paginations related items
+        def pagy_scope; end
 
-          paginated_records
+        def pagy_results
+          @pagy_results ||= pagy(pagy_scope, items: per_page, page: page)
+        end
+
+        def paginated_records
+          @paginated_records ||= pagy_results.second
+        end
+
+        def pagination_links
+          meta = pagy_metadata(pagy_results.first, urls: true)
+
+          {
+            first: meta[:first_url],
+            last: meta[:last_url],
+            prev: meta[:prev].nil? ? nil : meta[:prev_url],
+            next: meta[:next].nil? ? nil : meta[:next_url],
+          }
         end
 
         def per_page
--- a/app/controllers/api/public/v1/application_controller.rb
+++ b/app/controllers/api/public/v1/application_controller.rb
@@ -10,66 +10,6 @@ module API
         def jsonapi_404
           render jsonapi: nil, status: :not_found
         end
-
-        def jsonapi_pagination(collection)
-          collection.present? && pagy_scope.present? ? pagination_links : {}
-        end
-
-      private
-
-      # child must define pagy_scope method for paginations related items
-        def pagy_scope; end
-
-        def pagy_results
-          @pagy_results ||= pagy(pagy_scope, items: per_page, page: page)
-        end
-
-        def paginated_records
-          @paginated_records ||= pagy_results.second
-        end
-
-        def pagination_links
-          meta = pagy_metadata(pagy_results.first, urls: true)
-
-          {
-            first: meta[:first_url],
-            last: meta[:last_url],
-            prev: meta[:prev].nil? ? nil : meta[:prev_url],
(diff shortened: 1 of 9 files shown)


defong:
[Pagination isn't fully implemented on the `/public/v1` endpoints.](https://trello.com/c/sMsoC3wr/625-pagination-isnt-fully-implemented-on-the-public-v1-endpoints)

defong:
Not merging until @duncanjbrown agrees to **NOT** roast me for in place of a turkey for Xmas :christmas_tree: :santa: :gift: :turkey:

app/controllers/api/public/v1/application_controller.rb
```diff
@@ -11,12 +11,32 @@ def jsonapi_404
...
+        end
+
+        def pagination_links
+          meta = pagy_metadata(pagy_results.first, urls: true)
 
-          paginated_records
+          {
+            first: meta[:first_url],
+            last: meta[:last_url],
+            prev: meta[:prev].nil? ? nil : meta[:prev_url],
```
darokel, line 37:
```suggestion
meta[:prev].presence
```

app/controllers/api/public/v1/application_controller.rb
```diff
@@ -11,12 +11,32 @@ def jsonapi_404
...
+        def jsonapi_pagination(collection)
+          collection.present? && pagy_scope.present? ? pagination_links : {}
+        end
+
       private
 
-        def paginate(scope)
-          _pagy, paginated_records = pagy(scope, items: per_page, page: page)
+      # child must define pagy_scope method for paginations related items
+        def pagy_scope; end
```
darokel, line 21: It's starting to feel like these pagy related methods should live in a separate module as there's a handful of them woven into our `ApplicationController`. These methods seem to just customise Pagy to work to our requirements.

defong: Done it in this commit https://github.com/DFE-Digital/teacher-training-api/pull/1690/commits/eb291d9ed80ae11ca31f9360a62623b5918194df

darokel (approved):

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -2,16 +2,17 @@ module API
...
+          render jsonapi: paginated_records,
                  include: include_param,
                  class: API::Public::V1::SerializerService.call
         end
 
       private
 
-        def courses
-          @courses ||= CourseSearchService.call(
+        def pagy_scope
```
gpeng, line 14: I think this was better as it was before. `paginated_records` in the module needing `page_scope` out here in the controller doesn't seem as 'obvious' as just calling `paginate(courses)`

app/controllers/concerns/pagy_pagination.rb
```diff
@@ -0,0 +1,63 @@
...
+  extend ActiveSupport::Concern
+
+  def jsonapi_pagination(collection)
+    collection.present? && pagy_scope.present? ? pagination_links : {}
+  end
+
+private
+
+  # child must define pagy_scope method for paginations related items
+  def pagy_scope; end
```
gpeng, line 11: Still not sure this is the clearest way to do it but if we are going to use this then we should raise something here if the including class is required to implement this. Otherwise it will fail silently. I always use `NotImplementedError` but that's not really what it's supposed to be for I don't think. Sounds right though :)

gpeng (commented):
Can we test the pagination a bit less? Seems like we're testing the same code with the same tests 3 times. Can we test the module itself and then just test the controller `include?(PagyPagination)` or something like that?

app/controllers/concerns/pagy_pagination.rb
defong:
``` ruby
def pagy_scope; 
  raise NotImplementedError, "must define pagy_scope method"
end
```

you mean do something like this?

gpeng: Yes. Well, yes if we're going to take this approach. I still think having an explicit `paginate(scope)` method is clearer in the controllers.

defong: opted with `paginate(scope)`

app/controllers/api/public/v1/courses_controller.rb
defong: opted with `paginate(scope)`

gpeng (approved):

PR #1697, opened 2020-12-18
Terraform commands in Makefile

vigneshmsft:
Make commands to deploy to qa, staging and production.
Be able to create and destroy review apps.

### DIFF
--- a/terraform/workspace_variables/review.tfvars
+++ b/terraform/workspace_variables/review.tfvars
@@ -1,0 +1,13 @@
+#PaaS
+cf_space                   = "bat-qa"
+paas_web_app_instances     = 1
+paas_web_app_memory        = 512
+paas_worker_app_instances  = 1
+paas_worker_app_memory     = 512
+paas_postgres_service_plan = "tiny-unencrypted-11"
+paas_redis_service_plan    = "tiny-5_x"
+
+paas_app_config = {
+  RAILS_ENV                = "review"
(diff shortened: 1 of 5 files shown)


felixtheflex (approved):

terraform/workspace_variables/review.tfvars
```diff
@@ -0,0 +1,18 @@
...
+
+paas_app_config = {
+  RAILS_ENV                = "review"
+  RAILS_SERVE_STATIC_FILES = true
+}
+
+# StatusCake
+statuscake_username = "Review"
+statuscake_password = "Review"
+statuscake_alerts   = null
```
saliceti, line 18: You may want to add `default = {}` in variables.tf and remove this
Then you can even remove `count  = var.statuscake_alerts != null ? 1 : 0`

terraform/workspace_variables/review.tfvars
```diff
@@ -0,0 +1,18 @@
...
+paas_postgres_service_plan = "tiny-unencrypted-11"
+paas_redis_service_plan    = "tiny-5_x"
+
+paas_app_config = {
+  RAILS_ENV                = "review"
+  RAILS_SERVE_STATIC_FILES = true
+}
+
+# StatusCake
+statuscake_username = "Review"
```
saliceti, line 16: Can we add a default in variables.tf instead?

saliceti (approved):

PR #1714, opened 2021-01-06
[2829] - Add courses count to Public API

TobyRet:
### Context

- We accidentally added the Provider courses count from the v2 API in https://github.com/DFE-Digital/teacher-training-api/pull/1713 (soz)
- What we actually want to do is add the _courses_ count to the meta object when the Public API receives course search queries (which is currently the behaviour in v3)

- Remove Provider courses count from the Public API
- Return the (search) courses count in the meta object from the Courses Controller. This matches the behaviour in the v3 API.

### Trello
https://trello.com/c/e5NdFola/2829-dev-expose-required-attributes-on-the-v1-public-api

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/courses_controller.rb
+++ b/app/controllers/api/public/v1/courses_controller.rb
@@ -5,6 +5,7 @@ module API
         def index
           render jsonapi: paginate(courses),
                  include: include_param,
+                 meta: { count: courses.count("course.id") },
                  class: API::Public::V1::SerializerService.call
         end
 
--- a/spec/controllers/api/public/v1/courses_controller_spec.rb
+++ b/spec/controllers/api/public/v1/courses_controller_spec.rb
@@ -202,6 +202,19 @@ RSpec.describe API::Public::V1::CoursesController do
           )
         end
       end
+
+      context "courses count" do
+        it "returns the course count in a meta object" do
+          get :index, params: {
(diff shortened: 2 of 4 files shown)


spec/controllers/api/public/v1/courses_controller_spec.rb
```diff
@@ -202,6 +202,19 @@
           )
         end
       end
+
+      context "courses count" do
```
TobyRet, line 206: @gpeng - noticed that there are no request specs, only controller specs. I've followed convention in this case but wondering this and some of the other specs in this file should be moved to a request spec?

gpeng: Yes. Not sure what happened there. We do 'do' request specs generally. I'll get them moved if we ever have a lull/firebreak. This is good for now though.

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -5,6 +5,7 @@ class CoursesController < API::Public::V1::ApplicationController
         def index
           render jsonapi: paginate(courses),
                  include: include_param,
+                  meta: { count: courses.count("course.id") },
```
gpeng, line 8: I think the indenting is off 1 here.

gpeng (approved):

PR #1716, opened 2021-01-08
[755] Clean up include examples for public api end points

darokel:
### Context

- https://trello.com/c/Hpj3nXtX/755-m-refactor-include-schema-definition-setup

- Refactors some of the documented public endpoint examples to correctly show their relationships/include options
- Previously, any endpoint with an optional include query parameter would return a random `Resource` defined in the swagger template, this change makes it more explicit what to render in that file.

### Guidance to review

- Fire up the API docs locally
- Assert any endpoint with optional includes shows the correct structure in the response example

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/swagger/public_v1/component_schemas/CourseRelationships.yml
+++ b/swagger/public_v1/component_schemas/CourseRelationships.yml
@@ -2,15 +2,9 @@
 description: "This schema is used to describe associations that can be returned with a course."
 type: object
 properties:
+  recruitment_cycle:
+    $ref: "#/components/schemas/Relationship"
   accredited_body:
     $ref: "#/components/schemas/Relationship"
-  locations:
-    $ref: "#/components/schemas/RelationshipList"
-  modern_languages:
-    $ref: "#/components/schemas/RelationshipList"
-  provider_locations:
-    $ref: "#/components/schemas/RelationshipList"
   provider:
     $ref: "#/components/schemas/Relationship"
-  subjects:
-    $ref: "#/components/schemas/RelationshipList"
--- a/spec/docs/provider_suggestions_spec.rb
+++ b/spec/docs/provider_suggestions_spec.rb
@@ -23,12 +23,6 @@ describe "API" do
 
         run_test!
       end
-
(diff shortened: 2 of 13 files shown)


darokel:
[[M] Refactor include schema definition setup](https://trello.com/c/Hpj3nXtX/755-m-refactor-include-schema-definition-setup)

gpeng (approved):
This looks good to me. Can we get @defong to review before we merge though just to be on the safe side.

defong (changes requested):
The structure looks rightish... but they also looks 

- bit part

image

- half filled 

image



Missing example for bad request?
image

For this endpoint `/recruitment_cycles/{year}/providers/{provider_code}/courses `
the parameters does not match the examples

- parameters does not have `include`
- example has `included`

darokel:
@defong Pushed up the changes https://github.com/DFE-Digital/teacher-training-api/pull/1716. I removed the 400 as at the moment it does not return a response. We have a ticket somewhere to introduce a generic error handling method which may make sense to use at that point and update this but out of scope for this ticket.

swagger/public_v1/component_schemas/CourseRelationships.yml
```diff
@@ -2,15 +2,9 @@
...
-    $ref: "#/components/schemas/RelationshipList"
-  modern_languages:
-    $ref: "#/components/schemas/RelationshipList"
-  provider_locations:
-    $ref: "#/components/schemas/RelationshipList"
   provider:
-    $ref: "#/components/schemas/Relationship"
-  subjects:
-    $ref: "#/components/schemas/RelationshipList"
+    $ref: "#/components/schemas/Relationship"
```
defong, line 10: new line

defong (approved):
> This looks good to me. Can we get @defong to review before we merge though just to be on the safe side.

toothcombed

spinoffs https://trello.com/c/x28JcW5k/874-api-docs-response-examples-are-wrong @gpeng

PR #1718, opened 2021-01-10
[2842] - Add search for Provider functionality to Public API 

TobyRet:
### Context

- It's not possible to search for Providers - this breaks the 'search by provider' functionality on Find Teacher Training.

- Implement missing 'search by provider' functionality

### Trello
https://trello.com/c/ftLKFssd/2842-add-missing-serializer-relationship-definition-to-public-api

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/providers_controller.rb
+++ b/app/controllers/api/public/v1/providers_controller.rb
@@ -3,6 +3,8 @@ module API
     module V1
       class ProvidersController < API::Public::V1::ApplicationController
         def index
+          return render(status: :bad_request) if invalid_search?
+
           render jsonapi: paginate(providers),
           include: params[:include], class: API::Public::V1::SerializerService.call, fields: fields
         end
@@ -23,7 +25,9 @@ module API
       private
 
         def providers
-          @providers ||= if sort_by_provider_ascending?
+          @providers ||= if search_for_provider?
+                           recruitment_cycle.providers.search_by_code_or_name(params[:search])
+                         elsif sort_by_provider_ascending?
                            recruitment_cycle.providers.by_name_ascending
                          else
                            recruitment_cycle.providers.by_name_descending
@@ -50,6 +54,14 @@ module API
 
         def sort_field
           @sort_field ||= Set.new(params.dig(:sort)&.split(","))
+        end
+
+        def search_for_provider?
+          params[:search].present?
+        end
+
+        def invalid_search?
+          search_for_provider? && params[:search].length < 2
         end
 
         def provider_fields
--- a/spec/requests/api/public/v1/providers/index_spec.rb
+++ b/spec/requests/api/public/v1/providers/index_spec.rb
@@ -1,0 +1,122 @@
+require "rails_helper"
+
+describe "GET public/v1/recruitment_cycle/:recruitment_cycle_year/providers" do
+  let(:organisation) { create(:organisation) }
+  let(:recruitment_cycle) { find_or_create :recruitment_cycle }
+
+  let(:provider) {
+    create(:provider,
+           provider_code: "1AT",
+           provider_name: "First provider",
+           organisations: [organisation],
+           contacts: [contact])
+  }
+
+  let(:contact) { build(:contact) }
+
+  let(:json_response) { JSON.parse(response.body) }
+
+  def perform_request
+    get request_path
+  end
+
+  subject do
+    perform_request
+
+    response
+  end
+
+  context "Searching for a provider" do
+    let(:base_provider_path) { "/api/public/v1/recruitment_cycles/#{recruitment_cycle.year}/providers" }
+    let(:provider_two) do
+      create(:provider,
+             provider_code: "2AT",
+             provider_name: "Second provider",
+             organisations: [organisation],
+             contacts: [contact])
+    end
+
+    before do
+      provider
+      provider_two
+    end
+
+    context "Searching for a provider by its full name" do
+      let(:request_path) { "#{base_provider_path}?search=Second provider" }
+
+      it "Only returns data for the provider" do
+        perform_request
+
+        expect(json_response["data"].count).to eq(1)
+        expect(json_response["data"].first).to have_attribute("code").with_value("2AT")
+      end
+    end
+
+    context "Searching for a provider by its lower case full name" do
+      let(:request_path) { "#{base_provider_path}?search=second provider" }
+
+      it "Only returns data for the provider" do
+        perform_request
+
+        expect(json_response["data"].count).to eq(1)
+        expect(json_response["data"].first).to have_attribute("code").with_value("2AT")
+      end
+    end
+
+    context "Searching for a provider by part of its name" do
+      let(:request_path) { "#{base_provider_path}?search=provider" }
+
+      it "Returns data for the matching providers" do
+        perform_request
+
+        expect(json_response["data"].count).to eq(2)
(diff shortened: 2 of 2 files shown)


app/serializers/api/public/v1/serializable_location_status.rb
```diff
@@ -9,6 +9,12 @@ class SerializableLocationStatus < JSONAPI::Serializable::Resource
         attributes :publish,
                    :status
 
+        belongs_to :location do
+          data do
+            @object.site
+          end
+        end
+
```
defong, line 17: Documentation needs updating

app/serializers/api/public/v1/serializable_course.rb
```diff
@@ -16,6 +16,12 @@ def enrichment_attribute(name, enrichment_name = name)
           data { @object.accrediting_provider }
         end
 
+        has_many :location_statuses do
+          data do
+            @object.site_statuses
+          end
+        end
+
```
defong, line 24: Documentation needs updating

ie 
image

defong (changes requested):
Documentation needs updating


> Currently it is not possible to retrieve locations (known as Sites) from the Public API via a Course
Does this not serve it purpose?

TobyRet:
> https://api.publish-teacher-training-courses.service.gov.uk/api-reference.html#recruitment_cycles-year-providers-provider_code-courses-course_code-locations-get

@defong 
I guess so but seems a little efficient as we would need to retrieve the Provider first, and filter the courses again.

This is where the API call is happening in Find. Some filtering of SiteStatuses required also - https://github.com/DFE-Digital/find-teacher-training/blob/master/app/view_objects/results_view.rb#L283

Happy to take your steer on this though 👍

app/serializers/api/public/v1/serializable_course.rb
TobyRet: You mean this?
https://github.com/DFE-Digital/teacher-training-api/pull/1718/files#diff-4427947212e294ec5b97c1345d508c970654f96036e1f9e7bf130ccb460843ebR48

defong:
> > https://api.publish-teacher-training-courses.service.gov.uk/api-reference.html#recruitment_cycles-year-providers-provider_code-courses-course_code-locations-get
> 
> @defong
> I guess so but seems a little efficient as we would need to retrieve the Provider first, and filter the courses again.
> 
> This is where the API call is happening in Find. Some filtering of SiteStatuses required also - https://github.com/DFE-Digital/find-teacher-training/blob/master/app/view_objects/results_view.rb#L283
> 
> Happy to take your steer on this though +1

@misaka @gpeng to steer this forward, as this affects the overall architectural design profoundly

app/serializers/api/public/v1/serializable_course.rb
defong: I think also this files needs to be changed

https://github.com/DFE-Digital/teacher-training-api/blob/master/swagger/public_v1/api_spec.json

misaka:
> > [api.publish-teacher-training-courses.service.gov.uk/api-reference.html#recruitment_cycles-year-providers-provider_code-courses-course_code-locations-get](https://api.publish-teacher-training-courses.service.gov.uk/api-reference.html#recruitment_cycles-year-providers-provider_code-courses-course_code-locations-get)
> 
> @defong
> I guess so but seems a little efficient as we would need to retrieve the Provider first, and filter the courses again.

Inefficient I guess you mean? Ya, but it's not that bad really since the request will be filtered. As discussed in Slack, this is our attempt at avoiding accidental N+1s which we were getting regularly with API v3 ... we could've spent more time to sort out all the includes to avoid them, but that was taking a lot of time and it seemed more efficient to change the API shape to stop them. Also, this rebalances the API a little allowing for more requests to come in, which means they can be balanced across servers better ... big long requests (think about serialisation, etc) make it harder to spread the load.

 
> This is where the API call is happening in Find. Some filtering of SiteStatuses required also - [https://github.com/DFE-Digital/find-teacher-training/blob/master/app/view_objects/results_view.rb#L283](https://github.com/DFE-Digital/find-teacher-training/blob/master/app/view_objects/results_view.rb?rgh-link-date=2021-01-11T09%3A39%3A36Z#L283)

If you're having to do filtering on the client side then we should consider, where it makes sense, adding server-side filtering. Is that the case?

app/serializers/api/public/v1/serializable_location_status.rb
TobyRet: I am now using another end point to fetch locations - changes no longer required

app/serializers/api/public/v1/serializable_course.rb
TobyRet: I am now using another end point to fetch locations - changes no longer required

spec/requests/api/public/v1/providers/index_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    context "Searching for a provider with an invalid query" do
+      context "query is empty" do
+        let(:request_path) { "#{base_provider_path}?search=" }
+
+        it "returns all providers" do
+          perform_request
+
+          expect(json_response["data"].count).to eq(2)
+        end
```
defong, line 108: Feels odd, that and invalid will return a set of result

app/controllers/api/public/v1/providers_controller.rb
```diff
@@ -23,7 +25,9 @@ def show
       private
 
         def providers
-          @providers ||= if sort_by_provider_ascending?
+          @providers ||= if search_for_provider?
+                           recruitment_cycle.providers.search_by_code_or_name(params[:search])
+                         elsif sort_by_provider_ascending?
```
defong, line 30: Are we sure that we want this endpoint to be mutually exclusive? @gpeng

app/controllers/api/public/v1/providers_controller.rb
```diff
@@ -52,6 +56,14 @@ def sort_field
           @sort_field ||= Set.new(params.dig(:sort)&.split(","))
         end
 
+        def search_for_provider?
+          params[:search].present?
```
defong, line 60: requires documentation up 

image

spec/requests/api/public/v1/providers/index_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+    get request_path
+  end
+
+  subject do
+    perform_request
+
+    response
+  end
+
+  context "Searching for a provider" do
```
defong, line 29: https://api.qa.publish-teacher-training-courses.service.gov.uk/api-reference.html#provider_suggestions-get

What's the overriding reason over using the other endpoint?

defong (changes requested):
> It's not possible to search for Providers - this breaks the 'search by provider' functionality on Find Teacher Training.

Does this serves it purposes?

https://api.qa.publish-teacher-training-courses.service.gov.uk/api-reference.html#provider_suggestions-get


These **public api** changes is starting to feel like we need to put our heads together to ensue that they are conforming to singular design purpose @gpeng

spec/requests/api/public/v1/providers/index_spec.rb
TobyRet: Actually that is true. I'll use that instead and close this PR

TobyRet:
Closing PR as changes no longer required

PR #1723, opened 2021-01-11
Add additional missing attributes to the teacher training public API

JR-G:
### Context

There are some further attributes that need to be added to the Public API

Add `:description`, `:provider_code` and `:provider_type`   to SerializableCourse

### Trello

https://trello.com/c/wsR2MIzj/2841-add-more-missing-attributes-to-teacher-training-public-api

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/serializers/api/public/v1/serializable_course.rb
+++ b/app/serializers/api/public/v1/serializable_course.rb
@@ -33,7 +33,8 @@ module API
                    :qualifications,
                    :scholarship_amount,
                    :study_mode,
-                   :uuid
+                   :uuid,
+                   :description
 
         attribute :about_accredited_body do
           @object.accrediting_provider_description
--- a/swagger/public_v1/api_spec.json
+++ b/swagger/public_v1/api_spec.json
@@ -394,6 +394,11 @@
             "type": "string",
             "description": "Generated summary of the course.",
             "example": "PGCE with QTS full time"
+          },
+          "description": {
+            "type": "string",
+            "description": "A combined string of qualification, study mode & programme type.",
+            "example": "PGCE with QTS full time teaching apprenticeship."
           }
         }
       },
--- a/spec/serializers/api/public/v1/serializable_course_spec.rb
+++ b/spec/serializers/api/public/v1/serializable_course_spec.rb
@@ -72,6 +72,7 @@ RSpec.describe API::Public::V1::SerializableCourse do
   it { is_expected.to have_attribute(:running).with_value(course.findable?) }
   it { is_expected.to have_attribute(:salary_details).with_value(course.latest_published_enrichment.salary_details) }
   it { is_expected.to have_attribute(:scholarship_amount).with_value(nil) }
+  it { is_expected.to have_attribute(:description).with_value(course.description) }
 
   context "when bursary amount is present" do
     let(:course) { create(:course, :with_accrediting_provider, :secondary, enrichments: [enrichment], subjects: [find_or_create(:secondary_subject, :classics)]) }


gpeng (approved):

app/serializers/api/public/v1/serializable_course.rb
```diff
@@ -111,6 +112,14 @@ def enrichment_attribute(name, enrichment_name = name)
           @object.description
         end
 
+        attribute :provider_code do
+          @object.provider.provider_code
+        end
+
+        attribute :provider_type do
+          @object.provider.provider_type
+        end
```
defong, line 121: No this cannot be here, we need to conform with https://jsonapi.org/format/#fetching-sparse-fieldsets to properly utilise it, 

The controller should be updated to something like: 

```ruby

def index
          render jsonapi: paginate(courses),
                 include: include_param,
                 meta: { count: courses.count("course.id") },
                 class: API::Public::V1::SerializerService.call,
                 fields: fields
        end
        def fields
          { providers: provider_fields } if provider_fields.present?
        end

        def provider_fields
          params.dig(:fields, :providers)&.split(",")
        end

```
So that you can use a URL like 
http://localhost:3001/api/public/v1/recruitment_cycles/2021/courses?&fields[providers]=provider_type,code&include=provider

See 

image

swagger/public_v1/api_spec.json
```diff
@@ -394,6 +394,30 @@
...
+            "type": "string",
+            "description": "Code that uniquely identifies this provider.",
+            "maxLength": 3,
+            "minLength": 3,
+            "example": "X99"
+          },
+          "description": {
+            "type": "string",
+            "description": "A combined string of qualification, study mode & programme type.",
+            "example": "PGCE with QTS full time teaching apprenticeship."
```
defong, line 420: @misaka should this be here? as it is a calculated value and should be off loaded to the client side

defong (changes requested):

app/serializers/api/public/v1/serializable_course.rb
TobyRet: @defong - I can understand the rationale for not exposing provider attributes on the Course model (we were simply replicating v3 behaviour). I'm not clear though whether introducing sparse fields for an associated object is more beneficial than calling `course.provider.code?` ... am I missing something?

swagger/public_v1/api_spec.json
misaka: Ya, I remember we had a discussion about this, and I think we wanted to have a standard way of summarising the course. We even suggested that maybe we want to call this `summary`, which might make more sense than `description`.

misaka: @defong Do we currently produce this in API v3?

app/serializers/api/public/v1/serializable_course.rb
TobyRet: We'll remove the `provider_code` and `provider_type` attributes

swagger/public_v1/api_spec.json
defong: It's rooted here https://github.com/DFE-Digital/teacher-training-api/blob/1cff52b7d5283c2fe2d9ba4f9094ebd0524c4a54/app/models/course.rb#L362-L366

TobyRet: @misaka @defong - have you come to a decision on the above?

defong: > Ya, I remember we had a discussion about this, and I think we wanted to have a standard way of summarising the course. We even suggested that maybe we want to call this `summary`, which might make more sense than `description`.

https://github.com/DFE-Digital/teacher-training-api/blob/f79edc40b78c8d844906f4e3239d315310649fd5/swagger/public_v1/api_spec.json#L393-L397

https://github.com/DFE-Digital/teacher-training-api/blob/2d9c96133190f82bce664de1a459b3f2a9c6976e/app/serializers/api/public/v1/serializable_course.rb#L110-L112

:facepalm:

defong (changes requested):
@misaka is forcing my :hand: :face_with_head_bandage:

TobyRet:
The summary attribute is already present so no need for this PR

PR #1725, opened 2021-01-12
[712] V1 public API smoke tests

jhenderson:
### Context
https://trello.com/c/uqDY2T7t/712-m-providers-endpoint-smoke-tests

- Modify smoke test Github action to dynamically update `Gemfile` so it can use `rspec` and `httparty` in any environment
- New `spec_helper_smoke.rb`
- Smoke test to check V1 providers endpoint returns HTTP status of 200

### Guidance to review
- Copy and paste the following into the existing `Gemfile` and run `bundle`:
```
gem "httparty"
gem "rspec"
```
- run `bundle exec rspec spec/spoke` - the test should pass
- run `RAILS_ENV=qa bundle exec rspec spec/spoke` - the test should pass
- run `RAILS_ENV=staging bundle exec rspec spec/spoke` - the test should pass
- run `RAILS_ENV=production bundle exec rspec spec/spoke` - the test should pass

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/smoke/api/v1/providers_spec.rb
+++ b/spec/smoke/api/v1/providers_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "spec_helper_smoke"
+
+describe "V1 Public API Smoke Tests", :aggregate_failures, smoke: true do
+  let(:recruitment_year) { Settings.current_recruitment_cycle_year }
+  let(:base_url) { Settings.publish_url.sub("www", "api") }
+
+  subject(:response) { HTTParty.get(url) }
+
+  context "providers" do
+    describe "GET /v1/recruitment_cycles/:recruitment_year/providers" do
+      let(:url) { "#{base_url}/api/public/v1/recruitment_cycles/#{recruitment_year}/providers?[per_page]=1" }
+
+      it "returns a HTTP status code of 200" do
+        expect(response.code).to eq(200)
+      end
+
+      it "returns at least one record" do
+        expect(response.parsed_response["data"].length).to be_positive
+      end
+    end
+  end
+end
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -11,15 +11,29 @@ jobs:
     name: smoke-tests-${{ github.event.inputs.environment }}
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Ruby 2.7.2
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: 2.7.2
+
+      - name: install bundler and gems
+        run: |
+          gem install bundler
+          echo 'gem "httparty"' >> Gemfile
+          echo 'gem "rspec"' >> Gemfile
+          bundle
+
       - uses: softprops/turnstyle@v1
         name: Wait for other runs
         env:
           GITHUB_TOKEN: ${{ secrets.ACTIONS_API_ACCESS_TOKEN }}
 
       - name: Smoke Tests ${{ github.event.inputs.environment }}
-        run: echo "Nothing to run yet 🚀"
+        run: RAILS_ENV=${{ github.event.inputs.environment }} ./bin/bundle exec rspec spec/smoke --format documentation
 
-      - name: 'Nofiy #twd_publish_register_tech on failure'
+      - name: 'Notify #twd_publish_register_tech on failure'
         if: failure()
         uses: rtCamp/action-slack-notify@master
         env:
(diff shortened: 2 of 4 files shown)


jhenderson:
[[M] Providers endpoint smoke tests](https://trello.com/c/uqDY2T7t/712-m-providers-endpoint-smoke-tests)

gpeng (approved):
This looks good if we've got the same approach working on register too. I still wonder if it would make it simpler to move the smoke tests into their own directory in the root. That would mean (I think, haven't tried it) that it could just have standard spec_helper and Gemfile in there and wouldn't need to worry about smoke tests in the normal app. Let's go with this for now anyway if you've got it working 👍

spec/smoke/api/v1/providers_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require "spec_helper_smoke"
+
+describe "V1 Public API Smoke Tests", :aggregate_failures, smoke: true do
+  let(:course_year) { Time.current.year }
```
defong, line 6:
```suggestion
  let(:recruitment_year) { Settings.current_recruitment_cycle_year }
```

The recruitment cycle year value changes around October time.

spec/smoke/api/v1/providers_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let(:base_url) { Settings.publish_url.sub("www", "api") }
+
+  subject(:response) { HTTParty.get(url) }
+
+  context "providers" do
+    describe "GET /v1/recruitment_cycles/:recruitment_year/providers" do
+      let(:url) { "#{base_url}/api/public/v1/recruitment_cycles/#{course_year}/providers" }
```
defong, line 13:
```suggestion
      let(:url) { "#{base_url}/api/public/v1/recruitment_cycles/#{recruitment_year}/providers?page[per_page]=1" }
```

There is no such thing as `course_year`.
It probably going to be better to return 1 item, rather then the default 500.

spec/smoke/api/v1/providers_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      it "returns a HTTP status code of 200" do
```
defong, line 15: It should also check at least 1 item is returned in data

defong (changes requested):
Looks good but the year must be anchored down.

jhenderson:
> This looks good if we've got the same approach working on register too. I still wonder if it would make it simpler to move the smoke tests into their own directory in the root. That would mean (I think, haven't tried it) that it could just have standard spec_helper and Gemfile in there and wouldn't need to worry about smoke tests in the normal app. Let's go with this for now anyway if you've got it working 👍

@gpeng Not a bad idea, but would prefer to let this go in for now and consider that later on.

spec/smoke/api/v1/providers_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    describe "GET /v1/recruitment_cycles/:recruitment_year/providers" do
+      let(:url) { "#{base_url}/api/public/v1/recruitment_cycles/#{recruitment_year}/providers?[per_page]=1" }
+
+      it "returns a HTTP status code of 200" do
+        expect(response.code).to eq(200)
+      end
+
+      it "returns a payload with data" do
+        expect(payload.size).to be > 0
+      end
```
defong, line 23: @AbigailMcP Similar kinda question..


https://github.com/DFE-Digital/teacher-training-api/blob/f94a639c47afe1fd66c2495f0fd16d60c3fe3b25/spec/smoke/api/v1/courses_spec.rb#L18-L20

spec/smoke/api/v1/providers_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+describe "V1 Public API Smoke Tests", :aggregate_failures, smoke: true do
+  let(:recruitment_year) { Settings.current_recruitment_cycle_year }
+  let(:base_url) { Settings.publish_url.sub("www", "api") }
+
+  subject(:response) { HTTParty.get(url) }
+
+  let(:payload) { JSON(response.body).fetch("data", []) }
+
+  context "providers" do
```
defong, line 13: @AbigailMcP Should your one has context as well?

defong (approved):
:shipit:

defong (approved):
:shipit:

spec/smoke/api/v1/providers_spec.rb
jhenderson: Happy to go along with this. I did try `expect(payload).to have_at_least(1).provider` which is valid but needs another rspec matcher library.

spec/smoke/api/v1/providers_spec.rb
AbigailMcP: I'll add!

PR #1749, opened 2021-01-29
Use separate publish_api_url setting

vigneshmsft:
### Context
Api URL is inferred by replacing `www.` with `api.` in `publish_url` setting.
This will not apply to all environments in PaaS as, qa,staging and sandbox don't have a `www` prefix in PaaS.
This is also causing smoke-tests to fail in sandbox env.

Use a separate variable with correct value for `publish_api_url`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


misaka (approved):
👍

PR #1775, opened 2021-02-12
[99] Add provider creation task

gpeng:
### Context

We need to onboard users and providers to the Publish/TTAPI sandbox environment.

Add task that:

* Accepts a CSV
* Creates an organisation
* Creates a provider
* Adds a location to the provider

This task is procedural stylee for now. We are currently discussing building some UI on the API for this sort of support work. This will serve as 'documentation' for creating a better structured service class in the future.

The data set on the provider and associated entities is minimal placeholder data. Sandbox users should probably be directed to update things like the address and location etc otherwise there will be a lot of courses at the same location in Find.

### Guidance to review

The tasks will run locally. Create appropriate CSV files as per the task description.

`bundle exec rake sandbox:create_providers\[./providers.csv\]`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/providers/create_fake_provider_service.rb
+++ b/app/services/providers/create_fake_provider_service.rb
@@ -2,6 +2,14 @@ module Providers
   class CreateFakeProviderService
     attr_reader :errors
 
+    DEFAULT_PROVIDER_ATTRIBUTES = {
+        address1: "1 Test Street",
+        address3: "Town",
+        address4: "County",
+        postcode: "M1 1JG",
+        region_code: "north_west",
+    }.freeze
+
     def initialize(recruitment_cycle:, provider_name:, provider_code:, provider_type:, is_accredited_body:)
       raise "Can only be run in sandbox or development" unless Rails.env.sandbox? || Rails.env.development? || Rails.env.test?
 
@@ -15,22 +23,14 @@ module Providers
     end
 
     def execute
-      if provider_already_exists?
-        errors << "Provider #{@provider_name} (#{@provider_code}) already exists."
-        return false
-      end
+      return false if provider_already_exists?
 
-      provider = @recruitment_cycle.providers.build(
+      provider = @recruitment_cycle.providers.build({
         provider_name: @provider_name,
         provider_code: @provider_code,
         provider_type: @provider_type,
         accrediting_provider: @is_accredited_body ? "accredited_body" : "not_an_accredited_body",
-        address1: "1 Test Street",
-        address3: "Town",
-        address4: "County",
-        postcode: "M1 1JG",
-        region_code: "north_west",
-      )
+      }.merge(DEFAULT_PROVIDER_ATTRIBUTES))
 
       organisation = Organisation.new(name: @provider_name)
       organisation.providers << provider
@@ -50,18 +50,20 @@ module Providers
         errors << "Unable to create Provider #{provider_name}: #{provider.errors.to_sentence}."
       end
 
-      if errors.any?
-        false
-      else
-        true
-      end
+      errors.empty?
     end
 
   private
 
     def provider_already_exists?
-      @recruitment_cycle.providers.exists?(provider_name: @provider_name) || \
-        @recruitment_cycle.providers.exists?(provider_code: @provider_code)
+      if @recruitment_cycle.providers.exists?(provider_name: @provider_name) || \
+          @recruitment_cycle.providers.exists?(provider_code: @provider_code)
+        errors << "Provider #{@provider_name} (#{@provider_code}) already exists."
+
+        true
+      else
+        false
+      end
     end
   end
 end
--- a/app/services/providers/create_fake_provider_service.rb
+++ b/app/services/providers/create_fake_provider_service.rb
@@ -47,11 +47,7 @@ module Providers
         region_code: provider.region_code,
       )
 
-      if provider.invalid?
-        errors << "Unable to create Provider #{provider_name}: #{provider.errors.to_sentence}."
-      end
-
-      errors.empty?
+      true
     end
 
   private
--- a/app/services/providers/create_fake_provider_service.rb
+++ b/app/services/providers/create_fake_provider_service.rb
@@ -24,6 +24,7 @@ module Providers
 
     def execute
       return false if provider_already_exists?
+      return false if attempting_to_make_lead_school_accredited_body?
 
       provider = @recruitment_cycle.providers.build({
         provider_name: @provider_name,
@@ -63,6 +64,12 @@ module Providers
         true
       else
         false
+      end
+    end
+
+    def attempting_to_make_lead_school_accredited_body?
+      if @provider_type == "lead_school" && @is_accredited_body
+        errors << "Provider #{@provider_name} (#{@provider_code}) cannot be both a lead school and an accredited body."
       end
     end
   end
--- a/app/services/providers/create_fake_provider_service.rb
+++ b/app/services/providers/create_fake_provider_service.rb
@@ -3,6 +3,8 @@ module Providers
     attr_reader :errors
 
     def initialize(recruitment_cycle:, provider_name:, provider_code:, provider_type:, is_accredited_body:)
+      raise "Can only be run in sandbox or development" unless Rails.env.sandbox? || Rails.env.development? || Rails.env.test?
+
       @recruitment_cycle = recruitment_cycle
       @provider_name = provider_name
       @provider_code = provider_code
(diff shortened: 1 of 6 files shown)


duncanjbrown:
Hey @gpeng! This makes sense to me. By way of learning a bit about TTAPI I did some refactoring here — wdyt? Could follow up with the same treatment for the Users import.

duncanjbrown:
For realistic data, should we create a handful of locations along with the provider?

gpeng:
[Sandbox provider rake task](https://trello.com/c/XpRZMAeh/1062-sandbox-provider-rake-task)

gpeng:
> Hey @gpeng! This makes sense to me. By way of learning a bit about TTAPI I did some refactoring here — wdyt? Could follow up with the same treatment for the Users import.

Thanks @duncanjbrown Looks fine. I was holding off splitting it into services for now. If we're going to add a UI for onboarding all users/providers we'll need a non-fake version of the service and we can probably use that for both types. We can extract that later though so happy to go with this approach if you want to start doing that now. The tests are failing on the `env` check though.

gpeng:
> For realistic data, should we create a handful of locations along with the provider?

I was trying to create the minimum possible that would all the user to be able to log in and edit the provider. We need at least one as the UI doesn't handle there being none very well. Another issue with this is the address. If we add them all with the same address then sandbox find is going to end up with a load of courses all at whatever address we choose. Requiring the sandbox users to add them manually might avoid this a bit 😬 I was going to use Faker but that would cause issues in this regard too. Would be good to work out a solution of this.

In the meantime though, if you feel it would be more useful for sandbox users to have a few pre-filled then yes, let's add them.

duncanjbrown:
> > Hey @gpeng! This makes sense to me. By way of learning a bit about TTAPI I did some refactoring here — wdyt? Could follow up with the same treatment for the Users import.
> 
> Thanks @duncanjbrown Looks fine. I was holding off splitting it into services for now. If we're going to add a UI for onboarding all users/providers we'll need a non-fake version of the service and we can probably use that for both types. We can extract that later though so happy to go with this approach if you want to start doing that now. The tests are failing on the `env` check though.

Awesome, thank you — it was a useful learning experience in any case! I'll fix the specs.

duncanjbrown:
> > For realistic data, should we create a handful of locations along with the provider?
> 
> I was trying to create the minimum possible that would all the user to be able to log in and edit the provider. We need at least one as the UI doesn't handle there being none very well. Another issue with this is the address. If we add them all with the same address then sandbox find is going to end up with a load of courses all at whatever address we choose. Requiring the sandbox users to add them manually might avoid this a bit 😬 I was going to use Faker but that would cause issues in this regard too. Would be good to work out a solution of this.
> 
> In the meantime though, if you feel it would be more useful for sandbox users to have a few pre-filled then yes, let's add them.

Nothing to hold this up — let's merge as-is and get feedback?

defong (changes requested):
image

One stone two tech leads... rejected

duncanjbrown:
My first @defong rejection 😊

spec/services/providers/create_fake_provider_service_spec.rb
```diff
@@ -0,0 +1,69 @@
+require "rails_helper"
+
+RSpec.describe Providers::CreateFakeProviderService do
+  let(:provider_name) { "Fake Provider" }
+  let(:provider_code) { "ABC" }
+  let(:provider_type) { "scitt" }
+  let(:recruitment_cycle) { RecruitmentCycle.current.presence || create(:recruitment_cycle) }
```
defong, line 7:
```suggestion
  let(:recruitment_cycle) { find_or_create :recruitment_cycle }
```

Maybe?

app/services/providers/create_fake_provider_service.rb
```diff
@@ -0,0 +1,69 @@
...
+        address2: provider.address2,
+        address3: provider.address3,
+        address4: provider.address4,
+        postcode: provider.postcode,
+        region_code: provider.region_code,
+      )
+
+      if provider.invalid?
+        errors << "Unable to create Provider #{provider_name}: #{provider.errors.to_sentence}."
+      end
```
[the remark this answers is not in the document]
duncanjbrown: Good point

app/services/providers/create_fake_provider_service.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      errors.empty?
+    end
+
+  private
+
+    def provider_already_exists?
+      if @recruitment_cycle.providers.exists?(provider_name: @provider_name) || \
+          @recruitment_cycle.providers.exists?(provider_code: @provider_code)
```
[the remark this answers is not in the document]
duncanjbrown: Point proven — updating!

app/services/providers/create_fake_provider_service.rb
```diff
@@ -0,0 +1,69 @@
...
+        address1: provider.address1,
+        address2: provider.address2,
+        address3: provider.address3,
+        address4: provider.address4,
+        postcode: provider.postcode,
+        region_code: provider.region_code,
+      )
+
+      if provider.invalid?
+        errors << "Unable to create Provider #{provider_name}: #{provider.errors.to_sentence}."
```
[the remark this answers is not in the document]
duncanjbrown: It's not actually possible to make an invalid provider with this interface, even if you knock out the name and the code, so I've removed this check

app/services/providers/create_fake_provider_service.rb
```diff
@@ -0,0 +1,69 @@
...
+    }.freeze
+
+    def initialize(recruitment_cycle:, provider_name:, provider_code:, provider_type:, is_accredited_body:)
+      raise "Can only be run in sandbox or development" unless Rails.env.sandbox? || Rails.env.development? || Rails.env.test?
+
+      @recruitment_cycle = recruitment_cycle
+      @provider_name = provider_name
+      @provider_code = provider_code
+      @provider_type = provider_type
+      @is_accredited_body = is_accredited_body
```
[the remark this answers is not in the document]
duncanjbrown: Added

duncanjbrown:
Should be ready for another look!

app/services/providers/create_fake_provider_service.rb
```diff
@@ -0,0 +1,70 @@
...
+        location_name: "Site 1",
+        address1: provider.address1,
+        address2: provider.address2,
+        address3: provider.address3,
+        address4: provider.address4,
+        postcode: provider.postcode,
+        region_code: provider.region_code,
+      )
+
+      errors.empty?
```
defong, line 50: Does this actually get used?

given 
```ruby
      return false if provider_already_exists?
      return false if attempting_to_make_lead_school_accredited_body?
```

spec/services/csv_imports/fake_providers_import_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      import = described_class.new("csv_file.csv")
+      import.execute
+
+      expect(import.results).to eq(["Created provider Provider A."])
+    end
+  end
+
+  context "two providers to import" do
+    let(:csv_content) do
+      "name,code,type,accredited_body\n\"Provider A\",ABC,scitt,false\n\"Provider B\",DEF,lead_school,true"
```
defong, line 36:
```suggestion
      "name,code,type,accredited_body\n\"Provider A\",ABC,scitt,false\n\"Provider B\",DEF,lead_school,false"
```

app/services/providers/create_fake_provider_service.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  private
+
+    def provider_already_exists?
+      if @recruitment_cycle.providers.exists?(provider_code: @provider_code)
+        errors << "Provider #{@provider_name} (#{@provider_code}) already exists."
+        true
+      else
+        false
+      end
```
defong, line 61:
```suggestion
      end
```

defong (approved):

app/services/providers/create_fake_provider_service.rb
duncanjbrown: I think I prefer the v explicit version?

defong: https://github.com/DFE-Digital/teacher-training-api/blob/be0bba87a456bedc02f6ad21a835ef57133689de/app/services/providers/create_fake_provider_service.rb#L64-L68

it justslooks odd with this underneath

PR #1756, opened 2021-02-03
Add an MCB command to backfill Sites with the TTWA or London Borough

davidgisbey:
### Context

On Find we're reworking the design of the Course page. Instead of providing locations, we're telling candidates what areas providers operate in so that they do not get misled into thinking that they are going to be exclusively training in the school they apply too.

In order to get get the travel to work areas/boroughs for London, we need to use Mapit https://mapit.mysociety.org/

In my previous PR i added two new columns for travel_to_work_area and london_borough  #1755 

### Changes proposed in this pull request

- Adds an MCB command 
- Updates the TTWA and London Borough  when address changes
- Exposes the TTWA and London Borough in the SiteSerialiser

### Guidance to review

https://trello.com/c/RpLNRN3n/2932-%F0%9F%97%BA-dev-%F0%9F%8F%88-include-additional-metadata-in-provider-location-response

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #1777, opened 2021-02-15
Ensure Travel to Work Area and London Borough get updated on lat/long change

davidgisbey:
### Context

I've already run the MCB script from #1756 which updates travel_to_work_area and london_borough for each site in the current cycle.

This is a bit of clean up work to ensure that when a provider updates a sites location, we also update the travel to work area and London borough.

- Add service which hits Mapit and updates a sites TTWA and London Borough
- Add a callback to the model that is triggered when lat/long are updated
- Add a worker which is called calls the above service on lat/long update
 
### Guidance to review

Most of this has previously been reviewed as part of #1756. I've dropped the mcb command because we don't need a bulk update going forward and it tanked the test coverage.

Once this is merged I'll have a look at the audits for site updates since I ran the mcb command and update them where appropriate.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -31,6 +31,7 @@ class Site < ApplicationRecord
 
   attr_accessor :skip_geocoding
   after_commit :geocode_site, unless: :skip_geocoding
+  after_commit :add_travel_to_work_area_and_london_borough
 
   def geocode_site
     GeocodeJob.perform_later("Site", id) if needs_geolocation?
@@ -40,6 +41,12 @@ class Site < ApplicationRecord
     full_address.present? && (
     latitude.nil? || longitude.nil? || address_changed?
     )
+  end
+
+  def add_travel_to_work_area_and_london_borough
+    return false unless needs_travel_to_work_area_and_london_borough_updated?
+
+    TravelToWorkAreaAndLondonBoroughJob.perform_later(id)
   end
 
   def full_address
@@ -81,5 +88,14 @@ private
     available_desirable_codes = available_codes & DESIRABLE_CODES
     available_undesirable_codes = available_codes & EASILY_CONFUSED_CODES
     available_desirable_codes.sample || available_undesirable_codes.sample
+  end
+
+  def needs_travel_to_work_area_and_london_borough_updated?
+    latitude.present? &&
+      longitude.present? &&
+      (
+        saved_change_to_longitude? ||
+        saved_change_to_latitude?
+      )
   end
 end
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -31,7 +31,7 @@ class Site < ApplicationRecord
 
   attr_accessor :skip_geocoding
   after_commit :geocode_site, unless: :skip_geocoding
-  after_commit :add_travel_to_work_area_and_london_borough
+  after_commit :add_travel_to_work_area_and_london_borough, if: :needs_travel_to_work_area_and_london_borough_updated?
 
   def geocode_site
     GeocodeJob.perform_later("Site", id) if needs_geolocation?
@@ -44,8 +44,6 @@ class Site < ApplicationRecord
   end
 
   def add_travel_to_work_area_and_london_borough
-    return false unless needs_travel_to_work_area_and_london_borough_updated?
-
     TravelToWorkAreaAndLondonBoroughJob.perform_later(id)
   end
 
--- a/app/services/travel_to_work_area_and_london_borough_service.rb
+++ b/app/services/travel_to_work_area_and_london_borough_service.rb
@@ -11,7 +11,7 @@ class TravelToWorkAreaAndLondonBoroughService
                      end
 
     site.update_column("travel_to_work_area", ttw_area)
-    site.update_column("london_borough", london_borough)
+    site.update_column("london_borough", london_borough.gsub(/ Borough Council| City Council| Corporation/, ""))
     site.save!(validate: false)
   rescue StandardError
     false
--- a/app/services/travel_to_work_area_and_london_borough_service.rb
+++ b/app/services/travel_to_work_area_and_london_borough_service.rb
@@ -7,12 +7,13 @@ class TravelToWorkAreaAndLondonBoroughService
     ttw_area = json.dig(json.keys.first, "name")
 
     london_borough = if ttw_area == "London"
-                       json.dig(json.keys.second, "name")
+                       json
+                       .dig(json.keys.second, "name")
+                       .gsub(/ Borough Council| City Council| Corporation/, "")
                      end
 
     site.update_column("travel_to_work_area", ttw_area)
-    site.update_column("london_borough", london_borough.gsub(/ Borough Council| City Council| Corporation/, ""))
-    site.save!(validate: false)
+    site.update_column("london_borough", london_borough)
   rescue StandardError
     false
   end
--- a/app/services/travel_to_work_area_and_london_borough_service.rb
+++ b/app/services/travel_to_work_area_and_london_borough_service.rb
@@ -1,0 +1,19 @@
+class TravelToWorkAreaAndLondonBoroughService
+  def self.add_travel_to_work_area_and_london_borough(site:)
+    url = URI("#{Settings.mapit_url}/point/4326/#{site.longitude},#{site.latitude}?type=TTW,LBO&api_key=#{Settings.mapit_api_key}")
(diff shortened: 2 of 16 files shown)


app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    ttw_area = json.dig(json.keys.first, "name")
+
+    london_borough = if ttw_area == "London"
+                       json.dig(json.keys.second, "name")
+                     end
+
+    site.update_column("travel_to_work_area", ttw_area)
+    site.update_column("london_borough", london_borough.gsub(/ Borough Council| City Council| Corporation/, ""))
+    site.save!(validate: false)
```
defong, line 15:
```suggestion
    site.update_column({ travel_to_work_area: ttw_area, london_borough: london_borough)
    
```

Don't think you ned this now
`site.save!(validate: false)`
https://apidock.com/rails/ActiveRecord/Persistence/update_columns

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -0,0 +1,19 @@
+class TravelToWorkAreaAndLondonBoroughService
+  def self.add_travel_to_work_area_and_london_borough(site:)
+    url = URI("#{Settings.mapit_url}/point/4326/#{site.longitude},#{site.latitude}?type=TTW,LBO&api_key=#{Settings.mapit_api_key}")
+    response = Net::HTTP.get(url)
+    json = JSON.parse(response)
+
+    ttw_area = json.dig(json.keys.first, "name")
+
+    london_borough = if ttw_area == "London"
+                       json.dig(json.keys.second, "name")
```
defong, line 10:
```suggestion
                       json.dig(json.keys.second, "name")&.gsub(/ Borough Council| City Council| Corporation/, "")
```

spec/models/site_spec.rb
```diff
@@ -190,5 +190,57 @@
...
+        it { should be(false) }
+      end
+
+      context "longitude and latitude have been updated to nil" do
+        before do
+          site.update(latitude: nil, longitude: nil)
+        end
+
+        it { should be(false) }
+      end
```
defong, line 227: Maybe ...

```ruby

it "triggers add_travel_to_work_area_and_london_borough on after_commit" do
  expect(site).to not_receive(:add_travel_to_work_area_and_london_borough)
  site.save
end

```

config/settings/development.yml
```diff
@@ -5,3 +5,4 @@ mcbg:
 system_authentication_token: "Ge32"
 log_level: debug
 gcp_api_key: please_change_me
+mapit_url: https://mapit.mysociety.org
```
TobyRet, line 8: As the Mapit host is the same regardless of environment it can go in `config/settings.yml` and other environments will inherit this value.

Are we planning to have a Mapit secret per environment going forwards?

spec/services/travel_to_work_area_and_london_borough_service_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+        \"id\": 2504,
+        \"codes\": {\"unit_id\": \"11164\", \"ons\": \"00BK\", \"gss\": \"E09000033\", \"local-authority-eng\": \"WSM\", \"local-authority-canonical\": \"WSM\"},
+        \"name\": \"#{london_borough}\",
+        \"country\": \"E\",
+        \"type_name\": \"London borough\",
+        \"generation_low\": 1,
+        \"country_name\": \"England\",
+        \"type\": \"LBO\"
+        }
+      }"
```
defong, line 62: Can we try `{}.to_json`
so we lose some of those escapes...

app/models/site.rb
```diff
@@ -82,4 +89,13 @@ def pick_next_available_code(available_codes: [])
     available_undesirable_codes = available_codes & EASILY_CONFUSED_CODES
     available_desirable_codes.sample || available_undesirable_codes.sample
   end
+
+  def needs_travel_to_work_area_and_london_borough_updated?
+    latitude.present? &&
+      longitude.present? &&
+      (
+        saved_change_to_longitude? ||
```
TobyRet, line 97: Should this be `travel_to_work_area.present?`, `london_borough.present?`, `saved_change_to_travel_to_work_area?` and `saved_change_to_london_borough?`

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -0,0 +1,19 @@
...
+                     end
+
+    site.update_column("travel_to_work_area", ttw_area)
+    site.update_column("london_borough", london_borough.gsub(/ Borough Council| City Council| Corporation/, ""))
```
TobyRet, line 14: I see the 'Borough' string is being removed - do we need to backfill this change on prod?

TobyRet: Also this substitution does not work on "Royal London Borough of ..."

Might need to extract and unit test?

TobyRet (changes requested):
Looking good. There is an issue with the regex but other than that, some minor points

app/services/travel_to_work_area_and_london_borough_service.rb
davidgisbey: I had a quick test on Mapit and it doesn't actually return Royal London Borough for any of them (Kingston, Kensington etc.)

image

image

image

config/settings/development.yml
davidgisbey: No we're not actaully. Will make the changes.

app/models/site.rb
davidgisbey: Shouldn't this trigger when the lat/long is updated not when the attribute changes? It's not input anywhere on the UI so the only way it can be triggered is postcode update -> geolocation update -> ttwa update

app/services/travel_to_work_area_and_london_borough_service.rb
davidgisbey: I just backfilled prod 
image

app/models/site.rb
TobyRet: Ah yeah good point - forgot about geocoding

app/models/site.rb
```diff
@@ -31,6 +31,7 @@ class Site < ApplicationRecord
 
   attr_accessor :skip_geocoding
   after_commit :geocode_site, unless: :skip_geocoding
+  after_commit :add_travel_to_work_area_and_london_borough
```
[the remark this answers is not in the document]
davidgisbey: dun

app/services/travel_to_work_area_and_london_borough_service.rb
davidgisbey: done

app/services/travel_to_work_area_and_london_borough_service.rb
davidgisbey: yea boi. you're right

spec/services/travel_to_work_area_and_london_borough_service_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+             latitude: 51.498160,
+             longitude: -0.129900)
+    }
+
+    let(:invalid_site) do
+      build(:site, latitude: "this is not a latitude")
+    end
+
+    let(:valid_site_mapit_endpoint) do
+      URI("https://mapit.mysociety.org/point/4326/#{valid_site.longitude},#{valid_site.latitude}?type=TTW,LBO&api_key=#{Settings.mapit_api_key}")
```
TobyRet, line 22: The mapit host can now be accessed via Settings

TobyRet (approved):

app/models/site.rb
```diff
@@ -42,6 +43,12 @@ def needs_geolocation?
     )
   end
 
+  def add_travel_to_work_area_and_london_borough
+    return false unless needs_travel_to_work_area_and_london_borough_updated?
```
defong, line 47:
```suggestion
```

Don't thing you need that now after 

`   after_commit :add_travel_to_work_area_and_london_borough, if: :needs_travel_to_work_area_and_london_borough_updated?`

defong (approved):

PR #1786, opened 2021-02-24
[3077] - Add MCB command for adding travel to work data to Sites (Version 2)

TobyRet:
### Context

Resurrecting this PR from the dead with fixes 😸 

Whilst working on exposing travel to work data on Find noticed some Courses seemed to be displaying incorrect info (via their Sites)

Looking at the db it appears that there a number of Sites may have had their `london_borough` and `travel_to_work` values mixed up during the back fill.

The `TravelToWorkAreaAndLondonBoroughService` made the assumption that the info from the Mapit API is returned in order (travel to work info is always first) but when I pinged the API this was not the case. 

So for example, there are sites with travel_to_work values containing “Camden Borough Council” when they should actually have a london_borough value containing “Camden”

- Rather than fetch both the `travel_to_work_area` and `london_borough` info in a single request, split these out into separate requests to deal with the randomness

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


defong:
>     * This code works locally but I am getting some weird behaviour with the mcb spec and cannot, or debug it. @defong can you help? CI specs are failing with `Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected`


You touch mcb, you lost 1k of tests

image

TobyRet:
> > ```
> > * This code works locally but I am getting some weird behaviour with the mcb spec and cannot, or debug it. @defong can you help? CI specs are failing with `Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected`
> > ```
> 
> You touch mcb, you lost 1k of tests
> 

All good - issue is fixed now 👍

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -1,20 +1,62 @@
 class TravelToWorkAreaAndLondonBoroughService
   def self.add_travel_to_work_area_and_london_borough(site:)
-    url = URI("#{Settings.mapit_url}/point/4326/#{site.longitude},#{site.latitude}?type=TTW,LBO&api_key=#{Settings.mapit_api_key}")
+    new(site).add_travel_to_work_area_and_london_borough
+  rescue StandardError
```
gpeng, line 4: Can this be more specific or report the error? Not so much of an issue if this is only for the a one off backfill but if it's going to end up being used more widely than that this is going to swallow all errors.

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -1,20 +1,62 @@
...
+    false
+  end
+
+  def initialize(site)
```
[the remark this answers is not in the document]
TobyRet: Cool - done! 👍

app/services/travel_to_work_area_and_london_borough_service.rb
TobyRet: Yeah sorry I did not pay attention here. 
I think we actually want Sentry to deal with any issues right? 
I've removed the `rescue` and added some additional checks to ensure we only update a site upon receiving a `200` status
See https://github.com/DFE-Digital/teacher-training-api/commit/514c972f1e933e879aeb6cbab174e23cc4434b25

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -1,20 +1,60 @@
...
+
+      if london_borough
+        site.update_column("travel_to_work_area", "London")
+        site.update_column("london_borough", london_borough)
+        site.save!(validate: false)
+      end
+    elsif ttw_area
+      site.update_column("travel_to_work_area", ttw_area)
+      site.update_column("london_borough", nil)
+      site.save!(validate: false)
```
defong, line 22: `site.save!(validate: false)`
What kind of validation are you bypassing?
Is an actual need to bypass it?
What is the user case of `update_column`/`update_columns` given that it itself already bypass validation?

https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/Persistence.html#method-i-update_columns

If the site was never valid... then this should be a no ops?

https://github.com/DFE-Digital/teacher-training-api/blob/4463c0ec77df7a0659340e55b88837f4d6533a9b/app/models/site.rb#L18-L26

defong (changes requested):

app/services/travel_to_work_area_and_london_borough_service.rb
TobyRet: @defong If I recall, `validate: false` is there because of imported legacy sites where address data may be dodgy? Same approach is applied with geocoding

TobyRet: I think we originally applied the validations but later the decision was taken (Paul Hayes possibly?) 
to geocode all sites regardless because too many were being excluded

TobyRet: But yes I think we should be doing

```
  site.travel_to_work_area = "London"
  site.london_borough = london_borough
  site.save!(validate: false)
 ```

File: app/services/travel_to_work_area_and_london_borough_service.rb [hunk 4]
```diff
@@ -1,20 +1,60 @@
...
+  def get(attribute)
+    url = if attribute == :travel_to_work_area
+            travel_to_work_url
+          else
+            london_borough_url
+          end
+
+    response = Net::HTTP.get_response(url)
+
+    if response.code == "200"
```
defong, line 41: If its not a `200` do we need to log it?

ie.
Out of quotas?
Out of service?
Out of cash?

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -1,20 +1,60 @@
...
+    ttw_area = get(:travel_to_work_area)
+
+    if ttw_area == "London"
+      london_borough = get(:london_borough)
+
+      if london_borough
+        site.travel_to_work_area = "London"
+        site.london_borough = london_borough
+        site.save!(validate: false)
+      end
```
misaka, line 18: If area is "London" but it's not a london_borough, `call` returns `nil`, whereas if no `ttw_area` is provided `false` is returned. Correct? Does it matter?

TobyRet: Ah good spot - should be returning false - will change 👍

app/services/travel_to_work_area_and_london_borough_service.rb
TobyRet: Yeah good point.

I'm now generating an error that is captured by Sentry when the Mapit API returns a 403 or 5XX status

https://github.com/DFE-Digital/teacher-training-api/pull/1786/commits/963aa6fe0fcace559a69c3de4466d4a4902a8f44

defong (changes requested):

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -1,20 +1,81 @@
...
+
+  def get(attribute)
+    response = Net::HTTP.get_response(url_for(attribute))
+
+    if response.is_a?(Net::HTTPSuccess)
+      json = JSON.parse(response.body)
+      name_attr = json.dig(json.keys.first, "name")
+      name_attr = name_attr&.gsub(/ Borough Council| City Council| Corporation/, "") if attribute == :london_borough
+      name_attr
+    elsif response.is_a?(Net::HTTPForbidden) || response.is_a?(Net::HTTPServerError)
```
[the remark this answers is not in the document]
TobyRet: 👌

app/services/travel_to_work_area_and_london_borough_service.rb
```diff
@@ -1,20 +1,81 @@
...
+      false
+    end
+  end
+
+  def generate_sentry_error(response_code, attribute)
+    e = StandardError.new(
+      "Mapit API has returned status code #{response_code} for Site id #{site.id} whilst trying to obtain #{attribute}",
+    )
+    Raven.capture(e)
+  end
```
defong, line 60: An exception is instantiated but was not raised?

We need to either Capture message, or raise exception?

```suggestion
  def generate_sentry_error(response_code, attribute)
    message = "Mapit API has returned status code #{response_code} for Site id #{site.id} whilst trying to obtain #{attribute}"
    Raven.capture_message(message)
  end
  

```

```suggestion
  def generate_sentry_error(response_code, attribute)
    dead_bodies = StandardError.new(
      "Mapit API has returned status code #{response_code} for Site id #{site.id} whilst trying to obtain #{attribute}",
    )
    raise dead_bodies
  end
```

defong (changes requested):

app/services/travel_to_work_area_and_london_borough_service.rb
TobyRet: I've updated to use `capture_message`

gpeng (approved):

defong (approved):

PR #1803, opened 2021-03-04
Add subject_codes to public/v1 API

duncanjbrown:
### Context

The Apply sync job still uses the V3 API to fetch subject codes. With these on the `public/v1` API we can turn off the v3 API altogether and start to benefit from features of the V1 API like the `updated_since` filter for courses.

Add the field, a spec and documentation.

Introduce a pattern for release notes as discussed on Slack https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1614852108081800

### Guidance to review

Please be nice @defong 🙏

### DIFF
--- a/app/serializers/api/public/v1/serializable_course.rb
+++ b/app/serializers/api/public/v1/serializable_course.rb
@@ -111,6 +111,10 @@ module API
           @object.description
         end
 
+        attribute :subject_codes do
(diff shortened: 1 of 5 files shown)


defong (approved):
Have I ever been un-nice?

Actually lets not answer that one....

PR #1800, opened 2021-03-03
452 add gias edubase data

misaka:
### Context

This is a spike to import GIAS data and see how it compares to our existing Providers data. It's being considered for merging to `master` so we can run it in production.

### Changes proposed in this pull request

- Add webpacker and so the API is no longer strictly only an API.
- Add GIAS pages to import and explore GIAS data.
- Add DSI integration to ensure only admins can view the GIAS pages.

### Guidance to review

We want to make sure that only admins can view these pages, and that adding webpacker and everything else does not impact the API aspect of this service. This change should be pretty thoroughly tested before merging to master.

More details for how to test the importing, etc, to come.

### Checklist

- [ ] Make sure all information from the Trello card is in here
- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Tested by running locally

PR #1821, opened 2021-03-16
1114 dsi

defong:
### Context

This application was once an `api only` application.

Changes has been made to enable it to be a fully functional web application.

Most was ported from #1800 and strip back down to bare necessitates as the source was becoming a bit too encroach with becoming one of those one hit wonder.

Added common frontend stuff
Added DSI integrations

### Guidance to review
Pull down the repo and give it a spin,

Successfully `admin user` login will lead to a dead end/error page

There will be a follow up PR to further divide and partition #1800 

image

Successfully `user` login will lead to a 
image

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


defong:
using #1823  instead

PR #1833, opened 2021-03-26
[1114] Add frontend and DSI

gpeng:
### Context

We are adding a support interface to TTAPI and need to enable DfE Signin authentication. The app is currently API only so we also need to add the asset pipeline etc so as we can add a UI.

* Configure for DfE signin
* Add the non-API only parts of Rails that we require

### Guidance to review

This has been deployed to QA so can be reviewed there. Not working locally currently as we need further DSI configuration (at the DSI end).

This is an initial stab at setting this up and will be iterated on as we add functionality. There are issues to resolve with the UI, styles etc but the focus here is to get authentication setup.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


Gemfile
```diff
@@ -6,6 +6,9 @@ ruby "2.7.2"
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem "rails", "6.1.3.1"
 
+# Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
+gem "webpacker"
```
darokel, line 10: might be worth adding the version here, the latest pre-release 6.x in beta introduces some breaking changes from the current stable 5.x release

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,85 @@
...
+    <%= tag :meta, name: "ROBOTS", content: "NOINDEX, NOFOLLOW" %>
+    <%= favicon_link_tag asset_pack_path("media/images/favicon.ico") %>
+    <%= favicon_link_tag asset_pack_path("media/images/govuk-mask-icon.svg"), rel: "mask-icon", type: "image/svg", color: "#0b0c0c" %>
+    <%= favicon_link_tag asset_pack_path("media/images/govuk-apple-touch-icon.png"), rel: "apple-touch-icon", type: "image/png" %>
+    <%= favicon_link_tag asset_pack_path("media/images/govuk-apple-touch-icon-152x152.png"), rel: "apple-touch-icon", type: "image/png", size: "152x152" %>
+    <%= favicon_link_tag asset_pack_path("media/images/govuk-apple-touch-icon-167x167.png"), rel: "apple-touch-icon", type: "image/png", size: "167x167" %>
+    <%= favicon_link_tag asset_pack_path("media/images/govuk-apple-touch-icon-180x180.png"), rel: "apple-touch-icon", type: "image/png", size: "180x180" %>
+    <%= stylesheet_pack_tag "application", media: "all" %>
+    <%= javascript_pack_tag "application", defer: true %>
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
```
darokel, line 20: Given that this will be a support portal, do we actually need the analytics and tag manager stuff here?

darokel (approved):
Left a couple of comments but looks good otherwise

Gemfile
gpeng: Fixed

app/views/layouts/application.html.erb
gpeng: We do not. I've removed it. 👍

PR #1841, opened 2021-04-01
[1315] Provider show page with users tab

jhenderson:
### Context
https://trello.com/c/hpDqCfE2/1315-m-provider-show

- Add tab navigation to provider show page
- Add users index resource

Taking inspiration from Apply, when the user clicks on a provider, the show page lands on the "Details" tab (to be filled in later).

### Guidance to review
- Visit `/support/providers`
- Click on a provider
- Click on the "Users" tab
- Confirm table of users is displayed

Screenshot 2021-04-01 at 14 05 34

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/support/users/index.html.erb
+++ b/app/views/support/users/index.html.erb
@@ -1,0 +1,25 @@
+<table class="govuk-table">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th scope="col" class="govuk-table__header govuk-!-width-one-quarter">First name</th>
+      <th scope="col" class="govuk-table__header govuk-!-width-one-quarter">Last name</th>
+      <th scope="col" class="govuk-table__header govuk-!-width-two-quarters">Email</th>
+    </tr>
+  </thead>
+
+  <tbody class="govuk-table__body">
+    <% @users.each do |user| %>
+      <tr class="govuk-table__row user-row">
+        <td class="govuk-table__cell">
+            <span class="govuk-!-display-block govuk-!-margin-bottom-1"><%= user.first_name %></span>
+        </td>
+        <td class="govuk-table__cell">
+          <span class="govuk-!-display-block govuk-!-margin-bottom-1"><%= user.last_name %></span>
+        </td>
+        <td class="govuk-table__cell">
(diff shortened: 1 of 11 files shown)


jhenderson:
[[M] Provider show](https://trello.com/c/hpDqCfE2/1315-m-provider-show)

spec/support/page_objects/support/provider_show.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module PageObjects
+  module Support
+    class ProviderShow < PageObjects::Base
+      set_url "support/providers/{id}"
+
+      element :users_tab, ".app-tab-navigation__link", text: "Users"
+
+      def providers
```
darokel, line 10: I don't think is needed? I couldn't see any page object using it for this PR

app/views/support/users/index.html.erb
```diff
@@ -0,0 +1,32 @@
+<h1 class="govuk-heading-l"><%= @provider.provider_name %></h1>
+
+<%= render TabNavigation::View.new(items: [
```
darokel, line 3: I think we may be able to use the pattern from register to extend a layout which will just have this one component rendered. We do it like this in register for the trainee record page:

https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/views/layouts/trainee_record.html.erb

I can take a stab at it in my next PR if you want to leave it for now.

spec/support/page_objects/support/provider_show.rb
jhenderson: Goot spot. Copy and paste fail.

darokel (approved):

app/views/support/users/index.html.erb
jhenderson: Good shout. I've implemented the solution your suggested. Might need more refinements later on, but for now, it should do the job.

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,4 @@
 class ApplicationController < ActionController::Base
-  before_action :authenticate
```
darokel, line 2: Think we need this back in :)

PR #1850, opened 2021-04-08
[1317] Add edit provider name feature and summary list for providers

JonF1:
### Context
https://trello.com/c/cwKSUfTf/1317-s-add-edit-provider-name-feature-to-providers-pages

* Adds summary list component with change link.  Added name, code and last updated but can swap if other fields wanted. Link takes the user to form where they can change name. Cancel option to return if needed. Last updated date uses gov_uk format.
* Uniqueness and length validations. I set 100 chars limit as no provider name in db is more than 80 chars 
* Feature test
### Guidance to review
Go to `/support/providers` and select a provider
You may have to comment out `before_action :authenticate` in application controller

Screenshot 2021-04-13 at 15 47 18
Screenshot 2021-04-08 at 17 08 04

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


JonF1:
[[S] Add "Edit provider name" feature to providers' pages](https://trello.com/c/cwKSUfTf/1317-s-add-edit-provider-name-feature-to-providers-pages)

app/webpacker/styles/application.scss
```diff
@@ -1,3 +1,12 @@
...
 @import "~@ministryofjustice/frontend/moj/all";
+
+.app-provider-name {
+   display: inline;  
+   .app-provider-edit-link {
+    @include govuk-font($size: 19);
+     display: inline;
+     padding: govuk-spacing(2)
+   }
+}
```
darokel, line 12: line endings 🙈

darokel (commented):
Looking good. I wonder if we should just follow this pattern to allow names to be changed:

Screenshot 2021-04-09 at 16 01 43

We should probably start populating this summary view in the details tab and show their name there with the change link to the right? This would match our pattern in register.

JonF1:
> We should probably start populating this summary view in the details tab and show their name there with the change link to the right? This would match our pattern in register.

Ok I'll add blocked to the ticket

app/webpacker/styles/application.scss
JonF1: i fixed by removing all the css 😆

app/views/support/providers/edit.html.erb
```diff
@@ -0,0 +1,14 @@
...
+      
+  <%= f.govuk_error_summary %>
+
+  <%= f.govuk_text_field :provider_name, label: { text: "New Provider name", size: "s" }, width: 20 %>
+  <%= f.govuk_submit %>
+  <% end %>
+
+  <%= govuk_link_to(t("cancel"), support_provider_path) %>
+  </div>
+</div>
```
darokel, line 14: line ending

app/views/support/providers/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    key: 'Code',
+    value: @provider.provider_code
+  )
+
+  component.slot(
+    :row,
+    key: 'Last updated',
+    value: @last_updated
+  )
+end %>
```
darokel, line 20: line ending

app/views/support/providers/show.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= render GovukComponent::SummaryList.new do |component|
+  component.slot(
+    :row,
+    key: 'Name',
```
darokel, line 4: could we swap for double quotes

app/webpacker/styles/application.scss
```diff
@@ -1,3 +1,3 @@
 @import "./govuk_style_setup";
 @import "~govuk-frontend/govuk/all";
-@import "~@ministryofjustice/frontend/moj/all";
+@import "~@ministryofjustice/frontend/moj/all";
```
darokel, line 3: line ending

app/controllers/support/providers_controller.rb
```diff
@@ -26,5 +41,17 @@ def filters
     def filter_params
       params.permit(:text_search, :page, :commit)
     end
+
+    def last_updated_provider
```
darokel, line 45: this should go in a helper i think (and the included TimeFormat)

darokel (changes requested):
Looking good. Left a couple of comments to get this over the line

app/helpers/support_time_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module SupportTimeHelper
```
darokel, line 3: I think these helpers should be nested in a support directory to follow convention of other parts. `app/helpers/support/time_helper.rb` then you'd have:

```
module Support
  module TimeHelper
  end
end
```

darokel (approved):

PR #1858, opened 2021-04-14
[1427] Show a list of the providers a user has access to

AbigailMcP:
### Context

https://trello.com/c/HtLxI0c5/1427-s-display-a-list-of-providers-that-a-user-has-access-to

- Adds links to the users table on `support/providers/:id/users`, which redirect to:
- A new user show page on `support/users/:id`, which shows:
- A paginated list of the providers that user has access to in the current recruitment cycle, which looks something like:

Screenshot 2021-04-14 at 15 15 56

(page limit for pagination set artificially to four for the example screenshot, in reality it is 25 like elsewhere)

### Guidance to review

- Head to `support/providers/:id/users` for a provider
- Click on a user
- See that that user's providers are listed

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/support/providers/_list.html.erb
+++ b/app/views/support/providers/_list.html.erb
@@ -1,0 +1,26 @@
+<table class="govuk-table">
+  <thead class="govuk-table__head">
+    <tr class="govuk-table__row">
+      <th scope="col" class="govuk-table__header govuk-!-width-one-half">Provider name</th>
+      <th scope="col" class="govuk-table__header govuk-!-width-one-half">Provider code</th>
+    </tr>
+  </thead>
+
+  <tbody class="govuk-table__body">
+    <% @providers.each do |provider| %>
+      <tr class="govuk-table__row qa-provider_row">
+        <td class="govuk-table__cell">
+          <span class="govuk-!-display-block govuk-!-margin-bottom-1">
+            <%= govuk_link_to provider.provider_name, support_provider_path(provider) %>
+          </span>
+        </td>
+
+        <td class="govuk-table__cell">
+          <span class="govuk-!-display-block govuk-!-margin-bottom-1">
+            <%= provider.provider_code %>
+          </span>
+        </td>
+      </tr>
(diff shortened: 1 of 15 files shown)


AbigailMcP:
[[S] Display a list of providers that a user has access to](https://trello.com/c/HtLxI0c5/1427-s-display-a-list-of-providers-that-a-user-has-access-to)

spec/support/page_objects/support/user_show_page.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module PageObjects
+  module Support
+    class UserShow < PageObjects::Base
+      set_url "/support/users/{id}"
+
+      sections :provider_rows, PageObjects::Sections::Provider, ".qa-provider_row"
```
AbigailMcP, line 8: This had me stumped for a while - I originally called these sections `:providers` but that conflicted with api request specs that defined `@providers`. For example:

```
        before do
          @providers = Array.new(25) do |i|
            create(:provider, provider_code: "PROV#{i + 1}",
                   changed_at: (30 - i).minutes.ago,
                   sites: [])
          end
        end
```

in `spec/requests/api/v1/provider_spec.rb` 😱 

I've just changed the name of to `provider_rows` to avoid the conflict, but I don't feel great about the underlying!

app/views/support/providers/_list.html.erb
```diff
@@ -0,0 +1,26 @@
...
+
+        <td class="govuk-table__cell">
+          <span class="govuk-!-display-block govuk-!-margin-bottom-1">
+            <%= provider.provider_code %>
+          </span>
+        </td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
```
gpeng, line 26: Needs a newline

gpeng (approved):

dankmitchell (approved):

PR #1875, opened 2021-04-21
[1497] Validations added to urn

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/ymwbAv9m/1497-s-add-validation-to-urn-only-applicable-to-lead-school-providers)

- Added validation for presence of urn in lead schools, and a length range of between 5 and 6 characters for the urn if it is present

### Guidance to review

- Open your rails console `rails c`
- Create an invalid provider like so (a lead school provider without a URN, that is) 
- `Provider.create!(provider_type: "Y", recruitment_cycle: RecruitmentCycle.first)`
- You should see an error
- You can validate it by adding a urn, which will need to be 5-6 characters long. 
- No URN needed if the provider_type is anything but lead school

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -108,7 +108,7 @@ class Provider < ApplicationRecord
 
   validates :provider_name, length: { maximum: 100 }, on: :update, if: RecruitmentCycle.current_recruitment_cycle
 
-  validates :urn, length: { in: 5..6 }, if: -> { provider_type == "lead_school" }
+  validates :urn, length: { in: 5..6 }, if: :lead_school?, allow_nil: true
 
   validates :telephone, phone: { message: "^Enter a valid telephone number" }, if: :telephone_changed?
 
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -108,6 +108,9 @@ class Provider < ApplicationRecord
 
   validates :provider_name, length: { maximum: 100 }, on: :update, if: RecruitmentCycle.current_recruitment_cycle
 
+  validates :urn, length: { in: 5..6 }, unless: -> { urn.blank? }
+  validates :urn, presence: true, if: -> { provider_type == "lead_school" }
+
   validates :telephone, phone: { message: "^Enter a valid telephone number" }, if: :telephone_changed?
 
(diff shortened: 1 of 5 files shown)


defong:
[[S] Add validation to URN (only applicable to lead school providers)](https://trello.com/c/ymwbAv9m/1497-s-add-validation-to-urn-only-applicable-to-lead-school-providers)

app/models/provider.rb
```diff
@@ -108,6 +108,9 @@ def current_accredited_courses
 
   validates :provider_name, length: { maximum: 100 }, on: :update, if: RecruitmentCycle.current_recruitment_cycle
 
+  validates :urn, length: { in: 5..6 }, unless: -> { urn.blank? }
+  validates :urn, presence: true, if: -> { provider_type == "lead_school" }
```
defong, line 112: ummm not too sure....

only apply validation if `lead_school` and for it to be 5 or 6 in length and allow for nil...

stellenberger: Done 👍

spec/models/provider_spec.rb
```diff
@@ -28,6 +28,28 @@
     it { should have_many(:user_notifications) }
   end
 
+  describe "urn validations" do
+    context "when provider_type is not lead_school" do
+      let(:provider) { build(:provider, provider_type: "B") }
+      let(:provider_with_no_urn) { build(:provider, provider_type: "B", urn: nil) }
```
defong, line 34: Cna you use one if the traits...

spec/factories/providers.rb
```diff
@@ -21,6 +21,7 @@
     website { Faker::Internet.url }
     provider_type { :lead_school }
     accrediting_provider { "N" }
+    urn { Faker::Number.number(digits: 5).to_s }
```
defong, line 24: can you put that in a trait

```
urn { Faker::Number.number(digits: [5, 6].sample) }
```

dont think you need `to_s`

defong (changes requested):
almost

defong (approved):

PR #1882, opened 2021-04-22
[1503] Add validations to URN and UKPRN

avinhurry:
### Context

This PR prevents users from entering a URN or UKPRN in an incorrect format.

- Validate that the string is 5-6 numbers on URN (Site table)
- Validate that the string is 8 numbers on UKPRN (Provider table)
- Allow blank values
- Custom validators
- Add staging data dump file to .gitignore

### Guidance to review

- Pull down the code and ensure you have the latest migrations
- Boot up the rails console and navigate to a Site
- Ensure the URN will only be saved if it is nil, 5 or 6 numbers

- Repeat for UKPRN on the provider table. 
- Ensure the UKPRN will only be saved if it is nil, or 8 numbers

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -25,6 +25,8 @@ class Site < ApplicationRecord
                    inclusion: { in: POSSIBLE_CODES, message: "must be A-Z, 0-9 or -" },
                    presence: true
 
+  validates :urn, reference_number_format: { allow_blank: true, minimum: 5, maximum: 6, message: "^URN must be 5 or 6 numbers" }
+
   acts_as_mappable lat_column_name: :latitude, lng_column_name: :longitude
 
   scope :not_geocoded, -> { where(latitude: nil, longitude: nil) }
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ bin/fetch_config.rb
 
 # Ignore sanitised production data dump
 *.sql.gz
+*.sql
 
 # Exclude api doc application node modules
 docs/node_modules
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -108,11 +108,11 @@ class Provider < ApplicationRecord
 
   validates :provider_name, length: { maximum: 100 }, on: :update, if: RecruitmentCycle.current_recruitment_cycle
 
-  validates :urn, length: { in: 5..6 }, if: :lead_school?, allow_nil: true
-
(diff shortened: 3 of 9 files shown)


avinhurry:
[UKPRN/URN should only accept the characters 0-9](https://trello.com/c/jWUb0be7/1589-ukprn-urn-should-only-accept-the-characters-0-9)

defong (changes requested):
https://guides.rubyonrails.org/active_record_validations.html#performing-custom-validations

app/models/site.rb
```diff
@@ -4,6 +4,7 @@ class Site < ApplicationRecord
   include PostcodeNormalize
   include RegionCode
   include TouchProvider
+  include ActiveModel::Validations
```
darokel, line 7: This is already included by `ActiveRecord`

avinhurry: thanks, amended

app/validators/ukprn_validator.rb
```diff
@@ -0,0 +1,9 @@
+class UkprnValidator < ActiveModel::Validator
```
defong, line 1: Used the wrong base case....
 Should be `ActiveModel::EachValidator`

defong (changes requested):
this ties it down needlessly

defong (changes requested):
you only need one validator

app/validators/ukprn_urn_format_validator.rb
```diff
@@ -0,0 +1,9 @@
+class UkprnUrnFormatValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.blank?
```
defong, line 3: Can we make this an option like `allow_blank` so this can be overrided

app/validators/ukprn_urn_format_validator.rb
```diff
@@ -0,0 +1,9 @@
...
+
+    if value.length > options[:maximum] || value.length < options[:minimum] || !value.match?(/\A-?\d+\Z/)
```
defong, line 5: Why didn't you used all regex...

defong (changes requested):

defong (commented):
Almost can you add the respective tests here please

https://github.com/DFE-Digital/teacher-training-api/tree/master/spec/validators

defong (approved):

PR #1888, opened 2021-04-28
Add ADR for performance monitoring

archferns:
Add an ADR to record decision around using skylight for performance monitoring.

### DIFF
--- a/docs/adr/0008-use-skylight-for-performance-monitoring.md
+++ b/docs/adr/0008-use-skylight-for-performance-monitoring.md
@@ -1,0 +1,45 @@
+# 8. Use Skylight for performance monitoring
+
+Date: 2021-04-26
+
+## Status
+
+Accepted
+
+## Context
+
+We need to decide between using skylight versus sentry for performance monitoring.
+
+## Options
+
+### 1. [Skylight](https://www.skylight.io/)
+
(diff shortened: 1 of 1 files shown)


docs/adr/0008-use-skylight-for-performance-monitoring.md
```diff
@@ -0,0 +1,45 @@
...
+- Provides detailed insight into each request
+- Can be configured to trace a subset of requests
+
+#### Cons
+
+- Very expensive considering our daily hits (400k/day)
+
+## Decision
+
+[There was an attempt](https://github.com/DFE-Digital/teacher-training-api/pull/1860) to use sentry, but we managed to fill up our quota in less than a day. As such, we have decided to continue using skylight for performance monitoring.
```
misaka, line 45:
```suggestion
[There was an attempt](https://github.com/DFE-Digital/teacher-training-api/pull/1860) to use sentry, but we managed to fill up our quota in less than a day, and while Sentry does offer ways to reduce traffic by selective tracing, this would only allow us to trace less than 1% of our requests.

As such, we have decided to continue using skylight for performance monitoring.
```

misaka (approved):
Looks really good. Thanks @archferns ... we need to get some traction on doing these ADRs. :100:

PR #1910, opened 2021-05-11
[1678] Touch provider when course is updated

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/Dbm8Ct4t/1678-s-touch-provider-when-course-is-updated)

- Added TouchProvider module to the course model to ensure that the provider model records updates to the course, through `changed_at`. 

### Guidance to review

- Open your rails console `$ rails c`
- `$ course = Course.all.sample`
- `$ course.changed_at` => outputs sometime in the past
- `$ course.provider.changed_at` => outputs sometime in the past
- `$ course.name = "A new name!"`
- `$ course.save`
- `$ course.changed_at` => outputs today's date
- `$ course.provider.changed_at` => outputs today's date # this is the important part
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -2,6 +2,7 @@ class Course < ApplicationRecord
   include Discard::Model
   include WithQualifications
   include ChangedAt
+  include TouchProvider
   include Courses::EditOptions
   include StudyModeVacancyMapper
   include TimeFormat
--- a/spec/requests/api/v1/course_spec.rb
+++ b/spec/requests/api/v1/course_spec.rb
@@ -125,42 +125,6 @@ describe "Courses API", type: :request do
                               },
                             ])
       end
-
-      it "includes correct next link in response headers" do
-        timestamp_of_first_course = 10.minutes.ago
-        first_course = create(:course,
-                              :infer_level,
-                              course_code: "LAST1",
-                              provider: provider)
-
-        Timecop.travel(timestamp_of_first_course) do
-          create(:site_status, :published, course: first_course)
-        end
-
-        timestamp_of_last_course = 2.minutes.ago
-
-        last_course_in_results = create(:course,
-                                        :infer_level,
-                                        course_code: "LAST2",
-                                        provider: provider)
-        Timecop.freeze(timestamp_of_last_course) do
-          create(:site_status, :published, course: last_course_in_results)
-        end
-
-        get "/api/v1/#{current_year}/courses",
-            headers: { "HTTP_AUTHORIZATION" => credentials }
-
-        expect(response.headers).to have_key "Link"
-        url = url_for(
-          recruitment_year: current_year,
-          params: {
-            changed_since: timestamp_of_last_course.utc.strftime("%FT%T.%6NZ"),
-            per_page: 100,
-          },
-        )
-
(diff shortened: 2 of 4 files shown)


app/models/course.rb
```diff
@@ -2,6 +2,7 @@ class Course < ApplicationRecord
   include Discard::Model
   include WithQualifications
   include ChangedAt
+  include TouchProvider
```
TobyRet, line 5: @stellenberger - looks like we may overlapping cards! Was about to raise a PR for the same change
https://trello.com/c/RsbjA4sq/3299-incrementally-sync-courses-from-the-ttapi

My only suggestion here is rather than using the TouchProvider concern you can simply do
```
belongs_to :provider, touch: :changed_at
```

... this means the TouchProvider and TouchCourse concerns could be deleted in another PR but I'll let @gpeng @defong  and company make a call on this

spec/requests/api/v1/course_spec.rb
```diff
@@ -128,22 +128,22 @@ def get_course_codes_from_body(body)
 
       it "includes correct next link in response headers" do
         timestamp_of_first_course = 10.minutes.ago
-        Timecop.freeze(timestamp_of_first_course) do
-          first_course = create(:course,
-                                :infer_level,
-                                course_code: "LAST1",
-                                provider: provider)
+        first_course = create(:course,
```
TobyRet, line 131: mmm ... are the request specs necessary? If so I would imagine something similar will be required for v2, v3 and public api request specs

app/models/course.rb
gpeng: I prefer @TobyRet's suggestion. I'm not a fan of the `changed_at` field either but if we already have everything built around that rather than `updated_at` then we'll have to press on with that for now.

spec/requests/api/v1/course_spec.rb
gpeng: We are only really supporting the update_since stuff in /v3 and /public/v1 so we definitely don't need to update this version's spec. The existing specs for the other versions should be fine as is I would have thought 🤷

gpeng (commented):
^

app/models/course.rb
stellenberger: The `TouchProvider` module will only change the `changed_at` field for the provider when we touch it, and not the `updated_at` as it uses `update_columns`. I am happy to implement @TobyRet's suggestion, though it will end up changing both `updated_at` and `changed_at` fields for the provider when we touch it. If it's not important that both fields are updated, I'll push up @TobyRet's suggestion! 👍

spec/requests/api/v1/course_spec.rb
stellenberger: These specs were failing after the change, so its just a re-order of object creation outside of Timecop so they pass (we have a unique index on `changed_at`).
I can remove them altogether if they are obsolete?

defong: Whenever `Apply` kicks in then this will be obsolete

defong (commented):
https://github.com/DFE-Digital/teacher-training-api/pull/1910#pullrequestreview-657731241

^

defong (approved):

PR #1938, opened 2021-05-27
[1766] Add ability to add a provider from support UI

archferns:
### Context
New form to add a provider, with one organisation and site via the support UI.
 
1. Add new and create routes against support/providers
2. Add a couple of more validations against provider name and code

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


archferns:
[[L] Add provider](https://trello.com/c/YyTLWtbP/1766-l-add-provider)

app/controllers/support/providers_controller.rb
```diff
@@ -48,8 +63,24 @@ def provider
...
+    def update_provider_params
       params.require(:provider).permit(:provider_name)
     end
+
+    def create_provider_params
+      params.require(:provider).permit(:provider_name,
+                                       :provider_code,
+                                       :recruitment_cycle_id,
+                                       :email,
+                                       :telephone, sites_attributes: %i[id
```
darokel, line 75: Since this is a new site/organisation being created, do we need the id here?

app/views/layouts/_flash.html.erb
```diff
@@ -0,0 +1,25 @@
+<% flash.each do |key, value| %>
```
darokel, line 1: We've been re-using existing components from register so we don't rebuild things twice. Is there way we can re-use the flash component here? A bit of work was put it into it.

spec/features/support/providers/creating_a_provider_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  scenario "adds a new provider" do
+    when_i_fill_in_a_valid_provider_details
+    and_i_fill_in_a_valid_organisation_details
+    and_i_fill_in_a_valid_site_details
+    and_i_click_the_submit_button
+    then_i_am_redirected_back_to_the_provider_show_page
+    and_the_provider_is_created
+  end
+
+  context "invalid details" do
```
darokel, line 22: I think the consensus was to avoid using `context` wraps and potentially too many nested ones in feature specs to keep things simple

darokel (commented):
Nearly there but have suggested a few changes:

**Button**

I think we use apply's approach and have the button above the filter and result set

Screenshot 2021-06-02 at 10 41 08

app/views/layouts/_flash.html.erb
darokel: https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/components/notification_banner/view.rb and https://github.com/DFE-Digital/register-trainee-teachers/blob/master/app/components/flash_banner/view.rb

app/views/support/providers/new.html.erb
```diff
@@ -0,0 +1,54 @@
...
+
+      <fieldset class="govuk-fieldset">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+          <h1 class="govuk-fieldset__heading">Add a new provider</h1>
+        </legend>
+
+
+        <%= f.govuk_text_field :provider_name, label: { text: "Provider name", size: "s" } %>
+        <%= f.govuk_text_field :provider_code, label: { text: "Provider code", size: "s" } %>
+        <%= f.govuk_collection_select :recruitment_cycle_id,
```
darokel, line 15: I think we only want to use the current RecruitmentCycle so this field might be redundant

app/controllers/support/providers_controller.rb
archferns: Good point, removed the ids.

app/views/layouts/_flash.html.erb
archferns: Added these in from register

app/views/support/providers/new.html.erb
archferns: Removed this as well :+1:

darokel (approved):
👍

spec/components/flash_banner/view_preview.rb
```diff
@@ -0,0 +1,26 @@
...
+      render(FlashBanner::View.new(flash: flash(:warning)))
+    end
+
+    def with_info
+      render(FlashBanner::View.new(flash: flash(:info)))
+    end
+
+  private
+
+
```
defong, line 19: extra new line

defong (approved):

dankmitchell (approved):

PR #1939, opened 2021-05-27
[1785] Copy Allocations data into recruitment cycle 3

mooktakim:
### Context

https://trello.com/c/sJvRcyC6/1785-copy-allocations-data-into-recruitment-cycle-3

* Rake task `allocations:copy_allocations_data_to_cycle_year` to copy allocations from previous allocation year to current allocation year

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/services/allocation_copy_data_service.rb
+++ b/app/services/allocation_copy_data_service.rb
@@ -1,0 +1,82 @@
+class AllocationCopyDataService
+  include ServicePattern
+
+  def initialize(allocation_cycle_year:, summary: false)
+    @allocation_cycle_year = allocation_cycle_year
+    @summary = summary
+  end
+
+  def call
+    copied = 0
+    skipped = 0
+
+    Allocation.where(recruitment_cycle_id: previous_allocation_cycle.id).find_each do |prev_alloc|
+      if prev_alloc.confirmed_number_of_places.to_i.zero? ||
+          allocation_exists?(prev_alloc)
+        skipped += 1
+      else
+        provider = find_provider(prev_alloc)
+        accredited_body = find_accredited_body(prev_alloc)
+        create_allocation(prev_alloc, provider, accredited_body)
+        copied += 1
+      end
+    end
+
+    if @summary
+      puts "###################################"
+      puts "## Allocation copying task complete"
+      puts "# copied: #{copied}"
+      puts "# skipped: #{skipped}"
+      puts "###################################"
+    end
+  end
+
+private
+
+  def create_allocation(prev_alloc, provider, accredited_body)
+    Allocation.create!(
+      provider_code: prev_alloc.provider_code,
+      accredited_body_code: prev_alloc.accredited_body_code,
+      recruitment_cycle_id: allocation_cycle.id,
+
+      provider_id: provider.id,
+      accredited_body_id: accredited_body.id,
+
+      number_of_places: prev_alloc.confirmed_number_of_places,
+      confirmed_number_of_places: nil,
+    )
+  end
+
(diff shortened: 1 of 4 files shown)


mooktakim:
[Copy Allocations data into recruitment cycle 3](https://trello.com/c/sJvRcyC6/1785-copy-allocations-data-into-recruitment-cycle-3)

lib/tasks/allocations.rake
```diff
@@ -0,0 +1,6 @@
+namespace :allocations do
+  desc "Copy previous allocations data to allocation cycle year: rake allocations:copy_allocations_data_to_cycle_year"
+  task copy_allocations_data_to_cycle_year: :environment do
+    AllocationCopyDataService.call(allocation_cycle_year: Settings.allocation_cycle_year)
```
AbigailMcP, line 4: I _think_ (correct me if I'm wrong @gpeng ) we'll need to run this task before updating the `allocation_cycle_year`, otherwise the dates will visually change before the data is in. It might be simplest for this rake task to accept `allocation_cycle_year ` as an argument.

gpeng (approved):

app/services/allocation_copy_data_service.rb
```diff
@@ -0,0 +1,67 @@
...
+      number_of_places: prev_alloc.confirmed_number_of_places,
+      confirmed_number_of_places: nil,
+    )
+  end
+
+  def find_provider(prev_alloc)
+    # Fetch new provider with same code for current recruitment cycle
+    Provider.where(
+      provider_code: prev_alloc.provider_code,
+      recruitment_cycle_id: current_cycle.id,
```
defong, line 39:
```suggestion
      recruitment_cycle: current_cycle,
```

Shouldn't it be like that?

app/services/allocation_copy_data_service.rb
```diff
@@ -0,0 +1,67 @@
...
+  def allocation_exists?(prev_alloc)
+    Allocation.where(
+      provider_code: prev_alloc.provider_code,
+      accredited_body_code: prev_alloc.accredited_body_code,
+      recruitment_cycle_id: current_cycle.id,
+    ).exists?
+  end
+
+  def current_cycle
+    @current_cycle ||= RecruitmentCycle.where(year: @allocation_cycle_year).first!
```
gpeng, line 61: I think there's a `RecruitmentCycle.current` method. Maybe calling this `current_cycle` is a bit ambiguous as it may not be the current it's just the cycle that matches the allocation cycle.

app/services/allocation_copy_data_service.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def call
+    Allocation.where(recruitment_cycle_id: previous_cycle.id).find_each do |prev_alloc|
+      next if prev_alloc.confirmed_number_of_places.to_i.zero? ||
+        allocation_exists?(prev_alloc)
+
+      provider = find_provider(prev_alloc)
+      accredited_body = find_accredited_body(prev_alloc)
+      create_allocation(prev_alloc, provider, accredited_body)
+    end
```
AbigailMcP, line 16: Is there anything useful that we'd like to know at the end of running this task? Maybe how many we copied over vs skipped?

lib/tasks/allocations.rake
gpeng: Yes that's right I think. If we pass it as an argument my `current` comment up there might be more important.

mooktakim: Updated to allow you to pass in year.

PR #1940, opened 2021-05-27
Add visa booleans to the providers table

davidgisbey:
### Context

We're about to start some work on publish which allows providers to give information on whether they can sponsor skilled worker and/or student visas. This commit adds two booleans to the provider table to capture their response.

- Migration to add two booleans to the providers table to capture information on visas 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


db/migrate/20210527145331_add_visa_columns_to_the_database.rb
```diff
@@ -0,0 +1,6 @@
+class AddVisaColumnsToTheDatabase < ActiveRecord::Migration[6.1]
+  def change
+    add_column :provider, :can_sponsor_skilled_worker_visa, :boolean
```
dankmitchell, line 3: Do we want defaults here? I.e. ‘false’ Or are we okay with `nil`?

dankmitchell (commented):
One question cc. @gpeng

db/migrate/20210527145331_add_visa_columns_to_the_database.rb
davidgisbey: Happy to default to false, but i tend to prefer it being nil, just because it's making populating the radio buttons correctly easier.

```
If form.attr == true
  choose yes
elsif form.attr == false
  choose no 
else 
  no value 
end 
```

Let me know what you prefer 👍

darokel: @davidgisbey  this should ideally default to false and enabled when required otherwise you have 3 possible states to deal with as above. You can handle whether or not to display an unselected value in the view by using `false` which seems like a concern it should deal with

darokel: @davidgisbey I think these sort of fields should ideally default to false and enabled when required otherwise you have 3 possible states to deal with as above which adds unnecessary complexity. I might be missing some context but if it's for a new provider you can just initialise your form with empty values and handle validations etc. If it's editing an existing provider, wouldn't these fields be more of an opt-in? Ie "yes I can sponsor a skilled worker visa"

PR #1967, opened 2021-06-09
Add structured degree fields to the course table

davidgisbey:
### Context

Ahead of adding in the new degree requirements section, we need some new columns in the course database.

This PR adds 3 columns.

Then adds an enum for the degree_classification_required to the course model

- degree_classification_required - This will be an enum which captures all the grades or not_required
- additional_subject_requirements - Required to ensure the correct radio button is selected on the new/edit views
- subject_requirements - used to capture their free text answer

### Guidance to review

https://trello.com/c/KAPUFr0o/3494-add-structured-degree-requirements-fields-to-the-v2-ttapi

Suggestion on naming are more than welcome

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/models/course.rb
```diff
@@ -40,6 +40,13 @@ class Course < ApplicationRecord
     further_education: "Further education",
   }, _suffix: :course
 
+  enum degree_classification_required: {
```
gpeng, line 43: I don't think we need the `required` here. Also, register calls this `degree_grade`.

If this is an enum shouldn't it be an integer column and then map `two_one: 0` etc?

db/migrate/20210609081246_add_structured_degree_columns_to_the_course_table.rb
```diff
@@ -0,0 +1,7 @@
+class AddStructuredDegreeColumnsToTheCourseTable < ActiveRecord::Migration[6.1]
+  def change
+    add_column :course, :degree_classification_required, :string
+    add_column :course, :additional_subject_requirements, :boolean
```
gpeng, line 4: Not sure I understand this one. From the description is there a radio button that exposes `subject_requirements`? Can that be checked (or not) from the presence (or absence) of a value in `subject_requirements`. That sounds like a view concern if so that the API doesn't need to know about?

app/models/course.rb
gpeng: Just to clarify there, I wasn't suggesting this has to be named the same as register but it would be good to be consistent across 'all of the things :tm:' if we can.

db/migrate/20210609081246_add_structured_degree_columns_to_the_course_table.rb
davidgisbey: So it looks like this 
image

Just wondering how you would you select the correct radio button when editing though if you don't persist their  answer for no.

Just default the radio button to no if no value is present for subject_requirements? Or just don't have a default value selected for no when editing?

Or just infer that if you're on the edit page that they must have selected no if there is no value?

app/models/course.rb
davidgisbey: Yea no worries. Happy to change it to degree grade. I was just thinking it might get a bit funky down the line as degree_grade may be used in other places to collect a candidates degree grade rather than the required one to get onto a course.

defong (approved):
LGTM

PR #1969, opened 2021-06-09
Permit visa attributes in V2 provider updates

stevehook:
### Context

This is needed for Publish to be able to manage these new attributes. `Provider` attributes `can_sponsor_student_visa` and `can_sponsor_skilled_worker_visa` were added recently so that providers could declare information about the kind of visas they sponsor. This PR just attempts to make those attributes accessible via the `/api/v2/recruitment_cycles/:recruitment_cycle_year/providers/:provider_code` endpoint.

https://trello.com/c/I2K4fchq/3441-add-visa-flow-to-publish

- [x] Add `can_sponsor_student_visa` and `can_sponsor_skilled_worker_visa` to permitted params in `Api::V2::ProvidersController#update` (via `ProviderPolicy`)
- [x] Add `can_sponsor_student_visa` and `can_sponsor_skilled_worker_visa` to `DeserializableProvider` so that are received by the controller.

### Guidance to review

- I had to add the new attributes to the `except` lists for a few other `update_*` methods because they are all called from the same action and leaving as is causes an unsupported param exception.
- I couldn't see an obvious place to put a high-level spec that tests this particular feature other than `spec/features/api/v2/providers/update_spec.rb`. Have I missed something here?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -134,7 +134,7 @@ module API
       end
 
       def update_provider
-        return unless provider_params.values.any?
+        return if provider_params.values.all?(&:nil?)
 
         @provider.assign_attributes(provider_params)
         @provider.save
--- a/app/controllers/api/v2/providers_controller.rb
+++ b/app/controllers/api/v2/providers_controller.rb
@@ -187,6 +187,8 @@ module API
             :send_application_alerts,
             :ukprn,
             :urn,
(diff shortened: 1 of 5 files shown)


davidgisbey (approved):
Yea looks good. Sorry i did this for the degree stuff, but it slipped my mind when i was doing visas

darokel (approved):

PR #1983, opened 2021-06-16
[xxx] Run rubocop in CI

gpeng:
This should need https://github.com/DFE-Digital/teacher-training-api/pull/1982 to be merged before it will pass CI. 

### Context

It wasn't running. It is now.

* run it

### Guidance to review

* Look at the GitHub actions log and see if it ran. https://github.com/DFE-Digital/teacher-training-api/runs/2842537644?check_suite_focus=true (spoilers: it did and failed as we need the PR up there ^ merged)
* In the other apps we specify specific directories that should be linted. I think this belongs in the rubocop config rather than in the build command so I've not done that here. 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,6 +89,9 @@ jobs:
       env:
         GIT_BRANCH: ${{env.BRANCH_TAG}}
 
+    - name: Run Ruby Linter
+      run: docker-compose exec -T web /bin/sh -c 'bundle exec rubocop --format clang'
+
     - name: Run tests
       run: |
         docker-compose exec -T web /bin/sh -c 'bundle config --local disable_exec_load true'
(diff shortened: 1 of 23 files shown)


gpeng:
[Remove rubocop-govuk](https://trello.com/c/GmVrACXA/2011-remove-rubocop-govuk)

defong (changes requested):
Can we move this right at the top, as it literal the fastest no ops.

vigneshmsft (approved):
nice...didn't realise it wasn't being ran.

gpeng:
> Can we move this right at the top, as it literal the fastest no ops.

We can only move it a couple of steps up because we need the docker image to have been built. I do wonder if we should only deploy the review app after the tests have passed though. @vigneshmsft what do you think?

So we'd have:

* build the things
* lint
* test
* deploy (review app or to prod)
* alert

defong (approved):
We need to revisit devops/CI as this is kinda useless.

Other then failing it don't inform us what failed, self failures.

image

PR #1985, opened 2021-06-17
Fix inconsistency between V3 courses search and show endpoints

stevehook:
### Context

This PR attempts to fix a bug in Find. https://trello.com/c/sU1EgzBi/3503-missing-course-on-find

Certain courses that are returned when searching trigger a 404 when the user drills down into them. e.g. Harrow Collegiate Teaching School Alliance has a course called Music (A431) in their search results https://www.find-postgraduate-teacher-training.service.gov.uk/results?l=3&page=2&query=Harrow+Collegiate+Teaching+School+Alliance that leads to a _Page not found_.

After checking the TTAPI database and queries used to implement the search and drill down it looks like this course code is returning a 404 from the TTAPI because its latest (and only) `CourseEnrichment` is in the `withdrawn` state. This triggers the check on this line:
https://github.com/DFE-Digital/teacher-training-api/blob/master/app/controllers/api/v3/courses_controller.rb#L21

Add `published` scope to V3 courses `index` endpoint (`/api/v3/recruitment_cycles/:recruitment_cycle_year/providers/:provider_code/courses`). This makes it consistent with the show action in the same controller so that we don't get into a situation where a course is returned by the search API (index action) but then 404s for the show action.

### Guidance to review

- Does the change to the query make sense? Does it make the two actions consistent?
- Is there a better way to test this?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/api/v3/courses_controller.rb
+++ b/app/controllers/api/v3/courses_controller.rb
@@ -6,7 +6,11 @@ module API
       before_action :build_courses
 
       def index
-        course_search = CourseSearchService.call(filter: params[:filter], sort: params[:sort], course_scope: @courses)
+        course_search = CourseSearchService.call(
+          filter: params[:filter],
+          sort: params[:sort],
+          course_scope: @courses.published,
+        )
 
         render jsonapi: paginate(course_search),
                fields: fields_param,
--- a/spec/requests/api/v3/courses/index_spec.rb
+++ b/spec/requests/api/v3/courses/index_spec.rb
@@ -433,7 +433,7 @@ describe "GET v3/courses" do
       let(:request_path) { "/api/v3/courses?filter[subjects]=A1,B1" }
       let(:course_with_B1_subject) do
         create(:course,
-               enrichments: [published_enrichment],
+               enrichments: [build(:course_enrichment, :published)],
                site_statuses: [build(:site_status, :findable)],
                subjects: [create(:primary_subject, subject_code: "B1")])
       end
@@ -516,7 +516,7 @@ describe "GET v3/courses" do
                provider: provider_filtered_by,
                accrediting_provider: provider_filtered_by,
                site_statuses: [create(:site_status, :findable, site: site1)],
-               enrichments: [published_enrichment])
+               enrichments: [build(:course_enrichment, :published)])
       }
       let(:another_provider_course) {
         create(:course,
@@ -524,7 +524,7 @@ describe "GET v3/courses" do
                provider: another_training_provider,
                accrediting_provider: provider_filtered_by,
                site_statuses: [create(:site_status, :findable, site: site2)],
-               enrichments: [published_enrichment])
+               enrichments: [build(:course_enrichment, :published)])
       }
 
       before do
--- a/spec/requests/api/v3/providers/courses/index_spec.rb
+++ b/spec/requests/api/v3/providers/courses/index_spec.rb
@@ -225,7 +225,7 @@ describe "GET v3/recruitment_cycle/:recruitment_cycle_year/providers/:provider_c
           json_response = JSON.parse response.body
           expect(json_response["data"].count).to eq 1
           expect(json_response["data"].first)
-            .to have_attribute("name").with_value('Mathematics')
+            .to have_attribute("name").with_value("Mathematics")
         end
       end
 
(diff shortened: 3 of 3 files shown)


app/controllers/api/v3/courses_controller.rb
```diff
@@ -6,7 +6,11 @@ class CoursesController < API::V3::ApplicationController
       before_action :build_courses
 
       def index
-        course_search = CourseSearchService.call(filter: params[:filter], sort: params[:sort], course_scope: @courses)
+        course_search = CourseSearchService.call(
+          filter: params[:filter],
+          sort: params[:sort],
+          course_scope: @courses.published,
```
defong, line 12: Probably want to move that in to `CourseSearchService` defaulting with published `true`

defong (commented):
How many course does this issue affects?

gpeng (approved):
This looks good to me. I may not have been understanding what @defong was suggesting though.

app/controllers/api/v3/courses_controller.rb
stevehook: @defong & @gpeng encapsulating that extra condition in `CourseSearchService` would be neater but it looks like it would change behaviour, e.g. these two endpoints could get out of step https://github.com/DFE-Digital/teacher-training-api/blob/master/app/controllers/api/public/v1/providers/courses_controller.rb#L6-L16

I'll try to work out how many courses this issue applies to...

stevehook: It looks like there are around 246 courses in the current cycle that only have a withdrawn CourseEnrichment. I don't know how many of those are searchable.

defong: https://api.publish-teacher-training-courses.service.gov.uk/api-reference.html#schema-coursefilter

That should of covered us, as it purely a bug.

image

See this just suggest to me, that we haven't understood the root cause for the bug, as `findable` was a long establish meaning of being on `find` .
https://github.com/DFE-Digital/teacher-training-api/blob/5f2bbefe7a76c35b007396d0b86720811251a7a3/app/models/course.rb#L526-L531

stevehook: @defong I'm a bit confused I must admit. Having `findable` as the single determining factor would be nice but it seems more complicated than that. 

https://github.com/DFE-Digital/teacher-training-api/blob/master/app/controllers/api/v3/courses_controller.rb#L21 is filtering out courses unrelated to `findable`. Trouble is that check is only done on the `show` action, not on the `index`.

stevehook: @defong So do you think this is a data problem rather than something needing a code change?

dankmitchell:
@stevehook cc. @defong @gpeng 

This statement is the intended bevahior -

> After checking the TTAPI database and queries used to implement the search and drill down it looks like this course code is returning a 404 from the TTAPI because its latest (and only) CourseEnrichment is in the withdrawn state. This triggers the check on this line:
https://github.com/DFE-Digital/teacher-training-api/blob/master/app/controllers/api/v3/courses_controller.rb#L21

Withdrawing a course is UCAS Apply terminology for deleting since we can't remove courses from the V1 API after they have been sent so we had to create a new state of "withdrawn".

Post-UCAS I'd just like to trigger "discard" i.e. soft delete for this state.

In general "state" is something that needs refactoring to avoid all the confusion around findable, published etc.

Alot of clean up work todo post-UCAS.

app/controllers/api/v3/courses_controller.rb
defong: I think we need to understand how many course this has this issues, it maybe an more feasible to fix as a data issue.

Otherwise we do the code fix and follow it through in order to deem the accuracy of https://api.publish-teacher-training-courses.service.gov.uk/reporting

stevehook: @defong it looks like there are only two such courses in the current cycle.

Thats based on some rough-and-ready queries I just did: Provider 5A1 course G011 and provider 1EF course A431. (I ran a query that collects all the 'withdrawn' courses, defined as those that just have a withdrawn enrichment, and intersecting them with the results of the `findable` scope).

What do you think is the right thing to do here?

defong: Yeah I think fixing the underling data should be the path of least resistance.

stevehook:
As discussed with @defong on Slack, closing this PR in favour of a data fix.

defong:
Carded as https://trello.com/c/IrcA6ay4/2201-withdrawn-courses-needs-to-sites-with-no-vacancies-and-be-suspended

PR #1977, opened 2021-06-14
Add Structured GCSE Columns to course table

george-bit:
### Context

As part of a redesign on the front end of publish additonal values are required to capture
GCSE reqs in a structured format. 

https://trello.com/c/xCks76yv/3498-add-structured-gcse-requirements-fields-to-the-v2-ttapi

### Changes proposed in this pull request

This PR adds a migration for these values.

### Guidance to review

Do these names make sense?

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #1986, opened 2021-06-17
Exposing the structured GCSE fields on the V2 API

george-bit:
### Context

Carrying on from the attributes added on #1977  we need to expose these attributes on the V2 TTAPI for use in Publish.

This PR exposes the attributes and adds specs to cover this behaviour.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/factories/courses.rb
```diff
@@ -26,6 +26,13 @@
     additional_degree_subject_requirements { true }
     degree_subject_requirements { Faker::Lorem.sentence.to_s }
 
+    accept_pending_gcse { true }
+    accept_gcse_equivalency { true }
+    accept_english_gcse_equivalency { true }
+    accept_maths_gcse_equivalency { false }
+    accept_science_gcse_equivalency { false }
```
defong, line 33: Is there a particular reaon if not can we user `[true, false].sample`

defong (changes requested):
Failing test? Otherwise I think its fine

spec/factories/courses.rb
george-bit: My bad, added the factory in last so didn't check to see if it broke anything. Will fix as suggested 👍

george-bit: okay so had a bit of trouble testing the output of  `[true, false].sample` in the stubbed API response. Therefore, I've packaged these attributes as a trait on the courses factory. Is this okay?

defong (approved):

PR #1990, opened 2021-06-18
[2031] Add debug info to DuplicateUserError

FriedSock:
### Context
https://trello.com/c/KMR5Tpjc/2031-add-debug-info-to-error-message-for-error-duplicateusererror

It is impossible to debug this without more information.

Sentry just prints out the error message for the exception, (we could
also use something called raven_context to add some json to the 
exception, but that feature is undocumented so this feels safer). It
would be nice to have the email address as well, but this would be 
filtered out by our sentry config for PII reasons.

Adds all the attributes on DuplicateUserError to the message

### Guidance to review

Not easy to test this with sentry as we disable it on the review apps. If you want to see the error message in action, just raise one manually in the console. Or if you really want to have a fun Friday afternoon, set up your own trial sentry account, set the dsn on your local app and manually send the exception using `Raven.capture_exception`

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/services/authentication_service/duplicate_user_error.rb
+++ b/app/services/authentication_service/duplicate_user_error.rb
@@ -7,6 +7,13 @@ class AuthenticationService
       @existing_user_id              = existing_user_id
       @existing_user_sign_in_user_id = existing_user_sign_in_user_id
 
+      message += <<~DEBUG_INFO.insert(0, "\n")
+        user_id: #{@user_id},
+        user_sign_in_user_id: #{@user_sign_in_user_id},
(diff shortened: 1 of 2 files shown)


FriedSock:
[Add debug info to error message for error DuplicateUserError](https://trello.com/c/KMR5Tpjc/2031-add-debug-info-to-error-message-for-error-duplicateusererror)

AbigailMcP (approved):

PR #2002, opened 2021-06-24
[1976] Add EmailAddressValidator, from earlier EmailValidator

archferns:
### Context
Previously we used a validator defined as [EmailValidator](https://github.com/DFE-Digital/teacher-training-api/blob/1e02fc22d7f64e696b5fe1c9d6cd6dc78c93d444/app/validators/email_validator.rb),
which had to be discarded when we switched to usng DfE sign in,
which has a dependency on `validates_email` gem via `omniauth_openid_connect` gem.

However, validates_email relies on [Mail::Address](https://www.rubydoc.info/gems/mail/Mail/Address#initialize-instance_method) to validate addresses,
which isn't optimal for our use, as 'Person Example <person@example.com>' would be a
valid email address.

Hence to get around the clashing implementation, this commit renames
EmailValidator to EmailAddressValidator.

Additionally, it picks up the implementation directly from the [rails
guides](https://guides.rubyonrails.org/active_record_validations.html#custom-validators)
and adds a few test cases matching malformed data found among our
providers.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


archferns:
[[L] Fix validation on provider email](https://trello.com/c/dYKWlDZq/1976-l-fix-validation-on-provider-email)

defong (approved):

PR #2016, opened 2021-06-29
[2093] Added validation to the course model for site urn

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/WR5JVk06/2093-m-users-get-an-error-message-when-they-try-to-publish-a-course-if-there-are-sites-on-the-course-without-urn)

- Added validation to the Course Model, that checks that all sites have URNs

### Guidance to review

- Create a new course in the 2022 cycle, and make sure that at least one connected site does not have a URN

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -266,6 +266,7 @@ class Course < ApplicationRecord
   validate :validate_site_statuses_publishable, on: :publish
   validate :validate_provider_visa_sponsorship_publishable, on: :publish, if: -> { recruitment_cycle.after_2021? }
   validate :validate_provider_urn_ukprn_publishable, on: :publish, if: -> { recruitment_cycle.after_2021? }
+  validate :validate_all_sites_publishable, on: :publish, if: -> { recruitment_cycle.after_2021? }
   validate :validate_degree_requirements_publishable, on: :publish
   validate :validate_enrichment
   validate :validate_qualification, on: %i[update new]
@@ -831,6 +832,12 @@ private
     end
     @services.register(:content_status) do
       Courses::ContentStatusService.new
+    end
+  end
+
+  def validate_all_sites_publishable
+    if sites.any? { |s| s.urn.blank? }
+      errors.add(:sites, :site_urn_not_publishable)
     end
   end
 end
--- a/spec/controllers/api/v2/courses_controller_spec.rb
+++ b/spec/controllers/api/v2/courses_controller_spec.rb
@@ -4,38 +4,34 @@ describe API::V2::CoursesController, type: :controller do
   let(:site_status) { build(:site_status, :new) }
   let(:dfe_subject) { find_or_create(:primary_subject, :primary) }
   let(:provider) { create(:provider) }
-  let(:course) {
+  let(:course) do
     create(:course,
            site_statuses: [site_status],
            enrichments: [enrichment],
            subjects: [dfe_subject],
            provider: provider)
-  }
+  end
   let(:email) { "manage_courses@digital.education.gov.uk" }
   let(:sign_in_user_id) { "manage_courses_api" }
-  let(:existing_user) do
-    create(
-      :user, admin: true,
-             email: email,
-             sign_in_user_id: sign_in_user_id
-    )
-  end
+  let(:existing_user) { create(:user, admin: true, email: email, sign_in_user_id: sign_in_user_id) }
 
   before do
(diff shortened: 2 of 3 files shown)


stellenberger:
[[M] Users get an error message when they try to publish a course if there are sites on the course without URN](https://trello.com/c/WR5JVk06/2093-m-users-get-an-error-message-when-they-try-to-publish-a-course-if-there-are-sites-on-the-course-without-urn)

defong (changes requested):
Does not work.
no_urn

app/models/course.rb
```diff
@@ -812,4 +813,10 @@ def services
       Courses::ContentStatusService.new
     end
   end
+
+  def validate_all_sites_publishable
+    if sites.map(&:urn).include?(nil)
+      errors.add(:sites, "must have a Unique Reference Number (URN)")
```
darokel, line 819: Screenshot 2021-06-30 at 15 08 19
 I think this message needs to be set in the translation similar to https://github.com/DFE-Digital/teacher-training-api/pull/2022/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7R123

stellenberger: All done 👍

spec/controllers/api/v2/courses_controller_spec.rb
```diff
@@ -39,23 +39,56 @@
...
+
+        it "returns a validation error if any site does not have a URN" do
+          expect(NotificationService::CoursePublished).not_to receive(:call).with(course: course)
+          post :publish, params: {
+            recruitment_cycle_year: provider.recruitment_cycle.year,
+            provider_code: course.provider.provider_code,
+            code: course.course_code,
+          }
+          expect(response).to have_http_status(:unprocessable_entity)
+          expect(response.body).to match(/You must provide a Unique Reference Number/)
```
defong, line 90: Can we match the whole thing?

app/models/course.rb
```diff
@@ -261,6 +261,7 @@ def self.entry_requirement_options_without_nil_choice
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_site_statuses_publishable, on: :publish
   validate :validate_provider_visa_sponsorship_publishable, on: :publish
+  validate :validate_all_sites_publishable, on: :publish, if: -> { recruitment_cycle.year.to_i >= 2022 }
```
defong, line 264: Can we take this approach and refactor it along the lines of `COURSE_2022_REQUIREMENTS_REQUIRED_FROM`

https://github.com/DFE-Digital/teacher-training-api/blob/master/app/models/course.rb#L58
```suggestion
  validate :validate_all_sites_publishable, on: :publish, if: -> { recruitment_cycle.year.to_i >= COURSE_2022_REQUIREMENTS_REQUIRED_FROM }
```

To inform future selves, otherwise we end up with pepper shotgun with `202X` over the years

spec/controllers/api/v2/courses_controller_spec.rb
stellenberger: Done 👍

app/models/course.rb
stellenberger: done 👍

app/models/course.rb
```diff
@@ -812,4 +814,10 @@ def services
       Courses::ContentStatusService.new
     end
   end
+
+  def validate_all_sites_publishable
+    if sites.map(&:urn).include?(nil || "")
```
defong, line 819:
```suggestion
    if sites.any?() {|s| s.urn.blank?}
```

defong (approved):

app/models/course.rb
stellenberger: done 👍

app/models/course.rb
```diff
@@ -56,6 +56,7 @@ class Course < ApplicationRecord
   }.freeze
 
   DEGREE_REQUIREMENTS_REQUIRED_FROM = 2022
+  COURSE_2022_REQUIREMENTS_REQUIRED_FROM = 2022
```
jhenderson, line 59: I don't think the name of the constant makes any sense. Don't worry, I tired something similar and failed to come up with something meaningful. In the end, I just use the integer 2022 directly.

stellenberger: Reverted changes to just the integer 👍

jhenderson (approved):

PR #2018, opened 2021-06-29
Add Course#gcse_grade_required and expose in V2

duncanjbrown:
### Context

Reviewing https://github.com/DFE-Digital/publish-teacher-training/pull/1773/commits/90273c157e810fdfc9e783c18daf922672bfacd1 we decided that it was better if facts about providers lived down here in the API rather than up in Publish (and potentially then in Find/Apply/Register/etc etc).

We think the case where a provider has different GCSE requirements is extremely rare, so as a first cut this PR uses a list in a constant to put either a `4` or a `5` in a new `gcse_grade_required` field in the V2 API. Publish will consume this and render appropriate content.

https://trello.com/c/oLvRx47b/3497-add-the-structured-gcse-requirements-section-to-publish

### Checklist

- [ ] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history


app/models/course.rb
```diff
@@ -57,6 +57,10 @@ class Course < ApplicationRecord
 
   DEGREE_REQUIREMENTS_REQUIRED_FROM = 2022
 
+  # Most providers require GCSE grade 4 ("C"),
+  # but some require grade 5 ("strong C")
+  PROVIDERS_REQUIRING_GCSE_GRADE_5 = %w[U80 I30].freeze
```
defong, line 62: Will we need to consider this to be in the database at a later point in time?

Or are they just the only 2 out of 1100?

defong (commented):
Looks fine to me, through it breaks the normal as its the only thing that's hardcoded.

Deferring to @gpeng

app/models/course.rb
gpeng: I suppose this is pragmatic while we only have 2 🤷

gpeng (approved):

PR #2025, opened 2021-07-01
[2091] Publish locations allow providers to create more than 37 locations

archferns:
### Context
Providers will be allowed to add more than 37 locations (sites) next cycle. We should remove the validations enforcing this allowing them to do so.

1. The first 37 sites will continue to have a UCAS style code, i.e(one of A-Z, 0-9 or -), after which the next set of locations will have a sequential code from "AA", "AB" and so on.
2. v1 of the API will only fetch 37 sites, which have the UCAS code.
3. v2, v3 and public v1 will show all available sites.

### Checklist
- [x] Remove https://github.com/DFE-Digital/teacher-training-api/pull/2025/commits/7c4fcb4aa5bba5b24b3b247944253e3a9611d46d commit
- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -234,7 +234,7 @@ class Provider < ApplicationRecord
   end
 
   def can_add_more_sites?
-    sites.size < Site::POSSIBLE_CODES.size
+    true
   end
 
   # This reflects the fact that organisations should actually be a has_one.
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -229,10 +229,6 @@ class Provider < ApplicationRecord
     update_columns changed_at: timestamp
   end
 
-  def unassigned_site_codes
-    Site::POSSIBLE_CODES - sites.pluck(:code)
-  end
-
   def can_add_more_sites?
     true
   end
--- a/app/services/sites/code_generator.rb
+++ b/app/services/sites/code_generator.rb
@@ -1,0 +1,44 @@
+module Sites
+  class CodeGenerator
+    include ServicePattern
+
+    def initialize(provider:)
+      @provider = provider
+    end
+
+    def call
+      ucas_style_code.presence || highest_site_code.next
+    end
+
+  private
+
+    attr_reader :provider
+
+    def highest_site_code
+      return "Z" if existing_sequential_codes.blank?
+
+      existing_sequential_codes.max
+    end
+
+    def existing_sequential_codes
+      existing_site_codes - Site::POSSIBLE_CODES
+    end
+
+    def existing_site_codes
+      @existing_site_codes ||= provider.sites.pluck(:code)
+    end
+
+    def unassigned_ucas_style_site_codes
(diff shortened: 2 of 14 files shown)


archferns:
[[L] Publish locations - Allow providers to create more than 37 locations](https://trello.com/c/8RzgyjQF/2091-l-publish-locations-allow-providers-to-create-more-than-37-locations)

app/services/sites/code_generator.rb
```diff
@@ -0,0 +1,44 @@
...
+      Site::POSSIBLE_CODES - existing_site_codes
+    end
+
+    def ucas_style_code
+      @ucas_style_code ||= begin
+        available_desirable_codes = unassigned_ucas_style_site_codes & Site::DESIRABLE_CODES
+        available_undesirable_codes = unassigned_ucas_style_site_codes & Site::EASILY_CONFUSED_CODES
+
+        available_desirable_codes.sample || available_undesirable_codes.sample
+      end
```
archferns, line 41: Left in the existing code generation setup, and only sites created after these codes are used up, will be using sequential code generation

.gitignore
```diff
@@ -69,3 +69,5 @@ lock.json
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
+
+spec/examples.txt
```
gpeng, line 73: I don't think we need this.

app/models/provider.rb
```diff
@@ -229,12 +229,8 @@ def update_changed_at(timestamp: Time.now.utc)
     update_columns changed_at: timestamp
   end
 
-  def unassigned_site_codes
-    Site::POSSIBLE_CODES - sites.pluck(:code)
-  end
-
   def can_add_more_sites?
```
gpeng, line 232: Could we just remove this altogether if it's always true?

gpeng (changes requested):
Why have we included `-` in the codes?

gpeng:
> Why have we included `-` in the codes?

Ah... It was included in the UCAS format. 👍

app/models/provider.rb
archferns: it is being used in the v2 provider serializer, should we move it over there instead?
https://github.com/DFE-Digital/teacher-training-api/blob/c455e1f36bb1f46172d89790b733e82a94e055b8/app/serializers/api/v2/serializable_provider.rb#L6

.gitignore
archferns: Removed.

app/models/provider.rb
gpeng: Ok. We'll need to change publish to get rid of it in that case. Let's leave it for now and we'll do it in the tidy up.

gpeng (approved):

app/models/provider.rb
archferns: Carded up https://trello.com/c/EQEGSnMD/1664-remove-canaddmoresites-from-ttapi

PR #2026, opened 2021-07-01
Add legacy support for 'required_qualifications' attribute in public API

malcolmbaig:
### Context
We've updated course creation in Publish to take structured data for
required degree and GCSE information. This will eventually replace the
free text `required_qualifications` attribute.

To maintain the existing interface of the V1 public API, we need to
transform the structured data into a value that can be returned in the
`required_qualifications` field in scenarios where a Course has no value
for that attribute.

https://trello.com/c/LR2VwtVq

Do this by adding a RequiredQualificationsSummary object that's invoked
via the Course model. This object:

- Takes an instance of Course as an argument
- Looks for the `required_qualifications` enrichment attribute, and uses
  it if it has a value
- Otherwise generates summary text from the qualifications-specific
  attributes on Course

### Guidance to review
- Can be manually tested locally by creating a course in Publish and viewing it via the V1 public API. Could do this for both courses created with the free text field and the new structured qualifications form.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/required_qualifications_summary.rb
+++ b/app/models/required_qualifications_summary.rb
@@ -23,6 +23,11 @@ private
     output << gcse_equivalency_content
     output << "\n"
     output << (course.additional_gcse_equivalencies || "")
+    output << "\n"
+    output << degree_grade_content
+    output << "\n"
+    output << degree_subject_requirements_content
+    output << "\n"
   end
 
   def required_gcse_content
@@ -65,5 +70,23 @@ private
       maths: course.accept_maths_gcse_equivalency?,
       science: course.accept_science_gcse_equivalency?,
     }.select { |_k, v| v }.keys
+  end
+
+  def degree_grade_content
+    {
+      "two_one" => "An undergraduate degree at class 2:1 or above, or equivalent.",
+      "two_two" => "An undergraduate degree at class 2:2 or above, or equivalent.",
+      "third_class" => "An undergraduate degree, or equivalent. This should be an honours degree (Third or above), or equivalent.",
+      "not_required" => "An undergraduate degree, or equivalent.",
+      nil => "",
+    }[course.degree_grade]
+  end
+
+  def degree_subject_requirements_content
+    if course.additional_degree_subject_requirements?
+      course.degree_subject_requirements
+    else
+      ""
+    end
   end
 end
--- a/app/models/required_qualifications_summary.rb
+++ b/app/models/required_qualifications_summary.rb
@@ -1,0 +1,69 @@
+class RequiredQualificationsSummary
+  attr_reader :course
+
+  def initialize(course)
+    @course = course
+  end
+
+  def extract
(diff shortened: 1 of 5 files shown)


defong (commented):
I think it needs to be surfaced here 
https://teacher-training-api-pr-2026.london.cloudapps.digital/api-reference.html#schema-courseattributes

image

app/models/required_qualifications_summary.rb
```diff
@@ -0,0 +1,88 @@
+class RequiredQualificationsSummary
+  attr_reader :course
+
+  def initialize(course)
+    @course = course
+  end
+
+  def extract
```
defong, line 8: Can we use that? 
app/services/service_pattern.rb

Not fussed

app/models/required_qualifications_summary.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def required_gcse_content
+    case course.level
+    when "primary"
+      "Grade 4 (C) or above in English, maths and science, or equivalent qualification."
+    when "secondary"
+      "Grade 4 (C) or above in English and maths, or equivalent qualification."
+    else
+      ""
+    end
```
defong, line 37: https://github.com/DFE-Digital/publish-teacher-training/blob/master/app/components/gcse_requirements_component.rb#L10-L17
https://github.com/DFE-Digital/teacher-training-api/blob/master/app/models/course.rb#L439-L445

defong (dismissed):
Inconsistent `required_gcse_content`

malcolmbaig:
> I think it needs to be surfaced here
> https://teacher-training-api-pr-2026.london.cloudapps.digital/api-reference.html#schema-courseattributes
> 
> image

@defong I think that description for `required_qualifications` still holds. Do you think we should elaborate? Maybe something like:

> The minimum academic qualifications needed for this course. Uses the free-text value of the legacy `required_qualifications` attribute if present, otherwise returns a text summary based on the various GCSE/degree attributes of the course.

defong:
> > I think it needs to be surfaced here
> > https://teacher-training-api-pr-2026.london.cloudapps.digital/api-reference.html#schema-courseattributes
> > image
> 
> @defong I think that description for `required_qualifications` still holds. Do you think we should elaborate? Maybe something like:
> 
> > The minimum academic qualifications needed for this course. Uses the free-text value of the legacy `required_qualifications` attribute if present, otherwise returns a text summary based on the various GCSE/degree attributes of the course.

I was thinking about the named subjects, English, maths, and science, may need visiting as well. 🤔


Where do we stand on this documentation of things @gpeng

malcolmbaig:
@defong This PR doesn't interact with those subject fields. Happy to pick up any documentation updates we identify, but perhaps any changes related to those fields are outside the scope of this PR?

gpeng (approved):

PR #2027, opened 2021-07-01
Appease rubocop 1.18.1

archferns:
Fixed the issues raised by the updated version of rubocop.
Layout/LineEndStringConcatenationIndentation was unable to cope with
autocorrect, so that has been dealt by replacing the curly percent
literal with heredocs, in a migration and the provider model.


malcolmbaig (approved):
👍🏼

PR #2031, opened 2021-07-02
Remove validation for CourseEnrichment#required_qualifications from 2022

stevehook:
### Context

This validation for the presence of `CourseEnrichment#required_qualifications` should not be active after the 2021 recruitment cycle. It has been replaced by new attributes.

https://trello.com/c/bCnjk5oa/3561-remove-the-qualifications-needed-section-from-the-course-page-on-publish-for-2022-courses

- Make the `presence` validation conditional on recruitment cycle and update specs.

### Guidance to review

- Does this validation make sense?
- Can be manually tested by finding a draft 2022 course, setting it's _Qualifications needed_ to blank and attempting to publish.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/models/course_enrichment_spec.rb
+++ b/spec/models/course_enrichment_spec.rb
@@ -140,8 +140,11 @@ describe CourseEnrichment, type: :model do
 
   describe "required_qualifications attribute" do
     let(:required_qualifications_text) { "this course is great" }
+    let(:recruitment_cycle) { build(:recruitment_cycle, year: "2021") }
+    let(:provider) { build(:provider, recruitment_cycle: recruitment_cycle) }
+    let(:course) { build(:course, provider: provider) }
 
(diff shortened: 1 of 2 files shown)


spec/models/course_enrichment_spec.rb
```diff
@@ -215,7 +226,6 @@
 
     context "fee based course" do
       it { is_expected.to validate_presence_of(:fee_uk_eu).on(:publish) }
-      it { is_expected.to validate_presence_of(:required_qualifications).on(:publish) }
```
stevehook, line 218: As far as I can see the validation doesn't depend on the course being fee/salary-based so just removing this so that it doesn't break in 2022. Does that make sense?

gpeng (approved):

PR #2041, opened 2021-07-06
Remove validation for UCAS entry requirements for 2022 cycles courses

davidgisbey:
### Context

We shouldn't be validating that maths, english and science UCAS equivalency sections have been completed for rolled over courses and it's being deprecated from the next cycle onwards.

- Ensure that we are not validating the maths, english and science fields for courses in the 2022 cycle

### Guidance to review

There were some issues with tests failing due to race conditions. John very kindly added the 2nd commit that fixed them.


defong (approved):
I look forward to gutting this from our codebase

PR #2081, opened 2021-07-29
Bug fix - order courses correctly when ordering by distance and filtering by  'has_vacancies'

TobyRet:
### Context
Previously the `has_vacancies` filter was not taken into account when ordering by distance. This meant that some searches on Find would results in courses appearing higher up in search results even though they had sites with no vacancies

Squash that bug!

### Guidance to review
Make sense? [You can see a live example of the bug on prod](https://www.find-postgraduate-teacher-training.service.gov.uk/results?c=England&l=1&lat=51.4612794&lng=-0.1156148&loc=Brixton%2C+London%2C+UK&lq=Brixton%2C+London&rad=50&sortby=2&subjects%5B%5D=24)

|Find with bug|Find without the bug|
--- | --- |
|before_find_no_vacancies|find_after|

### Trello
https://trello.com/c/Vo3H6vAP/3453-find-locations-bug

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -91,6 +91,18 @@ private
     # Only running and published site statuses
     running_and_published_criteria = site_status[:status].eq(SiteStatus.statuses[:running]).and(site_status[:publish].eq(SiteStatus.publishes[:published]))
 
+    running_and_published_and_vacancy_criteria = if has_vacancies?
+                                                   running_and_published_criteria
+                                                     .and(site_status[:vac_status])
+                                                     .eq_any([
+                                                       SiteStatus.vac_statuses[:full_time_vacancies],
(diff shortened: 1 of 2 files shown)


app/services/course_search_service.rb
```diff
@@ -91,6 +91,18 @@ def locatable_sites
     # Only running and published site statuses
     running_and_published_criteria = site_status[:status].eq(SiteStatus.statuses[:running]).and(site_status[:publish].eq(SiteStatus.publishes[:published]))
 
+    running_and_published_and_vacancy_criteria = if has_vacancies?
```
darokel, line 94: Screenshot 2021-07-29 at 18 14 54

Can we extract method here so we don't add to the complaint above?

darokel (approved):

PR #2091, opened 2021-08-06
[2257] Send Big Query events

gpeng:
### Context

Sending web requests from TTAPI. Largely based on @FriedSock 's PR on publish https://github.com/DFE-Digital/publish-teacher-training/pull/1761

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/lib/big_query/request_event.rb
```diff
@@ -0,0 +1,52 @@
+module BigQuery
+  class RequestEvent
+    def initialize
+      @event_hash = {
+        environment: ENV["RAILS_ENV"],
+        occurred_at: Time.zone.now.iso8601,
+        event_type: "web_request",
+      }
```
darokel, line 8: is it worth `yielding self if block_given?` then you could do:

```
request_event = BigQuery::RequestEvent.new |event|
  event.with_request_details(request)
  event.with_response_details(response)
  event.with_user(current_user)
end
```

Just wondering if it'll help avoid always having to return the instance for each method definition?

darokel:
[Send BigQuery events from TTAPI](https://trello.com/c/t46M5T61/2257-send-bigquery-events-from-ttapi)

app/lib/big_query/request_event.rb
gpeng: I've updated with that suggestion @darokel 👍

darokel (approved):
Left one more cosmetic suggestion but all good

app/lib/big_query/request_event.rb
```diff
@@ -0,0 +1,53 @@
...
+      event_hash.merge!(
+        request_uuid: request.uuid,
+        request_path: request.path,
+        request_method: request.method,
+        request_user_agent: request.user_agent,
+        request_query: query_to_kv_pairs(request.query_string),
+        request_referer: request.referer,
+      )
+
+      self
```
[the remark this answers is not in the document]
gpeng: whoops. Yes, they can.

app/lib/big_query/request_event.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def with_request_details(request)
+      event_hash.merge!(
+        request_uuid: request.uuid,
+        request_path: request.path,
+        request_method: request.method,
+        request_user_agent: request.user_agent,
+        request_query: query_to_kv_pairs(request.query_string),
+        request_referer: request.referer,
+      )
```
misaka, line 22: For the other services we've also been adding `anonymised_user_agent_and_ip` for ... well ... an anonymised version of the user agent and IP. I know this is less useful for the API, but it could still be interesting and it would be more consistent. Worth adding?

gpeng: Can do

gpeng: I've updated to include that now @misaka if you could have another 👀

misaka: 👍

misaka (approved):
Events coming into BigQuery. LGTM! 🚢

PR #2092, opened 2021-08-06
Ensure courses UCL accredit require 5+ for GCSE requirements

davidgisbey:
### Context

UCL require 5+ for GCSEs for the courses they accredit.

Currently, we only check whether they are the provider. This has caused issues as a provider has noted that it says 4+ for their courses accredited by UCL.

https://trello.com/c/jYTgdnjE/3778-fix-issue-with-courses-ucl-accredit-not-showing-5-for-gcses

- Check if UCL is the accredited provider in the `gcse_grade_required` method. If so also return 5.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -442,7 +442,7 @@ class Course < ApplicationRecord
   end
 
   def gcse_grade_required
-    if PROVIDERS_REQUIRING_GCSE_GRADE_5.include?(provider_code)
+    if PROVIDERS_REQUIRING_GCSE_GRADE_5.any?(provider_code) || PROVIDERS_REQUIRING_GCSE_GRADE_5.any?(accrediting_provider&.provider_code)
       5
     else
       4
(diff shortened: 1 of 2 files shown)


app/models/course.rb
```diff
@@ -442,7 +442,7 @@ def gcse_science_required?
   end
 
   def gcse_grade_required
-    if PROVIDERS_REQUIRING_GCSE_GRADE_5.include?(provider_code)
+    if PROVIDERS_REQUIRING_GCSE_GRADE_5.include?(provider_code) || PROVIDERS_REQUIRING_GCSE_GRADE_5.include?(accrediting_provider&.provider_code)
```
TobyRet, line 445: Could potentially use `#any?` to make this more succinct?

TobyRet (approved):
Small comment otherwise LGTM

defong (approved):

app/models/course.rb
davidgisbey: have made the change

PR #2119, opened 2021-08-20
[2540] Course#updated_changed_at touches Provider

gpeng:
### Context

When a course has its vacancy status changed we expect the provider to be returned in a /providers API call with an updated_since filter. They currently are not as the `Provider#changed_at` is not updated when the `Course#site_statuses` are updated.

`touch_provider` when the `Course#update_changed_at` is called. `update_columns` skips callbacks so we need to call `touch_provider` explicitly.

### Guidance to review

Retrieve a course with vacancies in the console: `course = Course.with_vacancies.last`
Update the vacancy status: `course.site_statuses.each(&:no_vacancies!)`
Check the changed at on Course: `course.changed_at`
Check the changed at on Provider: `course.provider.changed_at`

They should match (pretty much). This provider should now appear on `/api/public/v1/recruitment_cycles/2021/providers?updated_since=<a bit ago>`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -378,6 +378,7 @@ class Course < ApplicationRecord
     # itself, so we don't want to alter the semantics of updated_at which
     # represents changes to just the course record.
     update_columns changed_at: timestamp
+    touch_provider
   end
 
   def study_mode_description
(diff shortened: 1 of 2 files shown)


gpeng:
[Provider endpoint updated_since bug](https://trello.com/c/32IIvr6L/2540-provider-endpoint-updatedsince-bug)

cwrw (approved):

PR #2124, opened 2021-08-23
[2256] Add BigQuery event update and event create events

gpeng:
### Context

We are sending events to BigQuery for analytics. We want to send create/update events for configured models.

* Add analytics config
* Add EmitsEntityEvents concern
* include it in ApplicationModel

### Guidance to review

* run the app
* create/update Provider in the console
* Inspect the sidekiq queue. Each action should create a job with a json payload

The analytics.yml is currently 'skeletal' awaiting some guidance on what we need to send.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/lib/big_query/entity_event.rb
```diff
@@ -0,0 +1,49 @@
+module BigQuery
+  class EntityEvent
+    EVENT_TYPES = %w[create_entity update_entity].freeze
+
+    def initialize
+      @event_hash = {
+        environment: ENV["RAILS_ENV"],
```
misaka, line 7: Just out of curiousity, is there a reason not to use `Rails.env`? Is it considered harmful in any way? Or just too new/a fad?

app/models/concerns/emits_entity_events.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+  def send_event(type, data)
+    return unless FeatureService.enabled?(:send_request_data_to_bigquery)
+
+    event = BigQuery::EntityEvent.new do |ee|
+      ee.with_type(type)
+      ee.with_entity_table_name(self.class.table_name)
+      ee.with_data(data)
```
misaka, line 31: Where does the request ID get set? I think we saw that Apply wasn't setting it so if this is copied from there ...

misaka:
We should test this out locally with the `publish_api_events_test` dataset.

app/models/concerns/emits_entity_events.rb
misaka: We discussed the challenge around this briefly. Publish uses [request_store](https://github.com/steveklabnik/request_store) to do this ([setting](https://github.com/DFE-Digital/publish-teacher-training/blob/42ab166f600a51a45ce8506faf42c080e0213ae5/app/controllers/application_controller.rb#L291), [retrieving](https://github.com/DFE-Digital/publish-teacher-training/blob/e7a7a1a420f885f6645f9ea7b50906a4c7a94fa7/app/models/base.rb#L15)). Which reminds me, we should probably be setting this in the headers of API calls to Apply and PAPI too.

gpeng: Should we do that as another PR so we can start getting the events in?

gpeng:
> We should test this out locally with the `publish_api_events_test` dataset.

I've sent a provider update and create to the test dataset. Looks to have made it ok.

app/models/concerns/emits_entity_events.rb
misaka: If you prefer ... so long as it doesn't get punted, I'd say.

PR #2133, opened 2021-08-27
[2392] Import existing db entities into BigQuery

gpeng:
### Context

We are sending create and update events to BigQuery.

* Add a `EmitsEntityEvents#send_import_event` - method to send an import event for a model
* Add config/analytics.yml which specifies the entities (db table names) and the fields that should be included in the entity event (generally, excluding PII)
* Add a rake task that will loop through the entities in the config file and queue up a job to send the entity.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -1,6 +1,7 @@
 ---
 :queues:
   - default
+  - mailers
   - geocoding
   - save_statistic
-  - mailers
+  - low_priority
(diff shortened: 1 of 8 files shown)


config/sidekiq.yml
```diff
@@ -1,6 +1,7 @@
 ---
 :queues:
   - default
+  - mailers
```
AbigailMcP, line 4: Just checking - did you mean to bump the mailers queue priority?

AbigailMcP (approved):
Cute! Tiny comment inline.

config/sidekiq.yml
gpeng: I did. It just seemed like sending emails to users was more important than the stats snapshot. Probably doesn't matter as they both have a tiny amount of jobs on them for the most part anyway. They're both almost always empty.

PR #2142, opened 2021-09-07
[2392] Throttle on Redis out of memory

gpeng:
### Context

The redis client will raise a `Redis::CommandError` when an operation is attempted that will cause it to breach its configured memory limit. When we ran this task on staging we almost reached that limit so as an extra bit of safety this commit will cause a sleep to allow Sidekiq to reduce the queue size back below the threshold. The staging data is effectively identical to production so it is unlikely that we'll hit this limit with more workers running in production. This should be sufficient to deal with an unexpected peak.

* Retry 5 times on Redis::CommandError
* Sleep for 10 seconds to give Sidekiq time to clear some jobs

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/lib/tasks/big_query.rake
+++ b/lib/tasks/big_query.rake
@@ -1,6 +1,9 @@
 require "csv"
 
 namespace :big_query do
+  ON_REDIS_ERROR_WAIT_TIME_IN_SECONDS = 10
+  ON_REDIS_ERROR_RETRY_ATTEMPTS = 5
+
   desc <<~DESC
     Send import events for configured entities
   DESC
@@ -19,7 +22,20 @@ namespace :big_query do
     classes.each do |c|
       puts "Queueing: #{c.count} #{c} entities"
 
-      c.find_each(batch_size: 200, &:send_import_event)
+      c.find_each(batch_size: 200) do |entity|
+        attempt_count = 0
+        begin
+          entity.send_import_event
+        rescue Redis::CommandError
+          attempt_count += 1
+          abort if attempt_count > ON_REDIS_ERROR_RETRY_ATTEMPTS
+
+          Rails.logger.info "sleeping for #{ON_REDIS_ERROR_WAIT_TIME_IN_SECONDS} - redis error"
+
+          sleep ON_REDIS_ERROR_WAIT_TIME_IN_SECONDS
+          retry
+        end
+      end
     end
   end
 end


lib/tasks/big_query.rake
```diff
@@ -19,7 +22,19 @@ namespace :big_query do
...
+          entity.send_import_event
+        rescue Redis::CommandError
+          attempt_count += 1
+          abort if attempt_count > ON_REDIS_ERROR_RETRY_ATTEMPTS
+
+          puts "sleeping for #{ON_REDIS_ERROR_WAIT_TIME_IN_SECONDS} - redis error"
+          sleep ON_REDIS_ERROR_WAIT_TIME_IN_SECONDS
+          retry
+        end
+      end
```
vigneshmsft, line 37: may be add an extra `sleep` after each batch to give the worker some breathing space regardless of the memory build up?

gpeng: Do you think we need to? This is only the enqueueing of the jobs so if Redis is coping with them I don't really see the benefit of slowing this process down.

lib/tasks/big_query.rake
```diff
@@ -19,7 +22,19 @@ namespace :big_query do
...
-      c.find_each(batch_size: 200, &:send_import_event)
+      c.find_each(batch_size: 200) do |entity|
+        attempt_count = 0
+        begin
+          entity.send_import_event
+        rescue Redis::CommandError
+          attempt_count += 1
+          abort if attempt_count > ON_REDIS_ERROR_RETRY_ATTEMPTS
+
+          puts "sleeping for #{ON_REDIS_ERROR_WAIT_TIME_IN_SECONDS} - redis error"
```
misaka, line 33: If I'm not mistaken, this kind of thing will get picked up by the GPaaS log processor and sent to Logit, which expects a JSON string. If we're going to emit this message it may be better to make it a proper log message.

lib/tasks/big_query.rake
misaka: This is a bit nuanced isn't it ... if we give artificially slow down the enqueuing job we might avoid running out of memory completely. That might be good because even thought we're handling errors here now, what happens to anything else that might be trying to enqueue a job?

misaka: I think the problem with the `sleep` approach is how do we know if it's enough. I think we have 4 options here:

1. Ensure we don't hit the memory limit here by checking the memory usage of Redis before enqueuing. This may not be perfect, something else might push it over the limit, but at least this approach is more adaptive and if we refrain from running two batch jobs at a time we should be alright.
2. Put similar protection wherever we enqueue jobs. This might be bad because not every branch of code can afford to stop and wait to retry.
3. Just whack a sleep in there and hope we get it right without slowing down the batch process too much.
4. Do nothing, move along, 🤞 it doesn't blow up.

I think I like option 1 the most.

vigneshmsft (approved):

lib/tasks/big_query.rake
gpeng: Discussed offline. Agreed not to do this.

PR #2202, opened 2021-09-30
[122] Save course after assigning program type

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/CKtm7FGo/122-bug-incorrect-route-being-set)

- Added `course.save` after assigning program type to persist to DB

### Guidance to review

- Go into rails console locally (make sure you have a DB dump of TTAPI) and run the following: 
```
> Course.find_by course_code: "L244"
> c = _
> c.program_type 
~> returns "scitt_programme"
> c.funding_type = c.funding_type
~> returns "fee"
> c.reload
> c.program_type  
~> returns "school_direct_training_programme"
```

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


stellenberger:
[[Bug] Incorrect route being set](https://trello.com/c/CKtm7FGo/122-bug-incorrect-route-being-set)

app/models/course.rb
```diff
@@ -540,6 +540,7 @@ def is_published?
   def funding_type=(funding_type)
     assign_program_type_service = Courses::AssignProgramTypeService.new
     assign_program_type_service.execute(funding_type, self)
+    save unless errors
```
gpeng, line 543: We shouldn't do this here as it's unexpected that setting an attribute will save the object. I think where you put it originally was more along the right lines. In fact, the calling code should use the service itself. This is a weird way to do it altogether 😆

app/services/courses/assign_program_type_service.rb
```diff
@@ -9,6 +9,7 @@ def execute(funding_type, course)
       when "fee"
         course.program_type = update_fee_program(course)
       end
+      course.save unless course.errors.any?
```
gpeng, line 12: We don't need the `course.errors.any?` bit. It would be better to just call `course.save` as that would make the return value of the method consistently a boolean that indicates whether it succeeded or not.

PR #2206, opened 2021-10-01
[2261] Stop logging bigquery

saliceti:
### Context
Filling up Logit with bigquery job log

Stop filling up Logit with bigquery job log by setting sidekig log level to WARN and bigquery job activejob log to null logger

### Guidance to review
Same implementation as find

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -3,6 +3,7 @@ if ENV.key?("VCAP_SERVICES")
     config.redis = {
       url: ENV.fetch("REDIS_WORKER_URL"),
     }
+    config.logger.level = Logger::WARN
 
     if Settings.bg_jobs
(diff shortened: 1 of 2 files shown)


gpeng (approved):

config/initializers/sidekiq.rb
```diff
@@ -3,6 +3,7 @@
     config.redis = {
       url: ENV.fetch("REDIS_WORKER_URL"),
     }
+    config.logger.level = Logger::WARN
```
gpeng, line 6: Do we want to lower the level of all sidekiq jobs? The big volume one is the BigQuery one and you've already handled that.

saliceti: This only removes the "start" and "stop" default sidekiq logs. Since we use activejob, we also get the "Processing" and "Processed" logs.

PR #2213, opened 2021-10-05
[xxx] add missing indexes

gpeng:
### Context

We use these fields for filtering in the CourseSearchService.

Add indexes for them. I'm not sure how much benefit we'll get from this as the fields all have low cardinality. We could consider adding compound indexes for the common combinations though.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


db/schema.rb
```diff
@@ -237,6 +243,40 @@
     t.index ["user_id"], name: "IX_organisation_user_user_id"
   end
 
+  create_table "pets2", id: false, force: :cascade do |t|
```
TobyRet, line 246: 'pets2'? 😆

db/schema.rb
```diff
@@ -345,6 +385,40 @@
     t.index ["typename"], name: "index_subject_area_on_typename", unique: true
   end
 
+  create_table "temp", id: false, force: :cascade do |t|
```
TobyRet, line 388: ❓

db/schema.rb
```diff
@@ -10,11 +10,12 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_07_12_082658) do
+ActiveRecord::Schema.define(version: 2021_10_05_153012) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "btree_gist"
+  enable_extension "citext"
```
TobyRet, line 18: Is this supposed to be here?

TobyRet (changes requested):
Sau would be proud of me ...

defong:
> Sau would be proud of me ...

I am very proud of you

db/schema.rb
gpeng: I have no idea what sort of foul play is at work here 🤣 I've removed it anyway.

dankmitchell (approved):

PR #2216, opened 2021-10-06
Add provider to V3 serializable course cache key

malcolmbaig:
So that updates to the provider bust the cache entry for serializable
courses.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


gpeng (approved):
I think changes on the course should update `changed_at` on the provider so we probably could use just the provider here. Not sure it would make much difference performance wise though.

PR #2240, opened 2021-10-13
[xxx] Missing personalisation first_name visibility

stellenberger:
### Context

- We are getting a consistent stream of errors where new users are being created without a first name, and passed into the email service. This is odd, as we have multiple validation checks on the user model to ensure they have a first name. 

- Gaining more visibility into the issue through raising errors, which *should* tell us roughly what the user object looks like when the method was being called, and at which point the data is incorrect. 
- Raised errors in the last instance if the `WelcomeEmailEmailer` is called without passing in a first name or if the `SendWelcomeEmailService` is called with a user that does not have a first name. 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/spec/services/send_welcome_email_service_spec.rb
+++ b/spec/services/send_welcome_email_service_spec.rb
@@ -116,7 +116,7 @@ describe SendWelcomeEmailService do
         expect(current_user_spy).to_not have_received(:update).with(hash_including(welcome_email_date_utc: Time.now.utc))
       end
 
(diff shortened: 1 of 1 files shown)


stellenberger:
[[Bug] Missing personalisation first_name](https://trello.com/c/juxs5ZRt/244-bug-missing-personalisation-firstname)

PR #2263, opened 2021-10-29
[320] Add support interface screen to edit uplifts

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/VRq3HZqn/320-add-support-interface-screen-to-edit-uplifts)

- Add `Allocations` tab on the support console
- Lists providers with allocations
- Allows CRU actions on AllocationUplift relation to Allocation

### Guidance to review

- Log into support console
- Go to allocations tab
- CRU allocation uplifts on each allocation for each provider listed

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support/allocation_uplifts_controller.rb
+++ b/app/controllers/support/allocation_uplifts_controller.rb
@@ -19,7 +19,7 @@ module Support
 
     def update
       if allocation_uplift.update(update_allocation_uplift_params)
-        redirect_to support_allocation_path(allocation_uplift.allocation)
+        redirect_to support_allocation_path(allocation_uplift.allocation), flash: { success: "Allocation Uplift was successfully updated" }
       else
         render :edit
       end
--- a/app/controllers/support/allocation_uplifts_controller.rb
+++ b/app/controllers/support/allocation_uplifts_controller.rb
@@ -10,6 +10,7 @@ module Support
 
     def create
       @allocation_uplift = AllocationUplift.new(create_allocation_uplift_params)
+      @allocation_uplift.allocation = allocation
       if @allocation_uplift.save
         redirect_to support_allocation_path(@allocation_uplift.allocation), flash: { success: "Allocation Uplift was successfully created" }
       else
@@ -28,7 +29,7 @@ module Support
   private
 
     def allocation_uplift
-      @allocation_uplift ||= AllocationUplift.find(params[:id])
+      @allocation_uplift ||= allocation.allocation_uplift
     end
 
     def allocation
--- a/app/controllers/support/allocation_uplifts_controller.rb
+++ b/app/controllers/support/allocation_uplifts_controller.rb
@@ -1,0 +1,47 @@
+module Support
+  class AllocationUpliftsController < ApplicationController
+    def edit
+      allocation_uplift
+    end
+
+    def update
+      if allocation_uplift.update(update_allocation_uplift_params)
+        redirect_to support_allocation_path(allocation_uplift.allocation)
+      else
+        render :edit
+      end
+    end
+
+    def new
+      @allocation_uplift = AllocationUplift.new(allocation: allocation)
+    end
+
+    def create
+      @allocation = Allocation.find(create_allocation_uplift_params[:allocation_id].to_i)
+      @allocation_uplift = AllocationUplift.new(allocation: @allocation, uplifts: create_allocation_uplift_params[:uplifts])
+      if @allocation_uplift.save
+        redirect_to support_allocation_path(allocation), flash: { success: "Allocation Uplift was successfully created" }
+      else
+        render :new
+      end
+    end
+
+  private
+
+    def allocation_uplift
+      @allocation_uplift ||= AllocationUplift.find(params[:id])
+    end
+
+    def allocation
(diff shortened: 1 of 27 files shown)


stellenberger:
[Add support interface screen to edit uplifts](https://trello.com/c/VRq3HZqn/320-add-support-interface-screen-to-edit-uplifts)

spec/features/support/allocations/allocation_uplift_edit_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "Edit an Allocation Uplift" do
+  let(:user) { create(:user, :admin) }
+  let(:provider) { create(:provider) }
+  let(:allocation) { create(:allocation, :with_allocation_uplift, provider: provider, number_of_places: 5) }
+  let(:new_allocation_uplift) { 500 }
```
malcolmbaig, line 9: I think with feature specs we can skip the use of `let` blocks — they're a layer of indirection that don't necessarily make the spec easier to follow. The Apply [testing styleguide](https://github.com/DFE-Digital/apply-for-teacher-training/blob/0efbef5707db151b128015c710011b8685b3485b/docs/testing-styleguide.md) outlines what I think we should be aiming for.

app/models/allocation.rb
```diff
@@ -16,6 +16,8 @@ class Allocation < ApplicationRecord
 
   enum request_type: { initial: 0, repeat: 1, declined: 2 }
 
+  scope :current_allocations, -> { RecruitmentCycle.find_by(year: Settings.allocation_cycle_year).allocations }
```
malcolmbaig, line 19: Might be simpler for this to be `where(recruitment_cycle: RecruitmentCycle.find_by(year: Settings.allocation_cycle_year))`

app/controllers/support/allocations_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    def allocation
+      @allocation ||= Allocation.find(params[:id])
+    end
+
+    def filtered_providers_with_allocations
+      Support::Providers::Filter.call(providers: find_providers_with_allocations, filters: filters)
+    end
+
+    def find_providers_with_allocations
+      Provider.where(id: Allocation.current_allocations.select(:provider_id)).order(:provider_name)
```
malcolmbaig, line 22: Could this be a scope on Provider, eg `Provider.with_allocations_for_current_cycle_year`?

The implementation could also be a join between providers and allocations, along with a where the scopes to the current cycle year.

config/routes.rb
```diff
@@ -29,6 +29,10 @@
     end
     resources :users, only: %i[index show new create]
 
+    resources :allocations, only: %i[index show]
+
+    resources :allocation_uplifts, only: %i[edit update create new]
```
malcolmbaig, line 34: For a bit of extra safety, shall we nest these under members of the allocations resource, so that you can scope retrieval of the uplifts in the controller by the allocation they belong to?

app/controllers/support/allocation_uplifts_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+        redirect_to support_allocation_path(allocation_uplift.allocation), flash: { success: "Allocation Uplift was successfully updated" }
+      else
+        render :edit
+      end
+    end
+
+  private
+
+    def allocation_uplift
+      @allocation_uplift ||= AllocationUplift.find(params[:id])
```
malcolmbaig, line 31: See related comment https://github.com/DFE-Digital/teacher-training-api/pull/2263#discussion_r744549630

mbocc (approved):
Added a PE uplift, and it works.Looks great, thank you!

spec/features/support/allocations/allocation_uplift_new_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "Edit an Allocation Uplift" do
+  let(:user) { create(:user, :admin) }
+  let(:provider) { create(:provider) }
+  let(:allocation) { create(:allocation, provider: provider, number_of_places: 5) }
+  let(:allocation_uplift_amount) { 5 }
```
malcolmbaig, line 9: Worth refactoring these lets out of the spec as well.

app/models/recruitment_cycle.rb
```diff
@@ -15,6 +15,11 @@ def current_recruitment_cycle
     end
     alias_method :current, :current_recruitment_cycle
 
+    def current_allocation_recruitment_cycle
```
gpeng, line 18: I don't think `RecruitmentCycle` should know about allocations. This should be on `Allocation` (or somewhere else)

gpeng: Is this only being used in the specs too? If so it should probably just be defined in the factory rather than out in the code.

gpeng (changes requested):

app/models/recruitment_cycle.rb
stellenberger: Sure, removed as not being used 👍

gpeng (approved):

PR #2266, opened 2021-11-05
[xxx] Validate accredited body if course not self accredited

stellenberger:
### Context

- We are getting situations where an accredited body is discarded, but courses that are accredited by them are still able to be published. 

 - Add a validation that prevents courses from being published, if their AB does not exist in the current cycle

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -304,6 +304,7 @@ class Course < ApplicationRecord
   validate :validate_subject_count
   validate :validate_subject_consistency
   validate :validate_custom_age_range, on: %i[create new], if: -> { age_range_in_years.present? }
+  validate :accredited_body_exists_in_current_cycle, on: :publish, unless: -> { self_accredited? }
   validates_with UniqueCourseValidator, on: :new
 
   validates :name, :profpost_flag, :program_type, :qualification, :start_date, :study_mode, presence: true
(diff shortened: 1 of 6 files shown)


stellenberger:
[[Bug] Validate that accredited body exists in current cycle when publishing a course](https://trello.com/c/hiqOeTJU/358-bug-validate-that-accredited-body-exists-in-current-cycle-when-publishing-a-course)

app/models/course.rb
```diff
@@ -869,4 +870,8 @@ def services
   def self_accredited_cannot_be_salary_funded
     errors.add(:program_type, "Salary is not valid for a self accredited course") if self_accredited? && funding_type == "salary"
   end
+
+  def accredited_body_exists_in_current_cycle
+    errors.add(:base, "The Accredited Body #{accredited_body_code} does not exist in this cycle") unless RecruitmentCycle.current.providers.find_by(provider_code: accredited_body_code)
+  end
```
malcolmbaig, line 876: Should this return early if accredited_body_code is blank?

spec/requests/api/v2/providers/courses/publish_spec.rb
```diff
@@ -140,6 +143,7 @@
             "Enter a course length",
             "Enter details about the salary for this course",
             "Enter GCSE requirements",
+            "The Accredited Body  does not exist in this cycle",
```
malcolmbaig, line 146: Rather than checking the presence of this error throughout the spec, it seems like this validation warrants its own context block, asserting that the error appears when the accredited body is not in the current cycle. 

Related comment: https://github.com/DFE-Digital/teacher-training-api/pull/2266/files#r745495169

app/models/course.rb
stellenberger: Yes, added a guard clause 👍

malcolmbaig (approved):

PR #2284, opened 2021-11-16
[0349] Rolling over providers beyond the rollver time period

defong:
### Context
Rolling over providers from bygone period
The rollover time period are ephemeral, and tied to `current` & `next` recruitment cycle concepts.

Instead of `current` & `next` recruitment cycle concepts replace it with `source` and `target` recruitment cycles as that will allow for `past` & `current` recruitment cycle pairing.

By allowing `source` and `target` recruitment cycles to be specified.
It removes the ephemeral nature of `rollover time period`.
Allowing source recruitment cycle providers to target recruitment cycle.
As these providers falls outside of `rollover time period`.

### Guidance to review

1. Restore sanitise copy of the database 
2. Then `bundle exec rails console` 
3. Then run the following Scenarios
"1DX" has 86 courses from last cycle but have not ran any of them previously

Scenario one
We realise that the "1DX" wanted both their provider and courses to be rollovered over a period of time
```ruby
RolloverService.call(provider_codes: ["1DX"], source_recruitment_cycle: RecruitmentCycle.current.previous, target_recruitment_cycle: RecruitmentCycle.current, force: :nil)
# 10am Rails -- Rollover: provider skipped, 0 sites copied, 0 courses copied in 0.435 seconds

RolloverService.call(provider_codes: ["1DX"], source_recruitment_cycle: RecruitmentCycle.current.previous, target_recruitment_cycle: RecruitmentCycle.current, force: :provider)
# 11am Rails -- Rollover: provider copied, 22 sites copied, 0 courses copied in 1.021 seconds

RolloverService.call(provider_codes: ["1DX"], source_recruitment_cycle: RecruitmentCycle.current.previous, target_recruitment_cycle: RecruitmentCycle.current, force: :provider_and_courses)
# 12am Rails -- Rollover: provider skipped, 0 sites copied, 86 courses copied in 5.730 seconds

```

Scenario two
We realise that the "1DX" wanted both their provider and courses to be rollovered in the first instance
```ruby 
RolloverService.call(provider_codes: ["1DX"], source_recruitment_cycle: RecruitmentCycle.current.previous, target_recruitment_cycle: RecruitmentCycle.current, force: :nil)
# 10am Rails -- Rollover: provider skipped, 0 sites copied, 0 courses copied in 0.435 seconds

RolloverService.call(provider_codes: ["1DX"], source_recruitment_cycle: RecruitmentCycle.current.previous, target_recruitment_cycle: RecruitmentCycle.current, force: :provider_and_courses)
# 11am Rails -- Rollover: provider copied, 22 sites copied, 86 courses copied in 6.477 seconds
```

Scenario three
We realise that the "1DX" wanted just their provider to be rollovered in the first instance
```ruby 
RolloverService.call(provider_codes: ["1DX"], source_recruitment_cycle: RecruitmentCycle.current.previous, target_recruitment_cycle: RecruitmentCycle.current, force: :nil)
# 10am Rails -- Rollover: provider skipped, 0 sites copied, 0 courses copied in 0.435 seconds

RolloverService.call(provider_codes: ["1DX"], source_recruitment_cycle: RecruitmentCycle.current.previous, target_recruitment_cycle: RecruitmentCycle.current, force: :provider)
# 11am Rails -- Rollover: provider copied, 22 sites copied, 0 courses copied in 1.021 seconds
```

Scenario four
- idempotent, running the command more then once is a noops
4. `bundle exec rake rollover:providers` should still work as expected

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


defong:
[New Horizons Training Partnership (1DX)](https://trello.com/c/DJiElVg8/349-new-horizons-training-partnership-1dx)

dankmitchell:
@defong the force was for providers without any running courses I.e. the intended to run this cycle but didn't the last cycle

app/services/courses/copy_to_provider_service.rb
```diff
@@ -5,7 +5,7 @@ def initialize(sites_copy_to_course:, enrichments_copy_to_course:)
       @enrichments_copy_to_course = enrichments_copy_to_course
     end
 
-    def execute(course:, new_provider:, force: false)
+    def execute(course:, new_provider:, force:)
       return nil unless course.rollable? || force
```
gpeng, line 9: the `nil` here is redundant.

app/services/rollover_service.rb
```diff
@@ -1,9 +1,12 @@
 class RolloverService
   include ServicePattern
 
-  def initialize(provider_codes:, force: false)
+  def initialize(provider_codes:, force: nil, source_recruitment_cycle: RecruitmentCycle.current_recruitment_cycle, target_recruitment_cycle: RecruitmentCycle.next_recruitment_cycle, logger: Rails.logger)
```
gpeng, line 4:
```suggestion
  def initialize(provider_codes:, force:, source_recruitment_cycle: RecruitmentCycle.current_recruitment_cycle, target_recruitment_cycle: RecruitmentCycle.next_recruitment_cycle, logger: Rails.logger)
``` 
Also I don't think we need to inject the logger here do we? We'll always be using `Rails.logger`

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -8,12 +8,12 @@ def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:,
       @logger = logger || Logger.new("/dev/null")
     end
 
-    def execute(provider:, new_recruitment_cycle:, force: false)
+    def execute(provider:, new_recruitment_cycle:, force:)
```
gpeng, line 11: I think the intention of `force` was just to allow us to roll providers that haven't been rolled over because they weren't rollable. I don't think we should overload this, especially with it being a boolean or a symbol. I'm not really sure what we're trying to achieve with the :provider and :provider_and_courses. It seems like we're trying to make this do too much.

spec/services/courses/copy_to_provider_service_spec.rb
```diff
@@ -32,7 +32,7 @@
   end
 
   it "makes a copy of the course in the new provider" do
-    service.execute(course: course, new_provider: new_provider)
+    service.execute(course: course, new_provider: new_provider, force: false)
```
gpeng, line 35: Why did we remove the default value and then have to start passing it everywhere?

gpeng (changes requested):

app/services/courses/copy_to_provider_service.rb
defong: :face_with_head_bandage: 

We can card this and leave it for another time as it solely tech debt

app/services/rollover_service.rb
defong: It was getting really confusing with 

All 3 of these was getting used
`puts`
`prints`
`Logger.new(STDOUT)`

as well as all the noise

image

image

And since reading the log will be an actual thing to be done, it should be more extensive tested, hence injection.

spec/services/courses/copy_to_provider_service_spec.rb
defong: It broke this for me 
The principle of least astonishment 

passing a `force = true` from the outer layer in i would of expected it to populate down to inner layer  

As a `false is a false` as well as `silent is false` so why  `silent`?

https://github.com/DFE-Digital/teacher-training-api/blob/master/app/services/providers/copy_to_recruitment_cycle_service.rb#L33

https://github.com/DFE-Digital/teacher-training-api/blob/master/spec/services/courses/copy_to_provider_service_spec.rb#L190-L207 

Enforce by the fact that there is test for it but no path of execution

app/services/providers/copy_to_recruitment_cycle_service.rb
defong: "1DX" has 86 courses from last cycle but have not ran any of them previously

`:provider` will be the 
> I think the intention of force was just to allow us to roll providers that haven't been rolled over because they weren't rollable.

`:provider_and_courses.` will be the above as well as actually rolling over the course that has not be ran previously


otherwise, on without any code change

```ruby
old_provider = RecruitmentCycle.current.previous.providers.find_by(code: "1DQ")

    copy_provider_to_recruitment_cycle = Providers::CopyToRecruitmentCycleService.new(
      copy_course_to_provider_service: nil,
      copy_site_to_provider_service: Sites::CopyToProviderService.new,
      logger: Logger.new(STDOUT),
    )

    copy_provider_to_recruitment_cycle.execute(
      provider: old_provider, new_recruitment_cycle: RecruitmentCycle.current, force: true,
    )

    new_provider = RecruitmentCycle.current.providers.find_by(code: "1DQ")

    old_provider.courses.each do |course|
      Courses::CopyToProviderService.new(
        sites_copy_to_course: Sites::CopyToCourseService.new,
        enrichments_copy_to_course: Enrichments::CopyToCourseService.new,
      ).execute(course: course, new_provider: new_provider, force: true)    
    end 

```

To be ran arbitrary per supports needs,

app/services/courses/copy_to_provider_service.rb
defong: https://trello.com/c/SkOktCFG/608-remove-pointless-return-nil

app/services/providers/copy_to_recruitment_cycle_service.rb
gpeng: I still think this is overloading 'force'. It's taking a lot of explanation which means the intention isn't clear. It's probably a mistake rolling the courses over at all in the provider service, that's where the complexity of all of this is coming from.

spec/services/courses/copy_to_provider_service_spec.rb
gpeng: Not to force is just the default operation so there's no need to pass a value. I agree you'd expect it to pass through to the inner service although as I've said, having the provider service call the course one is probably why all of this is getting so complicated.

defong:
Discussed with @gpeng we have opted for https://github.com/DFE-Digital/teacher-training-api/pull/2284#discussion_r753141637 using custom snippet of code as there is not enough well weighted reasons to endorse this change.

PR #2293, opened 2021-11-23
Rework DfE Sign-in for Support and Publish authentication

malcolmbaig:
### Context
We're merging Publish into TTAPI. Some initial work is required in the TTAPI codebase before we start migrating over functionality.

A DfE Sign-in flow that accommodates both existing Support users and future Publish users needs to be in place. Each sign in should redirect the user to an appropriate default start page.

https://trello.com/c/oSOo6lNC

- Add basic sign in flow for both Support and Publish 
- Treat `/publish` as the root for Publish, `/support/providers` as the root for Support

### Guidance to review
- Review of the diff is probably sufficient, given this is mostly plumbing at this stage
- Can be tested locally by switching to DfE Sign-in auth mode

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/views/sign_in/index.html.erb
```diff
@@ -6,9 +6,9 @@
     <h1 class="govuk-heading-l">Sign in to <%= I18n.t("service_name") %></h1>
 
     <% if AuthenticationService.dfe_signin? %>
-        <p class="govuk-body-m">Use DfE Sign-in to access your account. You need to be a registered administrator to login to these pages.</p>
+        <p class="govuk-body-m">Use DfE Sign-in to access your account.</p>
```
defong, line 9: image

I think we need to address the `admin` wording on the page

app/views/layouts/publish_interface.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= extends_layout :application do %>
```
defong, line 1: Missing page title

image

defong (approved):
Looks fine to me just some `admin` wordings

app/views/sign_in/index.html.erb
malcolmbaig: Was going to make some changes here, but on second thought let's do that as part of https://trello.com/c/i4t2NzH7/606-publish-ttapi-merge-update-sign-in-page-copy-on-ttapi.

malcolmbaig:
Minor clarification:

>and the jsonapi_rails gem expects the updated_at column to be updated

Rather, the custom cache key we specify expects the updated_at column to be updated.

PR #2308, opened 2021-11-29
[397] Fix Stats service to pass in correct cycle

george-bit:
### Context

The statistics for allocations on the performance Dashboard are coming 
through incorrectly. After looking into this the service is workingas it 
should be. It's just that because Allocations take place in June/July 
i.e. before cycle starts they are associated with previous year's 
recruitment cycle. 

The quick fix here is to pass in the previous year.  I've done this and run the numbers past Adam who says they all look as 
expect for the previous and current cycles.

### Guidance to review

Would this need a feature/system spec?

or should we backfill all of the allocations to match the revruitment cycle they are for?

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/services/statistic_service.rb
+++ b/app/services/statistic_service.rb
@@ -4,7 +4,7 @@ class StatisticService
       providers: ProviderReportingService.call(providers_scope: recruitment_cycle.providers),
       courses: CourseReportingService.call(courses_scope: recruitment_cycle.courses),
       publish: PublishReportingService.call(recruitment_cycle_scope: recruitment_cycle),
-      allocations: AllocationReportingService.call(recruitment_cycle_scope: recruitment_cycle),
+      allocations: AllocationReportingService.call(recruitment_cycle_scope: RecruitmentCycle.find_by(year: Settings.allocation_cycle_year)),
       rollover: RolloverReportingService.call,
     }
   end
--- a/spec/factories/recruitment_cycles.rb
+++ b/spec/factories/recruitment_cycles.rb
@@ -15,5 +15,9 @@ FactoryBot.define do
     trait :current_allocation do
       year { Settings.allocation_cycle_year }
     end
+
+    trait :previous_allocation_cycle do
+      year { (Settings.allocation_cycle_year - 1).to_s }
+    end
   end
 end
--- a/spec/services/statistic_service_spec.rb
+++ b/spec/services/statistic_service_spec.rb
@@ -3,6 +3,8 @@ require "rails_helper"
 describe StatisticService do
   let(:recruitment_cycle) { find_or_create(:recruitment_cycle) }
   let!(:previous_recruitment_cycle) { find_or_create(:recruitment_cycle, :previous) }
+  let!(:allocation_recruitment_cycle) { find_or_create(:recruitment_cycle, :current_allocation) }
+  let!(:previous_allocation_cycle) { find_or_create(:recruitment_cycle, :previous_allocation_cycle) }
 
   describe "#reporting" do
     subject { described_class.reporting(recruitment_cycle: recruitment_cycle) }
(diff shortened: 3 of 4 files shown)


defong (changes requested):
Can we have a token test since it affects an entire cycle presentation

defong:
[Fix Allocation tab on Service Performance Dashboard](https://trello.com/c/HMSkA7bZ/397-fix-allocation-tab-on-service-performance-dashboard)

app/services/statistic_service.rb
```diff
@@ -4,7 +4,7 @@ def self.reporting(recruitment_cycle:)
       providers: ProviderReportingService.call(providers_scope: recruitment_cycle.providers),
       courses: CourseReportingService.call(courses_scope: recruitment_cycle.courses),
       publish: PublishReportingService.call(recruitment_cycle_scope: recruitment_cycle),
-      allocations: AllocationReportingService.call(recruitment_cycle_scope: recruitment_cycle),
+      allocations: AllocationReportingService.call(recruitment_cycle_scope: recruitment_cycle.previous),
```
stellenberger, line 7: Use the `allocation_cycle_year` in the settings for all allocation related things

```suggestion
      allocations: AllocationReportingService.call(recruitment_cycle_scope: RecruitmentCycle.find_by(year: Settings.allocation_cycle_year)),
```

stellenberger (changes requested):

app/services/statistic_service.rb
george-bit: Aweesome will do, thanks 🙌

spec/factories/recruitment_cycles.rb
```diff
@@ -8,6 +8,10 @@
       year { (Settings.current_recruitment_cycle_year.to_i - 1).to_s }
     end
 
+    trait :year_before_last do
+      year { (Settings.current_recruitment_cycle_year.to_i - 2).to_s }
+    end
+
```
stellenberger, line 14: seems like the only time we use a RC two years prior is for the previous allocation cycle. perhaps `previous_allocation_cycle` is a better-suited name for this trait? 

```suggestion
   trait :previous_allocation_cycle do
      year { (Settings.allocation_cycle_year - 1).to_s }
    end

```

spec/services/statistic_service_spec.rb
```diff
@@ -23,7 +24,7 @@
     end
 
     it "calls the allocation reporting service" do
-      expect(AllocationReportingService).to receive(:call).with(recruitment_cycle_scope: recruitment_cycle)
+      expect(AllocationReportingService).to receive(:call).with(recruitment_cycle_scope: RecruitmentCycle.find_by(year: Settings.allocation_cycle_year))
```
stellenberger, line 27: perhaps we can put this into a variable using the factories, for consistency

```suggestion
      expect(AllocationReportingService).to receive(:call).with(recruitment_cycle_scope: allocation_recruitment_cycle)
```

There is also a trait for this, which makes it easier that finding the RC with the settings

stellenberger (changes requested):

spec/services/statistic_service_spec.rb
george-bit: There's a bit of subtelty here as `previous_recruitment_cycle` and `allocation_recruitment_cycle` are the same cycle and in this spec both names are relevant as only the AllocationReportingService would be for `allocation_recruitment_cycle` the rest would make more sense for `previous_recruitment_cycle`,  I've gone for a hybrid approach in the end.

defong (approved):
It got tests lets grab some :taco:

stellenberger (approved):
🚀

PR #2310, opened 2021-11-30
Add additional Persona users

malcolmbaig:
### Context
A persona-based authentication mode exists for use in development and
QA. Currently this supports sign in of a support user only.

https://trello.com/c/Ghmij2De

Enhance this with additional users ported across from the Publish
app. Each of these users maps to an anonymised account in the TTAPI
database dump available on Github.

### Guidance to review
- Can be accessed via `/sign-in` when running locally with persona auth mode enabled.
- Doesn't do much right now — you'll go to support if you sign in as Colin, everyone else will go to the blank Publish holding page.
- Have retained (but re-written) the lightweight feature spec that was already there. This is probably sufficient given this is an internally-facing bit of functionality. Also, our route definitions don't define the developer-mode endpoints unless the app is started in persona mode (rightly, imo), so we'd have to rewrite route definitions on the fly to test an actual persona sign in. Not worth the additional complexity I think.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/components/persona/view.html.erb
+++ b/app/components/persona/view.html.erb
@@ -1,15 +1,6 @@
-<h2 class="govuk-heading-m">Colin</h2>
-
-  <p class="govuk-body"><%= govuk_tag(text: "Admin", colour: "blue") %></p>
-
-  <p class="govuk-body">
-    Colin is a DfE support agent who has administrator access to all organisations.
-  </p>
-</h2>
-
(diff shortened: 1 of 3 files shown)


defong (approved):
Nice one... all looks good

signing-in

PR #2313, opened 2021-12-01
[319] Update pagy gem

george-bit:
### Context

Dependabot failed to update the Pagy gem as it introduced breaking changes

After updating the gem locally it looks as if certain syntax around Pagy has been updated. There's no longer a need to specify `urls: true` in concerns.  Also, there is no longer a need to include a fake protocol and domain of `http://test.host` for url prefixes in tests that use the gem.

### Guidance to review

Have I overlooked anything?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,7 @@ gem "active_model_serializers"
 gem "kaminari"
 
 # Pagination for API
-gem "pagy", "~> 3.13"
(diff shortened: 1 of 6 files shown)


stellenberger (approved):
lgtm

PR #2316, opened 2021-12-02
Revert "[319] Update pagy gem"

george-bit:
Reverts DFE-Digital/teacher-training-api#2313


defong (approved):

george-bit:
Fix found on #2317 instead

PR #2329, opened 2021-12-10
[457] (Part 3) Rake task to migrate userpremission from organisations

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/hlEKCLCm/457-rake-task-to-migrate-userpremission-from-organisations)

- A rake task to migrate the OrganisationUser records to UserPermission

### Guidance to review

- `bundle exec rake organisation_user_migrate:migrate_organisation_user_to_user_permission`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/lib/tasks/migrate_user_organisation_to_user_permission.rake
+++ b/lib/tasks/migrate_user_organisation_to_user_permission.rake
@@ -1,0 +1,19 @@
+namespace :organisation_user_migrate do
+  task migrate_organisation_user_to_user_permission: :environment do
+    total_bm = Benchmark.measure do
+      UserPermission.insert_all(
+        OrganisationUser.includes(:user, organisation: [providers: [:recruitment_cycle]]).all.flat_map { |organisation_user|
+          organisation_user.organisation.providers.where(recruitment_cycle: RecruitmentCycle.current).ids.map do |id|
+            {
+              user_id: organisation_user.user.id,
+              provider_id: id,
+              updated_at: Time.current,
+              created_at: Time.current,
+            }
+          end
+        },
+      )
+    end
+    puts total_bm.real
+  end
+end
--- a/lib/tasks/migrate_user_organisation_to_user_permission.rake
+++ b/lib/tasks/migrate_user_organisation_to_user_permission.rake
@@ -3,7 +3,7 @@ namespace :organisation_user_migrate do
     total_bm = Benchmark.measure do
       UserPermission.insert_all(
         OrganisationUser.includes(:user, organisation: [providers: [:recruitment_cycle]]).all.flat_map { |organisation_user|
-          organisation_user.organisation.providers.where(recruitment_cycle: RecruitmentCycle.current).ids.map do |id|
+          organisation_user.organisation.providers.ids.map do |id|
             {
               user_id: organisation_user.user.id,
               provider_id: id,
--- a/lib/tasks/migrate_user_organisation_to_user_permission.rake
+++ b/lib/tasks/migrate_user_organisation_to_user_permission.rake
@@ -2,7 +2,7 @@ namespace :organisation_user_migrate do
   task migrate_organisation_user_to_user_permission: :environment do
     total_bm = Benchmark.measure do
       UserPermission.insert_all(
-        OrganisationUser.includes(:user, organisation: [providers: [:recruitment_cycle]]).all.flat_map { |organisation_user|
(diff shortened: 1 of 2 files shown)


stellenberger:
[Rake task to migrate UserPremission from Organisations](https://trello.com/c/hlEKCLCm/457-rake-task-to-migrate-userpremission-from-organisations)

lib/tasks/migrate_user_organisation_to_user_permission.rake
```diff
@@ -0,0 +1,19 @@
+namespace :organisation_user_migrate do
+  task migrate_organisation_user_to_user_permission: :environment do
+    total_bm = Benchmark.measure do
+      UserPermission.insert_all(
+        OrganisationUser.includes(:user, organisation: [providers: [:recruitment_cycle]]).all.flat_map { |organisation_user|
+          organisation_user.organisation.providers.where(recruitment_cycle: RecruitmentCycle.current).ids.map do |id|
```
malcolmbaig, line 6: Do we want to scope this by recruitment cycle? By doing this, we're essentially removing user access to providers from previous recruitment cycles. Does that have any unintended side effects? Does it remove our ability to develop new functionality, eg—giving users a read-only view of previous recruitment cycles? Maybe it would be safer to preserve all existing permissions for now?

cc: @defong @darokel @dankmitchell

stellenberger: Thats a good point, I don't see any reason why we shouldn't keep all the existing records. I'll remove the `recruitment_cycle` from the `where`.

lib/tasks/migrate_user_organisation_to_user_permission.rake
```diff
@@ -0,0 +1,19 @@
+namespace :organisation_user_migrate do
+  task migrate_organisation_user_to_user_permission: :environment do
+    total_bm = Benchmark.measure do
+      UserPermission.insert_all(
```
darokel, line 4: Worth adding `unique_by: %i[ user_id provider_id ]` to this?

stellenberger: Added a unique by on the unique index 👍  Yes, this is probably going to be a one time script

malcolmbaig:
As part of this, it might be worth writing a task that checks the migrated data. That is, something which compares the providers our users have access to via the old table vs the new, and confirms they're the same. We'd want to test the migration task thoroughly on QA before running in prod—something which automates those checks would be quicker and more comprehensive. @stellenberger @darokel WDYT?

darokel:
> As part of this, it might be worth writing a task that checks the migrated data. That is, something which compares the providers our users have access to via the old table vs the new, and confirms they're the same. We'd want to test the migration task thoroughly on QA before running in prod—something which automates those checks would be quicker and more comprehensive. @stellenberger @darokel WDYT?

Yup sounds good to me

lib/tasks/validate_organisation_user_permission_migration.rake
```diff
@@ -0,0 +1,8 @@
+namespace :validate_organisation_user_permission do
+  task run_validation: :environment do
+    User.all.each do |user|
```
malcolmbaig, line 3:
```suggestion
    User.find_each do |user|
```
so we're loading smaller batches rather than all at once.

lib/tasks/validate_organisation_user_permission_migration.rake
```diff
@@ -0,0 +1,8 @@
...
+      raise "User #{user.id} has discrepancies with provider relationships" unless user.providers.ids.sort.uniq == user.providers_via_user_permissions.ids.sort.uniq
```
malcolmbaig, line 4: Could do `user.providers.ids - user.providers_via_user_permissions.ids == []` to skip the pair of sort operations executed per user.

lib/tasks/migrate_user_organisation_to_user_permission.rake
defong: this looks like it has been resolved

lib/tasks/migrate_user_organisation_to_user_permission.rake
defong: this looks like it has ben resolved via 
`unique_by: :index_user_permission_on_user_id_and_provider_id`

malcolmbaig (approved):
LGTM. Worth testing this thoroughly on QA before running on prod.

defong:
LGTM on localhost merging this in... hold tight

PR #2340, opened 2021-12-14
Migrate Provider and Contact details from Old Publish

malcolmbaig:
### Context
We're gradually moving Publish into the TTAPI codebase. This is the next set of code to be migrated across.

https://trello.com/c/wlq5VfsQ

- Move over most of the About Your Organisation section (everything except Visa sponsorship, which will follow in a later PR)
- Small improvements to application layout and sign in flow
- See commits for more detail

### Guidance to review
- In a review app or locally, navigate straight to the About Your Organisation section, eg— http://localhost:3001/publish/organisations/1O8/2022/details
- All edits and validation errors should match what's currently in Old Publish (except Visas, tbc)
- The behaviour in the controller actions is essentially a combination of the behaviour in Old Publish and in the V2 API endpoint that Old Publish calls. I've used a form object to combine these things — this seemed clearer and simpler than just dropping the required logic straight into the controller action.
- As per the guidelines [here](https://ukgovernmentdfe.slack.com/archives/C02D2L94TPH/p1637140879014200?thread_ts=1637140438.013100&cid=C02D2L94TPH), route structures and controller actions are a copy of what's in Old Publish currently. There's definitely scope to refactor some of this once the apps are merged, eg—having separate controllers and form objects for 'about your org' details vs contact details, using more Rails-like action names, etc.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/publish_interface/providers_controller.rb
+++ b/app/controllers/publish_interface/providers_controller.rb
@@ -25,8 +25,8 @@ module PublishInterface
     def update
       authorize @provider, :update?
 
-      @about_form = PublishInterface::AboutYourOrganisationForm.build_from_provider(@provider)
-      @about_form.save(provider_params)
+      @about_form = PublishInterface::AboutYourOrganisationForm.build_from_controller_params(provider_params)
+      @about_form.save
 
       if @about_form.valid?
         flash[:success] = I18n.t("success.published")
@@ -38,7 +38,7 @@ module PublishInterface
         )
       else
         @errors = @about_form.errors.messages
-        render provider_params["page"].to_sym
+        render page_param
       end
     end
 
@@ -62,12 +62,16 @@ module PublishInterface
     def provider_params
       params
         .fetch(:publish_interface_about_your_organisation_form, {})
+        .except(:page)
         .permit(
-          :email, :telephone, :urn, :website, :ukprn, :address1, :address2,
-          :address3, :address4, :postcode, :region_code,
-          :page, :train_with_us, :train_with_disability,
-          accredited_bodies: [:provider_name, :provider_code, :description]
+          *AboutYourOrganisationForm::FIELDS,
+          accredited_bodies: %i[provider_name provider_code description],
         )
+        .merge(provider: @provider)
+    end
+
+    def page_param
+      params.fetch(:publish_interface_about_your_organisation_form).fetch(:page)
     end
   end
 end
--- a/app/controllers/publish_interface/providers_controller.rb
+++ b/app/controllers/publish_interface/providers_controller.rb
@@ -18,6 +18,10 @@ module PublishInterface
       @about_form = PublishInterface::AboutYourOrganisationForm.build_from_provider(@provider)
     end
 
+    def contact
+      @about_form = PublishInterface::AboutYourOrganisationForm.build_from_provider(@provider)
+    end
+
     def update
       authorize @provider, :update?
 
@@ -58,7 +62,12 @@ module PublishInterface
     def provider_params
       params
         .fetch(:publish_interface_about_your_organisation_form, {})
-        .permit(:page, :train_with_us, :train_with_disability, accredited_bodies: [:provider_name, :provider_code, :description])
+        .permit(
+          :email, :telephone, :urn, :website, :ukprn, :address1, :address2,
+          :address3, :address4, :postcode, :region_code,
+          :page, :train_with_us, :train_with_disability,
+          accredited_bodies: [:provider_name, :provider_code, :description]
+        )
     end
   end
 end
--- a/app/controllers/publish_interface/providers_controller.rb
+++ b/app/controllers/publish_interface/providers_controller.rb
@@ -14,6 +14,10 @@ module PublishInterface
       flash.delete(:error_summary)
     end
 
+    def about
+      show_deep_linked_errors(%i[train_with_us train_with_disability])
+    end
+
   private
 
     def build_recruitment_cycle
@@ -29,6 +33,12 @@ module PublishInterface
       flash[:error] = { id: "provider-error", message: "Please enter a UKPRN before continuing" }
 
       redirect_to contact_provider_recruitment_cycle_path(@provider.provider_code, @provider.recruitment_cycle_year)
+    end
+
+    def show_deep_linked_errors(attributes)
+      return if params[:display_errors].blank?
+
+      @errors = @provider.errors.messages.select { |key| attributes.include? key }
     end
   end
 end
--- a/app/controllers/publish_interface/providers_controller.rb
+++ b/app/controllers/publish_interface/providers_controller.rb
@@ -15,7 +15,27 @@ module PublishInterface
     end
 
     def about
-      show_deep_linked_errors(%i[train_with_us train_with_disability])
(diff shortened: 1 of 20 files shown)


app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template ">
   <head>
-    <title><%= yield :page_title %></title>
+    <title><%= yield :page_title %> - Publish teacher training courses - GOV.UK</title>
```
defong, line 4: Identity crises

image

defong (changes requested):
Validation causes the changes lost


lost changes 2
lost changes

app/forms/publish_interface/about_your_organisation_form.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    def valid?
+      @provider.valid?
+    end
+
+    def save(params)
+      accredited_bodies_params = params.delete(:accredited_bodies)
+      provider_params = params.except(:page)
+
+      update_provider(provider_params)
```
darokel, line 55: could we wrap these two save calls so we just do it one database transaction?

spec/features/publish_interface/edit_provider_details_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  def when_i_visit_the_details_page
+    visit details_publish_provider_recruitment_cycle_path(@provider.provider_code, @provider.recruitment_cycle_year)
+    click_button "Sign in using DfE Sign-in"
+  end
+
+  def then_i_can_edit_info_about_training_with_us
+    click_link "Change details about training with your organisation"
+    fill_in "Training with you", with: ""
```
darokel, line 37: I think we've opted to use page objects to drive these interactions (there should already be some for reference) would be good if we could stick to this going forward? Otherwise it'll be more tech debt to deal with later

app/controllers/publish_interface/providers_controller.rb
```diff
@@ -30,5 +58,16 @@ def redirect_to_contact_page_with_ukprn_error
 
       redirect_to contact_provider_recruitment_cycle_path(@provider.provider_code, @provider.recruitment_cycle_year)
     end
+
+    def provider_params
+      params
+        .fetch(:publish_interface_about_your_organisation_form, {})
+        .permit(
```
darokel, line 65: I think the form object could declare all these fields in a constant eg `FIELDS` then we could just do a `permit(*PublishInterface::AboutYourOrganisationForm::FIELDS)` to permit whatever is declared from the form? Also means you could do the same with the `attr_accessor` in the form to use the same const

darokel (commented):
Left a few comments

app/forms/publish_interface/about_your_organisation_form.rb
```diff
@@ -0,0 +1,92 @@
...
+    attr_accessor :provider
+    attr_accessor(
+      :provider_code, :recruitment_cycle_year, :provider_name,
+      :train_with_us, :train_with_disability,
+      :email, :telephone, :urn, :website, :ukprn, :address1, :address2,
+      :address3, :address4, :postcode, :region_code
+    )
+
+    def self.build_from_provider(provider)
+      new(
```
darokel, line 15: adding to this https://github.com/DFE-Digital/teacher-training-api/pull/2340#discussion_r770526301 and an alternative,  you could probably just do something like `super(provider.attributes.slice(*FIELDS)` in `#new` and pass the provider in when you do `AboutYourOrganisationForm.new(provider)`

app/forms/publish_interface/about_your_organisation_form.rb
malcolmbaig: yes, can do

app/forms/publish_interface/about_your_organisation_form.rb
malcolmbaig: In order to address the [bug](https://github.com/DFE-Digital/teacher-training-api/pull/2340#pullrequestreview-833255533) identified by @defong  I think we need slightly different logic when building the form with the persisted provider vs building it with user-submitted params. I've opted to have two different build methods to accommodate this. WDYT?

darokel: I've got a PR to simplify it even further. Have a look at this https://github.com/DFE-Digital/teacher-training-api/pull/2354. It uses a pattern we developed at Register as we already solved these types of issues. Happy to discuss the approach

malcolmbaig: @darokel looks good. This PR is getting unwieldy — I suggest we card up refactoring to that pattern, and get this merged as-is (pending approval for current set of changes).

app/forms/publish_interface/about_your_organisation_form.rb
```diff
@@ -0,0 +1,89 @@
...
+    include ActiveModel::Validations
+
+    FIELDS = %i[
+      train_with_us train_with_disability email telephone urn website ukprn
+      address1 address2 address3 address4 postcode region_code
+    ].freeze
+
+    attr_accessor :provider
+    delegate :recruitment_cycle_year, :provider_code, :provider_name, to: :provider
+    attr_accessor :_accredited_bodies_data
```
defong, line 13: umm should that be prefixed with underscore?

maybe also move to private

app/forms/publish_interface/about_your_organisation_form.rb
```diff
@@ -0,0 +1,89 @@
...
+      @provider.save
+    end
+
+    def promote_errors
+      @provider.errors.each do |error|
+        errors.add(error.attribute, error.full_message)
+      end
+    end
+
+    class AccreditedBody
```
defong, line 83: class per file?

app/forms/publish_interface/about_your_organisation_form.rb
```diff
@@ -0,0 +1,89 @@
...
+      form = new(params)
+      form.provider = provider
+      form._accredited_bodies_data = accredited_bodies_params
+      form.provider.assign_attributes(params)
+      form
+    end
+
+    def accredited_bodies
+      _accredited_bodies_data.map do |ab|
+        AccreditedBody.new(
```
defong, line 38:
```suggestion
        AccreditedBody.new(params: ab)
```
So this class has less code

app/forms/publish_interface/about_your_organisation_form.rb
```diff
@@ -0,0 +1,89 @@
...
+      @provider.valid?
+    end
+
+    def save
+      Provider.transaction do
+        update_provider
+        update_accrediting_enrichment
+      end
+      promote_errors
+    end
```
defong, line 56:
```suggestion
    def save
      update_accrediting_enrichment
      if valid?
        Provider.transaction do
         @provider.save
        end
     end
   end
```

 then you can drop ` @provider.save` in `update_accrediting_enrichment`

`promote_errors` feeling a little misplaced as generally `model.save` & `model.valid?` sets the `model.error`

Suggest that we move it to `valid?` to mimic expectation

defong (approved):
A few minor bits.. lets get this in and deal with the aftermath :muscle:

PR #2344, opened 2021-12-15
Add custom error pages

malcolmbaig:
### Context
We don't have GOVUK style error pages for these HTTP response codes in TTAPI.

https://trello.com/c/4PUK3wCt

- Copy the ones from Publish, along with the routing and controller code.
- Tweak the service name in the header so it makes sense across Support, New Publish, and these error pages.

Screenshot_20211215_154004-1

Screenshot_20211215_154047

Screenshot_20211215_154129

### Guidance to review
- Should be fairly easy to trigger a 404 and 403 in the review app, a 500 probably easier to do locally
- If testing locally, you'll need to set `config.consider_all_requests_local` to `false` in `development.rb`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/publish_interface/providers_controller.rb
+++ b/app/controllers/publish_interface/providers_controller.rb
@@ -1,6 +1,6 @@
 module PublishInterface
   class ProvidersController < PublishInterfaceController
-    rescue_from ActiveRecord::RecordNotFound, with: :not_found
+    rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
     before_action :build_recruitment_cycle
     before_action :build_provider, except: [:index]
 
--- a/app/controllers/publish_interface/providers_controller.rb
+++ b/app/controllers/publish_interface/providers_controller.rb
@@ -17,8 +17,9 @@ module PublishInterface
   private
 
     def build_recruitment_cycle
-      cycle_year = params[:recruitment_cycle_year] || params[:year] || Settings.current_cycle
-      @recruitment_cycle = RecruitmentCycle.find_by(year: cycle_year)
+      cycle_year = params[:recruitment_cycle_year] || params[:year] || Settings.current_recruitment_cycle_year
+
+      @recruitment_cycle = RecruitmentCycle.find_by!(year: cycle_year)
     end
 
     def build_provider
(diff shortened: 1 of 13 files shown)


app/controllers/publish_interface/providers_controller.rb
```diff
@@ -1,6 +1,6 @@
 module PublishInterface
```
darokel, line 1: Might be outside of this PR - but what's the rationale for this module naming? Given that this repo will essentially be the interface for publish as a whole. Would be good if we could stick to module naming based on areas of the app eg `Admin`, `Providers`, `Organisations`, `Support` etc

darokel (approved):
Left a comment. Could be addressed outside of this PR

defong (commented):
Pretty sure the other PR covered this 
image
:thinking:

malcolmbaig:
> Pretty sure the other PR covered this 

Yes, covered here https://github.com/DFE-Digital/teacher-training-api/pull/2340

app/controllers/publish_interface/providers_controller.rb
malcolmbaig: This borrows an idea from Apply. To keep the monolith manageable, Apply adopted an [approach](https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/adr/0005-application-structure.md) of separating distinct parts of the app into namespaces (CandidateInterface, ProviderInterface, SupportInterface, VendorAPI, etc). The idea was to establish clear boundaries and keep things loosely coupled between them so the codebase stays manageable/easy to understand over time.

To date, the project's done a pretty decent job of conforming to the structure outlined in that ADR, roughly along the lines of:

- Each namespace
  - Views, view components, controllers, form objects, shared services, mailers
- App layer
  - Models, shared services, shared view components

There's also a lightweight [spec](https://github.com/DFE-Digital/apply-for-teacher-training/blob/6ca59790422975a3c3b97c12276e57e1b9b59d5d/spec/lib/namespaces_spec.rb) for checking we aren't using namespaced code across boundaries.

This may make sense for this project, as its future state is likely to be along the lines of:

- PublishInterface
- SupportInterface
- API
- FindInterface (tbc)

darokel: Might be worth having a little dev meeting to discuss the approach? Conscious that different teams in DFE use different approaches & methodologies. Given that some of the publish devs are from different teams it's probably worth deciding on what we want to use for this project going forward (and possibly drawing up new ADRs to explain why we went with said approach)

PR #2345, opened 2021-12-15
[448] Rescue from discarded provider in support providers controller

darokel:
### Context

https://sentry.io/organizations/dfe-bat/issues/2827983994/?project=1377944&referrer=slack
https://trello.com/c/oX1yp6n4/448-server-error-when-accessing-deleted-provider

This error was triggered by trying to access a provider which was discarded from the support console. This may have been a red herring and somebody accessing the provider's courses via the url directly as the providers list in the support console already filters out discarded providers.

The API already returns a 404 so I think this may have just been an edge case with the scenario described above. I've added a fallback for now.

Screenshot 2021-12-15 at 18 08 51

### Guidance to review

- Try to access the same url the sentry error got triggered by `https://localhost:3001/support/providers/17139/courses` and it should redirect to the support providers list

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support/courses_controller.rb
+++ b/app/controllers/support/courses_controller.rb
@@ -3,6 +3,9 @@ module Support
     def index
       @courses = provider.courses.order(:name).page(params[:page] || 1)
       render layout: "provider_record"
+    rescue ActiveRecord::RecordNotFound
+      flash[:warning] = "Provider not found"
+      redirect_to support_providers_path
     end
 
     def edit
--- a/spec/features/support/providers/courses/viewing_providers_courses_spec.rb
+++ b/spec/features/support/providers/courses/viewing_providers_courses_spec.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "Viewing a providers courses" do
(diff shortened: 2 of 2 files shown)


darokel:
[Server error when accessing deleted provider](https://trello.com/c/oX1yp6n4/448-server-error-when-accessing-deleted-provider)

app/controllers/support/courses_controller.rb
```diff
@@ -3,6 +3,9 @@ class CoursesController < SupportController
     def index
       @courses = provider.courses.order(:name).page(params[:page] || 1)
       render layout: "provider_record"
+    rescue ActiveRecord::RecordNotFound
```
defong, line 6: Do we need a token test?

defong (approved):
image

app/controllers/support/courses_controller.rb
darokel: done https://github.com/DFE-Digital/teacher-training-api/pull/2345/commits/d89be83b5f20606a93cb6a5318f8607b033b548a

PR #2351, opened 2021-12-17
[487] Fix start date to validate against incorrect date values

george-bit:
### Context

As some tech debt to the previous start date edit function introduced for courses in the support console, there needs to be some further validation to date format.

This PR adds validation to the course model to prevent users from submitting 'impossible' dates

image

### Guidance to review

I've tried to be succinct as possible here but it's ended up being kinda difficult. Becasue this validation is only relevant to user input it needs to be scoped to the update form only. Therefore, I've had add guard clauses in to prevent validation breaking a lot of other tests that trigger when start date value is `nil`

Currently validation does not allow retention of erroneous value in input field, I'm not sure it's worth the effort and amount of code to put this in but open to discussion.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support/courses_controller.rb
+++ b/app/controllers/support/courses_controller.rb
@@ -14,14 +14,9 @@ module Support
 
     def update
       @edit_course_form = Support::EditCourseForm.new(course)
-
       @edit_course_form.assign_attributes(update_course_params)
-      @edit_course_form.save
 
-      @edit_course_form.valid?
-      @edit_course_form.course_valid?
-
-      if @edit_course_form.errors.none?
+      if @edit_course_form.save
         redirect_to support_provider_courses_path(provider)
       else
         render :edit
@@ -40,7 +35,7 @@ module Support
 
     def update_course_params
       params.require(:support_edit_course_form).permit(
-        :course_code, :name,
+        *EditCourseForm::FIELDS,
         :"start_date(3i)", :"start_date(2i)", :"start_date(1i)"
       ).transform_keys { |key| start_date_field_to_attribute(key) }
     end
--- a/app/controllers/support/courses_controller.rb
+++ b/app/controllers/support/courses_controller.rb
@@ -9,11 +9,19 @@ module Support
     end
 
     def edit
-      course
+      @edit_course_form = Support::EditCourseForm.new(course)
     end
 
     def update
-      if course.update(update_course_params)
+      @edit_course_form = Support::EditCourseForm.new(course)
+
+      @edit_course_form.assign_attributes(update_course_params)
+      @edit_course_form.save
+
+      @edit_course_form.valid?
+      @edit_course_form.course_valid?
+
+      if @edit_course_form.errors.none?
         redirect_to support_provider_courses_path(provider)
       else
         render :edit
@@ -31,7 +39,7 @@ module Support
     end
 
     def update_course_params
-      params.require(:course).permit(
+      params.require(:support_edit_course_form).permit(
         :course_code, :name,
         :"start_date(3i)", :"start_date(2i)", :"start_date(1i)"
       ).transform_keys { |key| start_date_field_to_attribute(key) }
--- a/app/controllers/support/courses_controller.rb
+++ b/app/controllers/support/courses_controller.rb
@@ -31,7 +31,19 @@ module Support
     end
 
     def update_course_params
-      params.require(:course).permit(:course_code, :name, :start_date)
+      params.require(:course).permit(
+        :course_code, :name,
+        :"start_date(3i)", :"start_date(2i)", :"start_date(1i)"
+      ).transform_keys { |key| start_date_field_to_attribute(key) }
+    end
+
+    def start_date_field_to_attribute(key)
+      case key
+      when "start_date(3i)" then "day"
+      when "start_date(2i)" then "month"
+      when "start_date(1i)" then "year"
+      else key
+      end
     end
   end
 end
--- a/app/controllers/support/courses_controller.rb
+++ b/app/controllers/support/courses_controller.rb
@@ -42,9 +42,9 @@ module Support
 
     def start_date_field_to_attribute(key)
       case key
-      when "start_date(3i)" then "day"
-      when "start_date(2i)" then "month"
-      when "start_date(1i)" then "year"
+      when "start_date(3i)" then "start_date_day"
+      when "start_date(2i)" then "start_date_month"
+      when "start_date(1i)" then "start_date_year"
       else key
       end
     end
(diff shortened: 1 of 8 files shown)


defong:
[[Bug] Fix support console start date to validate against incorrect date values](https://trello.com/c/BdwCqzK2/487-bug-fix-support-console-start-date-to-validate-against-incorrect-date-values)

app/models/course.rb
```diff
@@ -765,6 +767,29 @@ def set_applications_open_from
...
+  end
+
+  def convert_start_date(date_args)
+    return true if date_args.any?(&:nil?)
+
+    begin
+      self.start_date = Date.civil(year.to_i, month.to_i, day.to_i)
+    rescue ArgumentError
+      false
+    end
```
defong, line 785: No, use of exception as a control flow is an antipattern.

defong (changes requested):

darokel (changes requested):
Discussed offline, think we should just implement via form object here. Would be good to keep validations out of the model given convo we had in Slack

app/controllers/support/courses_controller.rb
```diff
@@ -9,11 +9,19 @@ def index
...
-      if course.update(update_course_params)
+      @edit_course_form = Support::EditCourseForm.new(course)
+
+      @edit_course_form.assign_attributes(update_course_params)
+      @edit_course_form.save
+
+      @edit_course_form.valid?
+      @edit_course_form.course_valid?
+
+      if @edit_course_form.errors.none?
```
defong, line 24: This reads really messy

The mechanism of saving should of cause validation to occur which in turns leads to a boolean denoting has it been successful or a failure with errors

defong (changes requested):

app/controllers/support/courses_controller.rb
darokel: yup would be good if this is tucked away into the form so the controller just calls `save` and handles the outcome (redirect or re-render)

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,80 @@
+module Support
+  class EditCourseForm
+    include ActiveModel::Model
+    include ActiveModel::Validations
+
+    attr_accessor :course_code, :name, :day, :month, :year
```
darokel, line 6: Could we dine a const `FIELDS` for these like we do in the other forms. Then your `attr_accessor` could just be `attr_accessor(*FIELDS)` and your controller can permit the same fields

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,94 @@
...
+  private
+
+    def check_start_date
+      date_args = [year, month, day].map(&:to_i)
+
+      begin
+        Date.new(*date_args)
+      rescue ArgumentError
+        Struct.new(:day, :month, :year).new(day, month, year)
+      end
```
defong, line 54:
```suggestion
        Date.valid_date?(*date_args) ? Date.new(*date_args) : Struct.new(:day, :month, :year).new(day, month, year)
```

Pointless `block`

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,94 @@
...
+    def save
+      valid?
+      course_valid?
+      return false unless errors.none?
+
+      @course.save
+    end
+
+    def course_valid?
+      assign_attributes_to_course
```
defong, line 35: I don't agree with this, assigning attributes is one thing, checking is it valid is another. why coupled?

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,94 @@
...
+  class EditCourseForm
+    include ActiveModel::Model
+    include ActiveModel::Validations
+
+    FIELDS = %i[
+      course_code name
+    ].freeze
+
+    attr_accessor(*FIELDS)
+    attr_accessor :day, :month, :year
```
defong, line 11:
```suggestion
    attr_accessor :start_date_day, :start_date_month, :start_date_year
```

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,94 @@
...
+
+    def save
+      valid?
+      course_valid?
+      return false unless errors.none?
+
+      @course.save
+    end
+
+    def course_valid?
```
defong, line 34: you probably want

```ruby
def valid?
  super()
  course.valid?
  promote_errors
  errors.none?
end

```

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,94 @@
...
+          course_code: course_code,
+          name: name,
+        )
+      end
+    end
+
+    def promote_errors
+      @course.errors.each do |error|
+        errors.add(error.attribute, error.message)
+      end
```
defong, line 77:
```suggestion
      errors.merge!(course.errors)
```

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,94 @@
...
+          course_code: course_code,
+          name: name,
+          start_date: start_date,
+        )
+      else
+        @course.assign_attributes(
+          course_code: course_code,
+          name: name,
+        )
+      end
```
defong, line 71:
```suggestion

      attributes = {
          course_code: course_code,
          name: name,
      }
      
      attributes[:start_date: start_date] = start_date: start_date      if valid_date? || date_args_blank?
        
       course.assign_attributes(attributes)      
```

app/forms/support/edit_course_form.rb
```diff
@@ -0,0 +1,94 @@
...
+      begin
+        Date.new(*date_args)
+      rescue ArgumentError
+        Struct.new(:day, :month, :year).new(day, month, year)
+      end
+    end
+
+    def assign_attributes_to_course
+      # not ideal but whilst validation is in 2 places this should prevent overlap
+      # this can be removed when we migrate validation over to just the form
```
defong, line 59: Not sure

defong (changes requested):

app/forms/support/edit_course_form.rb
george-bit: because attributes are coupled to the form not the course model. In order to check the validity of attr on the course before saving need to assign them

app/forms/support/edit_course_form.rb
george-bit: So the problem with renaming these attrs is that it makes the model incompatible with:

```
      <%= f.govuk_date_field(:start_date,  legend: { text: "Start date" }, width: 10) %>
```
The `day`, `month` and `year` are methods which used when rendering the form after erros are hit.

With that in mind I don't think  we should rename these. Wdyt?

defong (approved):

darokel (approved):
Nice clean up. Would be good to re-organise some of the lets in the form spec but can be done at another time

PR #2358, opened 2021-12-22
[494] Refactor contact details

darokel:
### Context

- https://trello.com/c/DZv3JPgV/494-refactor-contact-details-to-use-baseproviderform

- Refactors the contact details in the about your organisation section to use new form/controller pattern
- Removes references to contact fields from about your organisation form/controller
- Tests the contact details in isolation in its own feature spec
- Re-list contact details specific validations in its form object. These can be removed from the model once this is live

### Guidance to review

- Updating the contact details works the same as before

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


darokel:
[Refactor Contact details to use BaseProviderForm](https://trello.com/c/DZv3JPgV/494-refactor-contact-details-to-use-baseproviderform)

app/models/provider.rb
```diff
@@ -140,7 +140,7 @@ def current_accredited_courses
 
   validates :provider_type, presence: true
 
-  validates :telephone, phone: { message: "^Enter a valid telephone number" }, if: :telephone_changed?
+  validates :telephone, phone: { message: "Enter a valid telephone number" }, if: :telephone_changed?
```
malcolmbaig, line 143: Does this have any side effects in Publish? The carat is a convention introduced by https://github.com/DFE-Digital/teacher-training-api/blob/master/Gemfile#L82, and I wonder if this changes the error messages Publish receives?

malcolmbaig: To clarify - this is relevant because Old Publish will continue serving contact-editing functionality until we start redirecting to New Publish.

config/routes.rb
```diff
@@ -33,12 +33,11 @@
       resources :recruitment_cycles, param: :year, constraints: { year: /#{Settings.current_recruitment_cycle_year}|#{Settings.current_recruitment_cycle_year + 1}/ }, path: "", only: [] do
         get "/about", on: :member, to: "providers#about"
         put "/about", on: :member, to: "providers#update"
-        get "/contact", on: :member, to: "providers#contact"
-        put "/contact", on: :member, to: "providers#update"
         get "/details", on: :member, to: "providers#details"
 
         scope module: :providers do
           resource :visas, only: %i[edit update], path: "/visas"
+          resource :contact, only: %i[edit update], path: "/contact"
```
malcolmbaig, line 40: Worth bearing in mind - this alters the path structure from Old Publish slightly (GET `/contact` in Old Publish vs. GET `/contact/edit` here). Not a huge deal but potentially makes writing controller redirects in Old Publish a bit more fiddly.

malcolmbaig (commented):
LG - couple of comments left inline.

app/controllers/publish_interface/providers/contacts_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+          flash[:success] = I18n.t("success.published")
+
+          redirect_to details_publish_provider_recruitment_cycle_path(
+            provider.provider_code,
+            recruitment_cycle.year,
+          )
+        else
+          render :edit
+        end
+      end
```
malcolmbaig, line 21: Both of these actions need to authorize access, like so https://github.com/DFE-Digital/teacher-training-api/pull/2358/files#diff-c8f9f288e415f332553b657513de9c2842f029c15424dac1462667cf024a26ddR22

malcolmbaig: Come to think of it, the Visas controller needs to do the same

darokel: done

app/models/provider.rb
darokel: i've just added messages unique to the form, undid the original changes of removing the carats

config/routes.rb
darokel: How fiddly? I think we should try to keep the new routes more RESTful. There's a bit of an anti-pattern we'd be porting over with the old routes and if there's an opportunity to improve them without too much cost I'd rather take advantage of it now

defong: users may have bookmarked these

dankmitchell: Add a simple redirect in this app from the old path to the new more restful URL?

darokel: I've just kept this backwards compatible for now https://github.com/DFE-Digital/teacher-training-api/pull/2358/commits/0e48b9b00e5f68204193f774f615192d56ed5f93, we can improve the routes after the migration

darokel:
@dankmitchell Do you want to give this a final pair of 👀

dankmitchell (approved):
LGTM

PR #2397, opened 2022-01-06
Match headers in Old and New Publish

malcolmbaig:
### Context
We're migrating Old Publish to New Publish one piece at a time. The top nav of both apps need to look the same and contain the same info.

Commits provide more detail, but essentially:

- Copy the implementation from Old Publish, reworking it to comply with the component-based header we have in this codebase
- Ensure this header works across Support, Publish Interface, and the sign in page
- Make sure we're linking to Old Publish as the homepage and notifications link, for now

### Guidance to review
- Should be straightforward to review in a Review app, worth covering sign in page, support, and a publish interface page via one of the personas (manual editing of the URL required to see anything other than the empty root page).

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


dankmitchell:
[Make navigation bar in New Publish the same as Old Publish](https://trello.com/c/J8qvzsEn/482-make-navigation-bar-in-new-publish-the-same-as-old-publish)

spec/components/phase_banner/view_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  }.each do |environment, colour|
+    it "renders a #{colour} phase banner for the #{environment} environment" do
+      allow(Settings.environment).to receive(:name).and_return(environment)
+      render_inline(described_class.new)
+
+      expect(page).to have_selector(".govuk-phase-banner .govuk-tag--#{colour}")
+    end
+  end
+
+  it "renders different message text if in sandbox_mode" do
```
darokel, line 21: This feels like it should be split across two scenarios?

darokel (approved):

PR #2408, opened 2022-01-10
[510] Display user fullname on support console and include provider list

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/BlEpvF93/510-modify-user-interface-on-sc)

- Change first/last name on user list to display fullname
- Display the provider list on the user page

### Guidance to review

- Log onto support console and see the changes on the User tab

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/webpacker/styles/application.scss
+++ b/app/webpacker/styles/application.scss
@@ -4,3 +4,11 @@
 
 @import "status-box";
 @import "summary-list";
+
+.fixed-table {
+  table-layout: fixed;
+}
+
+.break-email {
+  word-break: break-all;
+}
--- a/app/webpacker/styles/application.scss
+++ b/app/webpacker/styles/application.scss
@@ -4,11 +4,3 @@
 
 @import "status-box";
 @import "summary-list";
-
-.fixed-table {
-  table-layout: fixed;
-}
-
-.break-email {
-  word-break: break-all;
-}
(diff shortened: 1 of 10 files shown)


stellenberger:
[Modify 'User' interface on SC](https://trello.com/c/BlEpvF93/510-modify-user-interface-on-sc)

spec/features/support/users/users_list_spec.rb
```diff
@@ -9,6 +9,8 @@
     given_i_am_authenticated(user: user)
     when_i_visit_the_users_index_page
     then_i_should_see_a_table_of_users
+    and_i_click_the_full_name
```
darokel, line 12: this should probably read some thing like:

```
when_i_click_the_name
then_it_takes_me_to_the_users_page
```

darokel (commented):
I've left a couple of comments.

I think as a general rule we should avoid adding presentational logic in the database models and rely on ViewComponents to achieve this. It's probably fine here for this case but if we need to do something similar we should divert 

Using ViewComponents also means you could test the presentational logic directly in the spec for the component and keep the feature spec addressing the scenario at a higher level

app/webpacker/styles/application.scss
```diff
@@ -4,3 +4,11 @@
 
 @import "status-box";
 @import "summary-list";
+
+.fixed-table {
```
darokel, line 8: This feels out of place here in the main `application.scss`. It should probably live in a `utilities` sub directory

spec/features/support/users/users_list_spec.rb
stellenberger: Agreed, added in

darokel (approved):

mbocc (approved):
Fab work, all works and looks great!

PR #2418, opened 2022-01-12
[545] bug header spacing bug

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/KMrL2eOn/545-bug-header-spacing-bug)

- Comment out the navigation bar for time being (until more items are introduced)
- Fix the spacing issue on the header

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/components/header/style.scss
```diff
@@ -38,3 +38,7 @@
     border-bottom-color: govuk-colour("purple");
   }
 }
+
+#navigation {
```
darokel, line 42: I think we can just use a govuk utility class to zero out the margin `govuk-!-margin-bottom-0` on the element or something. Also we should probably avoid targeting id selectors due to the specificity

darokel (changes requested):

stellenberger:
[[Bug] Header spacing bug](https://trello.com/c/KMrL2eOn/545-bug-header-spacing-bug)

stellenberger:
[Remove the gray NavigationBar with one item from Support Console](https://trello.com/c/BZWIN6gl/549-remove-the-gray-navigationbar-with-one-item-from-support-console)

darokel (approved):
Not sure why that database diagram is being added. Could we remove it from the diff?

PR #2443, opened 2022-01-19
Verify authorization in PublishInterface

malcolmbaig:
### Context
The original Trello card called for better authorization specs. I looked at a few different ways of doing this and settled on a different approach to help ensure we're authorizing access correctly.

https://trello.com/c/nb8N8Hk3

Writing specs for each authorization case is time consuming and easily missed. Instead, use Pundit's built-in `verify_authorized` method in the PublishInterfaceController as an `after_action`. This raises an error if we forget to call authorize in any action. If we don't need authorization (eg—static content linked from the footer, cc: @defong ), we can call `skip_authorization`.

Paired with a very light feature spec, this should give us enough certainty that we're authorizing correctly, without the ongoing need to keep our specs up to date.

### Guidance to review
- Suitable approach?

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/publish_interface/providers_controller.rb
+++ b/app/controllers/publish_interface/providers_controller.rb
@@ -4,7 +4,9 @@ module PublishInterface
     before_action :build_recruitment_cycle
     before_action :build_provider, except: [:index]
 
-    def index; end
+    def index
+      authorize :provider, :index?
+    end
 
     def details
       authorize @provider, :show?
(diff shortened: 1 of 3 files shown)


darokel (approved):

darokel:
Tbh it probably makes more sense to do it this way and assume every action should be locked down unless explicitly skipped

PR #2518, opened 2022-02-15
Update details.html.erb

Kateharries:
Removed the 'View on a course' grey text box

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


app/views/publish/providers/details.html.erb
```diff
@@ -151,11 +151,5 @@
     <% end %>
   </div>
```
defong, line 153:
```suggestion
```

PR #2526, opened 2022-02-22
[621] Port over locations editing from course page

darokel:
### Context

- https://trello.com/c/ZNQzq1dZ/621-migrate-courses-course-show-page-updating-locations

- Port over and simplify locations selection logic for the course
- Refactor markup to use govuk form builder
- Add test coverage with feature spec

### Guidance to review

Original implementation here: https://qa.publish-teacher-training-courses.service.gov.uk/organisations/2AT/2022/courses/AK12/locations

Ported version: https://teacher-training-api-pr-2526.london.cloudapps.digital/publish/organisations/2AT/2022/courses/AK12/locations

Check:

- You can update locations
- Error message is triggered if you try to remove all locations

FYI the full address is shown in the hint for the location in new publish which includes the location name but is excluded in old publish. Not sure which is right, @mbocc or Adam to confirm?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -345,6 +345,10 @@ en:
               too_long: "Reduce the word count for fee details"
             financial_support:
               too_long: "Reduce the word count for financial support"
+        publish/course_location_form:
+          attributes:
+            site_ids:
+              no_locations: "Select at least one location"
   errors:
     messages:
       email: "^Enter an email address in the correct format, like name@example.com"
--- a/app/components/notification_banner/view.rb
+++ b/app/components/notification_banner/view.rb
@@ -68,7 +68,7 @@ module NotificationBanner
     end
 
     def success_banner?
-      type == :success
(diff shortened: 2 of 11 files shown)


config/locales/en.yml
```diff
@@ -345,6 +345,10 @@ en:
               too_long: "Reduce the word count for fee details"
             financial_support:
               too_long: "Reduce the word count for financial support"
+        publish/course_location_form:
```
goodviber, line 348: Namespacing seems unusual?

goodviber (approved):
lgtm, just the one comment

config/locales/en.yml
darokel: We need to specify the module path so the translations can be picked up correctly by Rails. It expects this pattern when customising activemodel error messages under module namespaces

PR #2618, opened 2022-04-20
[900] Only return non discarded sites for a course

darokel:
### Context

https://trello.com/c/V6Mm4POW/900-issue-with-location-1n1

Discarded sites were still appearing under the course's details page. Eg "Hanson Academy" is discarded but still appears here https://www2.publish-teacher-training-courses.service.gov.uk/publish/organisations/1N1/2022/courses/238T/details.

- Only return non discarded sites for a course

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


darokel:
[Issue with location (1N1)](https://trello.com/c/V6Mm4POW/900-issue-with-location-1n1)

app/models/course.rb
```diff
@@ -106,7 +106,7 @@ def set_subject_position(course_subject)
   has_many :site_statuses
   accepts_nested_attributes_for :site_statuses
   has_many :sites,
-           -> { distinct.merge(SiteStatus.where(status: %i[new_status running])) },
+           -> { kept.distinct.merge(SiteStatus.where(status: %i[new_status running])) },
```
defong, line 109: Can we have a token test?

darokel: i don't really see it adding much value? Behaviour has already been tested in other models. I'd just be re-testing implementation logic of the gem at this point

defong (approved):

PR #2638, opened 2022-05-06
Add rollover env

RMcVelia:
### Context

Create rollover env to do a dry run of the rollover process

https://trello.com/c/N3XXsl9o/946-rollover-environment-for-ttapi-new-publish

Create required config files for the new env
Update Makefile
Add rollover to build-and-deploy workflow

Also,
-updated AZ gh secret
-checked AZ app/infra kv up to date

### Guidance to review

confirm config/settings/rollover.yml
confirm AZ app secrets 
manual make of rollover env 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/config/settings/rollover.yml
+++ b/config/settings/rollover.yml
@@ -1,0 +1,28 @@
+gcp_api_key: please_change_me
+publish_api_url: https://teacher-training-api-rollover.london.cloudapps.digital
+publish_url: https://teacher-training-api-rollover.london.cloudapps.digital
+find_url: https://find-rollover.london.cloudapps.digital
+
+# URL of this app for the callback after sigining in
+base_url: https://teacher-training-api-rollover.london.cloudapps.digital
+
+search_ui:
+  base_url: https://find-rollover.london.cloudapps.digital
+
+dfe_signin:
+  issuer: https://pp-oidc.signin.education.gov.uk
+  profile: https://pp-profile.signin.education.gov.uk
+  secret: please_change_me # Override with SETTINGS__DFE_SIGNIN__SECRET
+  user_search_url: https://pp-support.signin.education.gov.uk/users
+
+bg_jobs:
+  save_statistic:
+    cron: "0 0 * * *" # daily at midnight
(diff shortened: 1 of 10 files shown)


config/settings/rollover.yml
```diff
@@ -0,0 +1,28 @@
+gcp_api_key: please_change_me
```
darokel, line 1: Are we going to inject the values for these via `ENV`?

darokel: It might be worth checking what's in ENV["SETTINGS__*"] for the other environment so we can cover any missing values

darokel (approved):

config/settings/rollover.yml
RMcVelia: Yes, the value is in the app keyvault secret. The only difference between rollover and staging is that staging also has the BigQuery settings. As we aren't going to send data to BQ, that should be ok.

saliceti (approved):

saliceti (approved):

config/settings/rollover.yml
```diff
@@ -0,0 +1,28 @@
...
+publish_api_url: https://teacher-training-api-rollover.london.cloudapps.digital
+publish_url: https://teacher-training-api-rollover.london.cloudapps.digital
```
saliceti, line 3: Shouldn't it be the new publish domain?

terraform/modules/paas/main.tf
```diff
@@ -60,6 +60,13 @@ resource cloudfoundry_route web_app_cloudapps_digital_route {
   hostname = local.web_app_name
 }
 
+resource cloudfoundry_route web_app_cloudapps_digital_route2 {
```
saliceti, line 63: As discussed IRL, we could build a list of cloudapps hostnames and use it in the route resource above

RMcVelia: done

saliceti (approved):

PR #2644, opened 2022-05-10
458  Update rollover code so that user permissions are correctly maintained across cycles

goodviber:
### Context
At rollover we create new provider records. Now that we've updated the user permissions model, we need to ensure that permissions are correctly sustained across cycles.

What needs to be done
Update the relevant rollover code to re-create or update user_permissions records so that they point to the new provider records.
Add users to the new rolled over providers in `copy_to_recuitment_cycle_service.rb`
### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/spec/services/providers/copy_to_recruitment_cycle_service_spec.rb
+++ b/spec/services/providers/copy_to_recruitment_cycle_service_spec.rb
@@ -17,6 +17,7 @@ describe Providers::CopyToRecruitmentCycleService do
     }
     let(:provider) {
       create :provider,
+             :with_users,
              sites: [site],
              ucas_preferences: ucas_preferences,
              contacts: contacts
@@ -111,6 +112,12 @@ describe Providers::CopyToRecruitmentCycleService do
       service.execute(provider: provider, new_recruitment_cycle: new_recruitment_cycle)
 
       expect(new_provider.organisation).to eq provider.organisation
+    end
+
+    it "assigns the new provider to users" do
+      service.execute(provider: provider, new_recruitment_cycle: new_recruitment_cycle)
+
+      expect(new_provider.users).to eq provider.users
     end
 
     it "copies over the ucas_preferences" do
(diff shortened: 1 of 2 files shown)


goodviber:
[(Part 6) Update rollover code so that user permission are correctly maintained across cycles](https://trello.com/c/84nPOsbJ/458-part-6-update-rollover-code-so-that-user-permission-are-correctly-maintained-across-cycles)

spec/services/providers/copy_to_recruitment_cycle_service_spec.rb
```diff
@@ -113,6 +114,12 @@
       expect(new_provider.organisation).to eq provider.organisation
     end
 
+    it "assigns the new provider to users" do
+      service.execute(provider: provider, new_recruitment_cycle: new_recruitment_cycle)
+
+      expect(new_provider.users).to eq provider.users
```
defong, line 120: Can you give some stats for how many providers and users will be rollovered for next cycle?

defong (approved):
LGTM

spec/services/providers/copy_to_recruitment_cycle_service_spec.rb
goodviber: As in this? https://github.com/DFE-Digital/teacher-training-api/blob/8ad1ccd7a7f613634be09b883d33879a608fdca4/app/services/providers/copy_to_recruitment_cycle_service.rb#L38-L42

goodviber: Or do you mean in the factory trait?

defong: yeah

goodviber: I guess you could have `expect(new_provider.users.count).to eq 4` but not sure what value that adds?

PR #2662, opened 2022-05-18
replace internal BQ code with dfe-analytics gem

misaka:
### Context

We have a new gem [DFE-Analytics](https://github.com/DFE-Digital/dfe-analytics) to handle sending events our analytics platform! 

Use the new gem to send events to our analytics platform.

### Guidance to review

The major changes here is to remove the bespoke classes that emit events to BigQuery and replace them with the implementation in `dfe-analytics`. That implementation also requires that all fields be accounted for, either in the already-existing `analytics.yml` file or, for fields not being sent, in the new `analytics_blocklist.yml`.

This gem provides some additional rake tasks:

```bash
rake dfe:analytics:check                                                                  # Generate a new field blocklist containing all the fields not listed for sending to Bigquery
rake dfe:analytics:import_all_entities[sleep_time,batch_size]                             # Send Analytics events for the (allowlisted) state of all records in the database
rake dfe:analytics:import_entity[model_name,sleep_time,batch_size,start_at_id]            # Send Analytics events for the state of all records in a specified model
rake dfe:analytics:regenerate_blocklist                                                   # Generate a new field blocklist containing all the fields not listed for sending to Bigquery
```
and some RSpec matchers, which were used in specs in this PR.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


misaka:
This PR is awaiting changes to the `dfe_analytics` gem, which are currently PRed https://github.com/DFE-Digital/dfe-analytics/pull/11 and in review. Once those are live I'll make this PR official.

misaka:
[Roll out BigQuery client code across BAT services (dfe-analytics)](https://trello.com/c/neK2rSqu/563-roll-out-bigquery-client-code-across-bat-services-dfe-analytics)

config/settings.yml
```diff
@@ -79,7 +79,7 @@ google:
   bigquery:
     project_id: replaceme
     dataset: replaceme
-    api_json_key: ""
+    api_json_key: "{}"
```
defong, line 82: why empty object? can it not be nil?

spec/requests/api/dfe_analytics_events_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      get "/api/test", to: "dfe_analytics_events_test_api#test"
+      get "/api/public/v1/test", to: "dfe_analytics_events_test_public_api#test"
+    end
+  end
+
+  after do
+    Rails.application.reload_routes!
+  end
+
+  describe APIController do
```
defong, line 39: Should there be a normal none api?

misaka: There are feature tests.

defong (approved):

Gemfile
```diff
@@ -138,6 +141,8 @@ group :development, :test do
   # Help eliminate N+1 queries
   gem "bullet"
 
+  gem "debug"
```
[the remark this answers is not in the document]
misaka: what, really? Like, not a sarcastic "noice!"?

config/settings.yml
misaka: It should be an object, at least that's what the API key is that we use here. I don't remember what broke as a result of it being `nil`, but this is technically More Correct.

misaka:
Tested in review app with events going to `qa`.

misaka:
Discovered an issue while checking the API integration ... 

> NameError at /api/public/v1/recruitment_cycles/2022/courses
>
> undefined local variable or method `current_user' for #<API::Public::V1::CoursesController:0x00000000045628>

Looks like `dfe-analytics` has a hard dependency on `current_user`. This may require a change in the gem, although we could also work around it for now by implementing a `current_user` function that returns a dummy value for now. Not a huge fan of the latter though.

misaka:
@defong could I pls get a review of the last commit, make sure it makes sense?

misaka:
Just to show that this now works with the public API, here's the event in BigQuery:

image

app/controllers/api_controller.rb
```diff
@@ -1,9 +1,12 @@
...
+  include DfE::Analytics::Requests
   include ActionController::HttpAuthentication::Token::ControllerMethods
   include Pundit::Authorization
 
   # child must define authenticate method
   before_action :authenticate
   after_action :verify_authorized
+
+  # Used by dfe-analytics to (try and) log the user triggering an event.
+  def current_user; end
```
defong, line 11: This is the tech debt that stems from a 3rd party dependency that we have control with. It's also fostering bad practice.

Can we just update `dfe-analytics`?

Solve the upstream so we don't pollute the downstream. :skull_and_crossbones: :ocean:

defong (changes requested):
> Not a huge fan

Agreed

app/controllers/api_controller.rb
misaka: I was going with the advice given in the `dfe-analytics` gem which was to provide an empty `current_user` method:

https://github.com/DFE-Digital/dfe-analytics#controllers

```ruby
  # This method MUST be present in your controller and should return
  # either nil or an object implementing an .id method.
  #
  # def current_user; end
```

Can we just merge this in for now and look to fix later? I'm happy to have a convo with @duncanjbrown about whether this really is the right approach.

misaka: PR merge, will cut a new release shortly.

defong (approved):

PR #2667, opened 2022-05-19
[63] Set up smoke test for find

darokel:
### Context

We had an incident with a recent PR merge which affected the Find service's course page. We decided on having a smoke test as part of this repo's build process so we can catch any potential issues early when modifying code Find is dependant on.

- Add a smoke test to query Find's results page and check that we can view a course

### Guidance to review

- You can run the test locally just make sure you set the base url correctly otherwise view it as part of the build process

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


darokel:
[Setup smoke test for find](https://trello.com/c/R9Ekh7Fw/63-setup-smoke-test-for-find)

spec/smoke/find/course_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def response
+    @response ||= self.class.get("/results?age_group=primary&l=2&subject_codes%5B%5D=00").parsed_response
+  end
+
+  def courses
+    response.css(".app-search-results .app-search-results__item .app-search-result__item-title").map do |course|
+      Course.new(course.css(".govuk-link").first["href"])
+    end
```
defong, line 31: can you use the `data-qa="course__link"` instead

spec/smoke/find/course_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "nokogiri"
+require "spec_helper_smoke"
+
+class HtmlParserIncluded < HTTParty::Parser
```
defong, line 6: Can we break the class per file please

spec/smoke/find/course_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def courses
+    response.css(".app-search-results .app-search-results__item .app-search-result__item-title").map do |course|
+      Course.new(course.css(".govuk-link").first["href"])
+    end
+  end
+end
+
+describe "Find Service Smoke Tests", :aggregate_failures, smoke: true do
+  let(:base_url) { FindResultsPage.base_uri }
```
defong, line 36: So when this test runs is it agains't live which makes the test one phase behind?

spec/smoke/find/course_spec.rb
darokel: This is so specific to this test that i don't see it adding value having it in its own file?

spec/smoke/find/course_spec.rb
darokel: It's environment specific and will run after a deploy of the latest changes to that env is made. So will target find-{env}

defong (approved):

spec/smoke/find/course_spec.rb
darokel: done

PR #2668, opened 2022-05-19
51 new navigation

avinhurry:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


darokel:
@simonwhatley Just wanted to check - for the new nav (and based on what i’ve seen in some other services), shouldn’t the nav sit right under the top header and in between the beta message? 

Prototype currently has it the other way round:

Screenshot 2022-05-23 at 12 28 49

app/views/layouts/application.html.erb
```diff
@@ -48,9 +48,40 @@
...
-       <%= render TitleBar::View.new(title: @provider.provider_name) %>
+    <% if FeatureService.enabled?(:new_publish_navigation) %>
+
+      <% if current_user %>
+        <% if current_user.has_multiple_providers_in_current_recruitment_cycle? && @provider && !request.path.include?('support') %>
+         <%= render TitleBar::View.new(title: @provider.provider_name) %>
+        <% end %>
+      <% end %>
+
+      <% if @provider && !current_page?(root_path) && !@provider.rolled_over? && @provider.accredited_body? && @provider.recruitment_cycle && !request.path.include?('support') %>
```
darokel, line 59: There's a bit of a logic soup here and a couple ways to clean this up:

1. Put the different nav views in partials (conditional methods should probably go into helpers)
2. Put them into sep view components and encapsulate their conditional logic in their `render?` methods

darokel, line 59: There's a bit of a logic soup here and a couple ways to clean this up:

1. Put the different nav views in partials (conditional methods should probably go into helpers)
2. Put them into sep view components and encapsulate their conditional logic in their `render?` methods

simonwhatley:
> @simonwhatley Just wanted to check - for the new nav (and based on what i’ve seen in some other services), shouldn’t the nav sit right under the top header and in between the beta message?
> 
> Prototype currently has it the other way round:
> 
> Screenshot 2022-05-23 at 12 28 49

@darokel the nav should sit below the beta banner.

spec/features/publish/accepting_terms_spec.rb
```diff
@@ -20,6 +21,10 @@
     then_i_should_see_an_error_message
   end
 
+  def given_the_new_publish_flow_feature_flag_is_enabled
+    allow(Settings.features).to receive(:new_publish_navigation).and_return(true)
```
darokel, line 25: I think we could make use of the `enable_features` helper. Eg:

```ruby
before do
  enable_features(: new_publish_navigation)
end

after do
  disable_features(:new_publish_navigation)
end
```

darokel, line 25: I think we could make use of the `enable_features` helper. Eg:

```ruby
before do
  enable_features(: new_publish_navigation)
end

after do
  disable_features(:new_publish_navigation)
end
```

darokel:
Also I think the background colour of the nav should span across the whole page? eg:

Screenshot 2022-05-25 at 14 48 04

simonwhatley:
> Also I think the background colour of the nav should span across the whole page? eg:
> 
> Screenshot 2022-05-25 at 14 48 04

Design notes:

- the thin blue line below the black bar spans the full viewport width
- the background grey of the primary navigation spans the full viewport width
- if the user is in a single org, the thin grey line below the beta banner is removed

app/views/layouts/application.html.erb
avinhurry: Thanks. I've gone with option 1 and it looks much neater now!

spec/features/publish/accepting_terms_spec.rb
avinhurry: Good shout too - have done

spec/support/page_objects/publish/providers_show.rb
```diff
@@ -11,6 +11,15 @@ class ProvidersShow < PageObjects::Base
       element :users, ".app-status-box"
       element :accredited_courses, '[data-qa="courses-as-an-accredited-body"]'
       element :allocations, '[data-qa="allocations"]'
+
+      # The new elements introduced with the new publish navigation are below. When the feature flag is removed the
+      # duplicated elements above can be removed.
+
+      element :nav_organisation_details, :link, "Organisation details"
```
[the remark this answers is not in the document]
avinhurry: We can do, but can we do it when we get rid of the feature flag as `ProvidersShow < PageObjects::Base` will become obsolete. We could maybe turn that into `PrimaryNav`?

I'm not a huge fan of all the abstraction that comes with using page objects, they make the tests much harder to read and write. Do we gain much ROI from using them?

avinhurry:
Thanks @darokel. I'll make the other changes as part of new cards

spec/support/page_objects/publish/providers_show.rb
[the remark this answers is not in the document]
darokel: Given that these are navigational links and appear everywhere on the site, i think it would make sense that they're contained in their own class. The problem with listing the nav links on this provider show page is that it's incorrect, they're not just owned by this page.

Re: page objects - i mean it's just Ruby and modelling parts of the application so tests can be more focused (and easier) to write. We can have a discussion about it but it's an approach that's becoming more standardised across DFE repos I think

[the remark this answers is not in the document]
avinhurry: That makes sense - thanks. I've pushed up some new commits. ♻️

darokel (approved):

PR #2675, opened 2022-05-20
SPIKE add Google Tag Manager config

SamJamCul:
### Context
https://trello.com/c/mmx8JnUU/47-ga-spike-add-a-new-course-custom-event

I've started, but not really finished, spiking out GA (Google Analytics) and GTM (Google Tag Manager) integrations for the app. GA is the main place where all the data is sent - and GTM is used to join up in-app events with the analytics platform, which are fed through to GA. 

I've had to get access to GA, and with this I was able to create a GTM account specifically for Publish. I'm not entirely sure what access to the GTM access looks like though, and it might need to be recreated to so that people other than me can make changes to it. 

I've added a basic GA-all-pages tag, as shown in the GTM screenshot below. With the new config in the `app/views/layouts/application.html.erb`, this will emit events to GA that show which pages are accessed. 
image

The aim of the spike, however, was to get specific button events tracked, which I haven't managed to do yet. This should be a matter of creating custom event triggers in GTM, and connecting those up to the required elements in the app.

It would also be great if we could segment these events by environment, so that there's no cross-contamination from testing and production. This is possible in GTM (I think) by using different containers for each environment, and adding some config that points to the appropriate container depending on the env.

Thanks to @defong and @AbigailMcP for pointing me in good directions. When testing, make sure you don't have something blocking your browser from sending out JS script stuff (like uBlock). There are a set of GTM PRs on Register which I've used as reference, [take a look at them here](https://github.com/DFE-Digital/register-trainee-teachers/pulls?q=is%3Apr+GTM)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -17,11 +17,12 @@
     <%= favicon_link_tag asset_pack_path("media/images/govuk-apple-touch-icon-180x180.png"), rel: "apple-touch-icon", type: "image/png", size: "180x180" %>
     <%= stylesheet_pack_tag "application", media: "all" %>
     <%= javascript_pack_tag "application", defer: true %>
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','GTM-TSK2T46');</script>
   </head>
 
   <body class="govuk-template__body <%= yield :body_class %>">
     <noscript>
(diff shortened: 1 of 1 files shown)


defong:
Still to do
- look into implementing it for `Course creation journey`
- segment deployment environments


Otherwise it looks all good as first cut

defong:
Still to do
- look into implementing it for `Course creation journey`
- segment deployment environments


Otherwise it looks all good as first cut

defong:
[GA Spike - "Add a new course" custom event](https://trello.com/c/mmx8JnUU/47-ga-spike-add-a-new-course-custom-event)

SamJamCul:
So here's the update - I've been able to connect specific events up with GA. This is primarily done through the Google Tag Manager interface, although I've been testing things on a local version, which makes it easier to debug. 

You'll need access to our GA and the associated GTM account. Within GTM, I've set up certain triggers based on the appropriate buttons - but these could be more closely defined. 

**This event triggers when a `govuk-button` containing the text `Add a new course` is clicked**
image

image

We can tie events to various different behaviours, all through GTM. This may be a bit difficult to do without running a local version to test out the exact variable names, but it's possible to configure everything from GTM, without making any changes to the site code. 

We can create separate triggers for each event we want to track - for example, we can set triggers up for every time a form is submitted under a specific url pattern, if we want to capture the "Adding a course" workflow.

PR #2679, opened 2022-05-20
61 new nav org switcher

avinhurry:
### Context

Following on from #2668, this PR adds a `Change organisation` button for providers who have multiple orgs.

- Add dynamic view component 
-  Render component if: current user exists, feature flag enabled, user org count greater than 1, provider not nil, not on a support page
- Add view preview page
- Duplicate tests and mark old file as ready to delete once feature flag removed

### Guidance to review

- Login as a multi org user and ensure you can change orgs using the new link on any page
- Repeat for a non multi org user and ensure you cannot
- Repeat for any of the support pages and ensure you cannot

### Screenshot

image

### DIFF
--- a/app/components/title_bar/script.js
+++ b/app/components/title_bar/script.js
@@ -1,0 +1,1 @@
+import './style.scss'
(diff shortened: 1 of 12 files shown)


avinhurry:
[New nav org switcher](https://trello.com/c/uDO1G5Kc/61-new-nav-org-switcher)

goodviber (approved):

PR #2682, opened 2022-05-23
Remove breadcrumbs

avinhurry:
### Context

As a follow up from #2668 where the new navigation bar is created and #2679 where the organisation switcher is implemented. This PR removes the breadcrumbs if the `new_publish_navigation` feature flag is activated, as those are now redundant.

### Guidance to review

- Review app is deployed with the flag on. Compare with prototype.


avinhurry:
[Remove breadcrumbs](https://trello.com/c/bfh13Dls/62-remove-breadcrumbs)

goodviber (approved):
lgtm

PR #2703, opened 2022-05-27
[85] Application start date is causing issues

darokel:
### Context

- https://trello.com/c/YSy4Ef3g/85-application-start-date-is-causing-issues

Some courses can have their `applications_open_from_date` less than the next cycle's applications start date. This PR handles this scenario by adjusting them to the next cycle's date.

Also (perhaps controversially) I've introduced two new settings to keep track of the next cycle's applications start/end date. These dates seem to be floated around internally across conversations but it might be a good idea to have them concretely defined in the application. The tradeoff is another a pair of properties to change each time rollover comes around but i think it's worth it to catch & handle edge cases like the above. 

Also fixed another issue when creating a course with a method missing in the course decorator.

### Guidance to review

- This issue occurs in the rollover environment. Load the same url in the review app `publish/organisations/2CG/2023/courses/D730/outcome` and attempt to save. It errors in the rollover env but saves in the review app.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -197,7 +197,7 @@ class CourseDecorator < ApplicationDecorator
       "As soon as the course is on Find (recommended)"
     else
       year = recruitment_cycle.year.to_i
-      day_month = Date.parse(recruitment_cycle.application_start_date).strftime("%-d %B")
+      day_month = recruitment_cycle.application_start_date.strftime("%-d %B")
       "On #{day_month} when applications for the #{year} to #{year + 1} cycle open"
     end
   end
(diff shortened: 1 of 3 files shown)


darokel:
[Application start date is causing issues](https://trello.com/c/YSy4Ef3g/85-application-start-date-is-causing-issues)

defong (changes requested):
It's unnecessary to introduce additional `application_start_date` and `application_end_date` as firstly when a recruitment cycle is created for next cycle we need to set the `application_start_date` and `application_end_date`

PR #2704, opened 2022-05-27
106 allow admins to use the org switcher

avinhurry:
### Context

As part of the new publish navigation, we are reworking the `Training partners` page and removing recruitment cycle text.

- Remove guidance text
- Rename H2
- Remove recruitment cycle text from all nav bar sections unless we are in rollover
- Keep these changes behind the feature flag

### Guidance to review

I've activated the feature flag in the review app, so look at the nav bar on that and see if everthing still looks ok.


avinhurry:
Incorrect branch name - reponed in #2705

PR #2708, opened 2022-05-30
108 css tweaks to and around the nav bar

avinhurry:
### Context

A few small tweaks in preparation for switching on the new navigation.

Note that when this is merged the app-header line will cover the full page width. This is not behind a feature flag. 

- Extend app-header line to cover full page width
- Grey background spans full viewport width
- Reduce spacing between change org link and primary nav bar

Only display the phase banner border:
- If there is a current user
- if the current user is not an admin
- if the current user is associated with more than one provider

## Before

image

## After

image

### Guidance to review

- Recommend reviewing commit by commit 

- Take a look at the nav bar and change org link as a multi org user, single org user and accredited body 
- Do the same for the phase banner border
- Compare and contrast with the prototype

### DIFF
--- a/app/components/phase_banner/style.scss
+++ b/app/components/phase_banner/style.scss
@@ -1,0 +1,5 @@
+.app-phase-banner--no-border {
+  .govuk-phase-banner {
+    border-bottom: 0;
+  }
+}
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -64,7 +64,11 @@ module ApplicationHelper
       end
     end
   end
-# rubocop:enable Rails/HelperInstanceVariable
+  # rubocop:enable Rails/HelperInstanceVariable
+
+  def display_phase_banner_border?(user)
+    user && !user.admin? && (user.providers_via_organisations.where recruitment_cycle: RecruitmentCycle.current).one?
+  end
 
 private
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -67,7 +67,7 @@ module ApplicationHelper
   # rubocop:enable Rails/HelperInstanceVariable
 
   def display_phase_banner_border?(user)
-    user && !user.admin? && (user.providers_via_organisations.where recruitment_cycle: RecruitmentCycle.current).one?
+    user && !user.admin? && user.providers.where(recruitment_cycle: RecruitmentCycle.current).one?
   end
 
 private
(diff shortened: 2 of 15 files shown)


avinhurry:
[CSS tweaks to and around the nav bar](https://trello.com/c/LCa6fnp8/108-css-tweaks-to-and-around-the-nav-bar)

darokel:
Two things:

1) I noticed the back buttons in places where you can change info about courses is in the wrong place?

Eg prototype:

https://publish-courses-prototype.herokuapp.com/organisations/96d90282-3ce7-4fc7-aa60-e03d7bf8a0f1/cycles/2022/courses/0e65fd3a-6c56-4b9d-882d-441b09776823/special-educational-needs-disability?referrer=details

Current build:

https://teacher-training-api-pr-2708.london.cloudapps.digital/publish/organisations/1A4/2022/courses/J084/about

2) The active state seems to lose its activeness once you go further into that area. Eg course index the courses link is active but loses this state when you're viewing an individual course. I think it's meant to stay active if you're still in the area

app/helpers/application_helper.rb
```diff
@@ -64,7 +64,11 @@ def enrichment_summary(summary_list, model, key, value, fields, truncate_value:
       end
     end
   end
-# rubocop:enable Rails/HelperInstanceVariable
+  # rubocop:enable Rails/HelperInstanceVariable
+
+  def display_phase_banner_border?(user)
+    user && !user.admin? && (user.providers_via_organisations.where recruitment_cycle: RecruitmentCycle.current).one?
```
darokel, line 70:
```suggestion
user && !user.admin? && (user.providers.where(recruitment_cycle: RecruitmentCycle.current)).one?
```

We've updated to just `users.providers` now and not using the organisation stuff

app/components/phase_banner/style.scss
```diff
@@ -0,0 +1,5 @@
+.app-phase-banner--no-border {
+  .govuk-phase-banner {
+    border-bottom: 0;
```
darokel, line 3: I'm wondering if there's a govuk css utility class to apply this rather than define a separate stylesheet?

darokel (commented):
Looking good just a few comments

app/components/phase_banner/style.scss
avinhurry: I'm not 100% sure but this is the way that Apply have done it, and this is also how it's done in the prototype so i'm happy with this

darokel (approved):

PR #2718, opened 2022-06-01
117 replace master with main

avinhurry:
### Context

We want to use `main` not `master`, this will bring us in line with the other BAT repos.

Replace all incidences of `master` in actions and `Makefile` with `main`.

Once this is in, we'll need to tell GitHub `main` is the primary branch: https://github.com/github/renaming

Removing a non existent GitHub link while we are here. 

### Guidance to review

Have I missed anything?


avinhurry:
[Replace `master` with `main`](https://trello.com/c/bBq1KFsv/117-replace-master-with-main)

PR #2724, opened 2022-06-06
Upgrade to Rails 7

darokel:
### Context

https://trello.com/c/7SE5OL6t/53-publish-bump-to-rails-7

This PR upgrades the project's Rails version to 7 to take advantage of new features and be consistent with other services moving towards the same direction. 

This PR mainly contains changes to support the version bump by updating framework or app related code/tests which doesn't align or cause issues preventing the upgrade. As this project is now more or less a full Rails application, I've configured it to be as such.

### Notes

**Webpacker** 

This is being replaced for Esbuild but will be done in a follow up ticket https://trello.com/c/ynkLCrTx/147-publish-replace-webpacker-with-es-build

**Ruby** 

We're planning to upgrade Ruby to the latest stable version `3.1.0` but that will be done in a follow up card https://trello.com/c/jp9NbynT/124-publish-bump-to-ruby-310

**Dependencies**

We've had to fork the `open-api-rswag` gem in order to loosen some of the dependency versions to support this upgrade. The gem also appears to be no longer maintained

We've also had to bump `rspec-rails` to the latest release candidate as it provides a fix to this issue https://github.com/rspec/rspec-rails/issues/2531.

### Guidance to review

This is more of a regression check

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


darokel:
[Publish - Bump to Rails 7](https://trello.com/c/7SE5OL6t/53-publish-bump-to-rails-7)

config/database.yml
```diff
@@ -2,9 +2,6 @@ default: &default
   adapter: postgresql
   encoding: unicode
   pool: <%= ENV.fetch('DATABASE_CONNECTION_POOL_SIZE', ENV.fetch('RAILS_MAX_THREADS', 5)) %>
-  variables:
```
darokel, line 5: We decided to remove this in the tech leads catch up.

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,25 @@
+# Be sure to restart your server when you modify this file.
+
+# Define an application-wide content security policy.
+# See the Securing Rails Applications Guide for more information:
+# https://guides.rubyonrails.org/security.html#content-security-policy-header
+
+# Rails.application.configure do
```
defong, line 7: Don't we need this?

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,120 @@
...
+#
+# Uncomment each configuration one by one to switch to the new default.
+# Once your application is ready to run with all new defaults, you can remove
+# this file and set the `config.load_defaults` to `7.0`.
+#
+# Read the Guide for Upgrading Ruby on Rails for more info on each option.
+# https://guides.rubyonrails.org/upgrading_ruby_on_rails.html
+
+# Temporarily disable this until we clean up the reference data stuff
+Rails.configuration.active_record.belongs_to_required_by_default = false
```
defong, line 13: Is this actually default?
If not can we move to its lonesome?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_04_21_115255) do
+ActiveRecord::Schema[6.1].define(version: 2022_04_21_115255) do
```
defong, line 13:
```suggestion
ActiveRecord::Schema[7.0].define(version: 2022_04_21_115255) do
```

app/models/course_enrichment.rb
```diff
@@ -27,7 +27,6 @@ def draft?
   end
 
   # About this course
-  # TODO POST MIGRATION: Move out of this as it's handled in the form object
```
defong, line 30: ?

app/controllers/support/providers_controller.rb
```diff
@@ -15,16 +15,19 @@ def create
...
-        # for invalid forms in the support console.
-        @provider.errors.messages.each { |k, v|
-          case k
-          when :"sites.urn", :email, :telephone
-            @provider.errors.messages[k] = v.first.gsub("^", "")
-          else
-            @provider.errors.messages[k] = "#{k.to_s.gsub(/.\.^?/, ' ').humanize} #{v.first}"
-          end
-        }
+        # TODO: Move this into a form object, the error linking is currently broken and the code below violates
```
defong, line 18: image
Is it this?

defong (commented):
Looking good just a few questions

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,4 @@
 # Be sure to restart your server when you modify this file.
 
-# Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += %I[password email first_name last_name]
+# Configure parameters to be filtered from the log file. Use this to limit dissemination of
```
SamJamCul, line 3: of?

config/initializers/new_framework_defaults_7_0.rb
SamJamCul: How bad is turning this on? It should probably live in `applicaiton.rb` rather than keeping this file alive

config/initializers/content_security_policy.rb
darokel: This is a new generated file, it wasn't here before. It'll be up to us to decide what to configure

config/initializers/new_framework_defaults_7_0.rb
darokel: I think i'd prefer to keep it here since this file is supposed to help with easing the upgrade. This is usually on by default but had to turn it off as it requires refactoring of https://github.com/DFE-Digital/teacher-training-api/blob/main/app/services/subjects/creator_service.rb#L2 which creates subjects without subject areas. Maybe it should be combined with this file https://github.com/DFE-Digital/teacher-training-api/blob/main/app/services/subjects/subject_area_creator_service.rb to achieve the same thing. But keeping it here means we can just remove it once we've fixed it.

Ideally this should be on, i wouldn't want to put it anywhere else which could mean we've decided that it shouldn't be true.

app/models/course_enrichment.rb
darokel: I removed these as I realise we need to keep the related validations since they get enforced when a course is attempted to be published.

app/controllers/support/providers_controller.rb
darokel: Yup carded up here https://trello.com/c/7ovcLHFw/143-publish-refactor-provider-form-in-support-console. The error linking wasn't really working properly to begin with.

config/initializers/filter_parameter_logging.rb
darokel: not sure what happened, reverted to original comment

config/initializers/new_framework_defaults_7_0.rb
SamJamCul: Sounds good as long as we have a card in the backlog for it

spec/models/course_spec.rb
```diff
@@ -39,7 +39,7 @@
       timestamp = 1.hour.ago
       course.provider.update updated_at: timestamp
       course.touch
-      expect(course.provider.updated_at).to eq timestamp
+      expect(course.provider.updated_at).to be_within(1.second).of timestamp
```
defong, line 42: Probably should use timecop

defong (approved):
> We've had to fork the open-api-rswag gem in order to loosen some of the dependency versions to support this upgrade. The gem also appears to be no longer maintained

I think we need to start making concession with regards to the docs as a whole

spec/models/course_spec.rb
darokel: I think the issue here is precision. Ruby time is a little more precise than what's saved in the database. The assertions were off by nanoseconds and failing in CI. See here for reference: https://tddium.wordpress.com/2011/08/07/rails-time-comparisons-devil-details-etc/

darokel: I think the issue here might be precision. Ruby time seems to be a little more precise than what's saved in the database. The assertions were off by nanoseconds and failing in CI. See here for example: https://tddium.wordpress.com/2011/08/07/rails-time-comparisons-devil-details-etc/

PR #2734, opened 2022-06-08
Remove pointless return nils

avinhurry:
### Context

We don't need to `return nil`, `return` implicitly returns nil.

- Remove all instances of `return nil` in the codebase and replace with just `return`

### DIFF
--- a/docs/lib/govuk_tech_docs/open_api/renderer.rb
+++ b/docs/lib/govuk_tech_docs/open_api/renderer.rb
@@ -286,7 +286,6 @@ module GovukTechDocs
 
       def get_schema_name(text)
         unless text.is_a?(String)
-          return nil
         end
 
         # Schema dictates that it's always components['schemas']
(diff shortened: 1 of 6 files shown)


docs/lib/govuk_tech_docs/open_api/renderer.rb
```diff
@@ -286,7 +286,6 @@ def get_operations(path)
 
       def get_schema_name(text)
         unless text.is_a?(String)
```
darokel, line 288: I think we want this to be something like:

`return unless text.is_a?(String)`

darokel (approved):

docs/lib/govuk_tech_docs/open_api/renderer.rb
avinhurry: Good catch!

avinhurry: Updated

avinhurry:
[Publish - Remove pointless return nils](https://trello.com/c/APVaumBe/130-publish-remove-pointless-return-nils)

PR #2741, opened 2022-06-09
#126 Enable code coverage with SimpleCov

goodviber:
### Context
We've got simplecov in the project but it doesn't appear to be setup locally to generate a coverage report.

#### What needs to be done
Tweak the setup similar to register [spec_helper.rb](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/spec/spec_helper.rb#L3) so we can run and produce a report locally to see the coverage.
#### Tech notes
Can we check that the minimum threshold right now is active? Eg does the report match the threshold?
Can we increase the threshold?

Config Simplecov to remove existing filter and save artifact in github actions

### Guidance to review
Current local results 90% . Not sure we need the COVERAGE env var, as there is little extra overhead from running the coverage report as default
Screenshot 2022-06-09 at 12 03 21

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


.gitignore
```diff
@@ -20,7 +20,6 @@
 
 .byebug_history
 .ruby-version
-coverage/
```
defong, line 23: I think we want it to be ignored otherwise it a load of noise

defong (changes requested):
There is no need to store generated reports in repo....

.gitignore
darokel: Yup i think we should be able to generate during the build run (after tests run) and store the artefact for viewing. We'll likely need to use the `COVERAGE` env or something if we want some flexibility to toggle. Also  90% is much better than what i was expecting, maybe we should increase the threshold to match instead of the current `75`

goodviber: I agree we should gitignore the coverage files, however with the current test commands in 'build-and-deploy.yml' this does not then generate the coverage directory for the upload coverage task. I have seen it done in other projects, will have a dig around

goodviber: Screenshot 2022-06-10 at 16 20 28

darokel (approved):

spec/spec_helper.rb
```diff
@@ -4,19 +4,19 @@
...
-# If running specs in parallel this ensures SimpleCov results appears
-# upon completion of all specs
-if ENV["TEST_ENV_NUMBER"]
-  SimpleCov.at_exit do
-    result = SimpleCov.result
-    result.format! if ParallelTests.number_of_running_processes <= 1
+if ENV.fetch("COVERAGE", false)
+  require "simplecov"
+
+  SimpleCov.minimum_coverage 85
```
darokel, line 10: Can we go for 90?

defong (approved):

spec/spec_helper.rb
goodviber: ok

PR #2742, opened 2022-06-09
129 publish setup auto correct on erb ruby linting

SamJamCul:
### Context
https://trello.com/c/GBcowlqH/129-publish-setup-auto-correct-on-erb-ruby-linting

Adding erb linting, and including it in the default rake linting

Added the `erb_lint` gem, a bit of config for it, and a rake task `rake lint:erb` to run it. It autocorrects by default, although erblint's autocorrect is a bit anemic.

Second commit is the result of me running erblint for the first time, and cleaning up the output. 

### Guidance to review
Check every file?

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/views/publish/courses/_basic_details_tab.html.erb
```diff
@@ -146,14 +146,14 @@
...
-          <% row.action() %>
+          <% row.action %>
         <% end %>
       <% end %>
 
       <% summary_list.row(html_attributes: { data: { qa: "course__name" } }) do |row| %>
         <% row.key { "Title" } %>
         <% row.value { course.name } %>
-        <% row.action() %>
+        <% row.action %>
```
goodviber, line 156: is this a safe change?

SamJamCul: I think so, ruby shouldn't need the brackets. For that `govuk_summary_list` thing, the [proper way to skip actions](https://design-system.service.gov.uk/components/summary-list/) is to change the class for the row to include `govuk-summary-list__row--no-actions` to the element class. But because we have dynamic decisions on whether there's an action, it looks like it's just easier to say there's an action, but provide no content. 

To be honest, I'm not sure what the better way of doing it is, but removing the brackets doesn't seem to break it :D

SamJamCul:
I've added another commit that includes `erblint` as part of the build and deploy bits - and it also turns off a certain linter, which was a bit buggy when toggling autocomplete on or off

.github/workflows/build-and-deploy.yml
```diff
@@ -162,6 +162,9 @@ jobs:
     - name: Run Ruby Linter
       run: docker-compose exec -T web /bin/sh -c 'bundle exec rubocop --format clang'
 
+    - name: Run ERB Linter
```
goodviber, line 165: lgtm

PR #2751, opened 2022-06-14
Publish refactor pages helper

avinhurry:
### Context

We use site prism, which requires adding explicitly defining a page object every time we test a new page. We would like to start automating this process. 

This PR aims to address this by using metaprogramming to define these methods on the fly using the file path.

`Note: There are a few more ways we can refactor this such as combining the two method generators into one and moving the DfE sign in page object into the parent folder, however as this PR now touches 108 files I thought it would be good to do as a follow up PR.`

- Use the file paths `spec/support/page_objects/publish` and `spec/support/page_objects/support` to grab the file names, process them into a more respectable form, and use this as the method name when defining the method on the fly.

- Rework the existing specs to make use of this new automatic method generator.

### Guidance to review

- Are we happy with the naming conventions used

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


avinhurry:
[Publish - Refactor pages helper](https://trello.com/c/md5wMWNo/125-publish-refactor-pages-helper)

darokel:
Nice work on this so far. The logic seems to be the same so maybe extract it out so it's not duplicated then we can define any number of namespaces by just updating the location of each one via `https://github.com/DFE-Digital/teacher-training-api/blob/8312feaf03921c48f55554cb6d8825c51825c63b/spec/support/feature_helpers/publish_pages.rb#L3`

avinhurry:
> Nice work on this so far. The logic seems to be the same so maybe extract it out so it's not duplicated then we can define any number of namespaces by just updating the location of each one via
> 
> https://github.com/DFE-Digital/teacher-training-api/blob/8312feaf03921c48f55554cb6d8825c51825c63b/spec/support/feature_helpers/publish_pages.rb#L3

As discussed offline, this will be done in a follow up PR

PR #2755, opened 2022-06-16
143 update error messages for support/provider

goodviber:
### Context
The form to create the provider from the support console currently has some code which formats the error message in the controller [providers_controller.rb](https://github.com/DFE-Digital/teacher-training-api/blob/main/app/controllers/support/providers_controller.rb#L20). It violates modifying a frozen hash in Rails 7 and also appears to be a bit of a smell. We should handle this logic into a form object similarly to how we've done it in other areas of the support console.

What needs to be done
Create a form object to wrap creating the provider which should also handle assigning the nested attributes it requires (sites etc)
Use the form object in the view and controller
Success
The error messages are shown properly and correctly link to their fields
Tech notes
We already have a pattern for doing this, have a look at the location/sites form object

As we are no longer using the "^" format, it is simpler to update the error messages, and provide a custom error display box to order the messages list to shadow the form for the nested attributes although there may be a way to do this via the formbuilder.

image

### Guidance to review
Needs another ticket to remove all the "^" and the gem "custom_error_message"
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -5,7 +5,7 @@ class Organisation < ApplicationRecord
 
   has_and_belongs_to_many :providers
 
-  validates :name, presence: { message: "Organisation name can't be blank" }
+  validates :name, presence: { message: "Enter organisation name" }
 
   has_associated_audits
   audited
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -5,7 +5,7 @@ class Organisation < ApplicationRecord
 
   has_and_belongs_to_many :providers
 
-  validates :name, presence: true
+  validates :name, presence: { message: "Organisation name can't be blank" }
 
   has_associated_audits
   audited
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -5,7 +5,7 @@ class Organisation < ApplicationRecord
 
   has_and_belongs_to_many :providers
 
-  validates :name, presence: { message: "Enter organisation name" }
+  validates :name, presence: true
 
   has_associated_audits
   audited
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -136,7 +136,7 @@ class Provider < ApplicationRecord
 
   validates :provider_type, presence: { message: "Provider type can't be blank" }
 
-  validates :telephone, phone: { message: "^Enter a valid telephone number" }, if: :telephone_changed?
+  validates :telephone, phone: { message: "Enter a valid telephone number" }, if: :telephone_changed?
 
   # TODO: Remove this validation once the 2021 recruitment cycle is over
   validates :ukprn, reference_number_format: { allow_blank: true, minimum: 8, maximum: 8, message: "^UKPRN must be 8 numbers" }
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -130,23 +130,23 @@ class Provider < ApplicationRecord
 
   validates :email, email_address: true, if: :email_changed?
 
-  validates :provider_name, length: { maximum: 100 }, on: :update
+  validates :provider_name, presence: true, length: { maximum: 100 }
 
-  validates :provider_code, uniqueness: { scope: :recruitment_cycle }
+  validates :provider_code, presence: true, uniqueness: { scope: :recruitment_cycle }
 
   validates :provider_type, presence: { message: "Provider type can't be blank" }
 
   validates :telephone, phone: { message: "Enter a valid telephone number" }, if: :telephone_changed?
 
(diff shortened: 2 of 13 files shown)


goodviber:
[Publish - Refactor provider form in support console](https://trello.com/c/7ovcLHFw/143-publish-refactor-provider-form-in-support-console)

app/models/organisation.rb
```diff
@@ -5,7 +5,7 @@ class Organisation < ApplicationRecord
 
   has_and_belongs_to_many :providers
 
-  validates :name, presence: true
+  validates :name, presence: { message: "Enter organisation name" }
```
avinhurry, line 8: Can we add this to the translations under an appropriate heading, like the others?

app/models/provider.rb
```diff
@@ -130,23 +130,23 @@ def current_accredited_courses
...
   validates :email, email_address: true, if: :email_changed?
 
-  validates :provider_name, length: { maximum: 100 }, on: :update
+  validates :provider_name, presence: true, length: { maximum: 100 }
 
-  validates :provider_code, uniqueness: { scope: :recruitment_cycle }
+  validates :provider_code, presence: true, uniqueness: { scope: :recruitment_cycle }
 
-  validates :provider_type, presence: true
+  validates :provider_type, presence: { message: "Provider type can't be blank" }
```
avinhurry, line 137: same here

app/models/provider.rb
```diff
@@ -130,23 +130,23 @@ def current_accredited_courses
 
   validates :email, email_address: true, if: :email_changed?
 
-  validates :provider_name, length: { maximum: 100 }, on: :update
+  validates :provider_name, presence: true, length: { maximum: 100 }
```
avinhurry, line 133: Wondering why we're dropping the `on: :update`

app/models/provider.rb
```diff
@@ -130,23 +130,23 @@ def current_accredited_courses
...
 
-  validates :provider_code, uniqueness: { scope: :recruitment_cycle }
+  validates :provider_code, presence: true, uniqueness: { scope: :recruitment_cycle }
 
-  validates :provider_type, presence: true
+  validates :provider_type, presence: { message: "Provider type can't be blank" }
 
-  validates :telephone, phone: { message: "^Enter a valid telephone number" }, if: :telephone_changed?
+  validates :telephone, phone: { message: "Enter a valid telephone number" }, if: :telephone_changed?
```
avinhurry, line 139: Can we add this to the translations too please?

spec/models/contact_spec.rb
```diff
@@ -60,14 +60,14 @@
         contact.email = ""
         contact.valid?
 
-        expect(contact.errors[:email]).to include("^Enter an email address in the correct format, like name@example.com")
+        expect(contact.errors[:email]).to include("Enter an email address in the correct format, like name@example.com")
```
avinhurry, line 63: I think it's good not to use the translations for the strings in the tests 👍🏼

config/locales/en.yml
```diff
@@ -211,12 +211,12 @@ en:
         site:
           attributes:
             location_name:
-              blank: "^Enter a name"
+              blank: "Enter a name"
```
avinhurry, line 214: Shall we name this more explicitly since there are three fields relating to `name`, maybe `Enter location name` or `Enter site name`?

avinhurry (commented):
The error message text is looking good! 

The following links don't link to their respective fields: 

- Enter organisation name
- Enter a name
- Enter a building and street
- Enter a postcode
- Site code must contain only A-Z, 0-9 or -

I think the error messages should also be in the same order that their respective fields are on the page, i.e if you submit the form without anything the first error should be `Enter provider name`, it's currently `Enter organisation name`.

Let me know if anything is unclear or if you want to chat about any of it.

app/models/provider.rb
goodviber: We are creating a new provider, I think?

app/models/organisation.rb
goodviber: Done

app/models/provider.rb
goodviber: Done

app/models/provider.rb
goodviber: It overrides the default phone validator message which has the caret, so it should initially have been written `phone: true` , I think we are raising another ticket to deal with all the no longer needed caret mesages and gem

config/locales/en.yml
goodviber: Done

avinhurry (approved):

PR #2756, opened 2022-06-16
Remove provider personas when going to support

avinhurry:
### Context

`/support` leads us to support login. For personas (on qa and dev environments) we have 3 personas that are not eligible to login. These personas are for publish, not support. 

- Render the button to Login as an Admin (Colin) if the user is going to `/support`
- Add support to params and dynamically render true or false, based on the path and use this as the check for which button to render.
- Keep the previous journey if the user is going to Publish
- This change will only be applied to persona login

### Guidance to review

- Login to support, and ensure that you don't see any non eligible personas to login with.
- Login to Publish and ensure that the original persona login journey remains the same.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -14,7 +14,7 @@ module Authentication
   def authenticate
     if !authenticated?
       session["post_dfe_sign_in_path"] = request.fullpath
-      redirect_to sign_in_path
+      redirect_to sign_in_path({ support: params[:controller].start_with?("support") })
     end
   end
 end
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -12,9 +12,13 @@ module Authentication
   end
 
   def authenticate
-    if !authenticated?
+    unless authenticated?
       session["post_dfe_sign_in_path"] = request.fullpath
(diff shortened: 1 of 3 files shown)


avinhurry:
[Publish - Remove provider personas when going to support](https://trello.com/c/ovgt9xml/131-publish-remove-provider-personas-when-going-to-support)

avinhurry:
[Publish - Remove provider personas when going to support](https://trello.com/c/ovgt9xml/131-publish-remove-provider-personas-when-going-to-support)

app/controllers/concerns/authentication.rb
```diff
@@ -14,7 +14,7 @@ def authenticated?
   def authenticate
     if !authenticated?
       session["post_dfe_sign_in_path"] = request.fullpath
-      redirect_to sign_in_path
+      redirect_to sign_in_path({ support: params[:controller].start_with?("support") })
```
SamJamCul, line 17: Discussed on slack, would be good to only populate this param if we're in dev or qa, as it's non-functional in prod

SamJamCul (changes requested):

SamJamCul (approved):

PR #2771, opened 2022-06-21
Fix alternative support login

avinhurry:
### Context

Fix alternative support login.

### DIFF
--- a/app/controllers/concerns/authentication.rb
+++ b/app/controllers/concerns/authentication.rb
@@ -14,7 +14,7 @@ module Authentication
   def authenticate
     unless authenticated?
       session["post_dfe_sign_in_path"] = request.fullpath
-      if AuthenticationService.mode == :persona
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/authentication.rb
```diff
@@ -14,7 +14,7 @@ def authenticated?
   def authenticate
     unless authenticated?
       session["post_dfe_sign_in_path"] = request.fullpath
-      if AuthenticationService.mode == :persona
+      if AuthenticationService.mode == "persona"
```
SamJamCul, line 17: Could we do `if AuthenticationService.persona?` instead?

SamJamCul (approved):
Looks good, just one change

app/controllers/concerns/authentication.rb
avinhurry: Done

PR #2774, opened 2022-06-22
179 publish implement cookie banner

SamJamCul:
### Context
https://trello.com/c/dDSD6t3f/179-publish-implement-cookie-banner-under-googleanalytics-flag

Added a banner for cookies - this only appears when the `google_analytics` feature flag is set to true, which it is in `development.yml` but off otherwise. 

The only remaining issue is that the tests aren't working - I've tried running `yarn test` but it doesn't seem to like the Register test, so it probably needs a bit of tweaking. 

### Guidance to review
Load up the app in a dev environment, and you should see a cookie banner. You can view the cookie that gets set once you've accepted or rejected the banner - but either way, the banner shouldn't show up on subsequent page loads.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/javascript/cookie_banner.js
```diff
@@ -0,0 +1,65 @@
...
+      this.$hideButton = this.$afterConsentBanner.querySelector('button')
+
+      this.bindEvents()
+    }
+  }
+
+  bindEvents () {
+    this.$acceptButton.addEventListener('click', () => this.accept())
+    this.$rejectButton.addEventListener('click', () => this.reject())
+    this.$hideButton.addEventListener('click', () => this.hideAfterConsentBanner())
```
defong, line 32:
```suggestion
    this.$acceptButton.addEventListener('click', this.accept)
    this.$rejectButton.addEventListener('click', this.reject)
    this.$hideButton.addEventListener('click', this.hideAfterConsentBanner)
```

Don't think you need to wrap another function

app/javascript/utils/cookie_helper.js
```diff
@@ -0,0 +1,50 @@
...
+ *  @returns {string} Cookie string in key=value format
+ *  @returns {null} if Cookie is not found
+ */
+function getCookie (name) {
+  const nameEQ = name + '='
+  const cookies = document.cookie.split(';')
+  for (let i = 0, len = cookies.length; i < len; i++) {
+    let cookie = cookies[i]
+    while (cookie.charAt(0) === ' ') {
+      cookie = cookie.substring(1, cookie.length)
```
defong, line 14:
```suggestion
      cookie = cookie.substring(1, len)
```

Otherwise it was a bit pointless at the start.

defong (commented):
Is `app-js-only` that a pointless class?

app/javascript/cookie_banner.js
SamJamCul: That change seemed to break it! Getting a `TypeError` in the js

app/javascript/utils/cookie_helper.js
SamJamCul: isn't `len = cookies.length` checking how many cookies there are, and `cookie.length` looking at the length of each separate cookie?

app/helpers/cookies_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module CookiesHelper
+  def hide_cookie_banner?
+    request.cookie_jar[Settings.cookies.consent.name] =~ /yes|no/ || !FeatureService.enabled?(:google_analytics)
+  end
+
+  def google_analytics_allowed?
+    request.cookie_jar[Settings.cookies.consent.name] == "yes"
+  end
```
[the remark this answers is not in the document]
darokel: I think that method should go when we enable this in prod

app/views/layouts/_cookie_banner.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <% end %>
+  <% end %>
+<% end %>
+<%= render GovukComponent::CookieBannerComponent.new(html_attributes: {
+  "hidden" => true,
+  "data-module" => "govuk-cookie-after-consent-banner",
+}) do |cookie_banner| %>
+  <% cookie_preferences_link = govuk_link_to(t("cookies.after_consent.page_link_text")) %>
+  <%= cookie_banner.message(text: t("cookies.after_consent.message",
+                                    cookie_preferences_link: cookie_preferences_link).html_safe) do |message| %>
```
darokel, line 27: I think we'll need to handle sending them to the right link with a bit of logic around the feature flags. While `google_analytics` is off we're still using the simple cookie page but if it's on we'd want to send them to the cookie preference page which would be enabled by the same flag

SamJamCul: This looks like it's been handled in the routes file itself - the feature flag will toggle the `cookies_path` between different routes.

darokel: of course, i was the one who added it 🤦‍♂️

Dockerfile
```diff
@@ -48,6 +48,8 @@ ADD . $APP_HOME/
 COPY --from=middleman /public/ $APP_HOME/public/docs/
 
 RUN ls /app/public/ && \
+    yarn build && \
```
goodviber, line 51: WHOOP WHOOP!

goodviber (approved):
LGTM

PR #2776, opened 2022-06-23
[0189] Refactor rollover script to handle draft and individual courses

defong:
### Context
Rollover script to handle draft and individual courses

Refactor rollover script to handle draft and individual courses

### Guidance to review
```bash

# bundle exec rake 'rollover:provider[provider_code ,  course_codes, force]'
  bundle exec rake 'rollover:provider[provider_code, course_code1 course_code2 course_code3, true]'

```

#### usage
Go and create a recruitment cycle for next recruitment cycle
Find some providers with none published courses
And take a lucky dip with the below commands

```bash
bundle exec rake 'rollover:provider[provider_code, , ]'
bundle exec rake 'rollover:provider[provider_code, ,false]'
```

To rollover a `rollover-able provider` and all it `rollover-able courses`

```bash
bundle exec rake 'rollover:provider[provider_code, course_codes , ]'
bundle exec rake 'rollover:provider[provider_code, course_codes, false]'
```

To rollover a `rollover-able provider` and only the specified `rollover-able courses`

```bash
bundle exec rake 'rollover:provider[provider_code, ,true]'
```

To rollover a `rollover-able or non-rollover-able provider`

```bash
bundle exec rake 'rollover:provider[provider_code, course_codes1 course_codes2, true]'
```

To rollover a `rollover-able or non-rollover-able provider` and only the specified `rollover-able or non-rollover-able courses`

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/providers/copy_to_recruitment_cycle_service.rb
+++ b/app/services/providers/copy_to_recruitment_cycle_service.rb
@@ -40,12 +40,13 @@ module Providers
     end
 
   private
+    attr_reader :copy_course_to_provider_service, :copy_site_to_provider_service
 
     def copy_courses_to_new_provider(provider, new_provider)
       courses_count = 0
 
       provider.courses.each do |course|
-        new_course = @copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
+        new_course = copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
         courses_count += 1 if new_course.present?
       rescue Exception # rubocop: disable Lint/RescueException
         Rails.logger.fatal "error trying to copy course #{course.course_code}"
@@ -59,7 +60,7 @@ module Providers
       sites_count = 0
 
       provider.sites.each do |site|
-        new_site = @copy_site_to_provider_service.execute(site: site, new_provider: new_provider)
+        new_site = copy_site_to_provider_service.execute(site: site, new_provider: new_provider)
         sites_count += 1 if new_site.present?
       end
 
--- a/app/services/providers/copy_to_recruitment_cycle_service.rb
+++ b/app/services/providers/copy_to_recruitment_cycle_service.rb
@@ -49,7 +49,7 @@ module Providers
       provider.courses.each do |course|
         new_course = copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
         courses_count += 1 if new_course.present?
-      rescue Exception # rubocop: disable Lint/RescueException
+      rescue StandardError
         Rails.logger.fatal "error trying to copy course #{course.course_code}"
         raise
       end
--- a/app/services/providers/copy_to_recruitment_cycle_service.rb
+++ b/app/services/providers/copy_to_recruitment_cycle_service.rb
@@ -1,11 +1,8 @@
 module Providers
   class CopyToRecruitmentCycleService
-    attr :logger
-
-    def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:, logger: nil)
+    def initialize(copy_course_to_provider_service:, copy_site_to_provider_service:)
       @copy_course_to_provider_service = copy_course_to_provider_service
       @copy_site_to_provider_service = copy_site_to_provider_service
-      @logger = logger || Logger.new("/dev/null")
     end
 
     def execute(provider:, new_recruitment_cycle:, force: false)
@@ -51,7 +48,7 @@ module Providers
         new_course = @copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
         courses_count += 1 if new_course.present?
       rescue Exception # rubocop: disable Lint/RescueException
-        logger&.fatal "error trying to copy course #{course.course_code}"
+        Rails.logger.fatal "error trying to copy course #{course.course_code}"
         raise
       end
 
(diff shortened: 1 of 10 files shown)


defong:
[Refactor rollover script to handle draft and individual courses](https://trello.com/c/eFhDX9F9/189-refactor-rollover-script-to-handle-draft-and-individual-courses)

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -44,14 +42,45 @@ def execute(provider:, new_recruitment_cycle:, force: false)
...
+      end
+
+      courses
+    end
+
+    def courses_from_course_codes(provider, course_codes)
+      courses = provider.courses.where(course_code: course_codes.to_a.map(&:upcase))
+
+      if courses.count != course_codes.count
+        error_message = "error courses found has discrepancies (#{courses.count} vs #{course_codes.count})"
```
darokel, line 68:
```suggestion
error_message = "Error: discrepancy between courses found and provided course codes (#{courses.count} vs #{course_codes.count})"
```

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -44,14 +42,45 @@ def execute(provider:, new_recruitment_cycle:, force: false)
...
       courses_count = 0
 
-      provider.courses.each do |course|
-        new_course = @copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
+      courses.each do |course|
+        new_course = copy_course_to_provider_service.execute(course: course, new_provider: new_provider)
         courses_count += 1 if new_course.present?
-      rescue Exception # rubocop: disable Lint/RescueException
-        logger&.fatal "error trying to copy course #{course.course_code}"
+      rescue StandardError
```
darokel, line 82: I wonder if we should raise the actual error instead of masking it as just something went wrong?

darokel:
Nice work on breaking this out, left a couple of comments. Just a sanity check on my end, this should allow us to:

1) Maintain the existing behaviour eg running the script pre-rollover in all envs will copy over published courses as normal
2) When driven by the provider, we'd pass in the provider code and that individual course code. The provider would be skipped since they've already been rolled over but the course would be irrespective of state because of the `force` params right?

app/services/providers/copy_to_recruitment_cycle_service.rb
defong: I have no idea what sort of exceptions this pathway may encounter

darokel: It'll likely be an `ActiveRecord::RecordNotSaved` exception since that's what the service is calling `save!` after building the objects. We could just bubble that up? We could leave it for another card but it'll just make trying to reproduce why a particular course wasn't rolledover a bit of a faff.

app/services/providers/copy_to_recruitment_cycle_service.rb
defong: done

defong:
>     1. Maintain the existing behaviour eg running the script pre-rollover in all envs will copy over published courses as normal
Yes it still function as before

 
>     2. When driven by the provider, we'd pass in the provider code and that individual course code. The provider would be skipped since they've already been rolled over but the course would be irrespective of state because of the `force` params right?

With force as `true` and only `provider_code` provided the provider is rollover
with force as `true` and and `provider_code` & `course_code`s the provider & its courses is rollover

If the provider already exists it will be skipped or if the courses already exists it will be skipped.

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -44,14 +42,45 @@ def execute(provider:, new_recruitment_cycle:, force: false)
...
   private
 
-    def copy_courses_to_new_provider(provider, new_provider)
+    attr_reader :copy_course_to_provider_service, :copy_site_to_provider_service, :force
+
+    def courses_to_copy(provider, course_codes)
+      courses = []
+      if force
+        if course_codes.nil?
+          Rails.logger.info "no courses will be rollover"
```
goodviber, line 51: Should this be 'rolled over' ?

app/services/providers/copy_to_recruitment_cycle_service.rb
```diff
@@ -44,14 +42,45 @@ def execute(provider:, new_recruitment_cycle:, force: false)
...
+        courses = courses_from_course_codes(provider, course_codes)
+      end
+
+      courses
+    end
+
+    def courses_from_course_codes(provider, course_codes)
+      courses = provider.courses.where(course_code: course_codes.to_a.map(&:upcase))
+
+      if courses.count != course_codes.count
```
goodviber, line 67: unless?

goodviber (approved):
lgtm, just a couple of comments

app/services/providers/copy_to_recruitment_cycle_service.rb
defong: Done

app/services/providers/copy_to_recruitment_cycle_service.rb
defong: Done

PR #2810, opened 2022-07-05
Update rollover doc

avinhurry:
### Context

A small update to the rollover docs.

### Changes proposed

- Update the dates that we will be using
- Remove steps refering to google docs as we don't use these anymore


avinhurry:
[Rollover QA & Staging](https://trello.com/c/qYCkb4Xb/231-rollover-qa-staging)

PR #2818, opened 2022-07-06
Refactor and add separator to title bar items

avinhurry:
### Context

Add separator in between the title bar items, when there are multiple items and a small refactor.

#### Old

image

#### New
image


avinhurry:
[Add divider between change org and change recruitment cycle](https://trello.com/c/w8LTV0Dl/277-add-divider-between-change-org-and-change-recruitment-cycle)

PR #2824, opened 2022-07-07
[0243] Users should not see admin error

defong:
### Context
Users should not see admin error

Present a forbidden page if user is not an admin instead of redirecting to sign in when they are signed in
Removed the flash, as it presents information without boundary
Redirect to the publish root path as that is really what the user wants to go to

### Guidance to review
You need DfE Sign in to reproduce the original isses

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


defong:
[Users should not see admin error](https://trello.com/c/JwUImNIu/243-users-should-not-see-admin-error)

PR #2843, opened 2022-07-19
Add recruitment cycle scope to course fetch

SamJamCul:
### Context
https://trello.com/c/JKvoMiv7/316-bug-copying-course-content-only-copies-from-current-cycle

When copying content for a course, it is now scoped to a course that runs in the same recruitment cycle as the course being copied to. 

Trying copying course content across from courses from the same provider, but changing the recruitment cycle - changes to the current recruitment cycle's course content will not be copyable to the next recruitment cycle's courses. 

We don't really seem to have any tests for this, I'm not entirely sure where to start introducing them either!

### DIFF
--- a/app/services/courses/fetch.rb
+++ b/app/services/courses/fetch.rb
@@ -1,8 +1,8 @@
 module Courses
   class Fetch
     class << self
-      def by_code(provider_code:, course_code:)
-        RecruitmentCycle.current.providers
+      def by_code(provider_code:, course_code:, recruitment_cycle_year:)
+        RecruitmentCycle.find_by(year: recruitment_cycle_year).providers
           .find_by(provider_code:)
           .courses
           .find_by(course_code:)
--- a/spec/services/courses/fetch_spec.rb
+++ b/spec/services/courses/fetch_spec.rb
@@ -1,0 +1,18 @@
+require "rails_helper"
+
+describe Courses::Fetch do
+  describe ".by_code" do
+    let(:provider_code) { course.provider.provider_code }
+    let(:course_code) { course.course_code }
+    let(:recruitment_cycle_year) { course.recruitment_cycle.year }
+    let(:course) { create(:course) }
+
+    it "fetches a course by course_code" do
+      expect(described_class.by_code(
+        provider_code:,
+        course_code:,
+        recruitment_cycle_year:,
+      )).to eq(course)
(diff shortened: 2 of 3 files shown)


app/services/courses/fetch.rb
```diff
@@ -1,8 +1,8 @@
 module Courses
   class Fetch
     class << self
```
defong, line 3: I'm not too sure of the usefulness of this `service` maybe we should look at removing it altogether @darokel

darokel: Yup, we should get rid of it if we can. I think it was used as a way to deal with the lookup soup in old publish

spec/services/courses/fetch_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+describe Courses::Fetch do
+  describe "#by_code" do
+    let(:provider_code) { course.provider.provider_code }
+    let(:course_code) { course.course_code }
+    let(:recruitment_cycle_year) { course.recruitment_cycle.year }
+    let(:course) { create(:course) }
+
+    it "fetches a course by course_code" do
+      expect(::Courses::Fetch.by_code(
```
defong, line 11:
```suggestion
      expect(described_class.by_code(
```

spec/services/courses/fetch_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "rails_helper"
+
+describe Courses::Fetch do
+  describe "#by_code" do
```
defong, line 4:
```suggestion
  describe ".by_code" do
```

Class method?

app/services/courses/fetch.rb
defong: Yeah lets card that up

defong (approved):

spec/services/courses/fetch_spec.rb
SamJamCul: Sorted

spec/services/courses/fetch_spec.rb
SamJamCul: Done

app/services/courses/fetch.rb
SamJamCul: It's now in the dev backlog

PR #2857, opened 2022-07-25
Eager load app in CI

darokel:
### Context

Tries to fix flakey test https://github.com/DFE-Digital/publish-teacher-training/runs/7452055270?check_suite_focus=true


goodviber (approved):
Nice

defong (approved):

PR #2863, opened 2022-07-27
[xxx] Update support playbook

darokel:
### Context

Expands example for updating `program_type` on a course.

### DIFF
--- a/docs/support_playbook.md
+++ b/docs/support_playbook.md
@@ -43,6 +43,13 @@ end
 
 Change `program_type` to the preferred value on the course record.
 
+```ruby
+course = RecruitmentCycle.current.providers.find_by(provider_code: "2A5").courses.find_by(course_code: "P843")
+course.update(program_type: :school_direct_salaried_training_programme)
+# Don't forget to clear out values for other fields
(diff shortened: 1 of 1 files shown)


docs/support_playbook.md
```diff
@@ -43,6 +43,13 @@ end
 
 Change `program_type` to the preferred value on the course record.
 
+```ruby
+course = RecruitmentCycle.current.providers.find_by(provider_code: "2A5").courses.find_by(course_code: "P843")
+course.update(program_type: :school_direct_salaried_training_programme)
```
defong, line 48: Maybe we should use this 
https://github.com/DFE-Digital/publish-teacher-training/blob/6e27e07988e7425252198466500c4eab878437f0/app/services/courses/assign_program_type_service.rb

:thinking:

defong (approved):

docs/support_playbook.md
darokel: That seems to be used by another method https://github.com/DFE-Digital/publish-teacher-training/blob/49bb67828e09a088288b55c8d4facf745a0a4543/app/models/course.rb#L635. Not sure if we should add to this service as i'm unsure of the side effects at the moment. We could probably do it via another service but this seems quite specific to adhoc support requests

PR #2865, opened 2022-07-27
load cloudfront-rails in envs that use CloudFront

misaka:
This allows us to send the correct IPs to the analytics platform.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -133,7 +133,7 @@ gem "sprockets-rails", require: "sprockets/railtie"
 # for sending analytics data to the analytics platform
 gem "dfe-analytics", github: "DFE-Digital/dfe-analytics", tag: "v1.2.1"
 
-group :production do
+group :production, :qa, :sandbox, :staging do
   gem "cloudfront-rails"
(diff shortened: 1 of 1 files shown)


defong (commented):
> This gem is relevant in any deployed env in GPaaS.

What about sandboc and review apps?

misaka:
> What about sandboc and review apps?

This is enabled for the `review`, but good point about `sandbox`. Also maybe `rollover`? And is `research` used?

defong:
> > What about sandboc and review apps?
> 
> This is enabled for the `review`, but good point about `sandbox`. Also maybe `rollover`? And is `research` used?

Are you going to need it? Your PR descriptions imply one way and your PR changes inversely imply you don't.

defong (changes requested):
I suppose it should be set for the other deployed env.

As it stands it inconclusive.

defong (approved):

PR #2885, opened 2022-07-29
342 add cycle switcher to support console

avinhurry:
### Context

As we have now finished scoping the recruitment cycle year to the params in support, we need a way for support users to easily switch between cycles. 

We have a switcher in the app, we are building something similar for support users.

- Implement the cycle switcher html page for support
- Show the html page when switching between cycles
- Add the cycle switcher in the title bar in line with the prototype

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/support_title_bar/view.html.erb
+++ b/app/components/support_title_bar/view.html.erb
@@ -1,0 +1,22 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+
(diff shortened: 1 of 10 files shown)


avinhurry:
[Add cycle switcher to support console](https://trello.com/c/Y71WwNH4/342-add-cycle-switcher-to-support-console)

defong (approved):

PR #2891, opened 2022-08-01
Fix support updates in next recruitment cycle

avinhurry:
### Context

Fix support updates in next recruitment cycle for courses and locations

### DIFF
--- a/app/controllers/support/courses_controller.rb
+++ b/app/controllers/support/courses_controller.rb
@@ -25,10 +25,6 @@ module Support
 
   private
 
-    def recruitment_cycle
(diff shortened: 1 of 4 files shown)


app/controllers/support/courses_controller.rb
```diff
@@ -25,8 +25,12 @@ def update
 
   private
 
+    def recruitment_cycle
```
darokel, line 28: Could we move this and the one below into `SupportController` as they both inherit from it

avinhurry: Good shout. Done [here](https://github.com/DFE-Digital/publish-teacher-training/pull/2891/commits/2fbc00ca5d30ea288ce06c637bdfbea85cea0573)

darokel (approved):

PR #2902, opened 2022-08-03
Remove PE Allocations tab from support in the next recruitment cycle

avinhurry:
### Context

Removing the PE Allocations tab from support in the next recruitment cycle as this is no longer required.

### Guidance to review

- Login to support in the current cycle, check that the PE allocations tab is present
- Repeat for the next cycle and ensure it is not present 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/support/_menu.html.erb
+++ b/app/views/support/_menu.html.erb
@@ -1,5 +1,7 @@
+#The PE allocations tab can be deleted after rollover 2022
+
 <%= render TabNavigation::View.new(items: [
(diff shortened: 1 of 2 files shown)


avinhurry:
[Remove PE allocations on support](https://trello.com/c/0EKCwBBF/367-remove-pe-allocations-on-support)

defong (approved):

PR #2911, opened 2022-08-09
Change heading and column content for will it be on find next recruitment cycle

avinhurry:
### Context

We have decided to change the `Will it be on Find?` column on the courses table in the next cycle to match the one in the current cycle, which is `Is it on Find?`. 

This change was made so that users can switch between cycles and be able to see at a glance what is on Find as that is the header they would be used to seeing all year. 

- Change the `next_cycle_open_date` in `Settings.yml` to `2022-10-4`
- Change the `Will it be on Find?` header in the courses cycle to match what is displayed in the current cycle.
- Change the value in that column so that it makes sense in the new context
- Our recruitment cycle year is displaying incorrectly in the course creation flow. I've updated that too.  

### Guidance to review

The main thing that concerns me here is changing the date in `Settings.yml`. Is this change OK to make and has it been changed to the correct date?

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -22,7 +22,7 @@ class CourseDecorator < ApplicationDecorator
       if current_cycle_and_open?
         h.govuk_link_to("Yes - view online", h.search_ui_course_page_url(provider_code: provider.provider_code, course_code: object.course_code))
       else
-        "Yes - from #{Date::MONTHNAMES[Settings.next_cycle_open_date.to_date.month]}"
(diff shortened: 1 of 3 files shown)


avinhurry:
[Change heading and column content for 'Will it be on Find?' next recruitment cycle](https://trello.com/c/RD5dY8mQ/376-change-heading-and-column-content-for-will-it-be-on-find-next-recruitment-cycle)

defong (approved):

PR #2940, opened 2022-08-16
[407] Streamline course state values in API

darokel:
### Context

https://trello.com/c/HzBaBvhu/407-update-course-status-send-via-the-api

- Streamlines course state values sent via the API. Keeps the logic in the service the same as some of the values are still relevant in our system
- Update api docs

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/content_status_service.rb
+++ b/app/services/courses/content_status_service.rb
@@ -1,23 +1,12 @@
 module Courses
   class ContentStatusService
     def execute(enrichment:, recruitment_cycle:)
-      if enrichment.blank?
-        if recruitment_cycle.next?
-          :rolled_over
-        else
-          :empty
-        end
-      elsif enrichment.published?
-        :published
-      elsif enrichment.withdrawn?
-        :withdrawn
-      elsif enrichment.has_been_published_before?
-        :published_with_unpublished_changes
-      elsif enrichment.rolled_over?
-        :rolled_over
-      else
-        :draft
-      end
+      return :rolled_over if (recruitment_cycle.next? && enrichment.blank?) || enrichment&.rolled_over?
+      return :published if enrichment&.published?
+      return :withdrawn if enrichment&.withdrawn?
+      return :published_with_unpublished_changes if enrichment&.has_been_published_before?
+
+      :draft
     end
   end
 end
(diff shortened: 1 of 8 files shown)


darokel:
[Update course status send via the API](https://trello.com/c/HzBaBvhu/407-update-course-status-send-via-the-api)

swagger/public_v1/api_spec.json
```diff
@@ -2524,4 +2528,4 @@
       }
     }
   }
-}
+}
```
defong, line 2531: line ending

swagger/public_v1/api_spec.json
```diff
@@ -1154,7 +1152,9 @@
             "enum": [
               "scitt",
               "lead_school",
-              "university"
+              "university",
+              "unknown",
+              "invalid_value"
```
defong, line 1157: I don't think these are actually there any more.

app/services/courses/content_status_service.rb
```diff
@@ -1,23 +1,12 @@
...
-        :published_with_unpublished_changes
-      elsif enrichment.rolled_over?
-        :rolled_over
-      else
-        :draft
-      end
+      return :rolled_over if (recruitment_cycle.next? && enrichment.blank?) || enrichment&.rolled_over?
+      return :published if enrichment&.published?
+      return :withdrawn if enrichment&.withdrawn?
+      return :published_with_unpublished_changes if enrichment&.has_been_published_before?
```
defong, line 7: Is there an actually need for this any more?

It looks more complexed then it should be, since it gets remapped later.

darokel: Yeah it's still used by the enrichment stuff so i think we still need it. Basically the real change is just avoiding sending `empty` and `published_with_unpublished_changes` down the api because the latter is still used internally (which is why it's the only one remapped in the api serializer otherwise whatever value was determined by this service gets returned)

defong (approved):
Just a few minors bits to tend to

swagger/public_v1/api_spec.json
darokel: done

swagger/public_v1/api_spec.json
darokel: done

PR #2951, opened 2022-08-19
[403] Put find routes behind constraint

darokel:
### Context

We need to work on new Find behind basic auth to prevent the pages being indexed/crawled (since it's a public facing service). This PR adds a guard to only enable the Find routes in QA for now until we've at least migrated most of the public facing pages along with the sitemap stuff. 

https://trello.com/c/8LjaEwQe/403-applying-the-domain-constraint-for-the-new-find-url

- Split out routes into separate files
- Only enable Find routes for dev, review and qa
- Have also removed the find temp urls in all envs other than QA
- Set root path to find namespace if arriving from temp url

### Guidance to review

Check find routes still load in review
Will need to do a manual deploy to qa

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


darokel:
[Applying the domain constraint for the new find url](https://trello.com/c/8LjaEwQe/403-applying-the-domain-constraint-for-the-new-find-url)

config/routes/find.rb
```diff
@@ -1,3 +1,7 @@
+constraints(FindConstraint.new) do
```
defong, line 1: Do we still need this then?

:thinking:

darokel: Yup - this is what actually sets the root path to go to the find kickoff page (if coming from the temp url)

defong (approved):

PR #2979, opened 2022-09-05
#440 Update course copy service in Publish to handle visa fields

goodviber:
### Context
To simplify collecting visa information from providers, we're updating how and which providers provide their visa sponsorship information via their provider details page and when they create courses

Add visa fields to CopyToProviderService

### Guidance to review
Not sure if this is exactly correct, we have visa fields on both the course and provider, are we transitioning from one to the other? 
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/copy.rb
+++ b/app/services/courses/copy.rb
@@ -43,6 +43,12 @@ module Courses
       ["Other requirements", "other_requirements"],
     ].freeze
 
+    # TODO This is to be utilised when we add these to a course form
+    VISA_FIELDS = [
+      ["Can sponsor skilled worker visa", "can_sponsor_skilled_worker_visa"],
+      ["Can sponsor student visa", "can_sponsor_student_visa"],
+    ].freeze
+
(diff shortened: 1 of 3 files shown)


goodviber:
[Update course copy service in Publish to handle visa fields](https://trello.com/c/rnmfwsZF/440-update-course-copy-service-in-publish-to-handle-visa-fields)

SamJamCul (approved):

defong:
> Not sure if this is exactly correct, we have visa fields on both the course and provider, are we transitioning from one to the other?

I think this is only a small part of it, as this is for rollover.

@darokel can you provide a steer for this please.

darokel:
@goodviber We're introducing them on the course level as courses can have their own visa settings. Going forward, we're transitioning to using the settings exclusively from the courses. I think we'll also need to cover the Courses::Copy service too?

PR #2998, opened 2022-09-14
Introduce "Add course" label to the new course flow and add course name and code to edit course views

avinhurry:
### Context

We are introducing a "Add course" hint label at the top of all pages in the Add new course flow. 

- Create a reusable view component with the correct css.
- Utilise the view component for all the pages in the new course flow.
- Add view preview and specs.

### Guidance to review

- Have I missed any of the pages?

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/publish/courses/accredited_body/new.html.erb
+++ b/app/views/publish/courses/accredited_body/new.html.erb
@@ -4,6 +4,8 @@
   <%= govuk_back_link_to(@back_link_path) %>
 <% end %>
 
+<%= render CaptionText::View.new(text: "Add course") %>
+
 <%= render "publish/shared/errors" %>
 
 <fieldset class="govuk-fieldset">
--- a/app/views/publish/courses/accredited_body/new.html.erb
+++ b/app/views/publish/courses/accredited_body/new.html.erb
@@ -4,13 +4,12 @@
   <%= govuk_back_link_to(@back_link_path) %>
 <% end %>
 
-<%= render CaptionText::View.new(text: "Add course") %>
-
 <%= render "publish/shared/errors" %>
 
 <fieldset class="govuk-fieldset">
   <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
     <h1 class="govuk-fieldset__heading">
+      <%= render CaptionText::View.new(text: t("course.add_course")) %>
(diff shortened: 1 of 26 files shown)


avinhurry:
[[SNAG] - Add the "Add course" label in the new course flow](https://trello.com/c/wDlESLCl/497-snag-add-the-add-course-label-in-the-new-course-flow)

avinhurry:
[[SNAG] - Add the course name label in when editing sections](https://trello.com/c/ozCUVNKK/498-snag-add-the-course-name-label-in-when-editing-sections)

darokel:
I think the label is supposed to sit within the h1 based on the prototype:

Screenshot 2022-09-15 at 09 37 31

so this doesn't happen:

Screenshot 2022-09-15 at 09 36 15

If we could sort that out and the other comment this should be good to go

app/views/publish/courses/accredited_body/new.html.erb
```diff
@@ -4,6 +4,8 @@
   <%= govuk_back_link_to(@back_link_path) %>
 <% end %>
 
+<%= render CaptionText::View.new(text: "Add course") %>
```
darokel, line 7: Could we put the "Add course" text into `I18n` so we can update it once for different areas

avinhurry:
> I think the label is supposed to sit within the h1 based on the prototype:
> 
> Screenshot 2022-09-15 at 09 37 31
> 
> so this doesn't happen:
> 
> Screenshot 2022-09-15 at 09 36 15
> 
> If we could sort that out and the other comment this should be good to go

Ahhh 😱. Good spot, thanks. Done [here](https://github.com/DFE-Digital/publish-teacher-training/pull/2998/commits/c18acde1c223e5d4eeaae0ea0218ece5e45eb0a8)

app/views/publish/courses/accredited_body/new.html.erb
avinhurry: This is also done [here](https://github.com/DFE-Digital/publish-teacher-training/pull/2998/commits/c18acde1c223e5d4eeaae0ea0218ece5e45eb0a8)

darokel (approved):

PR #3024, opened 2022-09-28
[535] Return user providers based on cycle year

darokel:
### Context

https://trello.com/c/VtEmHLuc/534-associate-users-with-organisation-1tf

This PR fixes an issue where users are given access to a provider in the next cycle but the UI will still show the user has having 0 providers since it's looking in the current cycle.

- Change `user.providers.in_current_cycle` to `user.providers.for_cycle(cycle)` so it returns the user's providers based on the cycle provided

### Guidance to review

Add a user to a provider in the next cycle and assert they only have access to that provider in that cycle. Check the UI correctly returns their provider when viewing in the next cycle

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support/providers/users_controller.rb
+++ b/app/controllers/support/providers/users_controller.rb
@@ -7,6 +7,7 @@ module Support
       end
 
       def show
+        recruitment_cycle
         provider_user
       end
 
--- a/app/controllers/support/providers_controller.rb
+++ b/app/controllers/support/providers_controller.rb
@@ -48,7 +48,7 @@ module Support
     end
 
     def find_providers
-      recruitment_cycle.providers.order(:provider_name).includes(:courses, :users)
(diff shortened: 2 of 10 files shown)


darokel:
[Associate users with organisation (1TF)](https://trello.com/c/VtEmHLuc/534-associate-users-with-organisation-1tf)

defong (changes requested):
Does this needs updating as well?

https://github.com/DFE-Digital/publish-teacher-training/blob/ab70961cbcf2e0f7299a9c5255ca01c242c08c03/app/models/user.rb#L73-L79

darokel:
> Does this needs updating as well?
> 
> https://github.com/DFE-Digital/publish-teacher-training/blob/ab70961cbcf2e0f7299a9c5255ca01c242c08c03/app/models/user.rb#L73-L79

Actually no i don't think so

defong (approved):

PR #3028, opened 2022-09-28
#536 add campaign_name enum

goodviber:
### Context
Some courses are created as part of a campaign to improve recruitment. The Engineers Teach Physics course is part of such a campaign. It is ultimately just a physics course but it will need to have a different title as part of the campaign.
We'll need to build a new step in the course flow to allow providers to create these types of courses.
What needs to be done

Add a new field on the course table campaign_name? this would be an enum field which could contain other campaigns in the future eg engineers_teach_physics
Write database migrations
Success

We should have a way of tracking these types of courses either by the field above or some other way
we should be able to do something like course.engineers_teach_physics? or course.has_campaign? etc

Db migration for enum campaign_name
### Guidance to review
Should the enum be integer or string? 
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -48,6 +48,11 @@ class Course < ApplicationRecord
     not_required: 9,
   }
 
+  enum :campaign_name,
+    {
+      engineers_teach_physics: 0, # should this be a string?
+    }
+
   ENTRY_REQUIREMENT_OPTIONS = {
     must_have_qualification_at_application_time: 1,
     expect_to_achieve_before_training_begins: 2,
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -48,10 +48,7 @@ class Course < ApplicationRecord
     not_required: 9,
   }
 
-  enum :campaign_name,
-    {
-      engineers_teach_physics: 0, # should this be a string?
-    }
+  enum :campaign_name, COURSE_CAMPAIGNS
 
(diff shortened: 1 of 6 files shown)


goodviber:
[Add campaign enum field in course table](https://trello.com/c/EqYMAK6K/536-add-campaign-enum-field-in-course-table)

app/models/course.rb
```diff
@@ -48,6 +48,11 @@ class Course < ApplicationRecord
     not_required: 9,
   }
 
+  enum :campaign_name,
```
darokel, line 51: Yup i think we should start standardising enums in one place, eg define an initializer in `config/initializers/course_enums.rb` then something like:

```ruby
  COURSE_CAMPAIGN_ENUMS = {
    engineers_teach_physics: "engineers_teach_physics",
  }.freeze

COURSE_CAMPAIGNS = {
  COURSE_CAMPAIGN_ENUMS[:engineers_teach_physics] => 0,
}
```

then in `course.rb`:

```ruby
  enum :campaign_name, COURSE_CAMPAIGNS
```

it'll give us flexibility to easily reference and update the values (if we ever need to), eg `COURSE_CAMPAIGN_ENUMS[:engineers_teach_physics]`

darokel (approved):

PR #3044, opened 2022-10-05
[598] Update CopyToProviderService

darokel:
### Context

https://trello.com/c/xMAG1w3V/598-copy-over-courses-to-organisation-1yf

- Tweak the CopyToProviderService so that it can run in the current cycle to copy courses from one provider to another

### Guidance to review

We have already documented how to use this in https://github.com/DFE-Digital/publish-teacher-training/blob/main/docs/support_playbook.md (and have used it for a previous support ticket). 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/copy_to_provider_service.rb
+++ b/app/services/courses/copy_to_provider_service.rb
@@ -51,13 +51,17 @@ module Courses
     end
 
     def adjusted_applications_open_from_date(course, year_differential)
-      next_cycle = RecruitmentCycle.next_recruitment_cycle
+      return course.applications_open_from if next_cycle.blank?
 
       if course.applications_open_from + year_differential.year >= next_cycle.application_start_date
         course.applications_open_from + year_differential.year
       else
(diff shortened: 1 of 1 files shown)


app/services/courses/copy_to_provider_service.rb
```diff
@@ -51,13 +51,17 @@ def copy_latest_enrichment_to_course(course, new_course)
     end
 
     def adjusted_applications_open_from_date(course, year_differential)
-      next_cycle = RecruitmentCycle.next_recruitment_cycle
+      return course.applications_open_from if next_cycle.blank?
```
defong, line 54:
```suggestion
      return current_cycle.application_start_date if course.applications_open_from.blank? && next_cycle.blank?
      
      return course.applications_open_from if next_cycle.blank?
```

Maybe something like this?

defong (approved):
some suggestions

PR #3066, opened 2022-10-18
[xxx] Fix ETP issues

darokel:
### Context

Fixes https://sentry.io/organizations/dfe-teacher-services/issues/3678466933/?referrer=slack and a few issues with ETP


db/schema.rb
```diff
@@ -134,7 +134,7 @@
     t.string "additional_gcse_equivalencies"
     t.boolean "can_sponsor_skilled_worker_visa", default: false
     t.boolean "can_sponsor_student_visa", default: false
-    t.string "campaign_name", default: "no_campaign"
+    t.integer "campaign_name"
```
defong, line 137: Don't we need another migration to drop column and create column as the data type has changed?

goodviber (commented):
The deploy app is still not working correctly

db/schema.rb
darokel: done

darokel:
@goodviber should be fixed now, have a look

goodviber (approved):

PR #3073, opened 2022-10-21
[735]reset campaign name when changing from ETP

goodviber:
### Context
We need to make sure we're resetting the campaign_name back to nil when a user changes from ETP to a different subject.
What needs to be done

Handle this edge case where ever the subject is changed so that the field is set back to nil on the course
clear campaign_name params for new action and update course model in update action in subjects controller
### Guidance to review
Create new Physics course then change course from confirmation page, then change back to Physics course, ETP radio buttons should be blank. Do same with existing course from basics details page.
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/features/publish/courses/new_engineers_teach_physics_spec.rb
+++ b/spec/features/publish/courses/new_engineers_teach_physics_spec.rb
@@ -99,13 +99,13 @@ private
 
   def params_with_subject(level, subject_type)
     course_subject = course_subject(subject_type)
-    "course%5Bis_send%5D=%5B%220%22%5D&course%5Blevel%5D=#{level}&course%5Bmaster_subject_id%5D=#{course_subject.id}&course%5Bsubjects_ids%5D%5B%5D=#{course_subject.id}"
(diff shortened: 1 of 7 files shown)


goodviber:
[Reset campaign name field when changing from ETP](https://trello.com/c/7UCW31IK/735-reset-campaign-name-field-when-changing-from-etp)

spec/features/publish/courses/new_engineers_teach_physics_spec.rb
```diff
@@ -29,6 +29,39 @@
...
+    and_i_click_continue
+    then_i_am_met_with_the_age_range_page_with_latin(:secondary, :latin)
+
+    when_i_go_back
+    and_i_select_a_subject(:physics)
+    and_i_click_continue
+    then_i_am_met_with_the_engineers_teach_physics_page_with_no_etp(:secondary, :physics)
+    and_i_click_continue
+    then_i_see_an_error_message
+  end
```
avinhurry, line 63: Do we need this test if we have the other one? Is it testing anything that the other one isn't? It's a big feature test and our test suite is already quite slow!

avinhurry (approved):
Looks good to me, left a couple of comments at your discretion 👍🏼

spec/features/publish/courses/new_engineers_teach_physics_spec.rb
goodviber: The new path validates in the controller, the edit path validates in the form_object, so I think we should cover both

PR #3086, opened 2022-10-27
Implement engineers teach physics filter on Find

avinhurry:
### Context

In line with the "Engineers teach physics" work, we are adding a filter on Find to search for only these courses. This filter will appear when searching for the "Physics" subject.

These are the backend changes, see https://github.com/DFE-Digital/find-teacher-training/pull/1567 for frontend. 

- Add scope which returns courses with the campaign_name "engineers_teach_physics"

- Add the scope to the course search service

### Guidance to review

- Checkout this branch along with the respective branch on Find, [here](https://github.com/DFE-Digital/find-teacher-training/pull/1567)
- Perform a search for a Physics course on Find
- Try out the new filter

### Screenshot

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -30,6 +30,7 @@ class CourseSearchService
     scope = scope.with_degree_grades(degree_grades) if degree_grades.any?
     scope = scope.changed_since(filter[:updated_since]) if updated_since_filter?
     scope = scope.can_sponsor_visa if can_sponsor_visa_filter?
+    scope = scope.engineers_teach_physics if engineers_teach_physics_filter?
 
     # The 'where' scope will remove duplicates
     # An outer query is required in the event the provider name is present.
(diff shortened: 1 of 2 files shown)


app/models/course.rb
```diff
@@ -285,6 +285,10 @@ def new_draft_attributes
     )
   }
 
+  scope :engineers_teach_physics, lambda {
+    where(campaign_name: "engineers_teach_physics")
+  }
+
```
defong, line 291:
```suggestion
```

^^ That looks way better

spec/factories/courses.rb
```diff
@@ -50,6 +50,10 @@
       level { :secondary }
     end
 
+    trait :engineers_teach_physics do
+      campaign_name { "engineers_teach_physics" }
+    end
+
```
defong, line 56:
```suggestion
```

^^ That looks way better

app/models/course.rb
avinhurry: https://media.giphy.com/media/SwyTq2jJxc9im6BYnN/giphy.gif

spec/factories/courses.rb
avinhurry: https://media.giphy.com/media/SwyTq2jJxc9im6BYnN/giphy.gif

defong (approved):

PR #3113, opened 2022-11-08
781 fix subject order in check your answers and course details page

avinhurry:
### Context

The order of the subjects on the check your answers page and course details page should match the order they were entered.

```For example:

If the user chose Physics and Chemistry, the list should display:

- Physics
- Chemistry

If the user chose Modern Languages, the list should display:

- Modern Languages
- Language choice 1
- Language choice ...n

Where language choice is in alphabetical order as per the selection page.

If the user chose Physics and Modern Languages, the list should display:

- Physics
- Modern Languages
- Language choice 1
- Language choice ...n

If the user chose Modern Languages and Latin, the list should display:

- Modern Languages
- Language choice 1
- Language choice ...n
- Latin
```

Display subjects in the order they were chosen, as per the requirements above.

### Guidance to review

- Choose the subjects in the various scenarios above when adding a new course and ensure they are displayed in the same order as above.

- Choose the subjects in the various scenarios above when editing basic details on an existing course and ensure they are displayed in the same order as above.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -143,7 +143,7 @@ class CourseDecorator < ApplicationDecorator
 
     if main_subject_is_modern_languages?
       format_name(modern_language_subjects.push(additional_subjects).flatten.uniq.unshift(main_subject))
-    elsif !main_subject_is_modern_languages? && has_any_modern_language_subject_type?
+    elsif !main_subject_is_modern_languages? && modern_languages_subjects.present?
       format_name(additional_subjects.push(modern_language_subjects).flatten.uniq.unshift(main_subject))
     else
       format_name(additional_subjects.unshift(main_subject))
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -149,7 +149,7 @@ class CourseDecorator < ApplicationDecorator
 
       # The below is to ensure modern language subjects chosen appear above the second main subject chosen
 
-      if main_subject_id == 33
+      if main_subject_id == SecondarySubject.modern_languages.id
         ordered_subjects = additional_subjects.push(main_subject).reverse
       else
         ordered_subjects = additional_subjects.unshift(main_subject)
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -138,6 +138,27 @@ class CourseDecorator < ApplicationDecorator
     object.subjects.map(&:subject_name).sort.join("<br>").html_safe
   end
 
+  def chosen_subjects(params, course)
+    if params[:course][:master_subject_id].nil?
+      sorted_subjects
+    else
+      main_subject_id = params[:course][:master_subject_id]
+      main_subject = Subject.find(main_subject_id)
+      raw_additional_subjects = course.subjects.map { |subject| subject unless subject.id.to_s == main_subject_id }
+      additional_subjects = raw_additional_subjects.compact
+
+      # The below is to ensure modern language subjects chosen appear above the second main subject chosen
+
+      if main_subject_id == 33.to_s
+        ordered_subjects = additional_subjects.push(main_subject).reverse
+      else
+        ordered_subjects = additional_subjects.unshift(main_subject)
+      end
+
+      ordered_subjects.map(&:subject_name).join("<br>").html_safe
+    end
+  end
+
   def length
     case course_length.to_s
     when "OneYear"
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -138,12 +138,16 @@ class CourseDecorator < ApplicationDecorator
     object.subjects.map(&:subject_name).sort.join("<br>").html_safe
   end
 
-  def chosen_subjects(course)
-    return sorted_subjects if course.master_subject_id.nil?
+  def chosen_subjects
+    return sorted_subjects if master_subject_nil?
 
-    return format_name(additional_subjects.push(main_subject).reverse) if main_subject_is_modern_languages?
-
-    format_name(additional_subjects.unshift(main_subject))
+    if main_subject_is_modern_languages?
+      format_name(select_modern_language_subjects.push(additional_subjects).flatten.uniq.unshift(main_subject))
+    elsif !main_subject_is_modern_languages? && has_any_modern_language_subject_type?
+      format_name(additional_subjects.push(select_modern_language_subjects).flatten.uniq.unshift(main_subject))
+    else
+      format_name(additional_subjects.unshift(main_subject))
+    end
   end
 
   def length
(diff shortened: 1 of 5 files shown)


avinhurry:
[Fix subject order in check your answers and course details page](https://trello.com/c/SK81s9SE/781-fix-subject-order-in-check-your-answers-and-course-details-page)

app/views/publish/courses/_basic_details_tab.html.erb
```diff
@@ -21,7 +21,7 @@
       <% unless course.further_education_course? %>
         <% summary_list.row(html_attributes: { data: { qa: "course__subjects" } }) do |row| %>
           <% row.key { "Subject".pluralize(course.subjects.count) } %>
-          <% row.value { course.sorted_subjects } %>
```
defong, line 24: `sorted_subjects`
Can that be dropped?

app/decorators/course_decorator.rb
```diff
@@ -130,6 +130,27 @@ def sorted_subjects
     object.subjects.map(&:subject_name).sort.join("<br>").html_safe
   end
 
+  def chosen_subjects(course)
+    if course.master_subject_id.nil?
+      sorted_subjects
+    else
+      main_subject_id = course.master_subject_id
```
defong, line 137: May as well change `master_subject_id` to `main_subject_id`

app/decorators/course_decorator.rb
```diff
@@ -130,6 +130,27 @@ def sorted_subjects
...
+      main_subject = Subject.find(main_subject_id)
+      raw_additional_subjects = course.subjects.map { |subject| subject unless subject.id == main_subject_id }
+      additional_subjects = raw_additional_subjects.compact
+
+      # The below is to ensure modern language subjects chosen appear above the second main subject chosen
+
+      if main_subject_id == 33
```
darokel, line 144: Could we fetch the id instead of hardcoding

app/views/publish/courses/_basic_details_tab.html.erb
avinhurry: No, we need it as we would revert back to it if the `master_subject_id` is `nil`

https://github.com/DFE-Digital/publish-teacher-training/pull/3113/files#diff-3803c296e70f15d1a8fafa24c3e9f46a5b22f3d88d3eaead20735ea3ddaf4a73R134-R135

app/decorators/course_decorator.rb
avinhurry: What, you mean rename the column in the db?

app/decorators/course_decorator.rb
avinhurry: Not sure what you mean, something like this? 🤷🏼‍♂️ 

https://github.com/DFE-Digital/publish-teacher-training/pull/3113/commits/d5c39d3e57c18670f28f955f19bf047892573367

darokel: I think you can do `SecondarySubject.physics.id` which is available as a method

app/decorators/course_decorator.rb
defong: yes

avinhurry: Can do in a follow up PR

app/decorators/course_decorator.rb
avinhurry: lol - oh right. Makes sense. Done

app/decorators/course_decorator.rb
```diff
@@ -130,6 +130,27 @@ def sorted_subjects
     object.subjects.map(&:subject_name).sort.join("<br>").html_safe
   end
 
+  def chosen_subjects(course)
```
darokel, line 133: This method is handling quite a few things. I think we could break it up into something like:

```ruby
  def chosen_subjects(course)
    return sorted_subjects if course.master_subject_id.nil?

    return format_name(additional_subjects.push(main_subject).reverse) if main_subject?

    format_name(additional_subjects.unshift(main_subject))
  end

  def additional_subjects
    course.subjects.map { |subject| subject unless subject.id == main_subject.id }.compact
  end

  def main_subject
    Subject.find(course.master_subject_id)
  end

  def main_subject?
    main_subject.id == SecondarySubject.modern_languages.id
  end

  def format_name(subjects)
    subjects.map(&:subject_name).join("<br>").html_safe
  end

```

You may have to tweak and check your tests. Smaller methods are easier to grok and understand and having the `main_subject` as a little helper might help with other bits in this file.

avinhurry: Nice! Now it looks like a professional wrote it 😆 

From 16 lines to 3. 💯

darokel (approved):

app/models/course.rb
```diff
@@ -751,6 +751,18 @@ def remove_carat_from_error_messages
     end
   end
 
+  def select_modern_language_subjects
+    subjects.select { |subject| subject.type == "ModernLanguagesSubject" }
```
darokel, line 755:
```suggestion
subjects.where(type: "ModernLanguagesSubject") so it's a little more performant 
```

darokel: I would also just rename this to `modern_language_subjects`

app/models/course.rb
```diff
@@ -751,6 +751,18 @@ def remove_carat_from_error_messages
...
+  end
+
+  def master_subject_nil?
+    master_subject_id.nil?
+  end
+
+  def has_any_modern_language_subject_type?
+    subjects.any? { |subject| subject.type == "ModernLanguagesSubject" }
```
darokel, line 763: Could this just use the `modern_language_subjects` method above and check for `.present?`

app/models/course.rb
avinhurry: I've renamed the method. 

Using `.where` seems to produce the same result as `select` but i'm getting a weird error. Do you know why? 

```
undefined method `push' for #<ActiveRecord::AssociationRelation [#<ModernLanguagesSubject id: 34, type: "ModernLanguagesSubject", subject_code: "15", subject_name: "French">, #<ModernLanguagesSubject id: 36, type: "ModernLanguagesSubject", subject_code: "17", subject_name: "German">]>
```

app/models/course.rb
avinhurry: Yep, i've changed it.

avinhurry: https://github.com/DFE-Digital/publish-teacher-training/pull/3113/commits/79d82a6489f1860adb88518b69236e9bb13a1b32

app/models/course.rb
avinhurry: https://github.com/DFE-Digital/publish-teacher-training/pull/3113/commits/348be7ef5523a62883968f611d968ba4761845e0

PR #3118, opened 2022-11-09
[576] Migrate the deadline banners component

darokel:
### Context

https://trello.com/c/8vDGqCXG/576-deadline-banners

Migrate the deadline banner component from Find and related code.

### Guidance to review

This just migrates the banner itself and the tests. https://trello.com/c/jrhPxKuT/577-migrate-find-cycles-feature will provide the ability to preview the banner in different states in the service

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/find/cycle_timetable.rb
+++ b/app/services/find/cycle_timetable.rb
@@ -1,0 +1,162 @@
+module Find
+  class CycleTimetable
+    CYCLE_DATES = {
+      2021 => {
+        find_opens: Time.zone.local(2020, 10, 6, 9),
+        apply_opens: Time.zone.local(2020, 10, 13, 9),
+        first_deadline_banner: Time.zone.local(2021, 7, 12, 9),
+        apply_1_deadline: Time.zone.local(2021, 9, 7, 18),
+        apply_2_deadline: Time.zone.local(2021, 9, 21, 18),
+        find_closes: Time.zone.local(2021, 10, 4, 23, 59, 59),
+      },
+      2022 => {
+        find_opens: Time.zone.local(2021, 10, 5, 9),
+        apply_opens: Time.zone.local(2021, 10, 12, 9),
+        first_deadline_banner: Time.zone.local(2022, 8, 2, 9),
+        apply_1_deadline: Time.zone.local(2022, 9, 6, 18),
+        apply_2_deadline: Time.zone.local(2022, 9, 20, 18),
+        find_closes: Time.zone.local(2022, 10, 3, 23, 59, 59),
+      },
+      2023 => {
+        find_opens: Time.zone.local(2022, 10, 4, 9),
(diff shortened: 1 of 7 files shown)


darokel:
[Deadline banners](https://trello.com/c/8vDGqCXG/576-deadline-banners)

app/services/find/cycle_timetable.rb
```diff
@@ -0,0 +1,172 @@
...
+    def self.apply_reopens
+      date(:apply_opens, next_year)
+    end
+
+    def self.preview_mode?
+      Time.zone.now.between?(apply_2_deadline, find_closes)
+    end
+
+    def self.find_down?
+      return true if current_cycle_schedule == :today_is_after_find_closes
```
avinhurry, line 104: Do we need the return here, or can we just have `true if current_cycle_schedule == :today_is_after_find_closes` ?

avinhurry: and for the ones below

avinhurry (commented):
Can we have view previews since we've started adding them for most of the other components?

app/services/find/cycle_timetable.rb
darokel: it needs to be explicitly returned as the last line takes precedence

darokel:
> Can we have view previews since we've started adding them for most of the other components?

For this one I don't think we need need to as the cycles switcher will allow us to preview it (realistically) in different states. Otherwise we'd have to mock out the CycleTimetable model just for the sake of the preview which feels like duplicating what's already available.

app/services/find/cycle_timetable.rb
darokel: it does look wonky though as it could just be `current_cycle_schedule == :today_is_after_find_closes || Time.zone.now.between?(find_closes, find_reopens)` but

avinhurry (approved):

app/services/find/cycle_timetable.rb
```diff
@@ -0,0 +1,162 @@
...
+        first_deadline_banner: Time.zone.local(2022, 8, 2, 9),
+        apply_1_deadline: Time.zone.local(2022, 9, 6, 18),
+        apply_2_deadline: Time.zone.local(2022, 9, 20, 18),
+        find_closes: Time.zone.local(2022, 10, 3, 23, 59, 59),
+      },
+      2023 => {
+        find_opens: Time.zone.local(2022, 10, 4, 9),
+        apply_opens: Time.zone.local(2022, 10, 11, 9),
+
+        # NOTE: the dates from below here are not the finalised but are required
```
defong, line 24: :face_exhaling:

defong (approved):

PR #3140, opened 2022-11-18
[0827] The sort by field should not auto submit we should provide a submit button

defong:
### Context
Sort field

Amended to show the sort button
Removed the related js that auto submits

### Guidance to review

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/assets/stylesheets/components/find/_search-results.scss
+++ b/app/assets/stylesheets/components/find/_search-results.scss
@@ -30,7 +30,6 @@
   text-decoration: underline;
 }
 
-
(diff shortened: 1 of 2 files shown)


defong:
[The sort-by field should not auto-submit; we should provide a submit button](https://trello.com/c/pbujy3Cx/827-the-sort-by-field-should-not-auto-submit-we-should-provide-a-submit-button)

darokel (approved):

PR #3156, opened 2022-11-30
#849 results filter

goodviber:
### Context

We're migrating the Find codebase over to the Publish codebase. Now that the searching and the search results page is wired up. We need to make sure the filters still work as before.
**What needs to be done**

Make sure all the search result filters produce the correct relevant query params to filter the search result

When un-toggling any of the filters make sure it is correctly removed and the original request url (before the filter was applied) remains

**Success**

All the filters work as before.

Changes the qualifcations param for `other` to be more readable, and also adds `pgde_with_qts` to course_search_service.rb qualifications method.
### Guidance to review
GOTO /find, enter search params, should replicate results from running FIND locally
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/services/course_search_service.rb
```diff
@@ -11,13 +11,12 @@ def initialize(
     @sort = Set.new(sort&.split(","))
   end
 
-  PROVIDER_ASCENDING = Set["name", "provider.provider_name"].freeze
```
darokel, line 14: I think this set which fields to order on (eg course name and the provider name in this case). Without it, it's just ordering on the provider name:

Review app:

Screenshot 2022-12-02 at 12 16 19

Find:

Screenshot 2022-12-02 at 12 16 36

darokel: cc @defong

defong: This look like a bad merge...

defong (changes requested):
There should be minimal changes to app/services/course_search_service.rb

Use a shim or a facade pattern to transform the  course_search_service `filter` hash values

app/services/course_search_service.rb
goodviber: I think the sorting is done by `ascending_canonical_order`? It seems to be working ok now

darokel (approved):

app/services/course_search_service.rb
```diff
@@ -166,11 +165,11 @@ def locations_filter?
   end
 
   def sort_by_provider_ascending?
-    sort == CourseSearchService::PROVIDER_ASCENDING
+    sort == Set["0"] || sort == Set["name", "provider.provider_name"]
   end
 
   def sort_by_provider_descending?
-    sort == CourseSearchService::PROVIDER_DESCENDING
+    sort == Set["1"] || sort == Set["name", "-provider.provider_name"]
```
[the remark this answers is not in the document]
goodviber: Good spot

PR #3165, opened 2022-12-05
update for visa sponsorship issue

goodviber:
### Context
Instruction to fix visa sponsorship issue in support playbook
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/docs/support_playbook.md
+++ b/docs/support_playbook.md
@@ -1,5 +1,4 @@
-Support Playbook
-================
+# Support Playbook
 
 ## Removing an organisation
 
@@ -63,6 +62,7 @@ To change the accrediting body of a course, you can do the following:
 p = RecruitmentCycle.current.providers.find_by(provider_code: "1YP")
 p.courses.update(accredited_body_code: "1YK")
 ```
+
 ## Transfer courses to another provider
 
 ```ruby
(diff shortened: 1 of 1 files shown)


docs/support_playbook.md
```diff
@@ -120,8 +120,19 @@ copier = Courses::CopyToProviderService.new(sites_copy_to_course: Sites::CopyToC
...
+
+## Visa sponsorship issue
+
+If providers encounter an error message 'Select if visas can be sponsored' when publishing a course after correctly filling the visa option on basic details page, they need to have the provider sponsorship attributes updated from nil, as we are validating on the provider rather than the course at present.
+
+```ruby
+# Find provider to update
+provider = RecruitmentCycle.current.providers.find_by(provider_code: "1TZ")
+
+provider.update(can_sponsor_skilled_worker_visa: true, can_sponsor_student_visa: true)
```
avinhurry, line 137: Does it matter whether we set these to `true` or `false`?

darokel: When creating courses, depending on the provider type, these values (at the course level) will be defaulted to what the provider has set. (eg a scitt creating a course could have these defaulted to what the accrediting body has set even though it can be changed)

goodviber: No, the validation is ```def validate_provider_visa_sponsorship_publishable
    if provider.can_sponsor_student_visa.nil? || provider.can_sponsor_skilled_worker_visa.nil?
      errors.add(:base, :visa_sponsorship_not_publishable)
    end
  end```

goodviber: but is seems more logical?

PR #3167, opened 2022-12-06
[860] Match feature coverage for old find

darokel:
### Context

Now that most of the core logic from Find has been migrated into New Find, we need to make sure we have the same level of coverage for the feature specs.

This PR introduces new feature specs matching those in old Find refactored to work properly in the Publish codebase. It also makes small fixes to things which may have been missed in some parts of the code.

### Guidance to review

Mainly the specs

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


darokel:
[Check new find feature specs match coverage of old Find](https://trello.com/c/kctgtVnD/860-check-new-find-feature-specs-match-coverage-of-old-find)

spec/features/find/search/across_england/primary_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+    expect(page).to have_content("Primary courses with subject specialisms")
+  end
+
+  def and_i_click_find_courses
+    click_button "Find courses"
+  end
+
+  def when_i_click_find_courses
+    and_i_click_find_courses
+  end
```
defong, line 121: Something looks off

spec/features/find/search/across_england/primary_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+
+  def then_i_should_see_a_subjects_validation_error
+    expect(page).to have_content("Select at least one subject")
+  end
+
+  def when_i_select_the_primary_subject_textbox
+    check "Primary"
```
defong, line 128: Umm page object?

app/decorators/site_decorator.rb
```diff
@@ -0,0 +1,7 @@
+class SiteDecorator < Draper::Decorator
+  delegate_all
+
+  def full_address
```
defong, line 4: Do we really need the decorator for this?

app/views/find/courses/show.html.erb
```diff
@@ -1,21 +1,32 @@
...
+    <p class="govuk-body-l" data-qa="course__extended_qualification_descriptions"><%= @course.extended_qualification_descriptions %></p>
+  </div>
+</div>
 
 <%= render FindInterface::Courses::SummaryComponent::View.new(@course) %>
 
-<!--The reason for the Apply component duplication here is because we want to keep the warning text at the top but Keep the Apply button and end of cycle text at the bottom. Uncomment once CycleTimetable moved over-->
-
-<%# <%= render FindInterface::Courses::ApplyComponent::View.new(@course) if @course.has_vacancies? || !CycleTimetable.mid_cycle? %>
+<%= render FindInterface::Courses::ApplyComponent::View.new(@course) unless @course.has_vacancies? %>
```
defong, line 29: We want to show the thing unless there is vacancies?

spec/features/find/search/across_england/primary_spec.rb
darokel: Sorted

spec/features/find/search/across_england/primary_spec.rb
darokel: I've tried to match the feature specs as they are because some of them go back and forth in the scenarios (triggering errors) which causes the input ids to change. If we want to fully use the page objects for all of it we'll need to refactor some of the form markup to use the GOV UK form builder for consistency. We have an upcoming sprint focused on tech debt so we can address it then as part of a New Find tech debt clean up.

app/decorators/site_decorator.rb
darokel: Probably not but I don't think it should go into the model either as it's mainly for presentational purposes. Will leave it for now but happy to bin it later when we clean things up and after we have an alternative.

app/views/find/courses/show.html.erb
darokel: I think this was a mistake in old Find, have completed removed it so it only shows the one at the bottom.

spec/features/find/search/across_england/primary_spec.rb
defong: I'm not even sure should we stick to page object...

What do you think?
@avinhurry @goodviber

defong (approved):

spec/features/find/search/across_england/primary_spec.rb
darokel: I'm at a stage in life where there is beauty in simplicity. I don't think we need to take a hard stance on this. We'll use it where it's appropriate (drying things up) otherwise just keep things very simple. We can finalise in next breakdown.

PR #3170, opened 2022-12-06
[0872] Migrate the cookie banner

defong:
### Context

Cookie banner for `Find`

Use the existing Cookie banner

### Guidance to review
They share the same underlying logic, since the site will be hosted on different domain hence it will be applicable per domain, rather then coding for multiverse in the universe.
 
First cut out of a death from a thousand cuts?

No GA related items

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


defong:
[Migrate the cookie banner](https://trello.com/c/rGNHsDGI/872-migrate-the-cookie-banner)

spec/support/page_objects/find/age_groups.rb
```diff
@@ -10,7 +10,7 @@ class AgeGroups < PageObjects::Base
       element :primary, "#find-age-groups-form-age-group-primary-field"
```
goodviber, line 10: should we use data-qa to identify test elements?

defong: No, 

1. There are already existing attributes that can be used
2. They end up being bloating the html 
3. They increase payload
4. They end up being tech debt
5. They serve no value other then for testing

goodviber (approved):

PR #3159, opened 2022-11-30
Managing users in publish (Add user)

avinhurry:
### Context

Introducing a feature for providers to create Publish users and associate them with specific organisations. The users will also be sent an email, with the link and inviting them to sign up to DfE Sign in, if they haven't already.

### Changes proposed in this pull request

- Create new views, routes, controllers and form.
- Store changes in redis when submitting the form and going to the check page.
- Update the database with the changes stored in redis on submission of the check page.
- Values in the form are persisted if you cancel or go back from the check page. If you have accidently gone to the wrong provider, when you go to the correct one the form will be prefilled.
- Utilise the `UserAssociationService`.
- If the user does not exist on Publish, create the user and associate them with the provider.
- If the user does exist on Publish, find them and associate them with the provider.

### Guidance to review

Have a play around in the review app, compare with prototype, are there any differences or bugs?

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally

PR #3171, opened 2022-12-06
Managing users in publish (removing users)

avinhurry:
### Context

Following on from #3159 and #3166. The next feature in the managing users epic is the ability for providers to remove other provider users.

- Add routes for deleting
- Add the remove user link on the show page
- Add a `delete?` method to the provider policy

### Guidance to review

Try it out on the review app.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/publish/users_check_controller.rb
+++ b/app/controllers/publish/users_check_controller.rb
@@ -1,7 +1,8 @@
 module Publish
   class UsersCheckController < PublishController
+    before_action :authorize_provider
+
     def show
-      authorize(provider)
       @user_form = UserForm.new(current_user, user)
     end
 
@@ -9,10 +10,15 @@ module Publish
       @user_form = UserForm.new(current_user, user)
       if @user_form.save!
         UserAssociationsService::Create.call(user: @user_form.model, provider:) if @user_form.model.providers.exclude?(provider)
-        authorize(provider)
         redirect_to publish_provider_users_path(params[:provider_code])
         flash[:success] = "User added"
       end
+    end
+
+  private
+
+    def authorize_provider
+      authorize(provider)
     end
 
     def user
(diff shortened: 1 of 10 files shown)


app/controllers/publish/users_check_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Publish
+  class UsersCheckController < PublishController
+    def show
+      authorize(provider)
```
goodviber, line 4: could we move the authorize to a before_action?

app/controllers/publish/users_controller.rb
```diff
@@ -6,8 +6,54 @@ def index
       @users = provider.users
     end
 
+    def show
+      authorize(provider)
```
goodviber, line 10: same here?

app/controllers/publish/users_check_controller.rb
avinhurry: Good idea! 

Done here: https://github.com/DFE-Digital/publish-teacher-training/pull/3171/commits/1915b854a174dc708eb2593c223e7e7b1fd86728

app/controllers/publish/users_controller.rb
avinhurry: 👍🏼

spec/support/page_objects/publish/users_check.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module PageObjects
+  module Publish
+    class UsersCheck < PageObjects::Base
+      set_url "/publish/organisations/{provider_code}/users/check"
+
+      element :add_user, ".govuk-button", text: "Add user"
```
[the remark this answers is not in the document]
avinhurry: I don't think it makes sense to add more stuff to the html just for testing when we can test without it. We're also thinking of dropping the whole page objects pattern...

defong (approved):

simonwhatley (approved):

PR #3180, opened 2022-12-13
818 fix modern languages issue

goodviber:
### Context
When creating a course with two subjects, for example, English with Modern languages (French and Spanish), editing the course does not maintain the choice hierarchy.

In this example, Modern languages switches to be the first subject and English the second.
**What needs to be done**

The choice hierarchy needs to be maintained when editing subjects on a course.
Technical notes

This only seems to appear when setting ML as an additional subject alongside some other non language main subject
There might be some other underlying issue here as this scenario doesn't seem to set the course_subject positions correctly.

Also when editing from the confirmation page, the modern languages flow does not include the language option page.

Create a check in etp controller to fix language flow from confirmation page

Alter subject form to track master and subordinate subjects

List Modern languages together
### Guidance to review
From confirmation or basic details pages, change subjects to modern languages etc
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


goodviber:
[Changing subject does not maintain first and second subject order](https://trello.com/c/hsF1kHsx/818-changing-subject-does-not-maintain-first-and-second-subject-order)

simonwhatley:
I first created a Chemistry with Modern Languages (French and German) course. The list is in the correct order.

image

Before adding the course, I changed the subject order to Modern Languages (French and German) with Chemistry. The list on the 'check your answers' page is not in the correct order.

image

app/controllers/publish/courses/engineers_teach_physics_controller.rb
```diff
@@ -10,6 +10,10 @@ def new
         authorize(@provider, :can_create_course?)
         return if has_physics_subject?
 
+        if params[:goto_confirmation] && params[:course][:subjects_ids]&.include?(modern_languages_id)
```
darokel, line 13: Might be worth putting some of these checks into smaller methods, some of the actions in this controller are getting quite large. Eg I noticed `params[:course][:subjects_ids].include?(modern_languages_id)` is also used somewhere else in this file

darokel (approved):

app/controllers/publish/courses/engineers_teach_physics_controller.rb
goodviber: Done

simonwhatley (changes requested):
The first and second subject orders are now correct. However, the list of languages needs to be alphabetical:

image

In this example, the list of subjects should read:
- Modern Languages
- French
- German
- Chemistry

simonwhatley (approved):

PR #3194, opened 2022-12-19
869 update error messages

goodviber:
### Context
Recently, as part of an accessibility update, we added an "Error: " prefix to all inline error messages.

During this work, it was apparent that the error messages are not formatted correctly, and the content can be improved.
**What needs to be done**

correctly display error summaries as per the [GOV.UK](http://gov.uk/) Design System
correct display form field errors as per the [GOV.UK](http://gov.uk/) Design System
    update the error message content

Update error messages on find search flow and split messages in subjects controller
### Guidance to review
GOTO /find and continue with empty inputs
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/find/search/subjects_controller.rb
+++ b/app/controllers/find/search/subjects_controller.rb
@@ -18,6 +18,8 @@ module Find
             has_vacancies: default_vacancies,
           ))
         else
+          @subjects_form.errors.clear
+          @subjects_form.primary? ? @subjects_form.errors.add(:subjects, :primary_subject) : @subjects_form.errors.add(:subjects, :secondary_subject)
           render :new
         end
       end
--- a/app/controllers/find/search/subjects_controller.rb
+++ b/app/controllers/find/search/subjects_controller.rb
@@ -18,8 +18,6 @@ module Find
             has_vacancies: default_vacancies,
           ))
         else
-          @subjects_form.errors.clear
-          @subjects_form.primary? ? @subjects_form.errors.add(:subjects, :primary_subject) : @subjects_form.errors.add(:subjects, :secondary_subject)
           render :new
         end
       end
--- a/app/forms/find/subjects_form.rb
+++ b/app/forms/find/subjects_form.rb
@@ -4,7 +4,13 @@ module Find
 
     attr_accessor :subjects, :age_group
 
-    validates :subjects, presence: true
+    validate :subjects_have_been_selected
+
+    def subjects_have_been_selected
+      if subjects.blank? && age_group.present?
+        errors.add(:subjects, :"#{age_group}_subject")
+      end
+    end
 
     def primary?
       age_group == "primary"
--- a/app/forms/find/subjects_form.rb
+++ b/app/forms/find/subjects_form.rb
@@ -9,6 +9,8 @@ module Find
     def subjects_have_been_selected
       if subjects.blank? && age_group.present?
         errors.add(:subjects, :"#{age_group}_subject")
+      elsif subjects.blank?
+        errors.add(:subjects, :blank)
       end
     end
 
--- a/app/forms/find/subjects_form.rb
+++ b/app/forms/find/subjects_form.rb
@@ -2,15 +2,18 @@ module Find
   class SubjectsForm
     include ActiveModel::Model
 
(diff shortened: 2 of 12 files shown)


app/controllers/find/search/subjects_controller.rb
```diff
@@ -18,6 +18,8 @@ def create
             has_vacancies: default_vacancies,
           ))
         else
+          @subjects_form.errors.clear
+          @subjects_form.primary? ? @subjects_form.errors.add(:subjects, :primary_subject) : @subjects_form.errors.add(:subjects, :secondary_subject)
```
defong, line 22: This looks like it should belong to the form

goodviber: Yeah, I gave that a whirl, but the actual form class only has access to params once it is instantiated. I think we would require 2 forms, a primary and secondary etc

simonwhatley (changes requested):
**Find courses by location or provider**

The error summary should be full width
image

If I submit the form without selecting a radio, the error message shows (this works as intended). But, if I then select one of the options with a conditional reveal, the field inside should not be in the error state. The field in the conditional reveal is only in error when I try and submit without completing the field (i.e., search for a location or a provider).

image

**Select which age group you want to teach**

The error summary should be full width
image

app/controllers/find/search/subjects_controller.rb
defong:
```
module Find
  class SubjectsForm

    # remove
    # validates :subjects, presence: true

    validate :subjects_has_been_selected

    def subjects_has_been_selected
      if subjects.blank? && age_group.present?
        @subjects_form.errors.add(:subjects, "#{age_group}_subject")
      end
    end
   # brevity
  end
end
```

You just need to move your current logic down to the form, as you have already have the context on it.

goodviber: Nice

app/forms/find/subjects_form.rb
```diff
@@ -4,7 +4,15 @@ class SubjectsForm
...
 
-    validates :subjects, presence: true
+    validate :subjects_have_been_selected
+
+    def subjects_have_been_selected
+      if subjects.blank? && age_group.present?
+        errors.add(:subjects, :"#{age_group}_subject")
+      elsif subjects.blank?
+        errors.add(:subjects, :blank)
+      end
```
defong, line 14: I don't think it is possible scenario. Can we remove?

goodviber: From a scenario perspective I agree, but from a philosophical perspective we get 
```
 [6] pry(main)> s = Find::SubjectsForm.new
=> #<Find::SubjectsForm:0x0000000110919e78>
[7] pry(main)> s.valid?
=> true
```

defong: Can we add a test for it then its not obvious, why its there and the lack of tests doesn't help .

goodviber: We have a test for it? 
```
it "is not valid when subject codes are not present" do
        form = described_class.new

        expect(form.valid?).to be(false)
      end
```

defong: As far as I can, see we only use it here:

https://github.com/DFE-Digital/publish-teacher-training/blob/81b1d1585b35b798cc7b418258f4eccfd4d3f0a9/app/controllers/find/search/subjects_controller.rb#L9

https://github.com/DFE-Digital/publish-teacher-training/blob/81b1d1585b35b798cc7b418258f4eccfd4d3f0a9/app/controllers/find/search/subjects_controller.rb#L13


Can we fix this class altogether, as the both our discussion and tests highlight it tech debt rolling forward with us, and distracting us, so can we boy scout :

1. remove
```
     elsif subjects.blank?
        errors.add(:subjects, :blank)
      end
```

2. amend
```
   #  attr_accessor :subjects, :age_group
   # to
    attr_reader :subjects, :age_group
```

3. add
```
    def initialize(subjects:, age_group:)
      @subjects = subjects
      @age_group = age_group
    end
```

defong (changes requested):
Let fix the tech debts

PR #3205, opened 2022-12-23
577 migrate find cycles feature

syed87:
### Context

The following page and functionality need to be migrated to the new Find: [Recruitment cycles - Find postgraduate teacher training - GOV.UK](https://qa.find-postgraduate-teacher-training.service.gov.uk/cycles)

What needs to be done
Migrate the cycles page/functionality
Success
It's migrated

Add find/cycles page

### Guidance to review

Cycles page should look the same and provide the same features as cycles page in find project minus the ability to redirect results page to cycle-has-ended if 'find has closed' is selected on cycles page. This feature will be added at a later date.

### Checklist

✅ Make sure all information from the Trello card is in here
✅ Attach to Trello card
✅ Rebased main
✅ Cleaned commit history
✅ Tested by running locally

### DIFF
--- a/spec/features/find/switcher_cycles_spec.rb
+++ b/spec/features/find/switcher_cycles_spec.rb
@@ -9,34 +9,34 @@ feature "switcher cycle" do
 
   scenario "Mid cycle and deadlines should be displayed" do
     when_i_visit_switcher_cycle_page
-    and_i_choose_mid_cycle_and_deadlines_should_be_displayed_option
+    and_i_choose("Mid cycle and deadlines should be displayed")
     then_i_click_on_update_button
     and_i_should_see_the_success_banner
     and_i_visit_results_page
-    and_i_see_mid_cycle_and_deadlines_should_be_displayed_banner
+    and_i_see_deadline_banner("Apply now to get on a course starting in the 2023 to 2024 academic year")
   end
 
   scenario "Update to Apply 1 deadline has passed" do
     when_i_visit_switcher_cycle_page
-    and_i_choose_apply_1_deadline_has_passed_option
+    and_i_choose("Apply 1 deadline has passed")
     then_i_click_on_update_button
     and_i_should_see_the_success_banner
     and_i_visit_results_page
-    and_i_see_apply_1_deadline_has_passed_banner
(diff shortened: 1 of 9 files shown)


defong:
[Migrate Find `/cycles` feature](https://trello.com/c/jrhPxKuT/577-migrate-find-cycles-feature)

spec/features/find/switcher_cycles_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  scenario "Update to Apply 2 deadline has passed" do
+    when_i_visit_switcher_cycle_page
+    and_i_choose("Apply 2 deadline has passed")
+    then_i_click_on_update_button
+    and_i_should_see_the_success_banner
+    and_i_visit_results_page
+    and_i_see_deadline_banner("Courses are currently closed but you can get your application ready")
+  end
+
+  # scenario "Find has closed" do
```
darokel, line 37: @avinhurry FYI i think you may need to utilise this in your branch/PR?

darokel (approved):

spec/features/find/switcher_cycles_spec.rb
```diff
@@ -22,7 +22,7 @@
     then_i_click_on_update_button
     and_i_should_see_the_success_banner
     and_i_visit_results_page
-    and_i_see_deadline_banner("If you’re applying for the first time since applications opened in December 2022")
+    and_i_see_deadline_banner("If your application did not lead to a place and you’re applying again")
```
darokel, line 25: This test suite passes locally before this change so I'm not sure if this is the correct text which is supposed to be shown for this scenario?

darokel: I guess we should go with a piece of text which will only get shown in that scenario to check the logic? If that's the case then happy to go a head so we can get this in. The CI thing though 🤔

avinhurry: Whoops. I thought that was only shown in that scenario, but turns out it isn't. I've changed it to one that is. 

Happy to merge?

darokel: yup 👍

PR #3225, opened 2023-01-04
916 rename component files

goodviber:
### Context
Our publish components are setup to have the following folder structure:

/components/foo/view.html.erb
/components/foo/view.rb

This is redundant extra boilerplate and can be simplified to just:

/components/foo.html.erb
/components/foo.rb
What needs to be done

Update all the publish components to follow the preferred structure above

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


goodviber:
[Rename components to a more flat structure](https://trello.com/c/gwQCHgFm/916-rename-components-to-a-more-flat-structure)

app/components/degree_row_content.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+class DegreeRowContent < ViewComponent::Base
```
avinhurry, line 3: Is it worth putting all of these in a `Publish` folder to help keep them separate, since we have a folder for Find?

avinhurry (commented):
I think you missed a few? The components in these locations haven't been renamed: 

- `app/components/filters`
- `app/components/filters/provider_attributes`
- `app/components/filters/user_attributes`
-`app/components/find/courses`
- `app/components/find/header`
- `app/components/find/phase_banner`
- `app/components/find/utility/advice_component`
- `app/components/providers/provider_list`

And the respective tests.

app/components/degree_row_content.rb
darokel: Eventually i think we should do that for as many files as possible to create boundaries but can leave it for another ticket

goodviber: We are leaving the multi nested folders for now, there is some higher priority stuff to get on with....the app/components/foo.rb went back and forth for a bit, settled on present structure.

spec/components/find_interface/header/view_preview.rb
```diff
@@ -3,11 +3,11 @@
 module Find::Header
```
defong, line 3: hmmm this sticks out like a sore thumb :thinking:

goodviber: Yes, we need to rename the spec../find_interface.

goodviber: Done

defong (approved):
LGTM

PR #3235, opened 2023-01-06
Alphabetize users list

avinhurry:
### Context

We are sorting the users list alphabetically based on first name and then last name. 

- Sort the list using active record in the controller

### Guidance to review

Is the list ordered in the way we want it to be? 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/publish/users_controller.rb
+++ b/app/controllers/publish/users_controller.rb
@@ -3,7 +3,7 @@ module Publish
     before_action :authorize_provider
 
     def index
-      @users = provider.users
+      @users = provider.users.order("LOWER(first_name), LOWER(last_name)")
     end
 
     def show
--- a/app/controllers/publish/users_controller.rb
+++ b/app/controllers/publish/users_controller.rb
@@ -3,7 +3,7 @@ module Publish
     before_action :authorize_provider
 
     def index
-      @users = provider.users.order("LOWER(first_name), LOWER(last_name)")
+      @users = provider.users.in_name_order
     end
 
     def show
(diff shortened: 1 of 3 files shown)


avinhurry:
[Alphabetize users list](https://trello.com/c/nv3WVfGC/933-alphabetize-users-list)

app/controllers/publish/users_controller.rb
```diff
@@ -3,7 +3,7 @@ class UsersController < PublishController
     before_action :authorize_provider
 
     def index
-      @users = provider.users
+      @users = provider.users.order("LOWER(first_name), LOWER(last_name)")
```
darokel, line 6: I think this should be moved down into the model under a scope then it'll read a little better in the controller:

`provider.users.in_name_order` or something

avinhurry: 👍🏼 https://github.com/DFE-Digital/publish-teacher-training/pull/3235/commits/3478c10439b503e84e50944d923dec7321b8be0f

spec/models/user_spec.rb
```diff
@@ -289,4 +289,20 @@
       end
     end
   end
+
+  describe "#canonical_order" do
```
darokel, line 293:
```suggestion
describe '. in_name_order' do
```

darokel (approved):

spec/models/user_spec.rb
avinhurry: oops

simonwhatley (approved):

PR #3236, opened 2023-01-06
Prevent numbers in names

avinhurry:
### Context

We are preventing users from including numbers in names.

### Screenshot

image

Add a validation that prevents the addition of numbers in first name and last name

### Guidance to review

Is the content in the error message correct?

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/validators/name_validator.rb
+++ b/app/validators/name_validator.rb
@@ -1,0 +1,13 @@
+class NameValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    if is_invalid_name_format?(value)
+      record.errors.add(attribute, message: options[:message])
+    end
+  end
+
+private
+
+  def is_invalid_name_format?(value)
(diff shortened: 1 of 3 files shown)


avinhurry:
[Tighten name validations for add/edit user](https://trello.com/c/x8V1fcn4/937-tighten-name-validations-for-add-edit-user)

app/validators/name_validator.rb
```diff
@@ -0,0 +1,13 @@
+class NameValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    if is_invalid_name_format?(value)
```
darokel, line 3:
```suggestion
`if name_contains_digits?(value)`
```

avinhurry:
We've decided not to do this anymore as in some cases names contain numbers/uncommon characters. 

Until it becomes a problem we will allow these.

PR #3242, opened 2023-01-10
[876] Tighten up url boundaries

darokel:
### Context

This PR sets up the correct routing for Find in non dev environments and documents how to configure local urls for development.

- Put all app routes into their respective constraints to prevent accessing paths falling into the other's domain
- Fix all request/feature specs to enforce respective constraint
- Makes sure the correct error layouts are shown for each domain
- Documents how to set up local urls for development

### Guidance to review

This PR has been tested in the QA environment. Please pull it down, configure the urls as described in the documentation and try to access the services locally.

### Notes

~~We're seeing if we can get temp aliases for review apps for the migrated Find version so we can have:~~

~~`https://find2-pr-x.london.cloudapps.digital/` point to the app to activate the routing constraint for Find.~~

This already been setup now.

https://publish-teacher-training-pr-3242.london.cloudapps.digital/ - Should load publish
https://find2-pr-3242.london.cloudapps.digital - Should load Find

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


darokel:
[Implement and prepare the new Find URL for qa](https://trello.com/c/vOimr5Zk/876-implement-and-prepare-the-new-find-url-for-qa)

spec/support/stubbed_constraint.rb
```diff
@@ -0,0 +1,9 @@
+RSpec.configure do |config|
+  config.before(:each, :with_find_constraint) do
+    allow_any_instance_of(FindConstraint).to receive(:matches?).and_return(true)
+  end
+
+  config.before(:each, :with_publish_constraint) do
+    allow_any_instance_of(PublishConstraint).to receive(:matches?).and_return(true)
+  end
```
defong, line 8: Can we brack these up to there own file and auto mark it

```
  config.define_derived_metadata(file_path: Regexp.new("/spec/features/publish")) do |metadata|
    metadata[:with_publish_constraint] = true
  end

  config.before(:each, :with_publish_constraint) do
    allow_any_instance_of(PublishConstraint).to receive(:matches?).and_return(true)
  end
```

?

darokel: Nice idea. Done. I've removed the tag from `publish|find` feature directories

app/lib/find_constraint.rb
```diff
@@ -2,6 +2,6 @@
 
 class FindConstraint
   def matches?(request)
-    Settings.find_temp_url&.include?(request.host)
+    Settings.find_temp_url&.include?(request.host) || request.host.include?("find2-pr")
```
defong, line 5: Hmmm on second thought, maybe this is better, as we don't really know the review env url and we don't care enough for the test env.

```suggestion
    Settings.find_temp_url&.include?(request.host) || Rails.env.review? || Rails.env.test?
```

Then we can remove `StubbedFindConstraint` altogether and the noise it created.

darokel: I don't think this makes sense because we need to be explicit to only activate the constraint depending on the URL (which will be different for find and publish review apps) otherwise some of the routes will conflict if both constraints are activated without checking it. Same re: the test, the stub gives more granular control

guides/setup-development.md
```diff
@@ -94,3 +94,29 @@ Then run the following command to populate the database:
...
+}
+```
+
+And make sure to set the following settings in `config/settings/development.local.yml`:
+
+```ruby
+# Settings.base_url to https://publish.localhost
+# Settings.find_temp_url to https://find.localhost
+# Settings.use_ssl to true
+```
```
defong, line 120: Can we move that to `development.yml`?

Is this also used for DfESignIn?

defong (approved):

PR #3247, opened 2023-01-11
Update results page ordering options

avinhurry:
### Context

We are adding the ability to sort the results page on Find by course name.

- **Add course ordering (A-Z) (default)**

If identical course, ordering falls back to the provider, if identical providers, ordering falls back to the course code

- **Add course ordering (Z-A)**

If identical course, ordering falls back to the provider, if identical providers, ordering falls back to the course code

- Remove the unnecessary use of Sort. This was probably added for future proofing, but there has been no need for it and it just causes confusion.

- Remove the constants and compare directly with strings encompassed within methods. This will make it easier to remove the code relating to old find when we delete it.

### Guidance to review

Search for a few courses and check that the hierarchy above is correct.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/find/results/sort_by_component.html.erb
+++ b/app/components/find/results/sort_by_component.html.erb
@@ -7,7 +7,7 @@
       <%= form.label(:sortby, "Sorted by", class: "govuk-label govuk-!-display-inline-block") %>
       <%= form.select(
         :sortby,
-        options_for_select(results.sort_options, selected: params["sortby"] || "A"),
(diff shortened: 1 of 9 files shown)


avinhurry:
[Sort courses by course name](https://trello.com/c/QrMrqrVS/897-sort-courses-by-course-name)

defong (approved):

defong (approved):

PR #3248, opened 2023-01-15
939 old find params

goodviber:
### Context
We've changed Find query params in the migrated version to use the same search params as the CourseSearchService to make it easier to maintain.

We should try to handle URLs with the old query params as well.
What needs to be done

Figure out a way to check an incoming url for old find query params and transform/redirect the request to the new query params so the user still gets to the same place

Success

Given an old find url such as /results?age_group=secondary&l=2&subject_codes%5B%5D=W1&subject_codes%5B%5D=C1&subject_codes%5B%5D=08 the user should be taken to the new find equivalent
Tech notes

Could use a routing constraint to check for the old url within a redirect block in the routes with some additional logic to transform the url
We'll probably only need to check urls beginning with the /results* prefix.
I think we can only check one way eg first land on the page, once the user starts modifying their search they'll be using the new params anyway
Keep it simple this is just to enhance and preserve UX

Creation of a params mapping service to be called in the results :index
### Guidance to review
run the new results page in the browser with params from old find
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


goodviber:
[Handle old find query params](https://trello.com/c/coZIGMe8/939-handle-old-find-query-params)

darokel:
@goodviber Could you rebase so we could get the constraint in here to see the review app?

.github/actions/smoke-test/action.yml
```diff
@@ -40,14 +40,14 @@ runs:
       shell: bash
       run: RAILS_ENV=${{ inputs.environment }} ./bin/bundle exec rspec spec/smoke --format documentation
 
-    - name: 'Notify #twd_findpub_tech on failure'
```
defong, line 43: Umm can we revert these changes please

app/services/find/match_old_params.rb
```diff
@@ -0,0 +1,62 @@
...
+      "Other" => "pgce pgde",
+      "PgdePgceWithQts" => "pgce_with_qts",
+      "QtsOnly" => "qts",
+    }.freeze
+
+    def initialize(request_params)
+      @request_params = request_params
+    end
+
+    def call
```
defong, line 30: Can we break this method to smaller private methods, akin to the way you have tested it

@darokel I think we need to revisit https://github.com/DFE-Digital/publish-teacher-training/blob/main/config/rubocop/metrics.yml

It seems we induced needless tech debt.

goodviber: Done

.github/actions/smoke-test/action.yml
goodviber: vimmed

app/services/find/match_old_params.rb
```diff
@@ -0,0 +1,78 @@
...
+  private
+
+    def sortby_and_funding
+      @request_params["sortby"] = "distance" if @request_params["sortby"] == "2"
+      @request_params["funding"] = "salary" if @request_params["funding"] == "8"
+    end
+
+    def qualification_filters
+      if @request_params["qualifications"]
+        @request_params["qualification"] = @request_params.delete("qualifications")
```
defong, line 48: :woozy_face:

app/views/find/result_filters/_hidden_fields.html.erb
```diff
@@ -9,3 +9,9 @@
     <%= form.hidden_field(:subjects, multiple: true, value: subject_code) %>
   <% end %>
 <% end %>
+<!-- to preserve old find subject codes -->
```
defong, line 12:
```suggestion
<!-- NOTE: to preserve old find subject codes -->
```

app/views/layouts/find_layout.html.erb
```diff
@@ -53,7 +53,6 @@
         ),
       ) %>
     <% end %>
-
```
defong, line 56: Can we have it back please.

spec/services/find/match_old_params_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "rails_helper"
+# just a comment
```
defong, line 2:
```suggestion

```

app/services/find/match_old_params.rb
goodviber: I think that is correct?

app/views/layouts/find_layout.html.erb
goodviber: Done

app/services/find/match_old_params.rb
defong: I not sure what you are intending....

image

goodviber: x = Hash.new
x[:qualifications] = "awesome"
x[:qualification] = x.delete(:qualifications)

3.1.0 :006 > x
 => {:qualification=>"awesome"}

goodviber: So in the test 
```
context "with QUALIFICATION_FILTERS" do
      subject {
        described_class.call({
          "qualifications" => %w[Other PgdePgceWithQts QtsOnly],
        })
      }

      it "maps the old find params" do
        expect(subject).to eq({
          "qualification" => ["pgce pgde", "pgce_with_qts", "qts"],
```

app/views/find/result_filters/_hidden_fields.html.erb
goodviber: Done

spec/services/find/match_old_params_spec.rb
goodviber: Done

app/services/find/match_old_params.rb
goodviber: Its 'qualificationS' mapped to 'qualification'

defong: :woozy_face:

defong (approved):

PR #3273, opened 2023-01-23
[944] Set up different DFE namespace for Find

darokel:
### Context

With Find being merged into Publish, we're going to stream events into one table but setup a different namespace for Find. 

https://trello.com/c/6nkOr3CU/944-handle-dfe-analytics-configuration

- Sets up a namespace for Find
- Tweaks the DFE config to send jobs async via diff. queue 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/find/application_controller.rb
+++ b/app/controllers/find/application_controller.rb
@@ -1,5 +1,7 @@
 module Find
   class ApplicationController < ActionController::Base
+    include DfE::Analytics::Requests
+
     layout "find_layout"
     default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
@@ -8,6 +10,16 @@ module Find
 
     def render_feedback_component
       @render_feedback_component = true
+    end
+
+    # For DfE::Analytics as Find doesn't have a current_user
+    def current_user
+      nil
+    end
+
+    # DFE Analytics namespace
+    def current_namespace
+      "find"
     end
 
   private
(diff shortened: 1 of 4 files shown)


darokel:
[Handle DFE analytics configuration](https://trello.com/c/6nkOr3CU/944-handle-dfe-analytics-configuration)

app/controllers/find/application_controller.rb
```diff
@@ -10,6 +12,16 @@ def render_feedback_component
       @render_feedback_component = true
     end
 
+    # For DfE::Analytics as Find doesn't have a current_user
+    def current_user
```
duncanjbrown, line 16: You can leave this out as `dfe-analytics` will check whether the method exists before calling it:

https://github.com/DFE-Digital/dfe-analytics/blob/b916b441d1a5579465c52d0772d99dab8e072d73/lib/dfe/analytics/requests.rb#L21

duncanjbrown (commented):
Looking good! Do you need to include `DfE::Analytics::Requests` in the Publish and Support namespaces too?

darokel:
> Looking good! Do you need to include `DfE::Analytics::Requests` in the Publish and Support namespaces too?

This is already included in the `ApplicationController` specific to publish/support which they both inherit from https://github.com/DFE-Digital/publish-teacher-training/blob/a30b6a691cf7061a738859f0b0734f860323331b/app/controllers/application_controller.rb#L2

duncanjbrown (approved):
LGTM!

PR #3276, opened 2023-01-23
[xxx] Enable the sidekiq dashboard for admins

darokel:
### Context

Publish has had sidekiq setup for a while but we have never enabled the web interface to be available. This PR does that so we can diagnose jobs. Using the same constraint from Register.

### DIFF
--- a/app/lib/system_admin_constraint.rb
+++ b/app/lib/system_admin_constraint.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class SystemAdminConstraint
+  def matches?(request)
+    system_admin?(request)
+  end
+
+  def system_admin?(request)
+    signin_user = UserSession.load_from_session(request.session)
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -12,6 +15,8 @@
     match "/(*path)" => redirect { |_, req| "#{Settings.base_url}#{req.fullpath}" }, via: %i[get post put]
   end
 
+  mount Sidekiq::Web, at: "/sidekiq", constraints: SystemAdminConstraint.new
```
defong, line 18: This should sit in the `publish` route section right?

> Publish has had sidekiq setup for a while but we have never enabled the web interface to be available. This PR does that so we can diagnose jobs. Using the same constraint from Register.

darokel: agreed, moved

defong (approved):

PR #3127, opened 2022-11-14
Don't capitalise subjects unless the subject is a language

avinhurry:
### Context

In the entry requirements section, subjects which are also not a language should not be capitalised. 

### Screenshot

image

### Changes proposed in this pull request

- Add the subject codes which are a language to an array.
- Use this array to downcase all other subject names in the view.
- Ensure modern languages is not capitalised (using gsub).

### Guidance to review

- View the entry requirements on courses with the following criteria, and ensure the capitalisation is correct. 

`Note: You may have to insert the course in to the instance variable in `app/controllers/find/courses_controller.rb` and the provider on the view as the full search functionality of new Find hasn't been built yet.`
- A course with a single language subject
- A course with a single non language subject
- A course with two subjects, one language and one non language
- A modern languages course
- A modern languages other course

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally

PR #3303, opened 2023-01-30
Downcase non language subjects

goodviber:
### Context
When creating course names, the second subject should be lowercase unless it is a noun.

https://publish-courses-prototype.herokuapp.com/course-names
What needs to be done

Update the course name generator to be sentence case unless the subject is a noun (e.g., English, French, Spanish, etc)
Add downcasing method
### Guidance to review
Covered by existing tests I think
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/generate_course_name_service.rb
+++ b/app/services/courses/generate_course_name_service.rb
@@ -49,10 +49,10 @@ module Courses
       "#{formatted_subjects.first} with #{downcase_if_not_language(formatted_subjects.last)}"
     end
 
-    def downcase_if_not_language(str)
-      return str if LANGUAGES.any? { |s| str.match s }
+    def downcase_if_not_language(subject_name)
+      return subject_name if LANGUAGES.any? { |s| subject_name.match s }
 
-      str.downcase
+      subject_name.downcase
     end
 
     def formatted_subjects
--- a/app/services/courses/generate_course_name_service.rb
+++ b/app/services/courses/generate_course_name_service.rb
@@ -4,6 +4,20 @@ module Courses
   class GenerateCourseNameService
     include ServicePattern
 
+    LANGUAGES = %w[
+      English
+      German
+      Italian
+      Japanese
+      Mandarin
+      Russian
+      Spanish
+      French
+      Ancient Greek
+      Ancient Hebrew
+      Latin
+    ].freeze
+
     ENGINEERS_TEACH_PHYSICS_TITLE = I18n.t('courses.generate_course_name_service.etp_title')
     FURTHER_EDUCATION_TITLE = I18n.t('courses.generate_course_name_service.further_education_title')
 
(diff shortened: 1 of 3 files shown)


goodviber:
[Course name noun list](https://trello.com/c/Jh0HVSsj/894-course-name-noun-list)

app/services/courses/generate_course_name_service.rb
```diff
@@ -4,6 +4,20 @@ module Courses
   class GenerateCourseNameService
     include ServicePattern
 
+    LANGUAGES = %w[
```
avinhurry, line 7: Can we use this? 

https://github.com/DFE-Digital/publish-teacher-training/blob/ce9d02f49547aed3bf15fd4b8868cddc932a61a9/app/decorators/course_decorator.rb#L8

app/services/courses/generate_course_name_service.rb
```diff
@@ -32,7 +46,13 @@ def generated_title
 
       return formatted_subjects.first if formatted_subjects.length == 1
 
-      "#{formatted_subjects.first} with #{formatted_subjects.last}"
+      "#{formatted_subjects.first} with #{downcase_if_not_language(formatted_subjects.last)}"
+    end
+
+    def downcase_if_not_language(str)
```
avinhurry, line 52:
```suggestion
    def downcase_if_not_language(subject_name)
```

app/services/courses/generate_course_name_service.rb
goodviber: There are only 10 subjects in LANGUAGE_SUBJECT_CODES and 11 in LANGUAGES?

avinhurry: `LANGUAGE_SUBJECT_CODES` is missing ancient Greek and ancient Hebrew. We haven't had any of those courses for years so I think its OK to leave out? Otherwise we'll have to add more logic because I think `ancient` should be lowercase and `Greek` uppercase. 

I think you're also missing `English as a second or other language` in your list. 

Either way the only thing that uses that `LANGUAGE_SUBJECT_CODES` array is to downcase non language subjects on Find, so I think the lists should be the same. See: https://github.com/DFE-Digital/publish-teacher-training/pull/3127/files

goodviber: I think we have
```
def format_subject_name(subject)
      {
        'communication and media studies' => 'Media studies',
        'english as a second or other language' => 'English',
        'physics' => course.is_engineers_teach_physics? ? ENGINEERS_TEACH_PHYSICS_TITLE : subject.to_s
      }[subject.subject_name.downcase] || subject.to_s
    end
```

avinhurry: @goodviber had a chat with Simon and the two lists should be the same, so I think to keep things DRY we can stick to the old list and add the ancient subjects to it. `A1` and `A2`. 

Ignore what I said about `ancient greek` - it should be `Ancient Greek`

goodviber: and @simonwhatley list
```
- Ancient Greek
- Ancient Hebrew
- English
- French
- German
- Italian
- Japanese
- Latin
- Mandarin
- Russian
- Spanish

avinhurry (approved):

PR #3304, opened 2023-01-30
[0915] Refactor the page helper page object so it can build up namespaced pages

defong:
### Context
Build up namespaced pages

Refactor the page helper page object so it can build up namespaced pages
Only include page objects for the related items, previous everything was included
Moved/rename files so they are more aligned with the usages
### Guidance to review
Best to follow per commits

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


defong:
[Refactor the page helper page object so it can build up namespaced pages](https://trello.com/c/TunTZAVw/915-refactor-the-page-helper-page-object-so-it-can-build-up-namespaced-pages)

darokel (approved):
Looks good, nice work. I noticed the code to generate the method names is repeated for each page object type. I wonder if we could just do it in one file and give it an array of all the module names?

avinhurry (approved):

defong:
> Looks good, nice work. I noticed the code to generate the method names is repeated for each page object type. I wonder if we could just do it in one file and give it an array of all the module names?

I broke it to only load when a particular feature `auth|find|publish|support` spec is loaded rather then all at once 

There seems to be a conflict with https://github.com/DFE-Digital/teacher-training-api/blob/915-refactor-the-page-helper-page-object-so-it-can-build-up-namespaced-pages/spec/spec_helper.rb#L28 vs the design of `module` so left it as such as it was getting hard to understand the the method was created in the context of `rspec`

PR #3344, opened 2023-02-07
[1033] Amended to only change name when valid

defong:
### Context
Course name and subjects and persisted status

Persisted course name can only be updated if it is valid
Non persisted course can be updated regardless

### Guidance to review
It works for both new course and existing courses

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/assign_subjects_service.rb
+++ b/app/services/courses/assign_subjects_service.rb
@@ -16,11 +16,12 @@ module Courses
 
       update_subjects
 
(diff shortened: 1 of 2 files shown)


defong:
[Fix course saving before validation](https://trello.com/c/QP81kWaE/1033-fix-course-saving-before-validation)

darokel (approved):

PR #3363, opened 2023-02-16
Add synonyms column to provider table and populate from CSV

avinhurry:
### Context

Migration to add synonyms to the provider table and populate the column with data from a CSV.

- New column type is array
- Default type is empty array
- Send data to BigQuery
- Create a rake task to populate the column with data from the CSV 

### Guidance to review

🤷🏼 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


avinhurry:
[Import the provider synonyms and abbreviations CSV](https://trello.com/c/JAduCBhC/989-import-the-provider-synonyms-and-abbreviations-csv)

avinhurry:
@darokel this is the [first card](https://trello.com/c/JAduCBhC/989-import-the-provider-synonyms-and-abbreviations-csv) done. 

Do you want the second card contained within this PR and shipped as a complete working feature, or shall we merge this and work on the [second card](https://trello.com/c/lydmSeh6/990-update-provider-autocomplete-to-use-synonyms-and-abbreviations) in a separate PR. I'm not sure which approach is best here.

db/data/20230217112836_add_synonyms_data_to_provider.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AddSynonymsDataToProvider < ActiveRecord::Migration[7.0]
+  def up
+    csv_filepath = Rails.root.join(__FILE__.gsub('.rb', '.csv'))
```
darokel, line 5: I think this should be a rake task and then we have a migration to call it? Are we handling the import rules? Eg stop words etc? Re: the rake task, we may need to re-import this list if we ever want to add additional synonyms

darokel:
@avinhurry Happy for this to be its own card/PR I think

db/migrate/20230216112522_add_synonyms_to_provider.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddSynonymsToProvider < ActiveRecord::Migration[7.0]
+  def change
+    add_column :provider, :synonyms, :string, array: true, default: []
```
darokel, line 5: Do we need an index for this?

db/data/20230217112836_add_synonyms_data_to_provider.rb
avinhurry: Done [here](https://github.com/DFE-Digital/publish-teacher-training/pull/3363/commits/c5ca66cf368e8e0c25543862285ac5404abadb92). We can run this in prod when we merge.

db/schema.rb
```diff
@@ -257,6 +257,7 @@
     t.string "urn"
     t.boolean "can_sponsor_skilled_worker_visa"
     t.boolean "can_sponsor_student_visa"
+    t.string "synonyms", default: [], array: true
```
darokel, line 260: Could we add an index for this? I know we're not using it for search at the moment but just in case. `add_index :provider, : synonyms, using: 'gin'

defong: Discussed in the realm of real life, it not likely we will be using that as a lookup hence the need for indexes are not needed for the moment

defong: Sure lets add the index, but we need to be aware that this is an array and in future need to tailor queries appropriately.

darokel (approved):

db/schema.rb
avinhurry: I've added it now [here](https://github.com/DFE-Digital/publish-teacher-training/pull/3363/commits/509f76f6278a84095fbab2389ee13874de6c77d9) but just thinking, can't we add it if we decide we need it later on to avoid confusion and over engineering?

db/migrate/20230216112522_add_synonyms_to_provider.rb
avinhurry: Discussion below.

db/schema.rb
darokel: It's just one line to avoid having to remember later and potentially some headaches

PR #3370, opened 2023-02-20
[926] Add content about relocation payments to international candidates section

darokel:
### Context

https://trello.com/c/Tu6wycNm/926-add-content-about-relocation-payments-to-international-candidates-section

- Update the content for the international candidates component used via the Find course show page and the Publish course preview
- Also updates some existing content in this component to match with the prototype

### Guidance to review

Review app - https://find-pr-3370.london.cloudapps.digital

Check one of the qualifying courses for a relocation payment and one without. Assert the relocation text is shown correctly.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/find/courses/international_students_component/view.html.erb
+++ b/app/components/find/courses/international_students_component/view.html.erb
@@ -14,7 +14,7 @@
 
     <ul class="govuk-list govuk-list--bullet">
       <li>are an Irish citizen</li>
-      <li>have settled or pre-settled status under the <%= govuk_link_to("EU Settlement Scheme", "https://www.gov.uk/settled-status-eu-citizens-families") %> (you may still be able to apply)</li>
+      <li>have settled or pre-settled status under the <%= govuk_link_to("EU Settlement Scheme", t("links.eu_settlement_scheme")) %> (you may still be able to apply)</li>
       <li>have indefinite leave to remain in the UK</li>
     </ul>
 
@@ -24,7 +24,7 @@
       <%= t("find.international_candidates.entitlement.html") %>
     <% end %>
 
-    <p class="govuk-body">Learn more about <%= govuk_link_to("training to teach in England as an international student", "https://getintoteaching.education.gov.uk/non-uk-teachers/train-to-teach-in-england-as-an-international-student") %>.</p>
+    <p class="govuk-body">Learn more about <%= govuk_link_to("training to teach in England as an international student", t("find.get_into_teaching.url_train_to_teach_as_international_candidate")) %>.</p>
 
     <h3 class="govuk-heading-m">International qualifications</h3>
     <p class="govuk-body">If your qualifications come from a non-UK institution, we may want to see a ‘statement of comparability’ from UK ENIC. This will show us how they compare to UK qualifications.</p>
--- a/app/components/find/courses/international_students_component/view.html.erb
+++ b/app/components/find/courses/international_students_component/view.html.erb
@@ -20,9 +20,11 @@
 
     <%= t("find.international_candidates.#{visa_type}.#{sponsorship_availability}.html") %>
 
-    <p class="govuk-body">Alternatively, check if you’re eligible for a different type of visa, such as a Graduate or Family visa, that allows you to train to be a teacher without being sponsored.</p>
+    <% if course_has_relocation_entitlement? %>
+      <%= t("find.international_candidates.entitlement.html") %>
+    <% end %>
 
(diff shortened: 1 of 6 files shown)


darokel:
[Add content about relocation payments to international candidates section](https://trello.com/c/Tu6wycNm/926-add-content-about-relocation-payments-to-international-candidates-section)

Kateharries (approved):
I've fixed some typos (extra spaces)

app/components/find/courses/international_students_component/view.html.erb
```diff
@@ -20,9 +20,11 @@
 
     <%= t("find.international_candidates.#{visa_type}.#{sponsorship_availability}.html") %>
 
-    <p class="govuk-body">Alternatively, check if you’re eligible for a different type of visa, such as a Graduate or Family visa, that allows you to train to be a teacher without being sponsored.</p>
+    <% if course_has_relocation_entitlement? %>
+      <%= t("find.international_candidates.entitlement.html") %>
+    <% end %>
 
-    <p class="govuk-body">Find out more about <%= govuk_link_to("training to teach in England", "https://www.gov.uk/government/publications/train-to-teach-in-england-non-uk-applicants/train-to-teach-in-england-if-youre-a-non-uk-citizen") %>.</p>
+    <p class="govuk-body">Learn more about <%= govuk_link_to("training to teach in England as an international student", "https://getintoteaching.education.gov.uk/non-uk-teachers/train-to-teach-in-england-as-an-international-student") %>.</p>
```
avinhurry, line 27: Can we keep the links altogether in here? 

https://github.com/DFE-Digital/publish-teacher-training/blob/ce50d4876c7190b333541655fe4444b6714b993f/config/locales/find.yml#L96

avinhurry (commented):
Looks good, just wondering if we generate the text regardless of whether the subject is dominant or not?

darokel:
> Looks good, just wondering if we generate the text regardless of whether the subject is dominant or not?

Yup

avinhurry:
> > Looks good, just wondering if we generate the text regardless of whether the subject is dominant or not?
> 
> Yup

Think @Kateharries is double checking with the international team

avinhurry:
@darokel yeah we only want to render the text if its the first subject, not the second subject, something along these lines: https://github.com/DFE-Digital/publish-teacher-training/blob/ce50d4876c7190b333541655fe4444b6714b993f/app/components/find/courses/entry_requirements_component/view.rb#L30

app/components/find/courses/international_students_component/view.rb
```diff
@@ -28,6 +43,14 @@ def visa_type
         def sponsorship_availability
           @sponsorship_availability ||= course.public_send("can_sponsor_#{visa_type}") ? :available : :not_available
         end
+
+        def course_has_relocation_entitlement?
+          SUBJECTS_WITH_RELOCATION_ENTITLEMENT.include?(course_subject_codes.first)
```
avinhurry, line 48: I think there are some annoying bugs here: 

1. If the first subject is Physics which has relocation entitlement, and the second subject is Biology which doesn't that works fine, but when you change the first subject to another relocation entitlement course, such as Latin (I think) it shifts the second subject (Biology) which doesn't have relocation entitlement to the first position in the array which means the text does not generate 🤯 

2. If you have modern languages (French) as the first subject, which has relocation entitlement the content renders correctly, however if you then add Biology as the second subject (which does not have relocation entitlement) the content does not render. It looks like it pushes French to the bottom of the array, maybe because it's `type: "ModernLanguagesSubject"` 🤷🏼‍♂️ 

Might be worth adding a few tests for these?

avinhurry (changes requested):
Looks like there are a few annoying bugs here, comments in line

app/components/find/courses/international_students_component/view.rb
darokel: This is a different underlying issue I raised before in our chats about the subject ordering. I encountered it when working on generating titles. I won't tackle that here because it requires a little more investigation. i think i may have already carded it up but if not will do

avinhurry: Is it OK for candidates to see that content when they may not be entitled to the financial help? I guess the content says "May"

darokel: Technically they wouldn't (which in terms of the api is correct because we want to rely on the first subject). This only goes funny when Modern Language is paired with a secondary subject (where the secondary subject gets swapped into the first slot). I've (reluctantly) handled it in the component for now but we need to solve the root problem otherwise we'll keep applying bandages everywhere. I count 3 different places where we're more or less doing the same thing to handle the issue.

avinhurry (approved):

victoriawinlow (commented):
All good, thanks

PR #3379, opened 2023-02-22
remove fe level change link

goodviber:
### Context
When creating a further education course, the check your answers change link leads back to the course level page, an error is raised when filling in the form with Primary or Secondary options.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/views/publish/courses/confirmation.html.erb
```diff
@@ -31,10 +31,14 @@
...
-            visually_hidden_text: "level"
-          ) %>
+          <% if course.is_further_education? %>
+            <% row.action %>
+          <% else %>
+            <% row.action(
+              href: new_publish_provider_recruitment_cycle_courses_level_path(course.provider.provider_code, course.recruitment_cycle.year, params.to_unsafe_h.merge(goto_confirmation: true)),
+              visually_hidden_text: "level"
+            ) %>
+          <% end %>
```
defong, line 41: I think its a bad idea to allow for change link at any level

Also can we have tests

defong (changes requested):

app/views/publish/courses/confirmation.html.erb
goodviber: To confirm, you think we should remove the change link for all levels of course, FE, Primary, Secondary ?

defong: Yes we should remove the change link for at levels.

There is not enough work done to support switching from level to level as there will be values passed over that is not relevant for the level specified.

defong (approved):

app/views/publish/courses/confirmation.html.erb
darokel: Could we also remove the change link for the SEND on this page? It'll essentially allow changing the level too

goodviber: Just saw this, will do

darokel (approved):

PR #3400, opened 2023-03-06
[1107] Added provider store

defong:
### Context
Provider store

Added provider store

### Guidance to review
Part of the arc for locations

This is the store that will provide temporary storage via redis.

The store will be tied to a provider instead of the user, as the data relates to the `provider` and not the `user`

Boiler plate code

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


spec/services/stores/shared_examples/store.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    form_store_keys.each do |store_key|
+      context "when form_store_key is #{store_key}" do
+        let(:form_store_key) { store_key }
+
+        let(:value) { nil }
+
+        if described_class == Stores::UserStore
+          # TECH DEBT: [Stores::UserStore] This is error prone
+          let(:redis_key) { "#{identifier_model.id}_#{form_store_key}" }
```
defong, line 36: it gives

`1_user`

Whereby 
- `1` is an id number and not related to either an `user` or a `course` or a `provider`
- `user` is a form key not as a identifier

defong:
[Bulk Uploads - CSV Flow](https://trello.com/c/DgfeqeZ3/1107-bulk-uploads-csv-flow)

avinhurry (approved):

spec/services/stores/shared_examples/store.rb
defong: https://trello.com/c/DCiFvwcr/398-bug-user-store-identifier-id-is-not-unique

app/services/stores/user_store.rb
```diff
@@ -13,6 +13,7 @@ def store_keys
     private
 
     def identifier_id
+      # TECH DEBT: [Stores::UserStore] This is error prone
       identifier_model.id
```
darokel, line 17: Just trying to understand the issue a little more but why isn't this following the same format as the others?

defong: I think it was leftovers.

PR #3405, opened 2023-03-08
1117 gias import rake task

goodviber:
### Context
User need

As a user, I want to easily add school placements, so that candidates can understand where they may be placed
Why

We have used the term locations to refer to school placements and it’s become confusing. We’re updating the use of the term locations to schools as well as improve the flow to add new schools.

This ticket is to download a copy of the GIAS data and create a way to import (and re-import) the data into the Publish database
What needs to be done

We need to write a script which is able to import (and re-import) a dump of the GIAS data into a new table in the Publish database. See the dev checklist below.
Done when

We have a copy of the GIAS data dump in the repository
We have a way of importing (and re-importing) the data into the database

Rake task which calls CSVImports::GiasImport service.
### Guidance to review
There are a variety of rows in the fixtures csv file, I put them here due to size of the rows. Only 2 schools should be valid for upsert.

We only need to include schools that are open (status: 1, 3)

We can exclude all other columns that are not relevant to our data.

We can exclude:

independent schools
Welsh schools
nurseries (phase: 1)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


db/migrate/20230307132542_create_gias_school.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class CreateGiasSchool < ActiveRecord::Migration[7.0]
+  def change
+    create_table :gias_school do |t|
+      t.text 'urn', index: true, null: false
```
darokel, line 6: Should we enforce a unique index?

lib/tasks/gias_update.rake
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+desc 'Upsert GIAS data into GiasSchool'
+task :gias_update, [:csv_path] => [:environment] do |_, args|
+  csv_path = args.csv_path
```
darokel, line 7: I think we should default to a path here if you don't provide it

app/services/csv_imports/gias_import.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module CSVImports
+  class GiasImport
```
darokel, line 6: Can we use the `ServicePattern` here? https://github.com/DFE-Digital/publish-teacher-training/blob/c13e26f96a7aaccd9ceec1305c6aea966a6146f6/app/services/service_pattern.rb#L1  as we're moving to using `.call`

goodviber: Yep

db/migrate/20230307132542_create_gias_school.rb
goodviber: Good idea

spec/models/gias_school_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  it 'is invalid without town' do
+    subject.town = ''
+    expect(subject).not_to be_valid
+  end
+
+  it 'is invalid without postcode' do
+    subject.postcode = ''
+    expect(subject).not_to be_valid
+  end
```
defong, line 43: Can we test it this way round. as that is very fragile

spec/models/gias_school_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+describe GiasSchool do
+  subject do
+    described_class.new(
+      urn: '100000',
+      name: 'school name',
+      address1: 'the address',
+      town: 'anytown',
+      postcode: 'postcode'
+    )
```
defong, line 13: Can we use factories and fakers to generate these

spec/models/gias_school_spec.rb
goodviber: Sure

spec/models/gias_school_spec.rb
goodviber: Sure, although I am not certain we get any value from faking the attributes

darokel (approved):

PR #3408, opened 2023-03-09
[1133] Fix secondary subject not displaying

darokel:
### Context

When creating a new course and on the check your answers page, if you have selected two subjects and attempt to update them - you are taken back to the subjects form but the second subject is lost.

https://trello.com/c/EpcaUDU6/1133-bug-fix-subject-editing-in-new-course-flow

- Fix call to get subject ids. For some reason `pluck` returns an empty array here (even when there are two records present). 

Screenshot 2023-03-09 at 13 49 04

Given this will only be two subjects, switch to map to get the ids

- Have also moved the caddy process into the Procfile so it can all be fired up in one command

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -2,3 +2,4 @@ js: yarn build --watch
 css: yarn build:css --watch
 web: bin/rails server -p 3001
 worker: bundle exec sidekiq -t 25 -C config/sidekiq.yml
+caddy: caddy run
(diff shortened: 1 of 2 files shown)


darokel:
[[Bug] Fix subject editing in new course flow](https://trello.com/c/EpcaUDU6/1133-bug-fix-subject-editing-in-new-course-flow)

defong (approved):

PR #3411, opened 2023-03-10
docs for gias_import

goodviber:
### Context
We need to add support documention for the gias_import task and make logs messages visible in the console
Addition to guides/setup-development.md and log to $stdout
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/services/csv_imports/gias_import.rb
+++ b/app/services/csv_imports/gias_import.rb
@@ -46,6 +46,7 @@ module CSVImports
         end
         rows += 1
       end
+      Rails.logger = Logger.new($stdout)
       Rails.logger.info "Done! #{upserted} schools upserted"
       Rails.logger.info "Failures #{rows - upserted}"
       Rails.logger.info "Errors - #{errors.inspect}" if errors.any?
--- a/guides/setup-development.md
+++ b/guides/setup-development.md
@@ -4,7 +4,6 @@ Clone the repo:
 
     git clone git@github.com:DFE-Digital/publish-teacher-training.git
 
-
 ## Setup the application
 
 Run the following commands:
@@ -16,7 +15,6 @@ bundle exec rails db:setup
 ```
 
 ## Start the server
-
 
 You can use [Foreman](https://github.com/ddollar/foreman) or [Overmind](https://github.com/DarthSim/overmind) to run all the processes needed for local dev. Once you have either of the two, you can run:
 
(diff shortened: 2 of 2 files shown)


app/services/csv_imports/gias_import.rb
```diff
@@ -46,6 +46,7 @@ def call
         end
         rows += 1
       end
+      Rails.logger = Logger.new($stdout)
```
darokel, line 49: Not sure if this is a good idea to override the `Rails.logger` in the service. Wouldn't it be better if the service defined it's own instance eg `@logger = Logger.new($stdout)` in the initialiser and then use that to log?

guides/setup-development.md
```diff
@@ -95,6 +93,30 @@ Then run the following command to populate the database:
 psql manage_courses_backend_development < ~/Downloads/backup_sanitised.sql
 ```
 
+## Seeding GiasSchool Data
+
+You can obtain an updated csv from https://www.get-information-schools.service.gov.uk/Downloads the `establishment fields CSV` checkbox. This can be uploaded to the `csv` directory.
```
darokel, line 98: I wonder if we should split this text for running it with the current file (in which case you don't need to provide a csv path as it'll default) and then what to do if you need to use a different version?

darokel (commented):
Left a couple of bits

app/services/csv_imports/gias_import.rb
goodviber: Will give it a go

guides/setup-development.md
goodviber: If we upload a new file, would we also want to update the default path to point at the new file, or leave the default pointing at the original?

darokel: good point, we'd just update the path to point to the new file and swap it in. Then when running this in all other cases, you could just leave it to the defaults

app/services/csv_imports/gias_import.rb
```diff
@@ -46,9 +46,10 @@ def call
         end
         rows += 1
       end
-      Rails.logger.info "Done! #{upserted} schools upserted"
-      Rails.logger.info "Failures #{rows - upserted}"
-      Rails.logger.info "Errors - #{errors.inspect}" if errors.any?
+      logger = Logger.new($stdout)
```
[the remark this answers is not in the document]
goodviber: Done

PR #3418, opened 2023-03-14
[1108] Amended govuk components and related dependency

defong:
### Context

Upgrading govuk components to 4.0.0a2 as the alpha version has code that we can use to build our summary card. It's due for release soon, so we may as well get ahead of the curve. 

- Upgrade govuk components to 4.0.0a2
- Fix breaking changes 
- Fix breaking tests

### Guidance to review

- Have a play around with the app in areas where components are heavily used and check that everything still works as expected. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,7 @@ gem 'active_model_serializers'
 gem 'kaminari'
 
 # Pagination for API
-gem 'pagy', '~> 5.10'
+gem 'pagy', '~> 6.0'
 
 # JSON:API Ruby Client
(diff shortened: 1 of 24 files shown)


avinhurry:
[Upgrade govuk components gem to alpha](https://trello.com/c/DcJbd1Ka/1143-upgrade-govuk-components-gem-to-alpha)

darokel (approved):

PR #3431, opened 2023-03-18
1105 Bulk Uploads - Update current add location form

goodviber:
### Context
As a user, I would like an easy way to add lots of location data, so candidates can be better informed about where they may take their courses.

We’re adding bulk upload support via the support console to achieve the criteria above.

This ticket is to update the existing location form.
What needs to be done

Update the current location form in support to match the one in the prototype
Add a confirm page following adding the location so the provider can confirm their details
On the confirm page, cancel undoes the creation of the location. Add location saves it and the Save location and add another saves and takes them to the same form

Success

The user should be able to enter the new location details, confirm their changes and successfully save it to the provider.
Tech notes

We already have a pattern to support confirm page checks and undoing-ness. See the user stuff

Out of scope

Changing the design of the edit view for the location?

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


goodviber:
[Bulk Uploads - Update current add location form](https://trello.com/c/6ApH6J5H/1105-bulk-uploads-update-current-add-location-form)

app/controllers/support/providers/locations_check_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+          end
+          flash[:success] = t('support.providers.locations.added')
+        else
+          render template: 'support/locations/new'
+        end
+      end
+
+      private
+
+      def build_site_and_form
```
defong, line 25: is there a real need to build the site?

app/forms/support/location_form.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    validates :location_name,
+              :address1,
+              :address3,
+              :postcode,
+              presence: true
+    validates :postcode, postcode: true
+    validates :urn, reference_number_format: { allow_blank: true, minimum: 5, maximum: 6, message: I18n.t('activemodel.errors.models.support/location_form.attributes.urn.format') }
+    validate :site_name_is_unique
+    # should the URN also be unique?
```
defong, line 26: in which sense?

app/views/support/locations/new.html.erb
```diff
@@ -1,10 +1,13 @@
 <%= render PageTitle.new(title: "support.providers.locations.new") %>
 
-<h1 class="govuk-heading-l">Add a location</h1>
+<h1 class="govuk-heading-l">
+  <span class="govuk-caption-l"><%= "#{@provider.provider_name} (#{@provider.provider_code})" %></span>
+    Add location
```
defong, line 5: image

app/controllers/support/locations_controller.rb
```diff
@@ -54,6 +54,20 @@ def flash_resource
...
+        :address1,
+        :address2,
+        :address3,
+        :address4,
+        :postcode
+      )
+    end
+
+    # TODO: remove this when we align the edit form
+    def edit_site_params
```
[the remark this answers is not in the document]
goodviber: Nice

app/controllers/support/providers/locations_check_controller.rb
goodviber: I think so, we need to pass it to the form?

app/forms/support/location_form.rb
goodviber: Just a thought, if it stands for Unique Reference Number, should we validate uniqueness?

app/views/support/providers/locations_check/show.html.erb
```diff
@@ -0,0 +1,39 @@
+<%= render PageTitle.new(title: t("support.providers.locations.check")) %>
```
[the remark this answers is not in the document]
goodviber: oops

app/forms/support/location_form.rb
defong: No as we don't control the uniqueness.

app/controllers/support/providers/locations_check_controller.rb
defong:
```
@restores_from_store = location_form = LocationForm.new(provider)
@overwrite_from_store = LocationForm.new(provider, site)
```

app/forms/support/location_form.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      return '' if address.all?(&:blank?)
+
+      address.select(&:present?).join('<br>').html_safe
+    end
+
+    private
+
+    def site_name_is_unique
+      return unless Site.exists?(location_name:)
```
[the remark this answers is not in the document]
defong: Yes it that equivalent. But not this [one](https://github.com/DFE-Digital/publish-teacher-training/blob/f1aee75bd29f3c2710c104b91eb05ae705f7eb2f/app/models/site.rb#L23)

app/forms/support/location_form.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      errors.add(:location_name, 'Name is taken')
```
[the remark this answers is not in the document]
defong: you the best :muscle:

app/views/support/locations/new.html.erb
goodviber: Double oops, there was no back link on the prototype

app/views/support/locations/_form.html.erb
```diff
@@ -1,11 +1,10 @@
-<%= form_with model: [:support, site], local: true, url: form_url, class: "location-form" do |f| %>
-
+<%= form_with model: @location_form, local: true, url: form_url, class: "location-form" do |f| %>
   <%= f.govuk_error_summary %>
 
   <%= f.govuk_text_field :location_name, label: { text: "Name", size: "m" } %>
```
darokel, line 4: Looking at the prototype, i think all of these labels now need to be a size `s`. Eg:

```
<%= f.govuk_text_field :location_name, label: { text: "Name", size: "s" } %>
```

Prototype:

Screenshot 2023-03-22 at 09 49 53

The build:

Screenshot 2023-03-22 at 09 51 03

app/views/support/locations/new.html.erb
goodviber: I think that might be by design, as the back link could go to either the index page or the check answers page, where it clears the location_details

app/forms/support/location_form.rb
goodviber: Coolio

app/forms/support/location_form.rb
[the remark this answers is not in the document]
goodviber: I think this works ```return unless @identifier_model.sites.exists?(location_name:)```

app/views/support/locations/_form.html.erb
goodviber: Done

app/controllers/support/providers/locations_check_controller.rb
goodviber: LocationForm.new(provider) raises an Argument error?

defong: Oh right, sorry was comparing it to `Publish::CourseFundingForm`

app/views/support/locations/_form.html.erb
darokel: Just looking at the new review app and it's still not quite there. 

- URN hint text needs to be below the input
- Create button needs to say `Continue`
- The address labels are also bold on the prototype?

app/views/support/locations/_form.html.erb
```diff
@@ -1,11 +1,10 @@
...
-
+<%= form_with model: @location_form, local: true, url: form_url, class: "location-form" do |f| %>
   <%= f.govuk_error_summary %>
 
-  <%= f.govuk_text_field :location_name, label: { text: "Name", size: "m" } %>
+  <%= f.govuk_text_field :location_name, label: { text: "Name", size: "s" } %>
 
   <%= f.govuk_text_field(:urn,
-    label: { text: "Unique Reference Number (URN)", size: "m" },
+    label: { text: "Unique Reference Number (URN) (optional)", size: "s" },
```
avinhurry, line 7: I think the URN should have a width of 10 too, similar to [this](https://github.com/DFE-Digital/publish-teacher-training/blob/f1aee75bd29f3c2710c104b91eb05ae705f7eb2f/app/views/support/providers/locations/new_multiple/show.html.erb#L30-L32) (although it doesn't currently in the proto).

image

app/views/support/locations/_edit_form.html.erb
```diff
@@ -0,0 +1,31 @@
+<%= form_with model: [:support, site], local: true, url: form_url, class: "location-form" do |f| %>
```
darokel, line 1: Just realised, how comes we have two location forms? Can't the new and edit share the same one?

app/views/support/locations/_form.html.erb
goodviber: Will go with the proto for now, see what Simon says?

PR #3450, opened 2023-03-24
1155 edit/delete location flow

goodviber:
### Context
We need to update the existing location edit flow to match the one in the prototype. This is when you click edit for a single location from the location index view within a single provider.

We need to update the existing location delete flow to match the one in the prototype. This is when you click the delete link within a location edit view for a single provider.

LocationForm object, validations promoted from Site. Combine new and edit form partial, follow prototype flows for edit and delete.
### Guidance to review
As support user, edit and delete a location/school
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


goodviber:
[Bulk Uploads - Update current delete location flow](https://trello.com/c/oUDr5MWk/1156-bulk-uploads-update-current-delete-location-flow)

goodviber:
[Bulk Uploads - Update current edit location flow](https://trello.com/c/0KJN45zg/1155-bulk-uploads-update-current-edit-location-flow)

darokel:
A couple of UX quirks I noticed:

**On the show page for a school in this PR:**

Screenshot 2023-03-28 at 14 48 40

- Change link for code should not exist
- Page title should be {school} - {provider} - service

**On the edit view for this PR:**

Screenshot 2023-03-28 at 14 49 29

- Missing back link
- Provider and school should be heading not `Edit school details`
- Page title should be {school} - {provider} - service
- School code should be above URN and not editable
- URN field and hit space seems a little larger than prototype
- Address line 3 seems to be missing from the form
- Submit button should read `Update school`

goodviber:
> A couple of UX quirks I noticed:
> 
> **On the show page for a school in this PR:**
> 
> Screenshot 2023-03-28 at 14 48 40
> 
>     * Change link for code should not exist
> 
>     * Page title should be {school} - {provider} - service
> 
> 
> **On the edit view for this PR:**
> 
> Screenshot 2023-03-28 at 14 49 29
> 
>     * Missing back link
> 
>     * Provider and school should be heading not `Edit school details`
> 
>     * Page title should be {school} - {provider} - service
> 
>     * School code should be above URN and not editable
> 
>     * URN field and hit space seems a little larger than prototype
> 
>     * Address line 3 seems to be missing from the form
> 
>     * Submit button should read `Update school`

I think these are done now, I followed the pattern for the PageTitle from the users. The URN field is size small and width 10 ? Address line 3 is the town or city field I think.

The back button is a little tricky as when returning to the check your answers page we lose the session data, maybe we could track the referrer and do a conditional back path?

goodviber:
[Update the existing location form](https://trello.com/c/LTRF4vhG/1119-update-the-existing-location-form)

darokel:
> > A couple of UX quirks I noticed:
> > **On the show page for a school in this PR:**
> > Screenshot 2023-03-28 at 14 48 40
> > ```
> > * Change link for code should not exist
> > 
> > * Page title should be {school} - {provider} - service
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **On the edit view for this PR:**
> > Screenshot 2023-03-28 at 14 49 29
> > ```
> > * Missing back link
> > 
> > * Provider and school should be heading not `Edit school details`
> > 
> > * Page title should be {school} - {provider} - service
> > 
> > * School code should be above URN and not editable
> > 
> > * URN field and hit space seems a little larger than prototype
> > 
> > * Address line 3 seems to be missing from the form
> > 
> > * Submit button should read `Update school`
> > ```
> 
> I think these are done now, I followed the pattern for the PageTitle from the users. The URN field is size small and width 10 ? Address line 3 is the town or city field I think.
> 
> The back button is a little tricky as when returning to the check your answers page we lose the session data, maybe we could track the referrer and do a conditional back path?

Nice, two more things though:

- The school code field is still editable. In the prototype it's just rendered as a text
- Address line 3 was a new field added to support an extra line of address from GIAS data https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/models/site.rb#L30. Address line 4 is town or city i think. @defong can you confirm?

goodviber:
The School code input is read-only, but I can change it to text. A Site has 4 addresses and the postcode, so address 1, address 2, town, county, postcode matches I think?

app/views/support/providers/schools_check/show.html.erb
```diff
@@ -7,7 +7,7 @@
     <%= form_with(model: @school_form, url: support_recruitment_cycle_provider_check_school_path(@provider.recruitment_cycle_year, @provider), method: :put, local: true) do |f| %>
 
     <fieldset class="govuk-fieldset">
-      <legend class="govuk-fieldset__legend govuk-fieldset__legend--l"> <span class="govuk-caption-l"><%= t("support.providers.schools.new") %> - <%= @provider.provider_name %> </span>
+      <legend class="govuk-fieldset__legend govuk-fieldset__legend--l"> <span class="govuk-caption-l"><%= t("support.providers.schools.new") %> - <%= "#{@provider.provider_name} (#{@provider.provider_code})" %> </span>
```
darokel, line 10: I think there is a helper for this https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/decorators/provider_decorator.rb#L22

spec/forms/support/school_form_spec.rb
```diff
@@ -19,15 +19,79 @@
...
+    context 'with valid postcode' do
+      it 'is valid' do
+        params['postcode'] = 'tr11un'
+        expect(subject).to be_valid
+      end
+    end
+
+    context 'with invalid urns' do
+      it 'is invalid' do
+        params['urn'] = '123'
```
darokel, line 79: Wouldn't using the previous `is_expected.not_to allow_values` and the opposite be a bit more idiomatic for these assertions?

spec/support/page_objects/sections/location_form.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative 'base'
+
+module PageObjects
+  module Sections
+    class LocationForm < PageObjects::Sections::Base
```
darokel, line 7: I think this should be named `SchoolForm` now

darokel:
> The School code input is read-only, but I can change it to text? A Site has 4 addresses and the postcode, so address 1, address 2, town, county, postcode matches I think?

@simonwhatley The prototype still shows the form as Address Line 3, should it be Town or City? Which is the updated version?

app/views/support/providers/schools_check/show.html.erb
goodviber: Not sure it is in scope here ```>> provider.name_and_code
!! #<NameError: undefined local variable or method `provider' for #<ActionView::Base:0x00000000606558>
Did you mean?  provide
               @provider>
>> @provider.name_and_code
!! #<NoMethodError: undefined method `name_and_code' for #<Provider id: 18396,```

app/models/site.rb
```diff
@@ -67,6 +67,13 @@ def full_address
     address.select(&:present?).join(', ')
   end
 
+  def confirm_full_address
```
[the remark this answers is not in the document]
goodviber: in app/views/support/schools/confirm.html
```
component.with_row do |row|
              row.with_key { t("support.providers.schools.address") }
              row.with_value(text: @site.confirm_full_address)
```
it lists address lines vertically as per the prototype rather than comma separated

spec/forms/support/school_form_spec.rb
goodviber: When I thought I'd be clever and promote the errors form site with
```
 def validate_site
      skip = []
      return if site.valid?

      sites = @identifier_model.sites.where.not(id: nil)
      skip << :location_name unless sites.exists?(location_name:) || location_name.blank?
      skip << :code unless sites.exists?(code:)

      promote_errors(site.errors, skip)
    end

    def promote_errors(site_errors, skip)
      site_errors.each do |site_error|
        next if skip.include?(site_error.attribute)

        errors.add(site_error.attribute, site_error.message)
      end
    end
```
the matchers stopped working?

spec/support/page_objects/sections/location_form.rb
goodviber: Yep

goodviber: Deleted

simonwhatley:
> > The School code input is read-only, but I can change it to text? A Site has 4 addresses and the postcode, so address 1, address 2, town, county, postcode matches I think?
> 
> @simonwhatley The prototype still shows the form as Address Line 3, should it be Town or City? Which is the updated version?

Now that we use GIAS we need all address forms to have:

- address line 1 
- address line 2
- address line 3
- town or city
- county
- postcode

app/views/support/providers/schools_check/show.html.erb
darokel: `provider.decorate.name_and_code?`

app/models/site.rb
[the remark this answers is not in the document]
darokel: Could this go in the site decorator as it's specific to the presentation? We have some other address related things in there

darokel:
> > > The School code input is read-only, but I can change it to text? A Site has 4 addresses and the postcode, so address 1, address 2, town, county, postcode matches I think?
> > 
> > 
> > @simonwhatley The prototype still shows the form as Address Line 3, should it be Town or City? Which is the updated version?
> 
> Now that we use GIAS data we need all address forms to have:
> 
> * address line 1
> * address line 2
> * address line 3
> * town or city
> * county
> * postcode

Discussed offline. Address line 3 was added after the bulk upload work so we're keeping the existing address fields but will introduce this field as a separate ticket as we'll need to handle other areas (like the bulk upload) to make sure things work as expected.

app/views/support/providers/schools_check/show.html.erb
goodviber: If you put `decorates_assigned :provider` in the schools_check_controller.rb, it still doesn't work
```
[1] pry(#<#<Class:0x0000000115431938>>)> provider.name_and_code
NoMethodError: undefined method `name_and_code' for #<Provider id: 18396
```
am I missing something?

goodviber: I think it is because I am calling `def provider` already

app/models/site.rb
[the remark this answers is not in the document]
goodviber: but it has the same issue as the provider decorator

app/views/support/providers/schools_check/show.html.erb
darokel: I would have thought `provider.decorate.name_and_code` would work? Otherwise `ProviderDecorator.new(provider)` in your `def provider`

app/models/site.rb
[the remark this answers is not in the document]
darokel: I think draper allows you to just call `.decorate` on the model?

darokel (approved):
Nice work 👍

PR #3463, opened 2023-03-30
rm duplicate file and unnecessary attrs

goodviber:
### Context
Removing some leftovers
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/forms/support/school_form.rb
```diff
@@ -2,9 +2,8 @@
 
 module Support
   class SchoolForm < Form
-    MAIN_SITE = 'main site'
+    # MAIN_SITE = 'main site'
```
darokel, line 5: Do we need this? I'd prefer we delete rather than comment out

app/views/support/schools/confirm.html.erb
```diff
@@ -1,45 +0,0 @@
-<%= render PageTitle.new(title: "support.providers.schools.confirm") %>
```
defong, line 1: Do these translations needs to be removed as well?

app/forms/support/school_form.rb
goodviber: Yep

defong (approved):

PR #3481, opened 2023-04-07
#1202 Reconcile school form address fields

goodviber:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


goodviber:
[Reconcile school form address fields](https://trello.com/c/p5lszIcg/1202-reconcile-school-form-address-fields)

darokel:
@simonwhatley I think we need to update the example data used in the CSV upload. With the addition of this new Address Line 3, it incorrectly pre-populates the form like so:

Screenshot 2023-04-12 at 09 29 47

app/serializers/api/public/v1/serializable_location.rb
```diff
@@ -51,6 +51,10 @@ class SerializableLocation < JSONAPI::Serializable::Resource
         attribute :street_address_2 do
           @object.address2
         end
+
+        attribute :street_address_3 do
```
darokel, line 55: If we're exposing this as a new field in the public api, we ought to add some notes in api docs as well.

darokel (commented):
Looks good overall. Just left one comment

docs/source/release-notes.html.md.erb
```diff
@@ -5,6 +5,10 @@ weight: 2
 
 # Release Notes
 
+## 12 April 2023
+
+- Rename `address3` field to `town` and add new `address3` field. This is a string, referencing an optional address field.
```
darokel, line 10:
```suggestion
- We have introduced a new `Street address 3` attribute to the Location schema. This is a string, referencing an optional address field.
```

I don't think we changed the current `city` attribute in the swagger spec so we just need to mention that we've introduced a new public Address 3 field if that makes senses.

app/serializers/api/public/v1/serializable_location.rb
goodviber: Done

darokel (approved):
Might just need a double check from Simon but looks good to me

PR #3490, opened 2023-04-13
[1194] Added accredited provider id to provider

defong:
### Context
Accredited provider id
Add accredited provider id

### Guidance to review
skeleton code

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


defong:
[Update existing provider form](https://trello.com/c/pqPIn7dh/1194-update-existing-provider-form)

goodviber (commented):
would this need to be referenced on the api docs etc?

darokel:
> > would this need to be referenced on the api docs
> 
> Nope...
> 
> @darokel do you have perception for exposing this field on the api?

If we want register to exclusively use the api to retrieve these APs then we would need to expose it. Saying that I don't think we need to do it right now as we don't (yet) have that comprehensive list of APs with their ids.

db/migrate/20230413112141_add_accredited_provider_id_to_provider.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAccreditedProviderIdToProvider < ActiveRecord::Migration[7.0]
+  def change
+    add_column :provider, :accredited_provider_id, :integer
```
darokel, line 5: Do we need an index for this?

darokel: yup fair enough, just wondered whether we'd ever search on it but probably not yet

darokel (approved):

db/migrate/20230413112141_add_accredited_provider_id_to_provider.rb
defong: Cool lets stick with YAGNI until there is a plausible future for it :smile_cat:

PR #3505, opened 2023-04-18
Tighten UKPRN validation

avinhurry:
### Context

We are tightening the validation for UKPRN, as these should begin with a 1. There are several instances in the database where they do not. Tightened validation should try to help prevent these from getting into the database in future. 

- Tighten validation on UKPRN, preventing any value that does not begin with a 1 and is not 8 integers from being saved. 
- Change the content on the validation message

### Guidance to review

- Attempt to change the UKPRN to a non valid entry in the various places that we ask the question. 
 
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/spec/factories/providers.rb
+++ b/spec/factories/providers.rb
@@ -23,7 +23,7 @@ FactoryBot.define do
     website { Faker::Internet.url }
     provider_type { :lead_school }
     urn { Faker::Number.number(digits: [5, 6].sample) }
-    ukprn { Faker::Number.within(range: 10000000..19999999) }
+    ukprn { Faker::Number.within(range: 10_000_000..19_999_999) }
     accrediting_provider { 'N' }
     region_code { 'london' }
     association :recruitment_cycle, strategy: :find_or_create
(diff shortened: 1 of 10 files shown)


goodviber (approved):
lgtm

spec/factories/providers.rb
```diff
@@ -23,7 +23,7 @@
     website { Faker::Internet.url }
     provider_type { :lead_school }
     urn { Faker::Number.number(digits: [5, 6].sample) }
-    ukprn { Faker::Number.number(digits: 8) }
+    ukprn { '12345678' }
```
defong, line 26:
```suggestion
Faker::Number.within(range: 1000000..19999999)
```

avinhurry: https://github.com/DFE-Digital/publish-teacher-training/pull/3505/commits/5ca8dc5bccb8a6893c829ae24904163d1c183d62

defong (approved):

PR #3506, opened 2023-04-18
1120 js school search

goodviber:
### Context
As a user, I want to easily add school placements, so that candidates can understand where they may be placed
Why

We have used the term locations to refer to school placements and it’s become confusing. We’re updating the use of the term locations to schools as well as improve the flow to add new schools.

This ticket is to setup the JS search flow via the autocomplete lookup.
What needs to be done

We need to build a new step and page before the schools form which will allow users to search for schools via the autocomplete or go directly to the schools form if they want to manually enter. See dev checklist for possible steps.
Done when

We have a page which allows users to search for schools via autocomplete
We have a way for users to skip the step above and enter a school manually

Add autocomplete to search input on schools search page
### Guidance to review
visit `publish/organisations/1Z1/2023/schools/search` and enter minimum of 3 chars in search bar, dropdown suggestion list should appear
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/controllers/publish/providers/school_search_controller.rb
```diff
@@ -5,7 +5,15 @@ module Providers
     class SchoolSearchController < PublishController
       helper_method :query, :search_result_title_component
 
-      before_action :authorize_provider
+      before_action :authorize_provider, except: %i[suggest]
+
+      def suggest
```
darokel, line 10: I think we've decided to avoid adding custom methods in controllers and keep the actions as RESTful as possible. I'd opt we introduce a new api controller specifically for this like:

https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/controllers/api/schools_controller.rb

darokel:
@goodviber Had a quick play, I think on the JS once you select it should take you straight to the school form using the chosen school rather than do a "search" on that same value. No point in doing the search since we already know the school from the list

goodviber:
> @goodviber Had a quick play, I think on the JS once you select it should take you straight to the school form using the chosen school rather than do a "search" on that same value. No point in doing the search since we already know the school from the list

Have done this, it will cover the same behaviour in non JS I think.

app/controllers/api/public/v1/school_suggestions_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module API
+  module Public
+    module V1
+      class SchoolSuggestionsController < API::Public::V1::ApplicationController
```
darokel, line 6: Hate to be a pain but can we move this to just `API::SchoolSuggestionsController`. It won't be publicly available or documented but internal use

app/controllers/api/public/v1/school_suggestions_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+        def index
+          return render_json_error(status: 400, message: I18n.t('school_suggestion.errors.bad_request')) if invalid_query?
```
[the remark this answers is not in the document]
defong: Oh it just read nicer for me...

Thought we have https://www.rubydoc.info/gems/rubocop-rails/2.17.0/RuboCop/Cop/Rails/HttpStatus enforced...

app/controllers/api/public/v1/school_suggestions_controller.rb
goodviber: Coolio

app/javascript/publish/application.js
```diff
@@ -23,9 +23,20 @@ try {
...
+
   if (autocomplete && providerInput) {
     initAutocomplete(autocomplete, providerInput, providerTemplate)
   }
+  if (schoolAutocomplete && schoolInput) {
+    initAutocomplete(schoolAutocomplete, schoolInput, schoolTemplate, {
+      path: '/publish/suggestions',
+      suggestionTemplate: schoolSuggestions
+    })
+  }
```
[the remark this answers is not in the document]
goodviber: I have done a little refactor, but I wouldn't want to create another autocomplete.js . As it stands you can add any number of autocomplete functions just by adding a few lines to the application.js try block

app/controllers/api/school_suggestions_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module API
+  class SchoolSuggestionsController < API::Public::V1::ApplicationController
```
darokel, line 4:
```suggestion
class SchoolSuggestionsController < PublicAPIController
```

config/routes/publish.rb
```diff
@@ -59,6 +59,7 @@
 
   get '/accept-terms', to: 'terms#edit', as: :accept_terms
   patch '/accept-terms', to: 'terms#update'
+  get '/suggestions', to: '/api/school_suggestions#index'
```
darokel, line 62: this can go here https://github.com/DFE-Digital/publish-teacher-training/blob/1222-support-add-edit-accrediting-provider-step-1-non-js/config/routes/api.rb#L6

spec/factories/gias_schools.rb
```diff
@@ -8,6 +8,7 @@
     town { Faker::Address.city }
     postcode { Faker::Address.postcode }
     status_code { GiasSchool::ESTABLISHMENT_OPEN_STATUS_CODE }
+    searchable { %w[name urn postcode] }
```
darokel, line 11: Do we need to set this, the callback should generate one when it's created?

goodviber: Good point

config/routes/publish.rb
goodviber: Done

darokel (approved):
Think Simon needs a product review but looks good to me on dev side 👍

app/javascript/publish/application.js
```diff
@@ -23,8 +23,22 @@ try {
   const providerTemplate = (result) =>
     result && `${result.name} (${result.code})`
 
+  const schoolAutocomplete = document.getElementById('school-autocomplete')
```
darokel, line 26: We may need to clean this up a little we'll need one more of these for the accredited provider look up in upcoming tickets. Something to think about

defong (approved):

PR #3516, opened 2023-04-20
updated read me to help with caddy error

syed87:
### Context
I ran into some issues setting up caddy
I've added some suggestions if anyone gets an error message using caddy run


goodviber (commented):
I think we also should be running http://publish.localhost rather thatn https?

goodviber (approved):

PR #3536, opened 2023-04-28
[1196] Added provider check your answers page

defong:
### Context
Provider check your answers page
Added provider check your answers page
### Guidance to review

The journey is complete
image
image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support/providers/onboarding/checks_controller.rb
+++ b/app/controllers/support/providers/onboarding/checks_controller.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Support
+  module Providers
+    module Onboarding
+      class ChecksController < SupportController
+        def show
+          provider_form
+          provider_contact_form
+        end
+
+        def update
+          provider = recruitment_cycle.providers.new(provider_attributes)
+          return unless provider.save
+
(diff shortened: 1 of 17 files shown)


defong:
[Update existing provider form - check your answers page](https://trello.com/c/iADW35dr/1196-update-existing-provider-form-check-your-answers-page)

darokel (approved):
Two small comments (don't have to be addressed in this PR)

- once product approved, then i think we should probably just clean up and delete that old flow including feature flags etc so this becomes the default
- when i created a new provider, i noticed they were not returned when searching for their name/code in the support console but they're returned in the autocomplete lookup when logged in as an admin?

otherwise 👍 🧑‍🍳

defong:
> Two small comments (don't have to be addressed in this PR)
> 
>     * once product approved, then i think we should probably just clean up and delete that old flow including feature flags etc so this becomes the default
> 
>     * when i created a new provider, i noticed they were not returned when searching for their name/code in the support console but they're returned in the autocomplete lookup when logged in as an admin?
> 
> 
> otherwise +1 cook

Will card it up :beers:

PR #3537, opened 2023-04-28
[1197] edit contact details

goodviber:
### Context
User need

As a user, I want to be able to add providers in Publish, so they can be setup correctly to use the service.
Why

ITT reforms requires us to modify the way we register and onboard providers into Publish.

This ticket is to update the existing flow when editing a provider from the support console to use the newly built form.
What needs to be done

We need to update the existing form when editing a provider to match the one used in the new flow.
Extension of the support provider edit form
### Guidance to review
from `/support/2023/providers`  click on provider and change details in upper and lower part of the form
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


goodviber:
[Update edit provider flow](https://trello.com/c/irYEgOmi/1197-update-edit-provider-flow)

app/helpers/publish/course_helper.rb
```diff
@@ -3,7 +3,7 @@
 module Publish
   module CourseHelper
     def course_value_provided?(value)
-      value.presence || tag.span(t('course.value_not_entered'), class: 'govuk-hint').html_safe
+      value.presence || tag.span(t('value_not_entered'), class: 'govuk-hint').html_safe
```
darokel, line 6: this and the support one look identical, maybe we only keep one?

spec/helpers/support/value_helper_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Support
+  describe ValueHelper do
```
darokel, line 6: if we only keep one i don't think we need the same spec

config/routes/support.rb
```diff
@@ -10,6 +10,7 @@
       end
       resource :onboarding, only: %i[new create]
     end
+    resources :providers_contact_details, only: %i[edit update]
```
darokel, line 13: Just thinking, this could probably be nested under `resource :providers`? but just the non pluralised version `resource :contact_details, only: %i[edit update], path: 'contact-details'`

then you would get:

`/providers/122/contact-details/edit`

It's not a blocker but we still need the dash paths to be consistent with the other support paths

darokel (commented):
Looks good so far. Just left a couple of comments. One thing I noticed is when you try to edit a provider that is accredited but does not have an id, it still allows them to submit the form. I think for these existing ones we want to try and validate the field at the point of update so we can try to capture this information as early as possible (i think but check with Simon too)

app/validators/accredited_provider_id_format_validator.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class AccreditedProviderIdFormatValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
```
darokel, line 4: I think we already had some validation logic for this? @defong can you confirm

app/models/provider.rb
```diff
@@ -156,6 +156,8 @@ def current_accredited_courses
 
   validate :add_enrichment_errors
 
+  validates :accredited_provider_id, accredited_provider_id_format: { allow_blank: false }, on: :update, if: :accredited_provider?
```
darokel, line 159: I think this might introduce an additional problem if providers (not support) try to update some other info and it fails because of this validation. At the moment, this isn't exposed as a field for them to edit so it'll be a dead end. Maybe we allow a blank value but only try to validate if one is provided (in which case it will only trigger from the support side if that makes sense)

config/routes/support.rb
```diff
@@ -11,6 +11,7 @@
       resource :onboarding, only: %i[new create]
     end
     resources :providers, except: %i[destroy] do
+      resource :contact_details, only: %i[edit update], path: 'contact-details'
```
darokel, line 14: 👍

app/models/provider.rb
goodviber: Yeah it's tricky

goodviber: I think that should work

darokel (approved):
Look good to me, nice work. Might want to double check the AP id validation stuff, felt like I saw Sau implement that earlier somewhere 🤔

app/validators/accredited_provider_id_format_validator.rb
goodviber: We have a few format specs, maybe you are thinking of `spec/validators/ukprn_format_validator_spec.rb` ?

defong: Oh I just done an inline validation

https://github.com/DFE-Digital/publish-teacher-training/blob/b331261175e0b5a2a1ee9ca857a442fd8094161a/app/forms/support/provider_form.rb#L80-L95

app/views/support/providers/edit.html.erb
```diff
@@ -1,22 +1,50 @@
-<h1 class="govuk-heading-l">Edit provider details</h1>
+<%= render PageTitle.new(title: "support.providers.edit") %>
```
defong, line 1: missing `has_error`

https://github.com/DFE-Digital/teacher-training-api/blob/517e8163c7528ce8398d499f660746d7102dd0d1/spec/components/page_title_spec.rb

app/views/support/providers/edit_contact_details.html.erb
```diff
@@ -0,0 +1,50 @@
+<%= render PageTitle.new(title: "support.providers.edit_contact") %>
```
defong, line 1: missing `has_error`

https://github.com/DFE-Digital/teacher-training-api/blob/517e8163c7528ce8398d499f660746d7102dd0d1/spec/components/page_title_spec.rb

app/helpers/publish/value_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Publish
+  module ValueHelper
+    def value_provided?(value)
+      value.presence || tag.span(t('value_not_entered'), class: 'govuk-hint').html_safe
```
defong, line 6: Change `value_not_entered`  to `value_not_entered_html`

So it becomes
```suggestion
      value.presence || t('value_not_entered_html')
```

app/validators/accredited_provider_id_format_validator.rb
goodviber: I think we have to allow blank on update.

goodviber: I like that regex!

PR #3543, opened 2023-05-03
238 add aks review apps

smneal:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/lib/publish_constraint.rb
```diff
@@ -2,6 +2,6 @@
 
 class PublishConstraint
   def matches?(request)
-    Settings.base_url&.include?(request.host) || Settings.publish_api_url&.include?(request.host) || request.host.include?('publish-teacher-training-pr')
+    Settings.base_url&.include?(request.host) || Settings.publish_api_url&.include?(request.host) || request.host.include?('publish-teacher-training-pr') || request.host.include?('ptt-review')
```
darokel, line 5: I think we need to setup an alias for the find equivalent. So if we're going to have something:

`https://ptt-review-7777.cluster3.development.teacherservices.cloud` we'll need an alias for `https://find-review-7777.cluster3.development.teacherservices.cloud`

and whitelist in the `FindConstraint` so those routes can be activated

PR #3567, opened 2023-05-15
[1292] Fix school search bug

darokel:
### Context

https://trello.com/c/fmC7d0ae/1292-bug-adding-a-school-in-publish

Fixes a bug in the school autocomplete search where selecting a school from the dropdown can still trigger a search. The fix is to set the school id form the JS and skip the search all together unless a value isn't selected which should then trigger a search.

### Guidance to review

- Go to the school search
- Try searching for the URN `144065`
- Assert that if you select the provider it doesn't trigger search

I've left the current implementation in where if you do not select a school but search a query where we match on one result then also skip the search otherwise we'll just show the results page with one option.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -732,6 +732,7 @@ GEM
 PLATFORMS
   arm64-darwin-20
   arm64-darwin-21
+  arm64-darwin-22
   x86_64-darwin-19
   x86_64-darwin-20
   x86_64-darwin-21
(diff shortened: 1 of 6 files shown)


darokel:
[Bug - Adding a School in Publish](https://trello.com/c/fmC7d0ae/1292-bug-adding-a-school-in-publish)

PR #3575, opened 2023-05-18
1228 support edit accrediting provider 

goodviber:
### Context
User need

As a user, I want to easily add accrediting providers to lead schools, so that these providers can be setup correctly in Publish
Why

This will give users and providers flexibility in managing their APs.
What needs to be done

We need to implement the edit functionality for AP in the new AP tab for lead schools so the description can be changed.
Done when

We are able to edit an AP’s description from the lead school's AP tab

Update routes/form/controller to allow update of provider.accredited_provider_enrichments.

### Guidance to review
In support console, select a  provider with accredited providers, click 'Add accredited provider' and update description via Change link
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/forms/support/accredited_provider_form.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Support
+  class AccreditedProviderForm < Form
```
darokel, line 4: I removed this in an earlier PR as I feel we should just be using the same form for support/publish. It's essentially the same behaviour? Couldn't we use the existing one https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/forms/accredited_provider_form.rb

goodviber: Done

config/locales/en.yml
```diff
@@ -346,7 +346,12 @@ en:
         new:
           title: About the accredited provider
           hint: Tell candidates about the accredited provider. You could mention their academic specialities and achievements.
-          caption: *accredited_provider_caption
+          caption: "Add accredited provider - %{provider_name} (%{code})"
+        edit: 
+          title: About the accredited provider
+          hint: Tell candidates about the accredited provider. You could mention their academic specialities and achievements.
+          caption: "%{provider_name}"
+          updated: About the accredited provider updated
```
defong, line 354: Are you actually using any of these?

goodviber: Screenshot 2023-05-19 at 14 46 00

Yes, I think so.

defong: Whats with caption?

PR #3579, opened 2023-05-19
[1229] delete support accredited provider

goodviber:
### Context
User need

As a user, I want to easily add accrediting providers to lead schools, so that these providers can be setup correctly in Publish
Why

This will give users and providers flexibility in managing their APs.
What needs to be done

We need to implement the delete functionality for an AP in the new AP tab so they can be removed from the lead school
Done when

If an AP is removable, we are able to click the delete link and see a confirm page and remove the AP upon form submission
If an AP is not removable, we see a page with some content telling us the AP cannot be removed

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support/providers/accredited_providers_controller.rb
+++ b/app/controllers/support/providers/accredited_providers_controller.rb
@@ -51,7 +51,6 @@ module Support
 
       def delete
         cannot_delete
-        # @cannot_delete = provider.courses.exists?(accredited_provider_code: accredited_provider.provider_code)
       end
 
       def destroy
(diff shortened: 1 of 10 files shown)


goodviber:
[Support: Add & Edit Accrediting Provider - Remove an AP](https://trello.com/c/R8bpW8pS/1229-support-add-edit-accrediting-provider-remove-an-ap)

app/controllers/support/providers/accredited_providers_controller.rb
```diff
@@ -49,8 +49,30 @@ def update
         end
       end
 
+      def delete
+        @cannot_delete = provider.courses.exists?(accredited_provider_code: accredited_provider.provider_code)
```
darokel, line 53: I would just memoize this variable as a method then you could do `cannot_delete?`

darokel (approved):

PR #3588, opened 2023-05-24
[1321] Remove change vacancies from course show page

defong:
### Context

`change vacancies` link on course show page

Remove `change vacancies` link on course show page

#### Before
image

#### After
image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/feature_service.rb
+++ b/app/services/feature_service.rb
@@ -13,7 +13,7 @@ module FeatureService
 
       segments = feature_name.to_s.split('.')
 
-      segments.reduce(Settings.features) { |config, segment| config[segment] }
+      segments.reduce(Settings.features) { |config, segment| config[segment] } == true
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/feature_service.rb
```diff
@@ -13,7 +13,7 @@ def enabled?(feature_name)
 
       segments = feature_name.to_s.split('.')
 
-      segments.reduce(Settings.features) { |config, segment| config[segment] }
+      segments.reduce(Settings.features) { |config, segment| config[segment] } == true
```
goodviber, line 16: Why this change?

defong: The method signature is

`    def enabled?(feature_name)`

So it shoud only return a true or a false given enabled`?` 

Without it any value can be returned including `nil`

PR #3612, opened 2023-06-02
[xxx] A few snag fixes

darokel:
### Context

Fixes some of the issues spotted in the snag.

### DIFF
--- a/app/views/publish/courses/accredited_provider/new.html.erb
+++ b/app/views/publish/courses/accredited_provider/new.html.erb
@@ -23,6 +23,7 @@
           <% if @provider.accredited_bodies.length > 0 %>
             <%= render partial: "provider_suggestion", collection: @provider.accredited_bodies.sort_by { |k| k[:provider_name] }, locals: { form: fields } %>
 
+            <!-- TODO: Remove this field once feature flag is removed -->
             <% unless FeatureService.enabled?(:accredited_provider_search) %>
               <div class="govuk-radios__divider">or</div>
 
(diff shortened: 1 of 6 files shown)


app/views/publish/courses/accredited_provider/new.html.erb
```diff
@@ -23,24 +23,26 @@
           <% if @provider.accredited_bodies.length > 0 %>
             <%= render partial: "provider_suggestion", collection: @provider.accredited_bodies.sort_by { |k| k[:provider_name] }, locals: { form: fields } %>
 
-            <div class="govuk-radios__divider">or</div>
+            <% unless FeatureService.enabled?(:accredited_provider_search) %>
```
avinhurry, line 26: Makes sense, so we will remove the additional field when we remove the feature flag? Is it worth leaving a comment so we remember to remove it?

config/settings.yml
```diff
@@ -5,7 +5,7 @@ support_email: becomingateacher@digital.education.gov.uk
 campaign_email: itt.engagement@education.gov.uk
 
 search_ui:
-  base_url: http://localhost:3002
+  base_url: https://find.localhost
```
avinhurry, line 8: What's this?

defong: Can we get rid of it?

app/views/publish/courses/accredited_provider/new.html.erb
darokel: Yup, you just left one on this PR 😂

config/settings.yml
darokel: The `view on find link` in dev didn't load the correct url

app/views/publish/courses/accredited_provider/new.html.erb
avinhurry: lol I meant in the code?

config/settings.yml
darokel: It's still used

app/views/publish/courses/accredited_provider/new.html.erb
darokel: Screen Recording (Safari)

I also left it in the code 😜

avinhurry (approved):

PR #3616, opened 2023-06-05
[1332] add study sites

goodviber:
### Context
What needs to be done

For this ticket we’ll need to add a new tab and index page to render the list of the provider’s study sites.
Done when

    There is a link in the main nav for Study sites
    I can see a list of the provider’s study sites on this page
    I can add an individual study site manually

### Guidance to review
Click on Study sites on main nav, follow the add a study site flow...
### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


goodviber:
[Add study sites](https://trello.com/c/qI5MxOpo/1322-add-study-sites)

app/controllers/publish/providers/study_sites_check_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      private
+
+      def pundit
+        authorize provider, :show?
+      end
+
+      def new_form
+        @study_site_form = ::Support::SchoolForm.new(provider, site)
```
darokel, line 27: I think we should probably use the Publish `SchoolForm` rather than the support one?

app/controllers/publish/providers/study_sites_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      before_action :build_study_site, only: %i[new create]
+
+      def index
+        @study_sites = provider.study_sites.sort_by(&:location_name)
+      end
+
+      def show; end
+
+      def new
+        @study_site_form = ::Support::SchoolForm.new(provider, @site, params: gias_school_params)
```
darokel, line 17: Same here, we're mixing the support and publish `SchoolForm`. Should probably use the publish one?

app/controllers/publish/providers/study_sites_controller.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module Publish
+  module Providers
+    class StudySitesController < PublishController
```
darokel, line 5: Can this controller also re-use same logic as the `SchoolsController`?

spec/features/publish/managing_study_sites_spec.rb
```diff
@@ -0,0 +1,138 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+feature "Managing a provider's schools", { can_edit_current_and_next_cycles: false } do
```
darokel, line 5:
```suggestion
feature "Managing a provider's study sites", { can_edit_current_and_next_cycles: false } do
```

darokel (commented):
Review app looks good. My main concern is around the duplication. Can we tuck some shared logic into a module or something? Happy for that to be carded up separately so we can get this moving

app/controllers/publish/providers/study_sites_controller.rb
goodviber: No we dont validate the code on update, it is set on create only

app/controllers/publish/providers/study_sites_controller.rb
goodviber: I am not sure that is the way to go, we could do a module like so: 
```
module Locations
  extend ActiveSupport::Concern

  included do
    before_action :pundit
    before_action :site, only: %i[show delete]
    before_action :build_site, only: %i[new create]
  end

  def show; end

  def new
    @school_form = ::Support::SchoolForm.new(provider, @site, params: gias_school_params)
    @school_form.clear_stash
  end

  def edit
    @school_form = ::Publish::SchoolForm.new(site)
  end

  def create
    @school_form = ::Support::SchoolForm.new(provider, @site, params: site_params(:support_school_form))
    if @school_form.stash
      redirect_to publish_provider_recruitment_cycle_check_school_path
    else
      render :new
    end
  end

  def update
    @school_form = ::Publish::SchoolForm.new(site, params: site_params(:publish_school_form))

    if @school_form.save!
      course_updated_message('School')

      redirect_to publish_provider_recruitment_cycle_school_path(
        @school_form.provider_code, @school_form.recruitment_cycle_year, site.id
      )
    else
      render :edit
    end
  end

  def delete; end

  def destroy
    site.destroy!
    flash[:success] = 'School removed'
  end

  private

  def pundit
    authorize provider, :show?
  end

  def site_params(param_form_key)
    params.require(param_form_key).permit(::Publish::SchoolForm::FIELDS)
  end

  def gias_school_params
    return {} unless params[:school_id] || params[:study_site_id]

    gias_school.school_attributes
  end

  def gias_school
    @gias_school ||= GiasSchool.find(params[:school_id])
  end
end
```
but we still would need conditional redirects on create, update and destroy. And the form models would share the same name. To my eye cognitive load in following this is more than having seperate controllers?

spec/features/publish/managing_study_sites_spec.rb
goodviber: oops

goodviber:
[Search for study sites (js)](https://trello.com/c/Jn4XlcKv/1329-search-for-study-sites-js)

app/controllers/publish/providers/study_site_search_controller.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+module Publish
+  module Providers
+    class StudySiteSearchController < PublishController
```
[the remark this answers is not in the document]
goodviber: will have a look

app/controllers/publish/providers/study_sites_controller.rb
darokel: Yup fair enough

app/controllers/publish/providers/study_sites_controller.rb
darokel: So do we still need to use the support one?

goodviber: am looking at it

app/controllers/publish/providers/study_site_search_controller.rb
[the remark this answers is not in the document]
goodviber: again we would need to do some conditional redirects etc.

darokel (approved):

PR #3636, opened 2023-06-09
[1362] Fix apply button logic

darokel:
### Context

https://trello.com/c/tMH6Iwbq/1362-issue-with-publish-closed-courses-are-open-and-open-open-courses-closed

The apply button is still lingering off some old vacancy logic so causing confusion to providers with a warning message being incorrectly shown.

- Update the logic to render apply button
- Also revert the view logic to align application status but update the underlying method as it contains some other rules around the open date and the course being findable

### Guidance to review

- Open/Close a course and check apply button/warning text is shown as expected

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -228,5 +228,13 @@ FactoryBot.define do
     trait :engineers_teach_physics do
       campaign_name { :engineers_teach_physics }
     end
+
+    trait :open do
+      application_status { :open }
+    end
+
+    trait :closed do
+      application_status { :closed }
+    end
   end
 end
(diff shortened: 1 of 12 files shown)


spec/factories/courses.rb
```diff
@@ -228,5 +228,13 @@
...
       campaign_name { :engineers_teach_physics }
     end
+
+    trait :open do
+      application_status { :open }
+    end
+
+    trait :closed do
+      application_status { :closed }
+    end
```
defong, line 238: You don't need that.

defong (approved):

PR #3667, opened 2023-06-20
[19] Enable editing study sites on a course from the basic details page

avinhurry:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/views/publish/courses/_basic_details_tab.html.erb
+++ b/app/views/publish/courses/_basic_details_tab.html.erb
@@ -138,6 +138,13 @@
       end
     end
 
+    summary_list.with_row(html_attributes: { data: { qa: "course__study_sites" } }) do |row|
+      row.with_key { "Study site" }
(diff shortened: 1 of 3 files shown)


avinhurry:
[Add edit study sites form to course details page](https://trello.com/c/W6uFyTTp/19-add-edit-study-sites-form-to-course-details-page)

JR-G:
Looks good Avin! I've been having a play around on the review app. Would we want different copy here if there are no sites available yet?
image

tomas-stefano (approved):

avinhurry:
> Looks good Avin! I've been having a play around on the review app. Would we want different copy here if there are no sites available yet? image

Thanks @JR-G! I've fixed that quirk, there is also a similar problem in the Add course flow which we will need to address before we switch on the feature flag.

PR #3691, opened 2023-07-03
Fix the Publish course information bug

JR-G:
### Context

There are currently two bugs existing in Publish:

**Bug no1:**
The ordering of `course_enrichments` (_an enrichment stores additional information about a course for things like the interview process_) is not consistent across environments and there are currently two approaches to returning the latest `course_enrichment`:

- The **first approach** is to use `@course.enrichments.last[:about_course]`
- The **second approach** is to use `@course.about_course`

The first approach is what causes the inconsistency – locally this returns the expected result but in all envs up to prod it does not. The second is consistent because it utilises the `scope :most_recent, -> { order(created_at: :desc, id: :desc) }` scope on the `CourseEnrichment` model.

On this page we're using the second approach:
image

On this page (which you access when you go to update these values) we're using the first to pre-populate the text fields:
image

This is the cause of the bug. By using the first approach to pre-populate these fields, we're unintentionally pre-populating them with old enrichment data, then when the user goes to update the information it writes the old data to the latest course enrichment object.
image

**Bug no2:**
The second bug unfortunately creates a nasty combination with the first. Any changes to course enrichments is being instantly published on Find. Even if the user hasn't pressed 'Publish changes'.
image

 

- Ensured consistency in ordering approach by using the decorator methods where possible
- Introduced `published` decorator methods to distinguish between what enrichments have and haven't been published and to only display published enrichments on Find i.e.:
```
  def published_how_school_placements_work
    return nil if no_published_enrichment?

    object.current_published_enrichment[:interview_process]
  end
```
These utilise:
```
  def no_published_enrichment?
    object.current_published_enrichment.nil?
  end
```
`current_published_enrichment` is a new method on the `Course` model:
```
  def current_published_enrichment
    enrichments.where(status: 'published').order(last_published_timestamp_utc: :desc).first
  end
```
This required an update to the `new_draft_attributes` method to set the `last_published_timestamp_utc` to nil for new (draft) enrichments:
```
    def new_draft_attributes
      latest_published_enrichment = most_recent.published.first
      if latest_published_enrichment.present?
        latest_published_enrichment_attributes = latest_published_enrichment
                                                 .dup
                                                 .attributes
                                                 .with_indifferent_access
                                                 .except(:json_data)

        latest_published_enrichment_attributes[:status] = :draft
        latest_published_enrichment_attributes[:last_published_timestamp_utc] = nil
        latest_published_enrichment_attributes
      else
        { status: :draft, last_published_timestamp_utc: nil }.with_indifferent_access
      end
    end
```
- Updated `has_unpublished_changes?` to utilise this new concept and introduced a `draft_enrichment` method:
```
  def has_unpublished_changes?
    published? && draft_enrichment.present?
  end

  def draft_enrichment
    enrichments.draft.most_recent.first
  end
```
with `published?` being:
```
  def published?
    current_published_enrichment.present?
  end
```
- updated `last_published_at` to use the `current_published_enrichment` method:
```
  def last_published_at
    current_published_enrichment&.last_published_timestamp_utc
  end
```

-----
`has_unpublished_changes?` has now expanded to:
```
  def has_unpublished_changes?
    return false if all_enrichments_are_published?

    (published? && draft_enrichment.present?) || (last_published_at.present? && enrichment_not_withdrawn?)
  end
```

It needs to cover three scenarios from what I can understand:

1. Published course with no changes
2. Published course with unpublished changes
3. Withdrawn course

Which is in thanks to `app/views/publish/courses/_course_button_panel.html.erb`

### Guidance to review

This is probably in need of quite a refactor but this at least fixes the bug in the meantime. I've introduced a number of `published_[attribute]` methods on the `CourseDecorator` that are not being used. From manually testing it seems like we don't actually need to worry about the enrichment overwriting problem, that we're dealing with on course information. I'll 🪓  them if someone can confirm this.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -344,7 +344,7 @@ class CourseDecorator < ApplicationDecorator
   def published_how_school_placements_work
     return nil if no_published_enrichment?
 
-    object.current_published_enrichment[:interview_process]
+    object.current_published_enrichment[:how_school_placements_work]
   end
 
   def published_fee_uk_eu
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -329,6 +329,52 @@ class CourseDecorator < ApplicationDecorator
     object.enrichment_attribute(:financial_support)
   end
 
+  def published_about_course
+    return nil if no_published_enrichment?
+
+    object.current_published_enrichment[:about_course]
+  end
+
+  def published_interview_process
+    return nil if no_published_enrichment?
+
+    object.current_published_enrichment[:interview_process]
+  end
+
+  def published_how_school_placements_work
+    return nil if no_published_enrichment?
+
+    object.current_published_enrichment[:interview_process]
+  end
+
+  def published_fee_uk_eu
+    return nil if no_published_enrichment?
+
+    object.current_published_enrichment[:fee_uk_eu]
+  end
+
+  def published_fee_international
+    return nil if no_published_enrichment?
+
+    object.current_published_enrichment[:fee_international]
+  end
+
+  def published_fee_details
+    return nil if no_published_enrichment?
+
+    object.current_published_enrichment[:fee_details]
+  end
+
+  def published_financial_support
+    return nil if no_published_enrichment?
+
+    object.current_published_enrichment[:financial_support]
+  end
+
+  def no_published_enrichment?
+    object.current_published_enrichment.nil?
+  end
+
   def financial_incentive_details
     financial_incentive = object.financial_incentives.first
     bursary_amount = number_to_currency(financial_incentive&.bursary_amount)
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -330,49 +330,45 @@ class CourseDecorator < ApplicationDecorator
   end
 
   def published_about_course
-    return nil if no_published_enrichment?
+    return unless current_published_enrichment
 
-    object.current_published_enrichment[:about_course]
+    current_published_enrichment[:about_course]
   end
 
   def published_interview_process
-    return nil if no_published_enrichment?
+    return unless current_published_enrichment
 
-    object.current_published_enrichment[:interview_process]
+    current_published_enrichment[:interview_process]
   end
 
   def published_how_school_placements_work
-    return nil if no_published_enrichment?
+    return unless current_published_enrichment
 
-    object.current_published_enrichment[:how_school_placements_work]
+    current_published_enrichment[:how_school_placements_work]
   end
 
   def published_fee_uk_eu
-    return nil if no_published_enrichment?
+    return unless current_published_enrichment
 
-    object.current_published_enrichment[:fee_uk_eu]
+    current_published_enrichment[:fee_uk_eu]
   end
 
   def published_fee_international
-    return nil if no_published_enrichment?
+    return unless current_published_enrichment
 
-    object.current_published_enrichment[:fee_international]
+    current_published_enrichment[:fee_international]
   end
 
   def published_fee_details
-    return nil if no_published_enrichment?
+    return unless current_published_enrichment
 
-    object.current_published_enrichment[:fee_details]
+    current_published_enrichment[:fee_details]
   end
 
   def published_financial_support
-    return nil if no_published_enrichment?
+    return unless current_published_enrichment
 
-    object.current_published_enrichment[:financial_support]
-  end
-
-  def no_published_enrichment?
-    object.current_published_enrichment.nil?
+    current_published_enrichment[:financial_support]
   end
 
   def financial_incentive_details
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -543,7 +543,7 @@ class Course < ApplicationRecord
   end
 
   def last_published_at
-    current_published_enrichment&.last_published_timestamp_utc || enrichment_attribute(:last_published_timestamp_utc)
+    enrichments.maximum(:last_published_timestamp_utc)
   end
 
   def withdrawn_at
@@ -850,7 +850,7 @@ class Course < ApplicationRecord
   end
 
   def all_enrichments_are_published?
-    (enrichments.one? && published?) || enrichments.flat_map(&:status).none?('draft')
+    (enrichments.one? && published?) || enrichments.none? { |enrichment| %w[draft withdrawn].include?(enrichment.status) }
   end
 
   def assignable_after_publish(course_params, is_admin)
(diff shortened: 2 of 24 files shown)


avinhurry (approved):
Nice work on this! 🎉 

I tested both bugs both locally and on the review app, comparing it to QA and it looks good to me. Left a minor comment at your discretion. 

It might be worth getting a second pair of eyes on this?

app/models/course.rb
```diff
@@ -157,9 +157,10 @@ def new_draft_attributes
                                                  .except(:json_data)
 
         latest_published_enrichment_attributes[:status] = :draft
+        latest_published_enrichment_attributes[:last_published_timestamp_utc] = nil
```
duncanjbrown, line 160: Maybe to avoid bugs return the core attributes from the `if` and then merge in the overridden params once, eg

```ruby
new_enrichment_attributes = if latest_published_enrichment.present?
        latest_published_enrichment_attributes = latest_published_enrichment
                                                 .dup
                                                 .attributes
                                                 .with_indifferent_access
                                                 .except(:json_data)
  else
    {}
  end

new_enrichment_attributes.merge(
  { status: :draft, last_published_timestamp_utc: nil }
)
```

Also — could we use `current_published_enrichment` instead of `most_recent.published.first`?

app/models/course.rb
```diff
@@ -540,7 +541,7 @@ def gcse_grade_required
   end
 
   def last_published_at
-    latest_enrichment&.last_published_timestamp_utc
+    current_published_enrichment&.last_published_timestamp_utc || enrichment_attribute(:last_published_timestamp_utc)
```
duncanjbrown, line 544: any reason this couldn't be the max `last_published_timestamp_utc` from all associated enrichments?

app/models/course.rb
```diff
@@ -822,6 +833,18 @@ def latest_enrichment
...
+  def published?
+    current_published_enrichment.present?
+  end
+
+  def enrichment_not_withdrawn?
+    !enrichments.most_recent.first.withdrawn?
+  end
+
+  def all_enrichments_are_published?
+    (enrichments.one? && published?) || enrichments.flat_map(&:status).none?('draft')
```
duncanjbrown, line 845: what if one is withdrawn?

app/models/course.rb
```diff
@@ -670,7 +671,13 @@ def is_published?
   end
 
   def has_unpublished_changes?
```
duncanjbrown, line 673: I know this one's painful! 

Does it imply `published, but with unpublished changes`? or just that there are unpublished changes

Would `(published? && draft_enrichment.present?)` be true if there were an old draft that was never published followed by a later published enrichment? Should this return true in that case?

(This would also cause `all_enrichments_are_published?` not to be true)

then in `(last_published_at.present? && enrichment_not_withdrawn?)`, we're assuming that the potentially-withdrawn enrichment is the one that was published. I think i now see why we look at the "current" enrichment in `last_published_at`. This is confusing.

Does this method actually want to check whether there exists a draft enrichment created after the last time the course was published?

app/services/courses/content_status_service.rb
```diff
@@ -6,7 +6,7 @@ def execute(enrichment:)
       return :rolled_over if enrichment&.rolled_over?
       return :published if enrichment&.published?
       return :withdrawn if enrichment&.withdrawn?
-      return :published_with_unpublished_changes if enrichment&.has_been_published_before?
+      return :published_with_unpublished_changes if enrichment&.has_been_published_before? || (enrichment&.course&.enrichments&.most_recent && enrichment&.course&.enrichments&.many?)
```
duncanjbrown, line 9: why couldn't this be `enrichment.course.published? && enrichment.course.has_unpublished_changes?`?

app/services/courses/copy.rb
```diff
@@ -53,11 +53,14 @@ class Copy
...
       fields.filter_map do |name, field|
-        source_value = source_course.enrichments.last[field]
-        if source_value.present?
-          course.enrichments.last[field] = source_value
-          [name, field]
-        end
+        source_value = source_course.enrichment_attribute(field)
+        next if source_value.blank?
+
+        enrichments_to_update = course.enrichments.most_recent&.first
```
duncanjbrown, line 59: Should this be singular?

app/services/courses/copy.rb
```diff
@@ -53,11 +53,14 @@ class Copy
...
+        enrichments_to_update.assign_attributes(field => source_value)
+        enrichments_to_update.save
```
duncanjbrown, line 61: Is it right to save each attr in the loop, rather than gathering the enrichment data as a whole and saving it once?

app/views/find/courses/_about_course.html.erb
```diff
@@ -1,7 +1,9 @@
 <div class="govuk-!-margin-bottom-8">
   <h2 class="govuk-heading-l" id="section-about">Course summary</h2>
   <div data-qa="course__about_course">
-    <% if course.about_course.present? %>
+    <% if course.published_about_course.present? %>
+      <%= markdown(course.published_about_course) %>
+    <% elsif course.about_course.present? %>
```
duncanjbrown, line 6: Why do we check for both? To support preview?

JR-G: that's right!

app/decorators/course_decorator.rb
```diff
@@ -331,6 +331,52 @@ def financial_support
     object.enrichment_attribute(:financial_support)
   end
 
+  def published_about_course
+    return nil if no_published_enrichment?
```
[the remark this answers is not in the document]
JR-G: Annoyingly I lost my comment but I need form of guard clause in place on these methods due to the preview page. Unfortunately I can't call `dig` due to current_published_enrichment not being a `Hash`

app/models/course.rb
JR-G: I've refactored it to use the above but `current_published_enrichment` breaks a handful of specs, so I'd like to spend a bit more time understanding the code and making sure logic and behaviour is being sufficiently tested before making further changes

app/models/course.rb
JR-G: Oh yes the source of all pain. Both... well, it needs to cover multiple scenarios due to `app/views/publish/courses/_course_button_panel.html.erb:2`. 

I assume it is possible we could end up with a draft enrichment that is superseded, but my instinct is that this is an edge case, then again I wouldn't be too surprised!

inulty-dfe (approved):
Tested and approved. Refactoring will be considered later after rollover.

avinhurry:
[Hide degree requirement rows for TDA courses](https://trello.com/c/mQzGTkS9/1672-hide-degree-requirement-rows-for-tda-courses)

PR #3697, opened 2023-07-05
[318] Hide add course button when organisation details are missing (next cycle only)

avinhurry:
### Context

Before adding a course in the next cycle, you will have to have filled in some of the organisation details. If the organisation details have not been filled in the "Add course" button will be hidden. 

Hide the add course button under the following conditions:

- If the provider doesn't have any accredited providers attached and the provider is not an accredited provider
- If the provider doesn't have any schools
- If the provider doesn't have any study sites
- Link to these sections if the add course button is hidden

### Guidance to review

Product: As this is only enabled in the next cycle, it will be easiest to do a screenshare. 

Dev: As above or you can rollover your local environment and be sure to activate the `study_sites` feature flag.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/add_course_button.html.erb
+++ b/app/components/add_course_button.html.erb
@@ -10,17 +10,14 @@
 
 <% else %>
 
-<div class="govuk-inset-text">
+  <div class="govuk-inset-text">
     <p class="govuk-body">Before adding a course, you need to:</p>
-      <ul class="govuk-list govuk-list--bullet">
-
-        <% incomplete_sections_hash.each do |section, path| %>
-          <% if send(section) %>
-            <li>
-              <%= govuk_link_to("Add #{incomplete_section_article(section)} #{incomplete_section_label_suffix(section)}", path) %>
-            </li>
-          <% end %>
-        <% end %>
-      </ul>
+    <ul class="govuk-list govuk-list--bullet">
+      <% incomplete_sections.each do |section| %>
+        <li>
+          <%= govuk_link_to(section.name, section.path) %>
+        </li>
+      <% end %>
+    </ul>
   </div>
 <% end %>
--- a/app/components/add_course_button.html.erb
+++ b/app/components/add_course_button.html.erb
@@ -1,0 +1,26 @@
+<% if required_organisation_details_present? %>
+  <%= govuk_button_link_to(
+    "Add course",
+    new_publish_provider_recruitment_cycle_course_path(
+      provider_code: provider.provider_code,
+      recruitment_cycle_year: provider.recruitment_cycle_year
+    ),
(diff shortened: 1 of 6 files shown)


avinhurry:
[Hide add course button when organisation details are missing](https://trello.com/c/KifaTR39/318-hide-add-course-button-when-organisation-details-are-missing)

app/components/add_course_button.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  ) %>
+
+<% else %>
+
+<div class="govuk-inset-text">
+    <p class="govuk-body">Before adding a course, you need to:</p>
+      <ul class="govuk-list govuk-list--bullet">
+
+        <% incomplete_sections_hash.each do |section, path| %>
+          <% if send(section) %>
```
tomas-stefano, line 18: I'd change here to something like:

```
   incomplete_sections.each do |section|
      govuk_link_to("Add #{section.name}", section.path
   end
```
The method incomplete_sections would return a more rich object but just selecting the sections that are incomplete.

The advantage is that the code becomes cleaner and easy to understand (you won't have to call #send in the view for example) and also move some methods of the component to the rich object.

avinhurry: Done it like this, WDYT? 

https://github.com/DFE-Digital/publish-teacher-training/pull/3697/commits/a11cc1c47c580db96d407ca5a84b667e7ef0003d

tomas-stefano: Much better! I will have only a small tweak. Excellent work mate!

app/components/add_course_button.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+class AddCourseButton < ViewComponent::Base
+  include RecruitmentCycleHelper
+  attr_reader :provider
+
+  Section = Struct.new(:name, :path)
```
tomas-stefano, line 7: I'd add `keyword_init: true` so in the code becomes better to understand when you initialize `Section.new(name: 'aaa', path: 'bbb')

app/components/add_course_button.html.erb
avinhurry: 🙇🏼‍♀️

app/components/add_course_button.rb
avinhurry: Good idea, thanks! I definitely knew you could do that 😅 😉 

https://github.com/DFE-Digital/publish-teacher-training/pull/3697/commits/557ed4a11b9700b21c089a83714007485a61f2c8

tomas-stefano (approved):

PR #3705, opened 2023-07-06
[353] Fix study sites validation bug

avinhurry:
### Context

There is a bug which causes a study site validation to kick in when the course doesn’t have at least one study site attached and other attributes on the course basic details are updated.

- Remove the presenece validation on update from the course.

- Use a form object to validate presence when updating the form. 

### Guidance to review

**Product:** Ask for a screenshare as study sites are only active in the next cycle

**Dev:** Ensure you are in the next cyclea and study sites are activated, then head to a courses basic details page and attempt to update the study site with no sites selected.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/publish/courses/study_sites_controller.rb
+++ b/app/controllers/publish/courses/study_sites_controller.rb
@@ -16,14 +16,16 @@ module Publish
 
       def edit
         authorize(provider)
+
+        @course_study_site_form = CourseStudySiteForm.new(@course)
       end
 
       def update
         authorize(provider)
 
-        update_course_study_sites
+        @course_study_site_form = CourseStudySiteForm.new(@course, params: study_site_params)
 
-        if course.save
+        if @course_study_site_form.save!
           redirect_to details_publish_provider_recruitment_cycle_course_path(
             provider.provider_code,
             recruitment_cycle.year,
@@ -54,6 +56,12 @@ module Publish
 
       def error_keys
         [:study_sites]
+      end
+
+      def study_site_params
+        return { study_site_ids: nil } if params[:publish_course_study_site_form][:study_site_ids].all?(&:empty?)
+
+        params.require(:publish_course_study_site_form).permit(study_site_ids: [])
       end
     end
   end
--- a/app/forms/publish/course_study_site_form.rb
+++ b/app/forms/publish/course_study_site_form.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module Publish
+  class CourseStudySiteForm < BaseCourseForm
+    FIELDS = %i[study_site_ids].freeze
+
+    attr_accessor(*FIELDS)
+
+    validate :no_study_sites_selected
+
+    private
+
+    def compute_fields
+      { study_site_ids: course.study_site_ids }.merge(new_attributes)
+    end
+
+    def no_study_sites_selected
+      return if params[:study_site_ids].present?
+
+      errors.add(:study_site_ids, :blank)
+    end
+  end
+end
--- a/app/forms/publish/course_study_site_form.rb
+++ b/app/forms/publish/course_study_site_form.rb
@@ -6,18 +6,12 @@ module Publish
 
     attr_accessor(*FIELDS)
 
-    validate :no_study_sites_selected
+    validates :study_site_ids, presence: true
 
     private
 
     def compute_fields
       { study_site_ids: course.study_site_ids }.merge(new_attributes)
-    end
-
-    def no_study_sites_selected
-      return if params[:study_site_ids].present?
-
-      errors.add(:study_site_ids, :blank)
     end
   end
 end
--- a/app/views/publish/courses/study_sites/edit.html.erb
+++ b/app/views/publish/courses/study_sites/edit.html.erb
@@ -24,7 +24,7 @@
 
         <%= f.govuk_check_boxes_fieldset :study_site_ids, legend: { text: "#{render CaptionText.new(text: course.name_and_code)} Study sites".html_safe, tag: "h1", size: "l" } do %>
           <div class="govuk-hint">Select all that apply</div>
-          <% @provider.study_sites.sort_by(&:location_name).each_with_index do |study_site, index| %>
+          <% @provider.study_sites.sort_by(&:location_name).each do |study_site| %>
                     <%= f.govuk_check_box :study_site_ids,
                         study_site.id,
                         label: { text: study_site.location_name },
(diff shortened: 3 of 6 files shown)


avinhurry:
[Fix study sites validation bug (next cycle)](https://trello.com/c/bsLI0nHN/353-fix-study-sites-validation-bug-next-cycle)

app/views/publish/courses/study_sites/edit.html.erb
```diff
@@ -22,14 +22,13 @@
...
-        <%= f.govuk_check_boxes_fieldset :study_sites, legend: { text: "#{render CaptionText.new(text: course.name_and_code)} Study sites".html_safe, tag: "h1", size: "l" } do %>
+        <%= f.govuk_check_boxes_fieldset :study_site_ids, legend: { text: "#{render CaptionText.new(text: course.name_and_code)} Study sites".html_safe, tag: "h1", size: "l" } do %>
           <div class="govuk-hint">Select all that apply</div>
-          <% @provider.study_sites.sort_by(&:location_name).each_with_index do |site, index| %>
-                    <%= f.govuk_check_box :study_sites,
-                        site.id,
-                        label: { text: site.location_name },
-                        hint: { text: site.full_address },
-                        checked: f.object.study_sites.include?(site) %>
+          <% @provider.study_sites.sort_by(&:location_name).each_with_index do |study_site, index| %>
```
inulty-dfe, line 27: `index` is not being used here

app/forms/publish/course_study_site_form.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Publish
+  class CourseStudySiteForm < BaseCourseForm
+    FIELDS = %i[study_site_ids].freeze
+
+    attr_accessor(*FIELDS)
+
+    validate :no_study_sites_selected
```
inulty-dfe, line 9: Can't we use `validates :study_site_ids, presence: true`?

app/forms/publish/course_study_site_form.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Publish
+  class CourseStudySiteForm < BaseCourseForm
+    FIELDS = %i[study_site_ids].freeze
```
inulty-dfe, line 5: This seems like overkill. The class name is so specific that it would be weird to have any other fields here. It would necessitate renaming the class.

app/controllers/publish/courses/study_sites_controller.rb
```diff
@@ -55,6 +57,12 @@ def current_step
       def error_keys
         [:study_sites]
       end
+
+      def study_site_params
+        return { study_site_ids: nil } if params[:publish_course_study_site_form][:study_site_ids].all?(&:empty?)
```
inulty-dfe, line 62: Why are we returning `nil` here? `present?` is used to determine if the value is valid or not later and that returns false for `nil` and empty array.

inulty-dfe (changes requested):
I see that a lot of the code is following a pattern from the BaseModelForm. But, this form isn't really a mode. It's only for a specific attribute on a model. Some of the boilerplate seems unnecessary.

Happy to hear your thoughts!

app/views/publish/courses/study_sites/edit.html.erb
avinhurry: Good spot, removed here https://github.com/DFE-Digital/publish-teacher-training/pull/3705/commits/83ee84b24198987d1b1944e5a4e2257663be4420

app/forms/publish/course_study_site_form.rb
avinhurry: Yes, done here https://github.com/DFE-Digital/publish-teacher-training/pull/3705/commits/ec8986de6b21f8947653fcd83062cce30902a15b

app/forms/publish/course_study_site_form.rb
avinhurry: We use this pattern in almost all of our forms on Publish (over 20), we can discuss alternative approaches after rollover when we have more time.

app/controllers/publish/courses/study_sites_controller.rb
avinhurry: I think we need to return nil here now, as I’ve followed your suggestion to use `validates :study_site_ids, presence: true` in the form.

avinhurry:
> I see that a lot of the code is following a pattern from the BaseModelForm. But, this form isn't really a model. It's only for a specific attribute on a model. Some of the boilerplate seems unnecessary.
> 
> Happy to hear your thoughts!

@inulty-dfe this is a pattern we have been following on the Publish and Register services for a while now, see https://github.com/DFE-Digital/register-trainee-teachers/blob/main/app/forms/validate_publish_course_form.rb for example. Happy to discuss alternative approaches after rollover. CC @JR-G

app/controllers/publish/courses/study_sites_controller.rb
inulty-dfe: I looked into this. It appears Rails returns `[""]` for `study_site_ids` when no selection is made. This is probably an issue with http forms where the "key" must have some value in order to be sent in the request.

We could `params.require(:publish_course_study_site_form).permit(study_site_ids: []).compact_blank`.

```ruby
[1] pry(main)> h = { one: 1, nill: nil, empt: "" }
=> {:one=>1, :nill=>nil, :empt=>""}
[2] pry(main)> params = ActionController::Parameters.new(h)
=> #<ActionController::Parameters {"one"=>1, "nill"=>nil, "empt"=>""} permitted: false>
[3] pry(main)> params.compact_blank
=> #<ActionController::Parameters {"one"=>1} permitted: false>
[4] pry(main)> params
=> #<ActionController::Parameters {"one"=>1, "nill"=>nil, "empt"=>""} permitted: false>
[5] pry(main)> params.permit("one", "nill", "empt")
=> #<ActionController::Parameters {"one"=>1, "nill"=>nil, "empt"=>""} permitted: true>
[6] pry(main)> params.permit("one", "nill", "empt").compact_blank
=> #<ActionController::Parameters {"one"=>1} permitted: true>
```

inulty-dfe (approved):
Left a suggestion for `study_site_params` but otherwise approved

app/controllers/publish/courses/study_sites_controller.rb
avinhurry: I tried this, it doesn't return `nil`, so the form doesn't validate when submitting empty. 

Also keeping it as it is helps keep things consistent with what we already have in the codebase, e.g: 

- https://github.com/DFE-Digital/publish-teacher-training/blob/0686327a0d76956c0d157b36a8021987544a7452/app/controllers/publish/courses/schools_controller.rb#L70

- https://github.com/DFE-Digital/publish-teacher-training/blob/0686327a0d76956c0d157b36a8021987544a7452/app/controllers/publish/courses/study_mode_controller.rb#L34
 
- https://github.com/DFE-Digital/publish-teacher-training/blob/0686327a0d76956c0d157b36a8021987544a7452/app/controllers/publish/providers/skilled_worker_visa_controller.rb#L32
 
- https://github.com/DFE-Digital/publish-teacher-training/blob/0686327a0d76956c0d157b36a8021987544a7452/app/controllers/publish/providers/student_visa_controller.rb#L32

PR #3716, opened 2023-07-10
Introduce spec for copy service

JR-G:
### Context

Spec is not being tested


duncanjbrown (approved):
This looks good to me!

PR #3719, opened 2023-07-11
[489] QA AKS add monitoring and logging

smneal:
### Context
Enable monitoring for the QA AKS environment.

- Align the domain resource prefix naming
- Add code in the Makefile to create the monitoring resource group and action group
- Add Terraform code to deploy the monitoring resources

### Guidance to review
- `make qa_aks action-group-resources ACTION_GROUP_EMAIL=notificationemail@domain.com` to create the action group for a given subcription
- `make qa_aks deploy-plan` to see the monitoring and StatusCake tests that will be created

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


terraform/aks/statuscake.tf
```diff
@@ -0,0 +1,8 @@
+module "statuscake" {
+  for_each = var.statuscake_alerts
+
+  source = "./vendor/modules/aks//monitoring/statuscake"
+
+  uptime_urls    = each.value.website_url
+  contact_groups = each.value.contact_group
+}
```
thomasleese, line 8:
```suggestion
module "statuscake" {
  source = "./vendor/modules/aks//monitoring/statuscake"

  uptime_urls    = [module.web_application.probe_url]
  contact_groups = [151103]
}
```

This avoids you needing to set up any variables, for example if the probe path changes in the future you only need to set it once in the web app module.

terraform/aks/locals.tf
```diff
@@ -10,4 +10,5 @@ locals {
     REDIS_WORKER_URL = module.redis_worker.url
   }
   app_env_values = yamldecode(file(var.app_config_file))[var.app_environment]
+  infra_secrets  = yamldecode(data.azurerm_key_vault_secret.infra_secrets.value)
```
thomasleese, line 13: You could probably use the secrets module to simplify this avoid needing the `data` blocks, something like:

```tf 
module "secrets" {
  source = "git::https://github.com/DFE-Digital/terraform-modules.git//aks/secrets?ref=stable"

  azure_resource_prefix = var.azure_resource_prefix
  service_short         = var.service_short
  config_short          = var.config_short
}

infra_secrets  = yamldecode(module.secrets.map[var.key_vault_secret_infra_name)
```

terraform/aks/statuscake.tf
smneal: I've updated the contact groups since they are the same, however the `probe_url` doesn't use the services custom domain so doesn't test the services DNS and front door setup.

thomasleese: You may want both, at least that's how we've handled it in Apply for QTS. It helps to determine where a problem might be, if one is failing and not the other, etc:

https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/d7a0d1011290abf8f3ae2c649e79a412a12cef74/terraform/aks/monitoring.tf#L10-L11

terraform/aks/locals.tf
smneal: Updated as suggested.

terraform/aks/statuscake.tf
smneal: Updated

terraform/aks/variables.tf
```diff
@@ -118,3 +118,9 @@ variable "use_db_setup_command" {
 }
 
 variable "postgres_azure_maintenance_window" { default = null }
+
+# StatusCake variables
+variable "statuscake_alerts" {
+  type    = map(any)
+  default = {}
+}
```
thomasleese, line 126: I think this can go now.

saliceti (approved):

PR #3734, opened 2023-07-14
Use Rails default logger in development environment

inulty-dfe:
### Context

Rails Semantic Logger appears to less useful in development.
Much of the output is dedicated to printing the timestamp and the
server details on each line. This is not particularly useful in
development.

Also, the default logger conveys the seapartion of requests more
distinctly by separating each new http request with a new line.
This is particularly useful when the application triggers over three
consecutive redirects.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -89,9 +89,6 @@ gem 'govuk_notify_rails'
 gem 'sidekiq'
 gem 'sidekiq-cron'
 
-# Semantic Logger makes logs pretty
-gem 'rails_semantic_logger'
-
 # Render nice markdown
 gem 'redcarpet'
 gem 'rubypants'
@@ -150,6 +147,9 @@ gem 'rack-cors'
 
 group :production, :qa, :sandbox, :staging do
   gem 'cloudfront-rails'
+
+  # Semantic Logger makes logs pretty
+  gem 'rails_semantic_logger'
 end
 
 group :development, :test do
--- a/Gemfile
+++ b/Gemfile
@@ -145,7 +145,7 @@ gem 'http'
 # For configuring domains and assets
 gem 'rack-cors'
 
-group :review_aks, :production, :qa, :sandbox, :staging do
+group :production, :qa, :sandbox, :staging do
   gem 'cloudfront-rails'
 
   # Semantic Logger makes logs pretty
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -154,6 +151,9 @@ gem 'rack-cors'
 
 group :production, :qa, :sandbox, :staging do
```
smneal, line 152: @inulty-dfe I have a gut feeling that production_aks, qa_aks, sandbox_aks and staging_aks will also need adding here

inulty-dfe: This is interesting. These environments are only related to the deployment, no? I don't think the application should be aware of them. There shouldn't be any difference between `env` and `env_aks` from a bundling point of view.

smneal: We set the different envs in `config/environments` and settings in `config/settings` and some app config from `terraform/aks/workspace_variables/app_config.yml` so if any of those mean that what that's reading might be different then may be worth a look. I'm not overly familiar with Ruby/Rails/Rack etc.

inulty-dfe: This doesn't feel right from the application perspective. It's bleeding deployment configuration into application code. Functionally there are 3 environments in the Rails application. But the application has to be aware of 7 environment names. There is a `DEPLOY_ENV` in the Makefile. Why do we need to mention `_aks` in the `RAILS_ENV` and `RACK_ENV` environment variables in terraform?

config/application.rb
```diff
@@ -8,7 +8,8 @@
 require 'govuk/components'
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
-Bundler.require(*Rails.groups)
+groups = Rails.env[/_aks/] ? Rails.groups + [:production] : Rails.groups
```
inulty-dfe, line 11: This is a way to load the production gems when the `RAILS_ENV` is set to `_aks`.
Another approach would be to declare a variable in the Gemfile and splat it out as an argument to `group` method. But this requires code change if another `*_aks` is added.

```ruby
aks_envs = %i[production_aks qa_aks sandbox_aks review_aks staging_aks]

group(*aks_envs, :production, :qa, :sandbox, :staging) do
  gem ...
end
```

avinhurry (commented):
Looks like the deploy is failing on the PasS review app. Could it be worth waiting until we've completely migrated away from PaaS to remove the additional layer of complexity?

JR-G:
Removing the deploy label as this is firing 100's of times a day on sentry!

PR #3737, opened 2023-07-17
Filter for discarded providers in public API

inulty-dfe:
### Trello
https://trello.com/c/nR0xM6aq/266-delete-providers-that-have-no-region-code-on-manage-and-dont-exist-on-publish

### Context
Apply/Manage needs a way to identify `Provider`s that have been discarded. The current API filters out discarded providers at the definition of the association between `RecruitmentCycle` and `Provider`. We can `unscope` this exclusion in the API for the purposes of returning discarded providers.

This is a non-breaking change to the API. API consumers do not need to be made aware of this change. It can be an undocumented feature if that is preferred.  There is probably no reason a consumer of the API would need to access discarded `Provider`s

Provider API accepts `filter[discarded]=true` which returns discarded Providers.
When the `filter[discarded]=true` is passed, only providers that are discarded are returned. The attributes include `discarded_at`
Without the filter, `discarded_at` is not returned, keeping the current response consistent.

### Guidance to review
Is this desirable to be added to the Public API?
Could this be made available via another way?
Should this new filter be explicitly documented in public schema?
How should this filter be documented internally?

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/controllers/api/public/v1/providers_controller_spec.rb
+++ b/spec/controllers/api/public/v1/providers_controller_spec.rb
@@ -334,6 +334,15 @@ RSpec.describe API::Public::V1::ProvidersController do
           end
         end
 
+        let(:discarded_provider) do
+          create(:provider,
+                 provider_code: '3wc',
+                 provider_name: 'Discarded',
+                 discarded_at: 1.minute.ago,
+                 organisations: [organisation],
+                 contacts: [contact])
+        end
+
         let(:provider_names_in_response) do
           json_response['data'].map do |provider|
             provider['attributes']['name']
@@ -342,6 +351,7 @@ RSpec.describe API::Public::V1::ProvidersController do
 
         before do
           provider2
+          discarded_provider
 
           get :index, params: {
             recruitment_cycle_year: recruitment_cycle.year,
@@ -402,6 +412,14 @@ RSpec.describe API::Public::V1::ProvidersController do
 
           it "returns 'Second' provider only" do
             expect(provider_names_in_response).to eq([provider2.provider_name])
+          end
+        end
+
+        context 'passing in discarded param' do
+          let(:filter) { { discarded: 'true' } }
+
+          it "returns 'discarded' provider only" do
+            expect(provider_names_in_response).to eq([discarded_provider.provider_name])
           end
         end
       end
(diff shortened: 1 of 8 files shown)


tomas-stefano (approved):
LGTM (after checking the documentation part)

inulty-dfe:
Thanks Avin for launching the review app :taco: 

```shell
curl -H 'content-type: application/json' https://publish-review-3737.test.teacherservices.cloud/api/public/v1/recruitment_cycles/2023/providers?filter%5bdiscarded%5d=true | jq
```

spec/controllers/api/public/v1/providers_controller_spec.rb
```diff
@@ -404,6 +414,23 @@
...
+        context 'passing in discarded param' do
+          let(:filter) { { discarded: 'true' } }
+
+          it "returns 'discarded' provider only" do
+            expect(provider_names_in_response).to eq([discarded_provider.provider_name])
+            expect(Time.zone.parse(response.parsed_body.dig('data', 0, 'attributes', 'discarded_at'))).to be_within(1.second).of(discarded_provider.discarded_at)
+          end
+        end
+
+        context 'not passing in discarded filter' do
```
avinhurry, line 427: Tiny nit but can we keep it consistent? `param` vs `filter` on line 418?

avinhurry (commented):
I’ve tested locally and it works well. Thanks for adding a link in the comments too, very helpful. 

We historically documented all changes made in the API and included them in the release notes. I don’t see any reason to have it as a ‘hidden feature’ but it’s up to @JR-G now. It can be a bit fiddly to work with.

- https://github.com/DFE-Digital/publish-teacher-training/blob/main/docs/source/release-notes.html.md.erb

- https://github.com/DFE-Digital/publish-teacher-training/blob/main/swagger/public_v1/api_spec.json

avinhurry:
> @avinhurry I've updated the OpenAPI spec with the `discarded` filter and the `discarded_at` attributes.

Looks good! @inulty-dfe, should we update the release notes too? https://github.com/DFE-Digital/publish-teacher-training/blob/main/docs/source/release-notes.html.md.erb

avinhurry (approved):

PR #3746, opened 2023-07-19
Fix incomplete accredited provider preview link

avinhurry:
### Context

The incomplete accredited provider preview link was trying to take the user to the edit page of a non existing accredited provider. If the accredited provider does not exist under the AP tab, we should redirect them to add an AP.

image

This is a temporary fix until we resolve our data rot.

Redirect the user to the add accredited provider journey if the accredited provider doesn't exist under the accredited provider tab.

### Guidance to review

- Click the link (as seen in the screenshot above): https://publish-review-3746.test.teacherservices.cloud/publish/organisations/2BY/2024/courses/2XNN/preview

- Should the content of the link be updated to reflect that we are now taking them on an "Add accredited provider" journey?

### DIFF
--- a/app/components/course_preview/missing_information_component.rb
+++ b/app/components/course_preview/missing_information_component.rb
@@ -25,6 +25,10 @@ module CoursePreview
       FeatureService.enabled?(:course_preview_missing_information) && is_preview
     end
 
+    def accrediting_provider_present?(course)
+      course.provider.accredited_providers.include?(course.accrediting_provider)
+    end
+
     private
 
     def about_course_link = about_publish_provider_recruitment_cycle_course_path(provider_code, recruitment_cycle_year, course_code, goto_preview: true)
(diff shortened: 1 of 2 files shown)


app/components/course_preview/missing_information_component.rb
```diff
@@ -36,9 +40,14 @@ def train_with_disability_link = "#{about_publish_provider_recruitment_cycle_pat
     def train_with_us_link = "#{about_publish_provider_recruitment_cycle_path(provider_code, recruitment_cycle_year, course_code:, goto_preview: true)}#train-with-us"
 
     def about_accrediting_provider_link
-      return "#{edit_publish_provider_recruitment_cycle_accredited_provider_path(provider_code, recruitment_cycle_year, accredited_provider_code: @course.accredited_provider_code, goto_preview: true)}#accredited-provider-form-description-field" if FeatureService.enabled?(:accredited_provider_search)
-
-      "#{about_publish_provider_recruitment_cycle_path(provider_code, recruitment_cycle_year, course_code:, goto_preview: true)}#accrediting-provider-#{accrediting_provider.provider_code}"
+      if FeatureService.enabled?(:accredited_provider_search) && accrediting_provider_present?(course)
+        "#{edit_publish_provider_recruitment_cycle_accredited_provider_path(provider_code, recruitment_cycle_year, accredited_provider_code: @course.accredited_provider_code, goto_preview: true)}#accredited-provider-form-description-field"
```
inulty-dfe, line 44: You can use the `anchor` option here instead of interpolating the whole URL. Similar for above and below.

```suggestion
        edit_publish_provider_recruitment_cycle_accredited_provider_path(provider_code, recruitment_cycle_year, accredited_provider_code: @course.accredited_provider_code, goto_preview: true, anchor: 'accredited-provider-form-description-field')
```

app/components/course_preview/missing_information_component.rb
```diff
@@ -36,9 +40,14 @@ def train_with_disability_link = "#{about_publish_provider_recruitment_cycle_pat
...
+      elsif FeatureService.enabled?(:accredited_provider_search) && !accrediting_provider_present?(course)
+        publish_provider_recruitment_cycle_accredited_providers_path(provider_code,
```
inulty-dfe, line 46: This link brings the user to a place they can add an accredited provider to the courses provider. What if it's the other way around? What if the course doesn't have the correct accrediting provider but the courses provider does?

inulty-dfe (commented):
Question around handling another case. Maybe it cannot arrive at this point?

app/components/course_preview/missing_information_component.rb
avinhurry: Good question. The intention of the link here is to demonstrate to the provider that they have not filled out information about the accredited provider. The link takes them to a place where they can add this information. Ideally, we would take them to the edit page, but because this is a new section, many providers haven't populated it yet and so the edit page doesn't exist. This is a work around.

When the provider adds the accredited provider in that joruney and inputs the "about accredited provider" info (which is mandatory). That info would then display on the preview page in place of the link.

app/components/course_preview/missing_information_component.rb
avinhurry: Nice one, that's a lot better! 

https://github.com/DFE-Digital/publish-teacher-training/pull/3746/commits/5e68903d18b7865680824cb8abfdaf429c8eaa37

inulty-dfe (approved):
The situation I asked about with course provider/accredited_provider is complicated and not the scope of this change.

PR #3747, opened 2023-07-20
Invalidate incomplete course on publish

inulty-dfe:
### Trello
[Validate a course has an accrediting provider](https://trello.com/c/Ox93dbSJ/393-itt-reform-prevent-publication-of-an-incomplete-course)
[Users can add accredited provider to their account when selecting an accredited provider for a course](https://trello.com/c/YsMU17Nu/404-when-publish-users-click-change-links-for-study-site-or-accredited-provider-in-course-description-and-they-only-have-1-surface-a)

### Review App

**Course's accredited provider is no longer accredited.**
https://publish-review-3747.test.teacherservices.cloud/publish/organisations/1JE/2024/courses/3FNB
**Course has no accredited provider associated**
https://publish-review-3747.test.teacherservices.cloud/publish/organisations/1JE/2024/courses/2B76 (edited) 

### Context

When a course is published, the course must:
 - have an accrediting provider associated. 
 - the accrediting provider must be validated as accredited

1. When the accrediting provider doesn't exist, we direct the user to choose an accredited provider from the providers that are associated with the courses training provider.
2. When the associated accrediting provider is not in fact accredited, we direct the user to add a new accredited provider to
their provider account. This is based on the assumption that a provider will likely need to do this.

Add validations to the Course model to ensure that the Accrediting Provider is in a valid state for :publish

### Guidance to review

The PR only focuses on publishing a course. Some of the dependent data states offer problems. This PR does not address the fact that a training provider may be associated with accredited providers that are no longer accredited. Publishing a course assumes the accredited providers associated with the training provider are valid and correct. The course will not publish with an invalid accredited provider but the user **_can_** assign an unaccredited provider as the courses accredited provider.

I believe the `enrichment_error_url` method needs refactoring, along with many of the `ViewHelper` methods.

### UPDATE

Added [Users can add accredited provider to their account when selecting an accredited provider for a course](https://trello.com/c/YsMU17Nu/404-when-publish-users-click-change-links-for-study-site-or-accredited-provider-in-course-description-and-they-only-have-1-surface-a) ticket into this PR out of necesity.

1. When publishing a course, and the course has no accrediting provider
2. And the training provider has no accredited providers
3. Then we need to fix both these issues.

We cannot assign an accrediting provider to the course until we add an accredited provider to the training provider.

Error messages that appear:

-   if the course has an accredited provider but it is not accredited, we prompt the user to "**Update the accredited provider (it's no longer accredited)**"
-   if the course has no accredited provider we "**Select an accredited provider**"

Both these errors link to the same page. The Choose an accredited provider from the list of accredited providers associated with the training provider.

This list could be empty.

In any case, there is a link that allows the training provider to add a new accredited provider to their account. After doing this they will be able to choose a valid accredited provider.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -61,7 +61,8 @@ module ViewHelper
         required_qualifications: "#{base}/requirements?display_errors=true#required_qualifications_wrapper",
         age_range_in_years: "#{base}/age-range?display_errors=true",
         sites: "#{base}/schools?display_errors=true",
-        study_sites: (course.provider&.study_sites&.none? ? "#{provider_base}/study-sites" : "#{base}/study-sites").to_s
+        study_sites: (course.provider&.study_sites&.none? ? "#{provider_base}/study-sites" : "#{base}/study-sites").to_s,
+        accrediting_provider: accredited_provider_publish_provider_recruitment_cycle_course_path(course.provider_code, course.recruitment_cycle_year, course.course_code)
       }.with_indifferent_access[field]
     end
   end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -322,6 +322,7 @@ class Course < ApplicationRecord
   validates :sites, presence: true, on: %i[publish new]
   validates :study_sites, presence: true, on: %i[publish new], if: -> { recruitment_cycle_after_2023? }
   validates :subjects, presence: true, on: :publish
+  validates :accrediting_provider, presence: true, on: :publish
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_site_statuses_publishable, on: :publish
   validate :validate_provider_visa_sponsorship_publishable, on: :publish, if: -> { recruitment_cycle_after_2021? }
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -327,6 +327,7 @@ class Course < ApplicationRecord
   validate :validate_site_statuses_publishable, on: :publish
   validate :validate_provider_visa_sponsorship_publishable, on: :publish, if: -> { recruitment_cycle_after_2021? }
   validate :validate_provider_urn_ukprn_publishable, on: :publish, if: -> { recruitment_cycle_after_2021? }
+  validate :validate_accredited_provider_is_accredited, on: :publish
   validate :validate_degree_requirements_publishable, on: :publish
   validate :validate_gcse_requirements_publishable, on: :publish
   validate :validate_enrichment
@@ -902,6 +903,12 @@ class Course < ApplicationRecord
 
     latest_enrichment.valid?
     add_enrichment_errors(latest_enrichment)
+  end
+
+  def validate_accredited_provider_is_accredited
+    return if accrediting_provider.blank?
+
+    accrediting_provider.accredited? || errors.add(:accrediting_provider, :is_not_accredited)
   end
 
   def validate_enrichment_publishable
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -69,6 +69,7 @@ class Provider < ApplicationRecord
   end
 
   alias accrediting_providers accredited_providers
+  alias accredited? accredited_provider?
 
   delegate :year, to: :recruitment_cycle, prefix: true
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -499,6 +499,7 @@ en:
               blank: "^Select an applications open date"
             accrediting_provider:
               blank: "Select an accrediting provider"
+              is_not_accredited: "The assigned accrediting provider is not accredited"
             base:
               duplicate: "This course already exists. You should add further schools for this course to the existing profile in Publish"
               visa_sponsorship_not_publishable: "Select if visas can be sponsored"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -499,7 +499,7 @@ en:
               blank: "^Select an applications open date"
             accrediting_provider:
               blank: "Select an accrediting provider"
-              is_not_accredited: "The assigned accrediting provider is not accredited"
+              is_not_accredited: "Update the accredited provider (it's no longer accredited)"
             base:
               duplicate: "This course already exists. You should add further schools for this course to the existing profile in Publish"
               visa_sponsorship_not_publishable: "Select if visas can be sponsored"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -497,6 +497,8 @@ en:
               blank: "^Select a study mode"
             applications_open_from:
               blank: "^Select an applications open date"
+            accrediting_provider:
+              blank: "Select an accrediting provider"
             base:
               duplicate: "This course already exists. You should add further schools for this course to the existing profile in Publish"
               visa_sponsorship_not_publishable: "Select if visas can be sponsored"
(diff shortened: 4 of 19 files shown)


inulty-dfe:
### Screenshots

|No accredited provider| AP is not accredited|
|---|---|
|Screenshot from 2023-07-20 16-51-24|Screenshot from 2023-07-20 16-57-00|

app/models/course.rb
```diff
@@ -903,6 +905,12 @@ def validate_enrichment
     add_enrichment_errors(latest_enrichment)
   end
 
+  def validate_accredited_provider_is_accredited
+    return if accrediting_provider.blank?
```
inulty-dfe, line 909: This is handled by the `presence` validation

MaeveRoseveare:
@MylesJarvis yes I see! Should we help them rectify the issue, maybe:

Change the accredited provider for this course, the one you selected in no longer accredited

And link preferable goes to the page to update the AP.

MylesJarvis:
> @MylesJarvis yes I see! Should we help them rectify the issue, maybe:
> 
> Change the accredited provider for this course, the one you selected is no longer accredited
> 
> And link preferably goes to the page to update the AP.

I'm happy with this. @inulty-dfe let's go with Maeve's suggestion for that validation error.

inulty-dfe:
I've updated the screenshots to show the flow.

https://github.com/DFE-Digital/publish-teacher-training/pull/3747#issuecomment-1644237032

app/helpers/view_helper.rb
```diff
@@ -61,11 +61,23 @@ def enrichment_error_url(provider_code:, course:, field:, message: nil)
         required_qualifications: "#{base}/requirements?display_errors=true#required_qualifications_wrapper",
         age_range_in_years: "#{base}/age-range?display_errors=true",
         sites: "#{base}/schools?display_errors=true",
-        study_sites: (course.provider&.study_sites&.none? ? "#{provider_base}/study-sites" : "#{base}/study-sites").to_s
+        study_sites: (course.provider&.study_sites&.none? ? "#{provider_base}/study-sites" : "#{base}/study-sites").to_s,
+        accrediting_provider: accrediting_provider_error_links(message, course)
```
avinhurry, line 65: Spotted a quirk with this where if there are no accredited providers under the provider, the validation error takes you to the update page. There is no journey for this currently as there is nothing for the user to choose. The error should take you to the page to add a new accredited provider and the error message wording should reflect that too.

I think `study_sites` above is doing something similar, I know the whole method could do with a refactor but maybe we could replicate what we've done there for now to help with readability of the method as a whole?

validation link should be dynamiic

app/models/course.rb
```diff
@@ -322,10 +322,12 @@ def self.entry_requirement_options_without_nil_choice
   validates :sites, presence: true, on: %i[publish new]
   validates :study_sites, presence: true, on: %i[publish new], if: -> { recruitment_cycle_after_2023? }
   validates :subjects, presence: true, on: :publish
+  validates :accrediting_provider, presence: true, on: :publish
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_site_statuses_publishable, on: :publish
   validate :validate_provider_visa_sponsorship_publishable, on: :publish, if: -> { recruitment_cycle_after_2021? }
   validate :validate_provider_urn_ukprn_publishable, on: :publish, if: -> { recruitment_cycle_after_2021? }
+  validate :validate_accredited_provider_is_accredited, on: :publish
```
avinhurry, line 330: The behaviour of the validation works well, but I have a product question on where the link should take the user.

@MylesJarvis @Kateharries if a user tries to publish a course with an accredited provider that has lost it's accreditation, do we want to take the user to the page where they can add a completely new accredited provider, or do we want to take them to a page where they can update the course with an accredited provider that they already have? The content of the link implies to me that we are taking them to update rather than add.

image

app/models/provider.rb
```diff
@@ -69,6 +69,7 @@ def accredited_providers
   end
 
   alias accrediting_providers accredited_providers
+  alias accredited? accredited_provider?
```
avinhurry, line 72: It is so confusing isn't it! `accredited_provider`, `accredited`, `accrediting_provider`, `ratifying_provider` 🤦🏼‍♀️

avinhurry (commented):
A couple of quirks/questions in line

app/models/course.rb
MaeveRoseveare: This is my fault! I did my best in Kate's absence on Friday. But clearly, I don't know how Publish works!! I will wait for @Kateharries to advise. Sorry for causing confusion!

app/models/provider.rb
inulty-dfe: I added this cause it's `accredited_provider.accredited_provider?` Redundancy. It should be obvious that we're calling `#accredited?` on a provider.

app/models/course.rb
inulty-dfe: `if a user tries to publish a course with an accredited provider that has lost it's accreditation`

This should never happen. It's a bug downstream. We shouldn't actually be handling this here.

MylesJarvis:
One more thing from me, which is linked to @avinhurry's comment and gif. In his gif, he has AA Teamworks as a Lead Partner, so can add an Accredited Provider (AP) in the tab. Whereas in the preview app and whatever database is associated there, AA Teamworks are listed as an Accredited Provider.

When a provider is an AP, they self-accredit, which means that they should have an accredited_provider value of `nil`. I think the knock on impact of validating for an AP that is approved is that we're now checking every course for this.

AP erroring when no AP

We will need to scope this check in this PR to lead partners only, as the solution, I think?

app/models/course.rb
Kateharries: I think (for now) we should take them to the page where they need to add a new AP. Content should read: Add an accredited provider.

When they get to the new page, can we have some content on this page especially for this scenario? It could read:

[AP name] is no longer an accredited provider. You'll need to add another accredited provider and then go to the 'Basic details' tab in your course page to attach the new accredited provider to the course.

app/helpers/view_helper.rb
Kateharries: The content in the column next to 'Accredited provider' should read 'Select an accredited provider'.

JR-G:
I've hit an error – not sure if this is review app behaviour. 

Using [this example](https://publish-review-3747.test.teacherservices.cloud/publish/organisations/1JE/2024/courses/2B76/publish), I click 'Select an accrediting provider' and then without selecting the radio button I press 'Update accredited provider' and it errors. 

Just flagging now and I can try and dig into it myself. I also appreciate this is an edge case (if it is genuine it's not an ideal user experience but they can go back and just select it) and this is important so I'm happy to consider this not a blocker.

config/locales/en.yml
```diff
@@ -497,6 +497,9 @@ en:
               blank: "^Select a study mode"
             applications_open_from:
               blank: "^Select an applications open date"
+            accrediting_provider:
+              blank: "Select an accrediting provider"
+              is_not_accredited: "Update the accredited provider (it's no longer accredited)"
```
JR-G, line 502: deferring to @Kateharries & @MaeveRoseveare  on this one – had a look through [gov error messaging](https://design-system.service.gov.uk/components/error-message) do we need to slightly tweak the validation message?

Kateharries: @JR-G @MaeveRoseveare @MylesJarvis I think the message should just say 'Update the accredited provider' and then we have content that appears on the 'Accredited providers' page just for this scenario (see my other comment).

JR-G (approved):

config/locales/en.yml
MylesJarvis: Me and @Kateharries have agreed to remove the bit in brackets. If we've missed the boat on this PR as it's been approved by James, I'll do it immediately after we ship this.

MylesJarvis: We also discussed some improved copy changes we can make on this, that we will ticket up separately.

PR #3763, opened 2023-07-26
Add "Add study site" button on provider study site index

inulty-dfe:
### Trello
https://trello.com/c/fNfLmXTS/429-when-publish-users-click-change-links-for-study-site-in-course-description-and-they-only-have-1-surface-a-link-button-to-allow-t

### Context

At least one study site is required to be associated with a course before it can be published. Often a course will not have a  study site associated with it. 

On the screen where the provider can choose multiple study sites from the list of study sites associated with their provider account. It's likely that while choosing a study site for the course they are publishing, they will want to add a study site that does n't already associate with their provider.

This PR adds a convenience "Add study site" button to the Course study site choice page.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/publish/courses/study_sites/edit.html.erb
+++ b/app/views/publish/courses/study_sites/edit.html.erb
@@ -31,13 +31,15 @@
                         hint: { text: study_site.full_address } %>
           <% end %>
         <% end %>
-        <%= f.govuk_submit "Update study sites" %>
-        <%= govuk_button_link_to(
-          "Add study site",
-          publish_provider_recruitment_cycle_study_sites_path(course.provider_code, course.recruitment_cycle_year),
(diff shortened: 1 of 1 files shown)


avinhurry (commented):
@MaeveRoseveare you mentioned that the correct pattern is a link rather than a grey button. Should the link take you to the index page (the page where the grey button in this PR takes you) or to the study sites search page. If it takes you to the index page, it will mean that the user will have to click "Add study sites" once on the update page and then again on the index page. 

I feel like it might make sense to take them straight to the search page so they only have to click once, but keen to get your thoughts?

avinhurry (approved):

PR #3769, opened 2023-07-28
[414] Update rollover script to handle the correct `application_open_from` date

avinhurry:
### Context

Update rollover script to handle the correct `application_start_date`.

If the course `applications_open_from` date is the same as the current cycle `application_start_date` set the next recruitment cycle `application_start_date` as the course `applications_open_from` date 🤯 🤯 🤯 

### Guidance to review

You can try it with manual rollover in the review app. It should be easy to find a course with a `applications_open_from` date as the 11th of October. Currently this would be rolled over to the 11th of October the following year. With this change it be the 10th of October when rolled over.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/copy_to_provider_service.rb
+++ b/app/services/courses/copy_to_provider_service.rb
@@ -55,7 +55,7 @@ module Courses
       return course.applications_open_from if next_cycle.blank?
 
       if course_start_is_same_as_current_cycle_start?(course)
-        next_cycle.application_start_date
+        Find::CycleTimetable.apply_reopens.to_date
       else
         [course.applications_open_from + year_differential.year, next_cycle.application_start_date].max
       end
--- a/app/services/courses/copy_to_provider_service.rb
+++ b/app/services/courses/copy_to_provider_service.rb
@@ -54,7 +54,11 @@ module Courses
 
       return course.applications_open_from if next_cycle.blank?
 
-      [course.applications_open_from + year_differential.year, next_cycle.application_start_date].max
+      if course_start_is_same_as_current_cycle_start?(course)
+        next_cycle.application_start_date
+      else
+        [course.applications_open_from + year_differential.year, next_cycle.application_start_date].max
+      end
     end
 
     def copy_schools(course:, new_provider:, new_course:)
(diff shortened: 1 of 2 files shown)


avinhurry:
[Application open date on rolled over courses is incorrect](https://trello.com/c/0HiO6b7O/414-application-open-date-on-rolled-over-courses-is-incorrect)

app/services/courses/copy_to_provider_service.rb
```diff
@@ -65,6 +69,10 @@ def copy_schools(course:, new_provider:, new_course:)
       end
     end
 
+    def course_start_is_same_as_current_cycle_start?(course)
+      course.applications_open_from == current_cycle.application_start_date
```
JR-G, line 73: `current_cycle` is `RecruitmentCycle.current_recruitment_cycle` which is then set in `Settings` right?

Would it be possible to use `CycleTimetable.current_year` which would align us with Apply?

Could we then use the [apply_reopens](https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/services/find/cycle_timetable.rb#L101) method to set the date?

Appreciate I might not fully grasp the complexities of Publish course dates but happy to chat!

avinhurry: Yes that’s right.

It is possible, but it will add a lot of complexity and I’m not sure the benefits we gain from it will be worth it. The date is set manually in the cycle time table, the same way it is set in the recruitment cycle. 

What’s stopping us from going all the way and setting the `application_start_date` from the cycle time table when we create the recruitment cycle? That way we only need to keep track of and update the one date in the cycle time table. Even if we do this we would still need to consider rolling over courses where the `applications_open_from` date is different to the `application_start_date` as we would want to retain these.

My gut would be to merge this as it solves the problem at hand and add a card to the backlog to re-think the approach when we have more time but also happy to continue working on it.

avinhurry: Had a chat offline and decided to use the cycle switcher for now and consider alternative joined up approaches between the services in the future.

PR #3776, opened 2023-08-02
Freeze cycle switcher specs

JR-G:
### Context

Builds are failing:
image
This is because these tests are not frozen in time. The Apply 1 deadline banner has just gone live on production so these tests are failing because we never render the Apply 2 deadline banner due to the first branch of the `if` statement returning `true`.

Freeze the test at the start of the 2022 cycle.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/spec/features/find/switcher_cycles_spec.rb
+++ b/spec/features/find/switcher_cycles_spec.rb
@@ -3,6 +3,10 @@
 require 'rails_helper'
 
 feature 'switcher cycle' do
+  before do
+    Timecop.freeze(2022, 10, 12)
+  end
+
   scenario 'Navigate to /cycle' do
     when_i_visit_switcher_cycle_page
     then_i_should_see_the_page_title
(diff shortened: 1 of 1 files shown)


inulty-dfe (approved):

spec/features/find/switcher_cycles_spec.rb
```diff
@@ -94,6 +98,6 @@ def then_i_should_see_the_applications_closed_text
   end
 
   def and_i_should_see_the_correct_previous_recruitment_cycle_year
-    expect(page).to have_text("Previous cycle year#{RecruitmentCycle.current.year}") # After find reopens, the previous cycle year for the fake cycle becomes the current cycle year for the real cycle
+    expect(page).to have_text('Previous cycle year2023') # After find reopens, the previous cycle year for the fake cycle becomes the current cycle year for the real cycle
```
avinhurry, line 101: Isn't this going to fail next year?

avinhurry: Oh I see, you've frozen time so that will always be the year?

avinhurry: Yeah we should tidy those up too. I feel like it might not be better to freeze time, but lets just get this in for now!

avinhurry (approved):

PR #3783, opened 2023-08-09
[412] Populate accredited provider list based on recruitment cycle

avinhurry:
### Context

The list of accredited providers is not recruitment cycle dependant. We need to make it dependant so that if an accredited provider in the current cycle is not an accredited provider in the next cycle, they are not seen in the list in the next cycle. 

More importantly, we need to ensure that providers who wern't accredited providers in the current cycle but are in the next cycle, are shown in this list - as this would prevent providers from adding and publishing courses. 

Pass in the recruitment cycle and generate the list of accredited providers based on the recruitment cycle.

### Guidance to review

- Go to a provider which is not an accredited provider in both cycles
- Check that they are not in the list when adding an accredited provider
- Change them to an accredited provider in one cycle
- Check that they are now in the list but only for the cycle that you changed it. 

- Ensure the accredited provider flow still functions as expected. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/accredited_provider_suggestions_controller.rb
+++ b/app/controllers/api/accredited_provider_suggestions_controller.rb
@@ -5,7 +5,7 @@ module API
     def index
       return render_json_error(status: 400, message: I18n.t('accredited_provider_suggestion.errors.bad_request')) if invalid_query?
 
-      accredited_providers = AccreditedProviders::SearchService.call(query: params[:query]).providers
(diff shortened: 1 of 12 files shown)


avinhurry:
[ITT Reform: Ensure that the Accredited Provider autocomplete populates based on the 'is the organisation an accredited provider?' setting](https://trello.com/c/Tsh9B5SY/412-itt-reform-ensure-that-the-accredited-provider-autocomplete-populates-based-on-the-is-the-organisation-an-accredited-provider-se)

JR-G (approved):

PR #3802, opened 2023-08-25
[549] Update deadline banners/pages and fix cycle switcher

avinhurry:
### Context

Fixing the cycle switcher and updating our content on the banners and pages in line with multiple applications.

The cycle switcher was not working once we were in one of the "between" periods. The if statement would always return true for the first conditional because the second part of this "or" statement would always be true.

`current_cycle_schedule == :today_is_after_find_opens || Time.zone.now.between?(find_opens, apply_2_deadline)`

- Fix cycle switcher
- Update content on Find banners
- Update content on "Applications closed" page

### Guidance to review

image

image

image

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/find/deadline_banner_component.html.erb
+++ b/app/components/find/deadline_banner_component.html.erb
@@ -19,7 +19,7 @@
   <%= govuk_notification_banner(title_text: "Important") do |notification_banner| %>
     <% notification_banner.with_heading(text: "Courses are currently closed but you can get your application ready") %>
     <p class="govuk-body">Courses starting in the <%= cycle_timetable.cycle_year_range %> academic year are closed.</p>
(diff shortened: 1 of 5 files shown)


avinhurry:
[Update deadline banners and ensure they work](https://trello.com/c/XaNtdLv3/549-update-deadline-banners-and-ensure-they-work)

JR-G:
Just tested this and it's working but it's highlighted to me that you can access https://www.find-postgraduate-teacher-training.service.gov.uk/cycle-has-ended at any point in the cycle. Could we redirect away from this if the cycle isn't closed?

avinhurry:
> Just tested this and it's working but it's highlighted to me that you can access https://www.find-postgraduate-teacher-training.service.gov.uk/cycle-has-ended at any point in the cycle. Could we redirect away from this if the cycle isn't closed?

Good shout @JR-G

PR #3826, opened 2023-09-06
[633] Remove visa validation from provider level

avinhurry:
### Context

As part of an effort to increase the number of published courses, we are fixing a bug which prevents providers from publishing courses if `can_sponsor_student_visa` or `can_sponsor_skilled_worker_visa` is `nil` on the provider.

We have moved these to the course level and should validate on that level instead, as the provider has no way of updating on the provider level anymore.

- Remove the validation from the provider level
- Remove the method as it's no longer used

### Guidance to review

Try publishing the course `1I2` `V655` `2023` on QA and compare with the review app. You'll only be able to do this once, so be careful! 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -324,7 +324,6 @@ class Course < ApplicationRecord
   validates :accrediting_provider, presence: true, on: :publish, unless: -> { self_accredited? }
   validate :validate_enrichment_publishable, on: :publish
   validate :validate_site_statuses_publishable, on: :publish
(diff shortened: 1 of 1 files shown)


avinhurry:
[Move visa sponsorship validation to course level](https://trello.com/c/UhIPqbKp/633-move-visa-sponsorship-validation-to-course-level)

JR-G:
Let's close this for now and can reopen if we decide we need to amend the validation cc @MylesJarvis

PR #3861, opened 2023-10-01
Set a default value of false for the two visa sponsorship columns

JR-G:
### Context

We don't set a value in these two fields when we onboard a new provider, yet we are validating for the presence of these values in order for the provider to be able to publish. We also offer no way of the provider being able to set these values themselves. This is preventing publication of courses, without developer intervention 

### Guidance to review

Backfill to follow

### Link to trello ticket

https://trello.com/c/sjx3JbKL/770-update-the-publish-data-schema-to-set-cansponsorskilledvisas-and-cansponsorstudentvisa-to-false-upon-provider-account-creation

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


db/schema.rb
```diff
@@ -24,6 +24,21 @@
     t.string "ProductVersion", limit: 32, null: false
   end
 
+  create_table "access_request", id: :serial, force: :cascade do |t|
```
JR-G, line 27: We have a migration called `DeleteAccessRequestTable` that is creating the table 🤔

defong (approved):

PR #3880, opened 2023-10-06
[2282] Fix provider name mapping for autocomplete

darokel:
### Context

https://trello.com/c/ACtolcvq/2282-mulberry-college-of-teaching-does-not-show-on-find

Some providers could have leading/trailing whitespace in their names which breaks the autocomplete from setting them correctly. Here is an example of how the option is received by the js logic. 

Screenshot 2023-10-06 at 16 31 58

It fails the comparison check here https://github.com/DFE-Digital/publish-teacher-training/blob/422241cf81c6d2ff6eb3fe57d2efa3c5f3ecc9c9/app/javascript/find/dfe-autocomplete/dfe-autocomplete.js#L48

- Strip from source 

### Guidance to review

You can re-produce the issue by searching `5V5` in the Find provider lookup. If you try to go to the next step it'll trigger an error. Try this in the review app and it should be fixed.

### DIFF
--- a/app/helpers/find/provider_helper.rb
+++ b/app/helpers/find/provider_helper.rb
@@ -9,7 +9,7 @@ module Find
           'data-boost' => 1.5
         }
 
-        value = provider.provider_name
+        value = provider.provider_name.strip
         name = "#{value} (#{provider.provider_code})"
 
         [name, value, data]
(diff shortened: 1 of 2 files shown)


darokel:
[Mulberry College of Teaching does not show on Find](https://trello.com/c/ACtolcvq/2282-mulberry-college-of-teaching-does-not-show-on-find)

JR-G (approved):

PR #3918, opened 2023-11-06
Update width on provider index support page

avinhurry:
### Context

Update the width for a more balanced look, since the introduction of the UKPRN column.

#### Before

image

#### After

image


lsalivonaite-transform (approved):
Perfect!

JR-G (approved):

PR #3927, opened 2023-11-09
Add `make shell` command to Publish

JR-G:
### Context
We have a make shell command on Apply, and it makes life a bit easier if we need to do something in the console. We should do this on Publish!

Added additional steps in our `MAKEFILE`.

### Guidance to review
`make review_aks aks-console APP_NAME=3734 USE_DB_SETUP_COMMAND=false`

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


vassyz:
What's the diffrence between the `aks-console` and `shell` commands?

PR #3940, opened 2023-11-20
Fix CI

avinhurry:
### Context

CI is broken, this is the fix.


defong (approved):
Great work.. we are sick of 0.0.0

PR #3955, opened 2023-12-04
Update support_playbook.md

inulty-dfe:
### Context
`RecruitmentCycle.current.next` is not valid. Remove `next`

### DIFF
--- a/guides/support_playbook.md
+++ b/guides/support_playbook.md
@@ -114,10 +114,10 @@ The example below copies scheduled courses during rollover from one provider to 
 
 ```ruby
 # Find provider to copy courses from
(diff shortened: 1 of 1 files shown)


guides/support_playbook.md
```diff
@@ -114,10 +114,10 @@ The example below copies scheduled courses during rollover from one provider to
 
 ```ruby
 # Find provider to copy courses from
-provider = RecruitmentCycle.current.next.providers.find_by(provider_code: "1TZ")
```
avinhurry, line 117: This was probably written when the two cycles are open (during rollover) when there would be a next recruitment cycle.

avinhurry (approved):
💯

PR #3983, opened 2024-01-02
Enable entity_table_check_job

ericaporter:
### Context
EntityTableCheck is a job which sends a row_count, a checksum and a timestamp for each entity table sent to Big Query and it’s purpose is to serve as a tool for confirming database tables match Big Query tables and then help to debug where they don’t match.

This PR enables the job and schedules it to run daily at 00:30

### Guidance to review
Please check the scheduling looks correct and advise if there is a better time for the job to run. I went for low_priority as the queue and enabled it only in the production.yml.

N.B If it is possible, testing this PR prior to deployment would be ideal - I can help with that

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -138,7 +138,7 @@ gem 'jsbundling-rails', '~> 1.2'
 gem 'sprockets-rails', require: 'sprockets/railtie'
 
 # for sending analytics data to the analytics platform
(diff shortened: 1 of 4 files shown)


vassyz (approved):
Running fine on the review app. 👍

ericaporter:
Data populating in Big Query as expected

PR #4003, opened 2024-01-15
Refactor references to provider codes in view templates

inulty-dfe:
### Context

Remove explicit references to Provider and Course codes in view templates.

We are extracting the logic for storing configuration related to conditionally showing aspects of the views based on the provider and course.

Create a config file for what providers and courses should trigger the alternative view content.

`config/course_information.yml` holds the data mapping configurations for paricular `Provider`s, `provider_type`s `Course` `program_type`s

`Config::CourseInformation` is a class that encapsulates the loading and lookup behaviour providing an abstracted public interface for interacting with the config data. It is used throughout the application where provider / course specific checks need to happen.

### Question / Proposal
I would like to create a class that encapsulates the views config database. Something like `ViewConfig`.

What should we name it and where should it live in the directory structure?

### Trello 

[Trello Ticket](https://trello.com/c/HlLZsoz2/1141-find-remove-reference-to-provider-codes)

#### Course Information Edit page

**Shows Guidance (SCITT programme but not E65)**
 - https://publish-review-4003.test.teacherservices.cloud/publish/organisations/24S/2024/courses/3D8V/about

**Doesn't show Guidance (SCITT programme E65)**
 - https://publish-review-4003.test.teacherservices.cloud/publish/organisations/E65/2024/courses/38NP/about

### Links - Find

**Shows Guidance (University Provider but not B31)**
- https://find-review-4003.test.teacherservices.cloud/course/U80/2P35

**Shows Guidance (SCITT programme but not E65)**
 - https://find-review-4003.test.teacherservices.cloud/course/24S/3D8V

**no show guidance (scitt programme E65)**
 - Doesn't show guidance : https://find-review-4003.test.teacherservices.cloud/course/E65/38np

### Links Address

**X104 course**
- https://find-review-4003.test.teacherservices.cloud/course/28T/X104

**non X104 course** 
- https://find-review-4003.test.teacherservices.cloud/course/28T/368R

### Links Contact Form

**Contact Form (U80 provider)**
- https://find-review-4003.test.teacherservices.cloud/course/U80/2P35

**Email (non U80 provider)**
- https://find-review-4003.test.teacherservices.cloud/course/2Z3/B275

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/components/find/courses/about_schools_component/view.rb
+++ b/app/components/find/courses/about_schools_component/view.rb
@@ -29,13 +29,13 @@ module Find
         def show_higher_education_guidance?
           return false unless course.higher_education_programme?
 
-          course_information_config.show_placement_guidance?(:program_type, :higher_education)
+          course_information_config.show_placement_guidance?(:program_type)
         end
 
         def show_scitt_guidance?
           return false unless course.scitt_programme?
 
-          course_information_config.show_placement_guidance?(:program_type, :scitt_programmes)
+          course_information_config.show_placement_guidance?(:program_type)
         end
 
         private
--- a/app/components/find/courses/about_schools_component/view.rb
+++ b/app/components/find/courses/about_schools_component/view.rb
@@ -29,21 +29,19 @@ module Find
         def show_higher_education_guidance?
           return false unless course.higher_education_programme?
 
-          course_information_config(:higher_education, :except_provider_codes).exclude?(course.provider.provider_code)
+          course_information_config.placement(:program_type, :higher_education)
         end
 
         def show_scitt_guidance?
           return false unless course.scitt_programme?
 
-          course_information_config(:scitt_programmes, :except_provider_codes).exclude?(course.provider_code)
+          course_information_config.placement(:program_type, :scitt_programmes)
         end
 
         private
 
-        def course_information_config(*path)
-          @course_information_config ||= Rails.application.config_for(:course_information)
-
-          @course_information_config.dig(:where_you_will_train, *path)
+        def course_information_config
+          @course_information_config ||= Configs::CourseInformation.new(course)
         end
       end
     end
--- a/app/components/find/courses/about_schools_component/view.rb
+++ b/app/components/find/courses/about_schools_component/view.rb
@@ -29,13 +29,13 @@ module Find
         def show_higher_education_guidance?
           return false unless course.higher_education_programme?
 
-          course_information_config.placement(:program_type, :higher_education)
+          course_information_config.show_placement_guidance?(:program_type, :higher_education)
         end
 
         def show_scitt_guidance?
           return false unless course.scitt_programme?
 
-          course_information_config.placement(:program_type, :scitt_programmes)
+          course_information_config.show_placement_guidance?(:program_type, :scitt_programmes)
         end
 
         private
--- a/app/components/find/courses/about_schools_component/view.rb
+++ b/app/components/find/courses/about_schools_component/view.rb
@@ -21,8 +21,7 @@ module Find
 
         def render?
           published_how_school_placements_work.present? ||
-            program_type == 'higher_education_programme' ||
-            program_type == 'scitt_programme' ||
+            program_type.in?(%w[higher_education_programme scitt_programme]) ||
             study_sites.any? ||
             site_statuses.map(&:site).uniq.many? || preview?(params)
         end
--- a/app/components/find/courses/about_schools_component/view.rb
+++ b/app/components/find/courses/about_schools_component/view.rb
@@ -25,6 +25,26 @@ module Find
             study_sites.any? ||
             site_statuses.map(&:site).uniq.many? || preview?(params)
         end
+
+        def show_higher_education_guidance?
+          return false unless course.higher_education_programme?
+
+          course_information_config(:higher_education, :except_provider_codes).exclude?(course.provider.provider_code)
+        end
+
+        def show_scitt_guidance?
+          return false unless course.scitt_programme?
+
+          course_information_config(:scitt_programmes, :except_provider_codes).exclude?(course.provider_code)
+        end
+
+        private
+
+        def course_information_config(*path)
+          @course_information_config ||= Rails.application.config_for(:course_information)
+
+          @course_information_config.dig(:where_you_will_train, *path)
+        end
       end
     end
   end
--- a/app/components/find/courses/contact_details_component/view.rb
+++ b/app/components/find/courses/contact_details_component/view.rb
@@ -12,6 +12,14 @@ module Find
           super
           @course = course
         end
+
+        def show_address?
+          course.provider_code.in?(course_information_config(:only_provider_codes)) && course.course_code.in?(course_information_config(:only_course_codes))
+        end
+
+        def course_information_config(*path)
+          Rails.application.config_for(:course_information).dig(:show_address, *path)
+        end
       end
     end
   end
--- a/app/components/find/courses/contact_details_component/view.rb
+++ b/app/components/find/courses/contact_details_component/view.rb
@@ -7,6 +7,7 @@ module Find
         attr_reader :course
 
         delegate :provider, to: :course
+        delegate :contact_form?, :contact_form, :show_address?, to: :course_information_config
 
         def initialize(course)
           super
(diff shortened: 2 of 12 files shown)


JR-G (approved):
This looks good Iain and the correct content is rendering! 

Re the naming of `course_information` I wonder if we could use something like `provider_specific_information`/`custom_course_information`/`course_information_exceptions` etc. can't say any of my suggestions are amazing but something to make it a bit more explicit that we need to render something unique for these providers. wdyt?

app/components/find/courses/about_schools_component/view.rb
```diff
@@ -21,11 +21,24 @@ def initialize(course)
...
+          course.higher_education_programme? && view_config(:higher_education, :except_provider_codes).exclude?(course.provider.provider_code)
+        end
+
+        def show_scitt_guidance?
+          course.scitt_programme? && view_config(:scitt_programmes, :except_provider_codes).exclude?(course.provider_code)
+        end
+
+        private
+
+        def view_config(*path)
```
vassyz, line 39: Isn't the method's name a bit too generic? It does hardcode `course_information` and `where_you_will_train`. In the other `config_for` scenarios you are assigning it to `@course_information_config`.

app/components/find/courses/contact_details_component/view.rb
```diff
@@ -12,6 +12,30 @@ def initialize(course)
...
+        end
+
+        private
+
+        def contact_forms
+          course_information_config(:contact_forms).stringify_keys
+        end
+
+        def course_information_config(*path)
+          @course_information_config ||= Rails.application.config_for(:course_information)
```
vassyz, line 35: You are memoizing a `config_for` call? 🤔

inulty-dfe:
Yes naming things is the hard part here. I'm very open to suggestions!

app/components/find/courses/about_schools_component/view.rb
inulty-dfe: Yeah, needs more work.

app/components/find/courses/contact_details_component/view.rb
inulty-dfe: The `config_for` loads data from disk on every call.

inulty-dfe:
> This looks good Iain and the correct content is rendering!
> 
> Re the naming of `course_information` I wonder if we could use something like `provider_specific_information`/`custom_course_information`/`course_information_exceptions` etc. can't say any of my suggestions are amazing but something to make it a bit more explicit that we need to render something unique for these providers. wdyt?

`course_view_configurations`
`course_presentation_config`

app/components/find/courses/contact_details_component/view.rb
vassyz: If it gets called often, why not add it to application config?
```ruby
config.x.course_information = config_for(:course_information)
```

and

```ruby
def course_information_config(*path)
  Rails.application.config.x.course_information.dig(*path)
end
```

spec/features/publish/courses/editing_course_information_guidance_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      def given_a_provider_which_does_not_show_guidance
+        @provider = create(:provider, :scitt, provider_code: 'E65')
+      end
+    end
+
+    context 'when provider is not excluded via config' do
+      scenario 'Guidance is shown' do
+        given_a_provider_which_does_show_guidance
+        and_the_provider_has_a_course
+        and_i_am_authenticated_as_a_provider_user
```
avinhurry, line 55: Do you think it's worth putting the authentication in a before block since it's shared between all of these tests? 

Like: https://github.com/DFE-Digital/publish-teacher-training/blob/4057005e39d40301d9c68a37073898a451468bfb/spec/features/publish/managing_schools_spec.rb#L6-L10

config/course_information.yml
```diff
@@ -0,0 +1,18 @@
+shared:
+  where_you_will_train:
+    universities:
+      except_provider_codes:
+        - B31
```
avinhurry, line 5: @JR-G do you know if there are plans to remove exceptions for specific providers in the codebase?

JR-G: Personally I would rather we just remove all of them and make no exceptions for any of them – apart from the UCL one, since that is a different situation. 

However, happy to move ahead with this as an interim solution, if we can reach out to the two providers to understand why we had this in place.

avinhurry: I agree

inulty-dfe: Yeah, B31 is not a university so this is not technically relevant. I'm just reproducing what existed before.
This should just apply to all universities.

app/services/configs/course_information.rb
```diff
@@ -0,0 +1,25 @@
+module Configs
```
vassyz, line 1: If you created a class for this, why use the `yml` file? You can have the values in a hash here. Also, is there a requirement for this to be a "configuration"? To me it doesn't feel like something that should be in `/config`

app/components/find/courses/contact_details_component/view.rb
inulty-dfe: I ran some benchmarks and decided it isn't worth memoizing after all

spec/features/publish/courses/editing_course_information_guidance_spec.rb
inulty-dfe: I was having trouble changing the authenticated provider in the spec so I decided to create the provider before authenticating on each spec because of how the authenticating method is implemented. I didn't want to make changes to that.

app/services/configs/course_information.rb
vassyz: I don't have an issue with storing data in yaml files, but I'm not a fan of handling it as a configuration. You are calling it `static data` in your comment as well.

Does it not look a bit off having `Configs` here?:

```ruby
@course_information ||= Configs::CourseInformation.new(@course)
```

The way I'd see this working is something similar to the `ActiveYaml` implementation of [active_hash](https://github.com/active-hash/active_hash), potentially having a model for course information.

spec/services/configs/course_information_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+require 'rails_helper'
+
+describe Configs::CourseInformation do
+  let(:scitt_provider) { build(:provider, :scitt, provider_code:) }
+  let(:university_provider) { build(:provider, :university, provider_code:) }
+
+  let(:he_course) { build(:course, :with_higher_education, provider:, course_code:) }
+  let(:scitt_course) { build(:course, :with_scitt, provider:, course_code:) }
+
+  let(:type) { :provider_type }
```
vassyz, line 12: You don't need this I think, it's being defined in every context.

spec/services/configs/course_information_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+    let(:provider) { scitt_provider }
+    let(:course) { scitt_course }
+
+    context 'when provider has a contact form' do
+      let(:provider_code) { 'U80' }
+
+      it 'returns the contact form URL' do
+        obj = described_class.new(course)
+
+        expect(obj.contact_form).to match(/^https/)
```
vassyz, line 122: `expect(obj.contact_form).to match(URI::DEFAULT_PARSER.make_regexp)`?

spec/services/configs/course_information_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    context 'when course and provider codes do not exist' do
+      it 'returns false' do
+        obj = described_class.new(course)
+
+        expect(obj.show_address?).to be(false)
+      end
+    end
+  end
+end
```
vassyz, line 158: You can make this shorter if you prefer
```ruby
describe '#show_address?' do
  subject { described_class.new(course).show_address? }

  let(:provider) { scitt_provider }
  let(:course) { scitt_course }

  context 'when course and provider codes exist' do
    let(:provider_code) { '28T' }
    let(:course_code) { 'X104' }

    it { is_expected.to be(true) }
  end

  context 'when course and provider codes do not exist' do
    it { is_expected.to be(false) }
  end
end
```

app/services/configs/course_information.rb
```diff
@@ -0,0 +1,34 @@
...
+      @course = course
+    end
+
+    def show_placement_guidance?(type)
+      case type
+      when :provider_type
+        subtype = @course.provider.provider_type
+      when :program_type
+        subtype = @course.program_type
+      end
```
vassyz, line 16: Should we not handle an `else` branch here even if there are currently only two possible types? The next line requires a `subtype` value.

app/services/configs/course_information.rb
vassyz: Thoughts on adding `data` in the namespace? For example `configs/data/course_information.yml`?

app/services/configs/course_information.rb
inulty-dfe: That's validating the `type` argument then, isn't it? I might consider that in a complicated method that accepts various types like String, Array or Numeric for the same argument, but I'm not inclined to do that here.

vassyz (approved):

PR #4011, opened 2024-01-18
Support getting resources from the current recruitment cycle

ollietreend:
## Context

The School Placements and Track & Pay services need to retrieve a list of providers from the [Teacher Training Courses API](https://api.publish-teacher-training-courses.service.gov.uk/docs/).

Within this app, providers belong to a recruitment cycle. The existing [GET /recruitment_cycles/{year}/providers](https://api.publish-teacher-training-courses.service.gov.uk/docs/api-reference.html#recruitment_cycles-year-providers-get) endpoint requires clients to specify which recruitment cycle they want to get providers for.

However in the case of School Placements and Track & Pay, we will always need to retrieve providers for the **current recruitment cycle**. Moreover, recruitment cycle years do not align with calendar years, meaning clients need to understand how to calculate the current recruitment cycle year is if that's what they need to retrieve.

Trello card: https://trello.com/c/0JIgI5Nj/127-use-the-current-recruitment-cycle-when-retrieving-providers-from-the-api

I'm proposing changes that should make the API and its documentation more explicit and predictable in what it supports.

### Feature: retrieve providers for the current recruitment cycle

I've added support for `"current"` to be specified in place of a recruitment cycle year. This value serves as an alias for the currently active recruitment cycle.

API clients can now call:

```
GET /recruitment_cycles/current/courses
GET /recruitment_cycles/current/providers
GET /recruitment_cycles/current/providers/locations
GET /recruitment_cycles/current/providers/courses
GET /recruitment_cycles/current/providers/courses/{code}/locations
```

These will return providers and courses for the current recruitment cycle. I have updated the API documentation to explicitly point out that this is a supported use of these endpoints.

### ⚠️ Breaking change: remove fallback for non-existent recruitment cycles

I have also removed 'fallback' behaviour which meant that requests for non-existent recruitment years automatically returned the current recruitment cycle. This behaviour was unusual and slightly unexpected, because it returned a `200 OK` response with no indication that you'd requested data associated with a resource that doesn't exist.

While this is a breaking change, I note that there were no explicit tests for this behaviour, nor was it included in the API documentation. It could be considered an undocumented 'feature' or a bug, depending on your point of view. So I've made the assumption that it is safe to remove the behaviour in favour of explicit support for the `"current"` endpoint.

For example, it is **no longer possible** to call:

```
GET /recruitment_cycles/some-random-string/providers
GET /recruitment_cycles/some-random-string/courses
GET /recruitment_cycles/some-random-string/providers/locations
GET /recruitment_cycles/some-random-string/providers/courses
GET /recruitment_cycles/some-random-string/providers/courses/{code}/locations
```

These requests will now result in a `404 Not Found` error response.

## Guidance to review

To test this locally:

1. Pull this branch:
    ```
    git checkout add-api-endpoints-for-current-year
    ```
2. Run the app:
    ```
    bin/dev
    ```
3. Request the "current" endpoint. You should receive a valid response.
    ```
    $ curl -Is http://publish.localhost:3001/api/public/v1/recruitment_cycles/current/providers | head -n 1
    HTTP/1.1 200 OK
    ```
4. Request an invalid recruitment cycle. You should receive an error response.
    ```
    $ curl -Is http://publish.localhost:3001/api/public/v1/recruitment_cycles/non-existent/providers | head -n 1
    HTTP/1.1 404 Not Found
    ```

## Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [N/A] Inform data insights team due to database changes

### DIFF
--- a/swagger/public_v1/api_spec.json
+++ b/swagger/public_v1/api_spec.json
@@ -1926,7 +1926,7 @@
             "name": "year",
             "in": "path",
             "required": true,
-            "description": "The starting year of the recruitment cycle.",
+            "description": "The starting year of the recruitment cycle. Also accepts \"current\" for the current recruitment cycle.",
             "example": "2020",
             "schema": {
               "type": "string"
--- a/swagger/public_v1/api_spec.json
+++ b/swagger/public_v1/api_spec.json
@@ -2081,7 +2081,7 @@
             "name": "year",
             "in": "path",
             "required": true,
-            "description": "The starting year of the recruitment cycle.",
+            "description": "The starting year of the recruitment cycle. Also accepts \"current\" for the current recruitment cycle.",
             "example": 2024,
             "schema": {
               "type": "string"
@@ -2155,7 +2155,7 @@
             "name": "year",
             "in": "path",
             "required": true,
-            "description": "The starting year of the recruitment cycle.",
+            "description": "The starting year of the recruitment cycle. Also accepts \"current\" for the current recruitment cycle.",
             "example": "2020",
             "schema": {
               "type": "string"
@@ -2340,7 +2340,7 @@
             "name": "year",
             "in": "path",
             "required": true,
-            "description": "The starting year of the recruitment cycle.",
+            "description": "The starting year of the recruitment cycle. Also accepts \"current\" for the current recruitment cycle.",
             "example": 2024,
             "schema": {
               "type": "string"
--- a/swagger/public_v1/api_spec.json
+++ b/swagger/public_v1/api_spec.json
@@ -1927,7 +1927,7 @@
             "in": "path",
             "required": true,
             "description": "The starting year of the recruitment cycle. Also accepts \"current\" for the current recruitment cycle.",
-            "example": "2020",
+            "example": 2024,
             "schema": {
               "type": "string"
             }
@@ -2156,7 +2156,7 @@
             "in": "path",
             "required": true,
             "description": "The starting year of the recruitment cycle. Also accepts \"current\" for the current recruitment cycle.",
-            "example": "2020",
+            "example": 2024,
             "schema": {
               "type": "string"
             }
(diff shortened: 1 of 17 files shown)


JR-G:
Re the breaking change – just for confirmation Apply & Register (the two confirmed stakeholders) are currently scoping by current recruitment cycle anyway.

JR-G:
@ollietreend change looks good and I'm in support of it – just one thing to discuss!

There are two additional endpoints scoped to a specific provider:
`api/public/v1/recruitment_cycles/2024/providers/provider-code/courses`
`api/public/v1/recruitment_cycles/2024/providers/provider-code/locations`

Do you foresee SP/T&P ever needing these? My instinct is that if we're introducing a pattern (or at least being more explicit about it) we should be consistent, however appreciate this might not be in scope for your requirements. Happy to chat!

swagger/public_v1/api_spec.json
```diff
@@ -1926,7 +1926,7 @@
             "name": "year",
             "in": "path",
             "required": true,
-            "description": "The starting year of the recruitment cycle.",
+            "description": "The starting year of the recruitment cycle. Also accepts \"current\" for the current recruitment cycle.",
             "example": "2020",
```
JR-G, line 1930: minor 🤏  `2023` has been bumped to `2024`, but `2020` has been left in place in a couple of examples. If we're touching the docs already, I wonder if it's good to have a clean slate and move everything to `2024`!

ollietreend: Changed in bf576799da257a2ceef7d279c1b994aad1afe1e8

ollietreend:
@JR-G thanks for your feedback. I've addressed your points, so the endpoints (and 'example' year!) should be more consistent now.

See the last two commits on this branch:

- [Add support for nested resources in "current" year](https://github.com/DFE-Digital/publish-teacher-training/pull/4011/commits/71f714e875328c4f9b990c0c80a2981dc1122db5)
- [Update example year to 2024](https://github.com/DFE-Digital/publish-teacher-training/pull/4011/commits/bf576799da257a2ceef7d279c1b994aad1afe1e8)

JR-G (approved):
🚢

PR #4018, opened 2024-01-23
Log env to the rails console

JR-G:
### Context

On Apply when you enter the rails console we output the environment, and make a distinction if it's prod. This is a useful visual indication! 

image

### DIFF
--- a/config/initializers/console.rb
+++ b/config/initializers/console.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module PublishConsole
+  def start
+    show_warning_message_about_environments
+    super
+  end
+
+  def show_warning_message_about_environments
+    if Rails.env.production_aks?
+      puts ('*' * 50).red
+      puts '** You are in the Rails console for PRODUCTION! **'.red
+      puts ('*' * 50).red
+    else
+      puts ('-' * 65).blue
+      puts "-- This is the Rails console for the #{Rails.env} environment. --".blue
+      puts ('-' * 65).blue
+    end
+  end
+end
+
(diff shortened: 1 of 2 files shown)


config/initializers/console.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module PublishConsole
+  def start
+    show_warning_message_about_environments
+    super
+  end
+
+  def show_warning_message_about_environments
+    if Settings.environment.name == 'beta'
```
avinhurry, line 10: I think our check for production should be `Rails.env.production_aks?`, but the linter complains. What do you think?

`W: Rails/UnknownEnv: Unknown environment production_aks. Did you mean production?`

avinhurry: Like this: https://github.com/DFE-Digital/publish-teacher-training/pull/4020/files

JR-G: Ah I posted my response before your second comment came through, will update! Thanks Avin

config/initializers/console.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def show_warning_message_about_environments
+    if Rails.env.production_aks?
+      puts ('*' * 50).red
+      puts '** You are in the Rails console for PRODUCTION! **'.red
+      puts ('*' * 50).red
+    else
+      puts ('-' * 65).blue
+      puts "-- This is the Rails console for the #{Settings.environment.name} environment. --".blue
```
avinhurry, line 16: I think you can use this, probably better for consistency?

```suggestion
      puts "-- This is the Rails console for the #{Rails.env} environment. --".blue
```

avinhurry (approved):

PR #4055, opened 2024-02-15
lb/Add explicit support for getting providers for current year

elceebee:
### Context
The Teacher Training Courses API has implicit behaviour where requests against a non-existent recruitment cycle will silently fall back to the current recruitment cycle. [According to the docs](https://api.publish-teacher-training-courses.service.gov.uk/docs/api-reference.html#recruitment_cycles-year-providers-get), {year} is a required parameter. However it turns out we can actually pass "current" as the year and – [by chance](https://github.com/DFE-Digital/publish-teacher-training/blob/4d680dbe27cdb5300ef07e886ad791e438b0dd9a/app/controllers/api/public/v1/providers_controller.rb#L86-L90) – get the current year. This is undocumented behaviour, but something that [itt-mentor-services now relies upon](https://github.com/DFE-Digital/itt-mentor-services/blob/844909fd555423b56ab4cfd940344b99241461f5/app/services/provider/api.rb#L18).

In [previous PR](https://github.com/DFE-Digital/publish-teacher-training/pull/4011) we changed this behaviour to make it more explicit. We added support for passing in "current" as the recruitment cycle, and dropped the fallback behaviour for other non-existent recruitment cycles. And we updated the documentation to reflect this.

However this change had to be reverted shortly after deployment due to the breaking change it introduced – i.e. requests for invalid years generate a 404 Not Found response, where they previously silently fell back to the current year. It turned out other API consumers were dependent on the previous behaviour.

This PR is almost identical to the first one, but without the breaking change. 

[Trello card](https://trello.com/c/Wq8mQZzB)

- Tests and documentation that make it explicit that 'current' is a valid 'year' and results in the `current_recruitment_cycle`
- Tests and documentation that make it explicit that the default / fallback behaviour for an invalid year is `current_recruitment_cycle`

### Guidance to review

Check out this branch
- Request the 'current' end point, you should receive a valid response
```
$ curl -Is http://publish.localhost:3001/api/public/v1/recruitment_cycles/current/providers | head -n 1
HTTP/1.1 200 OK
```

- Request an invalid year, you should receive a valid response.
```
$ curl -Is http://publish.localhost:3001/api/public/v1/recruitment_cycles/non-existent/providers | head -n 1
HTTP/1.1 200 OK
```

The content of both responses should be the same. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

(no database changes)


spec/controllers/api/public/v1/providers/locations_controller_spec.rb
```diff
@@ -66,4 +66,36 @@
...
+      before do
+        provider.sites << build_list(:site, 5, provider:)
+
+        get :index, params: {
+          recruitment_cycle_year: '1066',
+          provider_code: provider.provider_code
+        }
+      end
+
+      it 'returns locatinos for current recruitment year' do
```
JR-G, line 96: minor typo – `locations`

spec/controllers/api/public/v1/courses_controller_spec.rb
```diff
@@ -24,6 +24,20 @@
         provider.courses << build_list(:course, 2, provider:)
       end
 
+      context 'when "current" is specified as recruitment cycle' do
+        let!(:next_cycle) { create(:recruitment_cycle) }
```
JR-G, line 28: Is this required?

elceebee: no, it is not! Thanks for spotting that.

JR-G (approved):

PR #4071, opened 2024-02-22
Decrease the find radius search range to 50 miles

avinhurry:
### Context

As part of the Find hack day, we are decreasing the location range that Find uses when searching by postcode from 50 miles to 10 miles.

### Guidance to review

🚢


avinhurry:
This was a hack day idea, closing. this now.

tomas-stefano (approved):

PR #4101, opened 2024-03-06
[1265] Fix opt-in and opt-out of notifications

avinhurry:
### Context

Fixing the "opt-in and opt-out of notifications" feature. 

- Updating the regex pattern to specifically use `[^/]+` to match characters up to but not including the next `/` . This change fixes an encoding issue encountered in the previous version where `%2F` (the URL-encoded version of `/`) was being incorrectly included in the matched sections of the URL. 

- Also prefixing the match with `publish`, which is the correct URL.

### Before

image

image

### After

image

image

### Guidance to review

- Try to switch notifications on and off.
- Ensure the back link on that page works as expected. 
- It appears this is cycle independent. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/view_objects/notifications_view.rb
+++ b/app/view_objects/notifications_view.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class NotificationsView
-  ORGANISATION_URL_PATTERN = %r{/organisations/(\S+)/?}
+  ORGANISATION_URL_PATTERN = %r{/publish/organisations/([^/]+)/}
 
   def initialize(
     request:,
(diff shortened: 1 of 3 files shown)


avinhurry:
[Add unsubscribe link to Publish notifications](https://trello.com/c/bXeUuwkV/1265-add-unsubscribe-link-to-publish-notifications)

elceebee (approved):
Nice work! I've tested it locally and everything works as expected.

PR #4112, opened 2024-03-14
Strip whitespace before adding new users to the database

avinhurry:
### Context

Users were being added with unintentional leading and proceeding white space, causing login issues.

Strip the white space before and after the email address.

### Guidance to review

[Test it out here](https://publish-review-4112.test.teacherservices.cloud/support/2024/users/new)

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support/users_controller.rb
+++ b/app/controllers/support/users_controller.rb
@@ -22,8 +22,6 @@ module Support
 
     def create
       @user = User.new(user_params)
-      @user.email.strip!
-
       if @user.save
         redirect_to support_recruitment_cycle_users_path(params[:recruitment_cycle_year])
       else
--- a/app/controllers/support/users_controller.rb
+++ b/app/controllers/support/users_controller.rb
@@ -22,6 +22,8 @@ module Support
 
     def create
       @user = User.new(user_params)
+      @user.email.strip!
+
       if @user.save
         redirect_to support_recruitment_cycle_users_path(params[:recruitment_cycle_year])
       else
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -5,7 +5,7 @@ class User < ApplicationRecord
   include PgSearch::Model
   include RecruitmentCycleHelper
 
-  before_save :downcase_email
+  before_save :downcase_email, :strip_email_whitespace
 
   has_many :organisation_users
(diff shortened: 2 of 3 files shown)


app/controllers/support/users_controller.rb
```diff
@@ -22,6 +22,8 @@ def edit
 
     def create
       @user = User.new(user_params)
+      @user.email.strip!
```
elceebee, line 25: There is a `downcase_email` in a `before_save` hook on the `User` model. My preference would be to have both the strip and the downcase happen at the same time, and on the model. There may be other methods for creating a user and we will always want the whitespace removed. 

What do you think?

elceebee: Sweet!

avinhurry: Yup defo agree, updated here: https://github.com/DFE-Digital/publish-teacher-training/pull/4112/commits/3661e071d72c5723c631c94002bdb93dd52d2711

app/models/user.rb
```diff
@@ -5,7 +5,7 @@ class User < ApplicationRecord
   include PgSearch::Model
   include RecruitmentCycleHelper
 
-  before_save :downcase_email
+  before_save :downcase_email, :strip_email_whitespace
```
elceebee, line 8:
```suggestion
  normalizes :email, with: -> email { email.strip.downcase }
```

That should work without the `before_save` hook, according the documentation.

avinhurry: This only possible in rails 7.1 **or** with the normalize gem. I don't think it's worth adding the gem just for this. 

https://www.shakacode.com/blog/rails-add-normalize-to-activerecord/#:~:text=Normalization%20is%20applied%20when%20the,subsequent%20querying%20using%20unnormalized%20values.

elceebee: Ooh right! I hadn't realised we weren't on Rails 7 yet. Sorry about that!

elceebee (approved):

app/models/user.rb
avinhurry: No probs, we need to do the upgrade!

PR #4131, opened 2024-03-27
[1499] tech debt part 1 remove subject_knowledge_enhancement_course_available from financialincentive

elceebee:
### Context

The changes in SKE policies made us realise that we have a field on the `FinancialIncentive` model that is not being used anywhere in the code base. 

- `subject_knowledge_enhancement_course_available` has not been updated since 2020. 
- There are currently no `FinancialIncentive` records where `subject_knowledge_enhancement_course_available` is set to true.

- This PR just removes all the references to this field. it does not remove the column. I'll do that in a second PR. 
- Rather than removing the field from the API, and possibly causing a consumer a headache, I've just changed it so the field always comes back nil. 

### Guidance to review

- I've checked in production that this is always `nil` or `false`. 
- Those subjects (22) where it is `false` are: "French", "German", "Spanish", "Italian", "Japanese", "Mandarin", "Russian", "Modern languages (other)", "Latin", "Ancient Greek", "Ancient Hebrew", "Geography", "Chemistry", "Computing", "Mathematics", "Physics", "Biology", "Design and technology", "Art and design", "English", "Music", "Religious education".  I can't see where it is used anywhere. Let me know if you can think of anything else!

### Trello card
https://trello.com/c/3vseuZV2

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [NA] Inform data insights team due to database changes


defong (approved):
This makes sense, as long as there is no need to bring to the awareness to the user the fact that SKE course is available for them. If memory serves me right the usage of SKE course is to bring up the skills of a potential teacher, ie they studied a French degree and they want to become a Modern language teacher which encompass other languages such as Chinese than partaking this SKE course will allow them to demonstrate that they have the required knowledge.

The problem you may encounter with deleting it all, is that at a later date requirements to highlight the SKE is needed again. Future problem that may not arise.

avinhurry:
[Tech debt: Remove subject_knowledge_enhancement_course_available from FinancialIncentive](https://trello.com/c/3vseuZV2/1499-tech-debt-remove-subjectknowledgeenhancementcourseavailable-from-financialincentive)

avinhurry (approved):

PR #4146, opened 2024-04-05
Just to make ruby version manager adaptable

tomas-stefano:
### Context

On RVM  I am getting:

```
 Unknown ruby interpreter version (do not know how to handle): ~>3.2.3
```

Removing the `~>` works.


elceebee (approved):
I've run this locally (I'm using asdf) and it doesn't break anything. 🚀

PR #4151, opened 2024-04-09
Replace 'PGCE with QTS' with 'QTS with PGCE'

tomas-stefano:
## Context

There is inconsistency of terminology between GiT and Find. GiT use “QTS with PGCE” (which is more clear/intuitive) whereas Find uses “PGCE with QTS”.

Replace "PGCE with QTS” with "QTS with PGCE".

## Gotcha

When it comes from qualifications is an easy change but because the terminology is inside of the course description from the TTAPI we need to replace part of the string.

## Guidance to review

1. All the places in Find/Publish being replaced?

## Link to Trello card

https://trello.com/c/eftEZsWL/1283-review-and-replace-existing-instances-of-pgce-with-qts-with-the-correct-terminology-of-qts-with-pgce-in-ui-only

### DIFF
--- a/app/components/find/courses/qualifications_summary_component/view.html.erb
+++ b/app/components/find/courses/qualifications_summary_component/view.html.erb
@@ -9,8 +9,8 @@
         It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you’d like to teach in.
       </p>
     <% end %>
-<% when "PGCE with QTS" %>
-    <%= govuk_details(summary_text: "PGCE with QTS") do %>
+<% when "QTS with PGCE" %>
+    <%= govuk_details(summary_text: "QTS with PGCE") do %>
       <p class="govuk-body">
         A postgraduate certificate in education (PGCE) with qualified teacher status (QTS) will allow you to teach in state schools in England and may allow you to teach in other parts of the UK.
       </p>
--- a/app/components/find/courses/qualifications_summary_component/view.html.erb
+++ b/app/components/find/courses/qualifications_summary_component/view.html.erb
@@ -12,13 +12,10 @@
 <% when "QTS with PGCE" %>
     <%= govuk_details(summary_text: "QTS with PGCE") do %>
       <p class="govuk-body">
-        A postgraduate certificate in education (PGCE) with qualified teacher status (QTS) will allow you to teach in state schools in England and may allow you to teach in other parts of the UK.
+       You need qualified teacher status (QTS) to teach in state schools in England. QTS may also allow you to teach in other parts of the UK.
       </p>
       <p class="govuk-body">
-        It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you’d like to teach in.
-      </p>
-      <p class="govuk-body">
-        Many PGCE courses include credits that count toward a Master’s degree.
+        This course also offers a postgraduate certificate in education (PGCE). PGCE courses can include credits that count towards a master’s degree.
       </p>
     <% end %>
   <% when "PGDE with QTS" %>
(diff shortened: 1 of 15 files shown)


tomas-stefano:
[Review and replace existing instances of "PGCE with QTS" with the correct terminology of "QTS with PGCE" [in UI only!]](https://trello.com/c/eftEZsWL/1283-review-and-replace-existing-instances-of-pgce-with-qts-with-the-correct-terminology-of-qts-with-pgce-in-ui-only)

app/decorators/course_decorator.rb
```diff
@@ -27,6 +27,10 @@ def find_url(provider = object.provider)
     h.search_ui_course_page_url(provider_code: provider.provider_code, course_code: object.course_code)
   end
 
+  def description
+    object.description.to_s.gsub('PGCE with QTS', 'QTS with PGCE')
```
inulty-dfe, line 31: does this need to be `gsub` or can it  be `sub`?

tomas-stefano: Can be any of them and will work. I am happy either one.

app/models/concerns/with_qualifications.rb
```diff
@@ -45,7 +45,7 @@ module WithQualifications
 
     # This field may seem like an unnecessary overhead when there is already a
     # database-backed `qualification` field. However it's misleading, from the
-    # point of view of the teacher training domain, to think of 'PGCE with QTS'
+    # point of view of the teacher training domain, to think of 'QTS with PGCE'
```
dcyoung-dev, line 48: On line 68 `when 'pgce_with_qts' then 'Postgraduate certificate in education (PGCE) with qualified teacher status (QTS)'` should we be updating this wording too?

app/components/find/courses/qualifications_summary_component/view.html.erb
```diff
@@ -9,16 +9,13 @@
...
       </p>
       <p class="govuk-body">
-        It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you’d like to teach in.
-      </p>
-      <p class="govuk-body">
-        Many PGCE courses include credits that count toward a Master’s degree.
+        This course also offers a postgraduate certificate in education (PGCE). PGCE courses can include credits that count towards a master’s degree.
       </p>
     <% end %>
   <% when "PGDE with QTS" %>
```
dcyoung-dev, line 21: Should we be updating "PGDE with QTS" to "QTS with PGDE" as well?

dcyoung-dev (changes requested):
I have concerns that we may be missing other instances of this wording.

app/components/find/courses/qualifications_summary_component/view.html.erb
tomas-stefano: > Should we be updating "PGDE with QTS" to "QTS with PGDE" as well?

I asked the team and they mentioned PGDE will be done in a separate card and separate work.

app/models/concerns/with_qualifications.rb
tomas-stefano: I pushed the fix. Is that the only place you found?

app/decorators/course_decorator.rb
tomas-stefano: I changed to use sub.

app/models/concerns/with_qualifications.rb
dcyoung-dev: This is the only place that I saw - I didn't look very far to be honest

app/components/find/courses/qualifications_summary_component/view.html.erb
dcyoung-dev: Fair enough - seems like an unnecessary split to me 🤷‍♂️

dcyoung-dev (approved):

inulty-dfe (approved):

PR #4159, opened 2024-04-12
Makefile environment names to be equal to Apply

tomas-stefano:
### Context

The Publish and Apply repos are managed by the same team and there are features that exist in one and not the other.

One such feature is convenient commands to login to console environments such as production and sandbox.

Login to any environment exactly like Apply:

```
 make qa get-cluster-credentials
 make staging get-cluster-credentials
 make sandbox get-cluster-credentials
 make production get-cluster-credentials
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,5 +268,5 @@ shell: aks-ssh
 .PHONY: worker-shell
 worker-shell: aks-worker-ssh
 
-.PHONY: aks-console
+.PHONY: console
 console: aks-console
--- a/Makefile
+++ b/Makefile
@@ -246,3 +246,27 @@ action-group-resources: set-azure-account # make env_aks action-group-resources 
 	echo ${RESOURCE_NAME_PREFIX}-${SERVICE_SHORT}-mn-rg
 	az group create -l uksouth -g ${RESOURCE_NAME_PREFIX}-${SERVICE_SHORT}-mn-rg --tags "Product=Find postgraduate teacher training" "Environment=Test" "Service Offering=Teacher services cloud"
 	az monitor action-group create -n ${RESOURCE_NAME_PREFIX}-${SERVICE_NAME} -g ${RESOURCE_NAME_PREFIX}-${SERVICE_SHORT}-mn-rg --action email ${RESOURCE_NAME_PREFIX}-${SERVICE_SHORT}-email ${ACTION_GROUP_EMAIL}
+
+.PHONY: qa
+qa: qa_aks
+
+.PHONY: staging
+staging: staging_aks
+
+.PHONY: sandbox
+sandbox: sandbox_aks
+
+.PHONY: production
+production: production_aks
+
+.PHONY: logs
+logs: aks-logs
+
+.PHONY: shell
+shell: aks-ssh
+
+.PHONY: worker-shell
+worker-shell: aks-worker-ssh
+
+.PHONY: aks-console
+console: aks-console


tomas-stefano:
[Tech debt: Update Publish repo Makefile to match Apply repo](https://trello.com/c/vvz8wc0C/1046-tech-debt-update-publish-repo-makefile-to-match-apply-repo)

Makefile
```diff
@@ -59,17 +59,17 @@ dv_review_aks: ## make dv_review_aks deploy APP_NAME=2222 CLUSTER=cluster1 USE_D
...
-staging_aks:
+staging:
 	$(eval include global_config/staging_aks.sh)
 
-sandbox_aks:
+sandbox:
 	$(eval include global_config/sandbox_aks.sh)
 
-production_aks:
+production:
```
avinhurry, line 72: Hmm, I know this is out of scope for this card but do you know if we can swap these around? I feel like we should be consistent.

image

avinhurry (commented):
Publish is: `make production_aks aks-console CONFIRM_PRODUCTION=TRUE`
Apply is: `make production shell CONFIRM_PRODUCTION=yes`

Are we making these consistent too, i.e `aks-console` and `shell`? I feel like that would be one of the most used commands so would be good to have that consistent if possible. 

Any thoughts?

Makefile
tomas-stefano: Humm the problem is to make the same as a Apply but Apply doesn't have the production_aks environment.

I am strongly thinking that it might be the best in Publish to have both. make production ... and make production_aks ...

Wdyt?

avinhurry: Why would we want both `make production` and `make production_aks` in Publish? Or I have I misunderstood?

tomas-stefano: Sorry I answer because I understood another thing yesterday.

tomas-stefano:
@avinhurry agreed I renamed aks-console to shell :)

inulty-dfe:
I would propose:

 - `aks-ssh` => `shell`
 - `ask-console` => `console`

Makefile
```diff
@@ -138,23 +138,26 @@ get-cluster-credentials: read-cluster-config set-azure-account ## make <config>
 	az aks get-credentials --overwrite-existing -g ${RESOURCE_NAME_PREFIX}-tsc-${CLUSTER_SHORT}-rg -n ${RESOURCE_NAME_PREFIX}-tsc-${CLUSTER}-aks
 	kubelogin convert-kubeconfig -l $(if ${GITHUB_ACTIONS},spn,azurecli)
 
-aks-console: get-cluster-credentials
```
dcyoung-dev, line 141: Will any of these changes affect the infra team?
I see in Apply we alias the commands https://github.com/DFE-Digital/apply-for-teacher-training/blob/0d9f56abc3c0bbe22d99fc904b955d0bfa466609/Makefile#L133

tomas-stefano: Good question. I added @RMcVelia. Can you review this PR please?

RMcVelia: If you change (instead of alias) then there are other places that the code will need to be changed. e.g. github workflows are hardcoded to env_aks

tomas-stefano: thank you @RMcVelia. I will alias everything

tomas-stefano:
@RMcVelia @dcyoung-dev @inulty-dfe I addressed the changes, can you review again please?

Makefile
```diff
@@ -246,3 +246,27 @@ action-group-resources: set-azure-account # make env_aks action-group-resources
...
+.PHONY: logs
+logs: aks-logs
+
+.PHONY: shell
+shell: aks-ssh
+
+.PHONY: worker-shell
+worker-shell: aks-worker-ssh
+
+.PHONY: aks-console
```
RMcVelia, line 271: should it be .PHONY: console ?

tomas-stefano: Yes! Good catch! I pushed the fix :)

RMcVelia (approved):

PR #4162, opened 2024-04-16
Rolled over courses are closed by default and automatically opened on publish

inulty-dfe:
### Context

When a course is rolled over, it is rolled over to the “rolled over” state.

It is currently rolled over with the same `application_status` it had in the previous cycle.

It doesn’t make sense from a data analysis point of view for the course to be `unpublished` and `open` at the beginning of the cycle.

Therefore we should roll over courses and set the course `application_status` to closed until it is `published`.

We automatically open a course during the publish step to avoid the multi-step process of 'publish', 'open', 'confirm'.

Now the provider needs only to publish a closed course and it will be open for applications pending it's scheduled open date.

When a provider publishes a course, they must "Open" the course, even if the course was open in the previous cycle.
We've changed this to:

 - Rollover all courses closed
 - automatically open the course when it is published

## Before

|Rolled Over|Published|
|---|---|
|1_unpublished|2_published|

|Opening|Opened|
|---|---|
|3_opening_course|4_opened|

## After

[Screencast from 21-06-24 16:34:43.webm](https://github.com/DFE-Digital/publish-teacher-training/assets/135042929/98276119-3935-413b-bd66-a22c6d761b9f)

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/spec/services/courses/copy_to_provider_service_spec.rb
+++ b/spec/services/courses/copy_to_provider_service_spec.rb
@@ -88,8 +88,6 @@ RSpec.describe Courses::CopyToProviderService do
   end
 
   context 'when the original course is open' do
-    let(:new_recruitment_cycle) { create(:recruitment_cycle, :next) }
-
     before do
       course.update(application_status: 'open')
     end
--- a/spec/services/courses/copy_to_provider_service_spec.rb
+++ b/spec/services/courses/copy_to_provider_service_spec.rb
@@ -87,6 +87,20 @@ RSpec.describe Courses::CopyToProviderService do
     end
   end
 
+  context 'when the original course is open' do
+    let(:new_recruitment_cycle) { create(:recruitment_cycle, :next) }
+
+    before do
+      course.update(application_status: 'open')
+    end
+
+    it "sets the new course's application_status to closed" do
+      service.execute(course:, new_provider:)
+
+      expect(new_course.application_status).to eq('closed')
+    end
+  end
+
   it 'leaves the existing course alone' do
     service.execute(course:, new_provider:)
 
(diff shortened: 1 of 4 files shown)


inulty-dfe:
[`application_status` is rolled over with previous years value](https://trello.com/c/I3QA3Mwb/1428-applicationstatus-is-rolled-over-with-previous-years-value)

tomas-stefano (commented):
From the code point of view it looks good. It is unclear if there is any impact on rollover for a course being closed?

spec/services/courses/copy_to_provider_service_spec.rb
```diff
@@ -87,6 +87,20 @@
     end
   end
 
+  context 'when the original course is open' do
+    let(:new_recruitment_cycle) { create(:recruitment_cycle, :next, application_start_date: '2023-09-01') }
```
avinhurry, line 91:
```suggestion
    let(:new_recruitment_cycle) { create(:recruitment_cycle, :next) }
```

avinhurry (approved):

inulty-dfe:
> From the code point of view it looks good. It is unclear if there is any impact on rollover for a course being closed?

I've illustrated the expected steps for providers to take when launching a course in the new cycle:

1. Publish
2. Open

Every course will require step 2 now because all copied courses will be copied closed.

spec/services/courses/copy_to_provider_service_spec.rb
```diff
@@ -87,6 +87,20 @@
     end
   end
 
+  context 'when the original course is open' do
+    let(:new_recruitment_cycle) { create(:recruitment_cycle, :next) }
```
avinhurry, line 91: I don't think this line is needed

avinhurry (approved):
Looks good. 

Can we update the PR description 🙏🏼 

> When a provider publishes a course, they must "Open" the course, even if the course was open in the previous cycle.

PR #4196, opened 2024-05-10
Record search results analytics

dcyoung-dev:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/controllers/find/results_controller.rb
```diff
@@ -11,6 +11,15 @@ def index
       @filters_view = ResultFilters::FiltersView.new(params: matched_params)
       @courses = @results_view.courses.page params[:page]
       @number_of_courses_string = @results_view.number_of_courses_string
+
+      track_search_results(number_of_results: @results_view.course_count,
+                           course_codes: @courses.pluck(:course_code).uniq)
```
tomas-stefano, line 16: 2 minor questions just to double check:

1. Does @results_views.courses_count ignore the paginations and pages and send the total total total? 
2. Does the course codes sent to Bigquery will be include all course codes from all pages? 

I could investigate but probably will be faster to ask 😅

tomas-stefano: Read the PR description ` of the current page ` cool! Miss that.

tomas-stefano (approved):

app/controllers/find/results_controller.rb
dcyoung-dev: Yeh the `course_count` is the total of all the results i.e. 1000 courses.
The course codes only show the codes for the page i.e. should only be ~25 codes per page

PR #4203, opened 2024-05-14
Execute a connection to the database on healtcheck

tomas-stefano:
### Context

The active? method only returns true if a connection was established prior to the call, which is not guaranteed.

Making an explicit query will make that return true or false and also guarantees that we can query the DB.

### Trello card

 [Investigate and fix intermittently failing smoke tests on Publish](https://trello.com/c/GJuy7YIg/1629-investigate-and-fix-intermittently-failing-smoke-tests-on-publish)

### DIFF
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -29,8 +29,9 @@ class HeartbeatController < ActionController::API
   private
 
   def database_alive?
-    ActiveRecord::Base.connection.active?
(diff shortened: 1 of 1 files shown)


avinhurry:
[Investigate and fix intermittently failing smoke tests on Publish](https://trello.com/c/GJuy7YIg/1629-investigate-and-fix-intermittently-failing-smoke-tests-on-publish)

PR #4207, opened 2024-05-15
[1567] full-time part-time study change from radio buttons to checkboxes

elceebee:
### Context

from the [Trello ticket](https://trello.com/c/zTDCLOFE):
Following a review of historic design/research findings:

1. Change from radio buttons to checkboxes
2. Removing the ‘Full time or Part time’ field
3. Amend wording to correlate to an either/or choice

Replace radio buttons with checkboxes when selecting study pattern, for both editing a course and creating a new course.

https://github.com/DFE-Digital/publish-teacher-training/assets/44073106/8bcf5122-fbf5-47d6-8d02-60c0aa86b7ef

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [NA] Inform data insights team due to database changes

### DIFF
--- a/app/views/publish/courses/study_mode/new.html.erb
+++ b/app/views/publish/courses/study_mode/new.html.erb
@@ -4,7 +4,7 @@
   <%= govuk_back_link_to(@back_link_path) %>
 <% end %>
 
-<%= render "publish/shared/errors", error_id: "#course-study-mode-field-error" %>
+<%= render "publish/shared/errors" %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
--- a/app/views/publish/courses/study_mode/new.html.erb
+++ b/app/views/publish/courses/study_mode/new.html.erb
@@ -1,10 +1,10 @@
-<% content_for :page_title, title_with_error_prefix("Full time or part time", @errors && @errors.any?) %>
+<% content_for :page_title, title_with_error_prefix(t("page_titles.study_mode.new"), @errors && @errors.any?) %>
 
 <% content_for :before_content do %>
   <%= govuk_back_link_to(@back_link_path) %>
 <% end %>
 
-<%= render "publish/shared/errors" %>
+<%= render "publish/shared/errors", error_id: "#course-study-mode-field-error" %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -34,6 +34,9 @@ en:
       apprenticeship_and_visa: "Teaching apprenticeship and %{visa_type} visas updated"
       visa: "%{visa_type} visas updated"
   page_titles:
+    study_mode:
+      edit: Study pattern
+      new: Study pattern
     funding_type:
       edit: "Funding type"
     skilled_worker_visas:
@@ -292,6 +295,10 @@ en:
     update: "Update user"
   publish:
     providers:
+      study_mode:
+        form:
+          select_all_that_apply: Select all that apply
+          study_pattern: Study pattern
       study_sites:
         index:
           add_study_site: &add_study_site "Add study site"
@@ -832,7 +839,7 @@ en:
         publish/course_study_mode_form:
           attributes:
             study_mode:
-              blank: "Select full time or part time"
+              blank: You must choose a study pattern. Select all that apply.
         publish/course_withdrawal_form:
           attributes:
             confirm_course_code:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -130,8 +130,6 @@ en:
         label: "Part time"
       full_time:
         label: "Full time"
-      full_time_or_part_time:
-        label: "Full time or part time"
     apprenticeship:
       pg_teaching_apprenticeship:
         label: "Yes"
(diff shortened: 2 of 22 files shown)


spec/forms/publish/course_study_mode_form_spec.rb
```diff
@@ -13,23 +13,43 @@ module Publish
...
       end
     end
+
+    context 'when params are study mode is both part time and full time' do
+      let(:params) { { study_mode: %w[full_time part_time] } }
+
+      it 'does calls the course.ensure_site_statuses_match_study_mode' do
+        subject.save!
+
+        expect(course.study_mode).to eq 'full_time_or_part_time'
```
dcyoung-dev, line 41: I might be missing something here, how does checking the value of the `study_mode` relate to the `#ensure_site_statuses_match_study_mode`?

dcyoung-dev: I may have an answer to this question from past me https://dcyoung.dev/shorts/london-or-the-chicago-school-of-tdd

dcyoung-dev (commented):
This is way more complex than I'd have expected, there are some icky parts in this PR that are obviously from past decisions, it's scrambling my brain with the hoops you've had to jump through 🤯 good work 👌
I've left a couple of question about the specs but neither are blockers

spec/forms/publish/course_study_mode_form_spec.rb
```diff
@@ -13,23 +13,43 @@ module Publish
 
       describe '#save!' do
         it 'does not call the course.ensure_site_statuses_match_study_mode' do
-          expect(course).to receive(:changed?)
-          expect(course).not_to receive(:ensure_site_statuses_match_study_mode)
```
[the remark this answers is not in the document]
elceebee: The previous implementation meant that empty params gave valid? == true, so it reached the bit of logic that asked the question about 'changed?'. Now, the empty params returns valid? == false, so it doesn't hit `changed?`. It just doesn't save, which I this is a more reasonable expectation. 

I take your point about updating the name of the specs a bit though.

spec/forms/publish/course_study_mode_form_spec.rb
elceebee: The test is actually meant to test the 'save' method given certain params. 

And part of the change we've implemented going from radio to check boxes is whether or not selecting two boxes results in what previous happened when we selected one box.

ie, instead of `study_mode: 'full_time_or_part_time'` in the params, we end up `study_mode: ['full_time', 'part_time']`. So we want to return that given those new params, it saves as expected. 

I think you're right that it doesn't relate to the `ensure_site_statuses_match_study_mode` method -- that should be tested separately, explicitly. I'll make the change.

dcyoung-dev (approved):

elceebee:
[Publish: full time/part time study - change from radio buttons to checkboxes](https://trello.com/c/zTDCLOFE/1567-publish-full-time-part-time-study-change-from-radio-buttons-to-checkboxes)

tomas-stefano (approved):

config/locales/en.yml
```diff
@@ -832,7 +837,7 @@ en:
         publish/course_study_mode_form:
           attributes:
             study_mode:
-              blank: "Select full time or part time"
+              blank: Select a study pattern
```
inulty-dfe, line 840: I saw on the card @RebeccaCryan  suggested "You must choose a study pattern. Select all that apply". Was there a change of mind?

app/views/publish/courses/study_mode/new.html.erb
```diff
@@ -1,10 +1,10 @@
-<% content_for :page_title, title_with_error_prefix("Full time or part time", @errors && @errors.any?) %>
+<% content_for :page_title, title_with_error_prefix(t("page_titles.study_mode.new"), @errors && @errors.any?) %>
 
 <% content_for :before_content do %>
   <%= govuk_back_link_to(@back_link_path) %>
 <% end %>
 
-<%= render "publish/shared/errors" %>
+<%= render "publish/shared/errors", error_id: "#course-study-mode-field-error" %>
```
inulty-dfe, line 7: I don't understand why this is necessary to change. Does the existing error mechanism not work now or does it have something to do with accessibility?

inulty-dfe (changes requested):
Just a couple small concerns

config/locales/en.yml
elceebee: Yes, Slack thread: https://ukgovernmentdfe.slack.com/archives/C070ANGEZ8V/p1715964090673909

app/views/publish/courses/study_mode/new.html.erb
elceebee: `publish/shared/errors` doesn't work with the govuk checkboxes fieldset. It expects something other than the default field error id, breaking the href navigation from the error summary box to the fieldset. I've used the `govuk_uk checkboxes` as they were intended, which I think is better than writing bespoke html. Ideally, they would be paired with the `govuk_error_summary` component here (as we do for the edit side of things). But that would require refactoring the `CreationService`, which includes some funky error handling, and that is beyond the scope of the work we're doing here.

elceebee: I've reverted the code which fixes the error navigation for this and created a ticket for fixing it everywhere.

PR #4211, opened 2024-05-17
Add test for teacher degree apprentice and further education

tomas-stefano:
### Context

We should not show TDA option from add course - further education flow. 

The implementation does automatically under the file https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/models/concerns/courses/edit_options/qualification_concern.rb#L8

But it is best to increase our coverage for that scenario so this PR only adds a spec to guarantee the functionality is tested and working.

## Trello card

https://trello.com/c/1GZDGD5K/1676-remove-tda-option-for-further-education


tomas-stefano:
[Remove TDA option for further education](https://trello.com/c/1GZDGD5K/1676-remove-tda-option-for-further-education)

inulty-dfe (approved):

PR #4212, opened 2024-05-17
Set default degree requirements for TDA courses

tomas-stefano:
### Context

Because the degree apprenticeship will awards a degree then these fields needs to be the default when provider is adding the course

### Trello card

https://trello.com/c/5MJ3HpDI/1674-default-additional-degree-info-to-false-for-tda-courses

### DIFF
--- a/app/services/courses/creation_service.rb
+++ b/app/services/courses/creation_service.rb
@@ -37,6 +37,9 @@ module Courses
         course.program_type = 'teacher_degree_apprenticeship'
         course.can_sponsor_student_visa = false
         course.can_sponsor_skilled_worker_visa = false
+        course.additional_degree_subject_requirements = nil
+        course.degree_subject_requirements = nil
+        course.degree_grade = 'not_required'
       end
 
       AssignSubjectsService.call(course:, subject_ids:)
--- a/app/services/courses/creation_service.rb
+++ b/app/services/courses/creation_service.rb
@@ -37,8 +37,6 @@ module Courses
         course.program_type = 'teacher_degree_apprenticeship'
         course.can_sponsor_student_visa = false
         course.can_sponsor_skilled_worker_visa = false
-        course.additional_degree_subject_requirements = nil
-        course.degree_subject_requirements = nil
         course.degree_grade = 'not_required'
       end
 
--- a/spec/services/courses/creation_service_spec.rb
+++ b/spec/services/courses/creation_service_spec.rb
@@ -19,6 +19,27 @@ describe Courses::CreationService do
 
   let(:next_available_course_code) { false }
 
+  context 'when teacher degree apprenticeship course' do
+    let(:valid_course_params) do
+      {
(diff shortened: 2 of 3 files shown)


tomas-stefano:
[Default additional degree info to false for TDA courses](https://trello.com/c/5MJ3HpDI/1674-default-additional-degree-info-to-false-for-tda-courses)

app/services/courses/creation_service.rb
```diff
@@ -37,6 +37,9 @@ def build_new_course
         course.program_type = 'teacher_degree_apprenticeship'
         course.can_sponsor_student_visa = false
         course.can_sponsor_skilled_worker_visa = false
+        course.additional_degree_subject_requirements = nil
+        course.degree_subject_requirements = nil
```
avinhurry, line 41: Do we actually need line 40 and 41? They are nil by default

avinhurry (commented):
It's probably worth checking that we can publish a course without the degree validation kicking in but I am happy to add that as part of the course length card as we need to default that too before we can publish.

tomas-stefano:
> It's probably worth checking that we can publish a course without the degree validation kicking in but I am happy to add that as part of the course length card as we need to default that too before we can publish.

Yes, I stumbled that across, I will add to the course length card (which I am doing atm).

app/services/courses/creation_service.rb
tomas-stefano: True! I amended that :).

spec/services/courses/creation_service_spec.rb
```diff
@@ -19,6 +19,27 @@
...
+      }
+    end
+
+    it 'creates the teacher degree apprenticeship course' do
+      expect(subject.program_type).to eq('teacher_degree_apprenticeship')
+      expect(subject.funding_type).to eq('apprenticeship')
+      expect(subject.can_sponsor_student_visa?).to be false
+      expect(subject.can_sponsor_skilled_worker_visa?).to be false
+      expect(subject.additional_degree_subject_requirements).to be_nil
+      expect(subject.degree_subject_requirements).to be_nil
```
avinhurry, line 38: Can we remove from the test too please 🙏🏼

tomas-stefano: Humm, I left on purpose, I think the tests should be there to guarantee any new bugs to be introduced (because this field should be nil regardless of scenario in TDA).

tomas-stefano: Example:

Someone wake up in a morning 2 years from now and decided the DB default to this field to be true. Suddenly we have a bug on TDA and the tests should alert us about it.

avinhurry: Ok, makes sense! All good in that case. 

I did wonder if it might be worth explicitly mentioning that in a comment but I think it's self explanatory, re-reading the test.

avinhurry (approved):

PR #4221, opened 2024-05-21
1675 add 4 year default for tda courses

tomas-stefano:
### Context

We have defaulted all TDA courses to a length of 4 years. 

After this the course can be published. So we need to test the publishing mechanism for a TDA course.

## Observation

I am not sure why, but the course length has two patterns:

1. Enumerator: OneYear, TwoYears
2. Free text: 3 years, etc

I optin for a free text. Let me know if you disagree.

### Guidance to review

1. Does it save 4 years?
4. Does the API returns 4 years as free text?

### Trello card

https://trello.com/c/KRuhpAaL/1675-add-4-year-default-for-tda-courses

### DIFF
--- a/app/services/courses/creation_service.rb
+++ b/app/services/courses/creation_service.rb
@@ -39,6 +39,8 @@ module Courses
         course.can_sponsor_student_visa = false
         course.can_sponsor_skilled_worker_visa = false
         course.degree_grade = 'not_required'
+        course_enrichment = course.enrichments.find_or_initialize_draft
+        course_enrichment.course_length = '4 years'
       end
 
       AssignSubjectsService.call(course:, subject_ids:)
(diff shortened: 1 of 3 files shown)


tomas-stefano:
[Add 4 year default for TDA courses](https://trello.com/c/KRuhpAaL/1675-add-4-year-default-for-tda-courses)

elceebee (approved):
I've tested this locally and it works as expected. Nice one!

There is one odd thing, but it will be fixed with [this ticket](https://trello.com/c/XKuALLag), and @avinhurry and I already chatted about it:  If you edit salary, but then scroll up to course length, you can change to something else and save it. It appears to save successfully. But when you look at the summary page, it remains '4 years'. 

When I've finished my ticket, the course length section won't be on that page, so that behaviour won't exist. Just noting it here in case someone else comes across it and doesn't realise it's in hand.

PR #4224, opened 2024-05-22
Fix duplicate subject validation on secondary course selection

inulty-dfe:
### Context

The primary concern of this PR is to validate the Second level courses' subject inputs.

1. Validate duplicate Secondary Subject inputs
2. Validate missing master subject, even when subordinate subject supplied

When choosing two subjects to associate with a Second level course, the validation preventing a provider from choosing the same subject twice was not working.

When the validation was fixed, it was noticed that the form fields do not maintain the values that the user entered.

#### Add `subordinate_subject_id` attribute to Course
The form sends `master_subject_id`, `subordinate_subject_id` and `subject_ids` but does not process `subordinate_subject_id`.

This is because `master_subject_id` is a persisted column on the schema, and `subordinate_subject_id` is not even an attribute.

I've added `subordinate_subject_id` as an ActiveModel attribute on the `Course` so we can work with it explicitly.

This requires an `after_initialize` hook on the model to populate the property from the associated subjects.

To understand how this happens, you need to understand all about `Subjects`

But the long story short is that we can populate the form `subordinate_subject_id` with this attribute.

```ruby
"master_subject_id"=>"24", "subjects_ids"=>["24", "24"]
```

#### Differences between New and Edit

The Course controllers work significantly differently between creating a **new** course and **editing** and existing course. The two process share some of the same modules however.

**New Course**

```ruby
- SubjectsController#continue
- CourseBasicDetailConcern#build_new_course
- ::Courses::CreationService.call
- AssignSubjectsService
```

The problem here is that this service is called on every step of the course creation service with accumulating parameters from the view which need to be maintained over requests, including when using the back button. This makes it very messy and tricky to manage state.

The subject parameters are passed to `AssignSubjectsService`.
Not only is this service called on every step of the create path, it is also called during the editing path.

The params passed to it are an array of `subject_ids`

The `subject_ids` are built from the parameters in the URL for the new path.

How the `subject_ids` are built and managed is explained:

Select `master` dropdown and blank `subordinate`
`#=> subject_ids [1]`
Select `subordinate` dropdown and blank `master`
  `#=> subject_ids [2]`
  
Click back button after selecting only the `suboridinate` input, and the form populates the `master` input with the values supplied to the `subordinate` input previously.

If modern languages is one of the subject ids, the next page is multiple select of languages. These language ids need to be manages across pages in the URL like the original subject ids.

The subject ids array can have more than 3 elements.

The current implementation PREPENDS the language ids to this list of subject ids in the URL.

This means the master id could be the first, second or third+ position in the subject_ids array.

**Edit Course**

```ruby
- SubjectsController#continue
- CourseSubjectsForm#build_new_course
- AssignSubjectsService
```

#### AssignSubjectsService

This service receives an array of `subject_ids`

The subject_ids could contain
```ruby
[master, subordinate]
[subordinate]
[languages_id_1, master, subordinate]
[languages_id_1, languages_id_2, master, subordinate]
```

It's hard to work with such a variable set of values. It's better if the position of the values is relatively consistent.

The Course Creation process `unshifts` language ids onto the `subject_ids` array. I've changed it so they are pushed onto the end now.

The `secondary_subject_ids` are passed to the AssignSubjectsService without compacting away the missing values.

```ruby
[master, subordinate, languages_id_1]
[master, subordinate, languages_id_1, languages_id_2]
```

#### Errors

The Course creation / edit system depends heavily on `errors`. The errors on the object need to be managed carefully in order to show the right errors at the right time.

If we add custom, non-model errors to the object, and then call `#valid?` on the object, the custom errors are cleared and the model errors are applied.

```ruby
c = Course.new
c.errors
#=> []
c.errors.add(:prop, :validation)
=> #<ActiveModel::Error attribute=prop, type=valdation, options={}>
c.valid?
#=> false
 c.errors
=> #<ActiveModel::Errors [#<ActiveModel::Error attribute=course_code, type=blank, options={:on=>[:create, :update]}>, #<ActiveModel::Error attribute=maths, type=inclusion, o.................
c.errors.key?(:course_code)
#=> true
c.errors.key?(:prop)
#=> false
```

We must avoid calling valid? on an object that we have added custom errors to.

#### `course_subjects`

When we do `course.subjects = subjects` for a new record, that tries to save the records and raises an exception.

There is an ActiveRecord Validation on dupliate subjects, and also a database unique constraint on the foreign keys. So it's impossible to duplicate a subject on a course.

When this happens, the `course.subjects` does not maintain the two assigned subjects.

```ruby
 c = Course.last
 c.subjects
 #=> [#<SecondarySubject:0x00007f5be540b918 id: 10, type: "SecondarySubject", subject_code: "C1", subject_name: "Biology">,
 #<SecondarySubject:0x00007f5be540b7d8 id: 33, type: "SecondarySubject", subject_code: nil, su.......
subject = SecondarySubject.first
c.subjects = [subject, subject]
#=> ActiveRecord::RecordInvalid: Validation failed: Subject has already been taken

c.errors
#=> []

c.subjects
=> [#<SecondarySubject:0x00007f5be5400a18 id: 8, type: "SecondarySubject", subject_code: "W1", subject_name: "Art and design">]
```

It would be great to be able to use the assigned values to repopulate the form directly.

If we begin to use the `course.course_subjects` association throughout the application, we have more control over the assignment. We can build the association without creating objects.

```ruby
subject_ids.each do |subject_id|
  course.course_subjects.build(subject_id:)
end
```

```ruby
c = Course.new

c.course_subjects.build(subject_id: subject.id)
=> #<CourseSubject:0x00007f5be53758c8 id: nil, course_id: nil, subject_id: 8, created_at: nil, updated_at: nil, position: 0>

c.course_subjects.build(subject_id: subject.id)
=> #<CourseSubject:0x00007f5be5373208 id: nil, course_id: nil, subject_id: 8, created_at: nil, updated_at: nil, position: 1>

c.course_subjects
=> [#<CourseSubject:0x00007f5be53758c8 id: nil, course_id: nil, subject_id: 8, created_at: nil, updated_at: nil, position: 0>,
 #<CourseSubject:0x00007f5be5373208 id: nil, course_id: nil, subject_id: 8, created_at: nil, updated_at: nil, position: 1>]
```

We can use this to populate the form.

The main benefit of always using `course.course_subjects` over `course.subjects` is consistency between `new` and `persisted` records.

#### New
Validating Primary, Further Education and Secondary Subject inputs

[Screencast from 30-05-24 14:53:28.webm](https://github.com/DFE-Digital/publish-teacher-training/assets/135042929/cfe09e99-fc49-42ae-aafd-68248cc1b1a5)

#### Edit
Validating  Secondary Subject inputs

[Screencast from 30-05-24 14:57:16.webm](https://github.com/DFE-Digital/publish-teacher-training/assets/135042929/0f4fdda4-4db8-42f5-a6f7-7d2212985bd2)

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/models/course.rb
```diff
@@ -18,6 +18,19 @@ class Course < ApplicationRecord
...
+  after_initialize do
+    if is_primary?
+      self.master_subject_id ||= course_subjects.first&.subject_id
+    else
+      sub = course_subjects.select(&:position)
+
+      self.master_subject_id ||= sub.first&.subject&.id
+
+      self.subordinate_subject_id ||= sub&.second&.subject&.id
+    end
```
inulty-dfe, line 31: This does not account for Further Education, but there is no form inputs for further Education level courses. The `master` and `subordinate` properties are only used for input and validation.

app/services/courses/assign_program_type_service.rb
```diff
@@ -17,8 +17,6 @@ def execute(funding_type, course)
       when 'fee'
         course.program_type = calculate_fee_program(course)
       end
-      # NOTE: This looks like unwarranted side effects
-      course.save
     end
```
inulty-dfe, line 20: This is a hangover from a previous version of the application.

This service gets called during the building of a new course. When `course.save` is called, `validate` is called on the course, and then all custom errors added to the model are cleared and only the validations defined on the model are run.

This is not what we want when incrementally building up errors in the new course flow.

app/services/courses/creation_service.rb
```diff
@@ -26,6 +26,12 @@ def build_new_course
       course = provider.courses.new
       course.assign_attributes(course_attributes.except(:subjects_ids, :study_mode))
 
+      if course_attributes[:master_subject_id].blank? && course_attributes[:subordinate_subject_id].present?
+        course.errors.add(:subjects, :course_creation)
+      elsif subject_ids.present? || course_attributes[:level] == 'further_education'
+        AssignSubjectsService.call(course:, subject_ids:)
+      end
+
```
inulty-dfe, line 34: This is here so we can avoid validating the subjects unless under specific circumstances.

We don't want to call `AssignSubjectsService.call(course:, subject_ids:)` when we have not reached the subjects form (choosing primary / secondary level)
We don't want to call it if there are no subject ids either, except in the case when the level is `further education` because no subject ids are passed for this level type and the Service is responsible for assigning the FurtherEducation subject to that course.

app/services/courses/creation_service.rb
```diff
@@ -43,9 +49,8 @@ def build_new_course
         course_enrichment.course_length = '4 years'
       end
 
-      AssignSubjectsService.call(course:, subject_ids:)
+      course.valid?(:new) if course.errors.blank?
```
inulty-dfe, line 52: Avoid calling `valid?` on a model when you've added custom errors.

Calling `valid?` will clear your custom errors unless they are defined on the model as validations.

spec/features/publish/courses/editing_subject_spec.rb
```diff
@@ -13,30 +13,69 @@
...
     then_i_am_met_with_course_details_page
-    and_i_should_see_a_success_message
+    and_i_should_see_subject_updated_success_message
+  end
+
+  scenario 'updating a course with two secondary subjects' do
+    and_there_is_a_secondary_course_i_want_to_edit
+    and_the_course_has_two_subjects
+    when_i_visit_the_edit_course_subject_page
+    then_i_should_see_populated_selects('Drama', 'Latin')
```
elceebee, line 32:
```suggestion
    then_i_see_populated_selects('Drama', 'Latin')
```

I won't mark them all, but just to keep us from having to change this later if we introduce the rubocop rule that prohibits it, let's get rid of the 'should's in the tests

spec/features/publish/courses/editing_subject_spec.rb
```diff
@@ -13,30 +13,69 @@
...
+    when_i_select_a_master_subject(:latin)
+    when_i_select_a_subordinate_subject(:business_studies)
+    and_i_click_continue
+    then_i_am_met_with_course_details_page
+    and_i_should_see_subjects_updated_success_message
```
elceebee, line 37:
```suggestion
    and_i_see_subjects_updated_success_message
```

spec/features/publish/courses/editing_subject_spec.rb
```diff
@@ -13,30 +13,69 @@
...
+    when_i_select_a_master_subject(:modern_languages)
     and_i_click_continue
     then_i_am_met_with_the_modern_languages_page
   end
 
+  scenario 'updating secondary subject modern languages with multiple languages' do
+    and_there_is_a_secondary_course_i_want_to_edit
+    and_the_course_has_two_languages
+    when_i_visit_the_edit_course_subject_page
+    then_i_should_see_populated_selects('Drama', 'Modern Languages')
```
elceebee, line 53:
```suggestion
    then_i_see_populated_selects('Drama', 'Modern Languages')
```

spec/features/publish/courses/editing_subject_spec.rb
```diff
@@ -80,24 +146,43 @@ def then_i_am_met_with_course_details_page
...
+        "courses/#{course.course_code}/modern-languages?",
+        params_with_subject(extra_subject).to_s
+      ].join
+    )
     expect(page).to have_content('Languages')
   end
 
+  def then_i_am_met_with_an_error
+    within('.govuk-error-summary') do
+      expect(page).to have_content('Select a subject')
```
elceebee, line 163: This is nitpicking, but I think we should make sure we see the error in the summary and on the input -- So could do `expect(page).to have_content('Select a subject).twice` Or you can be more specific and check it is in the error summary as on the input field.

spec/features/publish/courses/editing_subject_spec.rb
```diff
@@ -13,30 +13,69 @@
     when_i_select_a_primary_subject('Primary with English')
     and_i_click_continue
     then_i_am_met_with_course_details_page
-    and_i_should_see_a_success_message
+    and_i_should_see_subject_updated_success_message
```
elceebee, line 16: I can see we have covered the further education scenarios in the unit testing -- are there features tests that cover that journey? (they might exist, but just haven't been changed, so I don't see them in this PR). Would be good to add a create / update scenario with FE if they don't already exist. That way, if we ever find ourselves in a position to do some refactoring, we can be confident we haven't broken any essential journeys.

elceebee (approved):
There are a few little things, but no blocks. You have navigated this nightmare terrain as well as possible.

spec/features/publish/courses/editing_subject_spec.rb
inulty-dfe: Good shout, I'll remember to do that in future

spec/features/publish/courses/editing_subject_spec.rb
inulty-dfe: I agree, and had to restrain myself from changing them initially.

inulty-dfe:
I'm merging now. We have tomorrow to pickup any issues. And these changes target the course creation / editing process which is highly unlikely to cause problems overnight.

inulty-dfe:
It's worth mentioning that there is an edge-case issue identified by @elceebee where setting master subject to Modern Languages and subordinate subject to Mathematics, if clicking back link, to the subjects page after selecting languages, the master and subordinate subjects are switched.

This is considered an edge case but should be documented here.

PR #4227, opened 2024-05-23
Move about this course section to its own page

elceebee:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


elceebee:
[Publish: Cut down Course Information - move Course information to its own page](https://trello.com/c/UWGjbohe/1572-publish-cut-down-course-information-move-course-information-to-its-own-page)

config/locales/en.yml
```diff
@@ -315,6 +315,34 @@ en:
...
+          candidate_say_li_4: the qualification and experience they will have at the end of the course.
+          remember_to: "Remember to:"
+          remember_to_li_1: keep it brief and to the point, people struggle with long blocks of writing online
+          remember_to_li_2: use bullet points, headings and paragraphs to make your writing easy to read
+          remember_to_li_3: spell out acronyms the first time you use them, for example, ITT, NQT, SCITT
+          remember_to_li_4: >
+            link to your organisation's website for people who want more detail about who you are and what you do.
+          view_examples: View examples of great course summaries
+          changes_not_saved: Your changes are not yet saved
+          copied_fields_from: We've copied this field from %{name_and_code}.
```
elceebee, line 343: Change to 'We have' instead of We've

app/views/publish/courses/about_this_course/_guidance.html.erb
```diff
@@ -0,0 +1,23 @@
+<p class="govuk-body"><%= t("publish.providers.about_course.edit.description") %></p>
+
+<p class="govuk-body"><%= t("publish.providers.about_course.edit.candidates_say") %></p>
+
+<ul class="govuk-list govuk-list--bullet">
+  <li><%= t("publish.providers.about_course.edit.candidate_say_li_1") %></li>
+  <li><%= t("publish.providers.about_course.edit.candidate_say_li_2") %></li>
+  <li><%= t("publish.providers.about_course.edit.candidate_say_li_3") %></li>
+  <li><%= t("publish.providers.about_course.edit.candidate_say_li_4") %></li>
```
inulty-dfe, line 9: Is there a reason this isn't scoped to `publish.courses.about_course`. I don't think this is important, just wondering.

elceebee: It's how the edit pages are scoped in the translation file, but it would be nice to change in the future.

app/views/publish/courses/about_this_course/edit.html.erb
```diff
@@ -0,0 +1,71 @@
...
+        <%= govuk_back_link_to(
+              back_link_path(param_form_key: f.object_name.to_sym,
+                             params:,
+                             provider_code: @provider.provider_code,
+                             recruitment_cycle_year: @course.recruitment_cycle_year,
+                             course_code: @course.course_code)
+            ) %>
+      <% end %>
+
+      <%= f.govuk_error_summary %>
```
avinhurry, line 34: This should be full width: 

image


Similar to this: 

image

avinhurry (commented):
Looking good so far, i've tested with rollover and it works as expected. Added a few minor comments in line.

I also noticed a quirk, i'm not sure if it's expected?

You have the copy course option when you come to the page: 

image

but when validation kicks in, you lose it 🤔 

image

app/views/publish/courses/about_this_course/edit.html.erb
elceebee: hmmm.... good question. I'll make the change

elceebee: Actually, the error component needs to sit within the form I think.

elceebee: Also, it's not how it works on any of the existing pages. I think if we want to change it, it should be done in a separate ticket and fix throughout.

elceebee:
> Looking good so far, i've tested with rollover and it works as expected. Added a few minor comments in line.
> 
> I also noticed a quirk, i'm not sure if it's expected?
> 
> You have the copy course option when you come to the page:
> 
> image
> but when validation kicks in, you lose it 🤔
> 
> image

This is a bug in the copy content component -- it happens in all the areas it is used. I think to fix it here would open a can of worms and make this PR unmanaageable. I'll create a separate bug ticket so it can be addressed throughout.

avinhurry (approved):
Awesome work! Love the video too 🎉

PR #4223, opened 2024-05-21
[1697] TDA edge cases

avinhurry:
### Context

We need to ensure the attributes are correct for TDA courses and non TDA courses.

### Changes proposed in this pull request

There are certain attributes which are set by default on TDA courses, so when a user changes from a non TDA course to a TDA course we need to ensure these attributes are correctly set:

- `study_mode: 'full_time'`
- `funding_type: 'apprenticeship'`
- `can_sponsor_student_visa: false`
- `can_sponsor_skilled_worker_visa: false`

When a user changes from a TDA course to a non TDA course, we need to step the user back through the options which were previously defaulted.

The controllers are heavy, we will think of a good way to refactor in a following PR as there are a lot of changes in here already. 

### Guidance to review

Try and break it. Contact me if you'd like a demo.

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history
- [x] Tested by running locally
- [x] Inform data insights team due to database changes

PR #4250, opened 2024-06-04
Remove rollover activation from review app

avinhurry:
### Context

Remove the rollover activation in the review app.

This was used in this commit: https://github.com/DFE-Digital/publish-teacher-training/pull/4223/commits/d8d32025e8a95087af672a685a75d6a1b7fbf841

It should have been removed before merging, so it doesn't confuse people testing on other review apps.

### DIFF
--- a/config/settings/review_aks.yml
+++ b/config/settings/review_aks.yml
@@ -13,13 +13,6 @@ search_ui:
 
 features:
   teacher_degree_apprenticeship: true
-  rollover:
-    # Normally a short period of time between rollover and the next cycle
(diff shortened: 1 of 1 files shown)


tomas-stefano (approved):
With the rollover upcoming, the rollver active in review apps, might not be a problem. But all good to approve :)

avinhurry:
> With the rollover upcoming, the rollver active in review apps, might not be a problem. But all good to approve :)

Na we should remove it because the review app should reflect our live service, and we also need to rollover the database for it to work properly.

PR #4255, opened 2024-06-05
[1719] Add blazer to Publish

avinhurry:
### Context

Add blazer to publish.

- Put behind an admin constraint
- Only allow selected admin users to access (hardcoded ID's)
- We will set an environment variable `BLAZER_ALLOWED_IDS` in production.

### Guidance to review

https://publish-review-4255.test.teacherservices.cloud/blazer

- Does it work?
- Is it secure enough?

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history


avinhurry:
[Add blazer to Find / Publish service for support](https://trello.com/c/NYobnXeY/1719-add-blazer-to-find-publish-service-for-support)

app/lib/blazer_admin_constraint.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  private
+
+  def authorised_for_blazer?(admin_user)
+    allowed_ids.blank? || allowed_ids.split(',').include?(admin_user.id.to_s)
+  end
+
+  def allowed_ids
+    ENV.fetch('BLAZER_ALLOWED_IDS', nil)
```
elceebee, line 19: Why have we chosen to store this in an environment variable? 

I think I would prefer it to be a column on the user model. Then we (as devs) can just use the rails console to add / remove blazer access, rather than the rather cumbersome step of updating environment variables and redeploying the app. If we keep the logic on the user, we can essentially delete lines 12 - 19 and change lines 8-9 to:
`User.admins_with_blazer_access.find_by(email: singin_user.email).present?` 
(where admins_with_blazer_access) is a scope on the User model)

The implementation here also means our default position is that all admin users have blazer access (if there is no env var). Is that what we want, or would we prefer that blazer access is always restricted to particular users? 

I'm not going to block this over it, but I'm curious as to why we have chosen to implement it this way. In general, it feels like making this dependent on an env var spreads the logic about who is / isn't a blazer user to more places than it needs to be. Making it harder to debug and read in the future.

app/lib/blazer_admin_constraint.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class BlazerAdminConstraint
+  def matches?(request)
+    signin_user = UserSession.load_from_session(request.session)
+    return false if signin_user.blank?
+
+    admin_user = User.admins.kept.find_by(email: signin_user.email)
```
inulty-dfe, line 8: [`exists?` would be cheaper here](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F)

avinhurry: Thanks - have opened a new PR and updated there: https://github.com/DFE-Digital/publish-teacher-training/pull/4277/commits/7e4476256b149765ec57dfb4a058683aad519915

avinhurry:
[Add blazer to Find / Publish Support Console](https://trello.com/c/NYobnXeY/1719-add-blazer-to-find-publish-support-console)

app/lib/blazer_admin_constraint.rb
avinhurry: Agreed to go with this approach offline.

avinhurry:
Closing this, in favour of https://github.com/DFE-Digital/publish-teacher-training/pull/4277/files

avinhurry:
[Add blazer to Find / Publish Support Console](https://trello.com/c/NYobnXeY/1719-add-blazer-to-find-publish-support-console)

avinhurry:
[Determine the security requirements and implications of using Blazer in Apply Support Console](https://trello.com/c/sqtnQIBI/103-determine-the-security-requirements-and-implications-of-using-blazer-in-apply-support-console)

PR #4261, opened 2024-06-07
Refactors Program Types and Funding Types

dcyoung-dev:
### Context

While investigating how the Funding and Program types, it was difficult to follow all the logic that relied on program_types.

This PR introduces a Program model and models for each type of Program. These hold the information about their funding types, if they are salaried, and if the sponsor visas.

This keeps all the logic for each of these in the same place rather than peppered throughout the Course model

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -519,16 +519,12 @@ class Course < ApplicationRecord
     :not_running
   end
 
-  def funding_type
-    return if program_type.nil?
+  def program
+    Program.from_type(program_type) if program_type.present?
+  end
 
-    if school_direct_salaried_training_programme? || scitt_salaried_programme? || higher_education_salaried_programme?
-      'salary'
-    elsif pg_teaching_apprenticeship? || teacher_degree_apprenticeship?
-      'apprenticeship'
-    else
-      'fee'
-    end
+  def funding_type
+    program.funding_type if program.present?
   end
 
   def is_fee_based?
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -288,21 +288,7 @@ class Course < ApplicationRecord
   }
 
   scope :with_funding_types, lambda { |funding_types|
-    program_types = []
-
-    program_types << :school_direct_salaried_training_programme if funding_types.include?('salary')
-
-    program_types << :pg_teaching_apprenticeship if funding_types.include?('apprenticeship')
-
-    if funding_types.include?('fee')
-      %i[
-        higher_education_programme
-        scitt_programme
-        school_direct_training_programme
-      ].each do |program_type|
-        program_types << program_type
-      end
-    end
+    program_types = Program.where_funding_types(funding_types)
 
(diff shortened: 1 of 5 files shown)


app/models/program.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
+
+class Program
+  def self.all
```
elceebee, line 4: What do you think about adding `class << self` so we don't have `self.` on all the methods. Just a personal preference, keeps things a bit easier to read for me, but I don't feel strongly about it.

app/models/course.rb
```diff
@@ -248,9 +248,8 @@ def new_draft_attributes
   scope :findable, -> { joins(:site_statuses).merge(SiteStatus.findable) }
   scope :with_vacancies, -> { joins(:site_statuses).merge(SiteStatus.with_vacancies) }
   scope :with_salary, lambda {
-    where(
-      program_type: %i[school_direct_salaried_training_programme pg_teaching_apprenticeship scitt_salaried_programme higher_education_salaried_programme]
-    )
+    program_types = Program.where_salaried
+    where(program_type: program_types)
```
tomas-stefano, line 252: Small thing I would put everything under one line. `where(program_type: Program.where_salaried)` as this is easy to read.

spec/models/program_spec.rb
```diff
@@ -0,0 +1,280 @@
...
+  end
+
+  describe '.sponsors_student_visa?' do
+    it { expect(described_class.sponsors_student_visa?).to be_falsey }
+  end
+
+  describe '.sponsors_skilled_worker_visa?' do
+    it { expect(described_class.sponsors_skilled_worker_visa?).to be_falsey }
+  end
+end
```
tomas-stefano, line 280: Should we split into multiple files, or is there a preference to keep all same classes under the same file and spec?

app/models/program.rb
dcyoung-dev: I don't love this syntax but have updated to match 👌

spec/models/program_spec.rb
dcyoung-dev: This was a preference over having the text `it 'returns false'` seemed a waste of key presses 🤣

app/models/course.rb
dcyoung-dev: I feel the separate lines are easier to read - not fussed either way 🤷‍♂️

app/controllers/publish/courses/funding_type_controller.rb
```diff
@@ -79,7 +79,7 @@ def next_path
       end
 
       def visa_sponsorship_path
-        if course.is_fee_based?
+        if course.fee_based?
```
elceebee, line 82: Yes!

elceebee (approved):
What a wonderful refactor! 🥇

PR #4284, opened 2024-06-18
[1749] Ensure back links redirect correctly on TDA courses

avinhurry:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


avinhurry:
[Ensure back links redirect correctly on TDA courses (part 2)](https://trello.com/c/s1vqY6uf/1749-ensure-back-links-redirect-correctly-on-tda-courses-part-2)

app/views/publish/courses/student_visa_sponsorship/edit.html.erb
```diff
@@ -9,14 +9,21 @@
...
           course.course_code
         )
     ) %>
+  <% elsif params[:previous_tda_course] == 'true' %>
+    <%= govuk_back_link_to(full_part_time_publish_provider_recruitment_cycle_course_path(
+              course.provider_code,
+              course.recruitment_cycle_year,
+              course.course_code,
+              previous_tda_course: true
+            )) %>
```
elceebee, line 26: This erb file is getting out of hand with logic. We've got 33 lines of an if / else statement here, making this very difficult to read. The erb files should just place the elements on the page and logic should be very minimal.

There are a few different ways to do it -- perhaps in the decorator, the form or the controller. 

But it should look more like:
```
  <%= govuk_back_link_to(back_link)
  # or   <%= govuk_back_link_to(course.back_link(:student_visa_sponsorship)) 
<% end %>
```

The logic itself is rather difficult to read. If you are checking something is not true and then checking that it is true in the next condition, this will be difficult for future you / other developers to comprehend later without drawing a complex venn diagram.

You should get the same results with 
```
def back_link
  if params[:back_to_basic_details].present?
    ...
  elsif params[:previous_tda_course] == 'true'
    ...
  elsif @provider.accredited_provider?
    ...
  else 
    ...
  end
end
```

app/views/publish/courses/study_mode/edit.html.erb
```diff
@@ -2,13 +2,22 @@
...
+          )) %>
+  <% else %>
+    <%= govuk_back_link_to(
+          details_publish_provider_recruitment_cycle_course_path(
+            course.provider_code,
+            course.recruitment_cycle_year,
+            course.course_code
+          )
+        ) %>
+  <% end %>
```
elceebee, line 20: Though the logic is less complex here (so I wouldn't necessarily block it over this), it's not a pattern we want to expand. Generally something to be cautious about. If your erbs are full of logic, are they actually just different views?

spec/features/publish/courses/editing_course_outcome_spec.rb
```diff
@@ -50,8 +50,21 @@
...
+        and_the_back_link_points_to_the_study_mode_page
+
+        and_i_click_back
+        then_i_am_on_the_study_mode_page
+
+        and_i_click_back
+        then_i_am_on_the_funding_type_page
+
+        and_i_update
+        and_i_update
```
elceebee, line 66:
```suggestion
        and_i_update
```
```suggestion
        when_i_update
        and_i_update
```

This might make it clear that you didn't just accidentally copy and paste twice.

elceebee (changes requested):
I think it's important to start using some cleaner patterns within the erb files.

app/views/publish/courses/student_visa_sponsorship/edit.html.erb
avinhurry: Discussed offline, and decided to go with: https://github.com/DFE-Digital/publish-teacher-training/pull/4284/commits/3acddbbd14034e43ec8b9a690907e634eefd2e3d

spec/features/publish/courses/editing_course_outcome_spec.rb
avinhurry: Done: https://github.com/DFE-Digital/publish-teacher-training/pull/4284/commits/7a007a75f39a628f5e01434428272e7d3e80323d

app/views/publish/courses/study_mode/edit.html.erb
avinhurry: Done here too: https://github.com/DFE-Digital/publish-teacher-training/pull/4284/commits/3acddbbd14034e43ec8b9a690907e634eefd2e3d

elceebee (approved):
Great work!

PR #4294, opened 2024-06-20
TDA A level requirements. - part 2 (second page in the wizard)

tomas-stefano:
### Context

We need to add the ability for providers to input A level requirements on TDA courses on Publish, in order to enable TDA applications.

A new view is created asking what subjects and grades are required, which is accessed when a user selects “Yes” to the previous page in the flow

The back link works and have the yes chosen.

### Guidance to review

1. Does validation works?
2. Error links work?
3. Back link works?


app/wizards/a_level_steps/what_a_level_is_required.rb
```diff
@@ -2,23 +2,23 @@
...
+    end
 
     # def next_step
-    #   if answer == 'yes'
-    #     :third_step
-    #   elsif answer == 'no'
-    #     :second_step
-    #   end
+    #  :exit
     # end
```
tomas-stefano, line 22: There is not next step here because I am adding on the next card (still working on it). If everything is valid you will see a not implemented error. Again next card, next PR.

app/views/publish/courses/a_level_requirements/what_a_level_is_required/new.html.erb
```diff
@@ -1,23 +1,48 @@
...
+<% content_for :page_title, title_with_error_prefix(t("course.#{@wizard.current_step.model_name.i18n_key}.heading"), @wizard.current_step.errors && @wizard.current_step.errors.any?) %>
 
 <% content_for :before_content do %>
-  <%= govuk_back_link_to("change-me") %>
+  <%= govuk_back_link_to(
+      publish_provider_recruitment_cycle_course_a_levels_are_any_a_levels_required_for_this_course_path(
+        @provider.provider_code,
+        @provider.recruitment_cycle_year,
+        @course.course_code,
+        are_any_a_levels_required_for_this_course: { answer: "yes" }
```
tomas-stefano, line 9: The next PR I am sorting out the params being passed from back and next steps. I wait for the third part because then I am having a deep thinking how to organise passing the params in each steps back and forward and then passing back to the instance of Wizard and the steps. So for now I just hardcoded the yes answer but this will be deleted/changed from this next PR.

elceebee: I don't have any particular ideas, but if you want to talk through your deep thinking, let me know, happy to chat it through before you tackle it.

elceebee (approved):
I have tested this locally and things work as expected. I have two questions:
- Is it deliberate that we do not have a cancel link on this page? 
- You are in conversation with Rose and Kellie about the fact that we don't have any length restriction on the 'Minimum grade required'. I'm happy for this to be merged in as is if we need to wait for a response, but we definitely want to address this before it goes live. 

Great work! 🚀

tomas-stefano:
> Is it deliberate that we do not have a cancel link on this page?

Yes, after some discussions with Kellie and Rose they opt-in to only add the cancel into the first step.

> You are in conversation with Rose and Kellie about the fact that we don't have any length restriction on the 'Minimum grade required'. I'm happy for this to be merged in as is if we need to wait for a response, but we definitely want to address this before it goes live.

Yes, if they reply back I will open another PR with the max length validation. (I will add one small card for it). I will merge this then.

PR #4295, opened 2024-06-20
Implement Copy from for Fees and financial support

inulty-dfe:
### Context

Provide the ability for providers to copy the fees and financial support content from one course in the same recruitment cycle to another.

Ensure that the copy feature is available to the user even after validation fails.
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


inulty-dfe:
[Add "Copy course content" to the remaining new pages split out from 'course length and fees'](https://trello.com/c/LHVFRFcR/1721-add-copy-course-content-to-the-remaining-new-pages-split-out-from-course-length-and-fees)

spec/features/publish/courses/editing_course_fee_details_copy_content_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+  def given_i_am_authenticated_as_a_provider_user
+    given_i_am_authenticated(user: create(:user, :with_provider))
+  end
+
+  def and_there_is_a_course_i_want_to_edit
+    given_a_course_exists(:secondary, additional_degree_subject_requirements: nil, degree_subject_requirements: nil)
+  end
+
+  def when_i_submit_without_data
+    fill_in 'Fees and financial support (optional)', with: 'a ' * 251
```
elceebee, line 40: If you are submitting without data, shouldn't it be:
```suggestion
    fill_in 'Fees and financial support (optional)', with: ''
```

inulty-dfe: Ah good spot. The name of the method is wrong, this is an optional field so need to provide too many words to trigger validation.

elceebee (approved):
Nice work!

avinhurry (approved):
Looks good to me! 

I noticed that if you try to copy content from a course which doesn't have any content in that field, it doesn't error. 

Do you think it should? If so i'll create a card for it.

inulty-dfe:
> Looks good to me!
> 
> I noticed that if you try to copy content from a course which doesn't have any content in that field, it doesn't error.
> 
> Do you think it should? If so i'll create a card for it.

That's a product question if you ask me

PR #4296, opened 2024-06-20
Rubocop changes

inulty-dfe:
### Context

Rubocop updated to v3 recently, some things were not attended to during the upgrade.

Also disable numbers in variables cop, it is not helpful

1. Rubocop RSpecRails exists now: Add it to the gem file and require it in rubocop.yml
2. Add `RSpecRails/HaveHttpStatus` and `RSpec/Dialect` back in
3. ignore restrictions around numbers in variable names.

### DIFF
--- a/config/rubocop/naming.yml
+++ b/config/rubocop/naming.yml
@@ -13,12 +13,9 @@ Naming/MethodName:
   Exclude:
     - 'app/models/accrediting_provider_enrichment.rb'
 
+# Allow numbers in variable names
 Naming/VariableNumber:
-  Exclude:
-    - 'app/serializers/api/public/v1/serializable_location.rb'
-    - 'app/serializers/api/public/v1/serializable_provider.rb'
-    - 'spec/serializers/api/public/v1/serializable_location_spec.rb'
-    - 'spec/serializers/api/public/v1/serializable_provider_spec.rb'
+  Enabled: false
 
 Naming/PredicateName:
   Exclude:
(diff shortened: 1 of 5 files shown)


config/rubocop/naming.yml
```diff
@@ -13,12 +13,9 @@ Naming/MethodName:
...
     - 'app/models/accrediting_provider_enrichment.rb'
 
+# Allow numbers in variable names
 Naming/VariableNumber:
-  Exclude:
-    - 'app/serializers/api/public/v1/serializable_location.rb'
-    - 'app/serializers/api/public/v1/serializable_provider.rb'
-    - 'spec/serializers/api/public/v1/serializable_location_spec.rb'
-    - 'spec/serializers/api/public/v1/serializable_provider_spec.rb'
+  Enabled: false
```
avinhurry, line 18: Did this cause a lot more complaints? Is it not a rule you see value in?

inulty-dfe: It make you write

```ruby
def my_method2025
```

Instead of 

```ruby
def my_method_2025
```

I don't see the value in enforcing this at all

inulty-dfe: I've just noticed this commit 
https://github.com/DFE-Digital/publish-teacher-training/commit/d20a13274f15fa92018a5f0ef0cfae13a8ed9b0e

inulty-dfe:
I've just noticed this commit 
https://github.com/DFE-Digital/publish-teacher-training/commit/d20a13274f15fa92018a5f0ef0cfae13a8ed9b0e

avinhurry (approved):

PR #4297, opened 2024-06-20
Activate rollover in this review app

avinhurry:
### Context

Adding this to allow people to play around and become familiar with how Publish will look during rollover.


avinhurry:
Testing is done and rollover into the 2024-2025 cycle is complete in all environments, closing this now as it's no longer required.

PR #4299, opened 2024-06-21
Find Domain change - We will have to sustain 2 domains until the domain redirect is applied

tomas-stefano:
## Context

Due to undergraduate courses appearing we will change Find domain.

Because this repo contain several domains into the same app we need to change the application logic to allow a new domain.

We’ll look to handle multiple domains (to enable a seamless transition from one domain to the other).

1. Allow multiple domains to function
2. Infra team to set up redirect from old to new domain
3. We remove this second domain and renamed the old domain to the new one in the app.

Allow multiple domains to function

### Guidance to review

1. Does it needs change anywhere else?

### DIFF
--- a/app/lib/find_constraint.rb
+++ b/app/lib/find_constraint.rb
@@ -2,6 +2,6 @@
 
 class FindConstraint
   def matches?(request)
-    Settings.find_url&.include?(request.host) || request.host.include?('find-pr') || request.host.include?('find-review')
+    Settings.find_url&.include?(request.host) || Settings.extra_find_url&.include?(request.host) || request.host.include?('find-pr') || request.host.include?('find-review')
   end
 end
--- a/app/lib/find_constraint.rb
+++ b/app/lib/find_constraint.rb
@@ -2,6 +2,6 @@
 
 class FindConstraint
   def matches?(request)
-    Settings.find_url&.include?(request.host) || Settings.extra_find_url&.include?(request.host) || request.host.include?('find-pr') || request.host.include?('find-review')
+    request.host.include?('find')
   end
 end
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -34,13 +34,7 @@ Rails.application.configure do
   config.assets.compile = false
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  config.action_controller.asset_host = proc do |_source, request|
-    if Settings.find_url.include?(request.host)
-      Settings.find_assets_url
-    elsif Settings.extra_find_url&.include?(request.host)
-      Settings.extra_find_assets_url
-    end
-  end
+  # config.action_controller.asset_host = proc { |_source, request| }
 
   # Specifies the header that your server uses for sending files.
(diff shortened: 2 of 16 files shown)


config/settings/review.yml
```diff
@@ -7,6 +7,7 @@ environment:
 
 publish_url: https://qa.publish-teacher-training-courses.service.gov.uk
 find_url: https://qa.find-postgraduate-teacher-training.service.gov.uk
+extra_find_url: https://qa.find-teacher-training-courses.service.gov.uk
```
saliceti, line 10: This is probably not used since the review domain is set to https://find-review-4299.test.teacherservices.cloud/

app/lib/find_constraint.rb
```diff
@@ -2,6 +2,6 @@
 
 class FindConstraint
   def matches?(request)
-    Settings.find_url&.include?(request.host) || request.host.include?('find-pr') || request.host.include?('find-review')
+    Settings.find_url&.include?(request.host) || Settings.extra_find_url&.include?(request.host) || request.host.include?('find-pr') || request.host.include?('find-review')
```
saliceti, line 5: The domain for review is https://find-review-4299.test.teacherservices.cloud/ so `request.host.include?('find-pr')` can be removed

saliceti: It also looks wrong in the test: https://github.com/DFE-Digital/publish-teacher-training/blob/52f20f4d0e29531b53754341d46be60606d529fb/spec/lib/find_constraint_spec.rb#L40

saliceti: Could the whole thing be simplified with something like `request.host.start_with? "find"` ?

tomas-stefano: Yes, it makes sense. I was conservative in the solution because I didn't wanna break things.

But I always up for simple solutions

tomas-stefano: Done!

config/settings/review.yml
tomas-stefano: With the simplified implementation I remove the extra_find_url. So now is good

saliceti:
I see the find_url setting referenced in multiple places, including https://github.com/DFE-Digital/publish-teacher-training/blob/f1d9f9d9819bbc85dc724984d57273acd657041c/config/initializers/cors.rb#L5
Do they need to be changed now? Or can they be changed after the domain change?

saliceti (approved):

config/environments/production.rb
```diff
@@ -34,9 +34,7 @@
   config.assets.compile = false
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  config.action_controller.asset_host = proc do |_source, request|
-    Settings.find_assets_url if Settings.find_url.include?(request.host)
-  end
```
tomas-stefano, line 39: Just for the record, the asset host in all environments is nil. This is a leftover from Gov.uk PaaS transition to AKS. So removing it to not confuse devs (like I got confused and wasting time).

saliceti (approved):

PR #4312, opened 2024-06-26
[1873] Render 404 if course should not be on Find

avinhurry:
### Context

Courses which are draft, rolled over and withdrawn should not be on Find. Currently you can access these courses if you manually type in the URL.

These courses are already hidden from the Find search and wont be shown in the results page.

Render a 404 if the course is in a state where it should not be on Find.

### Before

- https://qa.publish-teacher-training-courses.service.gov.uk/publish/organisations/4Y1/2024/courses/E588
- https://qa.find-postgraduate-teacher-training.service.gov.uk/course/4Y1/E588

image

### After

- https://publish-review-4312.test.teacherservices.cloud/publish/organisations/4Y1/2024/courses/E588
- https://find-review-4312.test.teacherservices.cloud/course/4Y1/E588

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/find/courses_controller.rb
+++ b/app/controllers/find/courses_controller.rb
@@ -14,6 +14,8 @@ module Find
         subjects: [:financial_incentive],
         site_statuses: [:site]
       ).find_by!(course_code: params[:course_code]&.upcase).decorate
+
+      render_not_found unless @course.is_published?
     end
   end
 end
--- a/spec/features/find/viewing_courses_in_various_states_spec.rb
+++ b/spec/features/find/viewing_courses_in_various_states_spec.rb
@@ -1,0 +1,111 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+feature 'viewing courses in various states' do
+  scenario 'viewing a published course' do
+    given_there_is_a_published_course
+    when_i_visit_the_course_page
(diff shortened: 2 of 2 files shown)


avinhurry:
[Courses in non-published states can be accessed directly (via URL) on Find](https://trello.com/c/rtE0IP4z/1873-courses-in-non-published-states-can-be-accessed-directly-via-url-on-find)

CatalinVoineag (approved):

app/controllers/find/courses_controller.rb
```diff
@@ -14,6 +14,8 @@ def show
         subjects: [:financial_incentive],
         site_statuses: [:site]
       ).find_by!(course_code: params[:course_code]&.upcase).decorate
+
+      render_not_found unless @course.is_published?
```
dcyoung-dev, line 18: Could we tack on `.published` to the `provider.courses` scope rather than handling this separately?
Alternatively should we use the `CourseSearchService` rather than directly accessing Courses 🤔

avinhurry: This will be considered as a potential follow up so that we have resolved the issue at hand.

PR #4313, opened 2024-06-26
[453] Omit course start day field on support

avinhurry:
### Context

Our support console in Publish allows someone to set a course as any day in any month.

However, if the provider went through the normal flow of creating a course, this is impossible. We default this to the first of the month. We should align the two.

We do not render this to the candidate, so it will not make any difference to candidates, but will help keep our data consistent.

There will be follow up work to set all course start days to the 1st of the month.

Remove the day field from support

### Before

image

### After

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/support/courses/edit.html.erb
+++ b/app/views/support/courses/edit.html.erb
@@ -20,7 +20,7 @@
         <% end %>
       </div>
 
-      <%= f.govuk_date_field(:start_date, legend: { text: "Start date" }) %>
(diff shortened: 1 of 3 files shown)


avinhurry:
[Publish Support Console allows DAY component of the course start date to be set](https://trello.com/c/jdZQA2Ai/453-publish-support-console-allows-day-component-of-the-course-start-date-to-be-set)

PR #4331, opened 2024-07-02
Move school placements into its own view

CatalinVoineag:
### Context

Currently, school placements are listed on course pages, which can be
problematic for candidates as there is quite a lot of content on this
page.
We want to move this information to a separate page.

New route/controller and view for the placements.
Link to the new view from the course show page, in find and publish

### Guidance to review

Review the show page of the course in publish(preview) and in find.
You should see a link to the school placements

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

https://github.com/DFE-Digital/publish-teacher-training/assets/11318084/b35474b1-9ff6-4d7f-b351-7fdd9ac206e4

### DIFF
--- a/app/components/find/courses/about_schools_component/view.rb
+++ b/app/components/find/courses/about_schools_component/view.rb
@@ -38,6 +38,21 @@ module Find
           course_information_config.show_placement_guidance?(:program_type)
         end
 
+        def placements_url
+          if course.has_unpublished_changes? || (course.is_published? && course.is_running?)
+            URI.join(
+              Settings.search_ui.base_url,
+              find_placements_path(course.provider_code, course.course_code)
+            ).to_s
+          else
+            placements_publish_provider_recruitment_cycle_course_path(
+              course.provider_code,
+              course.recruitment_cycle_year,
+              course.course_code
+            )
+          end
+        end
+
         private
 
         def course_information_config
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,4 +1,11 @@
 en:
+  shared:
+    courses:
+      placements:
+        heading: School placements at %{provider_name}
+        warning: You can’t pick which schools you will be in.
+        will_place_you: '%{provider_name} will place you in different schools you can travel to during your training.'
+        schools_to_be_placed: 'The schools you could be placed in are:'
   markdown_formatting:
     summary_text: Help formatting your text
     title: How to format your text
@@ -389,6 +396,9 @@ en:
   publish:
     providers:
       courses:
+        placements:
+          heading: School placements at %{provider_name}
+          back: Back to %{course_name} (%{course_code})
         description_content:
           course_information_heading: Course information
           about_course_label: About this course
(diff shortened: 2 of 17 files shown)


CatalinVoineag:
[Simplify course pages 3/6 - Move school placements to a separate page](https://trello.com/c/8sLNQ913/1761-simplify-course-pages-3-6-move-school-placements-to-a-separate-page)

app/components/find/courses/about_schools_component/view.rb
```diff
@@ -38,6 +38,21 @@ def show_scitt_guidance?
           course_information_config.show_placement_guidance?(:program_type)
         end
 
+        def placements_url
+          if course.has_unpublished_changes? || (course.is_published? && course.is_running?)
+            URI.join(
+              Settings.search_ui.base_url,
```
CatalinVoineag, line 44: I think this way of setting the URL breaks review app when it comes to redirection to find. Should work fine in production I think. We use the same logic for `View on find` button

elceebee: How odd...

config/locales/en.yml
```diff
@@ -1,4 +1,11 @@
 en:
+  shared:
+    courses:
+      placements:
+        heading: School placements at %{provider_name}
+        warning: You can't pick which schools you will be in.
```
elceebee, line 6: Tiny tiny little thing -- in this codebase, we seem to use smart quotes. In the other codebase, we would say 'cannot'. Might be worth checking with Ruby what she would prefer.

elceebee (approved):
Great work. 🚀

app/controllers/publish/courses_controller.rb
```diff
@@ -24,6 +24,11 @@ def show
       flash.delete(:error_summary)
     end
 
+    def placements
```
dcyoung-dev, line 27: How do we feel about introducing a new controller for Placements and using a `#show` action probably a `resource :placements` in the routes?

I know the Publish codebase has a number of "non-RESTful" actions and we don't need to add to them 🤞 if we don't have to.

config/routes/find.rb
```diff
@@ -13,6 +13,7 @@
   get '/privacy', to: 'pages#privacy', as: :privacy
   get '/terms-conditions', to: 'pages#terms', as: :terms
   get '/course/:provider_code/:course_code', to: 'courses#show', as: 'course'
+  get '/placements/:provider_code/:course_code', to: 'courses#placements', as: 'course_placements'
```
dcyoung-dev, line 16: Similarly to my other comment, is there a reason for not having this as
```suggestion
  get '/course/:provider_code/:course_code/placements', to: 'placements#show', as: 'course_placements'
```

kelliedesigner:
Looks good to me!

config/locales/en.yml
CatalinVoineag: Actually, in the ticket, it uses smart quotes. Good spot, thank you

app/controllers/publish/courses_controller.rb
CatalinVoineag: I've put the actions in their own controller(index action). I'm not sure about using resources, the way the routes are setup makes it difficult.

On find we need these two params `/course/:provider_code/:course_code/placements`
And on publish I have to use `/placements` rather than `/school-placements` as the path is already taken, the controller is called `school-placements`, it's an existing controller.

dcyoung-dev (approved):
Nice work 👌

PR #4360, opened 2024-07-10
[2027] Update apply deadline banner in publish

elceebee:
### Context
We’ve identified some banners in publish that need updating as part of the end of cycle.

This ticket is just related to the banner users see after 30 July, alerting them of the apply deadline.

We [agreed in this slack thread](https://ukgovernmentdfe.slack.com/archives/C077NRS8K8D/p1720520008204449) that we would use the same content as the banner we show candidates who are applying for courses

https://trello.com/c/myMbTZ4Q

- Updated text
- Added preview for deadline banner

Before
 image 

After
image 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/components/find/deadline_banner_component.rb
+++ b/app/components/find/deadline_banner_component.rb
@@ -22,6 +22,10 @@ module Find
       cycle_timetable.apply_deadline.to_fs(:govuk_date_and_time)
     end
 
+    def cycle_year_range
+      cycle_timetable.cycle_year_range
+    end
+
     def find_opens
       cycle_timetable.find_opens.to_fs(:month_and_year)
     end
--- a/spec/components/find/deadline_banner_component_preview.rb
+++ b/spec/components/find/deadline_banner_component_preview.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module Find
+  class DeadlineBannerComponentPreview < ViewComponent::Preview
+    def show_apply_deadline_banner
+      render(
+        Find::DeadlineBannerComponent.new(
+          flash_empty: true,
+          cycle_timetable: ShowApplyDeadlineBannerCycleTimetable
+        )
+      )
+    end
+
+    def show_cycle_closed_banner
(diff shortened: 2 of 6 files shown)


elceebee:
[Update Apply deadline banner in publish](https://trello.com/c/myMbTZ4Q/2027-update-apply-deadline-banner-in-publish)

app/components/find/deadline_banner_component.rb
```diff
@@ -22,6 +22,10 @@ def apply_deadline
       cycle_timetable.apply_deadline.to_fs(:govuk_date_and_time)
     end
 
+    def cycle_year_range
```
avinhurry, line 25: So that these dates don't get mixed up, Instead of defining another method. could we use: 

https://github.com/DFE-Digital/publish-teacher-training/blob/ffd59fa1d2d87c770e8dd1420dacb6c33c8c47bf/app/helpers/recruitment_cycle_helper.rb#L8-L10

spec/components/find/deadline_banner_component_preview.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Find
+  class DeadlineBannerComponentPreview < ViewComponent::Preview
```
avinhurry, line 4: Awesome! 🙌🏼

app/components/find/deadline_banner_component.rb
elceebee: I'm not sure the value of creating a new method. Would it be ok to just 
```suggestion
    def next_recruitment_cycle_period_text
      cycle_timetable.cycle_year_range
    end
```

avinhurry: Had a chat about this offline, we decided it would be a good idea to start referencing the cycle timetable for cycle related dates such as this.

avinhurry (approved):

PR #4364, opened 2024-07-11
Give a course with no enrichments being rolled over an enrichment

avinhurry:
### Context

When we manually rollover a course it has a `content_status` of `draft` but we want a manually rolled over course to have a `content_status` of `rolled_over`.

They are currently `draft` because they don't have any enrichments and in these cases we default the `content_status` to `draft`.

The code that causes the `content_status` to get set to rolled over is never hit in `content_service_status.rb` because there is no enrichment to `dup`.

Give a course being rolled over an enrichment if it doesn't have any enrichments.

### Guidance to review

Does this cause any unexpected side effects? 

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/courses/copy_to_provider_service.rb
+++ b/app/services/courses/copy_to_provider_service.rb
@@ -44,8 +44,9 @@ module Courses
     end
 
     def copy_latest_enrichment_to_course(course, new_course)
+      course.enrichments << CourseEnrichment.new(course:, status: 'draft') if course.enrichments.blank?
+
       last_enrichment = course.enrichments.most_recent.first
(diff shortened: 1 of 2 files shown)


avinhurry:
[Bug: A manually rolled over course is appearing as "Draft" rather than "Rolled Over" in the next cycle](https://trello.com/c/lqURRzFM/2006-bug-a-manually-rolled-over-course-is-appearing-as-draft-rather-than-rolled-over-in-the-next-cycle)

dcyoung-dev (approved):

PR #4367, opened 2024-07-11
Change tda to non tda and vice versa

tomas-stefano:
### Context

There were 4 bug cards from the bug party:

https://trello.com/c/9VaxXZ1f/2004-bug-investigate-recreating-403-access-issue-you-are-not-permitted-to-see-this-page-when-switching-a-tda-course-to-non-tda?filter=label:Squad%20B
https://trello.com/c/5HipohNv/2007-bug-investigate-broken-tda-course-that-wont-publish
https://trello.com/c/YuJcPxuE/1968-tda-publish-edge-case-reset-the-a-levels-when-user-changes-from-a-tda-course-to-a-non-tda-course
https://trello.com/c/n791FkBt/2013-bug-changing-back-to-tda-having-changed-to-non-tda-ends-up-being-unpublishable

These are always part of the same problem:

1. Changing TDA to non TDA course (in draft/rollover state).
2. Changing non TDA to TDA course (in draft/rollover state). 
3. Changing TDA to non TDA course (in published state). 
4. Changing non TDA to TDA course (in published state). 

### Changes implemented

We should allow to change from TDA <-> non TDA when draft or rollover. But not when the course is published.

### Guidance to review

1. Changing TDA to non TDA in draft/rollover state - can this be published?
5. Is the link hidden when is TDA published course?
6. Is the TDA option hidden when changing qualification on a published NON TDA course?


app/models/course.rb
```diff
@@ -811,8 +810,27 @@ def find_a_level_subject_requirement!(uuid)
...
+    when 'full_time'
+      site_status.update(vac_status: :full_time_vacancies)
+    when 'part_time'
+      site_status.update(vac_status: :part_time_vacancies)
+    when 'full_time_or_part_time'
+      site_status.update(vac_status: :both_full_time_and_part_time_vacancies)
+    else
+      raise "Unexpected study mode #{study_mode}"
+    end
+  end
```
tomas-stefano, line 832: This method was in a controller concern and was included in the Course model (only on course) so I remove the module and move the code here.

app/models/concerns/courses/edit_options/qualification_concern.rb
```diff
@@ -16,9 +16,13 @@ def qualification_options
         def qualifications_with_qts
           qts_list = Course.qualifications.keys.grep(/qts/)
 
-          return qts_list if tda_active?
+          qts_list -= %w[undergraduate_degree_with_qts] if !tda_active? || non_tda_published?
 
-          qts_list - %w[undergraduate_degree_with_qts]
+          qts_list
```
tomas-stefano, line 21: I really wish one dat to refactor the whole Edit options code because at least to me sounds wrong changing this module to remove an option in the view. And changing this module will affect different screens that uses the edit options.

I think the EditOption abstraction is bringing more problems than solutions. But maybe is just me. 

To solve this bug I have to add the condition above.

CatalinVoineag (commented):
The code looks good to me. And most of the bugs have been fixed

I still managed to change the course from TDA to something else when it's published but not as easy. Feels like something is changing when I get pundit policy tries to stop me.


https://github.com/user-attachments/assets/7a6528a0-fd9a-4cf7-8e8b-533ef4e06bee

tomas-stefano:
> The code looks good to me. And most of the bugs have been fixed
> 
> I still managed to change the course from TDA to something else when it's published but not as easy. Feels like something is changing when I get pundit policy tries to stop me.
> 
>  Peek.2024-07-12.11-15.mp4

Interesting. I did test for the published but not when is scheduled state. 🤔 
Thank you for spotting. I will add one feature test for this.

tomas-stefano:
@CatalinVoineag I pushed the fix. I optin to move everything to Pundit but to redirect away if you enter manually on the page.

CatalinVoineag (approved):
Nice 👍🏻

PR #4375, opened 2024-07-16
Add link to add A levels requirements from preview page

tomas-stefano:
### Context

When there are no A level and the user clicks to preview the course, a link to enter A levels should be there.

```
GIVEN I have created a TDA course
AND there are no A level requirements
WHEN I look at the preview and navigate to the A levels
THEN I should see a link to add A levels
```

### Changes proposed in this pull request

Screenshot 2024-07-16 at 10 00 58

### Guidance to review

1. Does it work on preview?

PR #4377, opened 2024-07-16
Fix wrong attribute name when previewing the course

tomas-stefano:
### Context

I added the wrong attribute name for #4375 

It should be A levels not GCSE.


CatalinVoineag (approved):
Nice, just need to fix a failed spec.

PR #4404, opened 2024-07-24
[1876] Ensure `program_type` is updated when `provider_type` changes

avinhurry:
### Context

- Currently when you update the `provider_type` on the provider the `program_type` on the course isn't updated.

Add an `after_save` callback to loop through and recalculate the `program_type` on courses after the `provider_type` has been changed.

### Guidance to review

- Check the `program_type` of a course in the rails console: `RecruitmentCycle.current.providers.find_by(provider_code: {provider_code}).courses.course_code(course_code: {course_code}).program_type`

- change the `provider_type` in support, e.g: `/support/2025/providers/21213/edit`

image

- Ensure the `program_type` has correctly updated on all the courses.

- Ensure the other cycle courses haven't inadvertently been affected.

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -13,6 +13,8 @@ class Provider < ApplicationRecord
   before_save :update_searchable, if: :accredited_provider?
   before_create :set_defaults
 
+  after_save :update_courses_program_type, if: :provider_type_previously_changed?
+
   has_associated_audits
   audited except: :changed_at
 
@@ -380,6 +382,13 @@ class Provider < ApplicationRecord
   end
 
   private
+
+  def update_courses_program_type
+    courses.each do |course|
+      Courses::AssignProgramTypeService.new.execute(course.funding_type, course)
+      course.save
+    end
+  end
 
   def accredited_provider_codes
     accrediting_provider_enrichments&.map(&:UcasProviderCode) || []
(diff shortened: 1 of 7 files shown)


avinhurry:
[Bug: `program_type` of a course isn't recalculated/updated](https://trello.com/c/jbRxSQxN/1876-bug-programtype-of-a-course-isnt-recalculated-updated)

app/models/provider.rb
```diff
@@ -13,6 +13,8 @@ class Provider < ApplicationRecord
   before_save :update_searchable, if: :accredited_provider?
   before_create :set_defaults
 
+  after_save :update_courses_program_type, if: :provider_type_previously_changed?
```
CatalinVoineag, line 16: Are there a lot of places where the `provider_type` can change and the `program_type` needs to be updated? Can `update_courses_program_type` method be included in a service/controller/form object when the `provider_type` changes? 

I personally don't like using `after_save` mainly because it's invoked every time you save 😄 and there can be unintended consequences.

If you think the columns need to be in sync at all times then it's fine, just thought I would ask 👍🏻

CatalinVoineag: OK, that's fine. I was just not sure if we need an after save but if the developer is changing things in the rails console than it's fine 👍🏻

CatalinVoineag (approved):

PR #4413, opened 2024-07-30
Git and provider website redirects

elceebee:
### Context

We have improved our course pages and we need a way of monitoring how our users are interacting with them

[Trello card](https://trello.com/c/y3NKamG9)

The redirect allows us to capture the initial click in big query before sending people out to the provider or get-into-teaching websites.

### Guidance to review
We can only tests that the redirects take us where we ultimately want to go. We will only know that they are being registered in BigQuery in production.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/concerns/get_into_teaching_redirect.rb
+++ b/app/controllers/concerns/get_into_teaching_redirect.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module GetIntoTeachingRedirect
+  extend ActiveSupport::Concern
+
+  def git_redirect
+    redirect_to I18n.t("get_into_teaching.#{git_url_locale_key}"), allow_other_host: true
+  end
+
+  private
+
+  def git_path_param
+    params[:git_path]
+  end
+
+  def git_url_locale_key
+    "url_#{git_path_param}"
+  end
+end
--- a/app/controllers/concerns/provider_website_redirect.rb
+++ b/app/controllers/concerns/provider_website_redirect.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module ProviderWebsiteRedirect
+  extend ActiveSupport::Concern
+
+  def provider_website
+    redirect_to provider.decorate.website, allow_other_host: true
+  end
+end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -856,6 +856,7 @@ en:
               taken: "Provider code already taken"
             provider_type:
               blank: "Provider type can't be blank"
+              format: Accredited provider cannot be a school
             email:
               blank: "^Enter email address"
             website:
@@ -883,8 +884,6 @@ en:
             accredited_provider_id:
               blank: Enter the accredited provider ID
               format: Enter a valid accredited provider ID - it must be 4 digits starting with a 1 or a 5
-            provider_type:
-              format: Accredited provider cannot be a school
         organisation:
           attributes:
             name:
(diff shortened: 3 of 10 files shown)


config/locales/en.yml
```diff
@@ -856,6 +856,7 @@ en:
               taken: "Provider code already taken"
             provider_type:
               blank: "Provider type can't be blank"
+              format: Accredited provider cannot be a school
```
elceebee, line 859: Unrelated change - just removed the duplicate key, `provider_type`

app/controllers/concerns/provider_website_redirect.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module ProviderWebsiteRedirect
+  extend ActiveSupport::Concern
+
+  def provider_website
+    redirect_to provider.decorate.website, allow_other_host: true
```
CatalinVoineag, line 7: I assume you need to have this controller that will actually redirect to where the user wants because we only capture requests to our own controllers with dfe_analytics? We don't capture requests to outside services?

CatalinVoineag (approved):
Nice work 👍🏻

app/controllers/concerns/provider_website_redirect.rb
elceebee: Yeah, that's exactly right. We do this currently with the apply button on the course pages as well, I think it is a nice little solution without having to set up beacons or fire off other events (it looks like @tomas-stefano set it up for the apply button). The analysts will be able to sort for clicks to `course/:provider_code/:course_code:/provider_url` for instance, and know how many people visited a provider's website.

CatalinVoineag: Interesting, thank you 👍🏻

tomas-stefano: Awesome idea. 🚀

tomas-stefano (approved):

config/routes/find.rb
```diff
@@ -15,6 +15,8 @@
   get '/course/:provider_code/:course_code', to: 'courses#show', as: 'course'
   get '/course/:provider_code/:course_code/placements', to: 'placements#index', as: :placements
   get '/course/:provider_code/:course_code/apply', to: 'courses#apply', as: :apply
+  get '/course/:provider_code/:course_code/git/:git_path', to: 'courses#git_redirect', as: :get_into_teaching_redirect
+  get '/course/:provider_code/:course_code/provider_website', to: 'courses#provider_website', as: :provider_website
```
dcyoung-dev:
```suggestion
  get '/course/:provider_code/:course_code/provider/website', to: 'courses#provider_website', as: :provider_website
```

Not a massive suggestion but thought this would leave the door open for adding other links related to the provider.

app/controllers/concerns/get_into_teaching_redirect.rb
```diff
@@ -0,0 +1,19 @@
...
+  extend ActiveSupport::Concern
+
+  def git_redirect
+    redirect_to I18n.t("get_into_teaching.#{git_url_locale_key}"), allow_other_host: true
+  end
+
+  private
+
+  def git_path_param
+    params[:git_path]
```
dcyoung-dev, line 13: should we validate the params we accept?

If we put something silly as the param where do we get redirected to? i.e. where would `course/:provider_code/:course_code/git/help` redirect to?

dcyoung-dev (commented):
A quick sanity check on the params accepted for the git endpoints

app/controllers/concerns/get_into_teaching_redirect.rb
elceebee: That will fail, but I think that's ok. It should fail if we don't have a locale. I thought about catching the failure and directing to the git home page, but that would hide the error.

dcyoung-dev (approved):

PR #4415, opened 2024-07-31
[2114] Add `course_type` column

avinhurry:
### Context

This is the first PR in our ITT Reform epic. 

Here we will categorise our courses by `course_type`. This will either be `postgraduate` or `undergraduate`.

- Add `course_type` column to `Course` table
- Default to `postgraduate`
- Store as an ~~`integer`~~`string`
~~- Add enum (`_prefix` required because of naming clash)~~

### Guidance to review

- ~~Does storing as an integer make sense to you? I considered storing as a string but decided to follow what was previously done in the codebase and to make some small performance gains 💪🏼~~ See PR comments.

- Does setting a default value make sense here? I did that because nearly all courses will be postgraduate. Only TDA courses will be undergraduate and we only currently have 1 of those.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


avinhurry:
[Add course_type column to course table](https://trello.com/c/YUilom7I/2114-add-coursetype-column-to-course-table)

db/migrate/20240731175623_add_course_type_to_courses.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddCourseTypeToCourses < ActiveRecord::Migration[7.1]
+  def change
+    add_column :course, :course_type, :integer, default: 0, null: false
```
dcyoung-dev, line 5: How do you feel about using a `string` here. The database will be easier to read `postgraduate` rather than `0`, `postgraduate` and `undergraduate` values are unlikely to change

app/models/course.rb
```diff
@@ -42,6 +42,11 @@ class Course < ApplicationRecord
     open: 1
   }, _prefix: :application_status
 
+  enum course_type: {
+    postgraduate: 0,
+    undergraduate: 1
+  }, _prefix: :type
+
```
dcyoung-dev, line 49: How do you feel about the methods being named `course.type_postgraduate?` and `course.type_undergraduate?`course.alternatively we could have `course.postgraduate_course_type?`, with a `course_type` suffix, or even `course.postgraduate?` without a prefix or suffix.

dcyoung-dev (commented):
In the Apply codebase we have a checklist to make sure we separate the migration from any code that uses the new column. How do you feel about splitting the migration into a separate deployable PR?

db/migrate/20240731175623_add_course_type_to_courses.rb
tomas-stefano: Should we add an index for this field? *I am not sure how will gonna be used so I didn't suggested but worth a thinking.

app/models/course.rb
tomas-stefano: I'd prefer by far `course.postgraduate?` simple, direct, clear and there won't be any clashes.

db/migrate/20240731175623_add_course_type_to_courses.rb
avinhurry: [I've converted it to a string](https://github.com/DFE-Digital/publish-teacher-training/pull/4415/commits/d4f054f17748643d10ae86d5e4dc5eaadf7fefee) @dcyoung-dev. I mentioned in my PR description that I don't feel strongly either way. Using strings makes reading the database by a non-dev slightly easier, but mapping the integers isn't that difficult. Using integers offers very slight performance gains, but it's probably not significant enough to matter in our case, however that's what was done previously and most recently for `application_status`.

@tomas-stefano, I don't think we need an index. We won't be querying by `course_type` very often, if at all. The only courses that aren't postgraduate are TDA courses, and we query those by `program_type`. Adding an index would make updates to the column slower without much benefit for reading the column, since we won't be reading it frequently.

But I'm open to hearing any counterarguments if anyone thinks there's a strong case for it.

app/models/course.rb
avinhurry: `course.postgraduate?` would definitely be my preference too but unfortunately it causes a naming clash:

`ArgumentError: You tried to define an enum named "course_type" on the model "Course", but this will generate a instance method "postgraduate?", which is already defined by another enum. (ArgumentError)`

I've gone with your suggestion @dcyoung-dev `course.postgraduate_course_type?` I think that sounds a little nicer than what I had.

avinhurry:
> In the Apply codebase we have a checklist to make sure we separate the migration from any code that uses the new column. How do you feel about splitting the migration into a separate deployable PR?

@dcyoung-dev yep that fine, happy to do that [i've removed everything apart from the migration ](https://github.com/DFE-Digital/publish-teacher-training/pull/4415/commits/8b16919d46731d09208e638218d4343fb35b956c)and will deploy that separately.

db/migrate/20240731175623_add_course_type_to_courses.rb
tomas-stefano: Ok, makes sense. For what I understand we will only need a index in Apply because we will be filtering in Manage.

app/models/course.rb
dcyoung-dev: That is a very interesting error - what other enum is generating a `postgraduate?` method 🧐

dcyoung-dev: UCAS strikes again - https://github.com/DFE-Digital/publish-teacher-training/blob/bcda42d57e6ef04931fef14a493dfef3e7179375/app/models/concerns/with_qualifications.rb#L30

I'm happy to introduce `course.postgraduate_course_type?` with the hope we can remove this other enum at some point

dcyoung-dev (approved):

PR #4421, opened 2024-08-01
[2116] Add `funding` column to `course`

avinhurry:
### Context

We will categorise our courses by a database backed `funding` column. This will either be `fee`, `salary` or `apprenticeship`.

We have `funding_type` in our implementation as it currently stands, however this is not a database backed value, it is calculated based on the `program_type`. We would like to change that.

- Set the default value to `not_set`
- Use string data type
- Prevent null columns
- Call the column `funding` to avoid conflicting with `funding_type`

### Guidance to review

~~Does it make sense to set the default value to `fee`? I did this as we don't want it to be `null` and `fee` is by far the most common `funding_type`.~~ Made the default `not_set` instead.

Following the pattern we used for `course_type` in #4415 we will use strings for readability.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


avinhurry:
[Add funding_type column to course table](https://trello.com/c/XyP7hlMm/2116-add-fundingtype-column-to-course-table)

db/migrate/20240801170318_add_funding_type_to_courses.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddFundingTypeToCourses < ActiveRecord::Migration[7.1]
+  def change
+    add_column :course, :funding, :string, default: 'fee', null: false
```
dcyoung-dev, line 5: The `default: 'fee'` is potentially dangerous. 
@elceebee had a great idea to add a "dummy" default value to make is obvious which records need updating. We could use something like "unknown"

dcyoung-dev: How do you feel about adding an index on this column as we'll regularly search based on this value 👌

avinhurry: I like @elceebee's idea to use a dummy value here. That will make it clear if the correct value hasn't been used yet. I've gone with `not_set` as the default.

An index also makes sense here 👍🏼 

https://github.com/DFE-Digital/publish-teacher-training/pull/4421/commits/380ced1a7ce8fa9e7011a1f4e33f6d1a817aaecf

dcyoung-dev (approved):

PR #4436, opened 2024-08-08
[2124] Add `funding` feature flag

avinhurry:
## Context

We are introducing a new database backed `funding` column on the `course` table, replacing the non database backed `funding_type`. This change will allow us to manage funding directly through the database.

To ensure a smooth transition, we will initially place this change behind a feature flag. This approach will allow us to:

- Control the rollout of the new funding column.
- Backfill the `funding` column with existing data through a data migration.
- Provide an easy way to revert the change if necessary.
- The feature flag will remain off in all environments until the migration and feature implementation are complete. Once the feature has been stable in production for a sufficient period, we will remove the feature flag.

## Changes proposed in this pull request

- Add the feature flag
- Set to inactive by default
- Add logic to use our existing implementation when the flag is off
- Add logic to use the db backed funding column when the flag is on

## Guidance to review

:shipit: 

## Things to check

- [x] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [x] If this code adds a column to the DB, decide whether it needs to be in analytics yml file or analytics blocklist
- [x] API release notes have been updated if necessary
- [x] Required environment variables have been updated added to the Azure KeyVault
- [x] Inform data insights team due to database changes
- [x] Make sure all information from the Trello card is in here
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally
- [x] Attach PR to Trello card

PR #4437, opened 2024-08-08
[2124] Utilise `funding` column and add tests

avinhurry:
## Context

Following #4436 we are now making use of the `funding` column. 

- Utilise `funding` column when the feature flag is on and the `funding` value is set to something other than the default (which is `not_set`). 

- If the feature flag is enabled and `course.funding` is `not_set` then use the previous implementation.

- If the feature flag is inactive, always use the previous implementation.

- Duplicate tests to continue to test the current implementation whilst testing the new implementation. The original tests can be removed along with the feature flag and the old code. Some whole files can be deleted, where this is the case it has been clearly indicated at the top of the file.

Once we merge this PR, we can switch on the feature flag. The courses will still be using the original implementation because all the records will have `funding` set to `not_set` . We can then run a backfill, after this point the new column will be in use.

image

## Guidance to review

Don't be put off by the large number of changes, they are straight forwardand I recommend going commit by commit.

- [This commit](https://github.com/DFE-Digital/publish-teacher-training/pull/4437/commits/ba164cde2bf9c1bf4efc97ca754ffb38a02455a4) disables the feature flag in the current tests to enable them to pass and test what they were previously testing.

- The large number of changes [in this commit](https://github.com/DFE-Digital/publish-teacher-training/pull/4437/commits/c62f6d3c50a4db7c8b39c98b73dfaca3a437ba49) are mainly tests which have been duplicated to test for the feature flag being enabled as well as disabled. The tests with the feature flag disabled can be removed along with the feature flag and our old implementation.

- [This commit](https://github.com/DFE-Digital/publish-teacher-training/pull/4437/commits/6050e7f22de00b52ba573b3e75301374e080b1b9) will enable us to safely use the new `funding` column as long as the value has been set.

## Things to check

- [x] Rebased main
- [ ] Cleaned commit history


avinhurry:
[Start utilising the new funding column](https://trello.com/c/hrm3YeLg/2124-start-utilising-the-new-funding-column)

CatalinVoineag (approved):
This looks good to me

PR #4443, opened 2024-08-13
[1884] Add 2026 cycle dates to codebase

elceebee:
## Context

As we approach the new cycle, we will need to have the next cycle defined, so adding it now. There are still some dates to be confirmed, but the opening dates and reject by default dates are agreed.

This PR keeps the dates in sync with the dates defined in Apply. See [this PR](https://github.com/DFE-Digital/apply-for-teacher-training/pull/9678)

- Added dates with an note about what needs to be confirmed later.
- Updated the 2025 dates that have been confirmed.

## Things to check

- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/services/find/cycle_timetable.rb
+++ b/app/services/find/cycle_timetable.rb
@@ -39,17 +39,15 @@ module Find
         find_opens: Time.zone.local(2024, 10, 1, 9), # CONFIRMED
         apply_opens: Time.zone.local(2024, 10, 8, 9), # CONFIRMED
         first_deadline_banner: Time.zone.local(2025, 7, 12, 9), # TBC
-        apply_deadline: Time.zone.local(2025, 9, 17, 18), # CONFIRMED
-        find_closes: Time.zone.local(2025, 9, 25, 23, 59, 59) # TBC
+        apply_deadline: Time.zone.local(2025, 9, 16, 18), # CONFIRMED
+        find_closes: Time.zone.local(2025, 9, 30, 23, 59, 59) # CONFIRMED
       },
       2026 => {
-        # NOTE: the dates from below here are not the finalised but are required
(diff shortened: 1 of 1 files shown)


app/services/find/cycle_timetable.rb
```diff
@@ -39,17 +39,15 @@ class CycleTimetable
         find_opens: Time.zone.local(2024, 10, 1, 9), # CONFIRMED
         apply_opens: Time.zone.local(2024, 10, 8, 9), # CONFIRMED
         first_deadline_banner: Time.zone.local(2025, 7, 12, 9), # TBC
-        apply_deadline: Time.zone.local(2025, 9, 17, 18), # CONFIRMED
-        find_closes: Time.zone.local(2025, 9, 25, 23, 59, 59) # TBC
+        apply_deadline: Time.zone.local(2025, 9, 16, 18), # CONFIRMED
+        find_closes: Time.zone.local(2025, 9, 30, 23, 59, 59) # TBC
```
avinhurry, line 43: Is the `find_closes` date for the 2025 cycle still not confirmed? 

Are the dates below for the 2026 cycle confirmed?

avinhurry: I'm being quite fussy because I know we don't usually look at these dates for a year after Find opens.

elceebee: We are publishing the 2025-26 dates on the gov uk pages and they have been ok-ed by Eliot. I suppose there is a possibility that the find_closes date will change based on the 20-27 find_open date, but that is a problem for 2027 us. https://ukgovernmentdfe.slack.com/archives/C077NRS8K8D/p1723540691902549

avinhurry (approved):

PR #4467, opened 2024-08-22
[2234] Ensure `funding` column updates correctly

avinhurry:
## Context

We have recently added a new `funding` column on the `Course` table which will replace our current implementation of `funding_type` which is not database backed. 

The correct value from the new column is displayed if the funding value has been updated to a real funding value, i.e `fee`, `salary` or `apprenticeship`. 

In this PR we need to ensure that the `funding` value can be updated when creating a new course and updating existing courses. **A backfill is required for this to work properly in production and other environment**s. I've completed the backfill on the review app (to enable more effective reviews).

Note that there are areas where `funding_type` is used for branching.

The branching for the old implementation of `funding_type` will be removed along with the feature flag, so a lot of this code is only temporary. 

- Ensure the funding column is updated when a new course is created
- Ensure the funding column is updated when an existing course is updated
- Ensure the branching relating to the funding all work correctly
- Ensure the appropriate `program_type` is assigned when funding is changed
- Ensure the previous implementation continues to work when the feature flag is disabled

## Guidance to review

The [review app ](https://publish-review-4467.test.teacherservices.cloud/publish/organisations/4A3/2024/courses)is in the correct state for testing.

- Create a course with various funding types
- Update the course with various funding types
- Ensure the column updates correctly
- Ensure the correct visa sponsorship is shown for each funding type. I.e student visa for fee paying courses and skilled worker for salaried courses 
- Ensure changing a TDA course to a non TDA course works as expected
- Ensure changing a non TDA course to a TDA course works as expected
- Ensure the old implementation continues to function as it did before when the `db_backed_funding_type` feature flag is disabled

## Things to check

- [x] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/publish/courses/funding_type_controller.rb
+++ b/app/controllers/publish/courses/funding_type_controller.rb
@@ -38,7 +38,7 @@ module Publish
       def funding_type_params
         return {} if params[:publish_course_funding_form].blank?
 
-        params.require(:publish_course_funding_form).permit(:funding_type, :previous_tda_course)
+        params.require(:publish_course_funding_form).permit(:funding_type, :funding, :previous_tda_course)
       end
 
       def handle_valid_form
@@ -59,7 +59,7 @@ module Publish
       end
 
       def process_previous_tda_course
-        @course_funding_form.save! if @course_funding_form.funding_type_updated?
+        @course_funding_form.save! if @course_funding_form.funding_updated?
         redirect_to full_part_time_publish_provider_recruitment_cycle_course_path(
           provider_code: course.provider_code,
           recruitment_cycle_year: course.recruitment_cycle_year,
@@ -69,7 +69,7 @@ module Publish
       end
 
       def next_path
-        if @course_funding_form.funding_type_updated?
+        if @course_funding_form.funding_updated?
           @course_funding_form.stash
           visa_page_path
         else
--- a/app/controllers/publish/courses/funding_type_controller.rb
+++ b/app/controllers/publish/courses/funding_type_controller.rb
@@ -38,6 +38,8 @@ module Publish
       def funding_type_params
         return {} if params[:publish_course_funding_form].blank?
 
+        # Remove funding_type from permitted params when the `db_backed_funding_type` feature flag is removed
+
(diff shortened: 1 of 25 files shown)


avinhurry:
[Ensure funding column updates correctly](https://trello.com/c/6WwH5cCF/2234-ensure-funding-column-updates-correctly)

app/controllers/publish/courses/funding_type_controller.rb
```diff
@@ -38,7 +38,7 @@ def update
       def funding_type_params
         return {} if params[:publish_course_funding_form].blank?
 
-        params.require(:publish_course_funding_form).permit(:funding_type, :previous_tda_course)
+        params.require(:publish_course_funding_form).permit(:funding_type, :funding, :previous_tda_course)
```
elceebee, line 41: How do you feel about putting a comment above this to the effect that when we get rid of `FeatureService.enabled?(:db_backed_funding_type)` we need to remove `:funding_type` from this line? It feels like something we'll forget otherwise. But if we are looking everywhere for `FeatureService.enabled?(:db_backed_funding_type)` and we find it in this comment, we'll remember. 

I'm not sure how people feel about comments / TODOs in the code. So if it's not the done thing, happy for this to be ignored. 

Basically, I can't find anything wrong with this and I'm just making a comment so you can tell I've done a thorough review.

elceebee (approved):
Great stuff. No notes. Testing in review app seems to work just fine.

app/controllers/publish/courses/funding_type_controller.rb
avinhurry: Haha, thats great. Thank you 🙏🏼 

I'm usually against comments in code, I think the code should be self explanatory but in this case we could easily miss it since the feature flag isn't directly referenced so happy to add the comment.

I'm planning to keep the feature flag in long enough to ensure that everything is working as expected but short enough that it's still fresh in my mind. So the comment will be removed soon anyways.

https://github.com/DFE-Digital/publish-teacher-training/pull/4467/commits/aeaaaac445b86c74e6d5775a9cac5eaadfff2d9a

PR #4468, opened 2024-08-22
Make sure undergraduate courses and non duplicates appear on the API

tomas-stefano:
## Context

We need to make sure undergraduate AND postgraduate courses are returned on the API.

This was necessary because the CourseSearchService is used in Find web service although there we will have to filter by course type always on the current designs.

Also added the order by to remove the duplications to be happening on the API

### DIFF
--- a/app/services/api_course_search_service.rb
+++ b/app/services/api_course_search_service.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class APICourseSearchService < CourseSearchService
+  def default_course_order(outer_scope)
+    outer_scope.order('course.id': :asc)
+  end
+
+  def filter_by_course_type(scope)
+    scope # we want to return all course types in the API
+  end
+end
--- a/app/services/api_course_search_service.rb
+++ b/app/services/api_course_search_service.rb
@@ -2,7 +2,7 @@
 
 class APICourseSearchService < CourseSearchService
   def default_course_order(outer_scope)
-    outer_scope.order('course.id': :asc)
+    outer_scope.order('course.created_at': :asc)
   end
 
   def filter_by_course_type(scope)
--- a/app/services/api_course_search_service.rb
+++ b/app/services/api_course_search_service.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class APICourseSearchService < CourseSearchService
-  def default_course_order(outer_scope)
+  def course_order(outer_scope)
     outer_scope.order('course.created_at': :asc)
   end
 
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -85,7 +85,9 @@ class CourseSearchService
 
   private
 
-  # Method signature
+  # This method provides a default ordering for courses.
+  # It can be overridden in subclasses to apply different ordering logic
+  # for various contexts, such as the API.
   def default_course_order(outer_scope)
     outer_scope
   end
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -75,7 +75,7 @@ class CourseSearchService
           outer_scope.order(:distance)
         end
     else
-      outer_scope = default_course_order(outer_scope)
+      outer_scope = course_order(outer_scope)
     end
 
     outer_scope
@@ -85,10 +85,9 @@ class CourseSearchService
 
   private
 
-  # This method provides a default ordering for courses.
-  # It can be overridden in subclasses to apply different ordering logic
+  # This method can be overridden in subclasses to apply different ordering logic
   # for various contexts, such as the API.
-  def default_course_order(outer_scope)
+  def course_order(outer_scope)
     outer_scope
   end
 
--- a/app/services/course_search_service.rb
+++ b/app/services/course_search_service.rb
@@ -21,7 +21,7 @@ class CourseSearchService
 
   def call
     scope = course_scope
-    scope = scope.with_course_type(course_type)
+    scope = filter_by_course_type(scope)
     scope = scope.with_salary if funding_filter_salary?
     scope = scope.with_qualifications(qualifications) if qualifications.any?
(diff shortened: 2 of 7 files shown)


tomas-stefano:
[Prevent Publish API from displaying duplicate courses](https://trello.com/c/bQIwfOeB/2105-prevent-publish-api-from-displaying-duplicate-courses)

tomas-stefano:
[TDA: Make sure undergraduate courses appears on the Publish API](https://trello.com/c/pglUOl4M/2204-tda-make-sure-undergraduate-courses-appears-on-the-publish-api)

app/services/api_course_search_service.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class APICourseSearchService < CourseSearchService
+  def default_course_order(outer_scope)
+    outer_scope.order('course.id': :asc)
```
tomas-stefano, line 5: I added course id ASC but let me know if you prefer any other way on the default order on the API. Happy to change.

tomas-stefano:
A context. I tried to refactor a lot of code and starting to separate many behaviour in a way that respects a clean code approach on #4466 BUT this will take more time and require to increase a huge amount of coverage for tests that should be there and it is not there, also will require more manual testing to make extra sure we don't break anything, so changing on these spec is the minimal of the minimal I could find to change the code.

app/services/api_course_search_service.rb
elceebee: One day, it would be nice to stop using consecutive numerical ids, so perhaps it would be better to order by `:created_at` or `:name`

inulty-dfe: Would `updated_at` be the best ordering?

tomas-stefano: Good suggestion. In related to your suggestion I prefer the created_at. I will add the changes.

app/services/course_search_service.rb
```diff
@@ -83,6 +85,15 @@ def call
 
   private
 
+  # Method signature
```
inulty-dfe, line 88: What does this comment mean?

spec/controllers/api/public/v1/courses_controller_spec.rb
```diff
@@ -66,6 +66,60 @@
...
+          ids = []
+
+          (1..10).each do |page|
+            get :index, params: {
+              recruitment_cycle_year: recruitment_cycle.year,
+              page:,
+              per_page: 3
+            }
+
+            ids << response.parsed_body['data'].pluck('id')
```
inulty-dfe, line 113: `+=` merges the array

app/services/api_course_search_service.rb
tomas-stefano: Done!

inulty-dfe (commented):
Would `updated_at` be the best ordering?

app/services/api_course_search_service.rb
tomas-stefano: > Would updated_at be the best ordering?
`updated_at DESC` I'd say yes. @dcyoung-dev wdyt?

app/services/course_search_service.rb
tomas-stefano: The term "method signature" refers to the definition of a method, specifically its name and parameters to be easily to be replaceable in subclasses. Here, the comment highlights the default_course_order method's signature, which includes the method name and its single parameter, outer_scope. This method currently returns outer_scope unchanged but can be overridden in subclasses to customize the course ordering logic. The signature sets up this method as a potential extension point for future customization. Because the Web ordering is different from the API ordering.

If you think is not necessarily I am happy to remove the comment.

spec/controllers/api/public/v1/courses_controller_spec.rb
tomas-stefano: True! Done

app/services/api_course_search_service.rb
dcyoung-dev: I'm happy with `id` or `created_at` `asc` both of these would preserve the pages in which they appear.
i.e. if more Courses are added they will be on the last page, all other Courses stay on the page they were on. 
If we order by a value that updates then the results on each page get mixed up again.

tomas-stefano: Ok, the `created at asc` is implemented so now all is good. Thank you all 🙏

app/services/course_search_service.rb
inulty-dfe: I think I would call that an interface. I would add more context to the comment, "Method signature" doesn't convey any of the intent you have explained her.

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -43,6 +43,60 @@
...
+            provider_code: provider.provider_code,
+            page:,
+            per_page: 3
+          }
+
+          ids += response.parsed_body['data'].pluck('id')
+        end
+
+        expect(ids.size).to eq(ids.uniq.size)
+        expect(ids).to eq(ids.sort)
```
elceebee, line 96: Isn't this just incidental because `created_at` and `id` are going to be ordered in the same way because we use consecutive numerical ids? Do you want to check that the records are sorted by created_at?

elceebee (approved):
Nice work. Just a minor thing about a spec, but not a blocker. 🚀

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
tomas-stefano: True it was a reminiscence of the order by ids. I chosen to deleted. Done

app/services/course_search_service.rb
tomas-stefano: Makes sense. I change to a more meaningful comment.

inulty-dfe (approved):

PR #4479, opened 2024-08-27
Enable TDA in staging and QA

tomas-stefano:
## Context

Enabling TDA in staging and QA.

## Why?

* To test integrations with Apply
* To facilitate demos
* For everyone in the team has access to how TDA is working though our apps

### DIFF
--- a/config/settings/qa_aks.yml
+++ b/config/settings/qa_aks.yml
@@ -31,6 +31,7 @@ basic_auth:
 
 features:
   send_request_data_to_bigquery: true
+  teacher_degree_apprenticeship: true
 
 find_valid_referers:
   - https://qa.find-teacher-training-courses.service.gov.uk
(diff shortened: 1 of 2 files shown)


tomas-stefano:
One of the reasons is to have TDA courses in QA so I can test https://github.com/DFE-Digital/apply-for-teacher-training/pull/9750

avinhurry (changes requested):
This PR is not necessary, `teacher_degree_apprenticeship` is already active in all environments:

https://github.com/DFE-Digital/publish-teacher-training/blob/main/config/settings.yml#L95-L96


We should remove it from: 

- https://github.com/DFE-Digital/publish-teacher-training/blob/07b33dac4d51b5a222ff5de49c0179a679f4f35f/config/settings/development.yml#L13

To avoid confusion.

tomas-stefano:
🤔 True!! I've made the confusion (I know that I coded! 🙈 )

Publish was active but Find was not active I thought was the feature flag, but I forgot we hardcode the recruitment cycle in the yml file.

I will close the PR

PR #4482, opened 2024-09-02
Add rake task to set selectable_school to true

inulty-dfe:
## Context

  We've identified 11 Provider who wish for their candidates to choose a
  school placement. This rake task updates the providers accordingly.

Add Rake task to update specific providers `selectable_school` to true.

## Things to check

- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/lib/tasks/update_selectable_schools.rake
+++ b/lib/tasks/update_selectable_schools.rake
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module DataMigrations
+  class SetSelectableSchoolsOnProviders
+    def change
+      provider_codes = %w[E65 2CG 2CH 2CJ 2CK 2C1 1EX 1QU CS01 2X4 1ZO]
+      Provider.where(provider_code: provider_codes).update_all(selectable_school: true) # rubocop:disable Rails/SkipsModelValidations
+    end
+  end
+end
+
+desc 'Set Selectable Schools on 11 providers'
+task set_selectable_schools: :environment do |_, _args|
+  DataMigrations::SetSelectableSchoolsOnProviders.new.change
+end
--- a/lib/tasks/update_selectable_schools.rake
+++ b/lib/tasks/update_selectable_schools.rake
@@ -4,7 +4,7 @@ module DataMigrations
   class SetSelectableSchoolsOnProviders
     def change
       provider_codes = %w[E65 2CG 2CH 2CJ 2CK 2C1 1EX 1QU CS01 2X4 1ZO]
-      Provider.where(provider_code: provider_codes).update_all(selectable_school: true) # rubocop:disable Rails/SkipsModelValidations
+      Provider.where(provider_code: provider_codes).update_all(selectable_school: true)
     end
   end
 end
--- a/spec/lib/tasks/update_selectable_school_rake_spec.rb
+++ b/spec/lib/tasks/update_selectable_school_rake_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'rake'
+
+describe 'set_selectable_schools' do
+  Rails.application.load_tasks if Rake::Task.tasks.empty?
+  subject(:set_selectable_schools_task) do
+    Rake::Task['set_selectable_schools'].invoke
+  end
+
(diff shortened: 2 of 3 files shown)


inulty-dfe:
[Add DataMigration to set `selectable school` on providers](https://trello.com/c/pDpPujWe/2203-add-datamigration-to-set-selectable-school-on-providers)

lib/tasks/update_selectable_schools.rake
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module DataMigrations
+  class SetSelectableSchoolsOnProviders
+    def change
+      provider_codes = %w[E65 2CG 2CH 2CJ 2CK 2C1 1EX 1QU CS01 2X4 1ZO]
+      Provider.where(provider_code: provider_codes).update_all(selectable_school: true) # rubocop:disable Rails/SkipsModelValidations
```
avinhurry, line 7: Do we need to update it for all 6 recruitment cycles? Could we just do it for the current and the next recruitment cycle? 

Also this updates 11 providers. in the PR description it says 7, did we find more that need to be updated?

avinhurry: Also, If we need to disable the cop could we do it in:

https://github.com/DFE-Digital/publish-teacher-training/blob/main/config/rubocop/rails.yml

So we can see which cops have been excluded. 

Would using a `find_each` with `update_columns` allow you to do it without disabling any cops? Not too fussed, just wondering what your thoughts were.

lib/tasks/update_selectable_schools.rake
```diff
@@ -0,0 +1,15 @@
...
+    end
+  end
+end
+
+desc 'Upsert GIAS data into GiasSchool'
```
avinhurry, line 12: Did you forget to change this?

spec/lib/tasks/update_selectable_school_rake_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'rake'
+
+describe 'set_selectable_schools' do
+  Rails.application.load_tasks if Rake::Task.tasks.empty?
+  subject do
+    Rake::Task['set_selectable_schools'].invoke
+  end
```
avinhurry, line 10:
```suggestion
  subject(:set_selectable_schools_task) do
    Rake::Task['set_selectable_schools'].invoke
  end
```

spec/lib/tasks/update_selectable_school_rake_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  let(:target_provider) { create(:provider, provider_code: 'E65', selectable_school: false) }
+  let(:ignore_provider) { create(:provider, provider_code: 'Z00', selectable_school: false) }
+
+  it 'updates the providers selectable_school to true' do
+    expect(target_provider.selectable_school).to be(false)
+    expect(ignore_provider.selectable_school).to be(false)
+
+    subject
```
avinhurry, line 19:
```suggestion
set_selectable_schools_task
```

lib/tasks/update_selectable_schools.rake
inulty-dfe: Yeah, I had 7 in my head since earlier in the process, it's actually 11.

As for the rubocop thing, I'm not too bothered about it at all. I'm happy to disable it throughout the application.

avinhurry: You don't have to disable it throughout the application, I think you can just disable it for the file like this: 

https://github.com/DFE-Digital/publish-teacher-training/blob/47b710e467c8014fbef01ac2a870cbfbf04a8cf1/config/rubocop/rails.yml#L13-L15

It's just nice to see everything that's disabled in one place.

inulty-dfe: I don't think it's a linting rule worth having. It's very well known what `update_all` does and anyone using it where they shouldn't ought to be picked up in a code review. I don't think we should have to selectively disable this rule.

inulty-dfe: Though I agree having all the bypasses in one place is a good idea

avinhurry (approved):

PR #4495, opened 2024-09-05
[2267] Implement `training_route`

avinhurry:
## Context

Adding a way to calculate `training_route` based on a courses `funding` and `course type`.

This is in line with Register, for the routes which apply to Publish.

https://becoming-a-teacher.design-history.education.gov.uk/becoming-a-teacher/routes-into-teaching/

image

Implement a hash map to easily return `training_route` based on the different combinations of course types and funding types.

## Guidance to review

- Do the values returned make sense to you? They are in line with Register to help make our services more consistent.

- Any thoughts on having this as  a database backed column? I decided not to do this, as `training_route` should always be calculated the same way, it didn't make sense to me to combine two strings and put it in a new column.

## Things to check

- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -14,7 +14,7 @@ class Course < ApplicationRecord
     %w[postgraduate fee] => 'fee_funded_initial_teacher_training',
     %w[postgraduate salary] => 'school_direct_salaried',
     %w[postgraduate apprenticeship] => 'postgraduate_teacher_apprenticeship',
-    %w[undergraduate salary] => 'teacher_degree_apprenticeship'
+    %w[undergraduate apprenticeship] => 'teacher_degree_apprenticeship'
   }.freeze
 
   after_initialize :set_defaults
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -10,6 +10,13 @@ class Course < ApplicationRecord
 
   A_LEVEL_ATTRIBUTES = %i[a_level_subject_requirements accept_pending_a_level accept_a_level_equivalency additional_a_level_equivalencies].freeze
 
+  TRAINING_ROUTE_MAP = {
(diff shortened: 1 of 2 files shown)


avinhurry:
[Implement training route calculation](https://trello.com/c/oIszp5R6/2267-implement-training-route-calculation)

app/models/course.rb
```diff
@@ -10,6 +10,13 @@ class Course < ApplicationRecord
 
   A_LEVEL_ATTRIBUTES = %i[a_level_subject_requirements accept_pending_a_level accept_a_level_equivalency additional_a_level_equivalencies].freeze
 
+  TRAINING_ROUTE_MAP = {
+    %w[postgraduate fee] => 'fee_funded_initial_teacher_training',
+    %w[postgraduate salary] => 'school_direct_salaried',
+    %w[postgraduate apprenticeship] => 'postgraduate_teacher_apprenticeship',
+    %w[undergraduate salary] => 'teacher_degree_apprenticeship'
```
dcyoung-dev, line 17:
```suggestion
    %w[undergraduate apprenticeship] => 'teacher_degree_apprenticeship'
```
I'd have assumed this should be an apprenticeship rather than salary 🤔

avinhurry: Both of these options would uniquely map to TDA but I think you're right, Apprenticeship does feel more correct.

I was previously going directly off of this mapping: https://becoming-a-teacher.design-history.education.gov.uk/becoming-a-teacher/routes-into-teaching/

but apprenticeships are salaried anyways. I've updated 👍🏼 

https://github.com/DFE-Digital/publish-teacher-training/pull/4495/commits/0b0659f1d2aee9c5c3198f2404f0696c334da741

PR #4514, opened 2024-09-11
Allow API to 404 when no recruitment cycle is found

dcyoung-dev:
## Context

Handles [Method Not Found Error](https://dfe-teacher-services.sentry.io/issues/4233014348/?alert_rule_id=11137790&alert_type=issue&notification_uuid=3d612f94-fbda-4c93-a06a-77e687acbfcf&project=1377944)

- Added a `RecruitmentCycle#current_recruitment_cycle!` method which will raise a `RecordNotFound` error if the cycle does not exist. The API will return a 404 at this point rather than continue to call `#providers` on a `nil` value

## Things to check

- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/application_controller.rb
+++ b/app/controllers/api/public/v1/application_controller.rb
@@ -10,7 +10,7 @@ module API
 
         def recruitment_cycle
           year = params[:recruitment_cycle_year]
-          @recruitment_cycle ||= RecruitmentCycle.find_by(year:) || RecruitmentCycle.current_recruitment_cycle
+          @recruitment_cycle ||= RecruitmentCycle.find_by(year:) || RecruitmentCycle.current_recruitment_cycle!
         end
       end
     end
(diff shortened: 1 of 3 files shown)


app/controllers/api/public/v1/application_controller.rb
```diff
@@ -10,7 +10,7 @@ class ApplicationController < PublicAPIController
 
         def recruitment_cycle
           year = params[:recruitment_cycle_year]
-          @recruitment_cycle ||= RecruitmentCycle.find_by(year:) || RecruitmentCycle.current_recruitment_cycle
+          @recruitment_cycle ||= RecruitmentCycle.find_by(year:) || RecruitmentCycle.current_recruitment_cycle!
```
avinhurry, line 13: A suggestion here (at your discretion) is to put this behind an environment check so that this would only return a 404 and not raise a sentry error in non prod environments because the recruitment cycle _should_ always be found in prod and we'd want to know if it wasn't there.

avinhurry (approved):
Approved with suggestion 👍🏼

PR #4517, opened 2024-09-12
[2313] Remove `db_backed_funding_type` feature flag and associated old logic (Part 1)

avinhurry:
## Context

Now that we have been using the funding column for a while, we are removing the `db_backed_funding` feature flag and the associated logic, which is used when the flag is switched off, as we won’t be going back.

- Remove the feature flag
- Remove the old logic (for when the feature flag is inactive)

## Guidance to review

These are all the test failures from making the small number of changes in the first commit.

Ensure everything still works as before. The end user should not notice any differences at all. Pay particular attention to changing the funding type, the TDA flow and creating courses.

## Things to check

- [x] Rebased main
- [x] Cleaned commit history


avinhurry:
[Remove db_backed_funding feature flag](https://trello.com/c/HLWpOtxO/2312-remove-dbbackedfunding-feature-flag)

app/models/course.rb
```diff
@@ -571,12 +571,7 @@ def student_visa?
 
   def funding_type
     ActiveSupport::Deprecation.new.warn('`funding_type` is deprecated and will be removed. Please use `funding` instead.')
-
-    if FeatureService.enabled?(:db_backed_funding_type)
-      not_set? ? program.funding_type : funding
```
dcyoung-dev, line 576: We definetly don't have any courses with the `not_set` value?

dcyoung-dev (approved):

app/models/course.rb
avinhurry: We do not:
```
Course.where(funding: 'not_set').count  # 0
Course.where(funding: ['fee', 'salary', 'apprenticeship']).count # 114869
Course.count # 114869
```

PR #4542, opened 2024-09-26
Setup test undergraduate courses data for Find

tomas-stefano:
## Context

For some time we won't have any TDA course so for now we need to generate many courses to test Find or Publish.

So this rake task create TDA courses and make sure the data is more correct (no fees, no secondary courses for primary schools, etc)

## Guidance to review

1. Run the rake task `rake undergraduate:create`
2. Does it work locally when searching on Find?

### DIFF
--- a/lib/tasks/undergraduate_courses.rake
+++ b/lib/tasks/undergraduate_courses.rake
@@ -20,17 +20,36 @@ namespace :undergraduate do
                  'Mathematics'
                end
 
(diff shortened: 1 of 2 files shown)


tomas-stefano:
[TDA: Setup the right courses data for Find & Apply](https://trello.com/c/plZ0ECRh/2354-tda-setup-the-right-courses-data-for-find-apply)

avinhurry (approved):

PR #4549, opened 2024-09-26
Show specific no results message when searching undergraduate courses

tomas-stefano:
## Context

When Teacher Degree Apprenticeships launches it is likely there will be no or very little courses available.

This may cause a dead end on the search results page, which isn’t a good user journey.

Adding a more detailed explanation could help users.

This is a short term solution and there are talks for a long term solution if this problem persists.

## Changes

screencapture-find-localhost-results-2024-09-26-17_50_59

## Guidance to review

1. Enter on review
2. Answer that you don't have a degree
3. See no results page and see content

### DIFF
--- a/config/locales/en/components/find/no_results_component.yml
+++ b/config/locales/en/components/find/no_results_component.yml
@@ -1,0 +1,7 @@
+en:
+  find:
+    results:
+      no_results_component:
+        undergraduate:
+          message_html: There are not many teacher degree apprenticeship (TDA) courses on the service at the moment. You can try again soon when there may be more courses, or get in touch with us at %{contact}.
+          find_out_more_about_tda: Find out more about teacher degree apprenticeship (TDA) courses.
--- a/config/settings/review_aks.yml
+++ b/config/settings/review_aks.yml
@@ -15,3 +15,5 @@ search_ui:
 authentication:
   secret: secret
   mode: persona
+
+current_recruitment_cycle_year: 2025
--- a/config/settings/review_aks.yml
+++ b/config/settings/review_aks.yml
@@ -15,5 +15,3 @@ search_ui:
 authentication:
   secret: secret
   mode: persona
-
-current_recruitment_cycle_year: 2025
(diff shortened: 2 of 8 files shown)


tomas-stefano:
[Find: Add initial content for "no courses found" for undergraduate courses](https://trello.com/c/lKbYQZHI/2210-find-add-initial-content-for-no-courses-found-for-undergraduate-courses)

config/settings/review_aks.yml
```diff
@@ -15,3 +15,5 @@ search_ui:
 authentication:
   secret: secret
   mode: persona
+
+current_recruitment_cycle_year: 2025
```
tomas-stefano, line 19: This is only to test the review apps. I will remove before merging.

config/locales/en/components/find/no_results_component.yml
```diff
@@ -0,0 +1,7 @@
+en:
+  find:
+    results:
+      no_results_component:
+        undergraduate:
+          message_html: There are not many teacher degree apprenticeship (TDA) courses on the service at the moment. You can try again soon when there may be more courses, or get in touch with us at %{contact}.
```
inulty-dfe, line 6: Why is "html" part of the key?

inulty-dfe (approved):
LGTM, just maybe remove the "html" from "message_html"

config/locales/en/components/find/no_results_component.yml
tomas-stefano: Because Rails will call html_safe as I am passing the link in the view

dcyoung-dev, line 6: It'd be good to get eyes on this from Content designers 👌

dcyoung-dev (approved):
Nice work 👌

config/locales/en/components/find/no_results_component.yml
tomas-stefano: All good and confirmed. I've asked earlier ✌️

PR #4551, opened 2024-09-26
[2397] Migration to remove default value (`not_set`) from `funding` column

avinhurry:
## Context

we added a default of `not_set` to the funding column to support migrating from the the calculated `funding_type` method. Anything that had `not_set` would use the old funding_type.

Now that we have we've backfilled everything so that there are no `not_set` funding values  we will remove that default.

- Migration to remove the default `not_set` value on funding
- Add a default funding value to course factory (fee) as this is what over 85% of courses are.
- Add funding to seeds as a value  is now explicitly required.
- Fix all the tests which fail due to this change

## Guidance to review

Are there any risks with this?

## Things to check

- [x] Rebased main
- [x] Cleaned commit history


avinhurry:
[Remove default (not_set) value from funding](https://trello.com/c/4monu2NW/2397-remove-default-notset-value-from-funding)

dcyoung-dev (approved):

PR #4554, opened 2024-09-27
2382 course fee is wrong when changing non tda to tda course

tomas-stefano:
## Context

Found one TDA course on QA/Production with fees on it.

TDA courses shouldn't have fees. This bug happens when the user switch from postgraduate to undergraduate courses.

On Find the fee shouldn’t be displayed as the course fee is only for fee funding courses.

The potential reason of this bug is when changing Non TDA to TDA course might have set the fee and not being erased when the course is changed to TDA

Erase fees if course is changed from Non TDA to TDA courses. If the user change back they will have to enter fees again.

## Guidance to review

1. Create a postgraduate courses
2. Enter fees
3. Switch to a TDA course (basic details -> qualifications)
4. Check if the fees are visible (or in the database)

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -505,6 +505,11 @@ class CourseDecorator < ApplicationDecorator
     )
   end
 
+  def no_fee?
+    object.teacher_degree_apprenticeship? ||
+      (fee_international.blank? && fee_uk_eu.blank?)
+  end
+
   private
 
   def not_on_find
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -506,8 +506,7 @@ class CourseDecorator < ApplicationDecorator
   end
 
   def no_fee?
-    object.teacher_degree_apprenticeship? ||
-      (fee_international.blank? && fee_uk_eu.blank?)
+    !object.fee?
   end
 
   private
--- a/spec/components/find/results/search_result_component_spec.rb
+++ b/spec/components/find/results/search_result_component_spec.rb
@@ -116,7 +116,7 @@ module Find
       end
     end
 
-    context 'when teacher degree apprenticeship courses has incorrectly fees' do
+    context 'when teacher degree apprenticeship course has incorrect fees' do
       it 'does not render fees' do
         course = create(:course, :published_teacher_degree_apprenticeship, enrichments: [create(:course_enrichment, fee_uk_eu: 9250)]).decorate
 
--- a/spec/components/find/results/search_result_component_spec.rb
+++ b/spec/components/find/results/search_result_component_spec.rb
@@ -118,7 +118,7 @@ module Find
 
     context 'when teacher degree apprenticeship course has incorrect fees' do
       it 'does not render fees' do
-        course = create(:course, :published_teacher_degree_apprenticeship, enrichments: [create(:course_enrichment, fee_uk_eu: 9250)]).decorate
+        course = create(:course, :apprenticeship, :published_teacher_degree_apprenticeship, enrichments: [create(:course_enrichment, fee_uk_eu: 9250)]).decorate
 
         result = render_inline(described_class.new(course:, results_view:))
         expect(result.text).not_to include('£9,250')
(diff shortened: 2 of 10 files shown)


tomas-stefano:
[Course fee is wrong when changing Non TDA to TDA course](https://trello.com/c/LMRRu1wb/2382-course-fee-is-wrong-when-changing-non-tda-to-tda-course)

app/decorators/course_decorator.rb
```diff
@@ -505,6 +505,11 @@ def course_fee_content
     )
   end
 
+  def no_fee?
+    object.teacher_degree_apprenticeship? ||
+      (fee_international.blank? && fee_uk_eu.blank?)
```
inulty-dfe, line 510: Courses have a funding column now, should that be used to determine if the course is fee based?

tomas-stefano: @inulty-dfe do you mean the no_fee only checks for `funding == 'fee'` or something similar? cc @avinhurry

tomas-stefano: @avinhurry what do you think about?

avinhurry: What is the intention with this `no_fee` method? @tomas-stefano

tomas-stefano: > I don't think we clear out the fee information when changing from a non TDA course to a TDA course? We should probably add a card to do that.

This PR does that (on tda attributes service class change)

tomas-stefano: > Should TDA courses never have the fee information inputted?

TDA specifically yes. Other future undergraduate courses I am not sure.

avinhurry: Hmm, i'd like to go with Iains suggestion of using `funding == fee` but i'm worried about edge cases. 

Why do we need this check? `object.teacher_degree_apprenticeship?`

Am I right in thinking that this check is only to catch the edge cases that are currently in? If so it might be better to fix those while their aren't that many. Maybe we should have a quick huddle.

spec/services/publish/courses/assign_tda_attributes_service_spec.rb
```diff
@@ -3,19 +3,54 @@
...
-      expect(described_class.new(course).call).to be_truthy
-      expect(course.study_mode).to eq('full_time')
-      expect(course.funding_type).to eq('apprenticeship')
-      expect(course.can_sponsor_skilled_worker_visa).to be(false)
-      expect(course.can_sponsor_student_visa).to be(false)
-      expect(course.additional_degree_subject_requirements).to be(false)
-      expect(course.degree_subject_requirements).to be_nil
-      expect(course.degree_grade).to eq('not_required')
-      expect(course.degree_type).to eq('undergraduate')
+    context 'when course does not have enrichments' do
```
avinhurry, line 19: Since we've specified postgrad below, should we specify the `degree_type` here?

tomas-stefano: Done

app/decorators/course_decorator.rb
tomas-stefano: @avinhurry @inulty-dfe that's done ✌️

spec/components/find/results/search_result_component_spec.rb
```diff
@@ -159,7 +169,7 @@ module Find
 
     context 'when there are no fees' do
       it 'does not render the row' do
-        course = create(:course, enrichments: [create(:course_enrichment, fee_uk_eu: nil, fee_international: nil)]).decorate
+        course = create(:course, :salary, enrichments: [create(:course_enrichment, fee_uk_eu: nil, fee_international: nil)]).decorate
```
avinhurry, line 172:
```suggestion
        course = create(:course, enrichments: [create(:course_enrichment, fee_uk_eu: nil, fee_international: nil)]).decorate
```

spec/factories/courses.rb
```diff
@@ -193,6 +193,18 @@
       end
     end
 
+    trait :fee do
```
avinhurry, line 196: 👍🏼

avinhurry (approved):

PR #4564, opened 2024-10-01
Feature flags must be available even if Find is down

inulty-dfe:
## Context

We have a redirect that comes into effect before we re-open Find. This redirect prevents us from reaching the Feature Flags on Find. We need to reach the feature flags to turn off bursary information

Skip the redirect on the feature flags controller

## Things to check

- [ ] Rebased main
- [ ] Cleaned commit history


inulty-dfe:
[Feature flags page was not accessible until Find opens](https://trello.com/c/s22iwN4T/2405-feature-flags-page-was-not-accessible-until-find-opens)

PR #4347, opened 2024-07-08
[1980] Set `personal_qualities` and `other_requirements` to nil during rollover

avinhurry:
### Context

We previously opened a PR to delete the keys. This causes validation to kick in and prevents the rollover script from executing properly. 

We still have `personal_qualities` and  `other_requirements` in the current cycle.

### Changes proposed in this pull request

 Set `personal_qualities` and `other_requirements` to `nil` during rollover.

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally
- [x] Inform data insights team due to database changes

PR #4566, opened 2024-10-01
Revert removing `personal_qualities` and `other_requirements` from rollover script

avinhurry:
## Context

We added this to the rollover script in #4347 to remove all the `personal_qualities` and `other_requirements` from all of the courses since we no longer surface the information.

Now that all courses have had this information erased, we should remove this from the rollover script as it is unnecessary and could confuse people in the future.


avinhurry:
[Revert removing personal_qualities and other requirements from rollover script](https://trello.com/c/Y2yDxeAe/1976-revert-removing-personalqualities-and-other-requirements-from-rollover-script)

tomas-stefano (approved):

PR #4569, opened 2024-10-02
[2415] Update `funding_type` to `funding`

avinhurry:
## Context

We have added a deprecation warning to `funding_type` and re-directed people to the database backed version (`funding`) instead.

We want to update our code to use the `funding` instead of `funding_type` so that it is cleaner and it doesn’t cause confusion in the future.

We are aiming for the low hanging fruits here, so if there are any that are particular difficult, this will be out of scope.

- We have updated `funding_type` to `funding` in the codebase (all or at least most)

- We receive less (or no) deprecation warnings relating to `funding_type`

## Guidance to review

Have a click around and ensure all functionality continues to work. Especially functionality related to funding, e.g creating a course, changing funding, changing to a TDA course etc.

`funding_type` is still coming through on the API

image

## Things to check

- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -83,7 +83,7 @@ class CourseDecorator < ApplicationDecorator
       'salary' => 'Salary',
       'apprenticeship' => 'Teaching apprenticeship - with salary',
       'fee' => 'Fee - no salary'
-    }[object.funding_type]
+    }[object.funding]
   end
 
   def subject_name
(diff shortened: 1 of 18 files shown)


avinhurry:
[Change easy `funding_type` to `funding`](https://trello.com/c/2eXbWQeP/2415-change-easy-fundingtype-to-funding)

elceebee (approved):
I've checked the api on the review app and see funding_type as expected. and I've  clicked around a lot on the review app as well. Looks good! Very happy to see the deprecation warnings disappear!

PR #4577, opened 2024-10-07
Use old financial support field only for 2024 courses and previous cycles only

tomas-stefano:
## Context

Provider raised a support ticket about financial support duplication, Reading the code I understood this field is only for "< 2025 courses"

Support ticket: 

https://trello.com/c/7pXkBeQ8/309-fees-and-financial-support-info-duplicated

### DIFF
--- a/app/components/shared/courses/financial_support/fees_and_financial_support_component/view.html.erb
+++ b/app/components/shared/courses/financial_support/fees_and_financial_support_component/view.html.erb
@@ -18,9 +18,4 @@
       <%= render partial: "shared/courses/financial_support/loan" %>
     <% end %>
   <% end %>
-
-  <% if financial_support.present? && !course.recruitment_cycle.after_2024? %>
-    <h3 data-qa="course__financial_support_details" class="govuk-heading-m">Financial support from the training provider</h3>
-    <%= markdown(financial_support) %>
-  <% end %>
 </div>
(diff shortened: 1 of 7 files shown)


app/components/shared/courses/financial_support/fees_and_financial_support_component/view.html.erb
```diff
@@ -19,7 +19,7 @@
     <% end %>
   <% end %>
 
-  <% if financial_support.present? %>
+  <% if financial_support.present? && !course.recruitment_cycle.after_2024? %>
```
dcyoung-dev, line 22: It's not possible to view courses from previous cycles - how do you feel about completely removing this from the view?

tomas-stefano:
[Remove duplicated financial support section from Find](https://trello.com/c/fEgmMhGb/2466-remove-duplicated-financial-support-section-from-find)

app/components/shared/courses/financial_support/fees_and_financial_support_component/view.html.erb
tomas-stefano: I was thinking that, but also I was thinking that providers are worry for tomorrow and I am sign off soon too. I can try give it a go for 10 minutes and see if is only few specs breaks to change.

tomas-stefano: Also the financial support is all over the data (on enrichments) and were copied over to this cycle. I think we should amend somehow as well.

tomas-stefano: > It's not possible to view courses from previous cycles - how do you feel about completely removing this from the view?

This is done and pushed.

dcyoung-dev (approved):

PR #4589, opened 2024-10-14
Added a AccreditedProviderUpdater service

dcyoung-dev:
## Context

We have identified several Providers that have incorrect Accredited Provider details. We want to be able to update these Providers and their Courses.

Added service to handle the logic involved in updating the Accredited Provider
This is loosely based on a [gist](https://gist.github.com/avinhurry/ab64d4dc83dbc018b1dbda1eb2195624) shared by @avinhurry 

## Guidance to review

N/A

## Things to check

- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/services/publish/accredited_provider_updater.rb
+++ b/app/services/publish/accredited_provider_updater.rb
@@ -15,22 +15,37 @@ module Publish
     end
 
     def update_provider
+      provider.update!(accrediting_provider: 'not_an_accredited_provider',
+                       accrediting_provider_enrichments: new_accrediting_provider_enrichments)
+    end
+
+    def update_courses
+      courses = provider.courses
+      courses.update_all(accredited_provider_code: new_accredited_provider.provider_code)
+    end
+
+    private
+
+    def new_accrediting_provider_enrichments
+      existing_accrediting_provider_enrichments = provider.accrediting_provider_enrichments || []
+
+      return existing_accrediting_provider_enrichments if new_accredited_provider_code_in_enrichments?(
+        existing_accrediting_provider_enrichments
+      )
+
       accredited_provider_enrichment = AccreditingProviderEnrichment.new(
         {
           UcasProviderCode: new_accredited_provider.provider_code,
           Description: ''
         }
       )
-      accrediting_provider_enrichments = provider.accrediting_provider_enrichments || []
-      accrediting_provider_enrichments << accredited_provider_enrichment
 
-      provider.update!(accrediting_provider: 'not_an_accredited_provider',
-                       accrediting_provider_enrichments:)
+      existing_accrediting_provider_enrichments << accredited_provider_enrichment
     end
 
-    def update_courses
-      courses = provider.courses
-      courses.update_all(accredited_provider_code: new_accredited_provider.provider_code)
+    def new_accredited_provider_code_in_enrichments?(accrediting_provider_enrichments)
+      ucas_provider_code = new_accredited_provider.provider_code
+      accrediting_provider_enrichments.map(&:UcasProviderCode).include?(ucas_provider_code)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


dcyoung-dev:
https://trello.com/c/Oi7GdhWA

app/services/publish/accredited_provider_updater.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def update_provider_and_courses
+      update_provider && update_courses
+    end
+
+    def update_provider
+      accredited_provider_enrichment = AccreditingProviderEnrichment.new(
+        {
+          UcasProviderCode: new_accredited_provider.provider_code,
+          Description: ''
```
tomas-stefano, line 21: It is not the first time I see this description field on enrichments. Why do they need to be empty here?

dcyoung-dev: I really don't know but they are set with a description key and I wanted to make sure it wasn't going to break anything else 🤞

app/services/publish/accredited_provider_updater.rb
```diff
@@ -0,0 +1,51 @@
...
+    attr_reader :provider, :new_accredited_provider, :recruitment_cycle
+
+    def initialize(provider_code:, recruitment_cycle_year:, new_accredited_provider_code:)
+      @recruitment_cycle = RecruitmentCycle.find_by!(year: recruitment_cycle_year)
+      @provider = Provider.find_by!(provider_code:, recruitment_cycle: @recruitment_cycle)
+      @new_accredited_provider = Provider.find_by!(provider_code: new_accredited_provider_code, recruitment_cycle: @recruitment_cycle)
+    end
+
+    def update_provider_and_courses
+      update_provider && update_courses
```
tomas-stefano, line 14: Does it matter if one returns true and the other false and when returns false we don't know what fail? 
(if was the first or second condition)

app/services/publish/accredited_provider_updater.rb
tomas-stefano: Yes makes sense. Would be interesting to take a look at the database just to understand why. 🤔 🤔 🤔

app/services/publish/accredited_provider_updater.rb
dcyoung-dev: Both the methods execute SQL directly so unlikely to return false and more likely to fail/error. 

We're under a lot of pressure to update these accredited providers beofre the end of the month so that Register data is accurate.

tomas-stefano (approved):

app/services/publish/accredited_provider_updater.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    def update_provider
+      provider.update_columns(accrediting_provider: 'not_an_accredited_provider',
+                              accrediting_provider_enrichments: new_accrediting_provider_enrichments)
+    end
+
+    def update_courses
+      courses = provider.courses
+      courses.update_all(accredited_provider_code: new_accredited_provider.provider_code)
```
tomas-stefano, line 24: If a provider can be accredited by multiple providers, this code could cause unintended behavior by replacing the accreditation for all courses, even those that might need to retain their original accredited provider.

Example: Provider A has courses that are accredited by Provider B and Provider C.
After running this code to set the accredited_provider_code to Provider D, all courses will be accredited by Provider D, which may overwrite the correct accreditation for the courses that were accredited by Provider B or Provider C.

This is my first time reading through the accredited provider on the Publish side so I could be missing some parts.

Is it expected?

dcyoung-dev: This is a good point. Maybe we should only update the courses that are accredited by the old accredited provider, I'll update the PR 👌

inulty-dfe: We would need to identify which accredited provider we were removing in order to specify which courses to update / not update.

For the immediate concern of updating specific providers to correct the data on behalf of the register team I think we can commit this service as is. 

In future, we could add a wizard service that prompted the support user to choose which courses should and should not have their accredited provider updated. This goes for the user permissions that exist on the providers, which also requires contextual selection.

PR #4607, opened 2024-10-21
Bump Ruby version to 3.3.5

gms-gs:
## Context

Ticket: https://trello.com/c/tnKwD1bs/173-tech-debt-update-ruby-version-to-335

As part of clearing tech debt we have bumped Ruby to 3.3.5

- Bump the Ruby version to 3.3.5
- Added `jemalloc` as recommended  here: https://guides.rubyonrails.org/tuning_performance_for_deployment.html#memory-allocators-and-configuration
- Further reading https://support.sentieon.com/appnotes/jemalloc/
- If you set LD_PRELOAD to the path of a shared object, that file will be loaded before any other library.

To validate that Ruby is setting correctly we ran the below, which shows us `/usr/lib/libjemalloc.so.2 => /usr/lib/libjemalloc.so.2 (0x7f80d6e34000)`

```
/app # ldd $(which ruby)
	/lib/ld-musl-x86_64.so.1 (0x7f80d70a7000)
	/usr/lib/libjemalloc.so.2 => /usr/lib/libjemalloc.so.2 (0x7f80d6e34000)
	libruby.so.3.3 => /usr/local/lib/libruby.so.3.3 (0x7f80d683c000)
	libz.so.1 => /lib/libz.so.1 (0x7f80d6822000)
	libgmp.so.10 => /usr/lib/libgmp.so.10 (0x7f80d67b8000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f80d70a7000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f80d651f000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x7f80d64fb000)
```
## Things to check

- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,12 +19,12 @@ FROM ruby:3.3.5-alpine3.20
 
 RUN apk add --no-cache libxml2
 
-RUN apk add --update --no-cache tzdata jemalloc && \
+RUN apk add --update --no-cache tzdata && \
     cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
     echo "Europe/London" > /etc/timezone
 
 RUN apk add --update --no-cache \
- postgresql-dev git ncurses shared-mime-info
+ postgresql-dev git ncurses shared-mime-info jemalloc
 
 ENV APP_HOME /app
 
@@ -40,6 +40,8 @@ RUN apk add --update --no-cache --virtual build-dependencies \
  bundle install --jobs=4 && \
  rm -rf /usr/local/bundle/cache && \
  apk del build-dependencies
+
+ENV LD_PRELOAD="/usr/lib/libjemalloc.so.2"
 
 COPY package.json yarn.lock ./
 RUN yarn install --frozen-lockfile && \


dcyoung-dev:
[Upgrade Ruby version 3.3.5](https://trello.com/c/4SJYg8bV/350-upgrade-ruby-version)

Dockerfile
Nitemaeric, line 1: Hmm, the Dockerfile here installs jemalloc, but it doesn't look to consume it. We'll need to update this.

Dockerfile
```diff
@@ -15,11 +15,11 @@ RUN bundle exec middleman build --build-dir=../public
 
 ###
 
-FROM ruby:3.2.4-alpine3.20
+FROM ruby:3.3.5-alpine3.20
 
 RUN apk add --no-cache libxml2
 
-RUN apk add --update --no-cache tzdata && \
+RUN apk add --update --no-cache tzdata jemalloc && \
```
Nitemaeric, line 22: Move this down to the end of line 27, as this line is specific to setting up `tzdata`.

Dockerfile
```diff
@@ -41,9 +41,11 @@ RUN apk add --update --no-cache --virtual build-dependencies \
  rm -rf /usr/local/bundle/cache && \
  apk del build-dependencies
 
+ENV LD_PRELOAD="/usr/lib/libjemalloc.so.2"
```
dcyoung-dev, line 44: Looking at a fresh Rails app, the Dockerfile doesn't contain this line. Is this because we're using Ruby alpine image? This is a wild guess and I'm only asking for my own knowledge 🤷‍♂️

tomas-stefano: Here: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/docker-entrypoint.tt#L5

dcyoung-dev: Nice one thanks for this - it makes much more sense.

PR #4616, opened 2024-10-24
Perf: Better response times from api/public/v1/courses

inulty-dfe:
## Context

Counting the number of courses in a recruitment cycle is very expensive. We do this on every request to the courses endpoint.

### This is an "agonising" endpoint in Skylight

image

## Loadtesting the change
The tests were run with Siege.

[Siege documentation](https://linux.die.net/man/1/siege) 

All tests were run against review apps. The review apps all had 1 replica with 512MiB memory.

### Before changes

```
inulty-dfe ~/Code/publish-teacher-training ( perf-api-providers) $ siege -c 10 -t 1m -if app/controllers/api/loadtest-urls-other-branch.txt

{       "transactions":                           91,
        "availability":                       100.00,
        "elapsed_time":                        59.50,
        "data_transferred":                    19.95,
        "response_time":                        5.97,
        "transaction_rate":                     1.53,
        "throughput":                           0.34,
        "concurrency":                          9.13,
        "successful_transactions":                31,
        "failed_transactions":                     0,
        "longest_transaction":                 19.52,
        "shortest_transaction":                 0.97
}71
}
```

### With just `loaded?` changes

```
inulty-dfe ~/Code/publish-teacher-training ( perf-api-providers) $ siege -c 10 -t 1m -if app/controllers/api/loadtest-urls-no-cache.txt

{       "transactions":                           75,
        "availability":                       100.00,
        "elapsed_time":                        59.49,
        "data_transferred":                    26.39,
        "response_time":                        7.16,
        "transaction_rate":                     1.26,
        "throughput":                           0.44,
        "concurrency":                          9.03,
        "successful_transactions":                45,
        "failed_transactions":                     0,
        "longest_transaction":                 19.76,
        "shortest_transaction":                 1.11
}
```

### With count query cached and `loaded?` changes

```
inulty-dfe ~/Code/publish-teacher-training ( perf-api-providers) $ siege -c 10 -t 1m -if app/controllers/api/loadtest-urls.txt

{       "transactions":                           94,
        "availability":                       100.00,
        "elapsed_time":                        59.86,
        "data_transferred":                    51.11,
        "response_time":                        6.07,
        "transaction_rate":                     1.57,
        "throughput":                           0.85,
        "concurrency":                          9.53,
        "successful_transactions":                84,
        "failed_transactions":                     0,
        "longest_transaction":                 13.18,
        "shortest_transaction":                 1.90
}
```

### Result

|Operation|Original|No cache changes|Cache changes|
|---|---|---|---|
| throughput |  0.34 | 0.44 | 0.85
| successful_transactions| 31 | 45 | 84 |
| data_transferred | 19.95 | 26.39 | 51.11|

### The URLs
The URLs used were variations on this list.
I added `no_cache` for the "no cache" test run.
I changed the PR number for the "original" test run.

image

- Use `loaded?` to avoid making requests to the database when we already have records in memory.
- Cache the course record count per recruitment cycle for a period of time.
- The `filter` field is not used in the cache key because it is ignored in the query. The API does not filter courses.

## Guidance to review

- How long should the course count be cached? I would suggest 1 day.

## Things to check

- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/api/public/v1/courses_controller.rb
+++ b/app/controllers/api/public/v1/courses_controller.rb
@@ -16,10 +16,12 @@ module API
         private
 
         def cached_course_count
-          if params[:no_cache]
+          year = permitted_params[:recruitment_cycle_year] || RecruitmentCycle.current.year
+
+          if permitted_params[:no_cache]
             courses.count('course.id')
           else
-            Rails.cache.fetch('api_course_count', expires_in: 5.minutes) do
+            Rails.cache.fetch("api_course_count_#{year}", expires_in: 5.minutes) do
               courses.count('course.id')
             end
           end
@@ -27,14 +29,18 @@ module API
 
         def courses
           @courses ||= APICourseSearchService.call(
-            filter: params[:filter],
-            sort: params[:sort],
+            filter: permitted_params[:filter],
+            sort: permitted_params[:sort],
             course_scope: recruitment_cycle.courses
           )
         end
 
         def include_param
-          params.fetch(:include, '')
+          permitted_params.fetch(:include, '')
+        end
+
+        def permitted_params
+          params.permit('page', 'no_cache', 'sort', 'per_page', 'courses', 'recruitment_cycle_year', 'include', 'filter' => %w[updated_since funding_type])
         end
       end
     end
--- a/app/controllers/api/public/v1/courses_controller.rb
+++ b/app/controllers/api/public/v1/courses_controller.rb
@@ -7,13 +7,23 @@ module API
         def index
           render jsonapi: paginate(courses),
                  include: include_param,
-                 meta: { count: courses.count('course.id') },
+                 meta: { count: cached_course_count },
                  class: API::Public::V1::SerializerService.call
         rescue ActiveRecord::StatementInvalid
           render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: :bad_request
         end
 
         private
+
+        def cached_course_count
+          if params[:no_cache]
+            courses.count('course.id')
+          else
+            Rails.cache.fetch('api_course_count', expires_in: 5.minutes) do
+              courses.count('course.id')
+            end
+          end
+        end
 
         def courses
           @courses ||= APICourseSearchService.call(
--- a/app/controllers/api/public/v1/courses_controller.rb
+++ b/app/controllers/api/public/v1/courses_controller.rb
@@ -18,12 +18,8 @@ module API
         def cached_course_count
           year = permitted_params[:recruitment_cycle_year] || RecruitmentCycle.current.year
 
-          if permitted_params[:no_cache]
+          Rails.cache.fetch("api_course_count_#{year}", expires_in: 5.minutes) do
             courses.count('course.id')
-          else
-            Rails.cache.fetch("api_course_count_#{year}", expires_in: 5.minutes) do
-              courses.count('course.id')
-            end
           end
         end
 
@@ -40,7 +36,7 @@ module API
         end
 
         def permitted_params
-          params.permit('page', 'no_cache', 'sort', 'per_page', 'courses', 'recruitment_cycle_year', 'include', 'filter' => %w[updated_since funding_type])
+          params.permit('page', 'sort', 'per_page', 'courses', 'recruitment_cycle_year', 'include', 'filter' => %w[updated_since funding_type])
         end
       end
     end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -594,7 +594,7 @@ class Course < ApplicationRecord
   end
 
   def last_published_at
-    return enrichments.max_by(&:last_published_timestamp_utc)&.last_published_timestamp_utc if enrichments.loaded?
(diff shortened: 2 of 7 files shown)


app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -7,24 +7,40 @@ class CoursesController < API::Public::V1::ApplicationController
...
 
         private
 
+        def cached_course_count
+          year = permitted[:recruitment_cycle_year] || RecruitmentCycle.current.year
+
+          if permitted[:no_cache]
+            courses.count('course.id')
+          else
+            Rails.cache.fetch("api_course_count_#{year}", expires_in: 5.minutes) do
```
Nitemaeric, line 24: We can let Rails handle the stringification our cache key here for our convenience. Not a blocker.

```suggestion
            Rails.cache.fetch([:api_course_count, year], expires_in: 5.minutes) do
```

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -7,24 +7,40 @@ class CoursesController < API::Public::V1::ApplicationController
...
           render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: :bad_request
         end
 
         private
 
+        def cached_course_count
+          year = permitted[:recruitment_cycle_year] || RecruitmentCycle.current.year
+
+          if permitted[:no_cache]
+            courses.count('course.id')
```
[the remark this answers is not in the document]
inulty-dfe: Yeah good point, I saw a discussion in `DfE::Analytics` around this too. I would be interested to see if we could optimise the query but I puropsely avoided getting into that here.

app/models/course.rb
```diff
@@ -585,6 +585,8 @@ def gcse_grade_required
   end
 
   def last_published_at
+    return enrichments.select(&:last_published_timestamp_utc).max_by(&:last_published_timestamp_utc)&.last_published_timestamp_utc if enrichments.loaded?
+
     enrichments.maximum(:last_published_timestamp_utc)
   end
```
Nitemaeric, line 591: This looks like it can be generalised for all associations.

Have you considered looking into [ActiveRecord/Associations/CollectionProxy](https://api.rubyonrails.org/v7.2/classes/ActiveRecord/Associations/CollectionProxy.html)? A `#maximum` method there might be able to address this issue across the entire app with a single change.

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -7,24 +7,40 @@ class CoursesController < API::Public::V1::ApplicationController
...
         rescue ActiveRecord::StatementInvalid
           render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: :bad_request
         end
 
         private
 
+        def cached_course_count
+          year = permitted_params[:recruitment_cycle_year] || RecruitmentCycle.current.year
+
+          if permitted_params[:no_cache]
```
Nitemaeric, line 21: Below might be something for another PR.

It might be more conventional to read the `Cache-Control` request header and return cached vs non-cached results.

I haven't quite dug into if that's indeed the right header, or what the header value should be (I _think_ `no-cache`?), and also if my assumption is considered standard practice, but it might be worth looking into.

inulty-dfe: I'm happy to remove this, I don't think it should be merged. Undocumented api params probably aren't a good idea.

As for the cache headers, I believe those (response headers) are for instructing how the client should cache, not how the server should respond.

app/models/course.rb
inulty-dfe: Yeah that looks tempting. but I would be careful about doing that. It's something we should discuss though, it seems like a useful thing to do.

app/controllers/api/public/v1/courses_controller.rb
Nitemaeric: I _think_ that [this section](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#cache_directives) is indicating that there are request header standards too, but yeah, I haven't seen it in practice yet. I wonder how the browser implements the "Disable cache" checkbox.

But yeah, I agree. Let's remove the `no_cache` for now.

app/controllers/api/public/v1/courses_controller.rb
```diff
@@ -7,24 +7,36 @@ class CoursesController < API::Public::V1::ApplicationController
...
         rescue ActiveRecord::StatementInvalid
           render json: { status: 400, message: 'Invalid changed_since value, the format should be an ISO8601 UTC timestamp, for example: `2019-01-01T12:01:00Z`' }.to_json, status: :bad_request
         end
 
         private
 
+        def cached_course_count
+          year = permitted_params[:recruitment_cycle_year] || RecruitmentCycle.current.year
+
+          Rails.cache.fetch("api_course_count_#{year}", expires_in: 5.minutes) do
```
tomas-stefano, line 21: Let's have a discussion about this expiration time?

tomas-stefano (approved):
Everything looks good. Great initiative. ✌️ 

Let's have a separate discussion about the cache & expiration.

PR #4620, opened 2024-10-25
Users should not see any providers without explicit association

inulty-dfe:
## Context

This PR fixes a bug that allowed any provider user to visit the courses pages (and more ) of any provider in the system. 
We've added an `authorize provider` call to a before action in the parent PublishController which authorizes all requests to any action in that controller.

Some of the tests had to be changed so that the user perfroming the action under test had an explicit association with the provider they are interacting with in the test.

image

## Things to check

- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/app/controllers/publish/providers_controller.rb
+++ b/app/controllers/publish/providers_controller.rb
@@ -101,7 +101,7 @@ module Publish
     private
 
     def provider
-      @provider ||= recruitment_cycle.providers.find_by!(provider_code: params[:provider_code] || params[:code])
+      @provider ||= recruitment_cycle.providers.find_by(provider_code: params[:provider_code] || params[:code])
     end
 
     def providers
--- a/app/controllers/publish/publish_controller.rb
+++ b/app/controllers/publish/publish_controller.rb
@@ -9,6 +9,9 @@ module Publish
     before_action :check_interrupt_redirects
     before_action :clear_previous_cycle_year_in_session, unless: -> { FeatureService.enabled?('rollover.can_edit_current_and_next_cycles') }
 
+    # Protect every action of a provider
+    before_action { authorize provider, :show? if provider }
+
     after_action :verify_authorized
 
     # DFE Analytics namespace
--- a/app/components/persona.html.erb
+++ b/app/components/persona.html.erb
@@ -1,4 +1,4 @@
-<%= form_tag("/auth/developer/callback") do %>
+<%= form_tag("/auth/developer/callback", method: :get) do %>
   <%= hidden_field_tag "email", email_address %>
   <%= hidden_field_tag "first_name", first_name %>
   <%= hidden_field_tag "last_name", last_name %>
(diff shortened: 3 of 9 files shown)


app/controllers/publish/publish_controller.rb
```diff
@@ -9,6 +9,9 @@ class PublishController < ApplicationController
     before_action :check_interrupt_redirects
     before_action :clear_previous_cycle_year_in_session, unless: -> { FeatureService.enabled?('rollover.can_edit_current_and_next_cycles') }
 
+    # Protect every action of a provider
+    before_action { authorize provider, :show? if provider }
```
Nitemaeric, line 13: Do we need this comment?

Also, it might be just me, but it might be easier to read either:

1. Create an intention-revealing method name for this:

```suggestion
    before_action :authorize_provider_sub_pages

    def authorize_provider_sub_pages
      authorize provider, :show? if provider
    end
```

3. Pull down the code into a multiline block:

```suggestion
    before_action do
      authorize provider, :show? if provider
    end
```

app/controllers/publish/providers_controller.rb
```diff
@@ -101,7 +101,7 @@ def search
     private
 
     def provider
-      @provider ||= recruitment_cycle.providers.find_by!(provider_code: params[:provider_code] || params[:code])
+      @provider ||= recruitment_cycle.providers.find_by(provider_code: params[:provider_code] || params[:code])
```
Nitemaeric, line 104: This seems like an important change.

Is the intention here to raise a 403 Forbidden error instead of a 404 Not Found error when a provider is not found within a recruitment cycle?

Nitemaeric (commented):
There's a few unrelated fixes to the PR title.

Not a blocker and I know it's easier to fix things once you run into them, rather than switch branches etc.

But overall, if we can try to separate out unrelated changes into their own small PRs, it'll make code review simpler as there is only a single storyline to follow.

Nitemaeric (approved):

app/controllers/publish/providers_controller.rb
inulty-dfe: We call `provider` in a before action now, whereas we didn't before.
Some actions do not require a provider, `/publisher/organisations` for example.
There is not provider parameter in this case and no Provider can be found.
That will trigger a 404.

I haven't thought enough about how to do this better.

app/controllers/publish/publish_controller.rb
inulty-dfe: I know I'm not as averse to code comments as other people are. I can remove the comment.

I wonder if there is more work to be done here if I need to add the `if provider` exists.

app/controllers/publish/providers_controller.rb
Nitemaeric: 🤔

PR #4634, opened 2024-10-30
Delegate all authorize provider in Publish to PublishController

inulty-dfe:


Nitemaeric:
I've just caught up with what you mentioned in the previous PR.

There are cases where the path does not contain `:recruitment_cycle_year`, therefore there is no `provider` instance found.

I think a controller concern for nested controllers would work well, alternatively, an additional subclass would work, but too many inheritance levels can become unwieldy.

```ruby
module AuthorisesProviderAccess
  extend ActiveSupport::Concern

  included do
    before_action do
      authorize provider, :show?
    end
  end
end

class Publish::Provider::UsersController < PublishController
  include AuthorisesProviderAccess # Add this to controllers where `provider` should be looked up.

  # ...
end
```

The inheritance alternative:

```ruby
class Publish::Provider::ApplicationController < Publish::PublishController
  before_action do
    authorize :provider, :show?
  end
end

class Publish::Provider::UsersController < Publish::Provider::ApplicationController
  # ...
end
```

Nitemaeric (approved):

PR #4636, opened 2024-10-30
Remove kaminari and only use pagy for frontend and api pagination

gms-gs:
## Context

Ticket: https://trello.com/c/ykmLnKYb/176-tech-debt-only-use-pagy-gem-for-pagination

Find and Publish used two gems for pagination, `kaminari ` for the frontend and `pagy ` for the api. We don't need to pagination gems. So we are removing `kaminari ` and using `pagy ` for all pagination.

In addition change the default pagination limit to 10: https://trello.com/c/ZJKmUqjA/48-improve-results-6-7-update-number-of-results-per-page-from-25-to-10

Remove all references to `kaminari ` and update code to use `pagy` implementation and update all pagination rendering to use the `govuk_pagination` component.

## Guidance to review

- Run the app and navigate through Find and Publish to pagination still working

### DIFF
--- a/app/controllers/concerns/pagy_pagination.rb
+++ b/app/controllers/concerns/pagy_pagination.rb
@@ -8,7 +8,7 @@ module PagyPagination
   end
 
   def paginate(scope)
-    @pagy_results ||= pagy(scope, limit: per_page, page:)
+    @pagy_results ||= pagy(scope, limit: per_page, page:, overflow: :exception)
 
     pagy_results.second
   end
--- a/config/initializers/pagy.rb
+++ b/config/initializers/pagy.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
 require 'pagy/extras/metadata'
+require 'pagy/extras/overflow'
 
+# `empty_page` is default for the UI
+# exception should be used in all API contexts
+Pagy::DEFAULT[:overflow] = :empty_page
 Pagy::DEFAULT[:limit] = 10


app/controllers/find/results_controller.rb
```diff
@@ -4,14 +4,16 @@ module Find
   class ResultsController < ApplicationController
     before_action :render_feedback_component
 
+    PAGY_PER_PAGE = 10
```
Nitemaeric, line 7: This value is reused quite a bit. It seems that 10 results is the most common amount per page? I haven't checked the other parts, but does it make more sense to set this as the global default for pagy?

Nitemaeric: Wait, wait.. Pagy comes with a way to set default values in the initializer. `Pagy::DEFAULT[:limit] = 10`, I believe. Check their docs :)

app/controllers/find/results_controller.rb
```diff
@@ -11,7 +11,7 @@ def index
 
       @results_view = ResultsView.new(query_parameters: matched_params)
       @filters_view = ResultFilters::FiltersView.new(params: matched_params)
-      @courses = @results_view.courses.page params[:page]
+      @pagy, @courses = pagy(@results_view.courses, page: params[:page])
```
Nitemaeric, line 14: I think the `page: params[:page]` is implicit for Pagy.

```suggestion
      @pagy, @courses = pagy(@results_view.courses)
```

inulty-dfe (commented):
Are we handling page overflow?

https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/config/initializers/pagy.rb#L4C1-L8C39

Nitemaeric (approved):

spec/features/publish/providers_index_spec.rb
```diff
@@ -117,7 +117,7 @@ def i_can_search_with_provider_details
   end
 
   def i_should_see_the_pagination_link
-    expect(publish_providers_index_page.pagination_pages.text).to eq('2 of 2')
+    expect(publish_providers_index_page.pagination_pages.text).to eq('2 of 4')
```
inulty-dfe, line 120: Why did this change?

inulty-dfe (commented):
Looks good to me. Just the question of why the page count changed in the test

spec/features/publish/providers_index_spec.rb
inulty-dfe: That's not in the PR description or the Trello ticket

inulty-dfe (approved):
Great work

PR #4638, opened 2024-10-30
Update rackup

Nitemaeric:
## Context

There are 6 [security dependabot alerts](https://github.com/DFE-Digital/publish-teacher-training/security/dependabot) related to `webrick` and `rexml` - These are dependencies of the `rackup` gem.

We should update the `rackup` gem to the latest version.


Nitemaeric:
Once we migrate from Sidekiq to Solid Queue and Active Job, we'll be able to upgrade.

cc: @DFE-Digital/publish-and-find

PR #4645, opened 2024-11-01
Replace instances of PGDE with QTS / change API course summary

tomas-stefano:
## Context

There is inconsistency of terminology between GiT and Find. GiT use “QTS with PGDE” (which is more clear/intuitive) whereas Find uses “PGDE with QTS”.

## Changes

This PR replaces all instances of `PGDE with QTS` to `QTS with PGDE`. 

## Coordination between other teams

It was decided by Apply and Find teach leads to change the API content of the course description to the expected ones 

1. PGCE with QTS becomes QTS with PGCE
2. PGDE with QTS becomes QTS with PGDE

In order to coordinate this change with this other team this PR:

1. Changes everything in the UI
2. Uses a feature flag for the `course.summary` field in the API.

So this is the plan to deploy this:

- [x] Merge the feature flag PR 
- [ ] Merge the code that uses the feature flag (This one!)
- [ ] Turn on the feature flag on QA when having a session with Register & Apply ([PR](https://github.com/DFE-Digital/publish-teacher-training/pull/4652))
- [ ] Have the Register & Apply sign off that everything worked
- [ ] Turn on the feature flag on Production ([PR](https://github.com/DFE-Digital/publish-teacher-training/pull/4653)) and also deploy the API docs change here about the content change on API on the history/updates section.
- [ ] Cleanup the feature flag (PR to be added here)

### DIFF
--- a/app/components/find/courses/qualifications_summary_component/view.html.erb
+++ b/app/components/find/courses/qualifications_summary_component/view.html.erb
@@ -20,7 +20,7 @@
 <% when "pgde_with_qts" %>
   <%= govuk_details(summary_text:) do %>
     <p class="govuk-body">
-      A postgraduate diploma in education (PGDE) with qualified teacher status (QTS) will allow you to teach in state schools in England and may allow you to teach in other parts of the UK.
+      Qualified teacher status (QTS) with a postgraduate diploma in education (PGDE) will allow you to teach in state schools in England and may allow you to teach in other parts of the UK.
     </p>
     <p class="govuk-body">
       It may also allow you to teach overseas, though you should always check what qualifications are needed in the country you’d like to teach in.
--- a/app/components/find/courses/qualifications_summary_component/view.html.erb
+++ b/app/components/find/courses/qualifications_summary_component/view.html.erb
@@ -1,6 +1,6 @@
-<% case find_outcome %>
-<% when "QTS" %>
-  <%= govuk_details(summary_text: "QTS") do %>
+<% case course.qualification %>
+<% when 'qts' %>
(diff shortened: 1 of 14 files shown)


tomas-stefano:
[Review and replace instances of "PGDE with QTS" to "QTS with PGDE" on Publish & Find](https://trello.com/c/fbggMEzh/33-review-and-replace-instances-of-pgde-with-qts-to-qts-with-pgde-on-publish-find)

app/components/find/courses/qualifications_summary_component/view.html.erb
```diff
@@ -1,24 +1,24 @@
-<% case find_outcome %>
-<% when "QTS" %>
-  <%= govuk_details(summary_text: "QTS") do %>
+<% case course.qualification %>
+<% when 'qts' %>
```
tomas-stefano, line 2: I changed the way it was done slightly. 

Before: it was using the translation to do a case for another translation.
After: Passing the course qualification and then translate.

I thought to split into partials but then I rollback and the code will be like 

View -> Component -> Another component -> Multiple partials

app/models/course.rb
```diff
@@ -516,11 +516,20 @@ def program_type_description
   def description
     return qualifications_description if teacher_degree_apprenticeship?
 
-    study_mode_string = (full_time_or_part_time? ? ', ' : ' ') +
-                        study_mode_description
     qualifications_description + study_mode_string + program_type_description
   end
```
tomas-stefano, line 520: I know the summary above is almost a duplication of this description but my plan is to remove this description field after the feature flag is removed.

PR #4650, opened 2024-11-04
Fix Placement school row not always displaying on the search results page

gms-gs:
## Context

Ticket: https://trello.com/c/suO632ld/190-placement-school-row-not-always-displaying-on-the-search-results-page

Removed some logic which didn't make sense having there in my opinion. The logic checked if the school had a website and if it did not hid the `Placement school` row. Removing that conditional logic from the view seems to fix the issue
 
- Remove the website checking logic

## Guidance to review

Visit to see fixed issue https://find-review-4650.test.teacherservices.cloud/results?age_group=secondary&applications_open=true&can_sponsor_visa=false&has_vacancies=true&l=3&provider.provider_name=Agnus+Dei+Catholic+Schools&subjects%5B%5D=W1&university_degree_status=true&visa_status=false to see the row now being show. 

Prod app to see the broken implementation: https://find-teacher-training-courses.service.gov.uk/results?age_group=secondary&applications_open=true&can_sponsor_visa=false&has_vacancies=true&l=3&provider.provider_name=Agnus+Dei+Catholic+Schools&subjects%5B%5D=W1&university_degree_status=true&visa_status=false

### DIFF
--- a/app/components/find/results/search_result_component.rb
+++ b/app/components/find/results/search_result_component.rb
@@ -39,7 +39,7 @@ module Find
 
       def location_label
         if course.fee_based?
-          t('.fee_based.location', count: sites_count)
+          t('.fee_based.location')
         else
           t('.salary_based.location', count: sites_count)
         end
--- a/config/locales/find.yml
+++ b/config/locales/find.yml
@@ -194,9 +194,7 @@ en:
       search_result_component:
         no_degree_required: No degree required
         fee_based:
-          location:
-            one: Placement school
-            other: Placement schools
+          location: Placement schools
         salary_based:
           location:
             one: Employing school


kelliedesigner:
Looks good to me!

While we are on this PR, it would be great to do a small update of removing the logic which currently changes the label "Placement schools" to be a singular if there is only one placement school, as the design intention was to have the label stay the same for all courses so that there is less changing to make it easier to compare courses.

Nitemaeric (approved):

app/components/find/results/search_result_component.html.erb
```diff
@@ -1,22 +1,17 @@
...
               <% else %>
-                <%= render partial: "find/results/by_country_or_provider",
-                           locals: { sites_count:, school_term: } %>
+                <%= render partial: "find/results/by_country_or_provider", locals: { sites_count:, school_term: } %>
               <% end %>
-
             <% else %>
-              Not listed yet
-            <% end %>
+            Not listed yet
```
Nitemaeric, line 14: We should look into proper localisation later on.

PR #4654, opened 2024-11-05
Update miscellaneous development setup

Nitemaeric:
## Context

Some minor cleanup.

- Add `.env.local` and `.env.*.local` to `.gitignore`.
  - The recommendation is to use `.env.development.local` for local ENV vars - namely to avoid any clash with other environments or in the very rare case that this file is added to git, it doesn't effect production - Unlike `.env` would.
- Update local default authentication mode to `persona`.
- Expose ports of `docker-compose.yml` services.

## Guidance to review

- Does this break any of your workflows?

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -1,10 +1,14 @@
 authentication:
   secret: secret
+  mode: persona
+
 mcbg:
   redis_password:
+
 system_authentication_token: "Ge32"
 log_level: debug
 gcp_api_key: please_change_me
+
 environment:
   label: "development"
   name: "development"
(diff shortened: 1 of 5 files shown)


Nitemaeric:
It seems we rely on docker for GitHub Actions builds 😅

config/settings/development.yml
```diff
@@ -1,10 +1,14 @@
 authentication:
   secret: secret
+  mode: persona
```
inulty-dfe, line 3: This may not be necessary anymore

https://github.com/DFE-Digital/publish-teacher-training/blob/main/bin/setup#L22-L31

Nitemaeric: Removed the setup code. :)

inulty-dfe (approved):
LGTM

PR #4666, opened 2024-11-08
Set Pagy limit to 10 and set overflow exception for API

gms-gs:
## Context

Ticket: https://trello.com/c/ZJKmUqjA/48-improve-results-6-7-update-number-of-results-per-page-from-25-to-10

Research has shown that candidates find it overwhelming to have too any results of courses per page. They have also said that having a long list of results makes it difficult to relate the results to the filters, and to use the filters.

This PR is to reduce the number of results per page from 25 courses to 10 courses.

Set `Pagy::DEFAULT[:limit] = 10` to ensure all instances of Pagy only show 10 items per page

## Guidance for review
Viewing the results page would result in only showing 10 items per page

### DIFF
--- a/spec/features/publish/providers_index_spec.rb
+++ b/spec/features/publish/providers_index_spec.rb
@@ -23,6 +23,7 @@ feature 'Providers index' do
     i_should_see_the_provider_list
     i_should_see_the_admin_search_box
     i_should_see_the_pagination_link
+    i_should_only_see_10_providers_per_page
     i_can_search_with_provider_details
 
     i_should_see_the_change_organisation_link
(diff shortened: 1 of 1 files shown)


spec/features/publish/providers_index_spec.rb
```diff
@@ -117,7 +117,7 @@ def i_can_search_with_provider_details
   end
 
   def i_should_see_the_pagination_link
-    expect(publish_providers_index_page.pagination_pages.text).to eq('2 of 2')
+    expect(publish_providers_index_page.pagination_pages.text).to eq('2 of 4')
```
tomas-stefano, line 120: Can we add a step to assert the number of courses within the search results too? 

Or something that might assert `Pagy::DEFAULT[:limit]` or just `10`. 

Keen to hear other thoughts :)

spec/features/publish/providers_index_spec.rb
```diff
@@ -23,6 +23,7 @@
     i_should_see_the_provider_list
     i_should_see_the_admin_search_box
     i_should_see_the_pagination_link
+    i_should_only_see_10_providers_per_page
```
tomas-stefano, line 26:
```suggestion
    i_should_only_see_10_results_per_page
```

tomas-stefano: Ignore this, I thought it was courses!

tomas-stefano (approved):

PR #4677, opened 2024-11-12
Use Pagy for pagination

gms-gs:
## Context

Ticket: https://trello.com/c/ykmLnKYb/176-tech-debt-only-use-pagy-gem-for-pagination

Find and Publish used two gems for pagination, `kaminari ` for the frontend and `pagy ` for the api. We don't need to pagination gems. So we are removing `kaminari ` and using `pagy ` for all pagination.

Remove all references to `kaminari ` and update code to use `pagy` implementation and update all pagination rendering to use the `govuk_pagination` component.

## Guidance to review

- Run the app and navigate through Find and Publish to pagination still working


Nitemaeric (approved):
Code-wise, this looks fine. I'm cautious about how many changes are going in all at the same time, based on the last deployment.

But if you're confident, let's do it. We can always roll this one back quite easily.

PR #4686, opened 2024-11-15
Update filter default: uncheck Study type & Qualifications checkboxes

tomas-stefano:
## Context

Currently the default view of filters has the ‘Study type’ and ‘Qualifications’ checkboxes checked to indicate that they are all selected.

We want to change this to all the checkboxes being unchecked by default, still to indicate that they are all selected.

*Before*
Screenshot 2024-11-15 at 12 05 15

*After*
Screenshot 2024-11-15 at 12 05 33

## Guidance to review

1. Do a Find search
2. Study type and qualification filters should be unchecked by default
3. Select all study types and qualfications
4. Study type and qualification filters should be checked
5. Now filter by specific study type or qualification
6. Does the filters still continuing to work as expected?

### DIFF
--- a/spec/support/filters_feature_specs_helper.rb
+++ b/spec/support/filters_feature_specs_helper.rb
@@ -9,4 +9,8 @@ module FiltersFeatureSpecsHelper
   def and_apply_the_filters
     find_results_page.apply_filters_button.click
   end
+
+  def query_params(uri)
+    ActionDispatch::Request.new({ 'QUERY_STRING' => uri.query }).query_parameters
+  end
 end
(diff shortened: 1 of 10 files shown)


tomas-stefano:
[Improve results - 0/7 Update how filters behave](https://trello.com/c/UuTHCi74/216-improve-results-0-7-update-how-filters-behave)

Nitemaeric (approved):
🙌

spec/support/filters_feature_specs_helper.rb
```diff
@@ -9,4 +9,8 @@ def when_i_visit_the_find_results_page
   def and_apply_the_filters
     find_results_page.apply_filters_button.click
   end
+
+  def query_params(uri)
+    ActionDispatch::Request.new({ 'QUERY_STRING' => uri.query }).query_parameters
```
inulty-dfe, line 14: Nice hacking. Would be nicer if we didn't need to create a request but heyho :)

inulty-dfe (approved):

kelliedesigner (approved):
LGTM!

PR #4680, opened 2024-11-14
916 publish clean up more paas code

RMcVelia:
## Context

Clean up github workflows and terraform config  

https://trello.com/c/DIeRBab2/916-publish-clean-up-more-paas-code

## Changes proposed in this pull request

Removing v2 and aks references

Initially will leave rails app references to _aks in place

## Guidance to review

make env deploy-plan
deploy review app (close PR to delete)

PR #4693, opened 2024-11-18
Remove references to "_aks" environment names

Nitemaeric:
## Context

We've [recently dropped](https://github.com/DFE-Digital/publish-teacher-training/pull/4680) "_aks" suffixes from AKS configurations.

We should remove the "_aks" suffixes from application code too.

- Remove previous non-aks environment configurations.
- Remove `_aks` suffix from environment names.

## Guidance to review

Let's see if `review` works :D


Nitemaeric:
[Tech debt: Remove "_aks" suffixes from environment names](https://trello.com/c/ruljxvNG/249-tech-debt-remove-aks-suffixes-from-environment-names)

tomas-stefano (approved):
From the app point of view, looks good 👍 

Is there anything we need to do on infra side or did they already remove references to `_aks` as well? 

Such as copy/moving the settings.yml for each environment and so on?

RMcVelia:
Could you also rename maintenance_page/manifests/env_aks to just env, or I can update on a separate pr if preferred.

Nitemaeric:
> Could you also rename maintenance_page/manifests/env_aks to just env, or I can update on a separate pr if preferred.

Done!

I'm curious about how it knows how to pick up the changes though. The workflow file has only `production` and `qa` as options, yet the directories were named `production_aks` and `qa_aks`. Was this a bug?

PR #2655, opened 2022-05-13
Switch new publish on for QA & Staging

darokel:
### Context

- https://trello.com/c/elvNaWG9/569-switch-www-and-www2

### Changes proposed in this pull request

- Updates the base url of the app
- Adds guard for handling redirects in prod behind a flag (currently off)

PR #4695, opened 2024-11-18
Remove www2 route redirection

Nitemaeric:
## Context

The `www2` host is no longer resolved within our DNS routes. We can safely drop this routing constraint/redirection.

The `www2` host was used as a preview host for the new publish application: https://github.com/DFE-Digital/publish-teacher-training/pull/2655

- Drop the `www2` route group.

## Guidance to review

- Try visiting https://www2.publish-teacher-training-courses.service.gov.uk/, it should fail. This means we can safely drop the host constraint and routes within the group.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,10 +10,6 @@ Rails.application.routes.draw do
     get '/', to: redirect('/docs/')
   end
 
-  constraints(host: /www2\./) do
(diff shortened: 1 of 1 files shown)


tomas-stefano (approved):
Sounds good to me. ✌️ I suggest add someone from infra to approve too. :)

RMcVelia (approved):

PR #4697, opened 2024-11-18
Split skylight metrics for review apps out per review app

Nitemaeric:
## Context

We can send our Skylight metrics for review apps out into separate environments. In the case that we want to understand the performance of a review app, we will be able to distinguish between each review app and its metrics.

- Send review app metrics to Skylight per PR number.

## Guidance to review

- See https://www.skylight.io/app/applications/2qpKL45hvoni/recent/5m/endpoints for example.

### DIFF
--- a/app/components/phase_banner.rb
+++ b/app/components/phase_banner.rb
@@ -7,7 +7,7 @@ class PhaseBanner < ViewComponent::Base
   end
 
   def environment_label
-    Settings.environment.label.capitalize
+    Settings.environment.label
   end
 
   def environment_colour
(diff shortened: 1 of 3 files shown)


inulty-dfe:
Nice work!

Will these environments stack up over time or will they delete themselvers?

Nitemaeric:
> Nice work!
> 
> Will these environments stack up over time or will they delete themselvers?

Here's what Skylight has previously told me.

> they automatically go out of scope if there is no data reported to them for 45 days

Which is a bit longer than we need, but better than never disappearing!

tomas-stefano (approved):

PR #4698, opened 2024-11-18
Rake task and service for publishing a course

inulty-dfe:
## Context

We sometimes need to publish a course through the console. The main scenario is for when the course is in a previous recruitment cycle.
It's good to have a stable service to run this in a repeatable and safe way

Extract a Service `Courses::PublishService` from the methods in the `Course` model.

The service accepts a user and a course
The task accepts a `uuid` and user email

- I created a Rake tasks which accepts an email address so we don't commit an admins email address to the code.
- I chose the uuid because courses' uuid are unique across all RecruitmentCycles.

I've used the service in the `Publish::CoursesController#publish` action

### DIFF
--- a/guides/support_playbook.md
+++ b/guides/support_playbook.md
@@ -28,6 +28,14 @@ RecruitmentCycle.current.providers.find_by(provider_code: "1JJ").discard
 
 If the organisation has running courses, you will get a validation error. In this scenario you just need to make sure there aren't any users attached to it.
 
+## Publishing a course from an old recruitment cycle
+
+
+Use the Rake task for publishing courses [PublishCourse Rake Task](../lib/tasks/publish_course.rake) 
+
+Get the uuid of the course you want to publish and choose a user (your own user ideally) to publish.
+
+
 ## Unpublishing a published course
 
 Sometimes providers will accidentally publish a course and would like to unpublish it.


inulty-dfe:
[Register request to publish course from 2024 cycle](https://trello.com/c/wYjP8CsD/237-register-request-to-publish-course-from-2024-cycle)

app/services/courses/publish_service.rb
```diff
@@ -0,0 +1,27 @@
...
+    UnpublishableError = Class.new(StandardError)
+
+    class << self
+      def publish(uuid:, user:)
+        Rails.logger.tagged('Course::PublishService') do |l|
+          l.info "publishing course uuid: #{uuid}"
+        end
+        course = Course.find_by(uuid:)
+
+        raise UnpublishableError unless course.publishable?
```
inulty-dfe, line 14: I don't think it should raise an error. I want to use this in the controller so I don't want to catch this error.

Should this
1) return false if the course is unpulishable?
2) return the course and expect the caller to check if the course was published?

tomas-stefano: I'd go for the solution 2.

This is how I'd write this:

```ruby
module Courses
  class CoursePublisher
    def initialize(course:, user:)
      @course = course
      @user = user
    end

    def call
      return false unless @course.publishable?

      publish_course
      notify_course_published

      @course
    end

    private

    def publish_course
      Course.transaction do
        @course.publish_sites
        @course.publish_enrichment(@user)
        @course.application_status_open!
      end
    end

    def notify_course_published
      NotificationService::CoursePublished.call(course: @course)
    end
  end
end
```

And on controller:

```ruby
def publish
  course = Course.find_by(uuid: params[:uuid])
  
  if result = Courses::CoursePublisher.new(course: course, user: current_user).call
    redirect_to course_path(result), notice: 'Course published successfully!'
  else
    redirect_to courses_path, alert: 'Course could not be published.'
  end
end
```

I put the notification out of the transaction. Not sure if it should though. 

Also Publish has this pattern of `include ServicePattern`. Not sure if I like to suffix everything with `Service` though.

inulty-dfe: Nice, I'll incorporate this into the PR

tomas-stefano (approved):
🚀 

What do you think to add a documentation or small content on support playbook?

PR #4701, opened 2024-11-19
Normalize provider names

Nitemaeric:
## Context

Provider names should not not contain multiple spaces.

- Normalize `Provider#provider_name` to de-dup whitespaces and strip excess ones.

## Guidance to review

- Should we add a data migration to clean up existing records?

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -64,6 +64,8 @@ class Provider < ApplicationRecord
            primary_key: :provider_code,
            inverse_of: :accrediting_provider
 
+  normalizes :provider_name, with: ->(value) { value.gsub(/\s+/, ' ').strip }
+
   def accredited_providers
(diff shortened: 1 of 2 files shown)


tomas-stefano (approved):
We can fix old data from old cycles but there would be no real consequence. So I am not sure. 

Question is the call `.normalize_attribute(:provider_name)` before save will be called somewhere else?

Nitemaeric:
> We can fix old data from old cycles but there would be no real consequence. So I am not sure.
> 
> Question is the call `.normalize_attribute(:provider_name)` before save will be called somewhere else?

I've run the following script to update all provider names to ensure they are all normalised:

```ruby
Provider.find_each do |provider|
  provider.update_column :provider_name, provider.provider_name
end
```

This prevents `updated_at` values from being modified.

PR #4704, opened 2024-11-20
Moving POROs out of app/models into app/lib

gms-gs:
## Context

We want to be moving Plain Old Ruby Objects (POROs) out of app/models and into app/lib (or a similar directory like app/services, app/concerns, or app/utils) in our application as it is recommended for better organisation and separation of concerns

- Move the PORO's from models to the lib directory

## Guidance to review

- Play with the review app and see if anything breaks

### DIFF
--- a/spec/lib/user_notification_preferences_spec.rb
+++ b/spec/lib/user_notification_preferences_spec.rb
@@ -69,7 +69,7 @@ describe UserNotificationPreferences do
 
   describe '#update' do
     let(:accredited_provider_one) { create(:provider, :accredited_provider) }
-    let(:accredited_provider2) { create(:provider, :accredited_provider) }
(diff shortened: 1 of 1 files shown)


tomas-stefano (approved):

spec/lib/user_notification_preferences_spec.rb
```diff
@@ -242,7 +240,7 @@
           expect(user_notifications.count).to eq(3)
           expect(user_notifications.map(&:provider_code))
             .to contain_exactly(
-              accredited_provider1.provider_code,
+              accredited_provider_one.provider_code,
               accredited_provider2.provider_code,
               accredited_provider3.provider_code
```
Nitemaeric, line 245: Can we change these too, for consistency please?

PR #4706, opened 2024-11-21
Partially disable semantic logger in development

tomas-stefano:
## Context

Disable partially the semantic logger in development in order to be as close as we can from Rails default in development.

Following semantic logger documentation and comment on the issue: https://github.com/reidmorrison/rails_semantic_logger/issues/213

There is no way to disable entirely rails semantic logger without the `require: false` as the issue above described. I opt-in for the solution proposed in the issue above.

The image below are from the docs: https://logger.rocketjob.io/rails.html

Screenshot 2024-11-21 at 12 30 32

## Guidance to review

1. The review app logs in JSON?
2. If you pull locally the logs are near to the Rails default as we can?

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -82,4 +82,9 @@ Rails.application.configure do
   config.authentication_token = ENV.fetch('AUTHENTICATION_TOKEN', 'bats')
 
   config.x.read_only_database_url = "postgres://localhost/#{ENV.fetch('DB_DATABASE', 'manage_courses_backend_development')}"
+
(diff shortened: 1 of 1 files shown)


Nitemaeric (approved):
I've tried locally. It looks good.

Things I'd change:

- Remove `request_id` from `development`. As far as I can tell, we're typically just making requests 1-by-1 ourselves.
- Remove PID - I tried playing around with this one for a bit, but I can't seem to get it working without digging further.

PR #4708, opened 2024-11-22
Stop pre-loading accredited course list associations

Nitemaeric:
## Context

The accredited provider's download of all courses they ratify times out for some providers. We need to investigate the cause and how to fix it.

- Remove `.includes` call when loading the accredited course list. More information in commit message: [137a46e](https://github.com/DFE-Digital/publish-teacher-training/pull/4708/commits/137a46e4adb8a4724f23d41e8620c6795009197e)
- Change the program type re-assignment based on fee to run before a database write occurs. (Whilst testing the `spec/services/exports/accredited_course_list_spec.rb` spec, it would fail _`course.program_type == course.program_type`_. This is because one of the instances was taken from the database, one ran the after_save callback.

## Guidance to review

- Visit http://publish.localhost:3000/publish/organisations/C58/2025/training-providers and download the CSV on the side. It should take around 3-5 seconds.

### DIFF
--- a/app/controllers/publish/training_providers/course_exports_controller.rb
+++ b/app/controllers/publish/training_providers/course_exports_controller.rb
@@ -16,11 +16,11 @@ module Publish
       private
 
       def courses
-        @courses ||= provider.current_accredited_courses.includes(:enrichments, :sites, :site_statuses)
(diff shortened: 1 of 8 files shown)


tomas-stefano (approved):
🚀

PR #4710, opened 2024-11-25
Undiscard the course when publishing

inulty-dfe:
## Context

  Courses may be #published? and still not appear in the API. When we
  publish a course, we should expect it to appear in the API.
  When publishing a course we should undiscard the course if it has been
  discarded

PunlishService will undiscard a course in the process of publishing it.


inulty-dfe:
[Undiscard a course when publishing](https://trello.com/c/6zWW6ob1/265-undiscard-a-course-when-publishing)

PR #4712, opened 2024-11-26
Rename the accredited_provider_id column

inulty-dfe:
## Context

Provider model has a column named `accredited_provider_id`.
The value represents an identifier used by Funding and Market Regs to discriminate between providers with similar names and such.

`_id` suffixes in Ruby on Rails conventionally denote a foreign key columns on the model.

We're changing this column name to avoid obvious misunderstandings.

Rename `accredited_provider_id` column to `accredited_provider_number`


inulty-dfe:
[Tech debt: Rename the accredited_provider_id column](https://trello.com/c/ibQZpJtO/267-tech-debt-rename-the-accreditedproviderid-column)

Nitemaeric (approved):
Still seeing `accredited_provider_id` in the codebase was a surprise. 😅 Now `accredited_provider_id` is actually what we thought it would be - ish - an AR id!

All good! Nice one.

PR #4713, opened 2024-11-26
Remove canonical-rails gem and handroll our own

gms-gs:
## Context

When upgrading to Rails 8 the `canonical-rails` gem had issues with compatibility, after looking into the gem it looked pretty simple enough to handroll ourselves and also remove code. 

Ticket: https://trello.com/c/kPLDsLbC/269-remove-gem-canonical-rails-and-handroll-our-own-implementation

Instead of using  `<%= canonical_tag %>` we handroll our own implementation for these tags

## Guidance to review

Review app tags handrolled:
Screenshot 2024-11-26 at 13 09 58

Prod app tags using the gem:
Screenshot 2024-11-26 at 12 50 08

### DIFF
--- a/spec/features/canonical_tags_spec.rb
+++ b/spec/features/canonical_tags_spec.rb
@@ -2,22 +2,92 @@
 
 require 'rails_helper'
 
-feature 'Canonical tags' do
-  context 'when visiting Find pages', :with_find_constraint do
-    scenario 'Find pages contain canonical tags' do
-      visit '/'
+feature 'Canonical tags', :with_find_constraint do
+  before do
+    given_i_have_courses
+    and_i_visit_the_home_page
+  end
 
-      expect(page).to have_css("link[rel='canonical'][href='http://www.example.com/']", visible: :all)
-      expect(page).to have_css("meta[property='og:url'][content='http://www.example.com/']", visible: :all)
+  describe 'Canonical tags on Find pages' do
+    scenario 'without query params' do
+      then_the_page_contains_canonical_tags_with_no_query_params
+    end
+
+    scenario 'with query parameters' do
+      and_i_select_the_across_england_radio_button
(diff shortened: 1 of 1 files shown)


app/views/layouts/find_layout.html.erb
```diff
@@ -4,7 +4,10 @@
     <title><%= yield :page_title %> - <%= t("service_name.find") %> - GOV.UK</title>
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
-    <%= canonical_tag %>
+
+    <% url = URI(request.url.split("?").first) %>
```
tomas-stefano, line 8: I didn't test but if the url has anchors, will work this line?

tomas-stefano: I think we could use `request.base_url + request.path` instead. Let me try and see.

app/views/layouts/find_layout.html.erb
```diff
@@ -4,7 +4,10 @@
...
+    <%= tag.link(href: url.path.ends_with?("/") ? url.to_s : "#{url}/", rel: "canonical") %>
+    <%= tag.meta(property: "og:url", content: url.path.ends_with?("/") ? url.to_s : "#{url}/") %>
 
     <%= tag.meta(name: "viewport", content: "width=device-width, initial-scale=1") %>
     <%= tag.meta(property: "og:image", content: image_path("govuk-opengraph-image.png")) %>
```
tomas-stefano, line 13: Because of importance (if we remove canonical tags would not be good). Can we add a test with two different pages as scenarios? Wdyt?

app/views/layouts/find_layout.html.erb
tomas-stefano: Yes, using `request.base_url + request.path` might work

tomas-stefano (approved):

PR #4714, opened 2024-11-26
Bump to rails version 8.0.0

gms-gs:
## Context
Ticket: https://trello.com/c/uEmppsew/262-upgrade-to-rails-8
As part of keeping the application up to date we decided to upgrade to Rails 8.0.0

Ran `rails app:update` which created/edited a lot of files.

- Updates many auto-generated config files, some changes just shift the content around to make it easier for future upgrades.
- Other small changes to the code to work with rails 8.
- Fix  start_date for the EditCourseForm. We can't pass Structs when updating records in rails 8 any more, we need to pass nil not Struct(day: 1, month: nil, year: nil) when creating or updating records.

## Guidance to review

Play through the review app and see if things are working as expected, however we can _hopefully_ rely on all tests passing as a good indicator of a successful upgrade.

### DIFF
--- a/app/models/accrediting_provider_enrichment.rb
+++ b/app/models/accrediting_provider_enrichment.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 class AccreditingProviderEnrichment
-  include ActiveModel::Validations
   include ActiveModel::Model
 
   # Pascal cased as the original is stored like so.
@@ -9,10 +8,8 @@ class AccreditingProviderEnrichment
 
   validates :Description, words_count: { maximum: 100 }
 
-  def initialize(attrs)
-    attrs.each do |attr, value|
-      send("#{attr}=", value) unless attr == 'errors'
-    end
+  def initialize(attributes = {})
+    super(attributes.with_indifferent_access.slice('UcasProviderCode', 'Description'))
   end
 
   class ArraySerializer


Gemfile
```diff
@@ -166,7 +163,7 @@ group :development, :test do
   gem 'amazing_print'
 
   # Help eliminate N+1 queries
-  gem 'bullet'
+  gem 'bullet', require: false
```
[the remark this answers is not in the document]
inulty-dfe: [Bullet supports Rails 8 now](https://github.com/DFE-Digital/publish-teacher-training/pull/4670)

app/controllers/publish/providers_controller.rb
```diff
@@ -24,7 +24,7 @@ def suggest
                        .provider_search(params[:query])
                        .limit(10)
                        .map { |provider| { code: provider.provider_code, name: provider.provider_name } }
-      render json: @provider_list
+      render plain: @provider_list.to_json
```
Nitemaeric, line 27: Why is this change required?

bin/dev
Nitemaeric, line 1: This changes the behaviour. After some research, I've found that this file is meant to be defined per project.

As some of the team use `bin/dev` to run the project, changing this will affect their workflows.

You might want to reconsider this one.

bin/setup
```diff
@@ -15,20 +13,22 @@ FileUtils.chdir APP_ROOT do
   # Add necessary setup steps to this file.
 
   puts "== Installing dependencies =="
-  system("yarn install")
```
Nitemaeric, line 18: As we use `yarn` to build our assets, let's keep this line in.

config/initializers/new_framework_defaults_8_0.rb
Nitemaeric, line 1: None of these changes are breaking for our application, we should drop this file and set our application to load 8.0 defaults.

app/controllers/publish/providers_controller.rb
Nitemaeric: 🤔 Do you know why it was doing that?

Does the recent update to AMS fix it?

inulty-dfe (commented):
Looks ok, I've clicked about in the review app. 

Are we not including the options in the environment files that are new in Rails 8?

bin/thrust
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
Nitemaeric: It _might_ be useful. It's definitely not being used here though.

I think it prefixes the `rails server` command in Rails 8's default `Dockerfile`.

cc: @gms-gs

Nitemaeric (approved):

PR #4723, opened 2024-11-28
Enable ActiveJob in skylight

inulty-dfe:
## Context

Sidekiq needs to be explicitly enabled in Skylight.

  https://www.skylight.io/support/background-jobs#using-skylightyml
  https://www.skylight.io/support/background-jobs#enabling-frameworks

Enable our Application Performance Monitoring tool to collect data on our background jobs

## Guidance to review

[Link to the review process](https://www.skylight.io/app/applications/P29Oy7glIJ6w/recent/5m/endpoints) (might disappear by the time you see this)

image

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -56,6 +56,7 @@ module ManageCoursesBackend
     config.skylight.environments += Settings.skylight.enable ? [Rails.env] : []
     config.skylight.logger = SemanticLogger[Skylight]
     config.skylight.log_level = :fatal
+    config.skylight.probes += ['active_job']
     config.skylight.native_log_level = :fatal
 
     config.view_component.preview_paths = [Rails.root.join('spec/components')]


config/skylight.yml
```diff
@@ -7,3 +7,4 @@ ignored_endpoints:
   - HeartbeatController#sha
 deploy:
   git_sha: <%= ENV["COMMIT_SHA"] %>
+enable_sidekiq: true
```
Nitemaeric, line 10: Let's drop this line:

> If you use Sidekiq with ActiveJob, you can enable either the Sidekiq config or the ActiveJob probe (read on to the next section), or both (it’s not necessary to do both, but it won’t hurt).

I was testing if I could grab the review PR number [here](https://github.com/DFE-Digital/publish-teacher-training/pull/4697), so we could separate the skylight environments per review app. Otherwise all review apps would feed into the same skylight environment.

inulty-dfe: I'm not 100% sure every job uses ActiveJob. In Apply, some jobs are ActiveJob and some are Sidekiq::Workers. I enabled both to make sure both work. I'm pretty sure all our jobs will be Sidekiq but not sure about DfE::Analytics

Nitemaeric: `DfE::Analytics` uses `ActiveJob`: https://github.com/DFE-Digital/dfe-analytics/blob/main/lib/dfe/analytics/analytics_job.rb

inulty-dfe:
I'm not seeing the worker process appear in the Skylight UI. Anyone got a clue why?

Nitemaeric:
> I'm not seeing the worker process appear in the Skylight UI. Anyone got a clue why?

Have any jobs been run? I don't think we have any worker jobs that fire.

inulty-dfe:
> > I'm not seeing the worker process appear in the Skylight UI. Anyone got a clue why?
> 
> Have any jobs been run? I don't think we have any worker jobs that fire.

Yeah that was it.

https://www.skylight.io/app/applications/P29Oy7glIJ6w/recent/5m/endpoints

Nitemaeric (approved):

PR #4724, opened 2024-11-29
Clear the accredited_provider_number when removing provider accreditation

inulty-dfe:
## Context

  Only Accredited providers should have an `accredited_provider_number`

Add simple UpdateProviderForm
It accepts attributes and uses `ActiveRecord::Base#assign_attributes` to assign the attributes. Not much needs to be tested here.
It removes the `accredited_provider_number` if the attribute assignment changes `accrediting_provider` to `'not_an_accredited_provider'`

## Guidance to review

The main point in the video below is that the Accredited provider number is not there when we change the provider back to an Accredited provider.

[Screencast from 29-11-24 15:09:10.webm](https://github.com/user-attachments/assets/0fcfd11e-f4b2-4faa-a8f1-b4f47061c121)

Pick it apart and recommend changes

### DIFF
--- a/app/components/providers/provider_list/view.html.erb
+++ b/app/components/providers/provider_list/view.html.erb
@@ -24,7 +24,7 @@
       end
 
       if @provider.accredited_provider?
-        component.with_row do |row|
+        component.with_row(html_attributes: { data: { qa: 'ap-number-row' } }) do |row|
           row.with_key { "Accredited provider number" }
           row.with_value { value_provided?(@provider.accredited_provider_number.to_s) }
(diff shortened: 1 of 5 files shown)


inulty-dfe:
[Clear the accredited_provider_number when removing accredited status from a provider](https://trello.com/c/O0FzXchN/271-clear-the-accreditedprovidernumber-when-removing-accredited-status-from-a-provider)

Nitemaeric:
This works fine, as of now. 1 thing to keep in mind though - I want us to move towards a ledger approach of accreditation. This way, we have a historical record of accreditation.

Nitemaeric (approved):

PR #4734, opened 2024-12-03
V2 results - Add all courses, pagination and card title

tomas-stefano:
## Context

Start to populate the `/v2/results`

This PR adds all courses and paginate all courses. And add only the course title (provider name, course name and code) to the summary card (for now).

screencapture-find-localhost-v2-results-2024-12-03-10_04_26

## Guidance to review

1. Access the review app /v2/results 
2. See all courses paginated

## What comes next?

Next I will start to implement the filters, one by one and later add the no course scenario along side with the content of amount of courses listed.

### DIFF
--- a/app/services/courses_query.rb
+++ b/app/services/courses_query.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class CoursesQuery
+  def self.call(params = {})
+    new(params:).call
+  end
+
+  attr_reader :scope
+
+  def initialize(params:)
+    @params = params
+    @scope = RecruitmentCycle
+             .current
+             .courses
+             .joins(:site_statuses)
+             .where(
+               site_statuses: {
+                 status: SiteStatus.statuses[:running],
+                 publish: SiteStatus.publishes[:published]
+               }
+             )
+  end
+
+  def call
+    @scope.distinct
+  end
+end
--- a/app/views/find/v2/results/index.html.erb
+++ b/app/views/find/v2/results/index.html.erb
@@ -1,0 +1,34 @@
+<div class="app-filter-layout">
+  <div class="app-filter-layout__filter">
+    <div class="app-filter__header">
+      <h2 class="govuk-heading-m">Filters</h2>
+
+      <div class="app-filter__content">
+      </div>
+    </div>
+  </div>
+  <div class="app-filter-layout__content">
+    <ul class="app-search-results">
+      <% @results.each do |course| %>
+        <li class="app-search-results__item" data-qa="course">
+          <%= govuk_summary_card(
+            classes: ["course-summary-card"],
+            title: govuk_link_to(
+              find_course_path(
+                provider_code: course.provider_code,
+                course_code: course.course_code
+              ),
+              class: 'govuk-link govuk-!-font-size-24',
+              data: { qa: 'course__link' }
+            ) do
+              content_tag(:span, course.provider.provider_name, class: 'app-search-result__provider-name', data: { qa: 'course__provider_name' }) +
+              content_tag(:span, course.name_and_code, class: 'app-search-result__course-name', data: { qa: 'course__name' })
+            end
+          ) %>
+        </li>
+      <% end %>
+    </ul>
+
+    <%= govuk_pagination(pagy: @pagy) %>
+  </div>
+</div>
--- a/app/views/find/v2/results/index.html.erb
+++ b/app/views/find/v2/results/index.html.erb
@@ -18,11 +18,11 @@
                 provider_code: course.provider_code,
                 course_code: course.course_code
               ),
-              class: 'govuk-link govuk-!-font-size-24',
-              data: { qa: 'course__link' }
+              class: "govuk-link govuk-!-font-size-24",
+              data: { qa: "course__link" }
             ) do
-              content_tag(:span, course.provider.provider_name, class: 'app-search-result__provider-name', data: { qa: 'course__provider_name' }) +
-              content_tag(:span, course.name_and_code, class: 'app-search-result__course-name', data: { qa: 'course__name' })
+              content_tag(:span, course.provider.provider_name, class: "app-search-result__provider-name", data: { qa: "course__provider_name" }) +
+              content_tag(:span, course.name_and_code, class: "app-search-result__course-name", data: { qa: "course__name" })
             end
           ) %>
         </li>
--- a/spec/services/courses_query_spec.rb
+++ b/spec/services/courses_query_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CoursesQuery do
+  let!(:findable_course) do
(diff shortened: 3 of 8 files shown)


app/views/find/v2/results/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+                course_code: course.course_code
+              ),
+              class: 'govuk-link govuk-!-font-size-24',
+              data: { qa: 'course__link' }
+            ) do
+              content_tag(:span, course.provider.provider_name, class: 'app-search-result__provider-name', data: { qa: 'course__provider_name' }) +
+              content_tag(:span, course.name_and_code, class: 'app-search-result__course-name', data: { qa: 'course__name' })
+            end
+          ) %>
+        </li>
```
tomas-stefano, line 28: For now I decided not to create a component just yet.

Until the need arrives ...

app/views/find/v2/results/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+            classes: ["course-summary-card"],
+            title: govuk_link_to(
+              find_course_path(
+                provider_code: course.provider_code,
+                course_code: course.course_code
+              ),
+              class: 'govuk-link govuk-!-font-size-24',
+              data: { qa: 'course__link' }
+            ) do
+              content_tag(:span, course.provider.provider_name, class: 'app-search-result__provider-name', data: { qa: 'course__provider_name' }) +
```
tomas-stefano, line 24: On the original search results the code is calling smart_quotes method from publish helper. I added as checklist in the outcome card:

Screenshot 2024-12-03 at 10 33 12

app/views/find/v2/results/index.html.erb
```diff
@@ -0,0 +1,34 @@
+<div class="app-filter-layout">
+  <div class="app-filter-layout__filter">
+    <div class="app-filter__header">
+      <h2 class="govuk-heading-m">Filters</h2>
+
+      <div class="app-filter__content">
+      </div>
+    </div>
+  </div>
+  <div class="app-filter-layout__content">
```
tomas-stefano, line 10: The handle of no results will arrive in another card.

app/views/find/v2/results/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+    <ul class="app-search-results">
+      <% @results.each do |course| %>
+        <li class="app-search-results__item" data-qa="course">
+          <%= govuk_summary_card(
+            classes: ["course-summary-card"],
+            title: govuk_link_to(
+              find_course_path(
+                provider_code: course.provider_code,
+                course_code: course.course_code
```
tomas-stefano, line 19: On the original implementation we are passing the search params to the show page. I added on the checklist of the outcome card and decided not to pass yet. Until we have the filters implemented at least.

app/services/courses_query.rb
```diff
@@ -0,0 +1,27 @@
...
+    @scope = RecruitmentCycle
+             .current
+             .courses
+             .joins(:site_statuses)
+             .where(
+               site_statuses: {
+                 status: SiteStatus.statuses[:running],
+                 publish: SiteStatus.publishes[:published]
+               }
+             )
```
tomas-stefano, line 21: One of the rules @Nitemaeric and I talked was to decrease the usage of scopes so moved the where here for the `findable` scope. This is easier to read and reduce the too much indirection when reading code.

tomas-stefano:
[Foundation for the pre-filter work - Add Pagination](https://trello.com/c/XOIfAURZ/258-foundation-for-the-pre-filter-work-add-pagination)

tomas-stefano:
[Foundation for the pre-filter work - Individual course card part 1 (only provider/course name and link)](https://trello.com/c/wcXpxOLx/259-foundation-for-the-pre-filter-work-individual-course-card-part-1-only-provider-course-name-and-link)

spec/services/courses_query_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+RSpec.describe CoursesQuery do
+  let!(:findable_course) do
+    create(:course, :with_full_time_sites)
+  end
+  let!(:another_course) do
+    create(:course, :with_full_time_sites)
+  end
+  let!(:non_findable_course) do
+    create(:course)
+  end
```
tomas-stefano, line 14: I chosen consciously to reach the DB instead of double the scope.

I chosen the trade-off of more slow tests over more stability and guarantee progression and behaviour of the course search.

I did via `let!` (let me know if you prefer other way of seeding different courses.

This will grow with the around 15 filters that will be added!

app/services/courses_query.rb
inulty-dfe: This is building on tech debt around the "running" of `SiteStatus`. We are supposed to be moving away from running sites and toward open / closed courses. The class should just be called `CourseSite` like the table now.

Where is `published_on_ucas` defined? I can't find it anywhere :thinking:

tomas-stefano: > This is building on tech debt around the "running" of SiteStatus. We are supposed to be moving away from running sites and toward open / closed courses.

Interesting. So the current scope `Course.findable` are using like that. So should we move away from findable scope and make open / closed? We do have the filter for open and closed applications to be added in a new ticket.

> Where is published_on_ucas defined? I can't find it anywhere 🤔

It is an enumerator.
```ruby
# site status
  31   enum :publish, {
  32     published: 'Y',
  33     unpublished: 'N'
  34   }, suffix: :on_ucas
```

tomas-stefano: Iain and I discussed this, I will create a card specifically to improve to have a course status. 

Would be awesome if we could do:

```ruby
     @scope = RecruitmentCycle
              .current
             .courses
             .where(status: 'published')
```

But we can't at the moment.

Whenever we change the course enrichment  to "publish" a course (or changes to withdrawn) we fill the status of "published" or "withdrawn".

inulty-dfe: So as you mentioned to me in a huddle, we need to be able to tell if the course is published or not.
Ideally we'd have a column 'status' or similar. Instead we have CourseEnrichment where the `status` is stored inside a jsonb object.

Happy to approve this now that we've discussed how we can improve it in future.

inulty-dfe (approved):
Great that we've begun to discuss some of the tech debt that this PR has highlighted.

spec/services/courses_query_spec.rb
Nitemaeric: I've found that the cleanest way to build these specs is as follows:

```ruby
RSpec.describe CoursesQuery do
  describe ".call" do
    subject(:results) { described_class.call(query) }

    context "when filtering by subject" do
      let!(:maths_subject) { ... }
      let!(:english_subject) { ... }

      let(:query) do
        { subject_ids: [maths_subject.id] }
      end

      it "returns only the courses with matching subject" do
        expect(call).to contain_exactly(maths_subject)
      end
    end
  end
end
```

Group the setup and filters applied next to each other.

Nitemaeric (approved):

PR #4736, opened 2024-12-03
Update Node to latest LTS: v22.x

Nitemaeric:


Nitemaeric:
We'll revisit this later. We should update our apps to use LTS node, but it is not a current priority.

PR #4738, opened 2024-12-03
Validate `accredited_provider_number` on update too

inulty-dfe:
## Context

The presence of an `accredited_provider_number` is validated when a provider is being onboarded but not when an existing provider is promoted from a Lead partner to an Accredited Provider

- Accredited provider cannot be a school
- Accredited provider number must be present when saving an Accredited provider.
- University accredited providers number format starts with a 1
- SCITT accredited providers number format starts with a 5
- Add specific Accredited provider number error messages for SCITT and University
- Update `scitt` and `university` provider factory traits - they are not automatically accredited providers anymore

## Guidance to review

Any tips on the form would be useful. How do we want to use forms going forward?

[Screencast from 06-12-24 15:20:51.webm](https://github.com/user-attachments/assets/ba73cd2a-144a-46ba-8bcd-8c25f7042c55)

### DIFF
--- a/app/forms/support/update_provider_form.rb
+++ b/app/forms/support/update_provider_form.rb
@@ -2,21 +2,31 @@
 
 module Support
   class UpdateProviderForm
-    attr_reader :provider
+    delegate :valid?, to: :provider
 
     def initialize(provider, attributes:)
       @provider = provider
       @attributes = attributes
+      assign_attributes
     end
 
     def save
-      @provider.assign_attributes(@attributes)
+      return false unless provider.valid?
+
+      provider.save!
+    end
+
+    private
+
+    attr_reader :provider, :attributes
+
+    def assign_attributes
+      @provider.assign_attributes(attributes)
       remove_accredited_provider_number
-      @provider.save
     end
 
     def remove_accredited_provider_number
-      return unless @provider.accrediting_provider_change&.[](1) == 'not_an_accredited_provider'
+      return unless @provider.accrediting_provider_changed?(to: 'not_an_accredited_provider')
 
       @provider.accredited_provider_number = nil
     end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -170,6 +170,13 @@ class Provider < ApplicationRecord
 
   validates :accredited_provider_number, accredited_provider_number_format: true, if: :accredited_provider?
 
+  validate :accredited_provider_type
+  def accredited_provider_type
+    return unless accredited_provider?
+
+    errors.add(:accrediting_provider, :invalid_provider_type) unless provider_type.to_s.in?(%w[scitt university])
+  end
+
   acts_as_mappable lat_column_name: :latitude, lng_column_name: :longitude
 
   before_discard do
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -168,7 +168,7 @@ class Provider < ApplicationRecord
 
   validate :add_enrichment_errors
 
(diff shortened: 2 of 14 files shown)


inulty-dfe:
[Bug: Validate presence of Accredited Provider Number on Provider Form](https://trello.com/c/6FV5RxM5/273-bug-validate-presence-of-accredited-provider-number-on-provider-form)

app/forms/support/update_provider_form.rb
```diff
@@ -2,21 +2,31 @@
...
+
+    def assign_attributes
+      @provider.assign_attributes(attributes)
       remove_accredited_provider_number
-      @provider.save
     end
 
     def remove_accredited_provider_number
-      return unless @provider.accrediting_provider_change&.[](1) == 'not_an_accredited_provider'
+      return unless @provider.accrediting_provider_changed?(to: 'not_an_accredited_provider')
```
Nitemaeric, line 29: <3

tomas-stefano (approved):

app/forms/support/update_provider_form.rb
```diff
@@ -2,21 +2,31 @@
...
       @attributes = attributes
+      assign_attributes
     end
 
     def save
-      @provider.assign_attributes(@attributes)
+      return false unless provider.valid?
+
+      provider.save!
+    end
```
tomas-stefano, line 17: In related to the PR description:

> Any tips on the form would be useful. How do we want to use forms going forward?

I'd not make the save here. On this example is simple that's why I approved but if becomes more complicated this class can be convulated (and then move to a service object to handle the transaction)

inulty-dfe: How do you propose the save should be made?

tomas-stefano: > How do you propose the save should be made?

Sure. I would create a Policy class and a service to handle the transaction (like in Apply we have in the provider interface).

```ruby
class AccreditingProviderPolicy
    def removed_accreditation?(provider)
      provider.accrediting_provider_changed?(to: 'not_an_accredited_provider')
    end
  end

  class ProviderPersistenceService
    def initialize(provider, attributes, accrediting_policy: AccreditingProviderPolicy.new)
      @provider = provider
      @attributes = attributes
      @accrediting_policy = accrediting_policy
    end

    def save
        provider.assign_attributes(attributes)
        apply_accrediting_policy
        provider.save!
      end
    rescue ActiveRecord::RecordInvalid => e
      handle_error(e)
      false
    end

    private

    attr_reader :provider, :attributes, :accrediting_policy

    def apply_accrediting_policy
      if accrediting_policy.removed_accreditation?(provider)
        provider.accredited_provider_number = nil
      end
    end
  end
end
```

Then you could have the call on the form or in the controller (1 or 2 below).

1. I do prefer in the controller but I can see the argument to be in the form:

```ruby
    def save
      return false unless valid?

      ProviderPersistenceService.new(provider, attributes).save
    end
```

2. In the controller:

```ruby
def update
    provider = Provider.find(params[:id])
    form = Support::UpdateProviderForm.new(provider, attributes: provider_params)

    if form.valid? && Support::ProviderPersistenceService.new(form.provider, provider_params).save
      redirect_to provider_path(provider), notice: 'Provider updated successfully.'
    else
      flash.now[:alert] = form.valid? ? 'Failed to save provider.' : 'Invalid provider data.'
      render :edit
    end
  end
```

tomas-stefano: I like Policy objects and I think we could have more on Publish by the way. The `AccreditingProviderPolicy` can be even receive the provider on the initialize.

PR #4750, opened 2024-12-05
Copy courses between providers in Support console

inulty-dfe:
## Context

Support are often asked to copy courses from one provider to another. This PR provides the ability for users to copy courses via the Support console.

#### Considerations

1. Copy all courses
2. Copy schools and study sites or not

- What if the target provider already has copied the courses?
  - The courses are not recopied
- What if the target provider already has copied the courses and deleted them?
  - The courses are not recopied
- What if there are 1000 courses
  - It will take a while to copy all the courses across.
- Could this be a background process?
  - Lets see if that is necessary

[Screencast from 09-12-24 15:27:18.webm](https://github.com/user-attachments/assets/065d4116-2ce3-45f0-8a0c-def86ad6bae4)

### DIFF
--- a/app/controllers/support/providers/copy_courses_controller.rb
+++ b/app/controllers/support/providers/copy_courses_controller.rb
@@ -4,25 +4,31 @@ module Support
   module Providers
     class CopyCoursesController < SupportController
       before_action :recruitment_cycle
+
       def new
-        @provider = Provider.find(params[:provider_id])
+        @target_provider = Provider.find(params[:provider_id])
+        @copy_courses_form = CopyCoursesForm.new(@target_provider)
       end
 
       def create
-        @provider = Provider.find(params[:provider_id])
-        @from_provider = recruitment_cycle.providers.find_by(provider_code: params[:course][:autocompleted_provider_code])
+        @target_provider = Provider.find(params[:provider_id])
+        @copy_courses_form = CopyCoursesForm.new(@target_provider, recruitment_cycle.providers.find_by(provider_code: params[:course][:autocompleted_provider_code]))
 
-        sites_copy_to_course = params[:schools] ? Sites::CopyToCourseService : ->(*) {}
+        if @copy_courses_form.valid?
+          sites_copy_to_course = params[:schools] ? Sites::CopyToCourseService : ->(*) {}
 
-        copier = ::Courses::CopyToProviderService.new(sites_copy_to_course:, enrichments_copy_to_course: Enrichments::CopyToCourseService.new, force: true)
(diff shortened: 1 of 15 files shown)


inulty-dfe:
[Copy courses from one provider to another in Support Console](https://trello.com/c/4sWKKFND/295-copy-courses-from-one-provider-to-another-in-support-console)

app/controllers/support/providers/copy_courses_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+        @target_provider = Provider.find(params[:provider_id])
+        @copy_courses_form = CopyCoursesForm.new(@target_provider, recruitment_cycle.providers.find_by(provider_code: params[:course][:autocompleted_provider_code]))
+
+        if @copy_courses_form.valid?
+          sites_copy_to_course = params[:schools] ? Sites::CopyToCourseService : ->(*) {}
+
+          copier = ::Courses::CopyToProviderService.new(sites_copy_to_course:, enrichments_copy_to_course: Enrichments::CopyToCourseService.new, force: true)
+
+          Provider.transaction do
+            @copy_courses_form.provider.courses.map do |course|
```
tomas-stefano, line 23: On the PR description it mentions "Copy all courses or some courses" but here is all courses. Can we remove the "some courses" from description just to avoid confusion?

I assume the copying specific courses would be done on a later interaction when we will have a support ticket?

app/controllers/support/providers/copy_courses_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+              copier.execute(course:, new_provider: @copy_courses_form.target_provider)
+            end
+          end
+
+          redirect_to support_recruitment_cycle_provider_path(recruitment_cycle.year, @copy_courses_form.target_provider.id)
```
tomas-stefano, line 28: Is there any indication everything went sucessfully? Should we add one?

tomas-stefano (approved):
Minor questions but tested and it worked ✌️

app/controllers/support/providers/copy_courses_controller.rb
inulty-dfe: Yeah I've not implemented "some courses" because unwanted courses can be deleted

app/controllers/support/providers/copy_courses_controller.rb
inulty-dfe: [Screencast from 10-12-24 10:47:38.webm](https://github.com/user-attachments/assets/f221dfa5-6e54-414a-abd0-99a61f84cb16)

inulty-dfe: I'll change it to redirect to the courses tab too

PR #4751, opened 2024-12-05
Foundation work for the pre filter work - add visa sponsorship filter

tomas-stefano:
## Context

On the new search results page we want to add filters.

This card is about adding the visa sponsorship filter to the new search results page.

When user loads this page without any visa sponsorship in the params, list courses that sponsor visa or not.

When user checks this filter it should return only courses that offers visa sponsorship.

screencapture-find-localhost-v2-results-2024-12-05-18_45_17

## Log screeenshot

I added the `CoursesQuery::Logger` class to help facilitate in a quick manner which filters the code applied to and also to log the query done by the `CoursesQuery` class. The query si already logged but in the midst of the many logs the page renders to which becomes difficult to spot quickly where is. This class only logs in development.

These are the screenshots:

**Without explain**

Screenshot 2024-12-09 at 11 30 29

**With explain**
Screenshot 2024-12-09 at 11 27 23

The explain is very big because we have a performance problem on the live search and on this as well (the Course#after_initialize is the culprit).

## Guidance to review

1. Visit /v2/results
2. Any bugs can you identify?
3. Are you happy with the rspec matcher created?
4. Are you happy with courses query logger created?


app/services/courses_query_logger.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class CoursesQueryLogger
```
tomas-stefano, line 3: I created this class because of my own necessity.

To see which filters is applied, inspect better in the log the whole query and also to explain to understand the query performance.

This is already helping me to see if the code applied the filter or not, see the query in the midst of the log chaos and also explain and understand a course problem with the courses <-> course_subjects <-> subjects (because of the after_initialize block in the course model).

app/views/find/v2/results/index.html.erb
```diff
@@ -1,12 +1,22 @@
...
+        <div class="app-filter__header">
+          <h2 class="govuk-heading-m">Filters</h2>
+        </div>
+
+        <div class="app-filter__content">
+          <%= form.submit "Apply filters", name: nil, class: "govuk-button", data: { qa: "apply-filters" } %>
+
+          <%= form.govuk_check_boxes_fieldset :can_sponsor_visa, multiple: false, legend: { text: t("helpers.legend.course_search_form.can_sponsor_visa_html"), size: "s" }, class: "app-filter__group govuk-checkboxes--small", data: { qa: "filters__visa" }, hidden: false do %>
+            <%= form.govuk_check_box :can_sponsor_visa, "true", name: "can_sponsor_visa", label: { text: t("helpers.label.course_search_form.can_sponsor_visa"), size: "s" } %>
+          <% end %>
```
tomas-stefano, line 14: For now I decided not to move this filter to a partial.

tomas-stefano:
[Foundation for the pre-filter work - Add visa sponsorship filter](https://trello.com/c/jfANwLC3/284-foundation-for-the-pre-filter-work-add-visa-sponsorship-filter)

spec/support/matchers/match_collection.rb
```diff
@@ -0,0 +1,55 @@
...
+    actual_ids = actual.pluck(:id)
+
+    case type
+    when :missing
+      expected.reject { |e| actual_ids.include?(e[:id]) }
+    when :extra
+      actual.reject { |a| expected_ids.include?(a[:id]) }
+    end
+  end
+end
```
tomas-stefano, line 55: I created this matcher because it was impossible to know what is wrong on the spec:

```sh
1) CoursesQuery.call when filter for visa sponsorship returns courses that sponsor visa
     Failure/Error: expect(results).to contain_exactly(course_that_sponsor_visa, another_course_that_sponsor_visa)

                 Expected [#<Course:0x0000000135c1aa08 @_already_called=nil, @_committed_already_called=nil, @_defer_touch_attrs=nil, @_new_record_before_last_commit=nil, @_skip_dirty_tracking=nil, @_start_transaction_state=nil, @_touch_attr_names=nil, @_touch_record=nil, @_trigger_destroy_callback=nil, @_trigger_update_callback=nil, @association_cache={ course_subjects: #<ActiveRecord::Associations::HasManyAssociation:0x0000000143b34

1000000 lines long.
```

tomas-stefano: Now is much better:

```
Expected collection contained:
         [{"can_sponsor_skilled_worker_visa"=>true,
         "can_sponsor_student_visa"=>false,
         :id=>4},
        {"can_sponsor_skilled_worker_visa"=>false,
         "can_sponsor_student_visa"=>true,
         :id=>5}]

       Actual collection contained:
         [{"can_sponsor_skilled_worker_visa"=>true,
         "can_sponsor_student_visa"=>false,
         :id=>4},
        {"can_sponsor_skilled_worker_visa"=>false,
         "can_sponsor_student_visa"=>true,
         :id=>5},
        {"can_sponsor_skilled_worker_visa"=>false,
         "can_sponsor_student_visa"=>false,
         :id=>6}]


       The missing elements were:
         []


       The extra elements were:
         [{"can_sponsor_skilled_worker_visa"=>false,
         "can_sponsor_student_visa"=>false,
         :id=>6}]
```

config/locales/find.yml
```diff
@@ -302,6 +302,13 @@ en:
         label:
           one: '%{count} mile'
           other: '%{count} miles'
+  helpers:
+    legend:
+      course_search_form:
+        can_sponsor_visa_html: Visa sponsorship<span class="govuk-visually-hidden"> filter</span>
```
inulty-dfe, line 308: TIL https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

app/services/courses_query.rb
```diff
@@ -22,6 +23,37 @@ def initialize(params:)
...
+
+    log_query_info
+
+    @scope
+  end
+
+  def visa_sponsorship_filter
+    return @scope if params[:can_sponsor_visa].blank?
+
+    filter_applied(name: :can_sponsor_visa)
```
inulty-dfe, line 37: `add_filter`?

app/services/courses_query.rb
```diff
@@ -5,10 +5,11 @@ def self.call(params = {})
     new(params:).call
   end
 
-  attr_reader :scope
+  attr_reader :scope, :params
 
   def initialize(params:)
     @params = params
+    @applied_filters = []
```
inulty-dfe, line 12: Is there such thing as "unapplied filters"?
If not, I'd just call this `filters`

inulty-dfe: Also, are we mixing domains here?
Filter sounds like a business domain term and queries are data layer.

This feels akin to using view logic in models

config/locales/find.yml
```diff
@@ -302,6 +302,13 @@ en:
         label:
           one: '%{count} mile'
           other: '%{count} miles'
+  helpers:
```
inulty-dfe, line 305: Why are we putting these "translations" under `'helpers.legend'`?

spec/support/matchers/match_collection.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+RSpec::Matchers.define :match_collection do |expected_collection, attribute_names: []|
```
inulty-dfe, line 3: Whats going on here? This isn't referenced in the code

inulty-dfe (commented):
Few suggestions around domain layers and questions.

config/locales/find.yml
tomas-stefano: To follow the same pattern as https://govuk-form-builder.netlify.app/building-blocks/localisation/

The reason I couldn't use the above is because govuk-form-builder don't translate if the `scope: ''` on the `form_with`.

spec/support/matchers/match_collection.rb
tomas-stefano: It is. More above: https://github.com/DFE-Digital/publish-teacher-training/pull/4751/files#diff-feadbcd7cd23dccc9c8311aea6c91bca39ac979ae097ec092b74e48b1168ec75R38

app/services/courses_query.rb
tomas-stefano: I could called applied scopes? Or applied parameters.

The only reason I am using to be more quick during development to know which things this query is searching - because the DB query might become too big with 16 filters from the view to understand.

app/forms/course_search_form.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class CourseSearchForm
```
Nitemaeric, line 3: I'm not sure I see the value in this form object.

It looks to filter down the object keys from query params, but that's all. Argument/parameter validation feels to me that it should be the responsibility of the query object itself?

app/services/courses_query_logger.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def call
+    log_filters
+    log_query
+  end
+
+  private
+
+  def log_filters
+    return unless Rails.env.local?
```
Nitemaeric, line 20: This and line 35 could be squashed into a single call in `#call` itself.

app/services/courses_query.rb
```diff
@@ -5,10 +5,11 @@ def self.call(params = {})
     new(params:).call
   end
 
-  attr_reader :scope
+  attr_reader :scope, :params
 
   def initialize(params:)
```
Nitemaeric, line 10: I've just realised, the `.call` argument signature differs from the `#initialize` signature.

I think there's more value in consistency.

app/services/courses_query.rb
```diff
@@ -22,6 +23,37 @@ def initialize(params:)
...
+          can_sponsor_skilled_worker_visa: true
+        )
+      )
+  end
+
+  private
+
+  def filter_applied(name:)
+    @applied_filters << { name:, value: params[name] }
+  end
```
Nitemaeric, line 54: Instead of creating an abstraction here, this could be the following instead:

```ruby
class QueryObject
  def initialize
    self.applied_filters = {}
  end

  def some_filter
    applied_filters[:some_filter] = params[:some_filter]

    @scope
  end

  private

  attr_accessor :applied_filters
end
```

or if you think a filter can be applied multiple times?

```ruby
class QueryObject
  def initialize
    self.applied_filters = []
  end

  def some_filter
    applied_filters << { name: :some_filter, value: params[:some_filter] }

    @scope
  end

  private

  attr_accessor :applied_filters
end
```

Should it?

app/services/courses_query_logger.rb
Nitemaeric: This class is coupled to the `CoursesQuery`, yet it is distinctly detached. Does it make sense to make this a subclass of `CoursesQuery` itself? `CoursesQuery::Logger` possibly?

app/services/courses_query_logger.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+class CoursesQueryLogger
+  def initialize(applied_filters, scope, explain: false)
+    @applied_filters = applied_filters
+    @scope = scope
+    @log_tag = self.class.name
+    @logger = Rails.logger
```
Nitemaeric, line 8: These 2 variables look like they don't need to be _variables_.

```ruby
def log_tag
  self.class.name
end

def logger
  Rails.logger
end
```

app/services/courses_query_logger.rb
Nitemaeric: Additionally, this does feel like it is a bit preemptive.

tomas-stefano: > You might want to show a screenshot of what the log output looks like in the PR description.

Done.

> Additionally, this does feel like it is a bit preemptive.

Maybe, although it already proves its usefulness so I plan to maintain. The advantage is this can be easily removed if don't prove to be useful anymore.

> This class is coupled to the CoursesQuery, yet it is distinctly detached. Does it make sense to make this a subclass of CoursesQuery itself? CoursesQuery::Logger possibly?

Done!

app/services/courses_query_logger.rb
tomas-stefano: Done

app/services/courses_query_logger.rb
tomas-stefano: Done

app/services/courses_query.rb
tomas-stefano: I didn't understood.

```ruby
 3 class CoursesQuery
   4   def self.call(params = {})
   5     new(params:).call
   6   end
   9
  10   def initialize(params:)
```

They have the same signature. Do you meant something else?

app/services/courses_query.rb
tomas-stefano: I removed this as per suggestion https://github.com/DFE-Digital/publish-teacher-training/pull/4751/files#r1875757673

app/services/courses_query.rb
tomas-stefano: The first one as Hash looks good to me. Done.

app/services/courses_query.rb
tomas-stefano: I renamed to `applied_scopes`.

app/forms/course_search_form.rb
tomas-stefano: I am not sure either the transformation from raw parameters into ruby objects (booleans, arrays, strings, value conversions, etc) should be in the query class. 

I know you wanna avoid indirection but then the Query class would have too much responsibility.

🤔 🤔 🤔 

In my opinion, the best we could do if we remove the form object and make the code clean:

```
class CoursesQuery
  class Params
    include ActiveModel::Attributes

    attribute :can_sponsor_visa, :boolean
  end

  def self.call(params = {})
    new(params:).call
  end

  attr_reader :scope, :params

  def initialize(params:)
    @params = CoursesQuery::Params.new(params)
```

Then in the future we could do something like:

```ruby
class Params
    include ActiveModel::Model
    include ActiveModel::Attributes

    attribute :can_sponsor_visa, :boolean
    attribute :age_group, :string
    attribute :applications_open, :boolean
    attribute :subjects, :array, default: []

    # Normalization and legacy param handling
    def initialize(params = {})
      super(normalize_params(params))
    end

    private

    def normalize_params(params)
      params.dup.tap do |p|
        p[:can_sponsor_visa] = p.delete(:visa_status) if p[:visa_status].present?
        p[:subjects] = Array(p[:subjects]).transform # transform subjects somehow
      end
    end
  end
```

Wdyt?

Nitemaeric: I'm not sure we need to do that either.

Let's assume params are directly consumed from the query params, from the following query:

```
?can_sponsor_visa=true&age_group=abc&applications_open=T&subjects[]=English&subjects[]=Mathematics
```

We can do:

```ruby
class CoursesQuery
  # Handling boolean params - Cast if not blank before usage
  # "applications_open" param would also work the same
  def sponsor_visa_filter
    return @scope if params[:can_sponsor_visa].blank?

    can_sponsor_visa = ActiveModel::Type::Boolean.cast(params[:can_sponsor_visa])

    @scope.where(can_sponsor_visa:)
  end

  # Allowlist/enum filter
  def age_group_filter
    return @scope unless params[:age_group].in?(%w[some list of valid values])

    @scope.where(age_group: params[:age_group])
  end
end
```

This way, the params are handled only if they are used and ignored if invalid/unused.

app/services/courses_query.rb
Nitemaeric:
```ruby
CoursesQuery.call(params) # accepts positional argument
CoursesQuery.new(params: params).call # accepts keyword argument
```

I'd do the following to keep them consistent:

```ruby
def self.call(...)
  new(...).call # pass everything received in .call to .new
end
```

app/forms/course_search_form.rb
tomas-stefano: I am not sure I agree but I think for the visa sponsorship only, it makes total sense for now. I will implement.

app/services/courses_query.rb
tomas-stefano: I see! Yes I will do.

app/services/courses_query.rb
```diff
@@ -22,6 +23,33 @@ def initialize(params:)
...
+        @scope.where(
+          can_sponsor_skilled_worker_visa: true
+        )
+      )
+  end
+
+  private
+
+  def log_query_info
+    CoursesQuery::Logger.new(@applied_scopes, @scope).call
```
Nitemaeric, line 53: I wonder if passing the query object itself makes sense?

```suggestion
    CoursesQuery::Logger.new(self).call
```

Nitemaeric (approved):
I'm good with these changes :)

app/services/courses_query.rb
tomas-stefano: Done

inulty-dfe (approved):

PR #4754, opened 2024-12-09
Im/Create ProviderPartnership association

inulty-dfe:
## Context

We have created the database tables for the `ProviderPartnership`.
This PR creates the associations and validations on those associations.

Next will be the controllers and routing

### DIFF
--- a/app/models/provider_partnership.rb
+++ b/app/models/provider_partnership.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 
 class ProviderPartnership < ApplicationRecord
+  belongs_to :training_provider, class_name: 'Provider'
+  belongs_to :accredited_provider, class_name: 'Provider'
 end
--- a/app/models/provider_partnership.rb
+++ b/app/models/provider_partnership.rb
@@ -3,4 +3,22 @@
 class ProviderPartnership < ApplicationRecord
   belongs_to :training_provider, class_name: 'Provider'
   belongs_to :accredited_provider, class_name: 'Provider'
+
+  validates :accredited_provider, uniqueness: { scope: %i[training_provider_id] }
+  validate :accredited_provider_must_be_accredited
+  validate :training_provider_must_not_be_accredited
+
+  def accredited_provider_must_be_accredited
+    return if accredited_provider.blank?
+
+    accredited_provider.accredited? ||
+      errors.add(:accredited_provider, :must_be_accredited)
+  end
+
+  def training_provider_must_not_be_accredited
+    return if training_provider.blank?
+
+    training_provider.accredited? &&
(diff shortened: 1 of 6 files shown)


inulty-dfe:
[Tech debt: Create conventional `has_many through:` for training provider -> accredited provider relationship](https://trello.com/c/FpYcSdO9/272-tech-debt-create-conventional-hasmany-through-for-training-provider-accredited-provider-relationship)

app/models/provider_partnership.rb
```diff
@@ -1,4 +1,24 @@
...
+    accredited_provider.accredited? ||
+      errors.add(:accredited_provider, :must_be_accredited)
+  end
+
+  def training_provider_must_not_be_accredited
+    return if training_provider.blank?
+
+    training_provider.accredited? &&
+      errors.add(:training_provider, :must_not_be_accredited)
+  end
```
tomas-stefano, line 23: What do you think we move to policy object? 

Just because it has roles and rules around the roles of a provider. 

The only thing missing is permission from a Policy object PoV but still sounds like.

```ruby
class AccreditingPolicy
  attr_reader :accredited_provider, :training_provider

  def initialize(accredited_provider:, training_provider:)
    @accredited_provider = accredited_provider
    @training_provider = training_provider
  end

  def accredited_provider_accredited?
    accredited_provider.accredited?
  end

  def training_provider_not_accredited?
    !training_provider.accredited?
  end
end
```

Then it could:

```ruby
validate :validate_partnership

  private

  def validate_partnership
    policy = AccreditingPolicy.new(
      accredited_provider: accredited_provider,
      training_provider: training_provider
    )

    errors.add(:accredited_provider, :must_be_accredited) unless policy.accredited_provider_accredited?
    errors.add(:training_provider, :must_not_be_accredited) unless policy.training_provider_not_accredited?
  end
```

inulty-dfe: I think there is a difference between policy and validation.

In my mind Policy is authorization. This is a data rule regardless of authorization.

This seems like a bit too much abstraction right now regardless of what we call it

Nitemaeric (approved):

app/models/provider_partnership.rb
tomas-stefano: No, the Policy Objects design pattern is not exclusively about authorization, although it is commonly used in that context. I think Pundit gem makes think Policy like authorisation but there are other cases:

```ruby
class DiscountPolicy
  def initialize(order)
    @order = order
  end

  def eligible?
    @order.total > 100 && @order.customer.loyalty_points > 500
  end
end
```

I will approve because I can see where you coming from in thinking too much abstraction but please consider if this start to become too complex and many other rules appears.

tomas-stefano (approved):

PR #4756, opened 2024-12-10
Td/foundation work applications open and send filter

tomas-stefano:
## Context

This PR adds the two filters to `/v2/results`:

* Applications open
* SEND (Special education needs)

Some caveats:

* In the v1 (live Find as it is now), courses with the status "open" show up in search results—even if they won’t actually accept applications until a future date. While you can add it to your application, Apply stops you from submitting it until then.
* Applications open on this PR will search by application status as it is on v1 (live) search. But will be reviewed before we changed to use the v2 on results. I've asked the team and the team is aware of the problem.

The SEND filter is behaving same as v1 and no issues.

screencapture-find-localhost-v2-results-2024-12-10-15_06_09
screencapture-find-localhost-v2-results-2024-12-10-15_06_25

## Guidance to review

1. Visit /v2/results
2. Does it works as expected?

### DIFF
--- a/app/services/courses_query.rb
+++ b/app/services/courses_query.rb
@@ -24,6 +24,8 @@ class CoursesQuery
 
   def call
     @scope = visa_sponsorship_scope
+    @scope = applications_open_scope
+    @scope = special_education_needs_scope
     @scope = @scope.distinct
 
     log_query_info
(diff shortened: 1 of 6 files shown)


tomas-stefano:
[Foundation for the pre-filter work - Add SEND filter](https://trello.com/c/HEPt3hAH/288-foundation-for-the-pre-filter-work-add-send-filter)

app/services/courses_query.rb
```diff
@@ -47,6 +49,22 @@ def visa_sponsorship_scope
       )
   end
 
+  def applications_open_scope
+    return @scope if params[:applications_open].blank?
+
+    @applied_scopes[:applications_open] = params[:applications_open]
+
+    @scope.where(application_status: Course.application_statuses[:open])
```
tomas-stefano, line 57: Like I mentioned on the PR description, I am maintaining the same functionality as the live results page and if we decide to change this change would be easier by just changing this line.

Nitemaeric (approved):

PR #4757, opened 2024-12-10
Td/foundation work study modes filter

tomas-stefano:
## Context

NOTHING TO SEE HERE (REVIEW https://github.com/DFE-Digital/publish-teacher-training/pull/4756 first)

NOTHING TO SEE HERE

## Guidance to review

NOTHING TO SEE HERE

### DIFF
--- a/app/services/courses_query.rb
+++ b/app/services/courses_query.rb
@@ -24,6 +24,7 @@ class CoursesQuery
 
   def call
     @scope = visa_sponsorship_scope
+    @scope = study_modes_scope
     @scope = applications_open_scope
     @scope = special_education_needs_scope
     @scope = @scope.distinct
@@ -47,6 +48,23 @@ class CoursesQuery
           can_sponsor_skilled_worker_visa: true
         )
       )
+  end
+
+  def study_modes_scope
+    study_modes = (params[:study_types].presence || []).compact_blank
+
+    return @scope if study_modes.blank?
+
+    @applied_scopes[:study_modes] = study_modes
+
+    case study_modes
+    when ['full_time']
+      @scope.where(study_mode: [Course.study_modes[:full_time], Course.study_modes[:full_time_or_part_time]])
+    when ['part_time']
+      @scope.where(study_mode: [Course.study_modes[:part_time], Course.study_modes[:full_time_or_part_time]])
+    else
+      @scope
+    end
   end
 
   def applications_open_scope
--- a/app/services/courses_query.rb
+++ b/app/services/courses_query.rb
@@ -24,6 +24,8 @@ class CoursesQuery
 
   def call
     @scope = visa_sponsorship_scope
+    @scope = applications_open_scope
+    @scope = special_education_needs_scope
     @scope = @scope.distinct
 
     log_query_info
@@ -45,6 +47,22 @@ class CoursesQuery
           can_sponsor_skilled_worker_visa: true
         )
       )
+  end
+
+  def applications_open_scope
+    return @scope if params[:applications_open].blank?
+
+    @applied_scopes[:applications_open] = params[:applications_open]
+
+    @scope.where(application_status: Course.application_statuses[:open])
+  end
+
+  def special_education_needs_scope
+    return @scope if params[:send_courses].blank?
+
+    @applied_scopes[:send_courses] = params[:send_courses]
+
+    @scope.where(is_send: true)
   end
 
   private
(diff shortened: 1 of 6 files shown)


tomas-stefano:
[Foundation for the pre-filter work - Add study type filter](https://trello.com/c/l4btu1x8/285-foundation-for-the-pre-filter-work-add-study-type-filter)

app/views/find/v2/results/index.html.erb
```diff
@@ -13,6 +13,11 @@
             <%= form.govuk_check_box :can_sponsor_visa, "true", multiple: false, label: { text: t("helpers.label.courses_query_filters.can_sponsor_visa"), size: "s" }, checked: params[:can_sponsor_visa].present? %>
           <% end %>
 
+          <%= form.govuk_check_boxes_fieldset :study_types, legend: { text: t("helpers.legend.courses_query_filters.study_type_html"), size: "s" }, class: "app-filter__group govuk-checkboxes--small", hidden: false do %>
+            <%= form.govuk_check_box :study_types, "full_time", label: { text: t("helpers.label.courses_query_filters.study_type_options.full_time"), size: "s" }, checked: params[:study_types].present? && params[:study_types].include?("full_time") %>
+            <%= form.govuk_check_box :study_types, "part_time", label: { text: t("helpers.label.courses_query_filters.study_type_options.part_time"), size: "s" }, checked: params[:study_types].present? && params[:study_types].include?("part_time") %>
+          <% end %>
```
tomas-stefano, line 19: Are we happy with these checked inteligence? It feels too much to abstract but also feels wrong to be here. But because we don't have a form object this needs to be here. Let me know your thoughts.

app/services/courses_query.rb
```diff
@@ -49,6 +50,23 @@ def visa_sponsorship_scope
       )
   end
 
+  def study_modes_scope
+    study_modes = Array(params[:study_types]).compact_blank
```
inulty-dfe, line 54: I'm not comfortable passing request params directly to a query object. We should be processing the params in the controller / mapping object.

app/views/find/v2/results/index.html.erb
inulty-dfe: I would probably put it in the controller and use instance variables. If the messy logic should go anywhere, move it as high up the stack as possible

app/services/courses_query.rb
Nitemaeric: What is the benefit of processing the params at the controller level? As this would de-couple the param handling from the query object, is there a case where we'd like to pass through non-processed params?

app/views/find/v2/results/index.html.erb
Nitemaeric: My gut feeling is the `checked: ` key shouldn't be needed. It doesn't do anything out of the ordinary.

app/services/courses_query.rb
inulty-dfe: ActiveSupport::Parameters are view level objects. Query objects are database level objects. Mixing the two blurs those lines between the frontend and the backend. I like there to be clear distinction between the two.

Nitemaeric: > It creates an implicit dependency between the Query object and the ActionController::Parameters class.

I'd disagree on the account of duck-typing. The underlying implementation at the Query Object level doesn't depend on the `params` being of type `ActionController::Parameters`, it only requires that the `params` respond to index accessors; `params[key]`.

app/views/find/v2/results/index.html.erb
tomas-stefano: For context, I did had the search form on an old PR:
https://github.com/DFE-Digital/publish-teacher-training/pull/4751/commits/d1b20e12d7aaa5299ce28cf4fa3c38128e33a144#diff-b05ade9e136163e376907cdd8caee4d0f89733201c361cdc6dcadfc7277d6635R9

This would be move there and the checked won't be needed and all params handling would be there in terms of formating and checked.

app/services/courses_query.rb
inulty-dfe: The hint is in the name `ActionController::Parameters`. I would like to keep ActionController constants in the request realm. If it's not related specifically to a HTTP request, I think it's better to not let these objects bleed into other domains

app/services/courses_query.rb
```diff
@@ -49,6 +50,22 @@ def visa_sponsorship_scope
       )
   end
 
+  def study_modes_scope
+    return @scope if params[:study_types].blank?
+
+    study_modes = params[:study_types]
+    @applied_scopes[:study_modes] = study_modes
+
+    case study_modes
```
Nitemaeric, line 59: I don't think this variable extraction is valuable enough to be worth the additional line.

Not a blocker though.

```suggestion
    @applied_scopes[:study_modes] = params[:study_types]

    case params[:study_types]
```

Nitemaeric (approved):
All else looks good to me!

app/services/courses_query.rb
tomas-stefano: Done

PR #4762, opened 2024-12-11
Add the page title and the results summary to the v2 results

tomas-stefano:
## Context

On the new search results page we want to add the description of “x courses found” or “no courses found” and we want to add the page title as well.

Scenarios:

* No courses found
* 1 course found
* X courses found
* 1,000 courses found

## What this PR does NOT do yet

Handle the no courses found message inside the search results (this will be done at later stage when all filters are done).

## Guidance to review

1. Visit /v2/results
2. Does it work for all scenarios?


tomas-stefano:
[Foundation for the pre-filter work - Search Results Header (and page title and no results scenario)](https://trello.com/c/waPQZrnn/290-foundation-for-the-pre-filter-work-search-results-header-and-page-title-and-no-results-scenario)

app/views/find/v2/results/index.html.erb
```diff
@@ -1,3 +1,9 @@
+<%= content_for :page_title, Find::ResultsSummary::View.new(courses_count: @courses_count).title %>
+
+<h1 class="govuk-heading-xl" data-qa="heading">
+  <%= render Find::ResultsSummary::View.new(courses_count: @courses_count) %>
+</h1>
+
```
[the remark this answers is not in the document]
tomas-stefano: Done! I put into the view but I think the helper will also be too much. I think now is good

app/controllers/find/v2/results_controller.rb
```diff
@@ -7,6 +7,7 @@ class ResultsController < Find::ApplicationController
 
       def index
         @courses = CoursesQuery.call(params:)
+        @courses_count = @courses.count
```
Nitemaeric, line 10: Do we need this variable? It feels like we could either call `@courses.count` or `@pagy.count` in the view directly. Is it to cache the value?

tomas-stefano: We use this count two times in the view and will send also for tracking. I think would be good to have this value.

For a while I was thinking that maybe courses query could have this value but we don't hold the courses query instance so I chosen this way.

Nitemaeric (approved):

PR #4766, opened 2024-12-13
Add HTTP request context to sentry

Nitemaeric:
## Context

We have a strange Sentry error that pops up frequently. We want to track down the cause of this issue, but we don't have enough information. It seems that the query parameters are causing the issue. Let's add them to our Sentry context.

- Add request contexts to Sentry.

## Guidance to review

- See the example event: https://dfe-teacher-services.sentry.io/issues/6138441690/events/a3c199c37edb4dfc8d5b36029fea10d7/?project=1377944&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=next-event&statsPeriod=14d&stream_index=0

### DIFF
--- a/app/controllers/find/results_controller.rb
+++ b/app/controllers/find/results_controller.rb
@@ -2,6 +2,7 @@
 
 module Find
   class ResultsController < ApplicationController
+    before_action :set_sentry_context
     before_action :render_feedback_component
 
     def index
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -40,6 +40,7 @@ gem 'jsonb_accessor'
 gem 'sentry-rails'
 gem 'sentry-ruby'
 gem 'sentry-sidekiq'
+gem 'stackprof'
```
tomas-stefano, line 43: Should we add only for development?

tomas-stefano (approved):
Looks good to me, just one minor question. Can you also put the benefits of adding the gem, the example outputs and etc?

PR #4774, opened 2024-12-16
Foundation for the pre filter work add Funding filter

tomas-stefano:
## Context

This PR adds the funding type filter on the v2 results page

## Caveats

For backwards compatibility the funding is string on the live site (v1 results page). This PR adds an array but also accept the funding as string for compatibility.

Screenshot 2024-12-16 at 11 37 57

## Guidance to review

1. Visit /v2/results
2. Does it work with all three checkboxes? Choosing one, two, three, zero?

### DIFF
--- a/app/controllers/find/v2/results_controller.rb
+++ b/app/controllers/find/v2/results_controller.rb
@@ -21,6 +21,7 @@ module Find
           :send_courses,
           :applications_open,
           :further_education,
+          :funding,
           study_types: [],
           qualifications: [],
           funding: []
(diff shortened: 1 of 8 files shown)


tomas-stefano:
[Foundation for the pre-filter work - Add fee or salary filter](https://trello.com/c/rxxk6lRT/289-foundation-for-the-pre-filter-work-add-fee-or-salary-filter)

app/controllers/find/v2/results_controller.rb
```diff
@@ -21,8 +21,10 @@ def search_courses_params
           :send_courses,
           :applications_open,
           :further_education,
+          :funding,
```
tomas-stefano, line 24: This one sounds duplicate but not! It is to be backwards compatible with the currently production page. See commit message and PR description.

Nitemaeric (approved):

PR #4786, opened 2024-12-19
Add `Settings.features.provider_partnerships` feature flag

inulty-dfe:
## Context

  This feature flag will be used to enable / disable routes and toggle the
  return values of certain methods so we can merge the updated provider
  partnerships code incrementally while maintaining the current
  behaviour.

  In test files we can have functional tests test the existing behaviour
  and the new implementation by switching on the feature flag
  for example:

    allow(Setttings.features).to receive(:provider_partnerships).and_return(true)


Nitemaeric (approved):

inulty-dfe:
https://trello.com/c/QnfYIYX6/332-add-feature-flag-to-merge-feature-concurrently

PR #4790, opened 2024-12-23
Td/53 improve filters add secondary filter part 2

tomas-stefano:
## Context

This PR adds the "Search" within the filter.

After talking with the tech lead we decide to add Stimulus only for this component and gradually switch over Stimulus.

Screenshot 2024-12-23 at 18 02 05

## Guidance to review

1. If you are local make sure your `config/settings/development.local.yml` has:
```yaml
features:
  v2_results: true
```
2. Visit `/v2/results`
3. Test the "Search" within the secondary filter
4. Also check the designs from the Trello ticket.


tomas-stefano:
Trello: https://trello.com/c/hnmuEiq8/53-improve-filters-add-secondary-filter

(the Github integration is not working)

app/views/find/v2/results/index.html.erb
```diff
@@ -18,9 +18,15 @@
...
-          <%= form.govuk_collection_check_boxes :subjects, form.object.secondary_subjects, :subject_code, :subject_name, legend: nil, include_hidden: false, small: true %>
-        <% end %>
+        <div class="filter-search govuk-form-group app-filter__group" data-controller="filter-search">
+          <%= form.govuk_check_boxes_fieldset :subjects, legend: { text: t("helpers.legend.courses_query_filters.secondary_html"), size: "s" }, hint: { text: t("helpers.hint.courses_query_filters.secondary"), class: "govuk-!-font-size-16" }, class: "app-filter__group govuk-checkboxes--small", hidden: false, multiple: false do %>
+            <% form.object.secondary_subjects.each do |subject| %>
+              <%= form.govuk_check_box :subjects,
+              subject.subject_code,
+              label: { text: subject.subject_name } %>
+            <% end %>
```
tomas-stefano, line 27: The govuk_collection_check_boxes with fieldset creates two classses `govuk-checkboxes` and  create more divs and form groups which makes the JS not work as expected. (i would have to call children many times. So I changed to loop through the secondary subjets.

Gemfile
```diff
@@ -193,7 +193,7 @@ group :development, :test do
   # Test framework
   gem 'rspec-rails', '7.1.0'
 
-  gem 'selenium-webdriver'
+  gem 'capybara-playwright-driver'
```
Nitemaeric, line 196: 🤔 I don't think we've committed to this change. Can we remove the playwright stuff for now please?

spec/spec_helper_system.rb
Nitemaeric, line 1: This file should also be removed from this PR, as we're not running system specs. But additionally, why the rename? The original file was only recently added by Iain.

tomas-stefano: spec helper is required first at rails helper - which means the support is requested first, then the selenium were loaded after.
In my view the system config should be done after everything is loaded.
Happy to move back although I don’t think it is the right order and can lead to oddd bugs hard to debug.

Nitemaeric: 🤔 Let's revisit this once the team is back from the holidays.

tomas-stefano: Ok, done

Nitemaeric (approved):

PR #4813, opened 2025-01-10
Replace #accredited_provider with #accredited on provider

inulty-dfe:
## Context

Providers are either Training providers or Accredited providers.
We currently use a column `accrediting_provider varchar(1)` to mark a provider as one or the other.
We use a Rails enum on top of this column which gives us more methods.

We get `accrediting_provider` which returns `"accredited_provider"` if the column value is `'Y'` or `"not_an_accredited_provider"` if the column value is `'N'`.
We also get methods 
#`accredited_provider?` which returns true or false.

We would like to simplify this by using a boolean column that returns true or false.

Every where we use `#accrediting_provider`, `#accredited_provider?`, etc, we can use `#accredited?`

## Guidance to review

The ProviderForm is a bit tricky. We submit a radio button with value `true` or `false`. The value is recieved as a string `'true'` or `'false'`. Both these values are truthy. We need to cast the value to a boolean to use it effectively in the class.
Normally we could pass this straight to an ActiveRecord model and it would be cast before insert but we aren't saving this directly to a db. 

### Data Migration
The PR includes a data migration that updates the `accredited` column based on the value of the `accrediting_provider` column.
This data migration will run as part of the CMD to start the docker container.

## Trello 
https://trello.com/c/tQIpgo0P/367-update-code-to-use-provideraccredited-instead-of-the-provider-enum-accreditingprovider

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -102,6 +102,7 @@ class Provider < ApplicationRecord
     accredited_courses.includes(:provider).where(provider: { recruitment_cycle: })
   end
 
+  scope :accredited, -> { where(accredited: true) }
   scope :by_provider_code, lambda { |provider_code|
     where('lower(provider_code) = ?', provider_code.to_s.downcase).first!
   }
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -287,6 +287,10 @@ class Provider < ApplicationRecord
     ).select('provider.*, COALESCE(a.courses_count, 0) AS included_accredited_courses_count')
   end
 
+  def self.accrediteds
+    %w[accredited_provider not_an_accredited_provider]
+  end
+
   def courses_count
     has_attribute?('included_courses_count') ? included_courses_count : courses.size
   end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -10,7 +10,7 @@ class Provider < ApplicationRecord
   include PgSearch::Model
   include VectorSearchable
 
-  before_save :update_searchable, if: :accredited_provider?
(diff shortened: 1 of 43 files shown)


app/models/provider.rb
```diff
@@ -295,6 +283,10 @@ def self.include_accredited_courses_counts(provider_code)
     ).select('provider.*, COALESCE(a.courses_count, 0) AS included_accredited_courses_count')
   end
 
+  def self.accrediteds
+    %w[accredited_provider not_an_accredited_provider]
+  end
```
Nitemaeric, line 288: I can't seem to find a call of this method.

spec/models/course/publishable_spec.rb
```diff
@@ -32,7 +32,7 @@
     end
 
     it 'is not publishable' do
-      course.accrediting_provider.not_an_accredited_provider!
+      course.accrediting_provider.toggle(:accredited)
```
Nitemaeric, line 35: I think it's better to be explicit here. `toggle` is a stateful method. We should be explicit in setting its value.

```suggestion
      course.accrediting_provider.accredited = false
```

app/models/provider.rb
inulty-dfe: It's a sneaky one
https://github.com/DFE-Digital/publish-teacher-training/pull/4813/files#diff-75e4e0f610e37aef5c89d0b3085b4f574370abddc824c4ef16afa422a03594caR48

Nitemaeric: 🤔 Hmm, definitely one difficult to debug.

But _anyway_, not something I think needs additional time spent on.

PR #4819, opened 2025-01-14
Change ratifying provider for courses

inulty-dfe:
## Context

Support Request

Update the ratifying provider for all courses of a given set of training providers in the current recruitment cycle.

## Trello
https://trello.com/c/wTPmCvLG/479-update-accrediting-provider-for-hubs

### DIFF
--- a/db/data/20250114155946_update_ratifying_provider.rb
+++ b/db/data/20250114155946_update_ratifying_provider.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class UpdateRatifyingProvider < ActiveRecord::Migration[8.0]
+  PROVIDERS = %w[5W1 1BJ 24H 4R4 3A7].freeze
+  TARGET_PROVIDER = '3A1'
+
+  def up
+    PROVIDERS.each do |training_provider_code|
+      UpdateRatifyingProviderForCourse.new(training_provider_code:, target_ratifying_provider_code: TARGET_PROVIDER).call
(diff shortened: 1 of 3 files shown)


db/data/20250114155946_update_ratifying_provider.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class UpdateRatifyingProvider < ActiveRecord::Migration[8.0]
+  PROVIDERS = %w[5W1 1BJ 24H 4R4 3A7].freeze
+  TARGET_PROVIDER = '3A1'
+
+  def up
+    PROVIDERS.each do |training_provider_code|
+      UpdateRatifyingProviderForCourse.new(training_provider_code:, target_ratifying_provider_code: TARGET_PROVIDER).call
```
Nitemaeric, line 9: This class is seemingly created specifically for this data migration. Does it make sense to add the class definition here as a nested class?

Nitemaeric: Even further, it doesn't look like we have plans to reuse that code. We can remove the class and call its implementation directly.

Nitemaeric: As for tests, I'm personally not too concerned - the logic here isn't too complex and we should have tested the migration locally.

Nitemaeric (approved):
Not enough to be a blocker though.

PR #4824, opened 2025-01-15
Bump Nokogiri to 1.18.1

inulty-dfe:
## Context

Version 1.18 of Nokogiri introduces separate versions of the gem for `gnu` and `musl`.
Our docker images are build on Alpine linux which uses `musl` so we need to explicitly declare a dependency on the `musl` version.

https://github.com/sparklemotion/nokogiri/blob/main/CHANGELOG.md#ruby

Bump Nokogiri to 1.18.1


Nitemaeric:
Do we also need to update the platform list?

inulty-dfe:
> Do we also need to update the platform list?

Can you elaborate? I don't know what you mean exactly

Nitemaeric:
> > Do we also need to update the platform list?
> 
> Can you elaborate? I don't know what you mean exactly

https://nokogiri.org/tutorials/installing_nokogiri.html#supported-platforms Nokogiri provides information about the platforms it supports and we can add them to our gemfile via `bundle lock --add-platform`.

inulty-dfe:
I don't think we need to do anything else here

Nitemaeric (approved):

PR #4825, opened 2025-01-15
Add course summary card to the new results page

tomas-stefano:
# Context

This PR adds all the course summary content into the search result.

The goal was:

* Simplify the views and the component and the locales in comparison from the old search result (you can compare with `SearchResultComponent` if you want)

If you wanna see the details of each field please read the commit message.

## Caveats

* There are N+1 queries on this PR - this will be addressed in future PR.

### DIFF
--- a/app/components/courses/summary_card_component.html.erb
+++ b/app/components/courses/summary_card_component.html.erb
@@ -1,0 +1,42 @@
+<li class="app-search-results__item">
+  <%= govuk_summary_card(classes: ["course-summary-card"], title:) do |card| %>
+    <%= card.with_summary_list(actions: false, classes: ["govuk-summary-list--no-border"]) do |summary_list| %>
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: location_key) %>
+        <% row.with_value(text: location_value) %>
+      <% end %>
+
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: fee_key) %>
+        <% row.with_value(text: fee_value) %>
+      <% end %>
+
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: length_key) %>
+        <% row.with_value(text: length_value) %>
+      <% end %>
+
+      <% if show_age_group_row? %>
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: age_group_key) %>
+          <% row.with_value(text: age_group_value) %>
+        <% end %>
+      <% end %>
+
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: qualification_key) %>
+        <% row.with_value(text: qualification_value) %>
+      <% end %>
+
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: degree_requirements_key) %>
+        <% row.with_value(text: degree_requirements_value) %>
+      <% end %>
+
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: visa_sponsorship_key) %>
+        <% row.with_value(text: visa_sponsorship_value) %>
+      <% end %>
+    <% end %>
+  <% end %>
+</li>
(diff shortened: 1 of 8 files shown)


app/components/courses/summary_card_component.rb
```diff
@@ -0,0 +1,226 @@
...
+    def location_value
+      return t('.location_value.not_listed') if @available_placements_count.zero?
+
+      if search_by_location?
+        safe_join(
+          [
+            t(
+              '.location_value.distance',
+              school_term:,
+              distance: content_tag(:span, pluralize(course.minimum_distance_to_search_location, 'mile'), class: 'govuk-!-font-weight-bold'),
```
tomas-stefano, line 39: the `minimum_distance_to_search_location` still doesn't exist. The idea is to return into the `SELECT` statement because we shouldn't be looping through all the placements in Ruby to calculate the minimum distance. 

This will be done when the location search is implemented.

app/components/courses/summary_card_component.rb
```diff
@@ -0,0 +1,226 @@
...
+      main_subject&.subject_name.in?(LANGUAGE_SUBJECTS)
+    end
+
+    def financial_incentive
+      @financial_incentive ||= main_subject&.financial_incentive
+    end
+
+    def main_subject
+      @main_subject ||= Subject.find_by(id: course.master_subject_id)
+    end
```
tomas-stefano, line 220: We need to create the "main subject" concept into the app.

For now I added here.

I will create a trello card and make sure it is finished before the prefilter work is delivered.

tomas-stefano:
https://trello.com/c/W9nszbPM/313-improve-results-update-results-and-course-summary-content

app/components/courses/summary_card_component.rb
```diff
@@ -0,0 +1,226 @@
+# frozen_string_literal: true
+
+module Courses
+  class SummaryCardComponent < ViewComponent::Base
+    attr_reader :course, :search_params, :available_placements_count, :minimum_distance_to_search_location
+
+    def initialize(course:, search_params:)
+      @course = course
+      @search_params = search_params
```
inulty-dfe, line 9: I think it's strange to see something called `search_params` in this component.

How about `location_details` and `visa_sponsorship`

app/models/course.rb
```diff
@@ -892,6 +892,44 @@ def name_and_code
     "#{name} (#{course_code})"
   end
 
+  def available_placements
```
inulty-dfe, line 895: Could this be better living in a module?

`AvailablePlacements.new(course:).call`

inulty-dfe (commented):
Looks good.

Couple of suggestions

app/components/courses/summary_card_component.rb
tomas-stefano: Makes sense. Only passes the params it uses :)

app/models/course.rb
tomas-stefano: It is a good idea, but this code will be removed because this should be in the db `SELECT` list on the courses query, so I am reluctantly to move to a class now.

app/components/courses/summary_card_component.rb
tomas-stefano: Done!

app/components/courses/summary_card_component.html.erb
```diff
@@ -0,0 +1,42 @@
+<li class="app-search-results__item">
+  <%= govuk_summary_card(classes: ["course-summary-card"], title:) do |card| %>
+    <%= card.with_summary_list(actions: false, classes: ["govuk-summary-list--no-border"]) do |summary_list| %>
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: location_key) %>
+        <% row.with_value(text: location_value) %>
+      <% end %>
```
Nitemaeric, line 7: I have a feeling we can remove the HTML from the ruby code.

```suggestion
      <% summary_list.with_row do |row| %>
        <% row.with_key(text: location_key) %>
        <% row.with_value do %>
          <%= location_value %>
          
          <% if location_hint %>
            <div class="govuk-hint govuk-!-font-size-16"><%= location_hint %></div>
          <% end %>
        <% end %>
      <% end %>
```

It should help with the template visualisation too. Now we now that all fields have a key, a value, and a possible hint.

tomas-stefano: Done!

Nitemaeric (approved):

PR #4826, opened 2025-01-15
Add Publish accredited partnerships

inulty-dfe:
## Context

We are changing how Provider partnerships are created and used in Publish.

This PR makes the new provider_partnerships system work behind a feature flag. Both provider enrichments and provider partnerships exist side by side so we can incrementally merge the feature while proving it works correctly.

Add the Provider Partnership system to the Publish namespace.

1. We are keeping the existing UI content "providers" not "partners" until we can have a content review.
2. The urls are changing to `/provider-partnerships` but the Accredited provider search maintains `accredited-providers`
3. Add new controllers and views to work alongside the existing implementation
4. Switch URLs based on the value of the feature flag
  5. Target of the "Accredited providers" link in the Navigation bar
  6. The target of the redirect when an accredited provider is chosen during a search
  7. Back links need to point to the new partnerships controller

I've tried to separate the two implementations as much as possible so if statements aren't littered around the code base. Almost every where we see `Settings.features.provider_partnerships` will need to be updated once the implemenetation is fully launched.

## Guidance to review

The feature is turned off - so the review app is using Enrichments.
If you want to test it with Partnerships, we can enable the feature

When the feature flag is enabled, it impacts on areas of the test suite that are not directly associated with this change.
I've added guards or alternative tests to test for both feature scenarios.

### Trello 

https://trello.com/c/GaeNMFMB/380-add-providerpartnerships-to-the-publish-namespace


app/controllers/publish/providers/accredited_partnerships/checks_controller.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Publish
+  module Providers
+    module AccreditedPartnerships
+      class ChecksController < ApplicationController
+        def show
+          provider_partnership_form
```
tomas-stefano, line 8: Wdyt? `before_action :set_provider_partnership_form, only: %i[show update]`

app/forms/provider_partnership_form.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class ProviderPartnershipForm < Form
```
tomas-stefano, line 3: Do we want to continue to inherit from `Form` class? cc @Nitemaeric

app/forms/provider_partnership_form.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  private
+
+  def assign_attributes_to_model
+    if model.persisted?
+      model.description = description
+    else
+      model.accredited_partnerships.build(accredited_provider_id: accredited_provider_id, description: description)
+    end
+  end
```
tomas-stefano, line 29: Is this method be called? On the controller you're calling save on the partnership object and not the form?

app/helpers/navigation_bar_helper.rb
```diff
@@ -13,9 +13,22 @@ def navigation_items(provider)
...
+      if provider.accredited_provider?
+        { name: t('navigation_bar.training_partners'), url: publish_provider_recruitment_cycle_training_providers_path(provider.provider_code, provider.recruitment_cycle_year) }
+      else
+        { name: t('navigation_bar.accredited_partnerships'), url: publish_provider_recruitment_cycle_accredited_partnerships_path(provider.provider_code, provider.recruitment_cycle_year) }
+      end
+    elsif provider.accredited_provider?
+      { name: t('navigation_bar.training_partners'), url: publish_provider_recruitment_cycle_training_providers_path(provider.provider_code, provider.recruitment_cycle_year) }
+    else
+      { name: t('navigation_bar.accredited_provider'), url: publish_provider_recruitment_cycle_accredited_providers_path(provider.provider_code, provider.recruitment_cycle_year) }
+    end
```
tomas-stefano, line 32: Can we achieve the same by?

```ruby
21   def partnership_link(provider)                                                 
  22     if Settings.features.provider_partnerships && !provider.accredited_provider? 
  23       { name: t('navigation_bar.accredited_partnerships'), url: publish_provider_recruitment_cycle_accredited_partnerships_path(provider.provider_code, provid     er.recruitment_cycle_year) }
  24     elsif provider.accredited_provider?                                          
  25       { name: t('navigation_bar.training_partners'), url: publish_provider_recruitment_cycle_training_providers_path(provider.provider_code, provider.recruitm     ent_cycle_year) }
  26     else                                                                         
  27       { name: t('navigation_bar.accredited_provider'), url: publish_provider_recruitment_cycle_accredited_providers_path(provider.provider_code, provider.recr     uitment_cycle_year) }                                                                                                                                         
  28     end
```

app/views/publish/providers/accredited_partnerships/_cannot_remove.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <p class="govuk-body">
+      If you need to change an accredited provider for a course which is published, please contact us at
+      <%= bat_contact_mail_to %>.
+    </p>
+
+    <p class="govuk-body">
+      <%= govuk_link_to(t("cancel"), publish_provider_recruitment_cycle_accredited_partnerships_path) %>
+    </p>
+  </div>
+</div>
```
tomas-stefano, line 29: Can we move to locales?

app/views/publish/providers/accredited_partnerships/checks/show.html.erb
```diff
@@ -0,0 +1,42 @@
...
+            component.with_row do |row|
+              row.with_key { "About the accredited provider" }
+              row.with_value { markdown @provider_partnership_form.description }
+              row.with_action(text: "Change", href: new_publish_provider_recruitment_cycle_accredited_partnership_path(goto_confirmation: true), visually_hidden_text: "accredited provider description")
+            end
+          end %>
+
+      <%= govuk_warning_text(text: "All users at #{@provider_partnership_form.provider_name} will be sent an email to let them know they’ve been added.") %>
+
+      <%= f.govuk_submit(t(".add")) %>
```
tomas-stefano, line 35: Can we move the rows to locales?

app/views/publish/providers/accredited_partnerships/edit.html.erb
```diff
@@ -0,0 +1,36 @@
...
+    <%= govuk_back_link_to(publish_provider_recruitment_cycle_accredited_partnerships_path) %>
+  <% end %>
+
+        <%= f.govuk_text_area(
+          :description,
+          label: { text: t(".title"), size: "l", tag: "h1" },
+          hint: -> { render "hint", hint: t(".hint") },
+          caption: { text: @provider_partnership_form.accredited_provider.provider_name, size: "l" },
+          max_words: 100,
+          rows: 10
```
tomas-stefano, line 24: Can we move the magic numbers to the form object?

app/forms/provider_partnership_form.rb
inulty-dfe: I need to inherit while we rely on the stashing of form content from step to step.
I'll continue to do that until we have an alternative

app/forms/provider_partnership_form.rb
inulty-dfe: Good point. I need to revise this area

app/helpers/navigation_bar_helper.rb
inulty-dfe: It's all getting deleted asap lads

app/controllers/publish/providers/accredited_partnerships/checks_controller.rb
inulty-dfe: I'm not excited by it tbh

app/views/publish/providers/accredited_partnerships/_cannot_remove.html.erb
inulty-dfe: I don't know what your referring to

app/views/publish/providers/accredited_partnerships/edit.html.erb
inulty-dfe: I don't really know what to do with it right now. I would like to ignore it for now

app/views/publish/providers/accredited_partnerships/_cannot_remove.html.erb
tomas-stefano: I mean all this content to move to locales:

```
<span class="govuk-caption-l"><%= @provider_partnership.accredited_provider.provider_name %></span>
      You cannot remove this accredited provider
    </h1>

    <p class="govuk-body">
      <%= @provider_partnership.accredited_provider.provider_name %> is an
      accredited provider for courses run by <%= @provider.provider_name %>. At least one of these courses is
      currently published on Find.
    </p>

    <p class="govuk-body">
      If you need to change an accredited provider for a course which is published, please contact us at
      <%= bat_contact_mail_to %>.
    </p>
```

Nitemaeric:
@inulty-dfe Adding a second Accredited Provider seems to retain the "About the Accredited Provider" text. Seems like a bug?

inulty-dfe:
> @inulty-dfe Adding a second Accredited Provider seems to retain the "About the Accredited Provider" text. Seems like a bug?

This is how the stashing works. It's an existing bug, behaves the same way on QA. I don't want to mess with it here.

Nitemaeric (commented):
All seems to work well! I think we need to get rid of the old data modelling first before I suggest we dive deeper into the code conventions.

Nitemaeric (approved):

app/controllers/publish/providers/accredited_partnerships/checks_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
+
+        def notify_accredited_provider_users
+          @partnership.accredited_provider.users.each do |user|
+            ::Users::OrganisationMailer.added_as_an_organisation_to_training_partner(
+              recipient: user,
+              provider: provider,
+              accredited_provider: @partnership.accredited_provider
+            ).deliver_later
+          end
```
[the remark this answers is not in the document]
inulty-dfe: I've removed it from the form.

app/views/publish/providers/accredited_partnerships/_cannot_remove.html.erb
inulty-dfe: Done :vulcan_salute:

inulty-dfe:
> All seems to work well! I think we need to get rid of the old data modelling first before I suggest we dive deeper into the code conventions.

Some of the existing patterns really prevent us from making improvements. Trying to migrate a seam of functionality at the same time as keeping in line with existing systems is hellish. But necessary.

It would be nice to tackle the form before the function

PR #4831, opened 2025-01-17
Revamp layouts

Nitemaeric:
## Context

1. Our `application.html.erb` layout has code that handles multiple namespace variants, namely the `support` vs `publish` namespaces. This creates increases the complexity of the layout and its less straight-forward about what should render where and when.

2. While working on layouts, we were given the heads up that we want to update the Find header.

3. Since we're working on a new view component, enhance the view component preview pages.

- Re-organise layouts.
  - Remove layout inheritance.
  - 1 layout per namespace.
  - Publish uses the `application.html.erb` layout. Top-level pages are generally Publish-facing pages.

- Reorganise stylesheets to align with namespace patterns.

  ```
  app
    controllers
      find
      publish
    assets
      stylesheets
        find
          application.scss
        publish
          application.scss
    views
      layouts
        find.html.erb
        application.html.erb # This should be "publish", but there 
  ```

- Change the Find and Support layouts to use the new `HeaderComponent` that includes the `service_navigation` component.

## Guidance to review

- The publish header is to stay as-is, while we figure out how we want the organisation-switcher to work with it.
- Does this break any shared pages? Any weird titles?

### Publish

CleanShot 2025-01-17 at 16 49 43@2x

### Find

CleanShot 2025-01-17 at 16 50 13@2x

### Support

CleanShot 2025-01-17 at 16 50 32@2x

### View component previews

CleanShot 2025-01-17 at 17 03 02@2x


Nitemaeric:
> Bit of a styling issue here when there is a lot in the title bar image

Interesting, I could have sworn that I didn't change that component. 😅

spec/features/support/view_components_spec.rb
```diff
@@ -2,22 +2,21 @@
 
 require 'rails_helper'
 
-RSpec.feature 'view components' do
```
inulty-dfe, line 5: I wouldn't remove the explicit receiver here. It's good practice to not rely on implicit receivers in the global namespace

inulty-dfe (approved):
Great work!

It feels like a massive improvement. And the commits were easy to follow.

:100:

spec/features/support/view_components_spec.rb
Nitemaeric: I moved the file and dropped the receiver at the same time and it fixed an issue and I just ran with it 😅 .

Brought back the receiver.

PR #4833, opened 2025-01-20
Add provider partnerships to support interface

inulty-dfe:


app/controllers/support/providers/accredited_partnerships/checks_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        def notify_accredited_provider_users
+          @partnership.accredited_provider.users.each do |user|
+            ::Users::OrganisationMailer.added_as_an_organisation_to_training_partner(
+              recipient: user,
+              provider: provider,
+              accredited_provider: @partnership.accredited_provider
+            ).deliver_later
+          end
+        end
```
tomas-stefano, line 53: We do have the same logic on the provider controller, can we move to a service? 
Maybe add to `app/notifiers` or something similar?

app/controllers/support/providers/accredited_partnerships_controller.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module Support
+  module Providers
+    class AccreditedPartnershipsController < ApplicationController
+      include ClearStashable
```
tomas-stefano, line 6: One line calling to clear the stash from the form, do we need the module even if is only one call to each controller?

app/controllers/support/providers/accredited_partnerships_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+          recruitment_cycle_year: @recruitment_cycle.year,
+          provider_id: @provider.id
+        )
+      end
+
+      private
+
+      def cannot_delete
+        @cannot_delete ||= provider.courses.exists?(accredited_provider_code: params[:accredited_provider_code])
+      end
```
tomas-stefano, line 81: The cannot delete sounds like an important rule:

```
app/controllers/publish/providers/accredited_partnerships_controller.rb
52:        cannot_delete
68:      def cannot_delete
69:        @cannot_delete ||= provider.courses.exists?(accredited_provider_code: params[:accredited_provider_code])

app/controllers/publish/providers/accredited_providers_controller.rb
43:        cannot_delete
47:        return if cannot_delete
59:      def cannot_delete
60:        @cannot_delete ||= provider.courses.exists?(accredited_provider_code: accredited_provider.provider_code)

app/controllers/support/providers/accredited_partnerships_controller.rb
60:        cannot_delete
65:        return if cannot_delete
79:      def cannot_delete
80:        @cannot_delete ||= provider.courses.exists?(accredited_provider_code: params[:accredited_provider_code])

app/controllers/support/providers/accredited_providers_controller.rb
52:        cannot_delete
56:        return if cannot_delete
71:      def cannot_delete
72:        @cannot_delete ||= provider.courses.exists?(accredited_provider_code: accredited_provider.provider_code)
```

Should we move this to a policy object?

app/controllers/support/providers/accredited_partnerships_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+        partnership = provider.accredited_partnerships.find_by(accredited_provider: partner)
+        @accredited_provider_form = ::ProviderPartnershipForm.new(current_user, partnership, params: accredited_provider_params)
+
+        if @accredited_provider_form.save!
+          redirect_to support_recruitment_cycle_provider_accredited_partnerships_path(
+            recruitment_cycle_year: @recruitment_cycle.year,
+            provider_id: @provider.id
+          )
+
+          flash[:success] = t('support.providers.accredited_providers.edit.updated')
```
Nitemaeric, line 52: Let's add the flash to the redirection.

```suggestion
          redirect_to support_recruitment_cycle_provider_accredited_partnerships_path(
            recruitment_cycle_year: @recruitment_cycle.year,
            provider_id: @provider.id
          ), flash: { success: t('support.providers.accredited_providers.edit.updated') }
```

app/controllers/support/providers/accredited_partnerships_controller.rb
```diff
@@ -0,0 +1,110 @@
...
+        return if cannot_delete
+
+        provider.accredited_partnerships.find_by(accredited_provider_id: partner.id).destroy
+
+        flash[:success] = t('support.providers.accredited_providers.delete.updated')
+
+        redirect_to support_recruitment_cycle_provider_accredited_partnerships_path(
+          recruitment_cycle_year: @recruitment_cycle.year,
+          provider_id: @provider.id
+        )
```
Nitemaeric, line 74: Here too, let's add the flash to the redirect call.

app/controllers/support/providers/accredited_partnerships_controller.rb
Nitemaeric: I think this one is blocking though. We really should have a proper policy for permissions.

inulty-dfe:
@Nitemaeric I'm not sure about the Policy thing. There are no permissions involved, it's just that it's not legal to remove accreditation while a provider is a ratifying provider. All the Policy objects seem to restrict actions based on the users admin status

Nitemaeric (approved):

PR #4843, opened 2025-01-22
Add postgis extension - a spatial database extender for PostgreSQL

tomas-stefano:
## Context

This PR adds the postgis extension to Publish/Find.

The postgis adapter does not support Rails 8 at the moment: https://github.com/rgeo/activerecord-postgis-adapter/issues/411#issuecomment-2606854358

So using the extension in it's raw mode.

I’ve encountered issues with the internal tables PostGIS uses for geospatial calculations when recreating the database.

Running the usual setup commands:

```
rails db:drop db:create db:schema:load db:migrate
```

leads to this error:

```
PG::DependentObjectsStillExist: ERROR:  cannot drop table spatial_ref_sys because extension postgis requires it
```

To work around this, I added the following initializer to ignore spatial_ref_sys during schema dumps:

```
ActiveRecord::SchemaDumper.ignore_tables |= ["spatial_ref_sys"]
```

## Guidance to review

    - Run in the review db: `SELECT postgis_full_version();` (e.g `ActiveRecord::Base.connection.execute("SELECT postgis_full_version();").values`)
    - run `SELECT *  FROM spatial_ref_sys LIMIT 10;`
    - run `SELECT * FROM geometry_columns  LIMIT 10;`
2. Does terraform includes postgis?


RMcVelia (approved):
terraform change looks good

Nitemaeric (approved):

formigarafa:
Hey, I've seen your comment on issue 441 of activerecord-postgis-adapter repo. I've been investigating myself on how to use postgis without that adapter gem. The adapter is great but I think it could be split into separate concerns, easier to upgrade.
Anyway, about your comment over there about spatial_ref_sys workaround and I don't know for sure but maybe it could be related with this adjustment they made: https://github.com/rgeo/activerecord-postgis-adapter/blob/master/lib/active_record/connection_adapters/postgis/database_statements.rb

I did not try myself but judging by the nature of your adjustments I feel this could resolve your problem.

PR #4852, opened 2025-01-24
Foundation work implement location search and radius filter

tomas-stefano:
## Context

For the new results page, we need to enhance the location search functionality:

Allow users to search for courses within a specific radius of a given location.
Sort the results by courses with the nearest available placements.

We are replacing the current search mechanism with PostGIS for geolocation searches, which offers improved performance and accuracy.

* Users can search for courses by location within a specified radius.
* If no radius is provided, a default radius is applied.
* Results are sorted by courses with the nearest placements.
* The implementation uses PostGIS for geolocation calculations.

## Things won't be done on this PR

* The site latitude and longitude geometry ST_MakePointM(float x, float y, float m) could be store in the Database but this will be done in a further iteration.
* This is only the backend, the interface will be done in a separate card and separate PR.

## Guidance to review

1. Visit the review and pass the latitude, longitude and location params: https://find-review-4852.test.teacherservices.cloud/v2/results?latitude=51.5345978&location=Some+Rd%2C+London+NW9%2C+UK&longitude=-0.1676324

### DIFF
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -2,6 +2,8 @@
 
 module Courses
   class Query
+    DEFAULT_RADIUS_IN_MILES = 10
+
     def self.call(...)
       new(...).call
     end
@@ -32,7 +34,7 @@ module Courses
       @scope = applications_open_scope
       @scope = special_education_needs_scope
       @scope = funding_scope
-      @scope = @scope.distinct
+      @scope = location_scope
 
       log_query_info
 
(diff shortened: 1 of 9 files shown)


tomas-stefano:
https://trello.com/c/12EKm1cu/390-foundation-work-implement-location-search-with-radius-and-sorting-for-new-results-page

app/services/courses/query.rb
```diff
@@ -125,6 +127,49 @@ def funding_scope
       @scope.where(funding: params[:funding])
     end
 
+    def location_scope
+      return @scope.distinct if params[:latitude].blank? || params[:longitude].blank?
+
+      radius_in_miles = (params[:radius] || DEFAULT_RADIUS_IN_MILES).to_f
+      radius_in_meters = radius_in_miles * 1609.34
+      latitude = params[:latitude].to_f
```
inulty-dfe, line 135: Would `Float(params[:latitude])` be preferrable here and below?

inulty-dfe (approved):
Looks ok to me

app/services/courses/query.rb
tomas-stefano: That's a good idea. Done!

PR #4864, opened 2025-01-27
Implement search by provider in the new results page

tomas-stefano:
## Context

Add (only the backend) for search a provider by name.

It is a exact match because there would be a JS automcomplete in the same the live site does. The non JS route will be visited in a further iteration of this work.

## Guidance to review

1. The new parameter: `v2/results?provider_name=Oxford+University`
2. The old parameter for backwards compatibiliy: `v2/results?provider.provider_name=Oxford+University`


tomas-stefano:
https://trello.com/c/SHvVgoPj/407-foundation-for-the-pre-filter-work-add-provider-search

Nitemaeric (commented):
If we're doing a lookup, would it make more sense to pass through an ID? This would allow for the JS and non-JS routes to pass separate keys, `provider_id` and `provider_name`. The former being passed through from the auto-complete value.

tomas-stefano:
> If we're doing a lookup, would it make more sense to pass through an ID? This would allow for the JS and non-JS routes to pass separate keys, `provider_id` and `provider_name`. The former being passed through from the auto-complete value.

So both, we search by provider name if there is no ID, or search through the ID if provider id is there? 
If so sounds good, I will action this.

inulty-dfe:
If we change to using `provider_id` are we intending to update all the other places where the provider search is used?

Nitemaeric:
> If we change to using `provider_id` are we intending to update all the other places where the provider search is used?

Possibly. The alternative I had in mind over `provider_id` is `provider_code`, but I'm not sure if they're unique or not.

inulty-dfe:
The name fills the autocomplete. The actual form sends the `provider_code`. If the controller receives a text fragment it produces a list of radios matching the text. That's basically what I delete from the add course flow

tomas-stefano:
@Nitemaeric / @inulty-dfe that's done

spec/services/courses/query_spec.rb
```diff
@@ -307,6 +307,82 @@ def initialize(course, minimum_distance_to_search_location:)
       end
     end
 
+    context 'when searching by provider' do
```
Nitemaeric, line 310: I think this is missing the logic of the current recruitment cycle. When searching by ID, you don't allow it to match with providers from another cycle. We should test this.

tomas-stefano: I agree. That's done.

Nitemaeric (approved):

app/services/courses/query.rb
```diff
@@ -150,6 +151,26 @@ def funding_scope
...
+    def provider_scope
+      return @scope if params[:provider_code].blank? && params[:provider_name].blank?
+
+      @applied_scopes[:provider] = params[:provider_code] || params[:provider_name]
+
+      providers = if params[:provider_code].present?
+                    RecruitmentCycle.current.providers.where(provider_code: params[:provider_code])
+                  else
+                    RecruitmentCycle.current.providers.where(provider_name: params[:provider_name])
+                  end
```
inulty-dfe, line 163: Should we be validating the input if it's not blank?

tomas-stefano: We already do some lines above

inulty-dfe (approved):

PR #4865, opened 2025-01-28
Drop column `accrediting_provider` on `provider` table

inulty-dfe:
## Context

We introduced the boolean column `accredited` to the Provider model/table to replace the text column `accrediting_provider`

The `accrediting_provider` columns needs to be dropped from the database.

- accrediting_provider column is ignored in Provider model
- removed from analytics
- dropped from the table

## Trello
https://trello.com/c/fHZxEH0A/398-drop-column-accreditingprovider-on-provider-table


spec/services/publish/accredited_provider_updater_spec.rb
```diff
@@ -13,9 +13,9 @@
     allow(Settings.features).to receive_messages(provider_partnerships: false)
   end
 
-  let(:provider_code) { 'A0' }
+  let(:provider_code) { 'R00' }
   let(:recruitment_cycle_year) { 2024 }
-  let(:new_accredited_provider_code) { 'A1' }
+  let(:new_accredited_provider_code) { 'S00' }
```
inulty-dfe, line 18: Provider codes are 3 characters long. We've had some issues with duplicate codes in the test suite so I've made them as unique as I can

Nitemaeric (approved):

PR #4866, opened 2025-01-28
Remove all `Provider#accrediting_provider` code

inulty-dfe:
## Context

We recently replaced the use of `accrediting_provider` with `accredited` on the Provider model.
Some of the code was not updated.
This PR finishes the migration by removing all references to the old column

Any remaining references to `accrediting_provider` should be relevant to the course only.

Change all places where `Provider#accrediting_provider` is used to determine if a provider is accredited to `Provider#accredited`

## Guidance to review

Has all references correctly been removed?

## Trello
https://trello.com/c/VhkOfHZG/410-remove-all-remaining-references-to-provideraccreditingprovider

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -979,7 +979,7 @@ en:
             provider_type:
               blank: "Provider type can't be blank"
               format: Accredited provider cannot be a school
-            accrediting_provider:
+            accredited:
               format: '%{message}'
               invalid_provider_type: 'Accredited provider must not have Provider type school'
             email:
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -979,7 +979,7 @@ en:
             provider_type:
               blank: "Provider type can't be blank"
               format: Accredited provider cannot be a school
-            accrediting_provider:
+            accredited:
               format: '%{message}'
               invalid_provider_type: 'Accredited provider must not have Provider type school'
```
Nitemaeric, line 984: Just a heads up, from some discussions with Simon, I am going to propose changing these provider types. Coupling the type and accreditation only leads to validation errors, we can infer the user-facing type based on accreditation and a lower-level provider type, such as "University" or "School" or "Institute".

Nitemaeric (approved):
Looks good - A minor comment about upcoming plans.

PR #4876, opened 2025-01-30
Replace Support and Advice in Find for TDA courses

inulty-dfe:
## Context

  TDA courses are not the responsibility of GiT. We do not want to
  invite candidates to seek advice from GiT for these courses

Create new partial that gets rendered to the Course page instead of the Support and Advice partial when the course is a TDA course.
Reuse existing styles but apply them to a new id selector.

## Guidance to review

[TDA course on Review App](https://find-review-4876.test.teacherservices.cloud/course/S72/G719?search_params=age_group%3Dsecondary%26applications_open%3Dtrue%26can_sponsor_visa%3Dfalse%26has_vacancies%3Dtrue%26l%3D2%26subjects%255B%255D%3DG1%26university_degree_status%3Dfalse%26visa_status%3Dfalse)

image

## Trello
https://trello.com/c/lL4j73nC/39-update-support-callout-box-for-tda-courses


config/locales/find.yml
```diff
@@ -209,6 +209,11 @@ en:
           uk_citizens: UK citizens
           non_uk_citizens: Non-UK citizens
           course_fees: The course fees for UK citizens in %{cycle_range} are %{fee}.
+      tda_advice:
+        heading: Teacher degree apprenticeships (TDAs)
+        content: On a teacher degree apprenticeship you’ll work in a school and earn a salary while getting a bachelor’s degree and qualified teacher status (QTS).
```
tomas-stefano, line 214: we have a similar content on, can we use?

232:          tda_with_qts_content_html: On a teacher degree apprenticeship you’ll work in a school and earn a salary while getting a bachelor’s degree and qualified teacher status (QTS). %{tda_link}

config/locales/find.yml
```diff
@@ -209,6 +209,11 @@ en:
...
+        link_text: Find out more about teacher degree apprenticeships.
+        link_url: https://getintoteaching.education.gov.uk/train-to-be-a-teacher/teacher-degree-apprenticeships
```
tomas-stefano, line 216: Same with above.
There is one component that uses. Maybe we can move this translations around?
```
       <%= t(                                                                     
         ".tda_with_qts_content_html",                                            
         tda_link: govuk_link_to(t(".find_out_more_about_tda"), t("find.get_into_teaching.url_tda"))
       ) %> 
```

config/locales/find.yml
inulty-dfe: They're in different components. I don't think it's worth the effort to make them the same value. They could easily change and that would be harder

Nitemaeric: If the text belongs to the components, let's keep them separate.

app/assets/stylesheets/components/_advice.scss
```diff
@@ -16,7 +16,8 @@
   margin-bottom: 0;
 }
 
-#section-advice-and-support {
+#section-advice-and-support,
+#section-tda-advice {
```
Nitemaeric, line 20: Can we use classes for styling instead of IDs please?

PR #4877, opened 2025-01-30
Add sorting to the new results page

tomas-stefano:
## Context

Following the same pattern
described on
https://becoming-a-teacher.design-history.education.gov.uk/find-teacher-training/sorting-by-course-name-in-the-search-results/

To add here later

## Guidance to review

To add here later

### DIFF
--- a/config/locales/find.yml
+++ b/config/locales/find.yml
@@ -403,7 +403,7 @@ en:
             other: "%{formatted_count} courses found"
   helpers:
     legend:
-      courses_query_filters:
+      courses_search_form:
         can_sponsor_visa_html: Visa sponsorship<span class="govuk-visually-hidden"> filter</span>
         primary_html: Primary<span class="govuk-visually-hidden"> filter</span>
         secondary_html: Secondary<span class="govuk-visually-hidden"> filter</span>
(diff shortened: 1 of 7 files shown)


tomas-stefano:
https://trello.com/c/UMhYViSt/413-foundation-for-the-pre-filter-work-add-sorting

config/locales/find.yml
```diff
@@ -403,7 +403,7 @@ en:
             other: "%{formatted_count} courses found"
   helpers:
     legend:
-      courses_query_filters:
+      courses_search_form:
```
tomas-stefano, line 406: I renamed because I realised courses_query_filters was not a good name. So changed the same name of the form object.

config/locales/find.yml
```diff
@@ -441,6 +446,14 @@ en:
           fee: Fee - no salary
           salary: Salary
           apprenticeship: Teaching apprenticeship - with salary
+        ordering:
+          location: Sorted by distance
```
tomas-stefano, line 450: This is not a label ALTHOUGH I'd though it would make sense to everything about ordering to be here.

We don't display labeling when sorting by distance

spec/services/courses/query_spec.rb
```diff
@@ -847,6 +859,52 @@ def initialize(course, minimum_distance_to_search_location:)
...
+            expect(results).to match_collection(
+              [
+                course_london_result,
+                course_canary_wharf_result,
+                course_lewisham_result
+              ],
+              attribute_names: %w[name minimum_distance_to_search_location]
+            )
+          end
+        end
```
tomas-stefano, line 888: This is important to have this test because on the location search the SQL needs to group the courses but the custom order needs to add the `SELECT provider.provider_name` which will cause SQL errors.  The location search does not need to include this select since is ordering by location.

PR #4882, opened 2025-02-03
Create Gias library

inulty-dfe:
## Context

Add a set of classes that will be used to download, transform and import GIAS school records.

The school attributes we want to import are limited to those columns we have on the `gias_school` table.

The CSV has bytes incompatible with UTF-8. The response is transcoded from `window-1252`

Add Gias module
```ruby
Gias::Downloader
Gias::Transformer
Gias::Importer
Gias::Log
```

Add `GiasImportJob` to stitch them all together.
This job is scheduled to run at 02:30 in production.
No need to run it in QA because the DB is restored from prod.

## Trello
https://trello.com/c/xz1ndaBH/357-1-2-sync-gias-data-daily-with-a-scheduled-job
https://trello.com/c/85ecj4GE/88-investigate-regularly-updating-gias-data-in-publish

## Guidance to review

Gias schools are being updated in the review app when scheduled using the cron worker
#We can run the job anytime from the sidekiq console

image

image

image


lib/gias.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+module Gias
```
Nitemaeric, line 3: Nitpick, but can this be `GIAS` instead and we add `GIAS` to our acronyms list.

inulty-dfe: Ok, but I can't see any advantage ever to strictly using acronym cases in code

inulty-dfe: Decided against this as we already have `GiasSchool` model. No point in making it less consistent

lib/gias/importer.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    attr_reader :csv_path
+
+    def call
+      Log.log('Gias::Importer', 'Importing GIAS schools...')
+
+      school_records = CSV.foreach(csv_path, headers: true).map(&:to_h)
+
+      GiasSchool.upsert_all(school_records, unique_by: :urn)
```
tomas-stefano, line 18: I didn't understand this part.

* We expected a CSV, does the CSV will have only the new schools from a given timeframe?
* How many records we have as base to insert or update at one go? Should we check database locking?

lib/gias/downloader.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'net/http'
+
+module Gias
+  class Downloader < Service
+    def call
+      yesterday = Time.zone.yesterday.strftime('%Y%m%d')
+      filename = "edubasealldata#{yesterday}.csv"
+      path = 'https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public'
```
tomas-stefano, line 10: Can we move this to a constant or to Settings?

lib/gias/downloader.rb
```diff
@@ -0,0 +1,43 @@
...
+      url = URI("#{path}/#{filename}")
+      csv_path = 'tmp/gias_school.csv'
+
+      Log.log('Gias::Downloader', "Downloading the new gias file for #{Time.zone.yesterday}", level: :info)
+
+      begin
+        # Chunking the response reduces the memory usage ~70MB
+        Net::HTTP.start(url.host, url.port, use_ssl: true) do |https|
+          req = Net::HTTP::Get.new(url.path)
```
tomas-stefano, line 19: We have `Faraday`, `http` gem and the ruby standard library for HTTP calls. We need to start talking about using one.

inulty-dfe: My thinking here is that if this is ever extracted to a gem, there is no http library dependencies.

lib/gias/transformer.rb
```diff
@@ -0,0 +1,119 @@
...
+      def open?
+        OPEN_SCHOOL_CODES.include? row.fetch('EstablishmentStatus (code)')
+      end
+
+      def in_england?
+        NON_ENGLISH_ESTABLISHMENTS.exclude?(row.fetch('TypeOfEstablishment (code)'))
+      end
+    end
+  end
+end
```
tomas-stefano, line 119: The school_row method is doing too much—retrieving geographic coordinates using SQL, transforming data, and processing CSV rows. This could be split.

can we consider move this to a class?

```ruby
module Gias
  class CoordinateTransformService # or GridToCoordinatesService
    def initialize(easting, northing)
      @easting = easting
      @northing = northing
    end

    def call
      result = ActiveRecord::Base.connection.execute(format(<<~SQL.squish, @easting, @northing))
        WITH point as (
          SELECT ST_AsText(ST_Transform(ST_SetSRID(ST_MakePoint(%f, %f), 27700), 4327)) as text
        )
        SELECT ST_Y(text) as latitude, ST_X(text) as longitude from point;
      SQL

      {
        latitude: result.first['latitude'],
        longitude: result.first['longitude']
      }
    end
  end
end
```

And call from the school_row.

Can we also move to a another class the validations?

```ruby
module Gias
  class SchoolValidator
# Code | EstablishmentStatus
      # 1    | Open
      # 3    | Open, but proposed to close
      OPEN_SCHOOL_CODES = %w[1 3].freeze
      # Code | Establishment type
      # 25   | Offshore schools
      # 30   | Welsh establishment
      # 37   | British schools overseas
      NON_ENGLISH_ESTABLISHMENTS = %w[25 30 37].freeze

    def self.valid?(row)
      new(row).valid?
    end

    def initialize(row)
      @row = row
    end

    def valid?
      has_coordinates? && school_in_scope?
    end

    private

    attr_reader :row

    def has_coordinates?
      row['Northing'].present? && row['Easting'].present?
    end

    def school_in_scope?
    # ...
    end
  end
end
```

tomas-stefano (commented):
Can we have a ADR about this work? Or have any high level of the mechanism.

"Every 2:30 a.m ..." or diagrams, any documentation could help.

lib/gias/importer.rb
tomas-stefano: I read after is 2:30 a.m, so no need to concern with locking. yet

lib/gias/downloader.rb
inulty-dfe: Done

lib/gias/transformer.rb
```diff
@@ -0,0 +1,127 @@
...
+          )
+          SELECT ST_Y(text) as latitude, ST_X(text) as longitude from point;
+        SQL
+
+        [result.first['latitude'], result.first['longitude']]
+      end
+    end
+
+    class School
+      include ActiveModel::Validations
```
tomas-stefano, line 82: Since you're not using the `#errors` or similar we don't need this inclusion. Or am I missing something?

inulty-dfe: Ah yeah, I went that way, then changed my mind

PR #4883, opened 2025-02-03
Improve navigation - Update Browse primary courses page

gms-gs:
## Context
Prototype: https://find-review-4453.test.teacherservices.cloud/subjects?age_group=primary
Ticket: https://trello.com/c/lolKTP1k/307-improve-navigation-update-browse-primary-courses-page

The basis of this ticket is to pretty much change some content, however Dan, Tomas and I have agreed we want to simplify the code for this. So we have gone with a v2 line for this work which will rework what we currently have to become easier to maintain and extend.

This PR creates a `SubjectsController` which will be responsible for the primary and secondary subject rendering (Only primary for this ticket, subjects is in an upcoming ticket), I also added a `PrimarySubjectsForm` which just ensures an option is checked before processing to the new v2 results page.

- Update the `Browse primary courses` quick link to take the user to the new `v2/primary` route instead of the old route
- Update the content
- Add form validation

## Guidance to review
No errors:
Screenshot 2025-02-03 at 18 59 26

With errors: 
Screenshot 2025-02-03 at 19 00 03

### DIFF
--- a/app/controllers/find/v2/subjects_controller.rb
+++ b/app/controllers/find/v2/subjects_controller.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Find
+  module V2
+    class SubjectsController < Find::ApplicationController
+      def primary
+        @form = Find::V2::Subjects::PrimarySubjectsForm.new(subject_params)
+        @primary_subject_options = Subject.primary
+      end
+
+      def submit
+        @form = Find::V2::Subjects::PrimarySubjectsForm.new(subject_params)
+
+        if @form.valid?
+          redirect_to find_v2_results_path({ subjects: @form.subjects })
+        else
+          @primary_subject_options = Subject.primary
+          render :primary
+        end
+      end
+
+      private
+
+      def subject_params
+        params.fetch(:find_v2_subjects_primary_subjects_form, {}).permit(subjects: [])
+      end
+    end
+  end
+end
--- a/app/forms/find/v2/subjects/primary_subjects_form.rb
+++ b/app/forms/find/v2/subjects/primary_subjects_form.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Find
+  module V2
+    module Subjects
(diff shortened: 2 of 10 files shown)


kelliedesigner:
LGTM!

app/controllers/find/v2/subjects_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      private
+
+      def subject_params
+        params.fetch(:find_v2_subjects_primary_subjects_form, {}).permit(subjects: [])
+      end
+
+      def search_query_params
+        {
+          subjects: @form.subjects,
+          university_degree_status: true,
```
tomas-stefano, line 31: This is an old parameter for showing only postgraduate. We will only show postgraduate after search for primary?

tomas-stefano: But we won't have the questions "Yes, I have a degree or am studying for one" on the pre-filtering. Sorry my confusion. And this will filter for only postgraduate courses. Also this is the old parameter. If we want to have only postgraduate, you need to use the minimum degree required as 1:1. 

My question would be the reasoning behing setting those for the user when the user didn't request that particularly. 

If I am the user and ask to browse Primary courses I want to see all Primary courses regarless of degree type or visa?

Let's talk on Slack.

app/forms/find/v2/subjects/primary_subjects_form.rb
```diff
@@ -0,0 +1,25 @@
...
+        validates :subjects, presence: true
+
+        def initialize(params = {})
+          @subjects = Array(params[:subjects]).compact_blank
+        end
+
+        def save
+          return false unless valid?
+
+          true
```
tomas-stefano, line 20: Can we just use 

```ruby
def save
  valid?
end
````

tomas-stefano: Or alias :)

tomas-stefano (approved):
🚀

PR #4884, opened 2025-02-04
Add latitude and  longitude columns to GiasSchool

inulty-dfe:
## Context

Gias School records have no columns for geographical reference. We will add an import library to update the GiasSchool table periodically and we will store the latitude and longitude of the school in the process.

Add `:latitude` and `:longitude` `:float` columns to the `gias_school` table

## Guidance to review

 - Is `float` the right data type?
 - Must inform analytics about the new columns


Nitemaeric (approved):
`float` works well for now. We may want to consider using PostGIS `point` types for coordinates in the future.

PR #4888, opened 2025-02-04
Uses dfe-autocomplete npm and rubygem

tomas-stefano:
## Context

For the pre-filtering work we want to use the dfe-autocomplete gem.

In the past, the dfe-autocomplete code was copied and paste from the gem to the app, and modified slightly

This PR replaces in a non intrusive way - meaning is not replacing the ruby part because I don't wanna change the code that will be deleted after pre-filtering work.

The JS modification required is already in the dfe-autocomplete library so all is well in the JS part.

**The old copied dfe-autocomplete is only used on provider name search on Find homepage.**

1. Add the dfe-autocomplete

## Guidance to review

1. Does the provider name search on Find homepage works as equal to production?
2. All the other autocomplete are not affected?


inulty-dfe:
This is strange...

We're using the repo tag 0.1.0 but we're using javascript from HEAD and the javascript version is `version "0.0.1"`?

tomas-stefano:
> This is strange...
> 
> We're using the repo tag `v0.1.0` but we're using javascript from `HEAD` and the javascript version is `version "0.0.1"`?

There are only modifications to JS. 

We can launch a v0.2.0 with no ruby updates and only the JS and then update the gem and the node to use the v0.2.0.

Wdyt?

tomas-stefano:
Done

inulty-dfe (approved):

PR #4921, opened 2025-02-14
Add no results content to the new results page

tomas-stefano:
## Context

On the new results page, when there are no results we need to display the same content as we display in the live site.

TDA content is added.
General no results content is added
Devolved nations content is added

## Guidance to review

1. When search for places in Northern Ireland and no results, does the Northern Ireland content appears?
2. When search for places in Scotland and no results, does the Scotland content appears?
3. When search for places in Wales and no results, does the Wales content appears?
4. When search for TDA (No degree required option) and no results, does the TDA appears?


tomas-stefano:
https://trello.com/c/54MxQ35g/412-foundation-for-the-pre-filter-work-add-no-results-content-to-the-new-search-results-page

PR #4926, opened 2025-02-17
Fix issue where you search, click on a course and return to search

tomas-stefano:
## Context

This PR fixes the bug on the pre-filtering work that you can't return to the search after you view a course.

Steps to replicate (on main branch)

1. Visit /v2/results
2. Apply any filters
3. Click on any course.
4. Then click “Back to results”.

The search on step 2 should be there.

The reason the bug exists si that back to results relies on the params in the url and this solution is flakey.

**Solution:** I used the referrer if is a permitted referrer.

**Solution:** I used the referrer if is a permitted referrer.

## Guidance to review

1. Is it secure to use the permitted referrer instead relying on the parameters?

### DIFF
--- a/app/views/find/courses/show.html.erb
+++ b/app/views/find/courses/show.html.erb
@@ -1,14 +1,25 @@
 <%= content_for :page_title, "#{@course.name_and_code} with #{smart_quotes(@course.provider.provider_name)}" %>
 
-<% if permitted_referrer? %>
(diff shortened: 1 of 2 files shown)


tomas-stefano:
https://trello.com/c/IsYgWG9L/415-foundation-work-fix-the-back-to-results-link-in-the-course-view-page

PR #4933, opened 2025-02-18
WIP: Debug rake task

gms-gs:
When running this rake task in a review app the other day, I noticed that once run ALL subjects would all be called `Physical education' and seemed to only update the last subject in the object. 

**FYI, please download the latest sanitised backup from Azure, before running this locally as you will have issues**


tomas-stefano:
Closing the PR so I can have a review app from the same data as QA to test the script!

PR #4935, opened 2025-02-19
Add provider search to the results page form

tomas-stefano:
## Context

This PR adds the ability to search by provider name to the form in the results page.

The field includes an autocomplete that drops down with all the list of providers

Content
<details> Search by training provider

<label> Enter provider name

## Caveats

Initially I did via a remote call via JS, but it sounds too slow for a simple search, like very noticeable slowness on my local machine. 

So I optin to have the list available at the client but cache to not hit the database again for a hour.

The one hour expiration was arbitrary and I need to align with tech lead if this is a good expiration rule (or should be higher or lower).

Screenshot 2025-02-19 at 10 03 19
Screenshot 2025-02-19 at 10 04 41

## Guidance to review

1. It is okay the cache providers list? For a hour?
2. Does the autocomplete works?
3. Is the style ok on mobile?


app/javascript/find/controllers/remote_autocomplete_controller.js
```diff
@@ -0,0 +1,73 @@
...
+    return result && typeof result.name === 'string' ? result.name : ''
+  }
+
+  inputValueTemplate (result) {
+    return this.suggestionTemplate(result)
+  }
+
+  onConfirm (result) {
+  }
+}
```
tomas-stefano, line 73: This class is not neeeded for this work. I just organised because initially I was trying to make the providers a remote autocomplete and rollback my changes but I think is useful to have this class where we can inherit from it if we need extend the remote autocomplete. 

So I left here.

app/javascript/find/controllers/locations_autocomplete_controller.js
```diff
@@ -1,60 +1,4 @@
...
-  }
-
-  suggestionTemplate (result) {
-    if (typeof result === 'string') {
-      return result
-    }
-
-    return result && typeof result.name === 'string' ? result.name : ''
-  }
+export default class extends RemoteAutocompleteController {
```
tomas-stefano, line 3: Empty but just open for extension.

See https://github.com/DFE-Digital/publish-teacher-training/pull/4935/files#r1961372936

app/javascript/find/application.js
```diff
@@ -9,15 +9,23 @@ import initAutocomplete from './autocomplete'
...
+import LocationsAutocompleteController from './controllers/locations_autocomplete_controller'
 
 window.Stimulus = Application.start()
 Stimulus.register('filter-search', FilterSearchController)
 Stimulus.register('subjects-autocomplete', SubjectsAutocompleteController)
+Stimulus.register('remote-autocomplete', RemoteAutocompleteController)
 Stimulus.register('locations-autocomplete', LocationsAutocompleteController)
+
+// V1
 Stimulus.register('providers-autocomplete', ProvidersAutocompleteController)
```
tomas-stefano, line 24: This controller v1, will be deleted after the work is live and working

app/javascript/find/controllers/remote_autocomplete_controller.js
```diff
@@ -0,0 +1,73 @@
...
+        })
+      })
+        .then((data) => {
+          const suggestions = Array.isArray(data?.suggestions) ? data.suggestions : []
+          populateResults(suggestions)
+        })
+        .catch(() => {
+          populateResults([])
+        })
+    }, 900)
```
inulty-dfe, line 57: Is this why it feels really slow? 900 ms is a lot of debounce, no?

tomas-stefano: 900 is what the live site `app/javascript/find/autocomplete.js` is doing. I am not sure what the number would be the best.

The slowness was felt if using the API when using one word and then two, after three the number of suggestions diminises: https://www.publish-teacher-training-courses.service.gov.uk//api/public/v1/provider_suggestions?query=Oxf

tomas-stefano: Done, changed to 200ms

PR #4936, opened 2025-02-19
Improve tracking on the prefiltering work

tomas-stefano:
## Context

Now that primary and secondary pages are working:

* Add utm_source and utm_medium to primary and secondary courses

And in the case of course vieweing:

* If the user viewed a course and clicked "Back to results" we send an event to DfE analytics

* Add tracking from the course view to the results
* Add tracking for the primary form
* Add tracking for the secondary form

## Guidance to review

1. Visit homepage
2. Click on primary link and check a primary course and find primary courses. What is the utm source and utm medium?
3. View a course clicking from the search results and click back to search resuits. A DfE analytics event should be send with the right track params

### DIFF
--- a/app/services/find/analytics/search_results_event.rb
+++ b/app/services/find/analytics/search_results_event.rb
@@ -3,7 +3,8 @@
 module Find
   module Analytics
     class SearchResultsEvent < ApplicationEvent
-      attr_accessor :search_params, :track_params, :total, :page, :results
+      attr_accessor :search_params, :total, :page, :results
+      attr_writer :track_params
 
       def event_name
         :search_results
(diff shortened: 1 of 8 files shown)


app/services/find/analytics/search_results_event.rb
```diff
@@ -29,6 +30,33 @@ def visible_courses
...
+          @track_params
+        end
+      end
+
+      def course_view_referer?
+        course_view_referer.present?
+      end
+
+      def course_view_referer
+        uri = URI.parse(request.referer)
```
inulty-dfe, line 51: Better to just return nil if unless uri is present rather than rescuing?

app/services/find/analytics/search_results_event.rb
```diff
@@ -29,6 +30,33 @@ def visible_courses
...
+      DEFAULT_TRACK_PARAMS = { utm_source: 'results', utm_medium: 'no_referer' }.freeze
+
+      def track_params
+        return DEFAULT_TRACK_PARAMS if request.referer.blank?
+
+        if course_view_referer?
+          { utm_source: 'course', utm_medium: 'course_view' }.merge(course_view_referer)
+        else
+          @track_params
+        end
```
inulty-dfe, line 43: Are we at the point of this being a Concern yet?

app/services/find/analytics/search_results_event.rb
tomas-stefano: It is done on the method above on #event_data. 
The rescue is more in case there is any error on parsing the referer.

inulty-dfe (approved):

PR #4937, opened 2025-02-19
Td/add rake task to create subject groups

tomas-stefano:
## Context

We need to create the subject groups for subjects in the secondary page.

screencapture-find-localhost-secondary-2025-02-19-12_00_44

## Guidance to review

2. If you visit the homepage and browse secondary courses, can you see all groups?

### DIFF
--- a/app/models/subject.rb
+++ b/app/models/subject.rb
@@ -32,7 +32,10 @@ class Subject < ApplicationRecord
   end
 
   def self.secondary_subjects_with_subject_groups
-    secondary.where.not(subject_group: nil)
+    secondary
+      .where.not(subject_group: nil)
+      .includes(:subject_group)
+      .reorder('subject_group.created_at ASC, subject.subject_name ASC')
   end
 
   def secondary_subject?
--- a/lib/tasks/populate_subject_groups.rake
+++ b/lib/tasks/populate_subject_groups.rake
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+desc 'Create SubjectGroups and populate with subjects'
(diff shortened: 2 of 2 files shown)


inulty-dfe:
I don't like this implementation, but I have to admit anything more would be premature optimisation.

I think migrating an order column with unique index would be better.

Maybe go back an ask once more for confirmation on the ordering before going with this approach?

tomas-stefano:
> I don't like this implementation, but I have to admit anything more would be premature optimisation.
> 
> I think migrating an order column with unique index would be better.
> 
> Maybe go back an ask once more for confirmation on the ordering before going with this approach?

Makes sense. But for the bug party tomorrow I need this. I can run the code on QA directly without merging though.

Let's hold this PR but I will ran this rake task in QA tomorrow for the bug party only.

tomas-stefano:
> I don't like this implementation, but I have to admit anything more would be premature optimisation.
> 
> I think migrating an order column with unique index would be better.
> 
> Maybe go back an ask once more for confirmation on the ordering before going with this approach?

Not just that, the subjects inside the group also has an order.

Maybe we shouldn't solve through the database but now I have other cards to finish.

inulty-dfe:
> > I don't like this implementation, but I have to admit anything more would be premature optimisation.
> > I think migrating an order column with unique index would be better.
> > Maybe go back an ask once more for confirmation on the ordering before going with this approach?
> 
> Not just that, the subjects inside the group also has an order.
> 
> Maybe we shouldn't solve through the database but now I have other cards to finish.

Ok, We can always revisit this later.

inulty-dfe (approved):

PR #4955, opened 2025-02-24
Add Engineers teach physics filter

tomas-stefano:
## Context

Add a new filter (same as the live site).

The filter is only visible when user is searching for physics.

Screenshot 2025-02-24 at 18 21 02

## Guidance to review

1. Search for physics
2. Does the new filter appears?
3. Check the new filter.
4. Does the Engineers teach Physics course appear?
5. If you uncheck Physics. 
6. Does the filter still appear?

### DIFF
--- a/app/views/find/v2/results/filters/_all.html.erb
+++ b/app/views/find/v2/results/filters/_all.html.erb
@@ -1,0 +1,78 @@
+<div class="app-filter__content">
+  <%= form.govuk_submit "Apply filters", name: "utm_medium", value: "apply_filters_top" %>
+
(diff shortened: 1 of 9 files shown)


app/views/find/v2/results/filters/_all.html.erb
```diff
@@ -0,0 +1,78 @@
+<div class="app-filter__content">
```
tomas-stefano, line 1: I decided to create this partial because it was too big too follow on index. Let me know your opinions.

Nitemaeric (approved):
Just 1 question, code is all good :)

inulty-dfe (approved):

tomas-stefano:
> Any reason to not call this just _filters.html.erb?

Naming the file _all.html.erb explicitly communicates that this partial is responsible for rendering all filters, not just a subset. This clarity can help developers understand the role of the partial at a glance.
We don't have any subset of filters and I don't think it will be, but I also thinks about resource/index or resource/all or resource/show when thinking in terms or organising partials.

PR #4960, opened 2025-02-25
Watch find and publish javascript in Procfile.dev

inulty-dfe:
## Context

bin/dev that runs the Procfile.dev doesn't reload publish javascript bundle in dev because the package.json script calls two commands joined by && and only watches the last one run (find)

  The previous command only watched the last command in the package
  script (find)

      yarn run build:js:publish && yarn run build:js:find

  We need to watch the builds separately to have them both reload

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,4 +1,5 @@
-js: yarn build:js --watch
+js-find: yarn build:js:find --watch
+js-publish: yarn build:js:publish --watch
 css: yarn build:css --watch
 web: bin/rails server -p 3001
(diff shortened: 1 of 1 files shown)


Procfile.dev
```diff
@@ -1,4 +1,5 @@
-js: yarn build:js --watch
+find-js: yarn build:js:find --watch
+publish-js: yarn build:js:publish --watch
```
Nitemaeric, line 2: Should we swap these to `js-find` and `js-publish` to go from most common to least common?

It also aligns with the script names.

Nitemaeric (approved):
No blockers. 1 suggestion :)

PR #4962, opened 2025-02-25
Don't show filters on mobile on page load

Nitemaeric:
## Context

The filters on mobile

- Default the filters to hidden to avoid any flickering
- Use CSS and a Stimulus "visibility" controller to control showing and hiding of a component via CSS classes

## Guidance to review

- Visit the results page on mobile
  - The filters modal should be hidden
- Visit the results page on desktop
  - The filters section should be visible
- Visit both with JavaScript disabled, it should show the filters above the results without the "close" and "Show filters" buttons.


app/views/find/v2/results/index.html.erb
```diff
@@ -16,7 +16,7 @@
   <%= form.hidden_field :utm_source, value: "results" %>
 
   <div class="app-filter-layout">
-    <div class="app-filter-layout__filter app-filter">
+    <div class="app-filter-layout__filter app-filter" hidden="hidden">
```
tomas-stefano, line 19: Tested and on mobile everything worked.

Although for non JS users the filter disappears because of this hidden attribute. Is there a way to add this attributes on the JS instead of hardcoding on document.load?

inulty-dfe:
Now's not the time, but this could have been implemented in CSS :smile:

app/views/find/v2/results/index.html.erb
Nitemaeric: Hmm, true...

It really just needs to be hidden for mobile. I might need to mess with the CSS.

tomas-stefano (approved):
It worked for me when I tested, although should we test on the "old live site" too? Or it doesn't affect?

Nitemaeric:
> It worked for me when I tested, although should we test on the "old live site" too? Or it doesn't affect? 

I added the changes for the current (old) site too.

PR #4965, opened 2025-02-26
Prefiltering work - Optimize the /results page 

tomas-stefano:
## Context

This PR optimizes the homepage and /results page by implementing pre-filtering improvements. The goal was to reduce database queries and enhance overall performance.

## Disclaimer

Due to time constraints, a comprehensive load test wasn't performed, and the current load test script wasn't fully optimized. I had to rely on the best approach available within the given time frame for analysis and optimization. All testing was done locally on my machine.

## TL; DR

Significant optimizations were made to the search functionality on Find, making searches faster and more efficient.

**Location-based searches** are now 60% faster (500 requests went from 218s → 86s).
**Non-location searches** have improved by 61% (500 requests went from 42s → 16s).

This means users will get quicker search results, making it easier to find courses. 🚀 (with the Google and location caveat)

But there is a worst case scenario: MANY different concurrent location searches at the same time. A low risk but still exists. 

## Comparison script

To simulate hundreds of requests, I created a basic script. This is not a proper load test but provides some insight into whether the optimizations are effective or not.

```ruby
# frozen_string_literal: true

require 'net/http'
require 'benchmark'
require 'uri'

# Non location searches
BASE_URL = 'https://find.localhost/results?utm_source=results&can_sponsor_visa=true&subjects%5B%5D=00&subjects%5B%5D=01&subjects%5B%5D=02&applications_open=true&subject_name=&subject_code=&location=&radius=10&provider_name=&provider_code=&utm_medium=search&order='
#BASE_URL = 'https://find.localhost/results?age_group=primary&applications_open=true&can_sponsor_visa=true&has_vacancies=true&l=2&subjects%5B%5D=00&subjects%5B%5D=01&subjects%5B%5D=02&university_degree_status=true&visa_status=true'

TOTAL_REQUESTS = 500
CONCURRENCY = 4

queue = Queue.new
TOTAL_REQUESTS.times { |i| queue << URI("#{BASE_URL}&page=#{i + 1}") }

success_count = 0
redirect_count = 0
error_count = 0
mutex = Mutex.new

time = Benchmark.realtime do
  threads = Array.new(CONCURRENCY) do
    Thread.new do
      until queue.empty?
        url = begin
          queue.pop(true)
        rescue StandardError
          nil
        end
        next unless url

        begin
          response = Net::HTTP.get_response(url)
          status = response.code.to_i

          mutex.synchronize do
            if status == 200
              success_count += 1
            else
              error_count += 1
            end
          end

          puts "Request to #{url} => Status: #{status}"
        rescue StandardError => e
          mutex.synchronize { error_count += 1 }
          puts "Request failed: #{e.message}"
        end
      end
    end
  end
  threads.each(&:join)
end

puts "\nTotal time: #{time.round(2)}s"
puts "200 responses: #{success_count}"
puts "Errors: #{error_count}"
```

For plot the graph I used GNU plot and create a script to check. For usage saved this to a file and when you wanna see the graph just CTRL + C to see the GNU plot.

```
brew install gnuplot
```

```
gem install gnuplot
```

```ruby
require 'gnuplot'

class ProcessUsage
  def self.get_usage
    rails_pid = `ps aux | grep '[p]uma' | awk '{print $2}'`.strip
    postgres_pid = `pgrep -n postgres`.strip

    return nil if rails_pid.empty? || postgres_pid.empty?

    rails_cpu = `ps -p #{rails_pid} -o %cpu=`.strip.to_f
    postgres_cpu = `ps -p #{postgres_pid} -o %cpu=`.strip.to_f
    rails_mem_percent = `ps -p #{rails_pid} -o %mem=`.strip.to_f
    postgres_mem_percent = `ps -p #{postgres_pid} -o %mem=`.strip.to_f

    total_memory = `sysctl -n hw.memsize`.strip.to_i / (1024 * 1024)  # Memory in MB

    rails_mem_mb = (rails_mem_percent / 100.0) * total_memory
    postgres_mem_mb = (postgres_mem_percent / 100.0) * total_memory

    {
      timestamp: Time.now.to_i,
      rails_cpu: rails_cpu,
      postgres_cpu: postgres_cpu,
      rails_mem: rails_mem_mb,
      postgres_mem: postgres_mem_mb
    }
  end
end

class CPUGraphPlotter
  def initialize
    @timestamps = []
    @rails_cpu = []
    @postgres_cpu = []
  end

  def plot(data)
    data.each do |entry|
      @timestamps << entry[:timestamp]
      @rails_cpu << entry[:rails_cpu]
      @postgres_cpu << entry[:postgres_cpu]
    end

    Gnuplot.open do |gp|
      Gnuplot::Plot.new(gp) do |plot|
        plot.title "Rails (Puma) & PostgreSQL CPU Usage"
        plot.xlabel "Time (seconds)"
        plot.ylabel "CPU Usage (%)"
        plot.grid
        plot.data << Gnuplot::DataSet.new([@timestamps, @rails_cpu]) do |ds|
          ds.with = "lines"
          ds.title = "Rails (Puma) CPU (%)"
          ds.linewidth = 2
        end
        plot.data << Gnuplot::DataSet.new([@timestamps, @postgres_cpu]) do |ds|
          ds.with = "lines"
          ds.title = "Postgres CPU (%)"
          ds.linewidth = 2
        end
      end
    end
  end
end

class MemGraphPlotter
  def initialize
    @timestamps = []
    @rails_mem = []
    @postgres_mem = []
  end

  def plot(data)
    data.each do |entry|
      @timestamps << entry[:timestamp]
      @rails_mem << entry[:rails_mem]
      @postgres_mem << entry[:postgres_mem]
    end

    Gnuplot.open do |gp|
      Gnuplot::Plot.new(gp) do |plot|
        plot.title "Rails (Puma) & PostgreSQL Memory Usage"
        plot.xlabel "Time (seconds)"
        plot.ylabel "Memory Usage (MB)"
        plot.grid
        plot.data << Gnuplot::DataSet.new([@timestamps, @rails_mem]) do |ds|
          ds.with = "lines"
          ds.title = "Rails (Puma) Memory (MB)"
          ds.linewidth = 2
        end
        plot.data << Gnuplot::DataSet.new([@timestamps, @postgres_mem]) do |ds|
          ds.with = "lines"
          ds.title = "Postgres Memory (MB)"
          ds.linewidth = 2
        end
      end
    end
  end
end

class Monitor
  def initialize
    @data = []
    @cpu_plotter = CPUGraphPlotter.new
    @mem_plotter = MemGraphPlotter.new
  end

  def run
    Signal.trap("INT") {
      puts "\nExiting monitor..."
      @cpu_plotter.plot(@data)  # Plot CPU graph
      @mem_plotter.plot(@data)  # Plot Memory graph
      exit
    }

    loop do
      usage = ProcessUsage.get_usage
      if usage
        @data.push(usage)
      else
        puts "Error: Could not find Puma or Postgres process."
      end
      sleep 1
    end
  end
end

monitor = Monitor.new
monitor.run
```

## Comparison with the live site (old version)

The live site performs approximately 96 queries per request. After pre-filtering, the new version performs only 9 queries per request (12 the first time, then 9 after caching).

Non-location searches:

Old site (500 requests): 
```
 https://find.localhost/results?utm_source=results&can_sponsor_visa=true&subjects%5B%5D=00&subjects%5B%5D=01&subjects%5B%5D=02&applications_open=true&subject_name=&subject_code=&location=&radius=10&provider_name=&provider_code=&utm_medium=search&order=&page=#{number-of-the-page}
```

```
Total time: 42.42s
200 responses: 500
Errors: 0
New site (500 requests):
```

Prefiltering after optimisation (500 requests):

```
Total time: 16.32s
200 responses: 500
Errors: 0
```

## GNU plot results

I monitored CPU and memory usage during tests using the ruby monitor.rb and ruby requests.rb scripts. Below are the results:

### For the old live site

CPU Usage (Rails App): 40-90%
CPU Usage (Postgres): 0-60%
Memory Usage (Rails App): up to 700MB
Memory Usage (Postgres): up to 200MB

old-live-site-cpu-usage

old-live-site-memory-usage

### Prefiltering BEFORE the optmisations

CPU Usage (Rails App): 0-90%, average 50% - database almost similar 
Memory Usage (Rails App): up to 500MB
Memory Usage (Postgres): up to 200MB

prefiltering-main-cpu-usage

prefiltering-main-memory-usage

### Prefiltering AFTER the optmisations

CPU Usage (Rails App): 0-90%, average reduced to 50% due to caching and fewer DB queries.
Memory Usage (Rails App): stabilized at 500MB
Memory Usage (Postgres): peak at 200MB, less frequent than before

prefiltering-optmisation-cpu-usage

prefiltering-optimisation-memory-usage

## Location search

Note: The location search on the first query may take slightly longer than the live site because the latitude and longitude calculation occurs after answering the first question. Once the /results page is loaded, the result is cached for 30 days.

The worst case scenario: we only have different location search on every request.

Because of this a pre-caution would be increasing the pods? I wanna hear your views.

### Location search graphs

The first request for location takes an average 600ms to 1000ms (because is asking Google the latitude and longitude) then the results are cached.

I did this monitoring to check the CPU usage of the Postgis calculation for geolocation data.

## Old site Location performance (live site)

500 requests to: `https://find.localhost/results?c=England&l=1&latitude=53.4807593&loc=Manchester%2C+UK&longitude=-2.2426305&lq=Manchester&radius=10&sortby=distance&page=#{page-number}`

```
Total time: 218.36s
200 responses: 500
Errors: 0
```

live-site-location-memory-usage

live-site-location-cpu-usage

## Prefiltering Location performance before optimisations

500 requests to: `https://find.localhost/results?utm_source=results&subject_name=&subject_code=&location=Manchester%2C+UK&radius=10&provider_name=&provider_code=&utm_medium=search&page=#{page-number}`

```
Total time: 86.62s
200 responses: 500
Errors: 0
```

prefiltering-before-optimisation-location-memory-usage
prefiltering-before-optimisation-location-cpu-usage

## Prefiltering Location performance after optimisations

500 requests to: `https://find.localhost/results?utm_source=results&subject_name=&subject_code=&location=Manchester%2C+UK&radius=10&provider_name=&provider_code=&utm_medium=search&page=#{page-number}`

```
Total time: 82.15s
200 responses: 500
Errors: 0
```

prefiltering-after-optimisation-location-memory-usage
prefiltering-after-optimisation-location-cpu-usage

## Guidance to review

1. Did I miss anything?
2. What about locations takes longer the first time? Do we need to increase the pods? Have a warm up the cache? Any ideas for the future?

## What do you think?

1. Do we need to increase pods size?
2. Do we need to increase memory size? The memory seems good after optmisations so everything looks good on Find side.


app/models/course.rb
```diff
@@ -204,7 +193,7 @@ def new_draft_attributes
     end
   end
 
-  has_one :latest_published_enrichment, -> { published.order('created_at DESC, id DESC').limit(1) },
+  has_one :latest_published_enrichment, -> { published.most_recent },
```
tomas-stefano, line 196: I did to compare and is the same as the one removed (but with the advantage of calling #preload or #includes or #eager_load)

```ruby
result = Courses::Query.new(params: {}).scope.select { |c| c.latest_published_enrichment != c.old_latest_published_enrichment }; puts result.size; puts result.first
0
=> nil
```

app/models/course.rb
```diff
@@ -27,17 +27,6 @@ class Course < ApplicationRecord
   audited
 
   attribute :subordinate_subject_id, :integer
-  after_initialize do
```
tomas-stefano, line 30: Everytime a Course was instantiate we were doing more queries to the DB so this was necessary to be moved. I moved the methods more below and didn't a way that all the specs passed.

app/models/course.rb
```diff
@@ -790,6 +779,29 @@ def withdraw
...
+    subjects_list.first&.subject_id
+  end
+
+  def assign_positioned_secondary_subject!(subjects_list:)
+    subjects_list.second&.subject&.id
+  end
+
+  def subordinate_subject_id
+    super || (is_primary? ? nil : assign_positioned_secondary_subject!(subjects_list: course_subjects.select(&:position)))
+  end
```
tomas-stefano, line 803: Something for the future - we need to change the wording "master" / "subordinate". I won't change to not loose focus on this PR but I think we should in the future

app/models/course.rb
```diff
@@ -905,32 +917,8 @@ def name_and_code
...
-                                AND site.address4 IN ('', NULL)
-                                AND site.postcode IN ('', NULL)
-                                AND site.latitude IS NULL
-                                AND site.longitude IS NULL"
-                              )
-  end
-
   def available_placements_count
-    @available_placements_count ||= available_placements.count
+    @available_placements_count ||= site_statuses.size
```
tomas-stefano, line 921: I verified and all the site statuses are similar to the query that was doing before. I had to call `.size` because this is preloaded from Active Record. If we use `.count` will be N+1 query on the search results.

app/components/courses/summary_card_component.rb
```diff
@@ -208,7 +207,15 @@ def financial_incentive
     end
 
     def main_subject
-      @main_subject ||= Subject.find_by(id: course.master_subject_id)
+      @main_subject ||= course.subjects.find { |subject| subject.id == course.master_subject_id }
+    end
+
+    # rubocop:disable Lint/UselessConstantScoping
+    NullEnrichment = Struct.new(:course_length, :fee_uk_eu, :fee_international, keyword_init: true)
```
tomas-stefano, line 214: I optin for the Null object approach. 

I will card up something to improve the tests and factories and replace this.

inulty-dfe (approved):
LGTM

Nitemaeric (approved):

app/services/courses/query.rb
```diff
@@ -53,6 +54,19 @@ def call
...
+      @scope.preload(
+        :site_statuses,
+        :latest_published_enrichment,
+        :provider,
+        subjects: [:financial_incentive]
+      )
+    end
+
+    def total
+      @scope.unscope(:order, :group).distinct.count(:id)
```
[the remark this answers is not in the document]
Nitemaeric: https://www.youtube.com/watch?v=H6juZ8c_Nu8

app/controllers/find/v2/results_controller.rb
```diff
@@ -12,9 +12,13 @@ def index
         @search_params = @search_courses_form.search_params
         @courses_query = ::Courses::Query.new(params: @search_params.dup)
         @courses = @courses_query.call
-        @courses_count = @courses.unscope(:order, :group).distinct.count(:id)
+        @courses_count = @courses_query.total
```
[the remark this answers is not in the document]
tomas-stefano: Renamed

app/services/courses/query.rb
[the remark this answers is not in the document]
tomas-stefano: Yes I agree but in order to apply the scopes and use count with distinct from Rails, Rails rewrites the select statement and because of distinct the count is done with the id. 

Maybe I could experiment

```
Course.where(id: @scope.unscope(:order, :group, :select).select(:id).distinct).count
```

This can sometimes be faster than COUNT(DISTINCT course.id)

Good idea, although I think for now I will go with the recent optimisations but we can opimitise the query even more defintely.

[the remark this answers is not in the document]
Nitemaeric: Good point.

Yeah, I think `count(:id)` is sufficient. No need to wrap.

PR #4972, opened 2025-02-28
Set search radius of 50 on large geocoded areas

inulty-dfe:
## Context

  When a user searches for Cornwall, we want to widen the default radius
  from 10 to 50 miles, otherwise many training locations in Cornwall are
  not returned in the search results

  We identify a large area by the presence of
  "administrative_area_level_2" in the types property returned by Google
  Geocoder API:

  ```
      "types": [
        "administrative_area_level_2",
        "political"
      ]
  ```

If a course search has been made relative to a location, if the location is a large area - according to Google - we increase the search radius to 50 miles from the default 10 miles. If an explicit radius is provided, that radius is used to filter the results.

## Guidance to review

https://github.com/user-attachments/assets/a822d676-b53e-4c92-bba5-49ed47dc9fd1

## Trello 
https://trello.com/c/huzf9Dlw/484-bug-searching-for-cornwall-the-default-radius-is-10miles-should-it-be-50

### DIFF
--- a/app/forms/courses/search_form.rb
+++ b/app/forms/courses/search_form.rb
@@ -116,6 +116,7 @@ module Courses
     RadiusOption = Struct.new(:value, :name, keyword_init: true)
     RADIUS_VALUES = [1, 5, 10, 15, 20, 25, 50, 100, 200].freeze
     DEFAULT_RADIUS = 10
+    LARGE_RADIUS = 50
 
     def radius_options
       RADIUS_VALUES.map do |value|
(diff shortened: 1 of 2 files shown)


Nitemaeric:
This looks to work well with counties. Nice.

Should we also default larger areas, like below, to 50 miles?

CleanShot 2025-02-28 at 15 20 32@2x

inulty-dfe:
> This looks to work well with counties. Nice.
> 
> Should we also default larger areas, like below, to 50 miles?
> 
> CleanShot 2025-02-28 at 15 22 03@2x

Who even knows where South East centroid is?

I'd put that in another ticket as it requires investigation

Nitemaeric (approved):

PR #4973, opened 2025-02-28
Show only open courses by default on Find searches

tomas-stefano:
## Context

Same as the live site, the applications open filter should be on by default.

## Guidance to review

1. Click in all homepage links (include the main search).
2. Does the applications open filter is on by default?

### DIFF
--- a/app/controllers/find/v2/primary_subjects_controller.rb
+++ b/app/controllers/find/v2/primary_subjects_controller.rb
@@ -13,7 +13,7 @@ module Find
 
       def submit
         if @form.valid?
-          redirect_to find_results_path({ subjects: @form.subjects }.merge(track_params))
+          redirect_to find_results_path({ subjects: @form.subjects, applications_open: true }.merge(track_params))
         else
           @primary_subject_options = Subject.primary
           render :index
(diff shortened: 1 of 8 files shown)


Nitemaeric (commented):
Hmm, this feels like we're doing a little too much to me.

Instead of sending the `applications_open: true` query param, could we default the value in the form object?

So we'd have:

```ruby
class Form
  def applications_open
    params[:applications_open] || true
  end
end
```

This way, the checkbox will default to checked when no params are sent?

Or is this something we're trying to prevent?

Nitemaeric (approved):
My line of thought is do whether we _always_ want to default to "Open" or just some specific links. If it is always, I think the approach I've suggested might be better, we can check for the string "false" to show all courses.

That being said, not a blocker.

tomas-stefano:
> My line of thought is do whether we _always_ want to default to "Open" or just some specific links. If it is always, I think the approach I've suggested might be better, we can check for the string "false" to show all courses.
> 
> That being said, not a blocker.

But that's the catch. there would never be string "false". If was I'd follow your approach as it is the simple solution.

PR #4974, opened 2025-02-28
Show filter button when there are no results

Nitemaeric:
## Context

When a user filters results and there are none left, the sort and filter buttons disappear. We have no need to change the sort order of 0 results, but we may want to change filters to get some results.

- Add the filter button to the no results condition

## Guidance to review

- Try searching for something with no results, e.g. Ancient Greek.
- You should see the filter button.


inulty-dfe (approved):
Don't like the duplication of the button tag but deduping isn't worth the hassle

PR #4978, opened 2025-02-28
Make sure all old parameters work with the prefiltering work

tomas-stefano:
## Context

After went through a CSV from Logit to get all parameters from the prod requests I saw some bugs around backwards compatibility:

* qualification is now qualifications
* study_type is now study_types
* lq is now location

Also there are values conversion for more meaninful values

## Guidance to review

1. Visit the live site and go to the results page
2. Check all filters
3. Copy and paste the url (without the domain) to the review app and see the filters applied.
4. Do the same for search by provider
5. Do the same for search by location


inulty-dfe:
Is it worth grouping all the old parameter tests together so they're easier to delete when we deprecate them?

PR #4989, opened 2025-03-04
Default radius if radius parameter can not be parsed

tomas-stefano:
## Context

Users bookmarked bad formatted URLs from the old live site.

With the prefiltering this won't happen anymore but if the user bookmarked their searches we need to make sure we don't break the prefiltering results.

Default radius if radius can't be parsed.

## Guidance to review

1. Attempt to visit a bad formatted radius with degrees on it `radius=10ºdegree_required=show_all_courses` 
2. Radius should use the default

### DIFF
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -226,7 +226,6 @@ module Courses
     def location_scope
       return @scope.distinct if params[:latitude].blank? || params[:longitude].blank?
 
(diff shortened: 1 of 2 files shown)


tomas-stefano:
https://trello.com/c/JDiZzVJr/297-find-results-param-degreerequired-is-being-parsed-as-20%E2%81%B0reerequired

Nitemaeric (approved):

PR #4990, opened 2025-03-04
Return the number of pods pre filtering launch

tomas-stefano:
## Context

The graphs are almost unchanged so we can safely return to 4 pods.

Nonetheles it is always good to be cautious in a launch of a redesign of the whole site and its search mechanisms. 

## Guidance to review

1. See Azure DB CPU percentage and mem used
2. See Redis cache size
3. See Grafana metrics
4. Are we good to return to 4 pods?


RMcVelia (approved):
couldn't see any reason not to decrease from the stats

Nitemaeric (approved):

PR #4997, opened 2025-03-05
Update fee content

gms-gs:
## Context

The content on the search results and course summary pages is not clear whether a course is fee or salary.

Update the content to match the designs here: https://trello.com/c/Jq22IvMJ/533-bug-fee-content-has-not-been-updated

## Guidance to review

- View a fee based course on the results page
- View that course and check the content matches

Screenshot 2025-03-05 at 13 02 59

Screenshot 2025-03-05 at 13 03 06

### DIFF
--- a/config/locales/find.yml
+++ b/config/locales/find.yml
@@ -28,8 +28,8 @@ en:
         salary: Salary
         apprenticeship: Salary (apprenticeship)
         fee:
-          uk_fees_html: "%{value} for UK citizens"
-          international_fees_html: "%{value} for Non-UK citizens"
+          uk_fees_html: "%{value} <b>fee</b> for UK citizens"
+          international_fees_html: "%{value} <b>fee</b> for Non-UK citizens"
           hint:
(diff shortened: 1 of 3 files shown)


config/locales/find.yml
```diff
@@ -28,8 +28,8 @@ en:
         salary: Salary
         apprenticeship: Salary (apprenticeship)
         fee:
-          uk_fees_html: "%{value} for UK citizens"
-          international_fees_html: "%{value} for Non-UK citizens"
+          uk_fees_html: "Fee - UK citizens %{value} "
+          international_fees_html: "Fee - Non-UK citizens %{value}"
```
Nitemaeric, line 32: This is the outdated solution 😅 It should be "£4,500 fee for UK citizens" as mentioned in the latest comment, [here](https://trello.com/c/Jq22IvMJ#comment-67c6c69fe0f23d91516fed34).

Nitemaeric (changes requested):

Nitemaeric (approved):
Code-wise, all good. A shame that the `value` and the `"fee"` are wrapped in separate `<b>` tags, but not a blocker.

If Product sign off, let's merge then.

PR #5035, opened 2025-03-11
change title of school placements

joegibb:
## Context

We continue to receive feedback from providers that the new changes to how we communicate location have resulted in confusion or inaccuracy.

One thing that needs to be improved is the content on the school placements page, to make it clearer that the schools have been attached to the course and not the provider.


joegibb:
keeps failing so restarting

PR #5036, opened 2025-03-11
change title on school placements page

joegibb:
## Context

We continue to receive feedback from providers that the new changes to how we communicate location have resulted in confusion or inaccuracy.

One thing that needs to be improved is the content on the school placements page, to make it clearer that the schools have been attached to the course and not the provider.


app/views/shared/courses/_placements.html.erb
```diff
@@ -1,5 +1,8 @@
 <h1 class="govuk-heading-xl">
-  <%= t(".heading", provider_name: course.provider_name) %>
+  <span class="govuk-caption-xl">
+    <%= t(".thing", provider_name: course.provider_name) %>
```
tomas-stefano, line 3: Can we rename to `.caption`, please? :)

PR #5037, opened 2025-03-11
Add missing Skip to main content link across Find

gms-gs:
## Context

The comply with accessibility standards, Find needs to have ‘Skip to main content’ links at the top of each page, which become visible using keyboard navigation, and on selection jump the user down to the page content and H1.

This link is missing on all pages of Find. This ticket is to add the link.

Following https://govuk-components.netlify.app/helpers/skip-link/ added the new component

## Guidance to review

Visit any find page hit TAB and then you will see the skip to main content link at the top of the page. If selected it will take you to the main content of that page

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,9 +28,9 @@
   </head>
 
   <body class="govuk-template__body <%= yield :body_class %>">
-    <%= render "layouts/add_js_enabled_class_to_body" %>
-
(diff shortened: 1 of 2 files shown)


Nitemaeric:
@gms-gs This might be out of scope, but it's curious how the "Skip content" for Publish is not the first element focused when we press "Tab". Do you think we should align both layouts?

Nitemaeric (approved):

PR #5048, opened 2025-03-13
Add rake task to create provider partnerships from Accredited enrichment

inulty-dfe:
## Context

  Training providers and Accredited providers have relationships defined in the `accrediting_provider_enrichments` column on the Training provider record.

  We need to recreate these relationships in the ProviderPartnership model before we release ProviderPartnerships.

  We have a data migration that does the same thing but we would like to
  be able to run this idempotent task as we approach the launch of
  provider partnerships

  This will help reduce te results of this query

  https://www.publish-teacher-training-courses.service.gov.uk/blazer/queries/105-training-providers-with-published-courses-and-no-partnership

Add a rake task to update the provider partnerships from enrichments

## Guidance to review

I ran this on QA

```
57 partnerships created!
```
The number of training providers with published courses and no partnership went from 49 to 0.

## Trello

https://trello.com/c/KjHGUvmB/563-add-rake-task-to-update-providerpartnerships-in-console


spec/lib/tasks/create_provider_partnerships_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'rake'
+
+describe 'publish_give_course' do
```
tomas-stefano, line 6:
```suggestion
describe 'creates provider partnerships' do
```
Or similar.

spec/lib/tasks/create_provider_partnerships_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  subject do
+    Rake::Task['provider_partnerships:create_from_enrichments'].invoke
```
tomas-stefano, line 8: Usually if this requires tests, I think it is better to move to a class.

spec/lib/tasks/create_provider_partnerships_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  let!(:accredited_provider) { create(:accredited_provider) }
+  let!(:description) { 'The Description' }
+  let!(:training_provider) { create(:provider, accrediting_provider_enrichments: [AccreditingProviderEnrichment.new(UcasProviderCode: accredited_provider.provider_code, Description: description)]) }
+
+  Rails.application.load_tasks if Rake::Task.tasks.empty?
+
+  it 'calls Courses::PublishService service' do
+    expect { subject }.to change(training_provider.accredited_partnerships, :count).by(1)
+  end
```
Nitemaeric, line 19: The description and the test don't match. It tests if partnerships of `training_provider` are increased by `1`.

I'd check if the accredited provider matches in the test too.

Nitemaeric (commented):
I agree with Tomas. This looks better off as a Service that the task calls.

spec/lib/tasks/create_provider_partnerships_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    Rake::Task['provider_partnerships:create_from_enrichments'].invoke
+  end
+
+  let!(:accredited_provider) { create(:accredited_provider) }
+  let!(:description) { 'The Description' }
+  let!(:training_provider) { create(:provider, accrediting_provider_enrichments: [AccreditingProviderEnrichment.new(UcasProviderCode: accredited_provider.provider_code, Description: description)]) }
+
+  Rails.application.load_tasks if Rake::Task.tasks.empty?
+
+  it 'calls Courses::PublishService service' do
```
Nitemaeric, line 17: Could we surface the expected behaviour in the spec description here, please?

```suggestion
  it 'creates accredited partnerships for training partners from their enrichments' do
```

spec/lib/tasks/create_provider_partnerships_spec.rb
inulty-dfe: Yeah we expect this to be removed in the coming week or two

Nitemaeric (approved):

tomas-stefano (approved):

PR #5050, opened 2025-03-14
changed content from courses are currently closed

joegibb:
## Context

TDA providers want to publish courses that are searchable on Find, but that cannot be applied to. They want to gauge interest in candidate demand without having to deal with applications.

As a result, they want the content returned for these courses, modified, to include a recommendation for candidates to contact the provider off-service via email.


inulty-dfe:
Do you think we want to put this text into the translation files?

Nitemaeric (approved):

PR #5062, opened 2025-03-18
Update Add multiple schools in Support interface

inulty-dfe:
## Context

Update "Add multiple schools" to provider in the Support console.

We are making this change to ensure that any schools added by the system to a provider are geneuine schools registered with GIAS.

- Remove the Add school form where custom attributes can be set on a providers schools
- All new schools added will match exactly the attributes of the school registered with GIAS

## Guidance to review

- Cannot submit with no URNs
- Cannot submit with more than 50 URNs
- Any values that don't match a GIAS URN are shown to not match any schools
- Any values that match URNs of schools the provider has already added are shown
- If no values correspond to a school that can be added, a button will take the user back to the Enter URNs page.

`Remove school` is a link to a GET endpoint. Is this normal?

## Trello
https://trello.com/c/NVx79IQE/539-update-add-multiple-schools-flow-for-support-interface


app/controllers/support/providers/schools/check_multiple_controller.rb
```diff
@@ -7,39 +7,60 @@ class CheckMultipleController < ApplicationController
...
+
+            @gias_schools = GiasSchool.where(urn: urn_form.fields[:values]).where.not(urn: existing_urns)
+
+            @unfound_urns = param_urns - @gias_schools.map(&:urn)
 
-        def school_details
-          @school_details ||= parsed_csv_school_form.school_details.map { |s| Site.new(s) }
+            @gias_schools.map do |school|
+              provider.sites.build(school.school_attributes)
+            end
```
tomas-stefano, line 62: I think this might be important to move to a class and do one unit test. Wdyt?

inulty-dfe: Yeah I think that's fair. I'm not sure where to put it. I expect it will be needed in the Publish module as well.

tomas-stefano (approved):

PR #5067, opened 2025-03-20
Do not show modern languages subject on autocomplete

tomas-stefano:
## Context

Modern languages is a "virtual" subject on our data that represents more like a subject group rather than one specific subject.

So removing from the autocomplete for now.

This changes the cache so we could:

1. Wait for the cache to expiry
2. Run `Rails.cache.delete('subjects:all')` to refresh the cache

## Guidance to review

1. Modern languages should not appear in the autocomplete when searching for a subject.

    - Visit review app
    - Search "Mode" on Subject. 
    - Only "Modern languages (other)" should appear.
    - Visit the results page
    -  Search "Mode" on Subject. 
    - Only "Modern languages (other)" should appear.

### DIFF
--- a/app/services/subjects_cache.rb
+++ b/app/services/subjects_cache.rb
@@ -31,7 +31,7 @@ class SubjectsCache
 
   def all_subjects
     Rails.cache.fetch('subjects:all', expires_in: expires_in) do
-      Subject.active.map do |subject|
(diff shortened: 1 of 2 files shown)


olioldfield (approved):
Looks good to me - thanks!

PR #5078, opened 2025-03-24
Fix support accredited provider autocomplete

inulty-dfe:
## Context

  The accredited provider search api requires the recruitment cycle. We
  now get the appropriate recruitment cycle year from the window.location (the url)

Get the required recruitment cycle year from the url to search the accredited providers

https://publish-review-5078.test.teacherservices.cloud/support/2025/providers/20584/accredited-providers/search

## Guidance to review

Is there a better way to get the recruitment cycle year?
Is the implementation sound?

## Trello

https://trello.com/c/mvWmywc7/588-autocomplete-is-not-working-for-provider-search

### DIFF
--- a/app/javascript/publish/autocomplete/accredited_provider.js
+++ b/app/javascript/publish/autocomplete/accredited_provider.js
@@ -5,8 +5,10 @@ const providerSuggestionTemplate = (result) => result && `${result.provider_name
 const onConfirm = (input) => (option) => (input.value = option ? option.id : '')
 
 function init () {
-  const recruitmentCycleYear = document.getElementById('accredited_provider_search_form_recruitment_cycle_year')?.value
-  if (!recruitmentCycleYear) return
+  const accreditedProviderSearchForm = document.querySelector('form[data-recruitment-cycle-year]')
+  if (!accreditedProviderSearchForm) return
+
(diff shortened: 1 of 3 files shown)


app/javascript/publish/autocomplete/accredited_provider.js
```diff
@@ -5,7 +5,7 @@ const providerSuggestionTemplate = (result) => result && `${result.provider_name
 const onConfirm = (input) => (option) => (input.value = option ? option.id : '')
 
 function init () {
-  const recruitmentCycleYear = document.getElementById('accredited_provider_search_form_recruitment_cycle_year')?.value
+  const recruitmentCycleYear = location.pathname.match(/\/support\/(\d{4,4})/)[1]
```
Nitemaeric, line 8: There's something fresh off the press for this actually: https://blog.cloudflare.com/improving-web-standards-urlpattern/

But... It seems support isn't universal yet... There are polyfills, but I'm cautious about adding them.

How about using `String#split` and fetching by index?

inulty-dfe: I've taken a completely different approach

Nitemaeric (approved):
If we were using stimulus here, it would look very similar, but slightly more concise.

inulty-dfe:
> If we were using stimulus here, it would look very similar, but slightly more concise.

I agree, we should create a policy of moving things to Stimulus if that's what we want to do.

PR #5079, opened 2025-03-24
Add script to download DB backup and restore to development

inulty-dfe:
## Context

It's good to test locally with up to date sanitized backup from prod. It's a bit annoying to have to download in the browser.

This script downloads the sanitzed backup, drops and recreates the database and seeds the local development database with the sanitized backup.

Add a convenience script to reload local db with sanitized backup.

## Guidance to review

A similar script already exists, I discovered after writing this one.
https://github.com/DFE-Digital/publish-teacher-training/blob/main/bin/download-nightly-backup

I don't think we need all those environment variables set. But lets test.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,6 +6,8 @@ coverage
 .git
 .github
 terraform
+# Better safe than sorry
+bin/restore-backup
 
 # work is my local project-specific non-committed workspace --Misha
 work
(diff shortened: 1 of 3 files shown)


Nitemaeric (commented):
Should this be one of the `make` commands?

Nitemaeric (approved):

inulty-dfe:
I should add this to the documentation

PR #5090, opened 2025-03-26
Make sure we smart quotes provider name before saving into the DB

tomas-stefano:
## Context

On our search results, providers which has single quotes on their names are not displaying smart quotes.

We should make sure smart quotes are saved into the database instead of changing on the presentation layer.

* Normalize the provider name attribute

## Guidance to review

1. When you save does the provider name is converted properly?

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -58,7 +58,24 @@ class Provider < ApplicationRecord
   has_many :accredited_partnerships, foreign_key: :training_provider_id, class_name: 'ProviderPartnership', inverse_of: :training_provider
   has_many :accredited_partners, through: :accredited_partnerships, class_name: 'Provider', source: :accredited_provider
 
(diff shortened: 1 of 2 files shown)


tomas-stefano:
https://trello.com/c/YY71CfYi/593-smart-quotes-on-provider-name-on-search-results

tomas-stefano:
The autocomplete works as expected around single and smart quotes too

Screenshot 2025-03-26 at 16 42 15

inulty-dfe (commented):
We're sure we'll **never** want the school name to be "unsmart" quoted?

tomas-stefano:
> We're sure we'll **never** want the school name to be "unsmart" quoted?

If we need we can make "unsmart" which could be the exception rather than the rule.

inulty-dfe (approved):

PR #5092, opened 2025-03-27
WIP

tomas-stefano:


tomas-stefano:
https://trello.com/c/Wh3VCuQs/427-flaky-tests-provider-code-is-taken-in-test-suite

inulty-dfe:
Is there another way to do this other than making DB queries until it works?

inulty-dfe:
> Is there another way to do this other than making DB queries until it works?

Would it be better to select all the provider codes at once and use a code that wasn't in the result rather than generate, query, generate, query?

tomas-stefano:
> > Is there another way to do this other than making DB queries until it works?
> 
> Would it be better to select all the provider codes at once and use a code that wasn't in the result rather than generate, query, generate, query?

Ok, I changed the implementation. Let me know your thoughts.

inulty-dfe:
> > > Is there another way to do this other than making DB queries until it works?
> > 
> > 
> > Would it be better to select all the provider codes at once and use a code that wasn't in the result rather than generate, query, generate, query?
> 
> Ok, I changed the implementation. Let me know your thoughts.

Lets keep an eye on build / test times and see if this is slows down the test suite in the coming weeks

inulty-dfe (approved):
Let's keep an eye on the test suite to see if this makes it much slower

PR #5095, opened 2025-03-27
Fallback to page 1 if page parameter can not be parsed

tomas-stefano:
## Context

Some requests are returning 500.because requests are adding custom value to the page parameter.O

e.g /results?page=something

This PR handle this parameter by fallback to first page if something is wrong with params[:page]

## Guidance to review

1. Visit `/results?page=something`
2. Does it fallback to the first page?
3. Apply filters and then change the page parameter again
4. Does it fallback to the first page?

### DIFF
--- a/app/controllers/find/results_controller.rb
+++ b/app/controllers/find/results_controller.rb
@@ -12,7 +12,8 @@ module Find
       @courses_query = ::Courses::Query.new(params: @search_params.dup)
       @courses = @courses_query.call
       @courses_count = @courses_query.count
-      @pagy, @results = pagy(@courses, count: @courses_count)
+
+      @pagy, @results = pagy(@courses, count: @courses_count, page:)
     end
 
     private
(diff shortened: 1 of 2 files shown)


tomas-stefano:
https://trello.com/c/W7X8EBuk/589-invalid-find-requests-handling

spec/requests/find/results_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Find
+  describe '/resulrs' do
```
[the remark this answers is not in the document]
tomas-stefano: Done!

app/controllers/find/results_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 module Find
   class ResultsController < Find::ApplicationController
+    rescue_from Pagy::VariableError, with: :render_not_found
```
Nitemaeric, line 5: Let's remove this.

tomas-stefano: Done

inulty-dfe (approved):
:rocket:

Nitemaeric (approved):

PR #5098, opened 2025-03-27
Fix failing system tests

gms-gs:
## Context

We disabled our system tests as we had very flakey tests and blocked deployment when we needed to release a feature

- Fixed failing system tests which needed to an extra stub 
- Remore v2 folder in specs
- Refactored `spec/system/find/results/tracking/search_results_tracking_spec.rb` to stop checking the queue but to now just check that `Find::Analytics::SearchResultsEvent` is called with new and with specific params.
- Due to the extreme flakiness of the tests I have added a `wait_for` block to ensure the exceptions are repeatedly evaluated until the underlying matcher passes or the configured timeout elapses. This has given consistent better results since being implemented.

Very rarely I have seen the below error in `search_results_enabled_spec `. This resolves itself after a rerun of failed jobs usually. I think it may make sense to see how much of a pain this in the future, as the occurrence is very minimal. Overall however tests passing has significantly improved, and we can now rely more on these system tests. This is the same as when a unit test or model test randomly fails but is very infrequent.
 
```
Selenium::WebDriver::Error::StaleElementReferenceError:
       stale element reference: stale element not found
         (Session info: chrome=134.0.6998.165); For d`
``` 
## Guidance to review

Run the system tests multiple times to check for consistent passing

## Checklist


tomas-stefano (approved):

inulty-dfe:
image

Nitemaeric (approved):

PR #5109, opened 2025-03-31
Move further education filter to the bottom

tomas-stefano:
## Context

Moved the further education as the last filter on the search results

## Guidance to review

1. Visit /results page
2. See further education filter (desktop, mobile)
3. It should be the last filter


tomas-stefano:
https://trello.com/c/MlD0NIGo/572-bug-further-education-filter-in-the-wrong-place

PR #5116, opened 2025-04-01
Move feature flags into Support namespace

Nitemaeric:
## Context

Feature flags are currently defined within Find, which require us to use basic authentication to authenticate the path.

As we have a Support console, we should rely on the Support User authentication to authenticate changes to feature flags.

- Clean up feature flag-specific components as they are not reused.
- Use `govuk_summary_card` in place of internal component.
- Move feature flags into the `Support` namespace.

## Guidance to review

- Visit https://publish-review-5116.test.teacherservices.cloud/support/feature-flags
- You should not be able to access it.
- Once you sign in, you will be able to.

## Checklist


app/controllers/support/feature_flags_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Support
+  class FeatureFlagsController < ApplicationController
+    def update
+      FeatureFlag.send(action, feature_name)
```
tomas-stefano, line 6: `action` and  `feature_name` comes directly from user params. 🤔 

I know that is an admin interface, but this is a security risk because it allows users to invoke any method on the FeatureFlag class, potentially leading to arbitrary code execution.

Can we use a before_action :verify_allowed_actions that verifies all possible actions before calling send?

Nitemaeric: Great idea! :) I'll update.

Nitemaeric: Decided to split out the `activate` and `deactivate` actions. The user will see be met with a 404 error in the case of sending an invalid action.

tomas-stefano (approved):

PR #5126, opened 2025-04-03
Move specs from features to system 1

inulty-dfe:
## Context

We are aiming to move away from feature tests and rely only on Rails system tests. This will allow us to remove DatabaseCleaner and move forward with more canonical testing methodologies.

Move the base feature tests to the system test directory.

## Checklist


Nitemaeric (approved):
🚀

PR #5132, opened 2025-04-04
Add some docs around Geolocation in the service

inulty-dfe:
## Context

Add a markdown doc describing what Geolocation/Geocoding we do and where to find it.

Why is `Query` in`CoordinatesQuery`, and not `Suggestions`

https://github.com/DFE-Digital/publish-teacher-training/blob/8f50906ed7afadd4fbb057afc551745a49cc9848/guides/geolocation.md

## Checklist

### DIFF
--- a/guides/geolocation.md
+++ b/guides/geolocation.md
@@ -1,0 +1,28 @@
+# Geolocation
+
+We use geolocation services in two ways. 
+
+1. Autocomplete a users text input when searching for a location.
+2. Retreiving the geographic coordinates of the corresponding place.
+
+### *Example:*
+- A user wants to find the courses near Bristol and order the courses by their distance to Bristol.
+
+#### Scenario
+1. First the user types 'Bristol' into the search for the location.
+2. This is the Google geocode service. It returns a matching place and location types (postcode, administrative_area_2, etc.).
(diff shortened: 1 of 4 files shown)


guides/geolocation.md
```diff
@@ -0,0 +1,27 @@
...
+1. Autocomplete a users text input when searching for a location.
+2. Retreiving the geographic coordinates of the corresponding place.
+
+### *Example:*
+- A user wants to find the courses near Bristol and order the courses by their distance to Bristol.
+
+#### Scenario
+1. First the user types 'Bristol' into the search for the location.
+2. This is the Google geocode service. It returns a matching place and the corresponding `place_id`.
+3. We use this `place_id` to get the coordinates of the place  from Google.
```
tomas-stefano, line 14: We abandoned the place id and now use the geocode through google using the query parameter:

https://github.com/DFE-Digital/publish-teacher-training/pull/5132/files#diff-29ce6a1dc9b6aa8328a3fb7b12a7d4698985f7035badf06b175deeafe4119b9aR43

We could return to use the place_id although there is no advantage.

tomas-stefano (approved):

PR #5143, opened 2025-04-07
As a Support User, I want to see a list of all Recruitment Cycles

gms-gs:
## Context

We want to be able to trigger Rollover via the Support Console.

Add a Recruitment Cycles index page to list all past and present cycles 

## Guidance to review

- Viisit publish
- Visit Settings
- Visit Recruitment Cycles tab

## Checklist

Screenshot 2025-04-07 at 22 10 44

### DIFF
--- a/app/controllers/support/recruitment_cycles_controller.rb
+++ b/app/controllers/support/recruitment_cycles_controller.rb
@@ -2,7 +2,9 @@
 
 module Support
   class RecruitmentCyclesController < ApplicationController
-    def index; end
+    def index
+      @recruitment_cycles = RecruitmentCycle.order(year: :desc)
+    end
 
     def new
       @support_recruitment_cycle_form = RecruitmentCycleForm.new
--- a/app/views/support/recruitment_cycles/index.html.erb
+++ b/app/views/support/recruitment_cycles/index.html.erb
@@ -6,6 +6,30 @@
       <h1 class="govuk-heading-l"><%= t(".page_title") %></h1>
 
       <%= govuk_button_link_to t(".add_recruitment_cycle"), new_support_recruitment_cycle_path %>
+
+      <%= govuk_table do |table| %>
+        <% table.with_head do |head| %>
+          <% head.with_row do |row| %>
+            <% row.with_cell text: RecruitmentCycle.human_attribute_name(:year) %>
+            <% row.with_cell text: RecruitmentCycle.human_attribute_name(:start_date) %>
+            <% row.with_cell text: RecruitmentCycle.human_attribute_name(:end_date) %>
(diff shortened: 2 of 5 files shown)


app/controllers/support/recruitment_cycles_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 module Support
   class RecruitmentCyclesController < ApplicationController
-    def index; end
+    def index
+      @recruitment_cycles = RecruitmentCycle.ordered_by_current_first
```
tomas-stefano, line 6: Do we need a scope if is only used here? Or is the idea to use somewhere? Wdyt? @Nitemaeric

app/models/recruitment_cycle.rb
```diff
@@ -10,6 +10,10 @@ class RecruitmentCycle < ApplicationRecord
 
   validates :year, presence: true
 
+  scope :ordered_by_current_first, lambda {
+    order(Arel.sql("CASE WHEN year = '#{Settings.current_recruitment_cycle_year}' THEN 0 ELSE 1 END, year DESC"))
```
tomas-stefano, line 14: Relying on `Settings.current_recruitment_cycle_year` means that when the new cycle starts we will have to change the yaml file and do a deployment.

🤔 🤔 🤔 

Would be good for us to consider to remove the usage of `Settings.current_recruitment_cycle_year` on any checks and use from the recruitment cycle object/timetable we will about to create 

As for this PR might be fine but we will have to revisit this code.

@Nitemaeric / @gms-gs about the above what are your thoughts?

app/controllers/support/recruitment_cycles_controller.rb
tomas-stefano: Yes, initially me too, but adds more indirection to the code. But it sounds okish for now

Nitemaeric: Why do we need to order by current first? Can't we `order(year: :desc)`?

app/views/support/recruitment_cycles/index.html.erb
```diff
@@ -1,3 +1,27 @@
...
+    <% head.with_row do |row| %>
+      <% row.with_cell text: t(".year") %>
+      <% row.with_cell text: t(".start_date") %>
+      <% row.with_cell text: t(".end_date") %>
+      <% row.with_cell text: t(".status") %>
+    <% end %>
+  <% end %>
+
+  <% table.with_body do |body| %>
+    <% @recruitment_cycles.each do |cy| %>
```
Nitemaeric, line 16: Can we always use explicit variable names please?

```suggestion
    <% @recruitment_cycles.each do |recruitment_cycle| %>
```

Nitemaeric (approved):

PR #5146, opened 2025-04-08
Update restore-backup script and instructions

inulty-dfe:
## Context

`#!/usr/bin/env` is more portable than `#!/bin/env`
The instructions for restoring the database weren't clear enough that only 1 option should be used.

## Checklist

### DIFF
--- a/bin/restore-backup
+++ b/bin/restore-backup
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
 
 # Download today's sanitized backup of the publish database to the Downloads directory
 # Drop and recreate the development and test databases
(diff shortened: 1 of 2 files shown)


bin/restore-backup
```diff
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
```
Nitemaeric, line 1: I ran into this today too.

I have some additional notes for improvement. In my case, I couldn't run this as `psql` defaults to the local unix socket. I need to pass through additional options to point to my local postgres container.

```sh
psql manage_courses_backend_development -h localhost -U postgres < ~/Downloads/sanitised_backup.sql
```

We might need to pass through arguments to the script.

Nitemaeric (approved):

bin/restore-backup
inulty-dfe: Would environment variables work for you?

image

Nitemaeric: 🤔 Yeah, why not. I was thinking to pass through arguments via "$@", but both work, I _think_.

PR #5148, opened 2025-04-08
Refactor methods around provider partnerships

inulty-dfe:
## Context

1. Removal of the last of the AccreditingProviderEnrichment code.

2. Clean up the Provider model:

We had `accrediting_providers` aliased to `accredited_providers` which called `accredited_partners`

 - accrediting_providers
 - accredited_providers
 - accrediting_bodies
 - accrediting_body

## Guidance to review

**Look at the commits one by one**

## Checklist


.github/workflows/build-and-deploy.yml
```diff
@@ -172,7 +172,7 @@ jobs:
         # Pass Settings compatible env vars to runs tests with
         # feature flag(s) enables
         # Separate mutliple flags with a space
-        feature-flags: ["", "SETTINGS__FEATURES__PROVIDER_PARTNERSHIPS=true"]
+        feature-flags: [""]
```
Nitemaeric, line 175: Let's remove this from the matrix entirely. Moving forward, let's test feature flags in a controlled environment. Having the flag on during the "Add support user" flow doesn't really give us any value.

app/components/course_preview/missing_information_component.rb
```diff
@@ -26,7 +26,7 @@ def render?
     end
 
     def accrediting_provider_present?(course)
-      course.provider.accredited_providers.include?(course.accrediting_provider)
+      course.provider.accredited_partners.include?(course.accrediting_provider)
     end
```
Nitemaeric, line 30: This logic seems weird. Should it not be `course.ratifying_provider.present?`? Why does the course preview need to be aware of partnerships?

.github/workflows/build-and-deploy.yml
inulty-dfe: I don't understand what you're saying about 

> Having the flag on during the "Add support user" flow doesn't really give us any value.

app/components/course_preview/missing_information_component.rb
inulty-dfe: I'll look into this

inulty-dfe: This method can go. It produces a link to edit the accredited provider description if one doesn't exist. But now we don't show that information anywhere anymore. We show the accredited providers own "About us" (train_with_us)

.github/workflows/build-and-deploy.yml
Nitemaeric: I mean, let's remove the practice of setting feature flags in the CI workflow level. If you need to test a feature flag flow, use a `context` which sets the feature flags.

This way, unrelated tests aren't run twice for, generally, zero benefit gained. There is the possibility that a change _somehow_ breaks something entirely unrelated, but it is far too infrequent for us to test this for every branch push.

inulty-dfe: Yeah, we can use it sparingly, but I don't see why we'd basically remove the ability to do that in those rare cases

Nitemaeric: 🤔 True. Happy to keep it in.

tomas-stefano (approved):

app/components/course_preview/missing_information_component.rb
Nitemaeric: Will you remove it on this PR or another?

inulty-dfe: I didn't realise I hadn't! :sweat_smile:

inulty-dfe: I made some rebasing mistakes and deleted my changes, I've added them back in

PR #5153, opened 2025-04-09
Improve results - Update no results page with radius quick links

gms-gs:
## Context

Research has shown that candidates respond well to no results when there are suggesting on how they might change their search, for example widen the search radius. This also aligns with the principles of [good services](https://www.gov.uk/service-manual/design/introduction-designing-government-services#:~:text=There%20are%20no%20dead%20ends,-Every%20user%20needs&text=If%20somebody's%20needs%20are%20outside,without%20knowing%20how%20to%20continue.) to have no dead ends.

Implementation
- On the no results index page we use stimulus to make an API call when no results are found for a search
- This calls `app/controllers/api/radius_quick_link_suggestions_controller.rb` via a stimulus controller `app/javascript/find/controllers/radius_quick_link_suggestions_controller.js`
- This will return links to an expanded search of possible courses based on the previous search params
- The reason why it is in an API call and not just called via the controller is that it had performance issues, so Dan suggested moving this into an API (similar to what Apply did), we also limit the returned courses to 101.
- We also have a loading indicator which lets the user know its loading suggestions while the call takes place
- This will return the links once completed
 
## Guidance to review

- Search for `Mandarin` and use the location of `Manchester, UK` and a search radius of `5 miles`
- You should then see quick links appear where search results would be
- Click these quick links to then search based on current params but with a wider radius search

## Checklist


app/components/courses/no_results_component.rb
```diff
@@ -24,6 +25,32 @@ def try_another_search_content
...
+      @search_params[:provider_name]
+    end
+
+    def search_by_location
+      @search_params[:location]
+    end
+
+    def no_results_quick_links
+      max_radius_results = Courses::Query.call(params: { **@search_params, radius: 200 })
+      quick_link_radiuses = [20, 50, 100, 200]
```
inulty-dfe, line 38: There is a list of radius options in the `SearchForm`. Can we try to find a way to associate these values with those so we're not defining search radii in two different places

app/components/courses/no_results_component.rb
```diff
@@ -24,6 +25,32 @@ def try_another_search_content
...
+
+      quick_link_radiuses.map do |radius|
```
inulty-dfe, line 40: Could `#group_by` be useful here?

spec/components/find/courses/no_results_component_spec.rb
```diff
@@ -3,17 +3,18 @@
...
-  include Rails.application.routes.url_helpers
-
   subject(:component) do
     render_inline(
-      described_class.new(country:, minimum_degree_required:, subjects:)
+      described_class.new(country:, minimum_degree_required:, subjects:, search_params:)
     )
   end
 
+  include Rails.application.routes.url_helpers
```
inulty-dfe, line 12: Why was this moved down?

inulty-dfe (commented):
Few suggestions / questions

app/components/courses/no_results_component.rb
```diff
@@ -24,6 +25,32 @@ def try_another_search_content
...
+    def search_by_provider_name
+      @search_params[:provider_name]
+    end
+
+    def search_by_location
+      @search_params[:location]
+    end
+
+    def no_results_quick_links
+      max_radius_results = Courses::Query.call(params: { **@search_params, radius: 200 })
```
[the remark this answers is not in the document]
inulty-dfe: If you have this query in the component, you need to make database queries in order to test it. Or mock the database

app/components/courses/no_results_component.html.erb
```diff
@@ -20,4 +20,21 @@
...
+  <% unless search_by_provider_name || !search_by_location %>
+    <% if @search_params[:subject_name].present? %>
+      <h3 class="govuk-heading-m">
+        <%= t(".try_browsing", subject_name: @search_params[:subject_name]) %>
+      </h3>
+    <% else %>
+      <h3 class="govuk-heading-m">
+        <%= t(".try_browsing_without_subject") %>
+      </h3>
+    <% end %>
```
Nitemaeric, line 33:
```suggestion
    <h3 class="govuk-heading-m">
      <% if @search_params[:subject_name].present? %>
        <%= t(".try_browsing", subject_name: @search_params[:subject_name]) %>
      <% else %>
        <%= t(".try_browsing_without_subject") %>
      <% end %>
    </h3>
```

app/javascript/find/application.js
```diff
@@ -18,6 +19,7 @@ Stimulus.register('remote-autocomplete', RemoteAutocompleteController)
 Stimulus.register('locations-autocomplete', LocationsAutocompleteController)
 Stimulus.register('visibility', VisibilityController)
 Stimulus.register('provider-autocomplete', ProviderAutocompleteController)
+Stimulus.register('radius-quick-link-suggestions', RaduisQuickLinkSuggestionsController)
```
inulty-dfe, line 22: Typo here

PR #5154, opened 2025-04-10
Automating Rollover 

tomas-stefano:
## Context

This PR adds the rollover to sidekiq.

* When the new recruitment cycle is created the Rollover service kicks in

## Observations

I decided to add to Publish the famous Apply `BatchDelivery` that I created in Apply some years ago.

The reasoning is that `RolloverJob` if I do this:

```ruby
class RolloverJob < ApplicationJob                                                   
  def perform                                                                        
    RecruitmentCycle.current_recruitment_cycle.providers.each do |provider|          
      RolloverProviderService.call(provider_code: provider.provider_code, force: false)                                                          
    end                                                                              
  end                                                                                
end
```

This will take 25 minutes to finish and timeout sidekiq.

Or if I do this:

```ruby
class RolloverJob < ApplicationJob                                                   
  queue_as :default                                                                  
                                                                                                                                                 
  def perform                                                                    
    RecruitmentCycle.current_recruitment_cycle.providers.each do |provider|      
      RolloverProviderJob.perform_later(provider.provider_code)                  
    end                                                                          
  end                                                                            
end 
```

It will attempt to enqueue jobs for all providers at once, causing the database to reach its maximum connection limit. 

With 800 providers, many of which have a broad range of courses (some with very few and others with many), the simultaneous job execution overwhelmed Sidekiq and the database, resulting in hundreds of errors - which retried and succeed at the end but at the cost of the DB erroring if there are requests from Find or Publish during the execution time.

To address this issue, I implemented a `BatchDelivery` mechanism that divides providers into smaller batches and staggers the execution of jobs over time. This ensures that jobs are processed gradually, preventing resource contention and improving system stability.

Key highlights:
- Providers are processed in batches of 10. - This was arbitrary because it gives a breathing room for all providers to execute succesfully.
- Each batch is staggered by 1-hour intervals to avoid overwhelming the database. - also 1 hour arbitrary for the same reason explained above
- The staggered execution significantly removes the likelihood of max connection errors while maintaining scalability within the app.

## Guidance to Review

1. Go to support
2. Go to Settings
3. Go to recruimtnet cycles
4. Add a new recruitment cycle
5. See `/sidekiq`

### DIFF
--- a/spec/jobs/rollover_job_spec.rb
+++ b/spec/jobs/rollover_job_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-require 'rails_helper'
-require 'sidekiq/testing'
+require "rails_helper"
+require "sidekiq/testing"
 
 RSpec.describe RolloverJob do
   let(:recruitment_cycle) { create(:recruitment_cycle) }
@@ -14,21 +14,21 @@ RSpec.describe RolloverJob do
 
   after { Sidekiq::Testing.disable! }
 
-  describe '#perform' do
-    it 'enqueues jobs with correct staggered timings' do
+  describe "#perform" do
+    it "enqueues jobs with correct staggered timings" do
       Timecop.freeze do
         current_time = Time.zone.now
         subject.perform
 
-        expect(Sidekiq::Queues['default'].size).to eq(15)
+        expect(Sidekiq::Queues["default"].size).to eq(15)
 
         providers[0..9].each do |provider|
           job = find_job(provider.provider_code)
 
           expect(job).to include(
-            'class' => 'RolloverProviderJob',
-            'args' => [provider.provider_code],
-            'enqueued_at' => current_time.to_f
+            "class" => "RolloverProviderJob",
+            "args" => [provider.provider_code],
+            "enqueued_at" => current_time.to_f,
           )
         end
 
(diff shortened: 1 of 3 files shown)


app/services/batch_delivery.rb
Nitemaeric, line 1: I see this more of a `lib` than a `service`. It's also app agnostic.

app/jobs/rollover_job.rb
```diff
@@ -3,5 +3,15 @@
...
   queue_as :default
 
-  def perform(recruitment_cycle_id); end
+  BATCH_SIZE = 10
+
+  def perform
+    relation = RecruitmentCycle.current_recruitment_cycle.providers
+
+    BatchDelivery.new(relation:, stagger_over: 1.hour, batch_size: BATCH_SIZE).each do |batch_time, providers|
+      providers.pluck(:provider_code).each do |provider_code|
```
Nitemaeric, line 12: I wonder if it would be cleaner to add a helper method to `ActiveRecord::Relation` or not? 🤔 

```suggestion
    RecruitmentCycle.current_recruitment_cycle.providers.in_batches_over_time(1.hour, batch_size: BATCH_SIZE) do |batch_time, providers|
      providers.pluck(:provider_code).each do |provider_code|
```

But I guess we land into the realm of monkey-patching again...

app/services/recruitment_cycle_creation_service.rb
```diff
@@ -10,12 +10,12 @@ def initialize(year:, application_start_date:, application_end_date:)
...
   def call
-    recruitment_cycle = RecruitmentCycle.create!(
+    RecruitmentCycle.create!(
       year: @year,
       application_start_date: @application_start_date,
       application_end_date: @application_end_date
     )
 
-    RolloverJob.perform_later(recruitment_cycle.id)
+    RolloverJob.perform_later
```
Nitemaeric, line 19: 🤔 I think it would be better to pass through the ID here, no?

Nitemaeric: Also, if you were to create multiple years or many years ahead (not saying that a user should be able to do so, but based on the code), it would always refer to the current cycle.

tomas-stefano: Then we would have to change the rollover code to handle this scenario. Can this be another PR and another trello card?

app/jobs/rollover_job.rb
tomas-stefano: > But I guess we land into the realm of monkey-patching again...
Yes, I'd avoid this path of adding objects to external libraries.

Wrapping and extending functionality is less intrusive and less error-prone.

app/services/recruitment_cycle_creation_service.rb
Nitemaeric: Sure. :)

Nitemaeric (approved):

PR #5165, opened 2025-04-16
Providers can't publish a course

tomas-stefano:
## Context

Card originated from an investigation on this support ticket.

One provider was unable to publish a course, but investigation reveals this is an issue affecting at least 14 providers and hundreds (265exactly) of courses.

The root cause is because the courses are missing accrediting provider partnerships, resulting in failed validations when course publishing.

One provider reported being unable to publish a course. The error message indicated the accrediting provider is not accredited or the partnership is missing.

Providers affected: 14

Published Courses affected: 265 published courses are currently invalid for re-publishing (update and publish again).

### DIFF
--- a/app/helpers/support/recruitment_cycles_helper.rb
+++ b/app/helpers/support/recruitment_cycles_helper.rb
@@ -1,0 +1,12 @@
+module Support
+  module RecruitmentCyclesHelper
+    def recruitment_cycle_status_tag(recruitment_cycle)
+      options = {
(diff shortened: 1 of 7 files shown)


tomas-stefano:
https://trello.com/c/STtxc85M/657-providers-cant-publish-a-course

tomas-stefano:
After this PR is merged this query can be ran:

```ruby
RecruitmentCycle.current.courses.joins(:provider).where.not(accredited_provider_code: nil).where(provider: { accredited: true }).update_all(accredited_provider_code: nil)
```

PR #5166, opened 2025-04-16
Separate specs between homepage and results page

inulty-dfe:
## Context

  Our system tests are unwieldy. I've tried to separate them into
  smaller more specific scenarios.

  One apparent discrimination I can make is to separate specs that
  happen on the home page and those that happen on the results page.

Break up the find system tests into logical files.

Separate them between homepage tests

|Page|Concept|
|---|---|
|**Home page**||
|| quicklinks|
|| tracking|
|| Searching by subject, location, both|
|**Results page**||
||filtering|
||sorting|
||no results|
||tracking|
||viewing course|

## Guidance to review

How do we manage shared methods?

## Checklist


spec/system/find/results_page/subject_and_location_spec.rb
```diff
@@ -94,50 +94,6 @@
     and_london_is_displayed_in_text_field
   end
 
-  context "searching from the homepage" do
```
inulty-dfe, line 97: Moved to the homepage directory

PR #5177, opened 2025-04-22
Split System tests and upload screenshots on failure

inulty-dfe:
## Context

Use the [actions/upload-artifact](https://github.com/actions/upload-artifact) to upload screenshots of test failures

> Zip archives
> When an Artifact is uploaded, all the files are assembled into an immutable Zip archive. There is currently no way to download artifacts in a format other than a Zip or to download individual artifact contents.

## Guidance to review

image

I tried this but it didn't work and I don't know how to debug it easily

```
        env:
          TEST_GROUP: ${{ matrix.tests.name }}

....

      - name: Upload screenshots
        if: ${{ failure() && startsWith(env.TEST_GROUP, 'System') }}
```

## Checklist

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -202,6 +202,13 @@ jobs:
           EXCLUDE_PATTERN: ${{ matrix.tests.exclude-pattern || ' ' }}
           FEATURE_FLAGS: ${{ matrix.feature-flags }}
 
+      - name: Upload screenshots
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


Nitemaeric (commented):
I think you can add an additional key to the matrix.

```yaml
- name: System - Other
  include-pattern: spec/system/.*_spec.rb
  exclude-pattern: spec/system/find/.*_spec.rb
  upload-screenshots: true
```

And then reference it below:

```yaml
if: ${{ failure() && matrix.tests.upload-screenshots }}
```

tomas-stefano (approved):

PR #5179, opened 2025-04-23
Implement mustache.js for JS template rendering

gms-gs:
## Context

We currently render content within stimulus controllers using raw HTMlL. We would prefer not to do this directly, it was discussed with the devs that we introduce https://github.com/janl/mustache.jsConnect your Github account for templating in JavaScript.

Extract raw HTML from our stimulus controllers and use `mustache.js` to manage this for us

## Guidance to review

- Play with the quick links on the no results page to check for any regression
- Play with the secondary courses filters on the results page to check for any regression

## Checklist

### DIFF
--- a/app/javascript/find/controllers/filter_search_controller.js
+++ b/app/javascript/find/controllers/filter_search_controller.js
@@ -1,8 +1,9 @@
 import { Controller } from '@hotwired/stimulus'
+import Mustache from 'mustache'
 
 // Connects to data-controller="filter-search"
 export default class extends Controller {
-  static targets = ['optionsList', 'searchInput', 'legend']
+  static targets = ['optionsList', 'searchInputContainer', 'legend', 'template', 'searchInput']
 
   static instanceCounter = 0
 
@@ -10,49 +11,28 @@ export default class extends Controller {
     const legend = this.element.querySelector('legend')
     legend.dataset.filterSearchTarget = 'legend'
 
-    // This will be unique for each filter-search instance
+    // Assign a unique ID for input/label association
     this.instanceId = this.constructor.instanceCounter++
 
-    const searchInput = this.createSearchInput()
-    this.optionsListTarget.before(searchInput)
-  }
+    const template = this.templateTarget.innerHTML
+    const html = Mustache.render(template, {
+      id: `${this.identifier}-${this.instanceId}-input`,
+      label: this.legendTarget.innerText
+    })
 
-  createSearchInput () {
-    const container = document.createElement('div')
-    container.classList.add('filter-search__search')
-
-    const inputId = `${this.identifier}-${this.instanceId}-input`
-    const labelText = `${this.legendTarget.innerText}`
-
-    container.innerHTML = `
-      <label for="${inputId}" class="govuk-label govuk-visually-hidden">
-        ${labelText}
-      </label>
-      <input type="search" id="${inputId}"
-        class="govuk-input govuk-!-margin-bottom-1"
-        autocomplete="off"
-        placeholder="Search"
-        data-action="input->${this.identifier}#search"
-        data-filter-search-target="searchInput">
-    `
-
-    return container
+    this.searchInputContainerTarget.innerHTML = html
   }
 
   search () {
-    const optionItems = this.optionsListTarget.children
+    const optionItems = this.optionsListTargets
     const searchValue = this.searchInputTarget.value.toLowerCase()
 
     this.toggleItems(optionItems, searchValue)
   }
 
   toggleItems (items, searchValue) {
-    Array.from(items).forEach(function (item) {
-      if (item.textContent.toLowerCase().indexOf(searchValue) > -1) {
-        item.style.display = ''
-      } else {
-        item.style.display = 'none'
-      }
+    Array.from(items).forEach(item => {
+      item.style.display = item.textContent.toLowerCase().includes(searchValue) ? '' : 'none'
     })
   }
 }
--- a/app/views/find/results/filters/_all.html.erb
+++ b/app/views/find/results/filters/_all.html.erb
@@ -21,12 +21,48 @@
     </div>
   <% end %>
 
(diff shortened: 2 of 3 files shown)


app/javascript/find/controllers/radius_quick_link_suggestions_controller.js
```diff
@@ -34,37 +33,27 @@ export default class extends Controller {
...
-    this.headingTarget.innerHTML = `<h3 class="govuk-heading-m">${headingText}</h3>`
-
-    const listItems = links.map(link =>
-      `<li class="govuk-list govuk-list--bullet"><a href="${link.url}" class="govuk-link">${link.text}</a></li>`
-    ).join('')
+    const template = document.getElementById('radius-suggestions-template').innerHTML
+    const rendered = Mustache.render(template, {
+      subjectName: this.subjectNameValue,
+      links: links
+    })
```
Nitemaeric, line 55: Let's use object shorthand where possible. I'll update our linter rules in a future PR.

```suggestion
    const rendered = Mustache.render(template, {
      subjectName: this.subjectNameValue,
      links
    })
```

app/components/courses/no_results_component.html.erb
```diff
@@ -28,6 +28,26 @@
     </p>
   <% end %>
 
-  <div data-radius-quick-link-suggestions-target="heading"></div>
+  <div data-radius-quick-link-suggestions-target="loading">
+    <p class="govuk-body"><%= t(".loading_suggestions") %></p>
+  </div>
+
   <div data-radius-quick-link-suggestions-target="container"></div>
```
Nitemaeric, line 35: Do we need 2 containers here?

app/javascript/find/controllers/radius_quick_link_suggestions_controller.js
```diff
@@ -34,37 +35,30 @@ export default class extends Controller {
...
+      .then(data => this.renderLinks(data))
       .catch(error => {
         console.error('Fetch error:', error)
         this.containerTarget.innerHTML = ''
       })
   }
 
+  renderLoading () {
+    const template = this.loadingTemplateTarget.innerHTML
+    this.containerTarget.innerHTML = Mustache.render(template, {})
```
Nitemaeric, line 47: Not a blocker, but I don't think there's value in extracting this out into it's own variable.

Good call on using targets instead of finding elements by ID.

```suggestion
    this.containerTarget.innerHTML = Mustache.render(
      this.loadingTemplateTarget.innerHTML,
      {}
    )
```

app/javascript/find/controllers/radius_quick_link_suggestions_controller.js
```diff
@@ -34,37 +35,30 @@ export default class extends Controller {
...
+      subjectName: this.subjectNameValue,
+      links
+    })
 
-    this.containerTarget.innerHTML = `
-      <ul class="govuk-list govuk-list--bullet">
-        ${listItems}
-      </ul>
-    `
+    this.containerTarget.innerHTML = rendered
```
Nitemaeric, line 62: Same here.

```js
this.contentTemplateTarget.innerHTML = Mustache.render(
  this.contentTemplateTarget.innerHTML,
  {
    subjectName: this.subjectNameValue,
    links
  }
)
```

app/views/find/results/filters/_all.html.erb
```diff
@@ -21,12 +21,46 @@
...
+      <label for="{{id}}" class="govuk-label govuk-visually-hidden">
+        {{label}}
+      </label>
+      <input
+        type="search"
+        id="{{id}}"
+        class="govuk-input govuk-!-margin-bottom-1"
+        autocomplete="off"
+        placeholder="Search"
+        data-action="input->filter-search#search">
```
Nitemaeric, line 35: Why did you remove `data-filter-search-target="searchInput"` from the attributes here?

app/javascript/find/controllers/filter_search_controller.js
```diff
@@ -1,58 +1,38 @@
...
 
   toggleItems (items, searchValue) {
-    Array.from(items).forEach(function (item) {
-      if (item.textContent.toLowerCase().indexOf(searchValue) > -1) {
-        item.style.display = ''
-      } else {
-        item.style.display = 'none'
-      }
+    Array.from(items).forEach(item => {
+      item.style.display = item.textContent.toLowerCase().includes(searchValue) ? '' : 'none'
```
Nitemaeric, line 35: Last thing, by adding the `searchInputTarget` back, these changes shouldn't be necessary again, right?

app/views/find/results/filters/_all.html.erb
```diff
@@ -21,12 +21,47 @@
     </div>
   <% end %>
 
-  <%= form.govuk_check_boxes_fieldset :secondary_subjects, legend: { text: t("helpers.legend.courses_search_form.secondary_html"), size: "s" }, hint: { text: t("helpers.hint.courses_search_form.secondary"), class: "govuk-!-font-size-16" }, class: "govuk-checkboxes--small", hidden: false, multiple: false, form_group: { class: "app-filter__group", data: { controller: "filter-search" } } do %>
-    <div class="filter-search scrollable-filter secondary-scrollable-filter" data-filter-search-target="optionsList">
+  <template id="filter-search-input-template">
```
Nitemaeric, line 24: Maybe this should be a stimulus target like the other file?

Nitemaeric (commented):
I think these are final comments.

app/javascript/find/controllers/filter_search_controller.js
```diff
@@ -1,58 +1,39 @@
...
-    return container
+    this.searchInputContainerTarget.innerHTML = html
   }
 
   search () {
     const optionItems = this.optionsListTarget.children
-    const searchValue = this.searchInputTarget.value.toLowerCase()
+    const searchInput = this.searchInputContainerTarget.querySelector('input')
 
+    const searchValue = searchInput.value.toLowerCase()
```
Nitemaeric, line 30:
```suggestion
    const searchValue = this.searchInputTarget.value.toLowerCase()
```

Nitemaeric (approved):

PR #5186, opened 2025-04-24
Geolocation cache control

inulty-dfe:
## Context

We have a cache mechanism for storing geocoding requests. We do not have anyway to manage or control this cache.

If we make an update to how the cache is built, we cannot reset the cache easily.

Add a set of pages that allow us to reset the geolocation cache from the support interface

## Checklist


inulty-dfe:
We may never need this

PR #5188, opened 2025-04-25
Add ability to show status of rollover progress

tomas-stefano:
## Context

As the process of the RolloverService is a long-running process, we want to be able to see how much progress has been made and how much is remaining.

The design should be simple. Just a summary list item that reads:

`3 of 823 providers (0.36%)`

No live/action cable - a simple display of progress information.

Screenshot 2025-04-25 at 10 57 21

## Guidance to review

1. Visit `/support/recruitment_cycles`
2. Does it works for past cycles?
3. Create a new cycle
4. Wait some minutes.
5. It is copying the providers? 
6. Refresh the page
7. It the numbers changing?


app/components/rollover_progress_component.rb
```diff
@@ -0,0 +1,34 @@
+class RolloverProgressComponent < ViewComponent::Base
+  attr_reader :current_cycle, :target_cycle
+
+  def initialize(current_cycle:, target_cycle:)
+    @current_cycle = current_cycle
+    @target_cycle = target_cycle
+
+    super
```
inulty-dfe, line 8: Call super before setting instance variables

inulty-dfe (changes requested):
Looks good except one comment

app/components/rollover_progress_component.rb
tomas-stefano: Ok, done.

PR #5200, opened 2025-04-28
Remove accredited partnership description step in support

inulty-dfe:
## Context

We are no longer displaying the Provider Partnership description in Find, we instead show the Accredited providers own "About us" content in it's place.

We are removing the step in the "Add accredited provider" flow where the users adds a description for the partnership.

1. Remove the AccredtiedProviderController actions `new`, `edit` and `update` which are used to add/edit the description.
2. Remove unused routes
3. Update the Support Search controller to redirect straight to the Checks controller
4. Update the backlinks on remaining pages
5. Remove ProviderPartnershipForm and ClearStashable, we don't need to stash any data anymore
6. Update Locales

## Guidance to review

I've moved the custom html from the publish interface into the `AccreditedProviderComponent`. This is a temporary implementation until it can be improved.

https://github.com/user-attachments/assets/989f04bd-0c2d-4c3c-9db6-4f241ac50c0e

## Trello

https://trello.com/c/eBqJPcqd/628-remove-the-description-step-in-creating-and-editing-provider-partnerships

## Checklist


app/controllers/support/providers/accredited_partnerships/checks_controller.rb
```diff
@@ -4,42 +4,36 @@ module Support
...
+          @partnership = provider.accredited_partnerships.build(accredited_provider_id: params[:provider_partnership][:accredited_provider_id])
+
           if @partnership.save
             notify_accredited_provider_users
-
+            flash[:success_with_body] = { "title" => t(".added"), "body" => @partnership.accredited_provider.provider_name }
             redirect_to support_recruitment_cycle_provider_accredited_partnerships_path(
               recruitment_cycle.year, provider.id
-            ), flash: { success: t(".added") }
+            )
```
Nitemaeric, line 23: Considering that line 11 puts the flash inline with the redirect, I think it might be worth deciding on a consistent format.

Nitemaeric (approved):
Nice one. There's so much code to all this.

app/controllers/support/providers/accredited_partnerships/checks_controller.rb
inulty-dfe: Yeah, I think setting the flash should be on it's own line to make it a first class operation. There's times when it's short and can conveniently be added to the redirect, but to be consistent, set it separately.

PR #5202, opened 2025-04-28
Clean up route constraints and reorganise routes

Nitemaeric:
## Context

When performing the Disaster Recovery exercise, we found that the application wouldn't respond to the "`temp`" hosts.

After inspection, it was found that 

- Remove `ApiConstraint`, `PublishConstraint`, and `FindConstraint` classes. We can perform a host constraint by simply doing `constraints: { host: array_of_valid_hosts }`.
- Remove `:with_find_constraint` and `:with_publish_constraint`, along with their support files that stub constraints.
- Extract API routes into 2 sets, internal APIs and external APIs.
  - Internal APIs are available on both the `Find` domain and `Publish` domain under the `/api/*` path.
  - External APIs are available on the `Settings.api_hosts` domains.
- Fix various specs.

## Guidance to review

- Does the app work on your local machine?
  - Do you need to configure the URLs or hosts?

## Checklist


config/initializers/content_security_policy.rb
```diff
@@ -8,16 +8,10 @@
...
 Rails.application.configure do
   config.content_security_policy do |policy|
-    local_domains = %w[https://publish.localhost https://find.localhost]
-    prod_domains = [
-      "https://*.find-teacher-training-courses.service.gov.uk",
-      "https://*.find-postgraduate-teacher-training.service.gov.uk",
-      "https://*.publish-teacher-training-courses.service.gov.uk",
-    ]
-    all_domains = local_domains + prod_domains
+    all_domains = [*Settings.find_hosts, *Settings.publish_hosts, *Settings.api_hosts]
```
Nitemaeric, line 11: API needs CSP headers, as the API docs are hosted on the `api_url`.

config/initializers/content_security_policy.rb
```diff
@@ -35,7 +29,7 @@
     policy.style_src   :self, *all_domains
 
     policy.frame_src   :self,
-                       *local_domains
+                       *all_domains
```
Nitemaeric, line 32: This may be worth removing.

config/initializers/cors.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
-  origin_urls = [Settings.base_url, Settings.find_url, Settings.extra_find_url].compact_blank
+  origin_urls = [Settings.find_url, Settings.publish_url].compact_blank
```
Nitemaeric, line 4: Allow APIs to be called from these domains.

config/initializers/content_security_policy.rb
tomas-stefano: Is it worth thinking to be strict like `policy.frame_ancestors :none`?

Nitemaeric: We could, yeah. Especially as we're not using embedded iframes.

tomas-stefano: Also frame_src is deprecated in CSP Level 2 in favor of frame-ancestors. See https://www.w3.org/TR/CSP2/#directive-frame-src

spec/rails_helper.rb
```diff
@@ -124,17 +124,53 @@
 
   ActiveJob::Base.queue_adapter = :test
 
+  config.before(:each, type: :request) do
+    service = self.class.metadata[:service]
+
```
tomas-stefano, line 129: Should we raise if is nil/blank?

tomas-stefano (approved):
After the build is green :)

PR #5203, opened 2025-04-28
Increase Capybara default max wait time to 10 seconds

inulty-dfe:
## Context

We're seeing a lot of stale element failures. Let's try change the Capybara config to help the tests pass.

Increase Capybara default max wait time to 10 seconds

If an element takes 3 seconds to appear, the test won't fail,
But if it fails for a good reason, it might take 10 seconds for it to fail.

Tradeoff.

## Checklist


inulty-dfe:
Deciding against this, lets monitor the flaky test failures

PR #5204, opened 2025-04-28
Make the table full width on recruitment cycles page

tomas-stefano:
## Context

Make the table full width on recruitment cycles support page.

### DIFF
--- a/app/views/support/recruitment_cycles/index.html.erb
+++ b/app/views/support/recruitment_cycles/index.html.erb
@@ -1,37 +1,33 @@
 <% content_for :page_title, t(".page_title") %>
 
 <div class="govuk-width-container">
-  <div class="govuk-grid-row">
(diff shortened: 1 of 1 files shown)


app/views/support/recruitment_cycles/index.html.erb
```diff
@@ -2,36 +2,34 @@
 
 <div class="govuk-width-container">
   <div class="govuk-grid-row">
```
Nitemaeric, line 4: We should remove this div. As far as I'm aware, the `govuk-grid` classes work similarly to Bootstrap, which the `row` class adds negative margins.

Nitemaeric (changes requested):

app/views/support/recruitment_cycles/index.html.erb
tomas-stefano: Sounds good!

PR #5207, opened 2025-04-29
Adjust model specs for faster CI runs

inulty-dfe:
## Context

Our CI distributes test files to test runners in order to balance the time it takes to run the full suite.

One of the files takes much longer than the others. This PR splits the test examples from that file into other files so the test suite can be better balanced.

I started thinking it was the course_spec taking a long time, but it turned out to be a few tests in the provider_spec.
I've left the changes in the course_spec in the PR as they are not a bad idea IMO.

Split up the `course_spec.rb`
stop using `is_expected` in `provider_spec.rb`, for some unknown reason each test took ~55 seconds to run.
I've combined 3 of the assertions into one example and they are much quicker.

|Before|After|
|---|---|
|4m23s|1m16s|
|image|image|

## Trello 
https://trello.com/c/UKVVrOOU/684-split-up-coursespecrb-for-faster-ci-runs

## Checklist


spec/models/course/scopes_spec.rb
```diff
@@ -0,0 +1,853 @@
...
+          name: "Course D",
+          provider: provider_b,
+        )
+      end
+
+      before do
+        course_d
+        course_c
+        course_a
+        course_b
```
tomas-stefano, line 82: Can we use `let!`?

spec/models/course/scopes_spec.rb
```diff
@@ -0,0 +1,853 @@
...
+    describe ".accredited_provider_order" do
+      subject { described_class.accredited_provider_order(provider.provider_name) }
+
+      let(:provider) { create(:provider) }
+      let(:delivered_course) { create(:course, provider:) }
+      let(:accredited_course) { create(:course, accrediting_provider: provider) }
+
+      before do
+        accredited_course
+        delivered_course
```
tomas-stefano, line 193: Also we can use `let!`

tomas-stefano (approved):

spec/models/course/scopes_spec.rb
inulty-dfe: I think they're called in this order so they're out of order before testing that the scope sorts them. I'ts not as simple as changing to `let!` I don't think

PR #5208, opened 2025-04-29
Specify the recruitment cycle when running automated rollover

tomas-stefano:
## Context

Originated from this PR comment:
https://github.com/DFE-Digital/publish-teacher-training/pull/5154#discussion_r2039199331

When running the rollover will be good to specify the recruitment cycle.

### DIFF
--- a/app/services/recruitment_cycle_creation_service.rb
+++ b/app/services/recruitment_cycle_creation_service.rb
@@ -10,12 +10,12 @@ class RecruitmentCycleCreationService
   end
 
   def call
-    RecruitmentCycle.create!(
+    recruitment_cycle = RecruitmentCycle.create!(
       year: @year,
       application_start_date: @application_start_date,
       application_end_date: @application_end_date,
     )
 
-    RolloverJob.perform_later
+    RolloverJob.perform_later(recruitment_cycle.id)
   end
 end
--- a/app/jobs/rollover_job.rb
+++ b/app/jobs/rollover_job.rb
@@ -5,12 +5,12 @@ class RolloverJob < ApplicationJob
 
   BATCH_SIZE = 10
 
-  def perform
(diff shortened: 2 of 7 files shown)


tomas-stefano:
https://trello.com/c/g8uGzmpn/647-make-sure-we-can-parameterize-the-recruitment-cycle-during-the-rollover-automation

app/services/recruitment_cycle_creation_service.rb
```diff
@@ -10,12 +10,12 @@ def initialize(year:, application_start_date:, application_end_date:)
...
   def call
-    RecruitmentCycle.create!(
+    recruitment_cycle = RecruitmentCycle.create!(
       year: @year,
       application_start_date: @application_start_date,
       application_end_date: @application_end_date,
     )
 
-    RolloverJob.perform_later
+    RolloverJob.perform_later(recruitment_cycle.id)
```
Nitemaeric, line 19: I wonder if we should add defensive measures here? If a user creates a `RecruitmentCycle` with the year being 2-3 years in the future, should the rollover script be run?

tomas-stefano: > I wonder if we should add defensive measures here? If a user creates a `RecruitmentCycle` with the year being 2-3 years in the future, should the rollover script be run?

If we wanna go this way, I'd recommend to not run the rollover at all when a cycle is created and we should add a button "Run rollover" with confirmation - on a recruitment cycle index/show action.

Nitemaeric: 👍 I think I'm good with that too. We can be explicit.

tomas-stefano: I will create a card then to add the button with confirmation and merge this PR.  Thanks!

PR #5217, opened 2025-05-01
Replace Microsoft feedback form with gov style form

sarahcrack:
## Context

Trello: https://trello.com/c/Ufq7dZxl/685-create-feedback-form-replace-microsoft-form-with-govuk-look-and-feel-form

Previously we worked to turn our feedback form into a Microsoft form (it was an email address that people sent feedback into) but now we’d like to make it a proper service based form where feedback can be seen on the service via the support console.

This PR creates the feedback form, feedback table and associated tests (linking the feedback to the support console will be done in another ticket). 
- add Feedback model with `ease_of_use` (string) and `experience` (text) fields
- create service feedback form (view, FeedbacksController, routes, add to analytics, add and organise locales for feedback form content and error messages)
- add error handling for blank submissions and success flash message on submission of feedback form
- create system test for feedback form and validation test
- swap Microsoft form link in header to new feedback form and change the behaviour to open form in the same tab
- create a feature flag (alongside a helper) to hide the gov style form in production until the work has been done to link the responses to the support console
- move all content into new locales file: `config/locales/en/find/feedbacks/new.yml`

https://github.com/user-attachments/assets/76da4795-ed48-4ca9-8291-d1c3cb6f3871

## Guidance to review
- feature flag currently hides the new gov style form in all environments except for `test`, `development` and `review` where the new gov style form is shown/used. To toggle the feature flag to show/hide the new form you can go to `config/settings.yml` and set `gov_style_feedback_form_enabled:` to `true` or `false`. When the feature flag is set to false, the old Microsoft form and 'open in new tab' behaviour will remain. 
- does the form work as it should, i.e. completing it successfully compared to when a user misses a question/tries to submit
- is the content of the form correct, including error messages?
- do responses successfully save to the database?
- are there any parts of the code that could be adapted or refactored to follow best practice?

## Checklist

- [x] I have created a feedback table in the database
- [x] I have created the feedback form that saves responses in the database
- [x] I have added appropriate error handling
- [x] I have created relevant tests
- [x] I have added the new feedback form link to the service
- [x] I have created a feature flag to hide the gov style form until the work has been done to link the responses to the support console.


app/helpers/find/feedback_form_helper.rb
```diff
@@ -0,0 +1,13 @@
+module Find::FeedbackFormHelper
+  def feedback_url
+    if Settings.features.gov_style_feedback_form_enabled
+      "./feedback/new"
```
Nitemaeric, line 4: We should use Rails's url helpers for internal urls.

```ruby
new_feedback_path
```

config/routes/find.rb
```diff
@@ -40,10 +40,10 @@
 
   resource :cookie_preferences, only: %i[show update], path: "/cookies", as: :cookies
   resource :sitemap, only: :show
-
   scope via: :all do
     match "/404", to: "errors#not_found"
     match "/500", to: "errors#internal_server_error"
     match "/403", to: "errors#forbidden"
   end
+  resource :feedbacks, only: %i[new create], path: "/feedback", as: :feedback
```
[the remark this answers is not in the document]
Nitemaeric: I think we should. The path should only be accessible if the flag is enabled.

Nitemaeric (commented):
As this will be considered development complete when merged, we should also enable this in the `development` environment.

app/controllers/find/feedbacks_controller.rb
```diff
@@ -0,0 +1,30 @@
+module Find
+  class FeedbacksController < ApplicationController
+    def new
+      @feedback_form = FeedbackForm.new
+    end
+
+    def create
+      @feedback_form = FeedbackForm.new(feedback_form_params)
```
Nitemaeric, line 8: I think we can remove this form object. We don't need to validate anything in addition to the model validations itself.

config/locales/en/find/feedbacks/new.yml
```diff
@@ -0,0 +1,29 @@
...
+  activemodel:
+    errors:
+      models:
+        find/feedback_form:
+          attributes:
+            ease_of_use:
+              blank: "Select how easy it is to find courses relevant to you"
+            experience:
+              blank: "Enter your feedback"
+              too_long: Feedback is too long (maximum is 1200 characters)
```
Nitemaeric, line 29: This should be in the `activemodel.yml` file.

inulty-dfe (changes requested):
I've added some comments. I'll review again after we discuss those.

app/forms/find/feedback_form.rb
```diff
@@ -0,0 +1,10 @@
+module Find
+  class FeedbackForm
```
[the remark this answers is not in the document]
sarahcrack: have now removed FeedbackForm object as per Dan's suggestion above 🙂

config/locales/en/find/feedbacks/new.yml
```diff
@@ -0,0 +1,29 @@
+en:
+  find:
+    feedbacks:
+      new:
+        title: Feedback
+        heading: Give feedback on Find teacher training courses
+        error_summary_title: There is a problem
```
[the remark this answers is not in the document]
sarahcrack: Turns out you can remove the title completely as it's included as part of the govuk_error_summary :)

app/controllers/find/feedbacks_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+        redirect_to find_root_path
+      else
+        render :new
+      end
+    end
+
+  private
+
+    def feedback_form_params
+      params.fetch(:feedback, {}).permit(:ease_of_use, :experience)
```
inulty-dfe, line 22: Where did the use of fetch come from here? We can remove this as we expect the form will be feedback and the attributes will be within that `feedback[ease_of_use]=very_easy`

app/helpers/find/feedback_form_helper.rb
```diff
@@ -0,0 +1,13 @@
+module Find::FeedbackFormHelper
```
inulty-dfe, line 1: Not sure what to do with these methods now the form is gone.

config/locales/en/find/feedbacks/new.yml
```diff
@@ -0,0 +1,18 @@
+en:
```
inulty-dfe, line 1: I think we can scope the files to the controller rather than the action.

`feedbacks.yml` instead of `feedbacks/new.yml`

config/routes/find.rb
```diff
@@ -36,10 +36,13 @@
...
-
   scope via: :all do
     match "/404", to: "errors#not_found"
     match "/500", to: "errors#internal_server_error"
     match "/403", to: "errors#forbidden"
   end
+
+  if Settings.features.gov_style_feedback_form_enabled
+    resource :feedbacks, only: %i[new create], path: "/feedback", as: :feedback
+  end
```
inulty-dfe, line 47: I'd group this with other `resource` calls above the error routes above.

Move it below `resource :sitemap`

inulty-dfe (commented):
Just a couple small things to change, nothing major and I think it's good to go

app/helpers/find/feedback_form_helper.rb
sarahcrack: thanks @inulty-dfe! I have just renamed to FeedbacksHelper if that works? We just need this helper until the feature flag is no longer needed, then it can be removed.

app/models/feedback.rb
```diff
@@ -0,0 +1,6 @@
+class Feedback < ApplicationRecord
+  enum :ease_of_use, %w[very_easy easy neither_easy_nor_difficult difficult very_difficult].index_by(&:itself)
+
+  validates :ease_of_use, presence: true
```
inulty-dfe, line 4: We should validate that the value of this is in the enum

inulty-dfe (approved):
Awesome work @sarahcrack :tada: :rocket:

PR #5220, opened 2025-05-02
Change internal content referring to Find and Apply

gms-gs:
## Context

Change the internal use of terms like Find and Apply on Publish because users do not necessarily know that language.

The references to find or changed for the new content found on this Lucid board [Change content to not use Find or Apply : Lucidspark](https://lucid.app/lucidspark/99751e9b-a00e-46e4-aa15-ef914fb814c9/edit?page=0_0&invitationId=inv_7d759170-4464-4599-8b0c-cdb4f7d77d0b#)

- Update content
- Move raw strings into translations (Apart from `CourseDecorator` as we plan to kill this off)

I have also removed some cases where `data-qa` attributes that are used for testing, when trying to remove them all it caused a lot of failures and needs a much larger rework as lots of other specs rely on these attributes and is outside the scope of this ticket. We can either do this in one big bang, create a dedicated ticket for this, or we slowly remove them as we touch more of this area of the app, as Publish side has not had as much love.

## Checklist

### DIFF
--- a/app/views/publish/courses/confirmation.html.erb
+++ b/app/views/publish/courses/confirmation.html.erb
@@ -26,13 +26,13 @@
 
       <%= govuk_summary_list(html_attributes: { data: { qa: "course__details" } }) do |summary_list| %>
 
-        <% summary_list.with_row(html_attributes: { data: { qa: "course__level" } }) do |row| %>
+        <% summary_list.with_row do |row| %>
           <% row.with_key { t(".subject_level") } %>
           <% row.with_value { course.level.humanize } %>
           <% row.with_action %>
         <% end %>
 
-        <% summary_list.with_row(html_attributes: { data: { qa: "course__is_send" } }) do |row| %>
+        <% summary_list.with_row do |row| %>
           <% row.with_key { t(".send") } %>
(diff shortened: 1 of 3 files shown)


app/decorators/course_decorator.rb
```diff
@@ -261,7 +261,7 @@ def age_range_in_years_and_level
 
   def applications_open_first_label(recruitment_cycle)
     if current_cycle?
-      "As soon as the course is on Find - recommended"
+      "As soon as the course is published - recommended"
```
tomas-stefano, line 264: Can we move this to translations?

app/decorators/course_decorator.rb
```diff
@@ -31,7 +31,7 @@ def description
   def on_find(provider = object.provider)
     if object.findable?
       if current_cycle_and_open?
-        h.govuk_link_to("Yes - view online", h.find_course_url(provider.provider_code, object.course_code))
+        h.govuk_link_to("View live course", h.find_course_url(provider.provider_code, object.course_code))
```
tomas-stefano, line 34: Can we move the link title to translations?

tomas-stefano (approved):

PR #5243, opened 2025-05-12
Use "available in publish from" field from recruitment cycle

tomas-stefano:
## Context

This PR uses the new field on recruitment cycle to display upcoming courses based on the recruitment cycle date field “available in publish from“.

Implement logic to filter and display only relevant future cycle courses.

Requirements:

Create a filtering mechanism that uses the available in publish from field to identify and display next cycle courses

Implement conditional display logic to hide current cycle if its date is nil

Add sorting functionality to present courses in chronological order by cycle date

## Acceptance Criteria

Accurately displays courses from upcoming recruitment cycles based on the date field from recruitment cycle.

Same feature from can_edit_current_and_next_cycles: true see https://github.com/DFE-Digital/publish-teacher-training/blob/main/guides/rollover.md  for reference

## What this PR doesn't do?

* Remove the feature flag `can_edit_current_and_next_cycles` this will done **after** this PR.
* Remove from the rollover docs (the docs will be updated when the automation rollover works)


spec/components/support_title_bar_spec.rb
```diff
@@ -21,7 +21,7 @@
 
   context "during rollover" do
     before do
-      allow(Settings.features.rollover).to receive(:can_edit_current_and_next_cycles).and_return(true)
+      create(:recruitment_cycle, :next, available_in_publish_from: 1.day.ago)
```
inulty-dfe, line 24: Would you consider stubbing `RecruitmentCycle.next_editable_cycles?` instead of creating a record just for this test? I'd like to think we can find ways to avoid our test suite getting really slow

inulty-dfe (commented):
Couple of concerns about test suite perfomance and readability

spec/components/support_title_bar_spec.rb
tomas-stefano: Done

inulty-dfe (approved):
Lets gooo

PR #5244, opened 2025-05-12
Add callout to GIT Funding and support if you're disabled page

gms-gs:
## Context

User research from GIT has suggested that people do not know what pages exist to help support disabled users. We also do not link to any pages to support disabled users. So at the point of need we should add a link to GITs help for funding and support for our users.

Add new call out box to provide support information

## Guidance to review

- Visit a course 
- Scroll down to `training with disabilities` section and select the link to find out more
- You should see the new call out box

Screenshot 2025-05-12 at 19 04 12

## Checklist


kelliedesigner (approved):
LGTM

PR #5259, opened 2025-05-15
Add Candidate model

inulty-dfe:
## Context

We are adding a Candidate model so users of Find can save courses and make use of other benefits of storing records associated with an account.

Add Candidate model

## Trello 

https://trello.com/c/l1j54Wdr/710-add-candidate-model

## Checklist


tomas-stefano (approved):
🚀

PR #5281, opened 2025-05-20
Ensure that saving GiasSchool creates pg_search_vectors

inulty-dfe:
## Context

Gias Schools are not searchable because when we save them, we aren't running the ActiveRecord Callbacks.

The CSV we receive has open schools with blank columns that we are currently validating the presence of.

We decided to use `upsert_all` because it was fast. But this is in a background job, run at night, once a week.
We will save each GiasSchool with callbacks that has changed during the import.
Remove presence validation on columns that are nullable in the source CSV.

## Trello
https://trello.com/c/5q1tqcZ8/725-gias-schools-are-not-searchable

## Checklist

### DIFF
--- a/lib/gias/importer.rb
+++ b/lib/gias/importer.rb
@@ -15,7 +15,13 @@ module Gias
 
       school_records = CSV.foreach(csv_path, headers: true).map(&:to_h)
 
-      GiasSchool.upsert_all(school_records, unique_by: :urn)
+      school_records.each do |school|
+        gs = GiasSchool.find_or_initialize_by(urn: school["urn"])
+        gs.assign_attributes(school)
+        if gs.changed?
+          gs.save!
+        end
+      end
 
       Log.log("Gias::Importer", "GIAS Data Imported!")
     end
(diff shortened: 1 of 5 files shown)


lib/gias/importer.rb
```diff
@@ -15,7 +15,21 @@ def call
...
+        begin
+          gs.save!
+        rescue ActiveRecord::RecordInvalid => e
+          Sentry.set_extras(
+            model: e.record.class.name,
+            attributes: e.record.attributes,
+            errors: e.record.errors.full_messages,
+            error_details: e.record.errors.details,
+          )
+          Sentry.capture_exception(e)
```
tomas-stefano, line 30: If thousands of transactions fail we will be alerting too much Sentry 🤔 🤔 🤔 I guess there is no escape in it?

tomas-stefano (approved):
Minor question to consider. Not blocking.

lib/gias/importer.rb
inulty-dfe: Yeah good point

tomas-stefano (approved):

PR #5284, opened 2025-05-21
Remove visa sponsorship deadline feature flag

CatalinVoineag:
## Context

This feature flag is not necessary anymore. The visa sponsorship deadline feature is permanent.

Removed `visa_sponsorship_deadline` feature flag where it's used

## Checklist

### DIFF
--- a/spec/features/publish/courses/new_tda_course_spec.rb
+++ b/spec/features/publish/courses/new_tda_course_spec.rb
@@ -186,6 +186,7 @@ feature "Adding a teacher degree apprenticeship course" do
     when_i_click_continue
     and_i_click_continue
     and_i_choose_to_sponsor_a_student_visa
+    and_i_choose_no_visa_deadline
     when_i_click_on_add_a_course
     then_i_see_the_correct_attributes_in_the_database_for_fee_paying
   end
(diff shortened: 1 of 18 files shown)


spec/features/publish/courses/new_tda_course_spec.rb
```diff
@@ -186,6 +186,7 @@
     when_i_click_continue
     and_i_click_continue
     and_i_choose_to_sponsor_a_student_visa
+    and_i_choose_no_visa_deadline
```
CatalinVoineag, line 189: Same here, the spec only passed because the feature flag was not set. Now we need to make a decision about the deadline

spec/features/publish/courses/new_visa_sponsorship_spec.rb
```diff
@@ -107,4 +109,12 @@ def when_i_choose_yes_for_skilled_worker_visa
...
     expect(page).to have_text("Check your answers")
   end
+
+  def and_i_select_no_student_visa
+    publish_courses_new_student_visa_sponsorship_page.no.click
+  end
+
+  def and_i_select_no_skillled_worker_visa
+    publish_courses_new_skilled_worker_visa_sponsorship_page.no.click
+  end
```
elceebee, line 119: The publish team is moving away from using these page objects, so I think these should be more like:
```suggestion
  def and_i_select_no_student_visa
     choose 'No'
  end

  def and_i_select_no_skillled_worker_visa
      choose 'No'
  end
```

inulty-dfe (approved):

PR #5289, opened 2025-05-23
Move Publish authentication resources to Publish module

inulty-dfe:
## Context

I noticed that all the DfE Signin, MagicLinks and Persona controllers were not namespaced.
I have moved them to a new `Publish::Authentication` module. This module encapsulates functionality shared by `Support` and `Publish`. Essentially, everyone logs into Publish, and some users can access support.

We will be implementing Authentication and Sessions for the Find service so this will help separate the Authentication system between Find and Publish/Support

Broken into 2 commits
**First Commit:**

1. The Authentication module is now a Concern
	1. Add `helper_method :current_user` and `before_action :authenticate` inside
	2. Move it out of the top level `ApplicationController`
	3. Include it directly into `Support::ApplicationController` (Already in `Publish::ApplicationController`)
	4. This removes Publish `Authentication` from being inherited by `Find::ApplicationController`
2. Move all Publish authentication controllers, views & services to `Publish::Authentication` namespace.
	1. SessionsController
	2. MagicLinksController
	3. SignInController
	4. UserSession

**Second Commit:**

 1. Move publish specific controllers into the Publish namespace.
	 1. `PagesController`
	 2. `CookiePreferencesController`.

## Trello
https://trello.com/c/V21YQURU/738-refactor-publish-support-authentication-in-preparation-for-find-authentication

## Checklist


app/controllers/concerns/platform/authentication.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Platform
+  module Authentication
+    extend ActiveSupport::Concern
+
+    included do
+      helper_method :current_user
+      before_action :authenticate
```
tomas-stefano, line 9: Should this module have `protect_from_forgery` ?

app/controllers/platform/application_controller.rb
```diff
@@ -0,0 +1,5 @@
+module Platform
+  class ApplicationController < ::ApplicationController
+    include Authentication
+  end
+end
```
tomas-stefano, line 5: I am not sure this controller add any value if we can include the module on both controller really 🤔

app/controllers/concerns/platform/authentication.rb
inulty-dfe: Good spot!

I can't find anywhere this is called in the code base right now.

tomas-stefano: True!! I forgot about that!

PR #5293, opened 2025-05-27
Upgrade QA DB

tomas-stefano:
## Context

QA is very slow and affecting smoke tests and deployment.

So let's upgrade postgres plan - this will help to unblock pipeline and make QA faster


inulty-dfe (commented):
:rocket:

RMcVelia (approved):

PR #5295, opened 2025-05-28
How placements work is missing styling on what we include in the section

sarahcrack:
## Context

Trello: https://trello.com/c/mlbayHJP/687-how-placements-work-is-missing-styling-on-what-we-include-in-the-section

On Publish, when you click the change link for 'How placements work', you come to a page where providers can enter information, and there is a details component to see what information would already be included on the page. The information in there is missing styling and should have the grey background and orange bar.

- the styling used for this section was coming from `_callout.scss` which was situated within the stylesheets/find folder and only imported within the stylesheet for Find; the styling was therefore not being applied within Publish.
- move `_callout.scss` out of Find folder and up one level to `stylesheets/components`; this makes sense as it is now being used by both services.
- import `_callout.scss`in `stylesheets/publish/application.scss` so the styling is applied correctly.
- adapt `_callout.scss` import path in `stylesheets/find/application.scss` following the move of the file to the components folder.

How the section looks following changes above:
Screenshot 2025-05-27 at 14 53 01

## Guidance to review
- On Publish sign in as Anne, Susy or Mary.
- Click on one of the courses in the list to view
- Scroll down to 'How placements work' section and click 'Change'.
- Click on 'See what we include in this section' and the callout box with correct styling should appear.

## Checklist

- [x] I have applied the correct styling to the callout box in Publish.


kelliedesigner:
Looks good to me!

PR #5320, opened 2025-06-09
Add primary navigation

gms-gs:
## Context

As part of the candidate accounts work, we want to add a primary navigation to allow candidates to navigate to search for courses and to view their saved courses.

We want to closly align with Apply design:
Screenshot 2025-06-11 at 13 34 12

Screenshot 2025-06-11 at 13 34 24

- Add primary navigation links (Using Apply as a reference, and you can see there is no primary navigation when the user is not logged in)
- Add saved courses page skeleton 
- Add redirecting of the user if the user attempts to visit the saved courses page without logging in

## Guidance to review

- Visit review app
- Enable feature flag if not activated
- Login
- Visit saved courses page
- Logout
- You should not see the primary navigation links
- You should not be able to visit the saved courses page

## Checklist

### DIFF
--- a/spec/system/find/candidates/saved_courses_spec.rb
+++ b/spec/system/find/candidates/saved_courses_spec.rb
@@ -1,8 +1,11 @@
 require "rails_helper"
 
 RSpec.describe "View pages" do
+  before do
+    FeatureFlag.activate(:candidate_accounts)
+  end
+
   scenario "As a Candidate, I can visit my saved courses" do
-    given_candidate_accounts_is_active
     when_i_visit_the_find_homepage
     then_i_see_a_login_button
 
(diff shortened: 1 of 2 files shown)


config/locales/en/layouts/find.yml
```diff
@@ -3,3 +3,6 @@ en:
     find:
       feedback: This is a new service &ndash; %{link_to} will help us to improve it.
       feedback_text: your feedback
+      primary_nav:
+        courses_html: "<strong>Courses</strong>"
+        saved_courses_html: "<strong>Saved courses</strong>"
```
inulty-dfe, line 8: This doesn't look right, but I know we have work to do on the HeaderComponent so we can address that later.

inulty-dfe (approved):

PR #5326, opened 2025-06-10
Change the accredited providers list to a table and adapt removal of accredited provider journey

sarahcrack:
## Context

Trello: https://trello.com/c/6rECXw4x/652-change-the-accredited-providers-list-to-a-table

We want to update the accredited providers list on Publish into a table format to align with the standard design of other pages e.g. Courses tab. 

- We should have a single column, “Name". This should be formatted as the Provider’s name with code. It should link to a “details” page of the accredited partner with only the provider name in the list.
- On the details page, it should show the accredited provider details and a remove link.
- Clicking on “Remove accredited provider” should show a success page.
- Only when confirmed does the partnership become removed.

- change accredited provider list for a training provider into a table: create 2 partials to render table format (using `govuk_components`) for accredited providers list in Publish. Removed the 'remove' button from the table list. 
- create new details page for accredited provider (show page) and add `show` route 
- add remove button to details page and remove from accredited provider list view
- create a new component to render accredited provider details; render the `AccreditedProviderDetails` component from the accredited partnerships show page
- add spec for `AccreditedProviderDetails` component
- add warning text to `can_remove` view as per content guidance
- adapt current feature test for accredited partnerships to account for the new journey in viewing/removing an accredited provider (via accredited provider details page)

_How the accredited provider list page used to look in Publish (please note it still looks like this in Support):_

Screenshot 2025-06-10 at 13 01 58

_How the 'remove accredited provider' journey looks now 
(N.B. the training provider is only exposed to a limited number of details about the accredited provider):_

https://github.com/user-attachments/assets/bf43d723-249e-44c9-9e8f-1dc0412f598a

## Guidance to review

PUBLISH
- login as Colin and search for Anthem SCITT (you can only remove an accredited provider if the training provider has no published courses linked to that accredited provider on Find e.g. has no courses listed).
- click on 'Accredited Provider' tab to view new accredited provider table
- click on an accredited provider to navigate to the details page (please check these are the details that need to be displayed - we can add or remove details here)
- click on 'Remove accredited provider' and confirm it has been removed from the table
- add an accredited provider back (just to double check the flow has not been affected). 

## Checklist

- [x] I have created a table to display the list of accredited providers
- [x] I have created a details page to display the accredited provider details, remove link and a back link
- [x] I have updated the content on the confirmation page to add a warning before removing the accredited provider
- [x] I have ensured that an accredited provider partnership can be deleted (and still added)
- [x] I have created relevant spec tests


app/components/providers/provider_list/view.html.erb
```diff
@@ -1,67 +1,98 @@
...
+       row.with_key { "Provider Code" }
+       row.with_value { @provider.provider_code }
 
+       row.with_action(
+         text: "Change",
+         href: edit_support_recruitment_cycle_provider_path(@provider.recruitment_cycle_year, @provider),
+         visually_hidden_text: "provider code",
+       )
+     end %>
+  <% component.with_row do |row|
```
inulty-dfe, line 21: This change has added a link to change the Provider code in Support. Was this intentional?

app/controllers/publish/providers/accredited_partnerships_controller.rb
```diff
@@ -5,7 +5,13 @@ module Providers
     class AccreditedPartnershipsController < ApplicationController
       helper_method :accredited_provider_id
 
-      def index; end
+      def index
+        @provider_partnerships = @provider.accredited_partnerships
+      end
+
+      def show
+        @provider_partnership = provider.accredited_partnerships.find_by!(accredited_provider: partner)
```
inulty-dfe, line 13: If you aren't using `@provider_partnership`, as in this case, it's ok to just set each provider to an instance variable. I know I told you different before, but this looks like a suitable exception.

```suggestion
        provider_partnership = provider.accredited_partnerships.find_by!(accredited_provider: partner)
        @accredited_provider = provider_partnership.accredited_provider
        @training_provider = provider_partnership.training_provider
```

app/views/publish/providers/accredited_partnerships/index.html.erb
```diff
@@ -1,30 +1,22 @@
...
 </div>
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
     <% if @provider.accredited_partnerships.none? %>
       <p class="govuk-body"><%= t(".no_providers", provider_name: @provider.provider_name) %>
       </p>
     <% else %>
-      <% @provider.accredited_partnerships.each do |partnership| %>
-        <%= render AccreditedProviderComponent.new(
```
inulty-dfe, line 23: Is this component being used by anything anymore?

inulty-dfe (commented):
Few questions / suggestions

app/views/publish/providers/accredited_partnerships/index.html.erb
sarahcrack: yes it's still being used on the support side as I haven't changed the accredited provider list into a table there - still looks like this: 

Screenshot 2025-07-04 at 12 00 43

app/components/providers/provider_list/view.html.erb
inulty-dfe: Yeah I understood that, but why has this file changed at all then if it's not being used by the accredited provider page?

sarahcrack: ah yes thanks @inulty-dfe I see what you mean! It was some leftover layout/styling changes from when I initially adapted the component to use in Publish. Have reverted all the changes for this file now so it's the same as main 👍

inulty-dfe (approved):
Great work :star2:

PR #5335, opened 2025-06-13
Refactor database restore script and add --no-download flag

tomas-stefano:
## Context

This PR refactors the database restore Bash script to improve readability. 

The script now supports a --no-download flag, allowing us to skip the backup download and decompression steps if the SQL file is already available locally.

## Changes

Strict mode: Enables set -euo pipefail for robust error handling.
Modularization: Breaks the script into clear functions for each major operation.
Argument parsing: Adds a --no-download flag and a help option, with usage instructions.
Help: Run with --help to see usage instructions.

### DIFF
--- a/bin/restore-backup
+++ b/bin/restore-backup
@@ -4,46 +4,106 @@
 # Drop and recreate the development and test databases
 # Restore the development database with today's sanitized backup.
 
-set -e
+set -euo pipefail
 
 log() {
   echo "$(date +'%Y-%m-%d %H:%M:%S') - $1"
 }
 
-log "Starting backup download..."
+usage() {
+  cat <<EOF
+Usage: $0 [--no-download] [--help]
 
-az storage blob download \
-  --account-name s189p01pttdbbkpsanpdsa \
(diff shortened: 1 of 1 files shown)


bin/restore-backup
```diff
@@ -4,46 +4,106 @@
...
+}
 
-log "Backup download completed."
+NO_DOWNLOAD=0
 
-log "Dropping and recreating development database..."
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --no-download)
+      NO_DOWNLOAD=1
```
inulty-dfe, line 29: Any change we can change this to `DOWNLOAD=0`?
Leave the option as `--no-download`, but use the inverse inside the script?

inulty-dfe (commented):
Only one small suggestion.

Very happy to see this option

bin/restore-backup
tomas-stefano: Ok sounds good, will do after lunch :)

inulty-dfe (approved):

PR #5353, opened 2025-06-20
Require system_retry_helper correctly

inulty-dfe:
## Context

I didn't commit the `.rb` on the end of the spec support file `require_relative`

This is required for some test runners that don't load the test files the same way as `bundle exec rspec` does.

## Checklist


inulty-dfe:
This actually works already and this PR is not needed....

PR #5358, opened 2025-06-20
Add new field to recruitment cycle (Part 1: Add new db column)

tomas-stefano:
## Context

During the course rollover process, support users need the ability to access next cycle course data in both the Support and Publish interfaces **before** it becomes available to general Publish users.

So I am adding a new field to represent that logic in a future PR.


db/schema.rb
```diff
@@ -393,6 +393,7 @@
     t.datetime "created_at", precision: nil, null: false
     t.datetime "updated_at", precision: nil, null: false
     t.date "available_in_publish_from"
+    t.date "available_for_support_users_date"
```
inulty-dfe, line 396:
```suggestion
    t.date "available_for_support_users_from"
```
?

inulty-dfe (commented):
I always wonder if we should only use datetime. What if we want to launch it at a specific time in the future?

inulty-dfe (approved):

PR #5359, opened 2025-06-20
Handle uncaught exceptions

inulty-dfe:
## Context

When a request is made to our app with an unexpected extension, we try to render a not found template, but the request is asking for a format that we don't provide.

Like, `find.com/x/y/z.json` will effectively try to render `404.json`. We crash when we can't find the matching template.

`include Errorable` in `Find::ApplicationController` and `ApplicationController` so it's available througout our controllers.
Move all `rescue_from` macros into the Errorable concern
Add 406 / not_acceptable error handling.
Add tests for 404, 406, 500 in Find and Publish

## Guidance to review

https://find-review-5359.test.teacherservices.cloud/results.json

image

We get a lot of these

image

This also fixes this

https://qa.publish-teacher-training-courses.service.gov.uk/publish/organisations/asdfasdf/2025/courses/J522

image

## Trello
https://trello.com/c/lFjkMINF/813-uncaught-exception-not-reported-to-sentry?filter=label:Bug%20%F0%9F%90%9E

## Checklist

### DIFF
--- a/spec/requests/find/errors_spec.rb
+++ b/spec/requests/find/errors_spec.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Testing Errors render", service: :find, type: :request do
+  describe "not_acceptable - GET /results.*" do
+    before do
+      create(:course, :secondary)
+    end
+
+    it "returns html response when json format is not found" do
+      get "/results.json"
+
+      expect(response).to have_http_status(:not_acceptable)
+      expect(response.body).to include("The format requested is not available")
+    end
+  end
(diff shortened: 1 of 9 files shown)


spec/requests/find/errors_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+    it "returns html response when json format is not found" do
+      get "/results.json"
+
+      expect(response).to have_http_status(:not_acceptable)
+      expect(response.body).to include("The format requested is not available")
+    end
+  end
+end
```
tomas-stefano, line 19: can we add a spec for each error?

PR #5365, opened 2025-06-23
Add excluded course param to search results

dcyoung-dev:
## Context

The Apply service would like to direct Candidates to "recommended courses" which would exclude courses that they have previously applied to. 

- Added the `excluded_courses` search param to the results page - this param is not used in the Find UI filters
- ~Controversially added a JSON response from the Results page.~

## Guidance to review

~The JSON response was put in to see how easy it could be. Ideally this would exist, allowing the Apply service to programatically check if there would be any results for a course recommendation query, before offering it to the Candidate.~

~This PR should act as a conversation on the possibilities/approach~

## Checklist

### DIFF
--- a/app/controllers/find/results_controller.rb
+++ b/app/controllers/find/results_controller.rb
@@ -53,7 +53,7 @@ module Find
         :subject_name,
         :university_degree_status,
         :applications_open,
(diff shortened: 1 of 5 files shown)


tomas-stefano (approved):
Small non blocking comment.

PR #5366, opened 2025-06-23
Add "available for support users" field to support interface

tomas-stefano:
## Context

During the course rollover process, support users need the ability to access next cycle course data in both the Support and Publish interfaces before it becomes available to general Publish users.

This PR adds this field to be set via support interface when creating a new cycle.

## Guidance to review

1. Enter on Support
2. Settings > Recruitment cycles -> Add recruitment cycle
3. Set the dates

**For now this work is only setup the field and not using. This will be part 3 (next PR!).;**

### DIFF
--- a/app/controllers/support/recruitment_cycles_controller.rb
+++ b/app/controllers/support/recruitment_cycles_controller.rb
@@ -21,6 +21,7 @@ module Support
           year: @support_recruitment_cycle_form.year,
           application_start_date: @support_recruitment_cycle_form.application_start_date,
           application_end_date: @support_recruitment_cycle_form.application_end_date,
+          available_for_support_users_from: @support_recruitment_cycle_form.available_for_support_users_from,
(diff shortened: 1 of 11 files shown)


inulty-dfe:
Does the date need to be validated that it is before some other date or after some other date?

tomas-stefano:
> Does the date need to be validated that it is before some other date or after some other date?

I am not sure I want to be that strict. Do you think I should do that this date should be before is available to publish users? I feel the user (probably Oli) will be conscious of this date. 🤔

inulty-dfe:
> > Does the date need to be validated that it is before some other date or after some other date?
> 
> I am not sure I want to be that strict. Do you think I should do that this date should be before is available to publish users? I feel the user (probably Oli) will be conscious of this date. 🤔

Yeah I think we should validate the form as much as we possibly can. If we know date A shouldn't be after date B, we should validate that, right?

PR #5373, opened 2025-06-26
Candidate can Sign In via One Login

inulty-dfe:
## Context

_Yesterday, July 1st, version all versions of the `omniauth-govuk-one-login` gem were yanked from RubyGems. I'm investeigating what is happening with this gem. For now, we're pulling the library in from the GitHub repo._

`Settings.one_login.enabled` must be true, for candidates can log in at `/auth/one-login`
If `Settings.one_login.enabled` is false, then candidates can log in at `/auth/find-developer`
This should never be true in Production.

|Environment| Settings.one_login.enabled|Details|
|---|---|---|
|Production|Yes|Obvious|
|Test|No|We cannot stub every request in system tests|
|Development|Can be|If you are testing the integration you can enable it. There will be documentation|
|Review|Can be|If you are testing the integration you can enable it. There will be documentation|

### Instructions will be added to documentation as part of this outcome

_This is informational only, this is all setup for the review app._

1. Create an integration environment [here](https://admin.sign-in.service.gov.uk/)
2. Create a key pair
3. Put the public key in the admin tool
4. Set the private key in the application with `export ONE_LOGIN_PRIVATE_KEY="$(cat private_key.pem)"`
5. Add redirect url to `http://find.localhost:3001/auth/one-login/callback` in the admin tool

### Update to how we handle errors.

Remove the `on_failure` callback and instead rely on OmniAuth redirecting errors to `/auth/failure` with a query string including `?message=some_error_happened&provider=one-login`

### Testing
This is the part I have spent most time on. I'm not totally sure how much I need to test.
OmniAuth doesn't really have a method for testing the strategy with you're application.

OmniAuth allows you to generate an error response which will render the `/auth/failure` endpoint with relevant params.
This is the best I think we can hope for. I've changed the error handling to depend directly on this functionality because I feel I couldn't test the `on_failure` callback as well as I'd like.

Allow users to sign in to Find as a Candidate authenticated by One Login.

## What this is not
 - Logout of One login
 - Recognise email changed for the user on login
 - check if user is already logged into One Login

## Trello

https://trello.com/c/3S0Q1nV9/823-sign-up-sign-in-with-one-login

## Checklist

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -56,6 +56,7 @@ if Settings.one_login.enabled
   begin
     private_key = OpenSSL::PKey::RSA.new(Settings.one_login.private_key.gsub('\n', "\n"))
   rescue StandardError => e
+    Sentry.capture_exception(e)
     Rails.logger.error(e)
   end
 
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -2,17 +2,15 @@
 
 require_relative "../../app/services/publish/authentication_service"
 
-if Publish::AuthenticationService.persona?
+OmniAuth.config.logger = Rails.logger
 
+if Publish::AuthenticationService.persona?
   Rails.application.config.middleware.use OmniAuth::Builder do
     provider :developer,
              fields: %i[uid email first_name last_name],
              uid_field: :uid
   end
 else
-
-  OmniAuth.config.logger = Rails.logger
-
   dfe_sign_in_issuer_uri = URI.parse(Settings.dfe_signin.issuer)
   dfe_sign_in_redirect_uri = URI.join(Settings.base_url, "/auth/dfe/callback")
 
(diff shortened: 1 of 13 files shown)


spec/requests/find/authentication_error_spec.rb
```diff
@@ -3,23 +3,23 @@
...
-
   before do
     create(:find_developer_candidate)
+    FeatureFlag.activate(:candidate_accounts)
 
     CandidateAuthHelper.mock_error_auth
   end
 
   describe "POST /auth/find-developer/callback" do
     it "triggers an error and renders the error page" do
```
langesi, line 14: Do you have a spec for the happy path? Sp when a candidate record is found/created, the session is set and the redirect is done correctly?

config/initializers/omniauth.rb
```diff
@@ -43,14 +41,44 @@
...
+                             URI::HTTP.build(host:, port:, path:)
+                           else
+                             URI::HTTPS.build(host:, port:, path:)
+                           end
+
+  # Load private key from environment variable
+  begin
+    private_key = OpenSSL::PKey::RSA.new(Settings.one_login.private_key.gsub('\n', "\n"))
+  rescue StandardError => e
+    Rails.logger.error(e)
```
langesi, line 59: Not sure if this is an issue, but if parsing fails, you log and continue, leaving private_key potentially nil. It’d be safer to re-raise so the app fails fast on misconfiguration

spec/requests/find/authentication_error_spec.rb
langesi: Oh sorry, missed that! Thanks :)

config/initializers/omniauth.rb
```diff
@@ -43,14 +41,45 @@
...
+  end
+elsif Rails.env.in?(%w[development test])
+  Rails.application.config.middleware.use OmniAuth::Builder do
+    provider(:find_developer,
+             name: "find-developer",
+             fields: %i[uid email],
+             uid_field: :uid,
+             path_prefix: "/auth",
+             callback_path: "/auth/find-developer/callback")
   end
```
tomas-stefano, line 84: I think is kind of a code smell to have those path strings here and in the code. 

Can we centralize everything in a object?

Example:

```ruby
# app/lib/auth_provider_config.rb
class AuthenticationConfig
  PROVIDERS = {
    one_login: {
      path: "/auth/one-login",
      callback_path: "/auth/one-login/callback"
    },
    find_developer: {
      path: "/auth/find-developer",
      callback_path: "/auth/find-developer/callback"
    }
    # Add more as needed
  }.freeze

  def self.path
    Settings.one_login.enabled ? PROVIDERS[:one_login][:path] : PROVIDERS[:find_developer][:path]
  end

  def self.callback_path
    Settings.one_login.enabled ? PROVIDERS[:one_login][:callback_path] : PROVIDERS[:find_developer][:callback_path]
  end

  def self.load_private_key
      # .... 
  end
end
```

Or alternatively, we can use the settings yaml to organise those and add all this options there? I strongly think we **should not hold any value** on this initializer.

PR #5384, opened 2025-07-01
Only show manually rollover button after is available to publish users

tomas-stefano:
## Context

Do not show the button "Roll over course" for publish users if "today's date" is before the recruitment cycle date that is available for publish users.

Only shows the button after the date is available for publish users.

Screenshot 2025-07-01 at 11 51 12

## Guidance to review

1. Run the rollover
2. Enter as provider user 
3. Enter on a provider page **on the current 2025 cycle**
4. See a rolled over courses from the current cycle (this course can be rolled over to the next cycle)
5. The button "Rollover course" should only appear after the recruitment_cycle.available_in_publish_from date.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -403,7 +403,7 @@ class Course < ApplicationRecord
   end
 
   def manually_rollable?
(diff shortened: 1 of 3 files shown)


langesi (approved):
LGTM

PR #5386, opened 2025-07-01
GOV.UK Rebrand

langesi:
## Context

There is an upcoming GOV.UK Rebrand.

- Enable the new rebrand by adding the `govuk-template--rebranded` css class.

## Guidance to review

- The new GOV.UK Rebrand must have gone live.
- The designs have been approved from our design team.


inulty-dfe:
@langesi I think we should add the deploy label to this so we can see it in a review app

PR #5389, opened 2025-07-02
Remove bug: Unexpected content when changing Modern Foreign language course in Publish

sarahcrack:
## Context

Trello: https://trello.com/c/nEFiztVI/779-bug-unexpected-content-when-changing-modern-foreign-language-course-in-publish

On the Modern Foreign Language course page in Publish there are a few things that need to be removed/edited:
1. There is an extra caption on the page saying 'Add course' which should not be there
2. There is a ^ symbol in both the error summary at the top and in the error message with the label
3.  There is an additional error message above the caption and H1 which should not be there

- remove the ^ symbol from error message in locale file
- remove the rendering of `error_messages` partial on the page which appears to add an additional error message which is not needed. The error message now only appears in the error summary and with the form label when a user does not select at least one language
- remove `CaptionText` from `form_fields` partial so that 'Add course' is not displayed when changing an existing course

'Add course' has been removed from the language selection page:
Screenshot 2025-07-02 at 13 05 47

The additional error message above the main content has been removed as well as the ^ symbol:
Screenshot 2025-07-02 at 13 06 21

## Guidance to review

- In Publish select a modern foreign language course from any provider
- Click on 'Basic details' tab
- On the subject row, click 'Change'
- Then click 'Update subject' on the next page
- QA: When changing an existing course, the caption on the Languages selection page should be of the course name and course code, and not display 'Add course'.
- QA: to check error messages, deselect all languages and click 'Save and publish changes'. When getting the error message, it should not contain the ^ symbol, and should only have the error message in the error summary and with the form label.

## Checklist

### DIFF
--- a/app/views/publish/courses/modern_languages/_form_fields.html.erb
+++ b/app/views/publish/courses/modern_languages/_form_fields.html.erb
@@ -1,7 +1,6 @@
 <%= render "publish/shared/error_wrapper", error_keys: [:modern_languages_subjects], data_qa: "course__modern_languages_subjects" do %>
   <fieldset class="govuk-fieldset">
     <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
-    <%= render CaptionText.new(text: t("course.add_course")) %>
       <h1 class="govuk-fieldset__heading" data-qa="page-heading">
         <% if course.course_code %>
           <span class="govuk-caption-l"><%= course.name_and_code %></span>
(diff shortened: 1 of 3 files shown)


tomas-stefano (approved):

tomas-stefano:
>  [Tech debt: Review and update seeds](https://trello.com/c/yW2cNnSV/966-tech-debt-review-and-update-seeds)

Wrong PR? I didn't understand 😅

PR #5392, opened 2025-07-03
Handle email change for candidate accounts

inulty-dfe:
## Context

A Candidate user can change their email address in One Login.
If they do that, we want to be able to authenticate them correctly and update their email address in our system.

Upon authentication, find the Authentication matching the subject_key of the One Login authentication hash.
Find the associated candidate. If the email does not match the auth hash, then update the email address.

## Trello 

https://trello.com/c/52YVeMdS/825-one-login-account-can-change-email-and-keep-same-candidate-account

## Checklist

### DIFF
--- a/app/services/find/candidate_authenticator.rb
+++ b/app/services/find/candidate_authenticator.rb
@@ -9,25 +9,46 @@ module Find
     # 1. Find existing Candidate via provider authentication
     # 2. Update Candidate email if it has changed
     # 3. Create a new candidate if one does not exist yet
+    # @return Candidate
     def call
-      email_address = oauth.info.email
-      authentication = ::Authentication.find_by(subject_key: oauth.uid)
-
       if authentication.present?
-        candidate = authentication.authenticable
+        sign_in!
+      else
+        sign_up!
+      end
+    end
+
+  private
+
+    # @return Candidate
+    def sign_up!
+      candidate = nil
+      Candidate.transaction do
+        candidate = Candidate.create(email_address:)
+
+        provider = ::Authentication.provider_map(oauth.provider)
+        candidate.authentications.build(provider:, subject_key: oauth.uid)
+        candidate.save!
+      end
+      candidate
+    end
+
+    # @return Candidate
+    def sign_in!
+      authentication.authenticable.tap do |candidate|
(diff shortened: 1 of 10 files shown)


app/services/find/candidate_authenticator.rb
```diff
@@ -0,0 +1,33 @@
...
+    def initialize(oauth:)
+      @oauth = oauth
+    end
+
+    # 1. Find existing Candidate via provider authentication
+    # 2. Update Candidate email if it has changed
+    # 3. Create a new candidate if one does not exist yet
+    def call
+      email_address = oauth.info.email
+      authentication = ::Authentication.find_by(subject_key: oauth.uid)
```
langesi, line 14: Same here. I suggest memoize to reduce method size 

```
Def authentication
 @authentication ||= ::Authentication.find_by(subject_key: oauth.uid)
 end
```

app/services/find/candidate_authenticator.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      if authentication.present?
+        candidate = authentication.authenticable
```
langesi, line 17: create new method named authenticate!

```
# Use existing account if successful and return it
# @return [Candidate] the authenticated Candidate object
def sign_in!
 authentication.authenticable.tap do |candidate|
  unless candidate.email_address.casecmp?(email_address)
   candidate.update(email_address:)
  end
 end
end
```

app/services/find/candidate_authenticator.rb
```diff
@@ -0,0 +1,33 @@
...
+        unless candidate.email_address.casecmp?(email_address)
+          candidate.update(email_address:)
+        end
+      else
+        Candidate.transaction do
```
langesi, line 22: create new method named sign_up!

```
# Create account if successful and return it
# @return [Candidate] the authenticated Candidate object
def sign_up!
 Candidate.transaction do
  Candidate.create(email_address:).tap do |candidate|
   provider = ::Authentication.provider_map(oauth.provider)
   candidate.authentications.build(provider:, subject_key: oauth.uid)
   candidate.save!
  end
 end
end
```

PR #5405, opened 2025-07-13
Add ability to save a course from the search results page

gms-gs:
## Context

As part of the candidate accounts work, we want to allow candidates to save a course for later on the search results page.

Reuse the save toggle partial to allow a user to save a course via the results page. To note I have used Stimulus only for the results page to save a course as there was complications with using a basic rails form due to form nesting with in the results page. 

## Guidance to review

- Sign in
- Save a course
- Ensure the state updates and its in your saved courses list
- If the user is not signed in and they try save a course it should take them to the homepage and show a flash message saying they must be signed in

## Checklist

### DIFF
--- a/app/controllers/find/application_controller.rb
+++ b/app/controllers/find/application_controller.rb
@@ -11,7 +11,7 @@ module Find
 
     default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
-    before_action :set_candidate
+    before_action :candidate
     before_action :redirect_to_cycle_has_ended_if_find_is_down
     before_action :redirect_to_maintenance_page_if_flag_is_active
     before_action :promote_flash_info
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/find/components/_save-course-button.scss
```diff
@@ -5,6 +5,29 @@
...
+    text-decoration-color: govuk-colour("black");
+  }
+}
+
+.results-save-course-button__unstyled-button {
+  float: right;
+
+  @include govuk-media-query($until: tablet) {
+    float: left;
+  }
```
inulty-dfe, line 30: `float` is a pretty un-used property in CSS these days, it's designed to allow content to reflow around the floated element, like an image in text where the text reflows around the image.

I think we would want to keep using flex for positioning when we don't have a specific use case for `float`

langesi (approved):
LGTM

PR #5414, opened 2025-07-15
Find developer is enabled when one-login is disabled except production

inulty-dfe:
## Context

  One Login should be enabled in any environment that Settings.one_login.enabled is true.
  FindDeveloper should replace One Login if Settings.one_login.enabled is false.
  Production should *never* have FindDeveloper enabled

#### What about QA and other environments?

  FindDeveloper logs the user in with a specific email, resolving to a specific user. Is this ok in QA, Staging, Sandbox?

If for any reason we disable One Login in a lower environment, we want the Find Developer strategy to be enabled.

## Guidance to review

Is there any environment other than Production that we should not enable the FindDeveloper Strategy?

## Checklist

### DIFF
--- a/app/lib/authentications/candidate_omni_auth.rb
+++ b/app/lib/authentications/candidate_omni_auth.rb
@@ -17,7 +17,7 @@ module Authentications
           redirect_uri:,
           private_key:,
         }
-      elsif Rails.env.local?
+      elsif !Rails.env.production?
         {
           name: "find-developer",
           fields: %i[uid email],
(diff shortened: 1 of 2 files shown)


inulty-dfe:
[Find Developer OmniAuth Strategy in all envs where One Login disabled except Production](https://trello.com/c/VopyEhKE/884-find-developer-omniauth-strategy-in-all-envs-where-one-login-disabled-except-production)

PR #5417, opened 2025-07-17
create imports table for tracking register school imports

tomas-stefano:
## Context

This PR introduces a new `imports` table, providing some groundwork for logging and reviewing imports/scripts operations (like  register school import). 

This is a key step for achieving long-term data traceability, improved diagnostics, and have an internal tech reporting.

## Usefulness & Benefits

- **Auditing**:  
  Each import, whether it succeeds or fails, is tracked persistently. 
This means past jobs can always be investigated, providing a clear audit trail for program changes and inquiries.

- **Error Diagnostics**:  
  By storing both summary and detailed data, admins and developers can quickly answer:
  - *What happened in this import?*
  - *Which rows failed, and why?*
  - *Are there patterns to failures across time?*


db/migrate/20250717005722_create_imports.rb
```diff
@@ -0,0 +1,12 @@
+class CreateImports < ActiveRecord::Migration[8.0]
+  def change
+    create_table :import do |t|
+      t.jsonb :short_summary
+      t.jsonb :full_summary
+      t.integer :import_type
```
langesi, line 6: Maybe just call it type. Then you could create new models based on it and add specific json accessors etc for the type of import

db/migrate/20250717005722_create_imports.rb
```diff
@@ -0,0 +1,12 @@
...
+
+      t.timestamps
+    end
+    add_index :import, :import_type
```
langesi, line 10: How about a composite index for id & type?
`add_index :import, [:id, :type]
`

db/migrate/20250717005722_create_imports.rb
tomas-stefano: I chosen import_type because type is rails keyword from STI so I am not sure about this, happy to discuss tomorrow in a huddle

db/migrate/20250717005722_create_imports.rb
tomas-stefano: That's a good idea but will we have many records? Probably yes if we decide to move many of our scripts/automation/imports/etc to this table.

PR #5418, opened 2025-07-17
Use imports table for tracking register school imports 

tomas-stefano:
## Context

WIP

WIP

## Guidance to review

WIP


app/models/import.rb
```diff
@@ -0,0 +1,9 @@
+class Import < ApplicationRecord
+  enum :import_type, {
```
langesi, line 2: How about making this a `type` column. You can spin off other models from it and create specific json_accessors and methods for each import type in the future

db/migrate/20250717005722_create_imports.rb
```diff
@@ -0,0 +1,12 @@
+class CreateImports < ActiveRecord::Migration[8.0]
+  def change
+    create_table :import do |t|
+      t.jsonb :short_summary
+      t.jsonb :full_summary
+      t.integer :import_type
+
+      t.timestamps
+    end
+    add_index :import, :import_type
```
langesi, line 10: Create composite index to load both type and id quicker together

db/migrate/20250717005722_create_imports.rb
```diff
@@ -0,0 +1,12 @@
+class CreateImports < ActiveRecord::Migration[8.0]
+  def change
+    create_table :import do |t|
+      t.jsonb :short_summary
+      t.jsonb :full_summary
+      t.integer :import_type
```
langesi, line 6: How bout using type column, you can spin other models off it and create specific json accessors and methods to the type of import

PR #5425, opened 2025-07-18
adding new JSON fields to course enrichment for long form course content

langesi:
## Context

Teacher-candidates need a richer, more structured “long-form” view of each course to help them decide which one to pick. Currently all that data lives as a single JSON blob on the CourseEnrichment model (course_enrichments table), and a handful of one-off keys already exist. To ship the new questions in our short-time window, we’ll continue to use that blob, extending it with new keys, accessors, validations and later UI components. 

1. New version column on course_enrichments
1.1 Type: integer, defaults to 1, not nullable
1.2 Tracks updates to the JSON schema so we can migrate old blobs safely in future
2. JSON Accessors for each new question field
2.1 Added in app/models/course_enrichment.rb
3. Validations
3.1 Presence and length limits on each new field in the model

## Guidance to review

Any suggestions or concerns are welcome!

## Checklist

- [ ] Plan how the JSON should be adapted to allow for long-form course content changes
- [ ] Implement new JSON accessors
- [ ] Change validations depending on Enrichment version
- [ ] Testing!


app/models/recruitment_cycle.rb
```diff
@@ -7,7 +7,7 @@ class RecruitmentCycle < ApplicationRecord
   # a site to a new recruitment_cycle
   has_many :courses, through: :providers
   has_many :sites, through: :providers
-
+  has_many :enrichments, through: :courses
```
tomas-stefano, line 10: Why do we need this association? :)

langesi: Thanks! I'll delete that. I did some queries while I was working on this PR where this helped. It slipped into the commit

It's not related to the PR.

db/migrate/20250716155729_add_version_for_course_enrichment.rb
```diff
@@ -0,0 +1,22 @@
...
+class AddVersionForCourseEnrichment < ActiveRecord::Migration[8.0]
+  # This migration adds a version column to the course_enrichments table.
+  # It sets the default version to 1 for existing records and updates
+  # records created after October 1, 2025, to have a version of 2.
+  def up
+    add_column :course_enrichment, :version, :integer, default: 1, null: false
+
+    # Ensure existing records have the initial version set
+    CourseEnrichment.update_all(version: 1)
+    CourseEnrichment.where("created_at > ?", Time.new(2025, 10, 1)).update_all(version: 2)
```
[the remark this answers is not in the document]
tomas-stefano: Ok, I asked because the update all will create a transaction and the DB will be update the hundred thousands of enrichments we have and during the commit there could be many reads in between. 
But the default is 1 so the update_all is not necessary?

Let's watch the deployment together when it happens  👀

PR #5431, opened 2025-07-21
Add data hub process summary table

tomas-stefano:
## Context

I am adding a new table.

Table: data_hub_process_summary
Purpose: This table stores a summary of each data process run (e.g. imports, syncs).

It's used to track, audit, and report on the outcome of key data operations like Register School Imports or GIAS sync tasks and any other big data process we do to improve our traceability and investigations of the incoming data.

### 📊 `data_hub_process_summary` schema overview

| Column         | Purpose                                                                 |
|----------------|-------------------------------------------------------------------------|
| `type`         | Enables STI; identifies what kind of process (e.g. `RegisterSchoolImportSummary`) |
| `status`       | Marks job outcome (e.g. `"completed"`, `"failed"`); more states will be added as needed |
| `started_at`   | When the process began                                                  |
| `finished_at`  | When the process completed                                              |
| `short_summary`| Key metrics (e.g. totals, counts, highlights)                          |
| `full_summary` | Full detail (e.g. per-row or per-provider breakdown)                   |
| `created_at`, `updated_at` | Standard Rails timestamps for auditing                      |

Usefulness & Benefits
Auditing:
Each script/import, whether it succeeds or fails, is tracked persistently.
This means past jobs can always be investigated, providing a clear audit trail for program changes and inquiries.

Error Diagnostics:
By storing both summary and detailed data, admins and developers can quickly answer:

What happened in this import?
Which rows failed, and why?
Are there patterns to failures across time?

## Guidance to review

1. Did I forget anything?


config/analytics.yml
```diff
@@ -84,6 +84,15 @@ shared:
...
     - vac_status
+  data_hub_process_summary:
+    - id
+    - type
+    - status
+    - started_at
+    - finished_at
+    - short_summary
+    - created_at
+    - updated_at
```
inulty-dfe, line 95: Is it useful to send this to analytics?

tomas-stefano: I think so, not the full summary though.

inulty-dfe: How? What will analytics do with this information?

langesi (approved):
LGTM

config/analytics.yml
tomas-stefano: I don't know. The main reason I send is to have at least a historical data "We run X at the time Y and it was successfull/failed/etc". 

That's why I am not sending the full_summary because that's only for us.

PR #5447, opened 2025-07-23
[Candidate] Track event saved course

gms-gs:
## Context

Whenever a candidate clicks “Save this course” on Find, we need to emit a structured analytics event into BigQuery so we can report on saved‐course behaviour. This event should include the candidate_id and the course_id, along with standard metadata (timestamp, user agent, referrer).

Create a new event type and and `after_action :send_saved_course_analytics_event, only: [:create]` so when saving the course from either results, course page we record the event

## Guidance to review

Run the app locally and save a course and view it in the console, can add print statements to see it easier

## Checklist

### DIFF
--- a/app/controllers/find/candidates/saved_courses_controller.rb
+++ b/app/controllers/find/candidates/saved_courses_controller.rb
@@ -13,10 +13,10 @@ module Find
 
         respond_to do |format|
           if saved_course
-            format.html { redirect_to_course(@course) }
+            format.html { redirect_to_course(course) }
             format.json { render json: { saved_course: saved_course.id }, status: :created }
           else
-            format.html { redirect_to_course(@course, error: t(".save_failed")) }
+            format.html { redirect_to_course(course, error: t(".save_failed")) }
             format.json { render json: { error: t(".save_failed") }, status: :unprocessable_entity }
           end
         end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -55,7 +55,7 @@ authentication:
   # mode: persona     # none critical systems, ie localhost
 
 one_login:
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/find/components/_save-course-button.scss
```diff
@@ -5,6 +5,27 @@
...
+    color: govuk-colour("black");
+  }
+}
+
+.results-save-course-button__unstyled-button {
+  float: right;
+
+  @include govuk-media-query($until: tablet) {
+    float: left;
+  }
```
tomas-stefano, line 28: I am not sure I fully understand the CSS changes to this PR that send events to Bigquery. Was there a reason?

app/controllers/find/candidates/saved_courses_controller.rb
```diff
@@ -2,21 +2,22 @@ module Find
...
+      after_action :send_saved_course_analytics_event, only: [:create]
 
       def index
         @saved_courses = @candidate.saved_courses
       end
 
       def create
-        course = Course.find(params[:course_id])
-        saved_course = SaveCourseService.call(candidate: @candidate, course:)
+        @course = Course.find(params[:course_id])
```
langesi, line 12: Suggestion, move it to a private and memoize

so

```
private

def course
 @course ||= Course.find(params[:course_id])
end
```

langesi: then just use the `course` method everywhere in the controller where you use the instance variable `@course` at the moment

langesi (commented):
Just 1 change then LGTM, let me know what you think

PR #5454, opened 2025-07-28
[Candidate] Track event Candidate applies

gms-gs:
## Context

When a signed‐in candidate clicks “Apply” on a course, emit a candidate_applies event into BigQuery. This event should include the [candidate.id](http://candidate.id/) and a flag indicating whether the course was previously saved, along with standard metadata (timestamp, user agent, referrer).

Add a new event to track a user has applied to a course 

## Guidance to review

- Enable candidate accounts
- Apply for a course
- You can add print statements to make it easier to see the events triggering 

## Checklist

### DIFF
--- a/app/controllers/find/track_controller.rb
+++ b/app/controllers/find/track_controller.rb
@@ -2,13 +2,21 @@
 
 module Find
   class TrackController < ApplicationController
+    ALLOWED_REDIRECT_HOSTS = %w[
+      gov.uk
+      getintoteaching.education.gov.uk
+      education-ni.gov.uk
+      teachinscotland.scot
+      educators.wales
+    ].freeze
+
     def track_click
       utm_content = params[:utm_content]
       url = params[:url]
 
       track_click_event(utm_content, url)
 
-      redirect_to url, allow_other_host: true
+      redirect_to safe_redirect_url(url), allow_other_host: true
     end
 
     def track_apply_to_course_click
@@ -16,10 +24,9 @@ module Find
       url = params[:url]
 
       track_click_event(utm_content, url)
-
       send_candidate_applies_analytics_event(params[:course_id])
 
-      redirect_to url, allow_other_host: true
+      redirect_to safe_redirect_url(url), allow_other_host: true
     end
 
   private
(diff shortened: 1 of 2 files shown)


app/controllers/find/track_controller.rb
```diff
@@ -6,9 +6,34 @@ def track_click
...
       redirect_to url, allow_other_host: true
     end
+
+    def track_apply_to_course_click
+      utm_content = params[:utm_content]
+      url = params[:url]
+
+      track_click_event(utm_content, url)
+
+      send_candidate_applies_analytics_event(params[:course], params[:candidate])
```
tomas-stefano, line 20: I think is best to use current_candidate if logged in, no? 

Otherwise this parameter can be override by any user.

app/controllers/find/track_controller.rb
```diff
@@ -6,9 +6,34 @@ def track_click
...
+
+    def track_click_event(utm_content, url)
+      Analytics::ClickEvent.new(utm_content:, url:, request: request).send_event
+    end
+
+    def send_candidate_applies_analytics_event(course)
+      Analytics::CandidateAppliesEvent.new(
+        request: request,
+        candidate_id: @candidate&.id,
+        course_id: course,
```
tomas-stefano, line 35: Can we rename the variable to course_id?

app/controllers/find/track_controller.rb
```diff
@@ -6,9 +6,34 @@ def track_click
...
+
+    def track_apply_to_course_click
+      utm_content = params[:utm_content]
+      url = params[:url]
+
+      track_click_event(utm_content, url)
+
+      send_candidate_applies_analytics_event(params[:course])
+
+      redirect_to url, allow_other_host: true
```
tomas-stefano, line 22: This line is vulnerable to **open redirect** attacks.

 We pass the `url` parameter from user input directly to `redirect_to` with `allow_other_host: true`, 
 enabling users to redirect anyone to arbitrary external sites (including attacker-controlled domains).

**Explanation of the problem:**  

The main issue is the *open redirect* vulnerability. By accepting a `url` parameter from user input and passing it straight to `redirect_to` with `allow_other_host: true`, the application allows attackers to craft special links that first hit your application, then immediately send users to any external site of their choosing. 

This makes phishing and other attacks much easier, since victims believe they're interacting with a trusted domain.

Risks:
- Attackers can create phishing URLs using our domain, tricking users into trusting fake sites.
- Can lead to theft of credentials, malware, or further attacks.
- **Do not use untrusted user input in redirects**. Only redirect to a known-safe set of destinations.

To prevent this, **never use untrusted or user-provided URLs for redirects**. If you must redirect, allow only a fixed subset of safe, pre-approved locations, or validate URLs rigorously before using them

Some references to help ❤️ 
https://www.stackhawk.com/blog/rails-open-redirect-guide-examples-and-prevention/
 https://dev.to/gregmolnar/open-redirect-vulnerabilities-in-rails-apps-g76
https://security.snyk.io/vuln/SNYK-RUBY-RAILS-1567785
https://www.acunetix.com/vulnerabilities/web/ruby-on-rails-url-redirection-to-untrusted-site-open-redirect-vulnerability-cve-2023-22797/

tomas-stefano (approved):

PR #5458, opened 2025-07-29
Optimise register import removing N+1 queries by bulk inmemory urn lookup

tomas-stefano:
## What

Refactors `SchoolCreator` to bulk-load GIAS schools and provider sites in a single query each, replacing repeated queries per URN with efficient hash lookups.

## Why

Less time is better and less queries to the database too.

## How to verify

- Run the batch import on many URNs and check DB logs: only two queries (plus insertions), never N+1.

## Benefits

The register import went down from around 30 minutes to 14 minutes.


langesi (approved):
LGTM

PR #5462, opened 2025-07-30
Longform: Update ‘Fees and financial support’ for Find and Publish

gms-gs:
## Context

We researched and tested improvements to the course pages on Find. This included restructuring sections and adding new sections, which requires changes to Publish.

This ticket is to update the ‘Fees and financial support’ page in Publish.

Update Publish and Find to reflect the new long from course content for fees and financial support page

## Guidance to review

- Ensure the feature flag is set
- Go through Publish and update the Fees and financial support content.
- Once update publish the course and view the changes in Find

## Checklist

### DIFF
--- a/spec/components/find/courses/fees_component/view_spec.rb
+++ b/spec/components/find/courses/fees_component/view_spec.rb
@@ -26,7 +26,7 @@ describe Find::Courses::FeesComponent::View, type: :component do
 
       result = render_inline(described_class.new(course))
       expect(result.text).not_to include("International students")
-      expect(result.text).to include("The course fees for UK citizens")
+      expect(result.text).to include("The course fee for UK citizens")
     end
   end
 end
--- a/spec/features/find/course/viewing_a_course_spec.rb
+++ b/spec/features/find/course/viewing_a_course_spec.rb
@@ -343,7 +343,7 @@ private
 
   def then_i_should_only_see_the_uk_fees
     expect(find_course_show_page).to have_content(
-      "The course fees for UK citizens in #{RecruitmentCycle.current.year} to #{RecruitmentCycle.current.year.to_i + 1} are £9,250",
+      "The course fee for UK citizens in #{RecruitmentCycle.current.year} to #{RecruitmentCycle.current.year.to_i + 1} are £9,250",
     )
 
     expect(find_course_show_page).not_to have_international_fees
--- a/spec/features/publish/viewing_a_course_preview_spec.rb
+++ b/spec/features/publish/viewing_a_course_preview_spec.rb
@@ -370,7 +370,7 @@ private
   end
 
   def and_i_see_the_the_course_fee
-    expect(page).to have_text "The course fees for UK citizens in #{course.recruitment_cycle.year} to #{course.recruitment_cycle.year.to_i + 1} are £100."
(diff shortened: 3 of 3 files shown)


app/controllers/publish/courses/course_content/fees_and_financial_support_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+        def edit
+          @course_content_course_fees_and_financial_support_form = CourseContentCourseFeesAndFinancialSupportForm.new(course_enrichment)
+          @copied_fields = copy_content_check(::Courses::Copy::FEES_FINANCIAL_SUPPORT_FIELDS)
+
+          @copied_fields_values = copied_fields_values if @copied_fields.present?
+          @course_content_course_fees_and_financial_support_form.valid? if show_errors_on_publish?
+        end
+
+        def update
+          @course_content_course_fees_and_financial_support_form = CourseContentCourseFeesAndFinancialSupportForm.new(
```
inulty-dfe, line 19: I'd suggest we namespace the forms as the controllers are namespaced

inulty-dfe (commented):
Can we have a commit that "installs" the feature flag "Add longform_content FeatureFlag". And if possible break the rest of this commit into smaller commits

langesi (changes requested):
Hi George,

Great work on this! I’ve left one comment inline on the PR. I think Iain touched on the most important parts already. A few higher-level thoughts around structure and versioning as earlier discussed:

1. New V2 surface should live separately:
To keep the old and new flows clearly delineated, I think it makes sense to introduce new partials/components for the V2 changes under their own namespace/path. That avoids clutter and eventually will make it easier to retire V1 components/partials. 

Example paths:

- Partial: `find/courses/v2/_fees_and_financials.html.erb`
- Component (if we do use components, though I lean towards partials for this): `Find::Courses::V2::FeeComponent`
- Form: `Publish::Course::FeesAndFinancialSupportForm`

As you see the Form does not have the V2 Namespace, that's because the namespace above doesn't exist yet. Adding the version would have only served naming continuity and not abstraction.

2. Partial vs component

Just a comment from me that no one asked for. I prefer partialsbecause they introduce less indirection, reduce the risk of parameter explosion, and keep simplicity. It's up to you if you'd like to use components and choose speed by just copying V1 components. But thought I just share my thoughts.

If anyone has strong objections to this approach, happy to discuss, otherwise I think it’ll keep the divergence manageable while making intent clear.

app/views/publish/courses/_description_content.html.erb
```diff
@@ -33,6 +33,49 @@
     action_visually_hidden_text: t("publish.providers.courses.description_content.interview_process_hidden_text"),
   ) %>
 
+  <% if FeatureFlag.active?(:long_form_content) || @recruitment_cycle.after_2026? %>
```
langesi, line 36: Can this be in a separate this partial? Splitting V1 and V2 by having different partials?

langesi (commented):
Would suggest we create seperate partials for V2. This will make it easier for us later to remove V1 and reduce risk that the wrong line will get removed

langesi (approved):
LGTM, well done!

PR #5466, opened 2025-08-01
feat(2424): PTT non-root user for AKS containers

pasitconsulting:
### Context

Requirement: to follows best practice for running containers securely on AKS hosts by running containers using a non-root user

https://trello.com/c/4o30khep/2424-security-aks-containers-run-as-non-root-user-ptt

- add new non-root user and group
- chmod minimum required files to allow non-rot- user to run the app
- amend Dockerfile to run the app as the new non-root user

### Guidance to review
Screenshot from 2025-08-01 11-36-51

https://github.com/DFE-Digital/publish-teacher-training/pull/5466/files

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,7 +55,7 @@ COPY --from=middleman /public/ $APP_HOME/public/docs/
 RUN ls /app/public/ && \
   yarn build && \
   bundle exec rake assets:precompile && \
-  rm -rf node_modules tmp
+  rm -rf node_modules tmp/*
 
 ARG COMMIT_SHA
 ENV COMMIT_SHA=${COMMIT_SHA}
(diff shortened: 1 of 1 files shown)


langesi (approved):
✅ Looks good to me and happy to approve.
Tested the app, logged into the pods, checked the CICD, Checked sidekiq page, should all be fine, I think!

Dockerfile
```diff
@@ -60,4 +60,14 @@ RUN ls /app/public/ && \
 ARG COMMIT_SHA
 ENV COMMIT_SHA=${COMMIT_SHA}
 
+# Create non-root user and group with specific UIDs/GIDs
+RUN addgroup -S appgroup -g 20001 && adduser -S appuser -G appgroup -u 10001
+
+# Change ownership only for directories that need write access
+RUN mkdir /app/tmp
```
RMcVelia, line 67: we seem to specifically delete the tmp directory on L58.
If we want to keep it maybe change L58 to 
rm -rf node_modules tmp/*
and remove L67

PR #5467, opened 2025-08-01
Increase main_app memory limit to 1.5Gi in QA configuration

tomas-stefano:
## Context

* Increase QA main_app memory limit to 1.5Gi.
* This change is in response to observed memory usage during recent register school import rake task, which peaked at approximately 557Mi locally. 
* Setting the memory limit to 1.5Gi provides a safe buffer for fluctuations and ensures reliable operation during heavy workloads.

## Why?

Helps prevent OOMKilled pod failures during imports.


RMcVelia (approved):

langesi (approved):
All good, as discussed

PR #5472, opened 2025-08-04
FlashBanner uses govuk_notification_banner helper

inulty-dfe:
## Context

We have a NotificationBanner component that seems to duplicate `govuk_notification_banner`. We can move away from this custom component by changing the places it's used.

Update `FlashBanner` to not depend on `NotificationBanner`

## Guidance to review

Bonus deleting old `data-ga` attributes from when we used Google Tag Manager

## Checklist


langesi (commented):
Great work, just some small comments

PR #5489, opened 2025-08-11
Fix rollover period

tomas-stefano:
## Context

The rollover period is any time before the cycle starts up to 1 month after the cycle started.

I confused October 31st with actual 30 September. Probably I was tired.


app/models/recruitment_cycle.rb
```diff
@@ -103,7 +103,7 @@ def year_range
   end
 
   def rollover_period_2026?
-    year.to_i == 2026 && Time.zone.now < application_start_date
+    year.to_i == 2026 && Time.zone.now < (application_start_date + 1.month)
```
inulty-dfe, line 106: What if the rollover period changes in future. I think it would be best to have this centralised somewhere.

```ruby
def rollover_end
  application_start_data + 1.month
end
```

inulty-dfe (changes requested):
I think we should begin to codify this concept of when rollover ends, at least centralise it

PR #5496, opened 2025-08-12
Add message_attribute option to word count validator

inulty-dfe:
## Context

We have an ask to make the WordCountValidator return error messages that don't conform to the existing implementation.

Currently, if the attribute has too many words, the error message is `"Reduce the word count for #{column_name}"`

We want to have the option of using a custom value in place of the `column_name`, such as the label of the input on the form.

Add `message_attribute` seems a useful API to customize the WordCountValidator while keeping the standard error message format and to reduce the repetition of "Reduce the word count for "

## Checklist


inulty-dfe:
Closing as we should be going for the design system

PR #5498, opened 2025-08-12
Add DB field for validating schools

tomas-stefano:
## Context

Add new field to the course table.

This field will be used as a flag for a mandatory action to schools (on rolled over period only) in order to publish, so that we can be confident providers have checked the right schools are the attached schools on courses after the changes.

## Guidance to review

1. Are you ok with the naming?


langesi (approved):
LGTM, don't forget to report the new column to the data team

PR #5507, opened 2025-08-14
Sync cycle times with apply

CatalinVoineag:
## Context

Apply and published have agreed to change cycle dates a bit.

The cycle dates for the next, next cycle (which will become the next cycle in a few months) have been confirmed with policy, apply and published, and updated in Apply.

## Checklist


langesi (approved):
Thanks Catalin! LGTM!

PR #5527, opened 2025-08-18
[Longform] Adding provider relationship to course_enrichment

langesi:
## Context

During the bug party publishing was blocked due to failing validations. After inspection I found that the validation should have been on the provider table.

1. Remove unused keys in json blob
2. Remove validation on those unused keys
3. Introduce new relationship to allow for delegation methods
4. Introduce delegation of fields that need to be validated during publishing
5. Introduce write delegation to allow for course enrichment tests to also work on provider table fields
6. Changing the eager load in the course_controller only for publishing of course to only include the latest draft enrichment, this allows for the validations to only run on the draft enrichment (At least when you run it the first time, running `valid?(:publish)` a second time will result in `false`
7. Specs

## Checklist

- [x] Remove unused keys in json blob
- [x] Remove validation on those unused keys
- [x] Introduce new relationship to allow for delegation methods
- [x] Introduce delegation of fields that need to be validated during publishing
- [x] Introduce write delegation to allow for course enrichment tests to also work on provider table fields
- [x] Changing eager load on publish course controller to allow for the publishing of courses
- [x] Specs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -187,7 +187,7 @@ group :development, :test do
   gem "rspec-wait"
 
   # Test framework
-  gem "rspec-rails", "8.0.1"
+  gem "rspec-rails", "8.0.2"
 
   gem "selenium-webdriver"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -633,7 +633,7 @@ GEM
     rspec-mocks (3.13.5)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.13.0)
-    rspec-rails (8.0.1)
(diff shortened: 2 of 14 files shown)


inulty-dfe:
We could do with some tests around the validations, that they appear in the error notice

langesi:
This PR is not really for the error messages, there is a seperate ticket that I will have a look at for this. The error messages that had to be fixed for this trello ticket were for pages and JSON keys that shouldn't exists. So I removed them. The error messages I added are not part of the ticket.

I'll write a spec in an upcoming PR @inulty-dfe

PR #5541, opened 2025-08-20
Where you will train displayed correctly in live and preview

inulty-dfe:
## Context

  "Where you will train" is displayed in course preview and live in Find

  - If it is rendered on Find, show only the published values for the fields
  - If in preview, show the latest enrichment values regardless if they are published.
  - If none of the values are present, show the missing information component

We only need to check the presence of the required fields, we omit the optional ones.

## Checklist


inulty-dfe:
[Bug: In the preview, it should enter the content if available, but it still has the link to enter details](https://trello.com/c/bzYlueIL/986-bug-in-the-preview-it-should-enter-the-content-if-available-but-it-still-has-the-link-to-enter-details)

PR #5546, opened 2025-08-21
Enable Longform Content flag in spec matrix

inulty-dfe:
## Checklist


tomas-stefano:
We don't have time travel in specs right?

Also if we change current_recruitment_cycle to 2026 will change anything?

inulty-dfe:
> We don't have time travel in specs right?
> 
> Also if we change current_recruitment_cycle to 2026 (only on this PR) will change anything?

Good point. Lets try that

PR #5552, opened 2025-08-21
Increase QA worker memory

tomas-stefano:
## Context

QA and review apps are being restarted during rollover.

And we are losing jobs. Increasing to test rollover on QA and see if is the root cause of it


langesi (approved):
Approved!

PR #5553, opened 2025-08-22
[Longform] Adding Salary details to main page in publish

langesi:
## Context

This is a bug that is also in V1 prior to the longform changes. The row for salaried courses was missing. If this does not get fixed before the new cycle rolled over courses could not change their salaried section

Adding salaried row to publish course as a row

## Guidance to review

Tested it already, just need a code review


tomas-stefano (approved):
Can you add a note to add a spec for it? System spec.

PR #5555, opened 2025-08-22
Increase word limit for disability support fields to 250 across forms…

langesi:
… and specs

## Checklist


tomas-stefano (approved):
Can we add a note to add a spec for it? Unit spec.

PR #5558, opened 2025-08-26
Fix wrong content on course preview page

gms-gs:
## Context

Wrong content on course preview page, It should say ‘30th September’

Having the value hardcoded is not ideal, so longer term we should agree on a way to automate this

Update the `Settings.next_cycle_open_date.to_date` value to be 30th September 

## Guidance to review

Create a scheduled course and ensure the date is 30th September  
Screenshot 2025-08-26 at 13 33 29

## Checklist

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -71,7 +71,7 @@ one_login:
   private_key: <%= ENV.fetch('ONE_LOGIN_PRIVATE_KEY', 'private_key').gsub("\n", '\n') %>
 
 current_recruitment_cycle_year: 2025
-next_cycle_open_date: 2025-10-1 # TBC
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -71,7 +71,7 @@ one_login:
   private_key: <%= ENV.fetch('ONE_LOGIN_PRIVATE_KEY', 'private_key').gsub("\n", '\n') %>
 
 current_recruitment_cycle_year: 2025
-next_cycle_open_date: 2025-10-1 # TBC
+next_cycle_open_date: 2025-09-30
```
tomas-stefano, line 74: Note: Hopefully, this on end of cycle outcome should be deleted

PR #5565, opened 2025-08-27
Test suite passes when Cycle changes from 2025 to 2026

inulty-dfe:
## Context

The publish service code base is built around the concept of a Recruitment Cycle.

Recruitment Cycles have transitions and intervals. The application behaviour can a lot after a transition from one interval to anoter. For example, when apply closes, the apply button on courses in find are all automatcally disabled / replaced with a notice.

If we write a test in the middle of the cycle asserting that we expect to see the apply button, it will pass, but when we are running the test after apply closes, that test will fail becuase the test suite by default recognises the Cycle timetable and knows not to render the apply button.

This is why we use Timecop to stub the current time in the application. We want to be able to pretend we are inthe rollover period in the test suite so we can test how the appplication behaves when it really is in the rollover period.

Sometimes when we write tests, we forget that the thing we're writnig tests for  will change behviour in a different interval of the cycle. Of ten this happens towards the end of the cycle when find closes and rollover begins. Becuase we don't test every feature under every time period, we can find we have a lot of failing tests when the time period changes significantly.

Another reason we see a lot of failing tests at the end/beginning of a cycle is that we test to make wide raching changes to the system to be launched in the new cycle. Tests that were written for the standard, existing behaviour will pass in the previous periods but now in the new cycle for the same period they will fail. 

These are the main issues that this PR is dealing with.

- Expand the API of the `Find::CycleTimetable` period methods to accept a year representing the recruitment cycle.
  - change ` def self.apply_opens` to `def self.apply_opens(year)` so we can get the time for apply opens in any cycle year
- Lock down tests to specific periods, and even specific periods in specific cycles.
- Add `.cycle_year_for_time` so we can calculate what recruitment cycle any given time is in.
- The availability of `.cycle_year_for_time` could be used to replace the Setting :current_recruitment_cycle_year. We can determine what cycle year we are in based on the current time of the application.
- Initialize the recruitment cycle factory dates with the CycleTimetable dates for the "current_year". This allows us to change the "current_year" and the CycleTimetable will automatically initialize the factory with the correct dates for the given year.
- Disable `SiteSetting` completely in the test environment. This means it doesn't interfere with the CycleTimetable and doesn't need to reach out to redis for a value. In test we manipulate the "current period" using tools other than `SiteSetting` so it is not necessary in the test environment. It's only used in non-prod envs like review apps and qa.
- Extend `rails_helper` by creating a `CycleTimetableHelpers` module and including (and extending) it in RSpec. This  makes `mid_cycle(year)`, `find_closes(year)` etc, available without having to call `Find::CycleTimetable.mid_cycle(2025)`. Extending the module makes those methods available outside the example group like in the metadata of an example/example group.

## Checklist

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -196,7 +196,6 @@ jobs:
         # Separate mutliple flags with a space
         feature-flags:
           - ""
-          - "SETTINGS__CURRENT_RECRUITMENT_CYCLE_YEAR=2026"
 
     steps:
       - name: Checkout
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -194,7 +194,9 @@ jobs:
         # Pass Settings compatible env vars to runs tests with
         # feature flag(s) enables
         # Separate mutliple flags with a space
-        feature-flags: [""]
+        feature-flags:
+          - ""
+          - "SETTINGS__CURRENT_RECRUITMENT_CYCLE_YEAR=2026"
 
     steps:
       - name: Checkout
--- a/app/lib/site_setting.rb
+++ b/app/lib/site_setting.rb
@@ -4,6 +4,10 @@ require "./config/initializers/redis"
 
 class SiteSetting
   def self.cycle_schedule
+    # ENV["ENABLE_SWITCHER"] allows us to selecctively enable the switcher in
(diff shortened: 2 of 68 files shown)


langesi (approved):
Looks good to me. I'm very happy with the use of metadata to Time travel. It makes it not just easier but also safer to time travel in tests. Huge win! Well done!

PR #5571, opened 2025-08-28
Improve performance of publishing a course with many schools

tomas-stefano:
## Context

Before this commit if you select all schools from a provider with thousand schools the publishing service would take more than a minute.

The reason is: N+1 query and N+1 update statement for each school!

Change to update all does solve the performance issue.


langesi (approved):
LGTM, well done!

inulty-dfe:
[Bug: "Check the schools for this course"](https://trello.com/c/OHKA7H2f/1017-bug-check-the-schools-for-this-course)

PR #5579, opened 2025-08-29
Add PromoteTrainingProviderAccreditation service

inulty-dfe:
## Context

Training Providers need to be promoted to Accredited providers in our system from time to time.
This process involves
 - Setting accredited = true
 - Deleting accredited_provider_code on all courses
 - Deleting accredited partnerships

Add Service and tasks to promote training providers

## Checklist

### DIFF
--- a/app/services/providers/promote_training_provider_accreditation.rb
+++ b/app/services/providers/promote_training_provider_accreditation.rb
@@ -21,7 +21,10 @@ module Providers
         end
 
         # Reset accredited provider on Courses
-        training_provider.courses.each { it.update!(accredited_provider_code: nil) }
+        training_provider.courses.each do
+          it.accredited_provider_code = nil
+          it.save!(validate: false)
+        end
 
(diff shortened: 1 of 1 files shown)


app/services/providers/promote_training_provider_accreditation.rb
```diff
@@ -0,0 +1,29 @@
...
+        log "#{@training_provider.name_and_code} has been upgraded to an accredited provider!"
+      end
+    end
+
+  private
+
+    def log(text)
+      Rails.logger.tagged("PromoteTrainingProviderAccreditation") { it.info(text) }
+    end
+    attr_reader :training_provider, :number
```
tomas-stefano, line 27: There was some PRs that @langesi asked to move those to the class declaration. Shall we follow the same pattern?

inulty-dfe: I don't see why we'd want to do that

PR #5582, opened 2025-09-01
Show all fees fields in course description

inulty-dfe:
## Context

Some of the Fees and Financial support input content is missing from the Course summary page.

### Display all the input content on the summary page

image

### Bug fix

`def markdown` returns an unsafe string if the argument is nil or blank. We expect that when we concatenate the result of `markdown(course.value) + markdown(course.other_value)` that the result is html_safe, even it one of the courses attributes is blank.

|Before|After|
|---|---|
|image|image|

## Checklist

### DIFF
--- a/app/helpers/publish_helper.rb
+++ b/app/helpers/publish_helper.rb
@@ -2,7 +2,7 @@
 
 module PublishHelper
   def markdown(source)
-    return "" if source.blank?
+    return "".html_safe if source.blank?
 
     render = Govuk::MarkdownRenderer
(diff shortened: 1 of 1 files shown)


inulty-dfe:
[Bug: Fees and financial support is fine on the live preview, and Find preview. But not on the 'Description tab'](https://trello.com/c/pYVnGYnP/1039-bug-fees-and-financial-support-is-fine-on-the-live-preview-and-find-preview-but-not-on-the-description-tab)

PR #5583, opened 2025-09-01
Enable One Login in Sandbox

inulty-dfe:
## Context

As part of the launch of Candidate Accounts in production, we want to test that the One Login integration is working in production.

We can safely do this in Sandbox.

## Checklist


langesi (approved):
All good

PR #5587, opened 2025-09-02
Move publish feature tests unrelated to courses to system directory

inulty-dfe:
## Context

We are moving Feature tests to the System test directory so our tests are consistent with the predominant spec arrangement recommended in Rails.

## Checklist


langesi (approved):
This is all good, just a couple of failing specs

PR #5606, opened 2025-09-09
Refactor provider details for find and publish

gms-gs:
## Context

The content on the provider details pages was not rendering the correct input values when previewing a course. I also noticed the logic was also set up when viewing provider details in find would render both v1 and v2 content.

Create a shared partial which can be used for both publish and find 

## Guidance to review

- View a v1 course in find and should show the v1 content for provider details
- Preview a v1 course in find and should show the v1 content for provider details
- View a v2 course in find and should show the v2 content for provider details
- Preview a v2 course in find and should show the v2 content for provider details

## Find 
Screenshot 2025-09-09 at 14 15 04

## Publish
Screenshot 2025-09-09 at 14 14 53

## Checklist

### DIFF
--- a/app/views/find/courses/providers/show.html.erb
+++ b/app/views/find/courses/providers/show.html.erb
@@ -19,15 +19,5 @@
 <% end %>
 
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <%= render Find::Courses::ContactDetailsComponent::View.new(@course) %>
-
-    <% if @course.provider.about_us.present? || @course.provider.value_proposition.present? %>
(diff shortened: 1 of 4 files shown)


app/controllers/publish/courses/providers_controller.rb
```diff
@@ -13,7 +13,7 @@ def show; end
 
       def build_course
         super
-        authorize @course
+        authorize @course.decorate
```
langesi, line 16: Looking at the `CourseBasicDetailConcern`, I believe the `@course` instance var should already be decorated. cCalling decorate here wouldn't serve a purpose, right?

PR #5609, opened 2025-09-10
Set up add course flow for DT page

gms-gs:
## Context

We have updated the course naming policy based on analysis and we need to add in specialisms for design and technology courses into Publish.

### User Story:
As a provider on the Find & Publish platform,

I want to select five new specialisms under “design and technology” when I create or edit a course,so that my course data is compliant with the updated DfE naming policy.

Update the publish course flow to have a specialisms page. This page is only show if design and technology is chosen on the previous subjects page. When the course is saved the naming convention of the course is also updated across Publish and Find e.g `Design and technology (Food technology and Product Technology)` if the specialisms selected are greater than 2 we just show `Design and technology` as per designs.

## Checklist

### DIFF
--- a/spec/factories/subjects/design_technology_subjects.rb
+++ b/spec/factories/subjects/design_technology_subjects.rb
@@ -2,11 +2,11 @@
 
 FactoryBot.define do
   design_technology_subjects = {
-    "Engineering" => "D420",
-    "Product design" => "D440",
-    "Food technology" => "D470",
-    "Textiles" => "D460",
-    "Graphics" => "D490",
+    "Engineering" => "DT420",
+    "Product design" => "DT440",
+    "Food technology" => "DT470",
+    "Textiles" => "DT460",
+    "Graphics" => "DT490",
   }
 
   factory :design_technology_subject do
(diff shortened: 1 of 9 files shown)


spec/factories/subjects/design_technology_subjects.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  design_technology_subjects = {
+    "Engineering" => "D420",
```
langesi, line 5: didn't the subject codes for these specialisms start with `DT`? Should this be changed in the Factory?

app/services/courses/generate_course_name_service.rb
```diff
@@ -53,6 +55,27 @@ def call
 
     attr_reader :course, :subjects
 
+    def generate_design_technology_title
```
langesi, line 58: This method seems rather complex with a couple of nested conditions, ideally it would be great if this were extracted into methods or even it's own class.

Either way, this is just a suggestion for the future. Due to time pressure I'd say leave this how it is as long as it works 👍

app/services/courses/generate_course_name_service.rb
```diff
@@ -53,6 +55,27 @@ def call
...
+      dt_title = dt_title_with_specialisms
+
+      if course.subordinate_subject_id.present?
```
langesi, line 61: Would suggest making this a guard clause for readability and reduction of nesting. This is not a priority though if you don't get to it

PR #5612, opened 2025-09-12
Time.zone.now and CycleTimetable determines the cycle year

inulty-dfe:
## Context

Until now, the truth in the system about what the current recruitment cycle year was defined in the Settings.

Settings.current_recruitment_cycle_year

```
# config/settings.yml
current_recruitment_cycle_year: 2025
```

This required us to make a PR to update this hard coded value when we wanted to open Find. "find opens" is the traditional moment for the new cycle to begin.

This change migrates the truth of the cycle year to the CycleTimetable. We take the current system time (Time.zone.now) and find the current cycle year based on that value.

## Test suite
In tests, we only need to travel to the time when a different cycle is active for the year to change. No more mocking `Settings.current_recruitment_cycle_year`. We do need to make sure that the `RecruitmentCycle` model exists in the DB when we time travel like this. (`find_or_create(:recruitment_cycle`)

### Next steps

These changes work for mid cycle, but we need to make further changes to make sure that other components are loaded at the right time.

For example, loading future recruitment cycles, we can get all the cycle years where find_opens is in the future, and use these years to fetch from the DB, instead of updating columns on the `RecruitmentCycle` record.

## Checklist

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -15,7 +15,7 @@ Provider.destroy_all
 User.destroy_all
 RecruitmentCycle.destroy_all
 
-current_recruitment_year = Settings.current_recruitment_cycle_year
+current_recruitment_year = Find::CycleTimetable.cycle_year_from_time(Time.zone.now)
 current_recruitment_cycle = RecruitmentCycle.create(year: current_recruitment_year, application_start_date: Date.new(current_recruitment_year.to_i - 1, 10, 9), application_end_date: Date.new(current_recruitment_year.to_i, 9, 30))
 next_recruitment_cycle = RecruitmentCycle.create(year: (current_recruitment_year.to_i + 1).to_s, application_start_date: Date.new(current_recruitment_year.to_i, 10, 8), application_end_date: Date.new(current_recruitment_year.to_i + 1, 9, 30))
 
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -15,7 +15,7 @@ Provider.destroy_all
 User.destroy_all
 RecruitmentCycle.destroy_all
 
-current_recruitment_year = Find::CycleTimetable.cycle_year_from_time(Time.zone.now)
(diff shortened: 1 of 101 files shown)


db/seeds.rb
```diff
@@ -15,7 +15,7 @@
 User.destroy_all
 RecruitmentCycle.destroy_all
 
-current_recruitment_year = Settings.current_recruitment_cycle_year
+current_recruitment_year = Find::CycleTimetable.cycle_year_for_time(Time.zone.now)
```
inulty-dfe, line 18: Redis is not necessarily available when we run this so we do not call `Find::CycleTimetable.current_year` because that calls `SiteSetting`.

spec/smoke/api/v1/providers_spec.rb
```diff
@@ -5,7 +5,7 @@
 describe "V1 Public API Smoke Tests", :aggregate_failures, :smoke do
   subject(:response) { HTTParty.get(url) }
 
-  let(:recruitment_year) { Settings.current_recruitment_cycle_year }
+  let(:recruitment_year) { Find::CycleTimetable.cycle_year_for_time(Time.zone.now) }
```
inulty-dfe, line 8: Redis is not available in smoke tests

spec/system/support/providers/switching_between_recruitment_cycles_spec.rb
```diff
@@ -25,27 +24,22 @@
     i_should_be_on_the_next_cycle_page
   end
 
-  scenario "viewing providers page when not in rollover", travel: mid_cycle do
+  scenario "viewing providers page when not in rollover" do
     given_we_have_a_next_cycle
-    and_there_are_two_recruitment_cycles
+    and_today_is_before_next_cycle_available_for_support_users_date
```
inulty-dfe, line 29: Don't need to `travel:` because we travel inside the spec when it matters

PR #5615, opened 2025-09-15
Remove candidate email_address from analytics

inulty-dfe:
## Context

  We are removing the email address until we can confirm beyond doubt
  that our recording of this information is reasonable unter data
  protection policy. We may need to provide a privacy policy before we
  can record such data.

Remove candidate email address from analytics until we can confirm it is acceptable to record  it with our current data protection guideline adherance.

## Checklist

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -9,9 +9,9 @@ shared:
     - updated_at
   candidate:
     - id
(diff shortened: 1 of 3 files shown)


langesi (approved):
LGTM

PR #5624, opened 2025-09-17
Add postgis adapter gem

tomas-stefano:
## Context

In order to do the optimisations for Find I need to create spatial fields and index them.

So we need to have available the postgis types to active record so the db/schema.rb is generated without problem. 

Otherwise we would have to change the schema format to sql...

This PR adds the postgis as adapter following [postgis readme](https://github.com/rgeo/activerecord-postgis-adapter)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,9 @@ gem "draper"
 # Custom attributes for endpoints
 gem "active_model_serializers"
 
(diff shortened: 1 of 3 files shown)


langesi (approved):
LGTM. Let's see if Prod agrees!

PR #5632, opened 2025-09-18
Dont allow discarded users to sign in on publish

gms-gs:
## Context

We noticed that in Publish when a user "Deletes" a user the User is then discarded. However the discarded user us still able to sign into Publish. 

The change is when a user attempts to sign in, we check if there is a user AND if the current_user attempting to sign in is discarded. If it is discarded we terminate their session and redirect to user_not_found_path.

**Note this is a temporary solution to the issue.** As destroying a user has downstream affects one being related to course enrichments so unable to just delete a user

## Guidance to review

Attempt to sign in with a discarded user on Publish

## Checklist


langesi (approved):
LGTM, thanks a lot! Quick work

PR #5639, opened 2025-09-22
Set the application Timezone to Europe/London

inulty-dfe:
## Context

`Time.zone.now` returns the current time in UTC. If we have code that actions something at 14:00, then during British Summer Time that action will take place an hour later.

## Checklist

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -116,7 +116,7 @@ class Provider < ApplicationRecord
 
   scope :changed_since, lambda { |timestamp|
     if timestamp.present?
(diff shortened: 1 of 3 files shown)


inulty-dfe:
A test failed that we wouldn't have expected to fail after setting the application TimeZone to 'Europe/London'.

The test failed in the API specs where an ISO8601 string was not compared to database records as we would have expected. This is certainly not going to be an isolated issue and indicates we require a lot of testing and investigation before setting the time zone application wide.

Closing until we schedule this work.

PR #5641, opened 2025-09-22
Scope the ProviderCache cache key to the current cycle year

inulty-dfe:
## Context

We cache the names and ids of providers for the provider search feature. The cache key doesn't consider the recruitment cycle year so when 9am on the day Find opens occurs, we will be serving providers from the old cycle for up to an hour.

Pass the current recruitment cycle year as a part of the cache key so when the cycle year changes the cache is rebuilt for the current year.

## Checklist

### DIFF
--- a/spec/services/providers_cache_spec.rb
+++ b/spec/services/providers_cache_spec.rb
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 RSpec.describe ProvidersCache do
-  let(:cycle_year) { 2026 }
+  let(:cycle_year) { Find::CycleTimetable.current_year }
   let(:cache_key) { "providers:list:#{cycle_year}" }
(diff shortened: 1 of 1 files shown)


langesi (approved):
LGTM

app/services/providers_cache.rb
```diff
@@ -8,7 +8,7 @@ def initialize(expires_in: 1.hour)
   end
 
   def providers_list
-    Rails.cache.fetch("providers:list", expires_in: expires_in) do
+    Rails.cache.fetch("providers:list:#{Find::CycleTimetable.current_year}", expires_in: expires_in) do
```
langesi, line 11: You could also use an array here for the cache key, happy with this though!

langesi (approved):

PR #5650, opened 2025-09-24
Optimising the course page

tomas-stefano:
## Context

This PR removes so many unnecessary hits to the DB by adding includes to the course page.

- Add more associations to includes and remove the N+1 queries on course enrichments

**Before**

```
(74 queries)
```

**After**:
```
(17 queries)
```

## Fragment Caching

Added fragment caching to the course page using changed_at (see TouchCourse module) so everytime is changed the cache is expired.

Before cache: 300ms-350ms
After cache: 80ms-90ms

### DIFF
--- a/app/views/find/courses/show.html.erb
+++ b/app/views/find/courses/show.html.erb
@@ -9,61 +9,63 @@
   <% end %>
 <% end %>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">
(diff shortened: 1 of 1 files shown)


langesi:
Hi @tomas-stefano  the last spec in Publish raises some red flags. A validation error on course/course enrichment appears if the wrong enrichment (most likely V1 rolled over enrichment) is eager-loaded. That could disrupt publishing for rolled-over courses (V1 → V2). I’ll need to investigate this before I can approve the PR.

tomas-stefano:
Closing the PR. I think is better we leave the optimisations as is.

PR #5659, opened 2025-09-26
The link in support to a provider in publish scopes the cycle

inulty-dfe:
## Context

The link in the support console that takes the support user to a provider in publish is not scoped to the recruitment cycle.
If a new provider is created in the next before the next cycle begins, the link will default to looking for the provider in the current cycle, but returns a 404.

## Checklist


langesi (approved):
Well done

PR #5668, opened 2025-09-30
Remove Settings.next_cycle_open_date

inulty-dfe:
## Context

We use `Settings.next_cycle_open_date` to communicate to publishers when their courses will be live.
We can use the CycleTimetable to calculate this instead and avoid a deployment to update.

## Guidance to review

image

## Checklist


inulty-dfe:
In fact, the View Course column should not have a date. It should have something else, because a course in a past cycle will never be published again

inulty-dfe:
> In fact, the View Course column should not have a date. It should have something else, because a course in a past cycle will never be published again

This is a support interface issue so impact is small. I'm going to go ahead with this change.

PR #5674, opened 2025-10-01
Restrict provider users to their own cycle

inulty-dfe:
## Context

Providers only ever have access to the current recruitment cycle and the next recruitment cycle when rollover is happening and it's available to provider users.

Our routes allow for anyone to put any valid year in the URL, and there's nothing to prevent providers from doing that.

- Add new providers scope to return only current and future providers, this will be used to get providers for users.
- Add `RecruitmentCycle.current_cycle` which returns a scope instead of executing a query. This makes efficient queries possible

## Checklist


inulty-dfe:
This should be scheduled along with other authorization tasks

PR #5677, opened 2025-10-01
Bug: Update sign in banner on candidate accounts

sarahcrack:
## Context

Trello: https://trello.com/c/WzcNKakt/1116-bug-update-sign-in-banner-on-candidate-accounts

This banner on the course page and the search page needs a link to sign in to make it clearer. It pops up when you are logged out but ‘heart’ a course on Find. It is a bit confusing to candidates and could be better with a link.

- as discussed, we are now directing users straight to Gov OneLogin rather than using a flash banner
- remove flash banner notification which stated that you needed to sign in
- add a link to the Gov.uk One Login sign in page when you try and save a course, but are not signed in (i.e. clicking on the ❤️)

## Guidance to review
- on Find, make sure you are signed out and try to save a course by clicking on the heart
- you should then be directed straight to the GovOneLogin page

https://github.com/user-attachments/assets/bfdeab70-b92f-4fb7-849b-e4f6114e05ac

## Checklist

### DIFF
--- a/app/controllers/concerns/find/authentication.rb
+++ b/app/controllers/concerns/find/authentication.rb
@@ -50,14 +50,16 @@ module Find
 
     def request_authentication
       session["return_to_after_authenticating"] = request.url
+      flash_info_message = t("find.concerns.authentication.unauthenticated_message")
 
       respond_to do |format|
         format.html do
-          redirect_to Settings.one_login.profile_url, allow_other_host: true
+          redirect_to find_root_path, flash: { info: flash_info_message }
         end
 
         format.json do
-          render json: { redirect: Settings.one_login.profile_url }, status: :unauthorized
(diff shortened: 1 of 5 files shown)


app/components/notification_banner.html.erb
```diff
@@ -6,7 +6,7 @@
   </div>
   <div class="govuk-notification-banner__content">
     <p class="govuk-notification-banner__heading">
-      <%= text %>
+      <%= text&.html_safe %>
```
inulty-dfe, line 9: This has the potential to be a security issue if someone puts something in the flash message from user input.

I think we should look at implementing something like what is proposed in this SO answer

https://stackoverflow.com/questions/26538891/flash-message-with-html-safe-from-the-controller-in-rails-4-safe-version

PR #5678, opened 2025-10-01
Fix start date options

tomas-stefano:
## Context

This PR fix how start date options are calculated and sliced for `Course` models. 

Previously, the code used the current timetable year for slicing, which caused failures when the current date was before the start of the recruitment cycle or in ambiguous cycle transitions. 

This led to incorrect start months being presented to users.

## How it works now

	•	The available start date options are always built for the course’s recruitment cycle (January–December + January–July of next year).
	•	The slicing logic now considers the current calendar date in context of the cycle boundaries:
	•	If today is before the cycle starts, the options always begin from January of the cycle year - all options are returned.
	•	If today is in the cycle year, slicing starts from today’s month of the cycle year.
	•	Otherwise, all options are returned.

### DIFF
--- a/app/models/concerns/courses/edit_options/start_date_concern.rb
+++ b/app/models/concerns/courses/edit_options/start_date_concern.rb
@@ -4,44 +4,39 @@ module Courses
   module EditOptions
     module StartDateConcern
       extend ActiveSupport::Concern
+
       included do
         def start_date_options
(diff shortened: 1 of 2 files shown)


tomas-stefano:
https://trello.com/c/fcUbhEme/1153-need-to-update-course-start-date-to-start-with-january-2026

PR #5684, opened 2025-10-02
Correct Omniauth failure param

inulty-dfe:
The param we receive to the failure endpoint is called strategy not
  provider.

  https://github.com/omniauth/omniauth/blob/master/lib/omniauth/failure_endpoint.rb#L43## Context

## Checklist


inulty-dfe:
We're going to park debugging the errors with One Login for now

PR #5691, opened 2025-10-07
Skip the AssignVisaSponsorshipApplicationDeadlineService without date

inulty-dfe:
## Context

  The AssignVisaSponsorshipApplicationDeadlineService only concerns the
  setting of the visa sponsorship date AFTER the provider decides there
  should be a visa sponsorship deadline date. We can avoid calling this
  service unless the date params are present.

## Checklist


sarahcrack (approved):
👍 LGTM

PR #5703, opened 2025-10-13
Td/19 tracking which information find users interact with redirect links

tomas-stefano:
## Context

As part of new and upcoming Find KPI dashboards, we want a proxy to measure which pieces of information users are engaging with on Find. We therefore want to be able to track all links (this PR).

But there are lots of other redirects that a candidate can interact with on Find. We would like to establish the same redirects for all of these other links OR use the new track_clicks to redo tracking for all information clicks.

## Guidance to review

1. All external links added points to the track click event and the redirect works?

### DIFF
--- a/app/controllers/track_controller.rb
+++ b/app/controllers/track_controller.rb
@@ -52,7 +52,11 @@ private
 
     uri = URI.parse(url)
 
-    return url if uri.relative? || host_allowed?(uri.host)
+    if uri.relative? && uri.host.blank? && uri.scheme.blank?
+      return url
+    end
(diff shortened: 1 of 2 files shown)


langesi:
@tomas-stefano  Found a security issue, try:

https://find-review-5703.test.teacherservices.cloud/track_click?url=//attacker.example.com&utm_content=test

URI.parse treats protocol-relative URLs (//evil.com) as relative, the controller will redirect to any host supplied with a leading //

will comment in a second where to change things to fix this

PR #5706, opened 2025-10-14
Service rebrand - colours and logo for Find and Publish

inulty-dfe:
## Context

This PR will update the Header in Find, Publish and Support to use the rebranded design for Govuk.

Find currently uses a service header provided by One Login.

- Update `govuk-one-login/service-header` package
  - We can now import the javascript directly from the npm package, so deleted the old one
  - copied across new html
- Update `govuk-frontend package`
- Pass in the rebranded html from the service-header (`rebranded-*`)
- Apply classes to form and button, and add styles so they apply to the button (only on links in the source)
- Create a view component preview for the header
  - move one-login styles to be shared by publish and find (needed for the preview)
- Update govuk-components to v5.11.1 (later versions require upgrading `view_components` gem.
  - This version updates the govuk_header to use the rebranded assets
- Added a preview for the one login header

## Not updated here

- Maintenence page to use updated govuk-frontend with new branding
- Service unavailable pages

## Guidance to review

Our Sign in / Sign out are buttons in Find. The One Login service-header styles do not work with buttons so we've had to conjure some custom styles to make the Sign in / Sign out buttons look like links in the One Login header.

Check multiple viewport sizes.
Check multiple focus / active states of the buttons

#### Signed Out

|Desktop|Mobile|
|---|---|
|image|image|

#### Signed In

|Desktop|Mobile|
|---|---|
|image|image|

## Publish
|Signed out|Signed in|
|---|---|
|image|image|

## Checklist

### DIFF
--- a/app/assets/stylesheets/one-login.scss
+++ b/app/assets/stylesheets/one-login.scss
@@ -1,16 +1,16 @@
 @import "govuk-one-login/service-header/dist/styles/service-header-no-imports";
 
-.rebranded-one-login-header__nav__link-content {
-  min-height: 30px;
-}
-
 .one-login-header__button-form {
   display: contents;
 }
 
 // Make a button look like a link
+// in the One Login Gov.uk header
 .one-login-header__button-link {
   display: inline;
+
+  font-family: "GDS Transport", arial, sans-serif;
+  -webkit-font-smoothing: antialiased;
 
   font-size: inherit;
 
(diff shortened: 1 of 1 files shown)


inulty-dfe:
[Implement new gov.uk designs on Publish and Find (branding)](https://trello.com/c/haapkOW7/1128-implement-new-govuk-designs-on-publish-and-find-branding)

langesi (approved):
Like it, amazing job. My suggestion is that we should have different colour of headers for different envs, but we can work on this separately

inulty-dfe:
> Like it, amazing job. My suggestion is that we should have different colour of headers for different envs, but we can work on this separately

I think that's a very sensible suggestion. I do not see a provision for that in the new designs. I will try get some feedback on that from the Design system people,

I have seen it on One Login site...

image

PR #5709, opened 2025-10-14
Add accredited provider content on accredited provider page

sarahcrack:
## Context

Trello: https://trello.com/c/9oGw8xTh/434-add-accredited-provider-content-on-accredited-provider-page

We are adding a call out box with content on the accredited providers information page where the accredited provider is different from the training provider. This is to explain to users the difference between an accredited provider and a training provider.

- created `WhatIsAnAccreditedProviderComponent` which can be rendered on the '/accredited-by' page on Find
- only applies to courses where the training provider is different from the accrediting provider

## Guidance to review

- choose a course that is accredited by a different organisation to the training provider
- go to the 'Course summary' and click on the provider in the 'Accredited by' row to reach the page in question - the callout box should be present as below:

Screenshot 2025-10-14 at 13 09 09

## Checklist


kelliedesigner:
Looks great thanks Sarah!

PR #5721, opened 2025-10-22
Display ratifying providers' value proposition in Find

inulty-dfe:
## Context

We were not pulling in the `value_proposition` column of the ratifying provider in Find.

## Guidance to review

bug

## Checklist

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -519,6 +519,10 @@ class CourseDecorator < ApplicationDecorator
     object.ratifying_provider_description
   end
 
+  def accrediting_provider_value_proposition
+    object.ratifying_provider_value_proposition
+  end
+
   def has_physical_education_subject?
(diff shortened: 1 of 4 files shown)


inulty-dfe:
[Bug: 'Accredited by' section is pulling through incorrect information for Saffron TSH](https://trello.com/c/MkueQCYL/1205-bug-accredited-by-section-is-pulling-through-incorrect-information-for-saffron-tsh)

app/decorators/course_decorator.rb
```diff
@@ -519,6 +519,10 @@ def about_accrediting_provider
     object.ratifying_provider_description
   end
 
+  def accrediting_provider_value_proposition
+    object.ratifying_provider_value_proposition
+  end
```
tomas-stefano: What is the real value of having this method? And also with different name?

inulty-dfe: Good question. I don't see much value at all

PR #5728, opened 2025-10-23
Do show degree subject requirements if there is any all the time

tomas-stefano:
## Context

We have this support ticket:  https://trello.com/c/nvz900fV/1045-incomplete-information-listed-for-ljmu-courses

The degree subject requirements are not showing for many courses (see support ticket for more examples).

## Solution

We should not distinguish to show degree subject requirements if is SKE subjects or not. We should always show the degree subject requirements if there is any, and only show SKE section for SKE subjects.


inulty-dfe (commented):
A little formatting suggestions

PR #5752, opened 2025-11-03
Fix auto publish for description fields

gms-gs:
## Context

When removing the long form course content feature flag I noticed when saving data via the course description fields that it was updating the changes on Find without hitting publish

Update the view logic to ensure its using the correct field data.

## Guidance to review

- Create a course, when drafting and updating the fields this i updating as expected
- Create a course and publish it
- When updating the description fields data and then viewing the course in Find the fields wont update. Only when you publish then you will see the changes. 

## Checklist

### DIFF
--- a/app/components/find/courses/summary_component/view.rb
+++ b/app/components/find/courses/summary_component/view.rb
@@ -8,7 +8,7 @@ module Find
         include ::ViewHelper
         include PreviewHelper
 
-        attr_reader :course
+        attr_reader :course, :enrichment
 
         delegate :accrediting_provider,
                  :provider,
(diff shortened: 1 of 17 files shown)


app/components/find/courses/summary_component/view.rb
```diff
@@ -93,11 +94,11 @@ def show_apply_from_row?
 
       private
 
-        def uk_fees(fee_uk = course.enrichment_attribute(:fee_uk_eu))
+        def uk_fees(fee_uk = enrichment&.fee_uk_eu.presence || course.published_fee_uk_eu.presence || (preview?(helpers.params) && course.fee_uk_eu))
           t(".fee_value.fee.uk_fees_html", value: content_tag(:b, number_to_currency(fee_uk.to_f))) if fee_uk.present?
         end
 
-        def international_fees(fee_international = course.enrichment_attribute(:fee_international))
+        def international_fees(fee_international = enrichment&.fee_international.presence || course.published_fee_international.presence || (preview?(helpers.params) && course.fee_international))
```
tomas-stefano: [Same as above](https://github.com/DFE-Digital/publish-teacher-training/pull/5752/files#r2502593844)

PR #5774, opened 2025-11-18
Lock Sidekiq version below 7

inulty-dfe:
## Context

  Sidekiq v7 requires Redis v7+ which we don't currently have access to.
  This is a precautionary version locking.

## Checklist


inulty-dfe:
[Lock Sidekiq version below 7](https://trello.com/c/kkcDM9MG/1269-lock-sidekiq-version-below-7)

PR #6042, opened 2026-04-09
Fix TypeError when geocoding returns nil coordinates

tomas-stefano:
## Context

When a user visits a course page with a `location` query param that Google Places fails to resolve (blank query, API error, or missing coordinates), `Geolocation::Address.query` returns an Address with `latitude: nil` and `longitude: nil`. `Find::CoursesController#distance_from_location` then passed these straight to `Courses::NearestSchoolQuery`, whose `select_sql` interpolates `Float(@longitude)` / `Float(@latitude)` into the `ST_MakePoint` call. `Float(nil)` raises `TypeError: can't convert nil into Float`, 500ing the course show page.

Guard the controller with an early return when the resolved address has no coordinates. The existing `Find::Courses::TrainingLocations::View` component already falls back to the "no address" branch when `@distance_from_location` is unset, so no view changes are needed.

Adds a regression spec that stubs `Geolocation::Address.query` to return nil coordinates and asserts `#show` does not raise.

### DIFF
--- a/app/controllers/find/courses_controller.rb
+++ b/app/controllers/find/courses_controller.rb
@@ -32,6 +32,8 @@ module Find
 
     def distance_from_location
       @address = Geolocation::Address.query(location_params)
+      return if @address.latitude.nil? || @address.longitude.nil?
+
       @distance_from_location ||= ::Courses::NearestSchoolQuery.new(
         courses: [@course],
         latitude: @address.latitude,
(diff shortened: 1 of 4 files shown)


app/controllers/find/courses_controller.rb
```diff
@@ -32,6 +32,8 @@ def location_params
 
     def distance_from_location
       @address = Geolocation::Address.query(location_params)
+      return if @address.latitude.nil? || @address.longitude.nil?
```
inulty-dfe: would it be better to define a method on the address for this?

tomas-stefano: Done

inulty-dfe (approved):

PR #6046, opened 2026-04-10
School remodel add urn to existing main sites

tomas-stefano:
## Context

As part of moving `Site` records onto GIAS-backed relationships, we need to populate `sites.urn` where it's currently `NULL` on current-cycle school sites. These are legacy "main site" records with the placeholder `code = "-"` (see `app/models/site.rb:46`, which exempts this case from URN presence validation). Reducing the manual cleanup before the full remodel is the goal.

Scope is **strictly the current recruitment cycle** — other cycles are not touched.

## What this PR does

Adds a new data migration: `db/data/20260410105723_backfill_main_site_urns.rb`.

For each current-cycle school site with `urn: nil`:

1. Normalizes the site's postcode (strip whitespace, upcase).
2. Queries `GiasSchool.available` (open / proposed_to_open / proposed_to_close) for schools whose normalized postcode matches.
3. If **exactly one** match → assigns the URN via `update_column` (skips validations, callbacks, and geocoding — address data is unchanged).
4. If **zero** or **multiple** matches → the record is logged to stdout for manual follow-up and not touched.

Safety properties:

- **Idempotent** — `where(urn: nil)` guarantees re-runs are no-ops on already-populated records, and never overwrite existing URNs.
- **Cycle-scoped** — uses `RecruitmentCycle.current.sites` so other cycles are untouched.
- **Conservative** — only auto-assigns on a single confident match against an active GIAS school.
- **Discard-aware** — soft-deleted sites are excluded via `Site.kept`.

Also establishes `spec/data/` as the convention for testing data migrations, with 11 examples covering:

- single-match backfill, idempotency, postcode normalization
- ambiguous / no-match / closed-school skipping
- cross-cycle isolation
- preservation of existing URNs
- study-site and discarded-site exclusions
- `down` irreversibility

## How matching was done

Normalized-postcode equality between `Site.postcode` and `GiasSchool.postcode` (both stripped of whitespace and upcased), filtered to `GiasSchool.available`. Only a single-match result triggers an assignment.

## Results (fill in after running on production)

- Current-cycle records updated automatically: **TBD**
- Current-cycle records left unresolved (no_match): **TBD**
- Current-cycle records left unresolved (ambiguous): **TBD**

The unresolved list is printed under `[BackfillMainSiteUrns] UNRESOLVED (...)` in the migration output and will be pasted into the ticket for @simonlange27.

## Running the migration

This migration is not auto-run on deploy — it needs to be invoked manually in each environment. From a web pod shell:

```sh
bundle exec rails data:migrate
```

Per environment (run from the repo root):

```sh
# QA
make qa get-cluster-credentials
make qa shell
# then inside the pod:
bundle exec rails data:migrate

# Staging
make staging get-cluster-credentials
make staging shell
bundle exec rails data:migrate

# Sandbox
make sandbox get-cluster-credentials
make sandbox shell
bundle exec rails data:migrate

# Production
make production get-cluster-credentials CONFIRM_PRODUCTION=yes
make production shell CONFIRM_PRODUCTION=yes
bundle exec rails data:migrate
```

Capture the `[BackfillMainSiteUrns] updated=... no_match=... ambiguous=...` summary and the unresolved list from each run — production output goes into this PR description and the ticket.

## Test plan

- [x] `bundle exec rspec spec/data/backfill_main_site_urns_spec.rb` — 11 examples, 0 failures
- [x] `bundle exec rubocop spec/data/backfill_main_site_urns_spec.rb db/data/20260410105723_backfill_main_site_urns.rb` — no offences
- [ ] Run on QA, verify update count and unresolved list
- [ ] Run on staging, verify update count and unresolved list
- [ ] Run on production, paste results into this PR and the ticket
- [ ] Hand unresolved list to @simonlange27 for manual cleanup

### DIFF
--- a/db/data/20260410105723_backfill_main_site_urns.rb
+++ b/db/data/20260410105723_backfill_main_site_urns.rb
@@ -1,0 +1,56 @@
+# frozen_string_literal: true
+
+class BackfillMainSiteUrns < ActiveRecord::Migration[8.1]
+  TAG = "[BackfillMainSiteUrns]"
+
+  def up
+    updated = []
+    no_match = []
+    ambiguous = []
+
+    RecruitmentCycle.current.sites
+      .kept
+      .where(site_type: :school, urn: nil)
+      .find_each do |site|
+        normalized = site.postcode.to_s.gsub(/\s+/, "").upcase
+        if normalized.blank?
+          no_match << site
+          next
+        end
+
+        matches = GiasSchool.available
+          .where("REPLACE(UPPER(postcode), ' ', '') = ?", normalized)
+          .pluck(:urn)
+
+        case matches.size
+        when 0
+          no_match << site
+        when 1
+          site.update_column(:urn, matches.first)
+          updated << [site, matches.first]
+        else
+          ambiguous << [site, matches]
+        end
+      end
+
+    say "#{TAG} updated=#{updated.size}, no_match=#{no_match.size}, ambiguous=#{ambiguous.size}"
+
+    if no_match.any?
+      say "#{TAG} UNRESOLVED (no_match):"
+      no_match.each do |site|
+        say "  provider_code=#{site.provider_code} site_id=#{site.id} location=#{site.location_name.inspect} postcode=#{site.postcode.inspect}"
+      end
+    end
+
+    if ambiguous.any?
+      say "#{TAG} UNRESOLVED (ambiguous):"
+      ambiguous.each do |site, urns|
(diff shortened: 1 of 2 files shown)


tomas-stefano:
https://trello.com/c/64lhHmSy/1584-1-school-remodel-add-urn-to-existing-main-sites

inulty-dfe:
What about empty string URNs or invalid ones?

tomas-stefano:
> What about empty string URNs or invalid ones?

Done. Now is much better. But still 30 unresolved — all in the no_match bucket (0 ambiguous). 

These are the remaining sites whose postcode didn't match any available GIAS school.

inulty-dfe (approved):

PR #6047, opened 2026-04-10
Denote non-uk financial incentives with major refactor of Financial Incenteives

inulty-dfe:
## Context

We calculate financial incentives and the text that the financial incentive should produce in the UI in different view components. This makes it difficult to understand and to change.

We'd like to move all the financial incentive calculations to a single place and use those calculations to produce the text that should appear in Find.

We have consulted some people from the relevant teams that are in charge of managing financial incentives to understand exactly what the rules are for awarding financial incentives and used this knowledge to produce a document explaining the rules.

1. Extract all the places where bursaries are calculated in to `CourseIncentive` module.
2. All the translations and content are in `CourseIncentive::View`
3. `Add some documentation explaining the formula for how bursaries are applied to courses based on the subjects

Show `for UK citizens` in the financial incentive text if the incentives only apply to UK citizens.

## Guidance to review

1. Create courses with a variety of subject combinations and test that the financial incentives are applied correctly
2. Test that the "for UK citizens" text appears on the courses where it is expected

### Take aways

 - We should find a way to reference subjects consistently
   - name
   - code
   - id

## Checklist


tomas-stefano (approved):

app/controllers/find/secondary_subjects_controller.rb
```diff
@@ -39,6 +37,18 @@ def formatted_secondary_subject_options
       end
     end
 
+    def financial_information(financial_incentive)
```
inulty-dfe, line 40: I don't like that this method is here and that we pass granular values into `CourseIncentive::View.hint_text` but I feel we've done enough here and need to move on

inulty-dfe:
[Change wording around bursaries](https://trello.com/c/0AhuenZN/1431-change-wording-around-bursaries)

inulty-dfe:
[Refactor Bursaries and subjects](https://trello.com/c/5ga23Hzx/1560-refactor-bursaries-and-subjects)

PR #6052, opened 2026-04-14
Concurrent logins & Permissions-Policy header

tomas-stefano:
## Context

Addresses two findings from ITHC FF25-567 (DfE Publish/Find IT health check).

**1. Concurrent logins (pp. 21–22)** — the same account could authenticate from multiple browsers simultaneously and every session stayed valid until the 30-minute inactivity timeout, undermining compromise detection and session auditing. Sign-in now destroys any existing `Session` rows for the authenticating `Candidate` (Find) or `User` (Publish) inside the same transaction that creates the new one, enforcing a single active session per user as recommended by the OWASP Session Management Cheat Sheet.

**2. Permissions-Policy header (pp. 18–20)** — the header was absent entirely. The previously-empty `config/initializers/permissions_policy.rb` now emits a restrictive baseline disabling `camera`, `microphone`, `geolocation`, `gyroscope`, `magnetometer`, `usb`, and `payment`, and limits `fullscreen` to `self`. None of these features are used by the app — the in-app "geolocation" is server-side Google Geocoding, not the browser API, and the only `navigator.*` usage is `navigator.clipboard`, which is not restricted.

The third ITHC finding (HSTS `preload` directive, same section of the report) is **not** in this PR — it's blocked on clarification from the infra/platform team about where the current `Strict-Transport-Security` header is emitted (Rails `force_ssl` is commented out, so it must be set upstream). The open questions and proposed options are documented in `guides/ithc-ff25-567-remediation.md §4` and will be handled in a follow-up PR.

## Behaviour change for end users

After this PR, signing in on a new device/browser silently invalidates any existing session the same user has elsewhere. On the old device, the next request returns the user to the sign-in page. This is intentional and aligns with OWASP guidance, but worth flagging to support so they can answer any "why was I signed out?" tickets.

## Test plan

- [ ] `bundle exec rspec spec/requests/find/sign_in_spec.rb spec/requests/publish/concurrent_sign_in_spec.rb spec/requests/security_headers_spec.rb` passes in CI.
- [ ] Full test suite green in CI.
- [ ] Manual: sign in as the same Candidate in two separate browsers (normal + incognito); refresh browser A → redirects to sign-in.
- [ ] Manual: same check for Publish via DfE Sign-in.
- [ ] Manual: `curl -sI https://<review-app-host>/ | grep -i permissions-policy` shows the new header with the expected directives on both Find and Publish pages.
- [ ] Manual sweep of Find and Publish in a browser to confirm no UI relies on a now-blocked browser feature (search, course listings, saved courses, publish provider admin).

### DIFF
--- a/config/initializers/permissions_policy.rb
+++ b/config/initializers/permissions_policy.rb
@@ -1,18 +1,41 @@
 # frozen_string_literal: true
 
-# Be sure to restart your server when you modify this file.
+# Emit a Permissions-Policy header that restricts powerful browser features
+# neither Find nor Publish use, per ITHC FF25-567.
+#
+# Rails' built-in `config.permissions_policy` DSL still emits the deprecated
+# `Feature-Policy` header name with the old `camera 'none'` syntax
+# (see actionpack lib/action_dispatch/http/permissions_policy.rb), which
+# does not satisfy the ITHC requirement. We set the modern header directly
+# via a small Rack middleware instead.
+#
+# "geolocation" here refers to the browser API — the app's own geolocation
+# features are server-side Google Geocoding calls and are unaffected.
 
-# Restrict powerful browser features that neither Find nor Publish use, per
-# ITHC FF25-567 (see guides/ithc-ff25-567-remediation.md). Geolocation here
-# refers to the browser API — the app's own geolocation features are
-# server-side Google Geocoding calls and are unaffected.
-Rails.application.config.permissions_policy do |policy|
-  policy.camera       :none
-  policy.microphone   :none
-  policy.geolocation  :none
-  policy.gyroscope    :none
-  policy.magnetometer :none
-  policy.usb          :none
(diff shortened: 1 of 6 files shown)


config/initializers/permissions_policy.rb
```diff
@@ -1,15 +1,23 @@
...
+Rails.application.config.action_dispatch.default_headers["Permissions-Policy"] = [
+  "camera=()",
+  "microphone=()",
+  "geolocation=()",
+  "gyroscope=()",
+  "magnetometer=()",
+  "usb=()",
+  "payment=()",
+  "fullscreen=(self)",
+].join(", ")
```
tomas-stefano, line 23: As the commit history tells I tried the syntax sugar DSL but Rails 8 don't set the headers we expect. See commit history for mode details.

inulty-dfe (approved):

PR #6057, opened 2026-04-15
Various content and UI changes to email alerts

inulty-dfe:
## Context

We have completed 90% of the email alerts feature. We have some snags that need to be corrected before we can launch.

- [x] Fix email criteria formatting
Remove the Subject text from the email body
- [ ] Unsubscribe button should fill width (email body)
We can't control the Notify email body formatting in the ways we'd like so that item cannot be complete.
- [x] Unsubscribe button should fill width (in service)
Apply 100% width on the button form
- [x] Hide Email courses link when no filters applied
Control when the email subscribe button appears
- [x] Email me courses like this should be in the filter panel
Move the link to the filter panel and move the clear all link next to the heading
- [x] Fix formatting on email alert confirmation summary table
Needs to be QA'd
- [x] Fix formatting in email alert create success flash banner
New lines respected and unsubscribe link added
- [x] Email alert index should be 2/3 width
Done

### Extra
Add coordinates, location and radius to FilterKeyDigest
Remove application_open from filter digest

### Hide 'Email me courses liek this' link conditionally

This is the most complicated part of this PR.

|Condition| Result|
|---|---|
|When a candidate is not logged in| Hide|
|When a search does not have specific filters |Hide|
|When a search contains the same specific filters as an existing email alert| Hide|

## Guidance to review

What if we add a new filter to Find? What affect would that have on the email alert digest?
## Checklist

### DIFF
--- a/app/models/candidate/email_alert.rb
+++ b/app/models/candidate/email_alert.rb
@@ -13,17 +13,6 @@ class Candidate
     validate :subscription_limit_not_reached
 
     scope :active, -> { where(unsubscribed_at: nil) }
-    scope :subscribed, -> { active }
-
-    def matches_search?(subjects:, search_attributes:)
-      other_attrs = search_attributes.to_h.stringify_keys
-      self.subjects.sort == Array(subjects).sort &&
-        normalize_filter_attrs(self.search_attributes) == normalize_filter_attrs(other_attrs)
-    end
-
-    def filter_key
-      [subjects.sort, normalize_filter_attrs(search_attributes)]
-    end
 
     def search_params
       params = search_attributes.symbolize_keys
--- a/app/models/candidate/email_alert.rb
+++ b/app/models/candidate/email_alert.rb
@@ -15,7 +15,7 @@ class Candidate
     scope :active, -> { where(unsubscribed_at: nil) }
 
     def search_params
-      params = search_attributes.symbolize_keys
+      params = (search_attributes || {}).symbolize_keys
       params[:subjects] = subjects if subjects.present?
       params[:longitude] = longitude if longitude.present?
       params[:latitude] = latitude if latitude.present?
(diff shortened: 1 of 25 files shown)


inulty-dfe:
[Updates to email alerts before going live](https://trello.com/c/gV75P1rr/1600-updates-to-email-alerts-before-going-live)

inulty-dfe:
[Updates to email alerts before going live](https://trello.com/c/gV75P1rr/1600-updates-to-email-alerts-before-going-live)

app/models/candidate/email_alert.rb
```diff
@@ -13,20 +13,9 @@ class EmailAlert < ApplicationRecord
...
-        normalize_filter_attrs(self.search_attributes) == normalize_filter_attrs(other_attrs)
-    end
-
-    def filter_key
-      [subjects.sort, normalize_filter_attrs(search_attributes)]
-    end
 
     def search_params
-      params = search_attributes.symbolize_keys
+      params = (search_attributes || {}).symbolize_keys
```
tomas-stefano, line 18: EmailAlert#search_params was made nil-safe this branch ((search_attributes || 
  {})), but RecentSearch#search_params still does search_attributes.symbolize_keys.
  Since `FilterKeyDigestable#compute_filter_key_digest` now routes through search_params, a RecentSearch with nil search_attributes raises on save. The  filter_key_digestable_spec "handles nil search_attributes" test was deleted.
  Validation probably prevents nil, but the asymmetry is a latent footgun — let's make both consistent.

tomas-stefano:
Off topic; Is the email preview should work on this PR? https://publish-review-6057.test.teacherservices.cloud/rails/mailers/email_alert_mailer/weekly_digest_few_courses

tomas-stefano (approved):
Minor considerations, looks good to me 🚀

PR #6058, opened 2026-04-16
School remodel backfill join tables

tomas-stefano:
## Summary

Implements a bulk SQL backfill that populates the new `provider_school` and `course_school` join tables from the existing `site` / `course_site` data model. School URNs on `site` are resolved to `gias_school.id` via a direct join on `urn`. The backfill is wrapped in a `DataHub::SchoolsBackfillProcessSummary` for observability and invoked via a rake task.

- **SQL-first**: two `INSERT ... SELECT ... ON CONFLICT DO NOTHING` statements do all the work — no Ruby per-row loops, no ActiveRecord instantiation
- **Idempotent by construction**: `ON CONFLICT DO NOTHING` covers both the composite unique index and the partial unique index (one main site per provider), so reruns are free
- **Parity with source**: all `course_site` rows are copied regardless of `publish` / `status` — the new `course_school` table has no publish/status column, so filtering here would destroy history; retrieval logic is a separate ticket
- **Skip reporting**: sites with no URN or an unresolvable URN are written to CSV files under `tmp/`, with paths and counts recorded on the process summary

### Approach — why bulk SQL

The dev DB holds 100k+ site rows and 600k+ course_site rows. 

A Ruby `.each` loop with individual `create!` calls would instantiate hundreds of thousands of ActiveRecord objects and run for minutes. The two SQL statements complete in seconds and use constant memory. `ON CONFLICT DO NOTHING` makes the process safely repeatable without needing to check for existing rows first.

### How to run

```
bundle exec rails schools_backfill:run
```

Outputs `short_summary` (counts) and `full_summary` (CSV paths) to stdout. The run is also recorded as a `DataHub::SchoolsBackfillProcessSummary` row (STI on `data_hub_process_summary`), observable via:

```ruby
DataHub::SchoolsBackfillProcessSummary.last.short_summary
# => {"provider_schools_inserted"=>108176, "course_schools_inserted"=>685256, "sites_skipped"=>..., "course_sites_skipped"=>...}
```

### What gets backfilled

| Source | Target | Join condition | Filters |
|--------|--------|---------------|---------|
| `site` | `provider_school` | `gias_school.urn = site.urn` | `site_type = 0` (school), not discarded, URN present and non-blank |
| `course_site → site` | `course_school` | `gias_school.urn = site.urn` | Same site filters; all course_site rows regardless of publish/status |

`site.code` maps directly to `site_code` on the new tables (no transformation). `site_code = '-'` represents the provider main site, enforced by the partial unique index from the previous PR.

### What gets skipped (and logged)

- Sites with `urn IS NULL` or `urn = ''` → reason: `no_urn`
- Sites whose URN doesn't match any `gias_school` row → reason: `urn_not_in_gias_school`
- Discarded sites and study sites are excluded from both inserts **and** from the skipped CSVs (they're expected exclusions, not actionable gaps)

### Performance trade-offs

- Single transaction wraps both tables — if the course_school insert fails, provider_school inserts roll back too. Acceptable because ON CONFLICT makes the full re-run cheap.
- `course_site` has no timestamps; `created_at` / `updated_at` on new rows are set to `NOW()` at insert time — downstream code should not use these as "when was the relationship established".
- CSVs live in `tmp/` — on containerised hosts they vanish on pod restart. Read them before cycling the pod, or pipe them to blob storage in a follow-up if needed.

## New files

| File | Purpose |
|------|---------|
| `app/models/data_hub/schools_backfill_process_summary.rb` | STI subclass with `jsonb_accessor` typed fields for counts + CSV paths |
| `app/services/data_hub/schools_backfill/executor.rb` | Core service: skip CSVs → bulk insert → finish summary |
| `lib/tasks/schools_backfill.rake` | `schools_backfill:run` thin wrapper |
| `spec/services/data_hub/schools_backfill/executor_spec.rb` | 13 examples: happy path, skipped rows, idempotency, process summary lifecycle, rollback on error, multi-cycle coverage |
| `spec/lib/tasks/schools_backfill_rake_spec.rb` | Smoke test for the rake task |

## Test plan

- [ ] `bundle exec rspec spec/services/data_hub/schools_backfill/executor_spec.rb spec/lib/tasks/schools_backfill_rake_spec.rb` — all green
- [ ] Run `bundle exec rails schools_backfill:run` on a seeded dev DB — confirm non-zero `provider_schools_inserted` / `course_schools_inserted`
- [ ] Re-run the rake task — confirm a second summary with both inserted counts at `0` (idempotency)
- [ ] Inspect a skipped-rows CSV: `head tmp/schools_backfill_skipped_sites_<id>.csv`
- [ ] `DataHub::SchoolsBackfillProcessSummary.last` shows `status: "finished"` with populated summaries
- [ ] `git diff db/schema.rb` is empty — no schema changes in this PR

### DIFF
--- a/app/models/data_hub/schools_backfill_process_summary.rb
+++ b/app/models/data_hub/schools_backfill_process_summary.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module DataHub
+  class SchoolsBackfillProcessSummary < ProcessSummary
+    jsonb_accessor :short_summary,
+                   provider_schools_inserted: [:integer, { default: 0 }],
+                   course_schools_inserted: [:integer, { default: 0 }],
+                   sites_skipped: [:integer, { default: 0 }],
+                   course_sites_skipped: [:integer, { default: 0 }]
+
+    jsonb_accessor :full_summary,
+                   skipped_sites_csv_path: :string,
+                   skipped_course_sites_csv_path: :string
+  end
+end
--- a/app/models/data_hub/schools_backfill_process_summary.rb
+++ b/app/models/data_hub/schools_backfill_process_summary.rb
@@ -9,7 +9,7 @@ module DataHub
                    course_sites_skipped: [:integer, { default: 0 }]
 
     jsonb_accessor :full_summary,
-                   skipped_sites_csv_path: :string,
-                   skipped_course_sites_csv_path: :string
+                   skipped_sites: [:jsonb, { array: true, default: [] }],
+                   skipped_course_sites: [:jsonb, { array: true, default: [] }]
   end
 end
--- a/app/services/data_hub/schools_backfill/executor.rb
+++ b/app/services/data_hub/schools_backfill/executor.rb
@@ -1,0 +1,143 @@
+# frozen_string_literal: true
+
+require "csv"
+require "fileutils"
+
+module DataHub
+  module SchoolsBackfill
+    class Executor
+      def initialize(tmp_dir: Rails.root.join("tmp"))
+        @tmp_dir = Pathname.new(tmp_dir)
+      end
+
+      def execute
+        process_summary = DataHub::SchoolsBackfillProcessSummary.start!
+
+        skipped_sites_csv_path        = @tmp_dir.join("schools_backfill_skipped_sites_#{process_summary.id}.csv")
+        skipped_course_sites_csv_path = @tmp_dir.join("schools_backfill_skipped_course_sites_#{process_summary.id}.csv")
+
+        ActiveRecord::Base.transaction do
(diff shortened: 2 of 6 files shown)


langesi (approved):
I’m good with this. Functionally it works as expected.

I do have one question though. why are we generating a CSV rather than storing the skipped-row data directly in the process summary? At the moment, only the file path is stored in full_summary. As this is in temp, this won't be persistent for long. Is that because the skipped data could become too large for the jsonb field?

inulty-dfe:
[3. School remodel - Backfill script for new join tables](https://trello.com/c/U3lCA2kt/1586-3-school-remodel-backfill-script-for-new-join-tables)

PR #6061, opened 2026-04-17
Remove Application open date filter and feature flag

inulty-dfe:
## Context

We have hidden the filter for courses application open date for a long time now, and we return the first day of the cycle as the value for this property in the API.

We are now removing the FeatureFlag that hides the filter in the UI and removing the filter itself.

Next step is to remove the property entirely from the system.

## Guidance to review

Remove feature flag
Update all default code paths to remove applications open
Update specs

## Checklist


inulty-dfe:
[Remove 'Applications open' filter Feature flag](https://trello.com/c/8ft8usJp/1092-remove-applications-open-filter-feature-flag)

tomas-stefano:
After this change, these still exist but may no longer reach a live code path in the add-course flow:                                          
  - `app/models/concerns/courses/edit_options/applications_open_concern.rb` — show_applications_open? is still piped through edit_options.rb, but the step it gated is gone from the add flow. Worth verifying whether any view still reads edit_options[:show_applications_open].               
  - `app/decorators/course_decorator.rb:223` — applications_open_first_label(recruitment_cycle) has no remaining callers I can find in the code.

gms-gs (approved):

inulty-dfe:
Thanks for the review @tomas-stefano , I've dealt with those two points

PR #6067, opened 2026-04-20
Implement add course flow wizard setup and Level step

gms-gs:
## Context

Implement DfE wizard for the add course flow. This PR adds the basic skeleton implementation for add course flow using the DfE wizard as well as the first step of this flow (Level step) 

- Add `refactored_add_course_flow` feature flag
- Update add course button logic to take user through DfE wizard route if activated
- Implemented the basic structure for the Add Course flow

## Guidance to review

- Check when flag is disabled u are taken through the old flow
- Check when flag is enabled u are taken through the new flow (Only goes up to the Level page)

## Checklist

### DIFF
--- a/app/components/add_course_button.rb
+++ b/app/components/add_course_button.rb
@@ -24,6 +24,7 @@ class AddCourseButton < ViewComponent::Base
       new_publish_provider_recruitment_cycle_course_wizard_path(
         provider_code: provider.provider_code,
         recruitment_cycle_year: provider.recruitment_cycle_year,
+        state_key: wizard_state_key,
       )
     else
       new_publish_provider_recruitment_cycle_course_path(
@@ -37,6 +38,10 @@ private
 
   def wizard_add_course_flow?
     FeatureFlag.active?(:wizard_add_course_flow)
+  end
+
+  def wizard_state_key
+    @wizard_state_key ||= SecureRandom.uuid
   end
 
   def accredited_provider
--- a/app/components/add_course_button.rb
+++ b/app/components/add_course_button.rb
@@ -19,7 +19,25 @@ class AddCourseButton < ViewComponent::Base
     @incomplete_sections ||= [school, accredited_provider].compact
   end
 
+  def add_course_path
+    if wizard_add_course_flow?
+      new_publish_provider_recruitment_cycle_course_wizard_path(
+        provider_code: provider.provider_code,
+        recruitment_cycle_year: provider.recruitment_cycle_year,
+      )
+    else
+      new_publish_provider_recruitment_cycle_course_path(
+        provider_code: provider.provider_code,
+        recruitment_cycle_year: provider.recruitment_cycle_year,
+      )
+    end
+  end
+
 private
+
+  def wizard_add_course_flow?
+    FeatureFlag.active?(:wizard_add_course_flow)
+  end
 
   def accredited_provider
     return if accredited_partner_present?
--- a/app/controllers/publish/course_wizards_controller.rb
+++ b/app/controllers/publish/course_wizards_controller.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+module Publish
+  class CourseWizardsController < ApplicationController
+    CACHE_EXPIRY = 1.hour
+
+    before_action :authorize_course_creation
+    before_action :set_wizard, except: [:new]
+
+    helper_method def current_step
+      @wizard.current_step
+    end
+
+    helper_method def current_step_name
+      params[:step]&.to_sym || :level
+    end
+
+    def new
+      return render_schools_messages unless provider.sites.any?
+
+      redirect_to publish_provider_recruitment_cycle_course_wizard_path(
+        provider_code: params[:provider_code],
+        recruitment_cycle_year: params[:recruitment_cycle_year],
+        step: :level,
+      )
+    end
+
+    def show; end
+
+    def create
+      update
+    end
+
+    def update
+      if @wizard.save_current_step
+        redirect_to @wizard.next_step_path
+      else
+        render :show
+      end
+    end
+
+  private
+
+    def authorize_course_creation
+      authorize(provider, :can_create_course?)
+    end
+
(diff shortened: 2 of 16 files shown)


gms-gs:
[Implement Refactored add course flow Feature Flag and Level page](https://trello.com/c/jZC1KNul/1626-implement-refactored-add-course-flow-feature-flag-and-level-page)

app/components/add_course_button.rb
```diff
@@ -19,8 +19,26 @@ def incomplete_sections
...
+        provider_code: provider.provider_code,
+        recruitment_cycle_year: provider.recruitment_cycle_year,
+      )
+    end
+  end
+
 private
 
+  def refactored_add_course_flow?
+    FeatureFlag.active?(:refactored_add_course_flow)
```
tomas-stefano: We are going to renamed the feature flag name. Minor.

app/controllers/publish/course_wizards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def render_schools_messages
+      flash[:error] = { id: "schools-error", message: "You need to create at least one school before creating a course" }
+
+      redirect_to publish_provider_recruitment_cycle_schools_path(
+        provider.provider_code,
+        provider.recruitment_cycle_year,
+      )
+    end
```
tomas-stefano, line 56: Can you flag this to Kellie. Sound a weird design rule.

app/controllers/publish/course_wizards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def set_wizard
+      state_store = CourseWizard::StateStores::CourseWizardStore.new(
+        repository: DfE::Wizard::Repository::Cache.new(
```
tomas-stefano, line 60: We don't need to change now, but maybe will be good to change to use redis directly to not put on redis-cache

app/controllers/publish/course_wizards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+          cache: Rails.cache,
+          key: "course_wizard_#{params[:provider_code]}_#{params[:recruitment_cycle_year]}",
+          expires_in: 1.hour,
```
tomas-stefano: Let's move this a constant.

And also because we are storing this temporarily I'd recommed to use also the state key https://github.com/DFE-Digital/dfe-wizard#multiple-wizard-instances-state_key
to have multiple instances of add course

app/wizards/course_wizard.rb
```diff
@@ -0,0 +1,32 @@
...
+    DfE::Wizard::StepsProcessor::Graph.draw(self) do |graph|
+      graph.root :level
+
+      graph.add_node :level, Steps::Level
+      graph.add_edge from: :level, to: :courses_index
+    end
+  end
+
+  def route_strategy
+    DfE::Wizard::RouteStrategy::ConfigurableRoutes.new(
```
tomas-stefano, line 18: If you go with the state key you might need to use the dynamic routes https://github.com/DFE-Digital/dfe-wizard#route-strategy

OR continue with configurable but then for each step you need to pass the new parameter.

app/wizards/course_wizard.rb
```diff
@@ -0,0 +1,32 @@
...
+      wizard: self,
+      namespace: "publish-provider-recruitment-cycle-course-wizard",
+    ) do |config|
+      config.default_path_arguments = {
+        recruitment_cycle_year: config.wizard.recruitment_cycle_year,
+        provider_code: config.wizard.provider_code,
```
tomas-stefano, line 24: It might work to pass state key here instead of changing to Dynamic.

app/views/publish/course_wizards/show.html.erb
```diff
@@ -0,0 +1,44 @@
...
+      <% end %>
+
+      <%= form.govuk_submit t("course_wizard.steps.level.submit") %>
+    <% end %>
+
+    <p class="govuk-body">
+      <%= govuk_link_to(t("course_wizard.steps.level.cancel"), publish_provider_recruitment_cycle_courses_path(params[:provider_code], params[:recruitment_cycle_year])) %>
+    </p>
+  </div>
+</div>
```
tomas-stefano, line 45: Observation: because the same route points to all the steps, later not on this PR you might need to:

```
render params[:step], if partial_exists?(params[:step])
```

app/wizards/course_wizard.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_accessor :recruitment_cycle_year, :provider_code
+
+  def steps_processor
+    DfE::Wizard::StepsProcessor::Graph.draw(self) do |graph|
+      graph.root :level
+
+      graph.add_node :level, Steps::Level
+      graph.add_edge from: :level, to: :courses_index
+    end
+  end
```
tomas-stefano, line 15: Let's add unit tests for the flow because the flow will be complicated later: https://github.com/DFE-Digital/dfe-wizard#navigation-matchers

tomas-stefano: Something to be based https://github.com/DFE-Digital/dfe-wizard/blob/main/spec/dummy-app-wizard-specs/add_course_wizard_spec.rb

app/controllers/publish/course_wizards_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+        provider.provider_code,
+        provider.recruitment_cycle_year,
+      )
+    end
+
+    def set_wizard
+      state_store = CourseWizard::StateStores::CourseWizardStore.new(
+        repository: DfE::Wizard::Repository::Cache.new(
+          cache: Rails.cache,
+          key: "course_wizard_#{params[:provider_code]}_#{params[:recruitment_cycle_year]}",
```
tomas-stefano: the key needs to be uniq, otherwize if we have multiple things that uses the same key and the key is expired all data expires.

tomas-stefano (approved):
🚀

PR #6068, opened 2026-04-20
fix: Live Course shows the course length before publishing

inulty-dfe:
## Context

Updating course Course Length doesn't publish the value until the course is republished.

Call `course_length_formatted` on the latest published course enrichment passed into the component instead of the course.

## Checklist

### DIFF
--- a/app/models/course_enrichment.rb
+++ b/app/models/course_enrichment.rb
@@ -164,4 +164,8 @@ class CourseEnrichment < ApplicationRecord
   def required_qualifications_needed?
     course&.provider&.recruitment_cycle&.year.to_i < Course::STRUCTURED_REQUIREMENTS_REQUIRED_FROM
   end
+
+  def standard_course_length?
+    course_length.in?(%w[OneYear TwoYears ThreeYears FourYears])
+  end
 end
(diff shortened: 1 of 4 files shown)


spec/system/publish/courses/editing_course_length_spec.rb
```diff
@@ -11,7 +11,7 @@
     then_i_an_error_message
   end
 
-  scenario "I update the course length to a standard length (eg Up to two years)" do
+  scenario "I update the course length to a standard length (eg Up to two years)", :js do
```
inulty-dfe, line 14: Needs poltergeist to open a new window

app/models/course_enrichment.rb
```diff
@@ -164,4 +164,19 @@ def withdraw
...
+      "1 year"
+    when "TwoYears"
+      "Up to 2 years"
+    when "ThreeYears"
+      "3 years"
+    when "FourYears"
+      "4 years"
+    else
+      course_length.to_s
+    end
```
tomas-stefano: I think there is already locales for those? Or we should create one.

tomas-stefano (approved):

PR #6071, opened 2026-04-21
Dual write provider schools

tomas-stefano:
## Summary

As part of the schools / main-site remodel, providers adding schools to their organisation now dual-write to both the legacy and new data models:

- **Legacy**: `Site` (unchanged — existing validations, geocoding, `Site#code` via `Sites::CodeGenerator` all behave as today).
- **New**: `Provider::School` (GIAS-backed join) — populated with the correct `gias_school_id` and `site_code` (`"-"` for the provider main site).

Scope is the **two user-facing flows** the ticket calls out:

1. **Single-school add** — provider or support user searches GIAS, confirms.
2. **Bulk URN add** — user pastes a URN list, confirms.

Each has a Publish (provider-facing) and Support (admin) controller, so four call sites total. The `DataHub::RegisterSchoolImporter` rake task is **out of scope**; anything it writes before switchover is picked up by `DataHub::SchoolsBackfill::Executor`.

## Done-when checklist

- [x] Adding a school to a provider writes to both data models (single-add)
- [x] Bulk URN import writes to both data models
- [x] Old and new write paths live in **separate service objects** (`ProviderSchools::LegacySiteCreator` and `ProviderSchools::Creator`) — the new path never reads legacy Site data
- [x] Code is structured so the old write path can be removed cleanly later (see *Switchover* below)
- [x] Old data model behaves exactly as today
- [x] New data model gets the correct `gias_school_id` and `site_code`
- [x] Failures are transactional — a failure in either write rolls back both, so we never silently create data in only one model
- [x] Specs cover both single-add and bulk URN flows, asserting rows exist in **both** tables
- [x] Switchover steps documented (below)

### Three services under `ProviderSchools::` and one shared utility under `Schools::`

- `ProviderSchools::LegacySiteCreator` — takes a built Site and saves it. One-file legacy island, deleted at switchover.
- `ProviderSchools::Creator` — takes `provider:` and `gias_school_id:`, writes the `Provider::School` row. Wraps in `@provider.with_lock` so concurrent adds to the same provider serialise instead of racing for the same code. Idempotent under `RecordNotUnique`. Uses `find_or_create_by!(gias_school_id:)` — a provider-school relationship is uniquely keyed by `(provider, gias_school)`.
- `Schools::CodeGenerator` — deterministic, provider-scoped site_code generator shared with the upcoming `CourseSchools::` work in the next PR. Reads the union of `provider.sites` + `provider.schools` so codes don't collide on either side during dual-write.

### Controller pattern (all four call sites)

```ruby
  provider_school = ::ProviderSchools::Creator.call(provider:, gias_school_id:)

  site.code = provider_school.site_code   # new → legacy, never the reverse
  ::ProviderSchools::LegacySiteCreator.call(site:)
end
```

The new write runs **first** and owns code generation; the legacy Site is set from the Provider::School's `site_code`. This is what keeps new-path inputs completely free of legacy-Site data — grep `@site.code | site.code` across the four controllers returns only the one *assignment* line above, which is the very line removed at switchover.

### Improvements over the legacy `Sites::CodeGenerator`

| Legacy | New `Schools::CodeGenerator` |
|---|---|
| `.sample` (non-deterministic) | First-available alphabetical (deterministic, predictable for users and tests) |
| Can hand out `"-"` | `"-"` hard-excluded (reserved for the main-site partial unique index) |
| Reads `provider.sites` only | Union of `provider.sites` + `provider.schools` |
| References `Site::POSSIBLE_CODES` etc. | Self-contained constants |
| `"Z"` sentinel in sequential branch | Explicit `"AA"` seed |
| No race protection | Caller (`ProviderSchools::Creator`) holds `@provider.with_lock` |

## Test plan

- [x] `bundle exec rspec spec/services/schools/code_generator_spec.rb` — 13 cases covering: empty provider, first-available pick, gap-filling, union read, numeric fallback, sequential fallback, `"AZ"→"BA"` wrap, `"-"` never emitted, nil/blank tolerance.
- [x] `bundle exec rspec spec/services/provider_schools/` — Creator + LegacySiteCreator unit specs.
- [x] `bundle exec rspec spec/system/publish/providers/schools/ spec/system/support/providers/schools/` — every add-school system spec asserts a matching `Provider::School` row exists with the correct `site_code` and `gias_school_id`.
- [x] Full sweep: `bundle exec rspec spec/system/publish/providers/ spec/system/support/providers/` — 110 examples, 0 failures.
- [x] Grep check: no leakage from legacy `Site` data into `ProviderSchools::Creator` inputs.
- [ ] Manual: add a school via Publish single-search, confirm `Site.last.code == Provider::School.last.site_code` and matching `urn`/`gias_school_id`.
- [ ] Manual: bulk URN paste (3+ URNs), confirm N `Site` rows and N `Provider::School` rows.
- [ ] Manual failure path: stub a `Provider::School` validation to fail, confirm no `Site` row is created either (transactional rollback).

## Switchover (when the new model goes live)

1. Delete `app/services/provider_schools/legacy_site_creator.rb` and its spec.
2. In each of the four schools/check controllers:
   - Remove the `ActiveRecord::Base.transaction` wrapper.
   - Remove the `@site.code = provider_school.site_code` (or `@school_form.fields[:code] = ...`) assignment.
   - Remove the `::ProviderSchools::LegacySiteCreator.call(...)` line.
   - For bulk flows, remove the legacy `schools` helper and the Site-building inside `update`. Rework the check-page view to iterate `gias_schools`.
   - For Support single-add, drop `@school_form.save!` (and the form itself, if unused elsewhere).
3. In `Schools::CodeGenerator`, drop the `provider.sites.pluck(:code)` term from the union.
4. Delete `Sites::CodeGenerator` along with the `Site` model.

`ProviderSchools::Creator` is untouched at switchover — it never read legacy Site data. `Schools::CodeGenerator` survives and continues to serve `CourseSchools::Creator` from the follow-up PR.

## Next PR

A follow-up will mirror this work for **course schools**: adding a school to a course will dual-write `CourseSite` (legacy) and `Course::School` (new). It will introduce `CourseSchools::Creator` and `CourseSchools::LegacyCourseSiteCreator`, reusing `Schools::CodeGenerator` unchanged.

### DIFF
--- a/app/controllers/publish/providers/schools/check_multiple_controller.rb
+++ b/app/controllers/publish/providers/schools/check_multiple_controller.rb
@@ -10,7 +10,16 @@ module Publish
         def show; end
 
         def update
-          schools.each(&:save!)
+          schools.each do |site|
+            ActiveRecord::Base.transaction do
+              ::ProviderSchools::LegacySiteCreator.call(site:)
+              ::ProviderSchools::Creator.call(
+                provider:,
+                gias_school_id: gias_schools_by_urn.fetch(site.urn).id,
+                site_code: site.code,
+              )
+            end
+          end
 
           schools_added_message(schools)
 
@@ -66,6 +75,10 @@ module Publish
           end
         end
         alias_method :load_schools, :schools
+
+        def gias_schools_by_urn
+          @gias_schools_by_urn ||= GiasSchool.where(urn: schools.map(&:urn)).index_by(&:urn)
+        end
 
         def unfound_urns
           @unfound_urns = urn_service[:unfound_urns]
--- a/app/controllers/publish/providers/schools/check_multiple_controller.rb
+++ b/app/controllers/publish/providers/schools/check_multiple_controller.rb
@@ -10,18 +10,19 @@ module Publish
         def show; end
 
         def update
-          schools.each do |site|
+          saved_sites = []
+
+          gias_schools.each do |gias_school|
             ActiveRecord::Base.transaction do
+              provider_school = ::ProviderSchools::Creator.call(provider:, gias_school_id: gias_school.id)
+
+              site = provider.sites.build(gias_school.school_attributes.merge(code: provider_school.site_code))
               ::ProviderSchools::LegacySiteCreator.call(site:)
-              ::ProviderSchools::Creator.call(
-                provider:,
-                gias_school_id: gias_schools_by_urn.fetch(site.urn).id,
-                site_code: site.code,
-              )
+              saved_sites << site
             end
           end
 
-          schools_added_message(schools)
+          schools_added_message(saved_sites)
 
           redirect_to publish_provider_recruitment_cycle_schools_path
         end
@@ -66,19 +67,14 @@ module Publish
           @urn_service ||= ProviderURNIdentificationService.new(provider, urn_form.values || []).call
         end
 
+        def gias_schools
+          @gias_schools ||= GiasSchool.where(urn: urn_service[:new_urns])
+        end
+
         def schools
-          @schools = begin
-            gias_schools = GiasSchool.where(urn: urn_service[:new_urns])
-            gias_schools.map do |gias_school|
-              provider.sites.build(gias_school.school_attributes)
-            end
-          end
+          @schools ||= gias_schools.map { |gias_school| provider.sites.build(gias_school.school_attributes) }
         end
         alias_method :load_schools, :schools
-
-        def gias_schools_by_urn
-          @gias_schools_by_urn ||= GiasSchool.where(urn: schools.map(&:urn)).index_by(&:urn)
-        end
 
         def unfound_urns
           @unfound_urns = urn_service[:unfound_urns]
--- a/app/services/provider_schools/creator.rb
+++ b/app/services/provider_schools/creator.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# Writes a Provider::School row against the new GIAS-backed data model.
+# Swallows RecordNotUnique so a re-run (e.g. race with the schools backfill
+# executor) is a no-op, matching the backfill's ON CONFLICT DO NOTHING.
+module ProviderSchools
+  class Creator
+    include ServicePattern
+
+    def initialize(provider:, gias_school_id:, site_code:)
+      @provider = provider
+      @gias_school_id = gias_school_id
+      @site_code = site_code
+    end
+
+    def call
+      @provider.schools.find_or_create_by!(gias_school_id: @gias_school_id, site_code: @site_code)
+    rescue ActiveRecord::RecordNotUnique
+      @provider.schools.find_by!(gias_school_id: @gias_school_id, site_code: @site_code)
+    end
+  end
+end
--- a/app/services/provider_schools/creator.rb
+++ b/app/services/provider_schools/creator.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
-# Writes a Provider::School row and owns site_code generation so the
-# new model never depends on legacy Site data. Locks the provider row
-# for the duration of the code-pick + insert so two concurrent adds
-# to the same provider can't hand out the same code. Idempotent under
-# RecordNotUnique (race with the backfill or another request).
 module ProviderSchools
+  # Writes a Provider::School row and owns site_code generation so the
+  # new model never depends on legacy Site data. Locks the provider row
+  # for the duration of the code-pick + insert so two concurrent adds
+  # to the same provider can't hand out the same code. Idempotent under
+  # RecordNotUnique (race with the backfill or another request).
   class Creator
     include ServicePattern
 
--- a/app/services/provider_schools/creator.rb
+++ b/app/services/provider_schools/creator.rb
@@ -1,22 +1,27 @@
 # frozen_string_literal: true
 
-# Writes a Provider::School row against the new GIAS-backed data model.
-# Swallows RecordNotUnique so a re-run (e.g. race with the schools backfill
-# executor) is a no-op, matching the backfill's ON CONFLICT DO NOTHING.
+# Writes a Provider::School row and owns site_code generation so the
+# new model never depends on legacy Site data. Locks the provider row
+# for the duration of the code-pick + insert so two concurrent adds
+# to the same provider can't hand out the same code. Idempotent under
+# RecordNotUnique (race with the backfill or another request).
 module ProviderSchools
   class Creator
     include ServicePattern
 
-    def initialize(provider:, gias_school_id:, site_code:)
+    def initialize(provider:, gias_school_id:)
       @provider = provider
       @gias_school_id = gias_school_id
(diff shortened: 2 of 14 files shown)


tomas-stefano:
https://trello.com/c/nLvjovlW/1587-4-school-remodel-add-new-school-to-provider

spec/services/schools/code_generator_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    it "returns 'AA' once all single-char codes are exhausted" do
+      (("A".."Z").to_a + ("0".."9").to_a).each { |c| seed_site(c) }
+
+      expect(described_class.call(provider:)).to eq("AA")
+    end
+
+    it "never returns '-' even when it is the only single-char code missing from used set" do
+      # seed everything BUT "-"; the legacy generator could pick "-" here, we must not
+      (("A".."Z").to_a + ("0".."9").to_a).each { |c| seed_site(c) }
```
inulty-dfe: Can we make this test setup faster? We seem to create 34 records a few times. Could `insert_all` make it better?

app/controllers/publish/providers/schools/check_multiple_controller.rb
```diff
@@ -10,9 +10,19 @@ class CheckMultipleController < ApplicationController
         def show; end
 
         def update
-          schools.each(&:save!)
+          saved_sites = []
 
-          schools_added_message(schools)
+          gias_schools.each do |gias_school|
+            ActiveRecord::Base.transaction do
+              provider_school = ::ProviderSchools::Creator.call(provider:, gias_school_id: gias_school.id)
```
inulty-dfe, line 17: Prefer imperative language which the command pattern tends to use. But it looks like this noun way is endemic in our codebase

https://refactoring.guru/design-patterns/command

```suggestion
              provider_school = ::ProviderSchools::Create.call(provider:, gias_school_id: gias_school.id)
```

app/services/provider_schools/creator.rb
```diff
@@ -0,0 +1,27 @@
...
+  class Creator
+    include ServicePattern
+
+    def initialize(provider:, gias_school_id:)
+      @provider = provider
+      @gias_school_id = gias_school_id
+    end
+
+    def call
+      @provider.with_lock do
```
inulty-dfe, line 18: Does this do anything? The lock is on the provider but the provider isn't being updated.

app/services/schools/code_generator.rb
```diff
@@ -0,0 +1,45 @@
...
+      return "AA" if multi_char_used.empty?
+
+      multi_char_used.max.next
+    end
+
+    def used_codes
+      @used_codes ||= (
+        @provider.sites.pluck(:code) +
+        @provider.schools.pluck(:site_code)
+      ).compact.reject(&:empty?).uniq
```
inulty-dfe:
```suggestion
      ).compact_blank.uniq
```

app/services/schools/code_generator.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+# Generates a site_code for a provider's next school relationship.
+# Deterministic (first-available) to keep behaviour predictable for
+# users and tests. Excludes "-" because it is reserved for the provider
+# main site under the partial unique index on provider_school
+# (site_code = '-'). Reads used codes from both the legacy site table
+# and the new provider_school table so codes don't collide on either
+# side during the dual-write period. Race safety is the caller's
+# responsibility (see ProviderSchools::Creator).
```
inulty-dfe: This should be above the class not the module

app/services/provider_schools/legacy_site_creator.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# Writes the legacy Site row for a provider-school addition. Kept isolated
+# from ProviderSchools::Creator so the old write path can be removed in a
+# single step once the new model is switched on.
```
inulty-dfe: This should be above the class not the module. LSP can read it that way

app/services/provider_schools/creator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# Writes a Provider::School row and owns site_code generation so the
+# new model never depends on legacy Site data. Locks the provider row
+# for the duration of the code-pick + insert so two concurrent adds
+# to the same provider can't hand out the same code. Idempotent under
+# RecordNotUnique (race with the backfill or another request).
```
inulty-dfe: Above class

app/services/schools/code_generator.rb
inulty-dfe: With comments separated
Image

Module
Image

Class
Image

app/services/provider_schools/creator.rb
tomas-stefano: It does. `with_lock` wraps the block in a transaction and issues SELECT … FOR UPDATE on the provider row — the lock is acquired regardless of whether we subsequently update the record. 

That's exactly what we want here: the lock serialises the read-then-write inside the block (pick next free site_code, then insert), so two concurrent "add school" requests for the same provider can't both read the same used_codes set and hand out the same code. 

The unique index would still reject the duplicate, but we'd be relying on RecordNotUnique rescue + retry instead of preventing the collision in the first place.

app/controllers/publish/providers/schools/check_multiple_controller.rb
tomas-stefano: Agreed the imperative form reads better for the command pattern — but as you noted, the *Creator / *Service noun form is the established convention across this codebase (e.g. `RolloverProviderService`, `Partnerships::SchoolCreator`, `Partnerships::SchoolDryRunCreator`, `Subjects::DesignTechnologySubjectCreatorService`, and most of app/services/). 

I'd rather stay consistent with that here and tackle the naming separately if/when we decide to standardise; a one-off Create would stand out more than it would clarify.

Happy to chat about it.

spec/services/schools/code_generator_spec.rb
tomas-stefano: Done

app/services/schools/code_generator.rb
tomas-stefano: Done

app/services/schools/code_generator.rb
tomas-stefano: Done

app/services/provider_schools/creator.rb
tomas-stefano: Done

app/services/provider_schools/legacy_site_creator.rb
tomas-stefano: Done

app/services/provider_schools/creator.rb
inulty-dfe: Ok, I did some reasearch and I think I understand now. :+1:

inulty-dfe (approved):

PR #6072, opened 2026-04-22
Add rate limiting to Sentry message reporting of OneLogin failure

inulty-dfe:
## Context

  Omniauth failures happen regularly due to session expiration and other
  normal issues. We want to silence normal occurances of this. We still
  want to report any high levels of this error so genuine issues don't
  get hidden

  We want to maintain a whitelist of error types that correspond the the
  types that the OneLogin strategy will raise. Because the errors
  correspond the the message of the error we keep a list of valid keys
  to create the redis bucket with.

## Guidance to review

Testing in QA.

I triggered the /auth/failure endpoint with inauthenticity token message 11 times.
Only 2 incitends were reported to Sentry.
This matches our expectations.
Every incident that is above the threshold within the window is reported.

image

image

## Checklist

### DIFF
--- a/app/controllers/find/authentication/sessions_controller.rb
+++ b/app/controllers/find/authentication/sessions_controller.rb
@@ -3,6 +3,37 @@ module Find
     class SessionsController < ApplicationController
       skip_before_action :verify_authenticity_token, only: %i[backchannel_logout]
 
+      KNOWN_ONE_LOGIN_ERROR_TYPES = %w[
+        openid_discovery
+        callback_state_mismatch
+        callback_access_denied
+        callback_invalid_request
+        callback_service_unavailable
+        callback_login_required
+        id_token_request
+        id_token_nonce_mismatch
+        id_token_iss_mismatch
+        id_token_aud_mismatch
+        id_token_iat_mismatch
+        id_token_exp_mismatch
+        id_token_vot_mismatch
+        userinfo_request
+        logout_token_exp_mismatch
+        logout_token_aud_mismatch
+        logout_token_iat_mismatch
+        logout_token_iss_mismatch
+        logout_token_sub_mismatch
+        logout_token_events_claim_mismatch
+        invalid_credentials
+        timeout
+        csrf_detected
+        invalid_response
+        connection_failed
+        invalid_authenticity_token
+      ].freeze
+
+      SAMPLED_ONE_LOGIN_ERROR_TYPES = %w[invalid_authenticity_token other].freeze
+
       def callback
         candidate = Find::CandidateAuthenticator.new(oauth: omniauth).call
 
@@ -31,12 +62,17 @@ module Find
       end
 
       def failure
-        Sentry.capture_message("One Login failure", tags: {
-          error_type: params[:message],
-          provider: params[:provider],
-        }, extra: {
-          session_id: request.session&.id&.public_id,
-        })
+        error_type = normalized_one_login_error_type
+        sampled = SAMPLED_ONE_LOGIN_ERROR_TYPES.include?(error_type)
+
+        if !sampled || ErrorReporting::RateLimiter.report?(key: "one_login:#{error_type}", threshold: 10)
+          Sentry.capture_message("One Login failure", tags: {
+            error_type:,
+            sample_rate: sampled ? 10 : 1,
+          }, extra: {
+            session_id: request.session&.id&.public_id,
+          })
+        end
 
         render "errors/omniauth"
       end
@@ -55,6 +91,11 @@ module Find
       end
 
     private
+
+      def normalized_one_login_error_type
+        candidate = params[:message].to_s.demodulize.underscore.sub(/_error\z/, "")
+        KNOWN_ONE_LOGIN_ERROR_TYPES.include?(candidate) ? candidate : "other"
+      end
 
       def logout_request(token)
         logout_utility.build_request(
--- a/app/controllers/find/authentication/sessions_controller.rb
+++ b/app/controllers/find/authentication/sessions_controller.rb
@@ -3,6 +3,7 @@ module Find
     class SessionsController < ApplicationController
       skip_before_action :verify_authenticity_token, only: %i[backchannel_logout]
 
+      ERROR_REPORTING_THRESHOLD = 10
       KNOWN_ONE_LOGIN_ERROR_TYPES = %w[
         openid_discovery
         callback_state_mismatch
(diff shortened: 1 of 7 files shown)


inulty-dfe:
[OneLogin Sentry errors](https://trello.com/c/B9G81bJr/1535-onelogin-sentry-errors)

app/controllers/find/authentication/sessions_controller.rb
```diff
@@ -31,12 +62,17 @@ def destroy
       end
 
       def failure
-        Sentry.capture_message("One Login failure", tags: {
-          error_type: params[:message],
-          provider: params[:provider],
```
tomas-stefano, line 36: The old Sentry.capture_message tagged provider: params[:provider]; the new version
   replaced tags with error_type + sample_rate and dropped provider. 
The `/failure` route is generic across providers (developer/dfe/one-login), and provider is low-cardinality — there's no cardinality reason to drop it. Looks accidental. Is it accidental or on purpose?

app/controllers/find/authentication/sessions_controller.rb
```diff
@@ -31,12 +62,17 @@ def destroy
...
-        }, extra: {
-          session_id: request.session&.id&.public_id,
-        })
+        error_type = normalized_one_login_error_type
+        sampled = SAMPLED_ONE_LOGIN_ERROR_TYPES.include?(error_type)
+
+        if !sampled || ErrorReporting::RateLimiter.report?(key: "one_login:#{error_type}", threshold: 10)
+          Sentry.capture_message("One Login failure", tags: {
+            error_type:,
+            sample_rate: sampled ? 10 : 1,
```
tomas-stefano: I don't fully understand this line.
`sample_rate: sampled ? 10 : 1` puts the threshold into a tag named "sample rate".

The limiter does not sample 1-in-10: once `≥10` events land in the window, report?

`returns true` for every subsequent event until they age out. 

So the effective rate is 1, not 10. Anyone extrapolating Sentry volume by ×10 will overcount badly. 

Rename the tag (e.g. rate_limited: true) or drop the numeric value.

app/lib/error_reporting/rate_limiter.rb
```diff
@@ -0,0 +1,29 @@
...
+        # Remove all members in a sorted set within the given scores.
+        r.zremrangebyscore(events_key, "-inf", now - window.to_i)
+        # Set a key's time to live in seconds.
+        r.expire(events_key, window.to_i + 60)
+        # Get the number of members in a sorted set.
+        r.zcard(events_key)
+      }.last
+
+      count >= threshold
+    rescue StandardError
```
tomas-stefano: Fail-open floods Sentry during a Redis outage.

 `rescue StandardError → true` means: if Redis is down, every sampled error reports —  the rate limiting disappears exactly when load may be highest. StandardError is also broad enough to swallow a programming error (typo'd method, nil) and silently always-report, hiding the bug. 

Consider rescuing Redis-specific errors only, and  logging the rescue so a broken limiter is observable.

app/controllers/find/authentication/sessions_controller.rb
```diff
@@ -31,12 +62,17 @@ def destroy
...
-        Sentry.capture_message("One Login failure", tags: {
-          error_type: params[:message],
-          provider: params[:provider],
-        }, extra: {
-          session_id: request.session&.id&.public_id,
-        })
+        error_type = normalized_one_login_error_type
+        sampled = SAMPLED_ONE_LOGIN_ERROR_TYPES.include?(error_type)
+
+        if !sampled || ErrorReporting::RateLimiter.report?(key: "one_login:#{error_type}", threshold: 10)
```
tomas-stefano: The first 9 invalid_authenticity_token / other events per hour produce zero Sentry events. 
That's the intention.

But a small-but-real regression in these buckets is invisible. Combined with the allowlist, a brand-new gem error type would land in other and be  suppressed until frequent. Can you confirm this blind spot is intended for these benign categories?

app/controllers/find/authentication/sessions_controller.rb
```diff
@@ -56,6 +92,11 @@ def backchannel_logout_utility
 
     private
 
+      def normalized_one_login_error_type
+        candidate = params[:message].to_s.demodulize.underscore.sub(/_error\z/, "")
```
tomas-stefano, line 97: `.demodulize` implies class-name messages, but I could not identify no producer exists in-repo — 
  sessions_controller.rb:96 + KNOWN list line 32
  OmniAuth's params[:message] values are snake_case symbols (csrf_detected, timeout,
   …) — never Foo::Bar. 

The .demodulize call and the spec value
  "ActionController::InvalidAuthenticityToken" indicate this was observed in production from a gem/middleware, but nothing in this repo references InvalidAuthenticityToken. Add a code comment documenting where that message
  originates so KNOWN_ONE_LOGIN_ERROR_TYPES can be kept in sync with the gem.

Also demodulizing a params might be dangerous without further verification list.

app/controllers/find/authentication/sessions_controller.rb
```diff
@@ -31,12 +62,17 @@ def destroy
...
-          error_type: params[:message],
-          provider: params[:provider],
-        }, extra: {
-          session_id: request.session&.id&.public_id,
-        })
+        error_type = normalized_one_login_error_type
+        sampled = SAMPLED_ONE_LOGIN_ERROR_TYPES.include?(error_type)
+
+        if !sampled || ErrorReporting::RateLimiter.report?(key: "one_login:#{error_type}", threshold: 10)
+          Sentry.capture_message("One Login failure", tags: {
```
tomas-stefano, line 70: Minor: threshold: 10 is hardcoded at the controller call site — consider a constant next to
  SAMPLED_ONE_LOGIN_ERROR_TYPES.

app/controllers/find/authentication/sessions_controller.rb
inulty-dfe: Ah I see what you mean. sample rate is misleading, it should be `threshold`

app/controllers/find/authentication/sessions_controller.rb
inulty-dfe: Fair. We now only threshold invalid_authenticity_token errors

app/controllers/find/authentication/sessions_controller.rb
inulty-dfe: This message comes from rails and it is in the form of ActionController::InvalidAuthenticityToken.

It's safe to demodularize strings that aren't in this format afaik

app/controllers/find/authentication/sessions_controller.rb
inulty-dfe: I've set a const

app/controllers/find/authentication/sessions_controller.rb
inulty-dfe: In production it doesn't actually show anything useful. We aren't using different providers, we only use OneLogin.

tomas-stefano (approved):

PR #6078, opened 2026-04-23
Dual write course schools

tomas-stefano:
## Summary

Continues the schools / main-site remodel. Adding or removing a school on a course now dual-writes to both models:

- **Legacy**: `SiteStatus` row (table `course_site`), written via the existing `Course#add_site!` / `remove_site!` path — unchanged behaviour for live reads.
- **New**: `Course::School` row (table `course_school`), with `site_code` copied from the matching `Provider::School` so codes stay consistent across `site`, `provider_school`, and `course_school` within a provider.

Both writes run inside the existing transaction in `Publish::Schools::UpdateCourseSchoolsService`. If the new-model write fails (e.g. no matching `Provider::School`), the legacy SiteStatus change is rolled back — the two models never silently diverge.

## Divergence from the ticket

The ticket asked for a **new flagged page** that sources its checkbox list from `Provider::School` instead of `Site`. That's dropped. After discussion:

- No feature flag.
- No new view/controller/form/route — same URL, same UX.
- Both **add** and **remove** on the existing flow dual-write.

Rationale: the dual-write is the data-safety bit; a flagged UI swap can ship later (or not) as its own ticket once reads migrate. Keeping the current UI wins us smaller scope and no URL churn.

## Done-when checklist

- [x] Old flow writes **both** old and new data (on add and on remove)
- [x] New course-school write path is implemented clearly (`CourseSchools::Creator` / `Remover`)
- [x] `gias_school_id` populated correctly in the new model
- [x] `site_code` copied from the matching `Provider::School`
- [x] Providers can only attach schools linked to their organisation (enforced by `Creator` — no Provider::School → `RecordNotFound` → rollback)
- [x] Failures handled sensibly (transactional; no silent half-writes)
- [x] Specs cover old-flow dual-write on add, dual-write on remove, site_code copy, and failure rollback
- [x] Switchover note below
- [ ] ~~New flagged view~~ — **descoped, see above**
- [ ] ~~Feature flag~~ — **descoped, see above**

### Services under `CourseSchools::`

Mirrors the `ProviderSchools::` layout from the previous PR:

- `CourseSchools::Creator` — writes `Course::School`. Looks up `Provider::School` for `(course.provider, gias_school)` and **copies** its `site_code`. Idempotent via `find_or_create_by!` + `RecordNotUnique` rescue. Raises `ActiveRecord::RecordNotFound` if no matching Provider::School.
- `CourseSchools::Remover` — `destroy_all` on matching `Course::School` rows. Idempotent (no-op on empty).
- `CourseSchools::LegacySiteStatusCreator` / `::LegacySiteStatusRemover` — one-line wrappers around `Course#add_site!` / `#remove_site!` (which remain private on purpose — called via `send`). Isolated in their own files so the legacy write path is deletable as a unit at switchover.

### Orchestration

`Publish::Schools::UpdateCourseSchoolsService` no longer uses the bulk `course.site_ids = [...]` assignment. It now diffs:

```ruby
to_attach_ids = desired_site_ids - current_site_ids
to_detach_ids = current_site_ids - desired_site_ids
```

and per site, inside the existing transaction:

```ruby
def attach_school(site, gias_schools_by_urn)
  ::CourseSchools::LegacySiteStatusCreator.call(course:, site:)

  gias_school = gias_schools_by_urn[site.urn]
  return unless gias_school                         # URN doesn't resolve → legacy-only, matches backfill skip semantics

  ::CourseSchools::Creator.call(course:, gias_school_id: gias_school.id)
end
```

Mirror `detach_school` for removes (`LegacySiteStatusRemover` + `Remover`).

### Why `site_code` is copied, not generated

A course can only teach at schools the provider has already added, and the provider-school row owns the code. `Schools::CodeGenerator` (from the previous PR) is for provider-level code generation; course-level `site_code` is always derivative. So `Creator` looks up the existing `Provider::School` and reuses its `site_code`.

Prerequisite invariant: every Site a course uses has a matching Provider::School. This holds because:
- Backfill (`DataHub::SchoolsBackfill::Executor`) has populated `provider_school` from historical `site` rows.
- Since the previous PR landed, every new provider school write dual-writes `Provider::School` via `ProviderSchools::Creator`.

If the invariant is ever broken (orphan Site), `Creator` raises `RecordNotFound`, the transaction rolls back, the user sees the error, and we notice rather than silently drift.

### Sites with unresolvable URNs

If a legacy Site has a blank URN or a URN with no `GiasSchool` row, the new-model write is skipped for that site (legacy-only). Matches `DataHub::SchoolsBackfill::Executor#fetch_skipped_course_sites` behaviour — the backfill's process summary remains the system-of-record for skips; we don't double-log.

## Test plan

- [x] `bundle exec rspec spec/services/course_schools/` — 14 unit specs: Creator (copy site_code, idempotent, RecordNotFound on missing Provider::School, RecordNotUnique race), Remover (idempotent, scoped), Legacy wrappers (attach/detach via add_site! / remove_site!).
- [x] `bundle exec rspec spec/services/publish/schools/update_course_schools_service_spec.rb` — 10 cases including dual-write on attach, dual-write on detach, transaction rollback when Creator raises.
- [x] `bundle exec rspec spec/system/publish/courses/editing_course_schools_spec.rb` — attach + detach scenarios both assert `course.schools` mirrors `course.sites` and `site_code` matches the Provider::School.
- [x] `bundle exec rspec spec/system/publish/courses/` — 238 examples, 0 failures (no regressions in neighbouring course flows).
- [x] Rubocop clean on all touched files.
- [ ] Manual: attach a school via the existing edit page, console-check `course.sites.pluck(:urn)` matches `course.schools.map { |s| s.gias_school.urn }` and `course.schools.pluck(:site_code)` matches the Provider::School codes.
- [ ] Manual: detach a school, confirm both the `SiteStatus` and `Course::School` row are gone (or SiteStatus suspended if the course is live, `Course::School` destroyed).
- [ ] Manual failure: temporarily destroy the relevant Provider::School, attempt an attach — transaction rolls back, no SiteStatus and no Course::School, form re-renders with error.

## Switchover note for the later cleanup ticket

> When course reads migrate to `Course::School`, the old write path comes out by:
> 1. Delete `app/services/course_schools/legacy_site_status_creator.rb` / `legacy_site_status_remover.rb` and their specs.
> 2. In `Publish::Schools::UpdateCourseSchoolsService#attach_school` / `#detach_school`, remove the `LegacySiteStatusCreator.call` / `LegacySiteStatusRemover.call` lines.
> 3. Drop `apply_publish_status_to_site_statuses` and anything else that maintains `site_statuses` / `course_site`.
> 4. Delete `Course#add_site!` / `#remove_site!` from the model (currently called via `send` from the legacy services).
>
> `CourseSchools::Creator` and `CourseSchools::Remover` are untouched at switchover — they never wrote legacy data.

## Related

- Depends on data shape from: [previous PR — dual-write provider schools (branch `td/dual-write-provider-schools`)]
- Backfill already in place: `DataHub::SchoolsBackfill::Executor` populates `course_school` from `course_site`.

### DIFF
--- a/app/services/publish/schools/update_course_schools_service.rb
+++ b/app/services/publish/schools/update_course_schools_service.rb
@@ -39,25 +39,41 @@ module Publish
       end
 
       def sync_schools
-        desired_site_ids = Array(params[:site_ids]).compact_blank.map(&:to_i)
-        current_site_ids = course.site_ids
+        return if sites_to_attach_ids.empty? && sites_to_remove_ids.empty?
 
-        to_attach_ids = desired_site_ids - current_site_ids
-        to_detach_ids = current_site_ids - desired_site_ids
-        return if to_attach_ids.empty? && to_detach_ids.empty?
-
-        sites_by_id = Site.where(id: to_attach_ids + to_detach_ids).index_by(&:id)
-        gias_schools_by_urn = GiasSchool
-          .where(urn: sites_by_id.values.map(&:urn).compact_blank)
-          .index_by(&:urn)
-
-        to_attach_ids.each { |id| attach_school(sites_by_id[id], gias_schools_by_urn) }
-        to_detach_ids.each { |id| detach_school(sites_by_id[id], gias_schools_by_urn) }
+        sites_to_attach_ids.each { |id| attach_school(sites_by_id[id]) }
+        sites_to_remove_ids.each { |id| detach_school(sites_by_id[id]) }
 
         course.sites.reload
       end
 
-      def attach_school(site, gias_schools_by_urn)
+      def submitted_site_ids
+        @submitted_site_ids ||= Array(params[:site_ids]).compact_blank.map(&:to_i)
+      end
+
+      def current_site_ids
+        @current_site_ids ||= course.site_ids
+      end
+
+      def sites_to_attach_ids
+        @sites_to_attach_ids ||= submitted_site_ids - current_site_ids
+      end
+
+      def sites_to_remove_ids
+        @sites_to_remove_ids ||= current_site_ids - submitted_site_ids
+      end
+
+      def sites_by_id
+        @sites_by_id ||= Site.where(id: sites_to_attach_ids + sites_to_remove_ids).index_by(&:id)
+      end
+
+      def gias_schools_by_urn
+        @gias_schools_by_urn ||= GiasSchool
+          .where(urn: sites_by_id.values.map(&:urn).compact_blank)
+          .index_by(&:urn)
+      end
+
+      def attach_school(site)
         ::CourseSchools::LegacySiteStatusCreator.call(course:, site:)
 
         gias_school = gias_schools_by_urn[site.urn]
@@ -75,7 +91,7 @@ module Publish
         )
       end
 
-      def detach_school(site, gias_schools_by_urn)
+      def detach_school(site)
         ::CourseSchools::LegacySiteStatusRemover.call(course:, site:)
 
         gias_school = gias_schools_by_urn[site.urn]
--- a/app/services/publish/schools/update_course_schools_service.rb
+++ b/app/services/publish/schools/update_course_schools_service.rb
@@ -22,7 +22,8 @@ module Publish
       def call
         ActiveRecord::Base.transaction do
           assign_attributes_to_course
-          update_site_statuses
+          sync_schools
+          apply_publish_status_to_site_statuses
           course.save!
         end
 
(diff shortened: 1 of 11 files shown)


app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -37,9 +38,53 @@ def assign_attributes_to_course
         course.assign_attributes(params.except(:site_ids))
       end
 
-      def update_site_statuses
-        course.site_ids = params[:site_ids]
+      def sync_schools
+        desired_site_ids = Array(params[:site_ids]).compact_blank.map(&:to_i)
```
langesi: Should we memoize?

Suggest name `submitted_site_ids`?

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -37,9 +38,53 @@ def assign_attributes_to_course
...
+        current_site_ids = course.site_ids
```
langesi: Also suggest this memoized, make the method smaller

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -37,9 +38,53 @@ def assign_attributes_to_course
...
 
+        to_attach_ids = desired_site_ids - current_site_ids
```
langesi: memoize and call it `def sites_to_attach_ids`?

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -37,9 +38,53 @@ def assign_attributes_to_course
...
+        to_detach_ids = current_site_ids - desired_site_ids
```
langesi: memoize and call it `def sites_to_remove_ids`?

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -37,9 +38,53 @@ def assign_attributes_to_course
...
+        return if to_attach_ids.empty? && to_detach_ids.empty?
+
+        sites_by_id = Site.where(id: to_attach_ids + to_detach_ids).index_by(&:id)
```
langesi: Also good candidate for memoization 

```
def sites
	@sites = Site.where(id: sites_to_attach_ids + sites_to_remove_ids)
end
```

langesi (commented):
Looks good so far, here are my comments on the code, will do some manual testing now and approve after.

langesi (approved):
Good work @tomas-stefano !  tested it in the review app by running your rake task and then verifying the changes, and it works.

I’m happy to merge this in. Please just take a look at the comments above.

I made a ticket (Schools data remodel 5.1) to remove course_school entries when a course_provider gets removed.

PR #6094, opened 2026-05-06
Fix untick school on published course

tomas-stefano:
## Context

On a published course, unticking a placement school and pressing **Update** appeared to do nothing — the school stayed attached. Sometimes it raised a 400 with `AASM::InvalidTransition` in logs.

Both symptoms are the same root cause: `Course#remove_site!` was calling `site_status.suspend!` on a row that the database said was already `:suspended` (a stale duplicate `course_site` row for the same `(course, site)` pair). AASM refused the transition, the transaction rolled back, and from the provider's perspective nothing changed.

## How to reproduce on main

1. Find a published course whose `course_site` table has more than one row for the same `(course_id, site_id)` — at least one running and at least one already-suspended (we have these in the wild; QA hit one on `1CS / 2026 / 2DGY`).
2. Visit `/publish/organisations/{provider}/{year}/courses/{code}/schools`.
3. Untick that school and press **Update**.
4. Either nothing changes (silent rollback) or the page renders 400 with the AASM error in the server log.

The bug is **pre-existing** — it has nothing to do with the recent dual-write work. The legacy `course.site_ids = ...` setter ultimately called the same `remove_site!` with the same `suspend!`. Nobody hit the duplicate-row state until QA did.

## Stack trace from QA

```
AASM::InvalidTransition - Event 'suspend' cannot transition from 'suspended'.
  app/models/course.rb:945:in `remove_site!'
  app/services/course_schools/legacy_site_status_remover.rb:18:in `call'
  app/services/publish/schools/update_course_schools_service.rb:95:in `detach_school'
  app/services/publish/schools/update_course_schools_service.rb:45:in `block in sync_schools'
  app/controllers/publish/courses/schools_controller.rb:30:in `update'
```

`Course#remove_site!` did `site_statuses.find_by!(site:)`, which is a SQL query with `LIMIT 1`. With duplicate rows in `course_site` for the same `(course, site)`, the row returned is non-deterministic — it can be the already-suspended one. Then `site_status.suspend!` raises because AASM `:suspend` only permits `:running → :suspended`.

## The fix

`Course#remove_site!` becomes:

```ruby
def remove_site!(site:)
  site_statuses.where(site:).destroy_all
end
```

- Destroys **every** `site_status` for the `(course, site)` pair — handles duplicates by definition.
- Idempotent: `destroy_all` on an empty relation is a no-op.
- No AASM event, so no transition trap.
- Audit history is preserved by the `audited` gem on the destroy event itself.

### What about `:suspended`?

It stays in the model. **`Course#withdraw`** still legitimately uses `:suspended` to mark a withdrawn course's site_statuses (a course-level "no longer live" concept, distinct from "this school was removed"). The AASM `start!` event also still permits `:suspended → :running` for legacy data that's already in `:suspended` state. We're only changing the *remove-a-school-from-course* path.

`:discontinued` is genuinely vestigial — nothing in the codebase ever transitions into it — and is a separate cleanup ticket.

### Why not keep suspend and just guard against duplicates?

We considered three options:

| Option | Verdict |
|---|---|
| `find_by!`-then-skip-if-already-suspended | Still leaves the duplicate-row state untouched; cleans up nothing; another callsite hitting the same data trips again. |
| Iterate, suspend each running, skip suspended | Cleans up partially; ends up with mixed running/suspended rows for the same pair; complicates re-add. |
| **Destroy_all** | Cleans up the whole pair, no AASM, no duplicates left behind, audit log retains history. |

Destroy is the smallest, most defensible change. `:suspended`-from-`remove_site!` was never a feature anyone relied on — it just happened to be the soft-delete strategy.

## Defensive secondary fix included

While debugging I also tightened `Publish::Schools::UpdateCourseSchoolsService#apply_publish_status_to_site_statuses` so it can never flip a `:suspended` or `:discontinued` row back to `:running` via `assign_attributes`:

```ruby
def apply_publish_status_to_site_statuses
  course.site_statuses.reload.new_or_running.each do |site_status|
    site_status.assign_attributes(site_status_attributes)
  end
end
```

`reload` to get DB truth, `new_or_running` so we never touch suspended/discontinued. This isn't strictly required given the destroy switch, but it stops `Course#withdraw`-induced suspended rows from being clobbered if `UpdateCourseSchoolsService` ever runs against a withdrawn-then-edited course.

### New system spec (drives the actual UI)

`spec/system/publish/courses/editing_course_schools_spec.rb` adds three scenarios:

- *I can detach a school from a published course (regression for QA bug)* — published course with two running site_statuses, untick one, assert site is gone, site_status row destroyed, basic-details page reflects the change.
- *Untick all then re-tick one school on a published course with three schools* — exact scenario you walked me through.
- *Duplicate stale site_status rows do not break the untick (AASM regression)* — seeds a duplicate `:suspended` row for the site about to be unticked. **This one fails on main with the same `AASM::InvalidTransition` you saw in logs**, and passes with the fix.

### Specs updated for new destroy semantics

- `spec/models/course_spec.rb` — the "suspends old site_status" assertions become "destroys old site_status".
- `spec/services/publish/schools/update_course_schools_service_spec.rb` — the regression context for unticking on a published course asserts row destruction.
- The "pre-existing suspended site_status doesn't get clobbered by apply" test stays — it now exercises the secondary defensive fix in `apply_publish_status_to_site_statuses`.

### Wider sweep

- `bundle exec rspec spec/services/publish/schools/ spec/services/courses/ spec/models/course/ spec/models/course_spec.rb spec/system/publish/courses/editing_course_schools_spec.rb` — 868 examples, 0 relevant failures (one pre-existing `geokit-rails/adapters/postgis` env load issue unrelated to this branch).
- Rubocop clean on all touched files.

### Manual verification on QA

After deploying this branch:

- [ ] Visit `/publish/organisations/1CS/2026/courses/2DGY/schools`, untick everything except one school, press Update — the page reflects only that one school attached.
- [ ] Run the duplicate-row scenario in rails console first (`SiteStatus.create!(course:, site:, status: :suspended, publish: :unpublished)` for a site that's already attached running) and repeat the untick — should still succeed.

## Risk

- **Low.** The change is to one method on `Course`. The legacy `course.site_ids = ...` flow that some other callers may rely on now destroys instead of suspends. Re-adding a previously-removed school works the same as before — `find_or_initialize_by(site:)` finds nothing, builds new, `start!` to `:running`. Audit log gets a destroy event in place of a suspend update event.
- The only behavioural change a downstream consumer could notice is that the row in `course_site` is gone instead of having `status='S'`. Nothing in the UI or in queries currently filters on `:suspended` from the remove path (verified by grep), and the only non-test reads of `:suspended` are from `Course#withdraw`, which is untouched.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -941,8 +941,12 @@ private
   end
 
   def remove_site!(site:)
-    site_status = site_statuses.find_by!(site:)
-    ucas_status == :new ? site_status.destroy! : site_status.suspend!
+    # Destroy every site_status for this (course, site) pair. The legacy
+    # suspend-then-reuse approach broke when stale duplicate rows existed
+    # (find_by! could return an already-suspended row, AASM raised
+    # InvalidTransition). Audit history is still preserved by the audited
+    # gem on the destroy event.
+    site_statuses.where(site:).destroy_all
   end
 
   def withdraw_latest_enrichment
--- a/app/services/publish/schools/update_course_schools_service.rb
+++ b/app/services/publish/schools/update_course_schools_service.rb
@@ -101,7 +101,12 @@ module Publish
       end
 
       def apply_publish_status_to_site_statuses
-        course.site_statuses.each do |site_status|
+        # Reload + scope to new_or_running so we never touch site_statuses
+        # that sync_schools just suspended or destroyed. Iterating the
+        # cached collection used to flip a freshly-suspended row back to
+        # running, leaving the unticked school still attached on the
+        # rendered page.
+        course.site_statuses.reload.new_or_running.each do |site_status|
           site_status.assign_attributes(site_status_attributes)
         end
       end
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -839,8 +839,8 @@ describe Course do
         expect(second_site_status.reload.status).to eq("running")
       end
 
-      it "sets old site_status to suspended" do
-        expect(first_site_status.reload.status).to eq("suspended")
+      it "destroys the old site_status when its site is removed" do
+        expect(SiteStatus.find_by(id: first_site_status.id)).to be_nil
       end
     end
   end
@@ -1466,15 +1466,19 @@ describe Course do
     context "for running courses" do
       let(:existing_site_status) { create(:site_status, :running, :published, site: existing_site) }
 
-      it "suspends the site when an existing site is removed" do
+      it "destroys the site_status when an existing site is removed" do
         expect { subject.sites = [] }
-          .to change { existing_site_status.reload.status }.from("running").to("suspended")
+          .to change { subject.reload.site_statuses.size }.from(1).to(0)
+        expect(SiteStatus.find_by(id: existing_site_status.id)).to be_nil
       end
 
-      it "adds a new site status and sets it to running when a new site is added" do
-        expect { subject.sites = [new_site] }.to change { subject.reload.site_statuses.size }.from(1).to(2)
+      it "destroys the old site status and creates a running one when sites change" do
+        old_id = existing_site_status.id
+        subject.sites = [new_site]
+        subject.reload
 
-        expect(existing_site_status.reload.status).to eq("suspended")
+        expect(SiteStatus.find_by(id: old_id)).to be_nil
+        expect(subject.site_statuses.size).to eq(1)
         expect(new_site_status.reload.status).to eq("running")
       end
     end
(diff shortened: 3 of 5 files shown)


tomas-stefano:
https://trello.com/c/dUWfJmNC/1662-bug-removing-schools-isnt-working

langesi (commented):
Agree 100% with you approach, looking forward to the sites_status table being retired! 

Will test it manually now, but from a code perspective I'm happy

langesi (approved):
Tested it manually:
1. Bulk Add (>30 and <30)
2. Bulk removal (>30 and <30)
3. Add single schools 
4. Remove single school
5. Remove and add single school

PR #6105, opened 2026-05-07
Store financial incentives by year with display control

langesi:
## Context

Financial incentives are currently modelled as one record per subject. Each year’s incentives replace the previous set, which means we cannot load the next year’s bursary/scholarship data ahead of time without immediately exposing it in Find and the public API.

This change lets us store one financial incentive per subject per recruitment year. It also adds an explicit `displayed` flag so future-year incentives can be loaded hidden, then manually released later by switching which incentive is displayed.

The important behaviour is that existing/current incentives keep showing after the migration, while newly created incentives default to hidden.

- Add `year` and `displayed` to `financial_incentive`.
- Backfill existing financial incentives to:
  - `year: RecruitmentCycle.current.year`
  - `displayed: true`
- Default new financial incentives to:
  - current recruitment cycle year
  - `displayed: false`
- Add database constraints:
  - one financial incentive per `subject_id` and `year`
  - one displayed financial incentive per subject
- Update associations:
  - `subject.financial_incentive` now returns only the displayed incentive
  - `subject.financial_incentive_records` returns all yearly records
  - `course.financial_incentives` continues to expose displayed incentives through subjects
  - `course.financial_incentive_records` exposes all yearly records
- Update `Subjects::FinancialIncentiveCreatorService` so it:
  - creates/updates incentives by `subject + year`
  - resets only the target year
  - does not overwrite other years
  - defaults imported incentives to hidden unless `displayed: true` is passed
- Update test factories and reference data to support yearly/displayed incentives.
- Add coverage for:
  - model validations and defaults
  - subject/displayed association behaviour
  - creator service year isolation
  - CourseIncentive ignoring hidden future incentives
  - public API subject and course serializers ignoring hidden future incentives
  - Find search results and course pages showing only displayed incentives
  - UK-only and non-UK eligible incentive wording

## Guidance to review

Please pay closest attention to:

- The migration:
  - existing incentives should remain visible after deploy
  - new incentives should default to hidden
  - indexes enforce the intended data model
- `Subject#financial_incentive`:
  - this is intentionally kept as the public “displayed incentive” association so existing Find/API code does not need widespread changes
- `Subjects::FinancialIncentiveCreatorService`:
  - it should now be safe to load a future year without affecting the currently displayed year
- The system specs in `spec/system/find/financial_incentives_content_spec.rb`:
  - they check both search results and course show pages
  - they cover wording differences for UK-only vs non-UK eligible incentives

## Checklist


langesi:
https://trello.com/c/CvP2hKYH

langesi (commented):
I chose to store the year as an integer rather than using the integer value of a recruitment cycle.

My thinking is that, where possible, financial incentive settings should not be tightly coupled to recruitment cycles.

The reason for this is that I’m planning to support a UI where incentives can be instantiated, created, or updated for the current year. I’d prefer not to couple more logic to the rollover functionality unless it’s necessary.

Let me know if you disagree with this approach, and if so, why.

db/migrate/20260507120000_add_year_and_displayed_to_financial_incentive.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    FinancialIncentive.reset_column_information
+    FinancialIncentive.update_all(displayed: true)
+
+    add_index :financial_incentive, %i[subject_id year], unique: true
+    add_index :financial_incentive,
+              :subject_id,
+              unique: true,
+              where: "displayed",
+              name: "index_financial_incentive_on_displayed_subject_id"
```
tomas-stefano, line 24: You know what is bothering me?

This migration assumes there is at most one financial_incentive row per subject_id today. It is "fine". 

But there is currently no DB constraint enforcing that — only find_or_initialize_by(subject:) in the import service. If production data has any duplicate rows for a subject, the partial unique index will fail to build and the deploy will be wedged.

Can you do one of the below, please?

  - Add a defensive pre-step that detects duplicates and deletes/aggregates them before adding the index (with raw SQL
  SELECT subject_id, COUNT(*)... HAVING COUNT(*) > 1), or
  - Run a one-off audit on prod and document in the PR/ADR that the data was verified clean before merge.

spec/system/find/financial_incentives_content_spec.rb
```diff
@@ -9,61 +9,82 @@
...
+        then_i_see_the_displayed_financial_incentive_is_for_uk_citizens
+        and_i_do_not_see_the_hidden_future_financial_incentive
+      end
+
+      scenario "search results render the displayed financial incentive when it is available to non-UK citizens" do
+        given_there_is_a_findable_course(bursary_amount: 4000, scholarship_amount: 2000, subject: :physics, non_uk_bursary_eligible: true, non_uk_scholarship_eligible: true)
+        and_there_is_a_hidden_future_financial_incentive
+        when_i_visit_the_find_results_page
+        then_i_see_the_displayed_financial_incentive_is_available_to_non_uk_citizens
+        and_i_do_not_see_the_hidden_future_financial_incentive
```
tomas-stefano, line 35: Maybe I understood wrong.

```ruby
  scenario "the course page renders the displayed financial incentive for UK citizens" do
    ...
    when_i_select_the_course
    then_i_see_the_displayed_financial_incentive_is_for_uk_citizens   # ← wrong page
  end
```
  `then_i_see_the_displayed_financial_incentive_is_for_uk_citizens` asserts on the string `"Scholarships of £2,000 or bursaries of £4,000 are available for UK citizens"`. 
That copy is produced by `CourseIncentive::View#hint_text`
  (app/view_objects/course_incentive/view.rb:57), which is rendered by `Find::Courses::SummaryComponent` and other summary  card components — i.e. the search results page, not the course details page.
  
  On the course details page, the bursary+scholarship copy is `find.financial_support.bursary_and_scholarship_amount` → `"Bursaries of £4,000 and scholarships of £2,000 are available to eligible trainees."` Different string entirely.

  So the two "the course page renders…" scenarios are asserting search-results copy after navigating to the course page. They should fail. Either the scenarios weren't run, or there's a stale-page issue.

tomas-stefano:
The PR has a checklist "I have removed the usage of data-qa attributes in HTML files and updated the corresponding
  tests." 

But the commit only stops the spec from using data-qa selectors. The attributes themselves are still in the
  views:
  - app/components/shared/courses/financial_support/bursary_component/view.html.erb:3,21
  - app/components/shared/courses/financial_support/scholarship_and_bursary_component/view.html.erb:3,21
  - app/views/shared/courses/financial_support/_loan.html.erb:3,14

I tried to pull locally and see on Github. Maybe there is more to push?

app/models/financial_incentive.rb
```diff
@@ -1,5 +1,21 @@
 # frozen_string_literal: true
 
 class FinancialIncentive < ApplicationRecord
+  attribute :displayed, default: false
+  attribute :year, default: -> { FinancialIncentive.current_year }
+
   belongs_to :subject
+
+  scope :displayed, -> { where(displayed: true) }
```
tomas-stefano, line 9: Naming a scope identically to a boolean column shadows nothing today, but it's a subtle footgun.
  FinancialIncentive.displayed is the scope; record.displayed is the column accessor; record.displayed? is the
  predicate. If anyone writes relation.displayed expecting "give me records where displayed is true", they get it — fine
   — but instance.displayed in some serializer context could become ambiguous. Consider renaming the scope to :visible,
  :active, or :on_display. Not blocking.

app/services/subjects/financial_incentive_creator_service.rb
```diff
@@ -213,20 +214,21 @@ def subject_and_financial_incentives
...
-      @financial_incentive.update_all(
+      # Reset the target year's records in case subjects
+      # from an earlier import are no longer being updated.
+      @financial_incentive.for_year(@year).update_all(
         bursary_amount: nil,
         scholarship: nil,
         early_career_payments: nil,
         non_uk_bursary_eligible: false,
         non_uk_scholarship_eligible: false,
+        displayed: @displayed,
```
tomas-stefano: Trying to re-read 3 times to see if I get it.

`update_all` skips validations. 
If `@displayed = true` and another year for any of these subjects already has `displayed = true`, this will violate the partial unique index and raise a `PG::UniqueViolation` instead of a friendly `ActiveRecord::RecordInvalid`.

app/services/subjects/financial_incentive_creator_service.rb
```diff
@@ -213,20 +214,21 @@ def subject_and_financial_incentives
     end
 
     def execute
-      # Reset all records' attributes in case subjects
-      # from previous year are not being updated
-      @financial_incentive.update_all(
+      # Reset the target year's records in case subjects
```
tomas-stefano: The reset loop nulls out bursary_amount, scholarship, etc., for every existing row in the target year, even for
  subjects no longer present in subject_and_financial_incentives[@year]. Those orphaned rows are left as fully-NULL but
  still displayed: `@displayed`. Is that intentional?

db/migrate/20260507120000_add_year_and_displayed_to_financial_incentive.rb
langesi: Thank you, good spot! 

I ran

```
SELECT
  subject_id,
  COUNT(*) AS financial_incentive_count,
  STRING_AGG(id::text, ', ' ORDER BY id) AS financial_incentive_ids
FROM financial_incentive
WHERE subject_id IS NOT NULL
GROUP BY subject_id
HAVING COUNT(*) > 1
ORDER BY financial_incentive_count DESC, subject_id;

```

with 0 results, however I do think that it would be great to have a guard. Especially considering if there is a rollback

spec/system/find/financial_incentives_content_spec.rb
langesi: Thanks! You’re right that the Spec was weak because it did not prove the browser had reached the course page before asserting the incentive copy.

The specific text (`"Scholarships of £2,000 or bursaries of £4,000 are available for UK citizens"`) is not only search-results copy. It is rendered by `CourseIncentive::View#hint_text`, and the course page also renders that through `Find::Courses::SummaryComponent`.

I’ve updated the spec so the course-page scenarios now wait for the course path after clicking through, then assert both pieces of visible course-page content:
- the incentive hint
- the financial support acordion text(`Bursaries of £4,000 and scholarships of £2,000 are available to eligible trainees."`)

I’ve also added negative assertions for the hidden future incentive amounts, so we can avoid targeting implementation details like `data-qa` or CSS classes

app/models/financial_incentive.rb
langesi: Reasonable, I'd argue though that this is quite common in rails. Like if I created a boolean named active, having a scope with the same name in the model would be consider normal, right? It's a common pattern, I'd argue

app/services/subjects/financial_incentive_creator_service.rb
langesi: I didn’t put much thought into this service because my plan was to build a UI that would eventually allow us to edit the financial incentives directly. For now, I kept the existing logic and only updated it to accept years as an argument.

That said, I think this was an oversight. I shouldn’t have made this depend on a future feature that doesn’t exist yet.

I think this service needs to be further updated.

app/services/subjects/financial_incentive_creator_service.rb
langesi: Good spot, didn't give this service enough attention, decided to change the way it works, it'll now instantiate each record and perform `display! `method to hide the old one and display the new one.

tomas-stefano (approved):

PR #6106, opened 2026-05-11
Add Age Range page to new course flow

gms-gs:
## Context

Implement Age Range step for Primary and Secondary courses

- Add age range step to the add course wizard
- Add specs for updated flow
- Add validation to the age range step

## Checklist

### DIFF
--- a/app/views/publish/course_wizards/steps/_age_range.html.erb
+++ b/app/views/publish/course_wizards/steps/_age_range.html.erb
@@ -1,0 +1,33 @@
+<% content_for :page_title, title_with_error_prefix(t("course_wizard.steps.age_range.heading"), current_step.errors.any?) %>
+
+<h1 class="govuk-heading-l">
+  <%= render CaptionText.new(text: t("course_wizard.steps.age_range.caption")) %>
+  <%= t("course_wizard.steps.age_range.heading") %>
+</h1>
+
+<% preset_options = wizard_repository.read[:level] == "primary" ? current_step.primary_age_range_options : current_step.secondary_age_range_options %>
+
+<%= form.govuk_radio_buttons_fieldset :age_range_in_years, legend: { hidden: true, text: t("course_wizard.steps.age_range.heading") } do %>
+  <% preset_options.each_with_index do |value, index| %>
+    <%= form.govuk_radio_button :age_range_in_years, value,
+          label: { text: value.humanize },
+          link_errors: index.zero? %>
+  <% end %>
+
+  <%= form.govuk_radio_divider %>
+
+  <%= form.govuk_radio_button :age_range_in_years, "other",
+        label: { text: t("edit_options.age_range_in_years.other.label") } do %>
+    <p class="govuk-body"><%= t("course_wizard.steps.age_range.other_hint") %></p>
+
+    <%= form.govuk_text_field :course_age_range_in_years_other_from,
+          label: { text: t("course_wizard.steps.age_range.other_from_label") },
+          class: "govuk-input govuk-input--width-2" %>
+
+    <%= form.govuk_text_field :course_age_range_in_years_other_to,
+          label: { text: t("course_wizard.steps.age_range.other_to_label") },
+          class: "govuk-input govuk-input--width-2" %>
+  <% end %>
+<% end %>
+
+<%= form.govuk_submit t("course_wizard.steps.age_range.submit") %>
--- a/app/views/publish/course_wizards/steps/_age_range.html.erb
+++ b/app/views/publish/course_wizards/steps/_age_range.html.erb
@@ -5,10 +5,8 @@
   <%= t("course_wizard.steps.age_range.heading") %>
 </h1>
 
-<% preset_options = wizard_repository.read[:level] == "primary" ? current_step.primary_age_range_options : current_step.secondary_age_range_options %>
-
 <%= form.govuk_radio_buttons_fieldset :age_range_in_years, legend: { hidden: true, text: t("course_wizard.steps.age_range.heading") } do %>
-  <% preset_options.each_with_index do |value, index| %>
+  <% current_step.preset_options.each_with_index do |value, index| %>
     <%= form.govuk_radio_button :age_range_in_years, value,
           label: { text: value.humanize },
           link_errors: index.zero? %>
--- a/app/wizards/course_wizard/repositories/course.rb
+++ b/app/wizards/course_wizard/repositories/course.rb
@@ -3,9 +3,7 @@
 class CourseWizard
   module Repositories
     class Course < DfE::Wizard::Repository::Cache
-      CACHE_EXPIRY = 24.hours
-
-      def initialize(provider_code:, recruitment_cycle_year:, state_key:, cache: Rails.cache, expires_in: CACHE_EXPIRY)
+      def initialize(provider_code:, recruitment_cycle_year:, state_key:, expires_in:, cache: Rails.cache)
         super(
           cache:,
           key: self.class.cache_key(provider_code:, recruitment_cycle_year:, state_key:),
--- a/app/wizards/course_wizard/repositories/course.rb
+++ b/app/wizards/course_wizard/repositories/course.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class CourseWizard
+  module Repositories
+    class Course < DfE::Wizard::Repository::Cache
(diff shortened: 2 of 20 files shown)


gms-gs:
[Implement Age Range step for Primary and Secondary courses](https://trello.com/c/6xlfSR2b/1663-implement-age-range-step-for-primary-and-secondary-courses)

config/locales/en/publish/course_wizard.yml
```diff
@@ -40,3 +40,33 @@ en:
         errors:
           master_subject_id:
             blank: Select a subject
+            same_as_subordinate_subject_id: The second subject must be different to the first subject
+      age_range:
+        caption: Add course
+        heading: Age range
+        other_hint: "Enter an age range in years, for example: 5 to 11. The course must cover 4 or more school years."
+        other_from_label: From
```
tomas-stefano, line 48: Maybe I understood wrong but "Enter an age in From" reads awkwardly. Suggest "Enter a from age" or "Enter an
  age in the 'From' field" — match whatever DfE content style guide says. Same for the To equivalents.

app/wizards/course_wizard/steps/secondary_subjects.rb
```diff
@@ -11,13 +11,21 @@ class SecondarySubjects
...
         SubjectsCache.new.secondary_subjects.map { |subject| [subject.subject_name, subject.id] }
       end
 
       def self.permitted_params
         %i[master_subject_id subordinate_subject_id]
       end
+
+      def validate_master_subject_id_is_not_same_as_subordinate_subject_id?
+        if master_subject_id == subordinate_subject_id
+          errors.add(:master_subject_id, I18n.t("course_wizard.steps.secondary_subjects.errors.master_subject_id.same_as_subordinate_subject_id"))
```
tomas-stefano: When the user submits the secondary subjects form without selecting anything, master_subject_id == subordinate_subject_id is nil == nil → true → they get both "Select a subject" and "The second subject must be different to the first  subject". Confusing error wall.

Fix: bail out when either side is blank, e.g.  return if master_subject_id.blank? || subordinate_subject_id.blank?

While you're here, I'd drop the trailing `?` from the method name — validate callbacks push errors, they don't return booleans, so the predicate convention is misleading.

app/wizards/course_wizard/steps/primary_subjects.rb
```diff
@@ -5,17 +5,17 @@ module Steps
     class PrimarySubjects
       include DfE::Wizard::Step
 
-      attribute :master_subject_id, :string
+      attribute :primary_master_subject_id, :string
```
tomas-stefano, line 8: We need to talk about master subject id...

  Asymmetric attribute naming —primary_master_subject_id vs master_subject_id

  PrimarySubjects was renamed master_subject_id → primary_master_subject_id (with
  corresponding view and locale rename), but SecondarySubjects still uses
  master_subject_id. The repository spec
  (spec/wizards/add_course_wizard/repositories/course_spec.rb:18-23) writes both
  keys to the same state store.

  Two concrete consequences:
  1. After a user navigates level → primary → back → level → secondary → ..., the  state store will contain both primary_master_subject_id (from the primary visit)  and master_subject_id (from the secondary visit). Whatever consumes this later to
  build the actual Course model needs to know which one to pick. Worth grepping the  eventual consumer.
  2. It's a naming asymmetry for the same domain concept. Pick a convention — either primary_master_subject_id/secondary_master_subject_id for both, or  master_subject_id for both — and stick to it.

spec/wizards/add_course_wizard/wizard/previous_step_spec.rb
```diff
@@ -32,4 +32,29 @@
...
+
+  context "from age range with secondary level" do
+    let(:current_step) { :age_range }
+
+    before do
+      state_store.write(master_subject_id: "123", subordinate_subject_id: "456")
+    end
+
+    it "goes back to secondary subjects" do
+      expect(wizard).to have_previous_step(:secondary_subjects)
```
tomas-stefano, line 72: set level: "secondary" in state, only master_subject_id and  subordinate_subject_id. The implicit "which attributes are present" → "which level you came from" inference is fragile. Could you add a regression test for the case   where both primary_master_subject_id and master_subject_id are present in the  cache?

app/wizards/course_wizard/repositories/course.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class CourseWizard
+  module Repositories
+    class Course < DfE::Wizard::Repository::Cache
+      CACHE_EXPIRY = 24.hours
```
tomas-stefano: app/wizards/course_wizard/repositories/course.rb:6 — CACHE_EXPIRY defined in two
  places - Both files now define CACHE_EXPIRY = 24.hours. The controller still passes
  expires_in: CACHE_EXPIRY explicitly, overriding the repository's default with an identical value. Single source of truth, please:
  - Either drop the controller constant and let the repository default apply, or
  - Use CourseWizard::Repositories::Course::CACHE_EXPIRY in the controller.

app/views/publish/course_wizards/steps/_age_range.html.erb
```diff
@@ -0,0 +1,33 @@
+<% content_for :page_title, title_with_error_prefix(t("course_wizard.steps.age_range.heading"), current_step.errors.any?) %>
+
+<h1 class="govuk-heading-l">
+  <%= render CaptionText.new(text: t("course_wizard.steps.age_range.caption")) %>
+  <%= t("course_wizard.steps.age_range.heading") %>
+</h1>
+
+<% preset_options = wizard_repository.read[:level] == "primary" ? current_step.primary_age_range_options : current_step.secondary_age_range_options %>
```
tomas-stefano: <% preset_options = wizard_repository.read[:level] == "primary" ?
  current_step.primary_age_range_options : current_step.secondary_age_range_options
  %>

  This is exactly the reason helper_method :wizard_repository got added in the
  controller. 

Reaching into the repository cache from the view is a layer-skip that bypasses the step object — and it's the only step that does so. Better:
  - Expose preset_options (or a primary?/secondary? predicate) on Steps::AgeRange,  drive it off the state store inside the step. The view becomes
  current_step.preset_options.each_with_index ....
  - Then the helper_method :wizard_repository change can go away.

app/wizards/course_wizard/steps/age_range.rb
```diff
@@ -0,0 +1,68 @@
...
+      include DfE::Wizard::Step
+
+      attribute :age_range_in_years, :string
+      attribute :course_age_range_in_years_other_from, :string
+      attribute :course_age_range_in_years_other_to, :string
+
+      validates :age_range_in_years, presence: { message: I18n.t("course_wizard.steps.age_range.errors.age_range_in_years.blank") }
+
+      with_options if: :age_range_other? do
+        validates :course_age_range_in_years_other_from, numericality: {
```
tomas-stefano, line 17: app/wizards/course_wizard/steps/age_range.rb:15-26 validates:
  - from: 0–46
  - to: 4–50

  app/services/courses/validate_custom_age_range_service.rb:6-9 validates:
  - from: 3–15
  - to: 7–19

  A user entering from=20, to=30 passes wizard numericality (both within outer
  bounds) but then hits "Age range must cover 4 or more school years" from the
  service. The outer numericality bounds aren't doing anything useful except
  catching truly absurd numbers. Either:
  - Tighten the wizard numericality to mirror the service's MIN/MAX constants (and
  let those constants be the single source of truth), or
  - Remove the bounds and let the service own range validation.

  Mixing two sets of rules invites drift; if the policy changes, somebody will update one but not the other.

config/locales/en/publish/course_wizard.yml
gms-gs: These error messages aren't in the old version of the add course flow. I just added them as it seemed odd we didn't have them in the first place. I spoke with Kellie and we will sort this out properly with Rose after bug party. We have a ticket for this so we can leave this for now

app/wizards/course_wizard/steps/secondary_subjects.rb
gms-gs: @tomas-stefano If the user submits no master_subject_id or subordinate_subject_id the error message is below "Select a subject", This is due to the ordering of the validations :

Screenshot 2026-05-13 at 11 40 14

tomas-stefano (approved):

PR #6112, opened 2026-05-15
Bump the RAM allocation to review app web pods to 1Gi

inulty-dfe:
## Context

  Our latest Prototype branch is restarting with OOM. We may be able to
  revert this if we find optimisations in that branch

Increase the max memory for review apps to 1Gi

## Checklist


gms-gs (approved):

langesi (dismissed):
Just do the change in the prototype branch. That should be enough as the deployment crispy uses the terraform config from that branch only and we won't upscale all new review branch apps

PR #6123, opened 2026-05-21
Implement Funding type step

gms-gs:
## Context

Add funding type step

- Add step
- Add view
- Add tests

## Guidance to review

Test that when going from qualifications to funding type this goes to funding type page **UNLESS a user selects `Teacher degree apprenticeship (TDA) with QTS` they should be taken to the course index page, as in the current app (check qa) it takes them to select `Employing schools` page and skips funding type** Which will be implemented when I get to that step.

## Checklist


gms-gs:
[Implement Funding type step](https://trello.com/c/BwOTtq4h/1685-implement-funding-type-step)

inulty-dfe (approved):

PR #6129, opened 2026-05-26
Update API to use schools remodel

inulty-dfe:
## Context

The aim of this PR is to update the API endpoints, services and serializers to return course and provider locations data to be sourced from the Course::School and Provider::School models instead of the Site and SiteStatus models in the 2026 cycle and beyond.

There are 2 locations endpoints in the TTAPI.

1. Provider locations ([API docs](https://api.publish-teacher-training-courses.service.gov.uk/docs/api-reference.html#recruitment_cycles-year-providers-provider_code-locations-get))
2. Course locations ([API docs](https://api.publish-teacher-training-courses.service.gov.uk/docs/api-reference.html#recruitment_cycles-year-providers-provider_code-courses-course_code-locations-get))

The course locations endpoint accepts an include to add a (deprecated) `location_status`. The location status is deprecated because all locations attached to a course are effectively published and with vacancies.

Each of these objects can include SiteStatus.

TTAPI locations endpoints for provider and course schools return records same as when they used the sites table.

 - Switch location endpoints and locations serialization to use Schools instead of Sites
 - Append " (Main site)" to the location name of all locations with site_code of "-" (main sites)
 - Transform `region_code` to the same format in the API as it was in the Sites implementation

### Bug fix

Our courses controller does not permit all necessary params for filtering courses as the API docs advertises. This PR includes a bug fix for this.

### Api provider schools

This course has been updated to allow no schools to be attached.

```
curl -H 'content-type: application/json'  https://publish-review-6129-api.test.teacherservices.cloud/api/public/v1/recruitment_cycles/2027/providers/B91/courses/T891/locations | jq '.'
{
  "data": [],
  "meta": {
    "has_course_schools": false
  },
  "jsonapi": {
    "version": "1.0"
  }
}
```
## Checklist


app/controllers/api/public/v1/providers/courses/locations_controller.rb
```diff
@@ -34,6 +52,10 @@ def provider
             def include_param
               params.fetch(:include, "")
             end
+
+            def schools_remodelled
```
langesi, line 79: Should we move this method into the Application controller as it is repeated a few times?

config/settings.yml
```diff
@@ -124,4 +124,6 @@ cookies:
   user_session:
     name: user_session
 
+schools_remodel_cycle_year: 2025
```
langesi: I know I'm flip flopping here, but as we are so close to the new cycle, I think we might as well just enable this for 2027 and show the old cycle for now.

This will also be safer as we can see during rollover if the API fails, which would decrease our risk of deploying to a live cycle

app/controllers/api/public/v1/providers/courses/locations_controller.rb
```diff
@@ -9,22 +9,40 @@ class LocationsController < API::Public::V1::ApplicationController
...
+                     expose: exposures,
                      class: API::Public::V1::SerializerService.call
             end
 
           private
 
             def locations
-              @locations ||= course&.sites
+              @locations ||= if schools_remodelled
+                               course&.schools
```
langesi: Course location needs to display provider_school if no course_school is available. 

Suggestion:

course.schools.presence || provider.schools

app/controllers/api/public/v1/providers/courses/locations_controller.rb
```diff
@@ -9,22 +9,40 @@ class LocationsController < API::Public::V1::ApplicationController
...
             end
 
             def location_statuses
               @location_statuses ||= course&.site_statuses
             end
 
             def course
-              @course ||= provider.courses.includes(site_statuses: [:site]).find_by(course_code: params[:course_code])
+              @course ||= if schools_remodelled
+                            provider.courses.includes(schools: %i[gias_school provider_school]).find_by(course_code: params[:course_code])
```
langesi, line 65: We might want to change the include statement if no school is attached to the course

app/controllers/api/public/v1/providers/courses/locations_controller.rb
langesi: Also do we append the 

"has_course_schools": boolean

to the location response.

langesi (commented):
Overall LGTM, just have a few comments. Let me know what you think

app/serializers/api/public/v1/school_location_status.rb
```diff
@@ -0,0 +1,31 @@
...
+    module V1
+      class SchoolLocationStatus
+        def initialize(course_school)
+          @course_school = course_school
+        end
+
+        delegate :id, to: :@course_school
+
+        def status
+          "running"
```
gms-gs: Should this be a in a constant?

app/serializers/api/public/v1/school_location_status.rb
```diff
@@ -0,0 +1,31 @@
...
+        end
+
+        def publish
+          "published"
```
gms-gs: Should this be a in a constant?

app/serializers/api/public/v1/school_location_status.rb
```diff
@@ -0,0 +1,31 @@
...
+        end
+
+        def vac_status
+          SiteStatus.default_vac_status_given(study_mode: @course_school.course.study_mode).to_s
+        end
+
+        def has_vacancies?
+          true
```
gms-gs, line 32: has_vacancies? always returns true. Should this be derived from the course school, vacancy status, or study mode? As written, every school location will be advertised as having vacancies regardless of its actual state. Maybe I am misunderstanding the purpose of this method

inulty-dfe: The concept of vacancies is deprecated, if a course is returned in the API it has vacancies by definition of it being publicly presented as a location on the course.

app/controllers/api/public/v1/providers/courses/locations_controller.rb
inulty-dfe: Maybe, but we don't expect it to be here long enough for it to matter IMO

app/controllers/api/public/v1/providers/courses_controller.rb
```diff
@@ -21,8 +21,9 @@ def show
         private
```
langesi, line 21: Can we add additional meta data to this endpoint indicating if the location is from course_school or provider_school? I think that's not in there yet, right?

So additional key value pair in the metadata:

"has_course_schools": boolean

This was a request from Apply if we could have an identifier in there.

langesi (approved):
Just one thing I'd like you to add then LGTM, feel free to merge after

app/controllers/api/public/v1/providers/courses_controller.rb
inulty-dfe: Yeah I just remembered I forgot this.

inulty-dfe:
[7. School remodel - Update API school locations for the school remodel outcome](https://trello.com/c/zbvrd9Jr/1667-7-school-remodel-update-api-school-locations-for-the-school-remodel-outcome)

PR #6132, opened 2026-05-27
Enable Airbyte in DfE::Analytics

asatwal:
## Context

Enable Airbyte in the Development and Review environments 

Enavle Airbyte in DfE::Analytics and generate the airbyte configuration

## Guidance to review

See documentation on Airbyte in DfE::Analytics:
https://github.com/DFE-Digital/dfe-analytics/blob/main/docs/airbyte.md

### DIFF
--- a/.github/workflows/delete-review-app.yml
+++ b/.github/workflows/delete-review-app.yml
@@ -49,6 +49,13 @@ jobs:
           echo "USE_DB_SETUP_COMMAND=false" >> $GITHUB_ENV
           echo "TF_STATE_FILE=pr-${PR_NUMBER}.tfstate" >> $GITHUB_ENV
 
+      - name: Set Airbyte
+        if: (contains(github.event.pull_request.labels.*.name, 'airbyte'))
+        run: |
+          echo "TF_VAR_pg_airbyte_enabled=true" >> $GITHUB_ENV
+          echo "TF_VAR_airbyte_enabled=true" >> $GITHUB_ENV
+          echo "TF_VAR_connection_status=active" >> $GITHUB_ENV
+
       - name: Delete Review App
         id: delete-review-app
         uses: DFE-Digital/github-actions/delete-review-app@master
@@ -63,3 +70,5 @@ jobs:
           resource-group-name: "${{ env.AZURE_RESOURCE_PREFIX }}-${{env.SERVICE_SHORT}}-${{ env.CONFIG_SHORT }}-rg"
           storage-account-name: "${{ env.AZURE_RESOURCE_PREFIX }}${{env.SERVICE_SHORT}}tfstate${{ env.CONFIG_SHORT }}sa"
           tf-state-file: "${{ env.TF_STATE_FILE }}"
+          gcp-project-id: rugged-abacus-218110
+          gcp-wip: projects/712009772377/locations/global/workloadIdentityPools/publish-teacher-training/providers/publish-teacher-training
--- a/terraform/aks/airbyte.tf
+++ b/terraform/aks/airbyte.tf
@@ -1,0 +1,92 @@
+provider "google" {
+  project = "rugged-abacus-218110"
+}
+
+# data "azurerm_key_vault_secret" "airbyte_client_id" {
+#   count = var.airbyte_enabled ? 1 : 0
+
+#   key_vault_id = data.azurerm_key_vault.key_vault.id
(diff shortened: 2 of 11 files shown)


hopep-mobilise (approved):
Rick has deployed and tested the review app. We will need this merged ready for Thursday.

terraform/aks/airbyte.tf
```diff
@@ -0,0 +1,92 @@
...
+  azure_resource_prefix = var.azure_resource_prefix
+  service_short         = var.service_short
+  service_name          = var.service_name
+  docker_image          = var.docker_image
+  postgres_version      = var.postgres_version
+  postgres_url          = module.postgres.url
+
+  host_name          = module.postgres.host
+  database_name      = module.postgres.name
+  workspace_id       = var.airbyte_enabled ? module.secrets.map.AIRBYTE-WORKSPACE-ID : null
```
langesi, line 56: Are hypen keys valid in terraform? Just checking that there are no parse errors

terraform/aks/airbyte.tf
```diff
@@ -0,0 +1,92 @@
...
+  client_id          = var.airbyte_enabled ? module.secrets.map.AIRBYTE-CLIENT-ID : null
+  client_secret      = var.airbyte_enabled ? module.secrets.map.AIRBYTE-CLIENT-SECRET : null
+  repl_password      = var.airbyte_enabled ? module.secrets.map.AIRBYTE-REPLICATION-PASSWORD : null
+  server_url         = "https://airbyte-${var.namespace}.${module.cluster_data.ingress_domain}"
+  connection_status  = var.connection_status
+  connection_streams = local.connection_streams
+
+  cluster           = var.cluster
+  namespace         = var.namespace
+  gcp_taxonomy_id   = "69524444121704657"
```
langesi, line 66: Checking if these values should be hardcoded. Our repo is open source

.github/workflows/delete-review-app.yml
```diff
@@ -49,6 +49,13 @@ jobs:
           echo "USE_DB_SETUP_COMMAND=false" >> $GITHUB_ENV
           echo "TF_STATE_FILE=pr-${PR_NUMBER}.tfstate" >> $GITHUB_ENV
 
+      - name: Set Airbyte
```
langesi, line 52: Nice to see that we can label Airbyte to deploy it. I however don't see what happens if the label is removed though. So what happens if someone removes the label before closing the PR?

langesi (approved):
Hope my comments make sense, just have a look at them, if they are all good, feel free to merge

terraform/aks/airbyte.tf
asatwal: Good point @langesi - Ideally these shouldn't be hardcoded, altough I don't think they give too much away.

terraform/aks/airbyte.tf
asatwal: In terraform hyphens are valid as string map keys. but not valid in terraform identifiers, such as variable names, local names, resource names, etc. Well that's my understanding anyway. Could you confirm please @hopep-mobilise ?

hopep-mobilise: Most rules in Terraform are naming conventions for readability and standardisation. However, hyphens are not allowed in resource names and data source names. Since AIRBYTE-WORKSPACE-ID is reference to an external secret there's no problem.

terraform/aks/airbyte.tf
hopep-mobilise: You are affectively giving some information away and I do think it would be better to have all these as hidden variables or secrets. I brought this up separately a few weeks ago and have created a card on DevOps to come up with a strategy and timeline to hide all such values across all services. For now it's not a big concern.

.github/workflows/delete-review-app.yml
hopep-mobilise: It's not clever enough to handle the label being removed. An improvement for the future perhaps.

asatwal: Not too sure what happens myself. It might even stop a cleanup of Airbyte when the PR is closed. @hopep-mobilise ?

hopep-mobilise: Yes, it will fail to clear up the AirByte infrastructure if the label is remobved.

PR #6139, opened 2026-05-29
Refactor search location value object

tomas-stefano:
## Summary

The question *"does this search have a usable location?"* was being answered in four different files, each with a slightly different definition. This consolidates it into one place — a small `Courses::SearchLocation` value object — and updates the four sites to ask it the same way.

## Why this matters

The recent `PG::UndefinedColumn` crash on the Find results page (hot-fixed in `td/fix-pg-undefined-column-distance-ordering`) was the loudest symptom of this. That fix patched one of the four places. The others kept producing quieter wrong-UI bugs:

- A confusing "Sort by: Course (a to z)" chip would appear after searching a non-UK location.
- The "Sort by" chip would not show up when a user picked a non-default sort on a geocoded search.

## What's in the box

**New:** `Courses::SearchLocation` — small value object. Its public method is `#located?`, which returns true only when both `latitude` and `longitude` are present.

**Updated** — the four sites that previously implemented the rule themselves now delegate to the value object:
- `Courses::OrderingStrategy`
- `Courses::SearchForm#ordering_filter_count`
- `Find::SearchParamDefaults`
- `Courses::ActiveFilters::HashExtractor`

**Bonus fix:** `Courses::ActiveFilters::Extractor` was being handed `search_params` with coordinates stripped (so they wouldn't render as filter chips), but it then used those stripped params to compute which order counted as "default" — so it could no longer tell a geocoded search apart. Pointed it at the form's full `search_params` for the defaults check only.

## Behaviour change to call out

Previously, a search with `short_address: "Manchester"` but no resolved coordinates (think: a stored email alert whose lat/lng columns were never populated) was treated as if `"distance"` were the natural default order. After this PR, only resolved coordinates count. The filter chip extractor is now honest: it won't claim `"distance"` is the default for a search the query can't actually run by distance.

A dedicated unit spec in `search_param_defaults_spec.rb` locks this in.

## Tests

- New unit specs for `Courses::SearchLocation`.
- New unit specs pinning down the `SearchParamDefaults` behaviour change.
- 7 new system specs in `spec/system/find/results/un_geocoded_location_spec.rb` covering the production trigger (`Atlantis` geocoding to ZERO_RESULTS) and six variations — page renders, no `PG::UndefinedColumn`, no misleading "Sort by" chip.
- 6 new system specs in `spec/system/find/results/filter_chip_consistency_spec.rb` asserting the form and the chip extractor agree on default-vs-applied across geocoded, un-geocoded, and no-location states.

To confirm the new system specs actually exercise the rule, I temporarily broke `SearchLocation#located?` to return the wrong answer and ran them — 8 of 13 went red. Restored.

## Class diagram — `Courses::SearchLocation` + `Courses::OrderingStrategy` and their callers

```mermaid
classDiagram
    direction LR

    class SearchLocation {
      +text: String
      +latitude
      +longitude
      +formatted_address: String
      +short_address: String
      +located?() Boolean
      +blank?() Boolean
      +label() String
      <<value object>>
      $from_params(params) SearchLocation
    }

    class OrderingStrategy {
      DEFAULT_ORDER = "course_name_ascending"
      LOCATED_ORDER = "distance"
      +call() String
      $default_for(search_location) String
      <<strategy>>
    }

    class SearchForm {
      +order() String
      +ordering_filter_count() Integer
      -search_location() SearchLocation
    }

    class SearchParamDefaults {
      +default_value?(key, value) Boolean
      -defaults() Hash
    }

    class HashExtractor {
      +call() Array~ActiveFilter~
      -default_order() String
      -search_location() SearchLocation
    }

    class Extractor {
      +call() Array~ActiveFilter~
      -param_defaults: SearchParamDefaults
    }

    SearchForm        --> SearchLocation   : builds (memoised)
    HashExtractor     --> SearchLocation   : builds from @attrs
    SearchParamDefaults ..> SearchLocation : builds inside defaults proc

    SearchForm           --> OrderingStrategy : .new(...).call  (resolves order)
    SearchForm           --> OrderingStrategy : .default_for     (ordering_filter_count)
    SearchParamDefaults  --> OrderingStrategy : .default_for     (defaults proc)
    HashExtractor        --> OrderingStrategy : .default_for     (default_order)

    OrderingStrategy --> SearchLocation : asks .located?
    Extractor       --> SearchParamDefaults : uses for default checks
    SearchForm      --> Extractor : passes form into Extractor.new
```

---

## Sequence diagram — "user searches with un-geocoded text" (the production trigger)

```mermaid
sequenceDiagram
    autonumber
    actor U as User
    participant C as Find::ResultsController
    participant F as SearchForm
    participant L as SearchLocation
    participant O as OrderingStrategy
    participant Q as Courses::Query
    participant E as ActiveFilters::Extractor
    participant D as SearchParamDefaults

    U->>C: GET /results?location=Atlantis
    C->>F: SearchForm.new(params + geocode result)
    Note over C,F: geocode returned ZERO_RESULTS → no latitude/longitude

    F->>L: build SearchLocation (text, no coords)
    F->>O: OrderingStrategy.new(search_location:, ...).call
    O->>L: located?
    L-->>O: false
    O-->>F: "course_name_ascending"  (DEFAULT_ORDER)

    F->>Q: Courses::Query.call(params: ...order=course_name_ascending)
    Q-->>F: results sorted by course name (no PG::UndefinedColumn)

    F->>E: build active_filters (passes self)
    E->>D: SearchParamDefaults.new(form.search_params)
    Note over D: defaults proc uses<br/>OrderingStrategy.default_for(SearchLocation.from_params(params))
    D->>L: built inside proc → located?
    L-->>D: false
    D-->>E: default order = "course_name_ascending"
    Note over E: order = default → suppress "Sort by" chip

    E-->>F: chips (no order chip)
    F-->>C: render
    C-->>U: 200 OK, no misleading chip
```

---

## What changed structurally

- One **predicate** — `SearchLocation#located?` — replaces four different definitions of "is this a location-based search?".
- One **mapping** — `OrderingStrategy.default_for(search_location)` — replaces four duplicates of `located? ? "distance" : "course_name_ascending"`.
- `Extractor` now sources defaults from `search_form.search_params` (full) so it can see coords that were intentionally stripped from chip iteration.
- `Courses::Query` / `SavedCourses::Query` still implement the same default-order rule via internal scope state — already correct, deliberately out of scope for this PR.

## Test plan

- [x] `bundle exec rspec spec/services/courses/ spec/services/find/ spec/services/saved_courses/ spec/forms/courses/` (677 examples)
- [x] `bundle exec rspec spec/system/find/results/un_geocoded_location_spec.rb spec/system/find/results/filter_chip_consistency_spec.rb` (13 examples)
- [x] Smoke check by breaking `SearchLocation#located?` and confirming the new system specs go red.

### DIFF
--- a/app/forms/courses/search_form.rb
+++ b/app/forms/courses/search_form.rb
@@ -102,10 +102,17 @@ module Courses
     end
 
     def order
+      # The user's explicit sort choice is preserved across requests EXCEPT
+      # the one-time transition from "no location" to "with location" — at
+      # that moment the carried-over order (course_name_ascending, the
+      # no-location default) should yield to the new location's natural
+      # default. Other incompatible combinations (distance without coords,
+      # fee ordering without fee funding) are still caught inside
+      # OrderingStrategy.
       OrderingStrategy.new(
         search_location: search_location,
         funding: funding,
-        current_order: location_category_changed? ? nil : super,
+        current_order: location_category_first_set? ? nil : super,
       ).call
     end
 
@@ -256,6 +263,18 @@ module Courses
         formatted_address: formatted_address,
         short_address: short_address,
       )
+    end
+
+    # True only when the user is moving from "no location" to "with
+    # location" on this request. Distinct from #location_category_changed?,
+    # which also fires when the category drifts between two location
+    # categories (locality ↔ regional) — that drift can be caused by Google
+    # returning different address_types for the same query and would
+    # otherwise silently override a sort the user just picked.
+    def location_category_first_set?
+      return false if attributes["previous_location_category"].nil?
+
+      previous_location_category.blank? && location_category.present?
     end
 
     def boolean_filter_count(value)
--- a/app/forms/courses/search_form.rb
+++ b/app/forms/courses/search_form.rb
@@ -277,8 +277,7 @@ module Courses
     end
 
     def ordering_filter_count
-      default_order = search_location.located? ? "distance" : "course_name_ascending"
-      order == default_order ? nil : 1
+      order == OrderingStrategy.default_for(search_location) ? nil : 1
     end
 
     def radius_filter_count
--- a/app/forms/courses/search_form.rb
+++ b/app/forms/courses/search_form.rb
@@ -103,7 +103,7 @@ module Courses
 
     def order
       OrderingStrategy.new(
-        location: geocoded_location,
+        search_location: search_location,
         funding: funding,
         current_order: location_category_changed? ? nil : super,
       ).call
@@ -248,10 +248,14 @@ module Courses
 
   private
 
-    def geocoded_location
-      return nil unless latitude.present? && longitude.present?
-
-      location
+    def search_location
+      @search_location ||= SearchLocation.new(
+        text: location,
+        latitude: latitude,
+        longitude: longitude,
+        formatted_address: formatted_address,
+        short_address: short_address,
+      )
     end
 
     def boolean_filter_count(value)
@@ -273,7 +277,7 @@ module Courses
     end
 
     def ordering_filter_count
-      default_order = location.present? ? "distance" : "course_name_ascending"
+      default_order = search_location.sortable_by_distance? ? "distance" : "course_name_ascending"
       order == default_order ? nil : 1
     end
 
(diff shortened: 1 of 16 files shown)


tomas-stefano:
https://trello.com/c/gq9Y7YkE/1697-solve-is-this-a-location-based-search-rule-in-find-in-the-codebase

inulty-dfe (approved):

PR #6143, opened 2026-06-01
Remove schools from a provider

tomas-stefano:
## What this does

NITE had about 1,250 schools added to their account by the Register import, but they only use 25. The long list makes it hard for them (and the support team) to manage schools and add them to courses. Removing schools one at a time in Publish would take far too long.

This adds a task that removes all of a provider's schools except the ones we want to keep.

## How it works

A new rake task, `remove_provider_schools`, takes:

- `PROVIDER_CODE` – the provider to clean up
- `YEAR` – the recruitment cycle
- `KEEP_CSV` – a CSV file listing the school URNs to keep

For NITE, the keep list is the 25 schools they have attached to courses.

For every school the provider has (in that year), it:

- keeps any school whose URN is in the CSV
- keeps the provider's main site
- skips and reports any school still attached to a course, so we never break a live course by accident
- removes the rest

Removal is a soft delete (discard). The schools disappear from Publish straight away but can be brought back if needed. Every removal is recorded in the audit log.

It also checks the keep list against the account and reports which of the 25 URNs were actually found and which were missing, so support can spot any data problems.

## Safety

- There is a dry run (`remove_provider_schools:dry_run`) that shows exactly what would happen without changing anything. Always run this first.
- A summary is saved for every run with counts and details (removed, skipped, kept).
- Nothing is hard deleted, so it can be undone.

## How to run

```
PROVIDER_CODE=XYZ YEAR=2026 KEEP_CSV=/path/to/keep.csv rake remove_provider_schools:dry_run
PROVIDER_CODE=XYZ YEAR=2026 KEEP_CSV=/path/to/keep.csv rake remove_provider_schools:run
```

Check the dry run summary first: removed count should be about 1,225, kept count should be 25 (look into anything missing), and review the skipped-with-courses list before doing the real run.

## Tests

New specs cover the filter, the discard, the dry run, the summary, the keep-list check, and the rake flow. All green.

### DIFF
--- a/app/models/data_hub/remove_provider_schools_process_summary.rb
+++ b/app/models/data_hub/remove_provider_schools_process_summary.rb
@@ -1,0 +1,16 @@
+module DataHub
+  class RemoveProviderSchoolsProcessSummary < ProcessSummary
+    jsonb_accessor :short_summary,
+                   removed_count: :integer,
+                   skipped_with_courses_count: :integer,
+                   kept_present_count: :integer,
+                   kept_missing_count: :integer,
+                   error: :string
+
+    jsonb_accessor :full_summary,
+                   removed: [:jsonb, { array: true, default: [] }],
+                   skipped_with_courses: [:jsonb, { array: true, default: [] }],
+                   kept_present: [:string, { array: true, default: [] }],
+                   kept_missing: [:string, { array: true, default: [] }]
+  end
+end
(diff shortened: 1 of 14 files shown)


tomas-stefano:
https://trello.com/c/ZJRqLxnm/1558-bulk-remove-schools-from-nite

inulty-dfe (approved):
This is more complicated than I expected 😆

PR #6144, opened 2026-06-02
Implement Course start date step

gms-gs:
## Context

Add Course start date step

- Add new step
- Add tests
- Update tests
- Refactor `provider` and `recruitment_cycle` to be referenced in `course_wizard.rb` t o reduce repetitive code duplication across steps.

## Guidance to review

Run through the flow compare to QA please note the below. Validate this is working as expected and compare with QA.

```
      graph.add_multiple_conditional_edges(
        from: :study_sites,
        branches: [
          # Further education does not require visa sponsorship in the
          # existing flow, so it goes straight to start date.
          { when: :further_education_level?, then: :start_date },
          # TDA also goes straight to start date.
          { when: :undergraduate_degree_with_qts?, then: :start_date },
        ],
        # TODO: Visa sponsorship steps for other routes will go here.
        default: :courses_index,
      )
```

## Checklist

### DIFF
--- a/app/wizards/course_wizard/steps/start_date.rb
+++ b/app/wizards/course_wizard/steps/start_date.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+class CourseWizard
+  module Steps
+    class StartDate
+      include DfE::Wizard::Step
+
+      attribute :start_date
+
+      validates :start_date, presence: { message: I18n.t("course_wizard.steps.start_date.errors.start_date.blank") }
+
+      def start_date_options
+        cycle_year = wizard.provider.recruitment_cycle.year.to_i
+        options = (1..12).map { |m| "#{Date::MONTHNAMES[m]} #{cycle_year}" } +
+          (1..7).map { |m| "#{Date::MONTHNAMES[m]} #{cycle_year + 1}" }
+
+        return options if persisted?
+
+        index = options.index(sliced_label_for_today(cycle_year))
+
+        index.blank? ? options : options[index..]
+      end
+
+      def self.permitted_params
+        [:start_date]
+      end
+
+    private
+
+      def sliced_label_for_today(cycle_year)
+        today = Time.zone.today
+        january_label = january_label_for(cycle_year)
+
+        if today.year < cycle_year
+          # We're before January starts, so slice at "January <cycle_year>"
+          january_label
+        elsif today.year == cycle_year
+          # In cycle year, slice at the actual month
+          "#{Date::MONTHNAMES[today.month]} #{cycle_year}"
+        else
+          # Default to first month
+          january_label
+        end
+      end
+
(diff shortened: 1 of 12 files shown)


gms-gs:
[Implement Course start date step](https://trello.com/c/4w5JXzBl/1704-implement-course-start-date-step)

app/wizards/course_wizard/steps/start_date.rb
```diff
@@ -0,0 +1,51 @@
...
+          january_label
+        end
+      end
+
+      def january_label_for(cycle_year)
+        "#{Date::MONTHNAMES[1]} #{cycle_year}"
+      end
+    end
+  end
+end
```
tomas-stefano, line 52: This line was copied verbatim from the legacy `Courses::EditOptions::StartDateConcern` (`app/models/concerns/courses/edit_options/start_date_concern.rb:14`), where it works because the concern is mixed into the Course `ActiveRecord` model — `persisted?` is `true` when editing an existing course.

But wizard steps include `DfE::Wizard::Step` → `ActiveModel::Model`, which are plain form `objects.persisted?` is never overridden anywhere in the app or gem, so it always returns false. The branch is unreachable.

  Why it matters: The intent of that guard is "if the user already chose a start date, show all options so their saved choice still renders." Because it's dead, a user who picked e.g. June 2026 and later revisits the step after the month rolls over gets an option list sliced to July 2026…, and their previously-selected month is no longer rendered at all. It's a low-frequency edge case (the wizard state expires in 24h, so it only bites at a month boundary), but it's a latent correctness bug and confusing dead code for the next reader.

  Fix: replace the inapplicable persisted? check with the actual "already chosen" condition:

`return options if start_date.present?`

app/wizards/course_wizard/steps/start_date.rb
```diff
@@ -0,0 +1,51 @@
...
+  module Steps
+    class StartDate
+      include DfE::Wizard::Step
+
+      attribute :start_date
+
+      validates :start_date, presence: { message: I18n.t("course_wizard.steps.start_date.errors.start_date.blank") }
+
+      def start_date_options
+        cycle_year = wizard.provider.recruitment_cycle.year.to_i
```
tomas-stefano: The wizard already exposes `recruitment_cycle_year` directly (attr_accessor, from the URL), and wizard.provider is itself resolved via recruitment_cycle.providers.find_by!. So this line triggers a provider lookup + an association load purely to recover a value the wizard already holds. By construction wizard.provider.recruitment_cycle.year ==  wizard.recruitment_cycle_year. Prefer:   cycle_year = wizard.recruitment_cycle_year.to_i

  (The legacy concern used  rovider.recruitment_cycle.year only because, in the Course model, provider is the natural association — that reasoning doesn't carry over to the wizard.)

tomas-stefano (approved):

PR #6145, opened 2026-06-02
Implement Visa sponsorship and Accredited provider step

gms-gs:
## Context

The add-course wizard previously sent most providers straight from study sites to the courses index. The legacy publish flow already asks about student visa sponsorship, and for school-based providers with multiple accredited partners it also asks them to choose an accredited provider.

This PR adds the **Student visas** step and aligns post-study-sites behavior with the legacy flow, including the accredited-provider selection path and study-sites skipping behavior when no study sites exist.

**Note:** If the user answers **Yes** to sponsoring student visas, they are sent to the **courses index** for now. A dedicated visa application deadline step is still a TODO in `course_wizard.rb`.

- Add `CourseWizard::Steps::VisaSponsorship` with validation for `can_sponsor_student_visa`
- Add `_visa_sponsorship.html.erb` (Yes/No radios, conditional overseas guidance, accrediting-provider inset)
- Add locale strings under `course_wizard.steps.visa_sponsorship`
- Route study sites → **visa sponsorship** by default (TDA and FE still go to start date)
- Route visa sponsorship → **start date** when No, or **courses index** when Yes (`visa_sponsorship_required?` in state store)
- Add/expand wizard helper logic for `accrediting_provider` so it supports:
  - single partner auto-selection
  - selected `accredited_provider_code` when multiple partners
- Add `CourseWizard::Steps::AccreditedProvider` and `_accredited_provider.html.erb`
- Route school-based providers with multiple accredited partners:
  - study sites → **accredited provider** → visa sponsorship
- Add `skip_when` behavior for study sites in the wizard graph and `skip_study_sites?` rule for providers with no study sites
- Add locale strings under `course_wizard.steps.accredited_provider`
- Update/add specs for:
  - visa sponsorship step and navigation/content permutations
  - accredited provider wizard step (unit + system)
  - updated graph next/previous step behavior
  - study-sites skip and routing cases
  - shared wizard test context/provider setup updates

## Guidance to review

- **`app/wizards/course_wizard.rb`**  
  Confirm graph transitions match legacy expectations:
  - study sites → start date (TDA/FE)
  - study sites → accredited provider (school-based + multiple partners)
  - study sites → visa sponsorship (default)
  - accredited provider → visa sponsorship
  - visa sponsorship → courses index/start date based on answer
  - study sites skipped when provider has no study sites

- **`app/wizards/course_wizard/steps/visa_sponsorship.rb`**, **`app/views/publish/course_wizards/steps/_visa_sponsorship.html.erb`**  
  Validate conditional copy/UI behavior for accredited vs school-based providers and accrediting partner messaging.

- **`app/wizards/course_wizard/steps/accredited_provider.rb`**, **`app/views/publish/course_wizards/steps/_accredited_provider.html.erb`**  
  Validate provider-selection UI and validation matches legacy “Accredited provider” behavior.

- **`app/wizards/course_wizard/state_stores/course_wizard_store.rb`**  
  Confirm `visa_sponsorship_required?` behavior is still correct for interim routing.

- **Specs**
  - `spec/system/publish/courses/add_course_wizard/visa_sponsorship_spec.rb`
  - `spec/system/publish/courses/add_course_wizard/accredited_provider_spec.rb`
  - `spec/system/publish/courses/add_course_wizard/study_sites_spec.rb`
  - `spec/wizards/add_course_wizard/steps/visa_sponsorship_spec.rb`
  - `spec/wizards/add_course_wizard/steps/accredited_provider_spec.rb`
  - `spec/wizards/add_course_wizard/wizard/next_step_spec.rb`
  - `spec/wizards/add_course_wizard/wizard/previous_step_spec.rb`

### DIFF
--- a/app/views/publish/course_wizards/steps/_visa_sponsorship.html.erb
+++ b/app/views/publish/course_wizards/steps/_visa_sponsorship.html.erb
@@ -1,0 +1,35 @@
+<% content_for :page_title, title_with_error_prefix(t("course_wizard.steps.visa_sponsorship.heading"), current_step.errors.any?) %>
+
+<h1 class="govuk-heading-l">
+  <%= render CaptionText.new(text: t("course_wizard.steps.visa_sponsorship.caption")) %>
+  <%= t("course_wizard.steps.visa_sponsorship.heading") %>
+</h1>
+
+<% if current_step.show_recruiting_from_overseas_guidance? %>
+  <p class="govuk-body">
+    <%= t("course_wizard.steps.visa_sponsorship.recruiting_from_overseas.prefix") %>
+    <%= govuk_link_to t("course_wizard.steps.visa_sponsorship.recruiting_from_overseas.link_text"), t("course_wizard.steps.visa_sponsorship.recruiting_from_overseas.link_url"), target: "_blank" %>.
+  </p>
+<% end %>
+
+<h2 class="govuk-heading-m">
+  <%= current_step.question %>
+</h2>
+
+<% if current_step.show_accrediting_provider_inset_text? %>
+  <div class="govuk-inset-text">
+    <% if current_step.accrediting_provider_can_sponsor_student_visa? %>
+      <%= t("course_wizard.steps.visa_sponsorship.inset.can_sponsor", provider_name: current_step.accrediting_provider_name) %>
+    <% else %>
+      <p class="govuk-body"><%= t("course_wizard.steps.visa_sponsorship.inset.cannot_sponsor", provider_name: current_step.accrediting_provider_name) %></p>
+      <p class="govuk-body"><%= t("course_wizard.steps.visa_sponsorship.inset.contact_provider", provider_name: current_step.accrediting_provider_name) %></p>
+    <% end %>
+  </div>
+<% end %>
+
+<%= form.govuk_radio_buttons_fieldset :can_sponsor_student_visa, legend: { hidden: true } do %>
+  <%= form.govuk_radio_button :can_sponsor_student_visa, true, label: { text: t("course_wizard.steps.visa_sponsorship.options.can_sponsor_student_visa.label") } %>
+  <%= form.govuk_radio_button :can_sponsor_student_visa, false, label: { text: t("course_wizard.steps.visa_sponsorship.options.cannot_sponsor_student_visa.label") } %>
+<% end %>
+
+<%= form.govuk_submit t("course_wizard.steps.visa_sponsorship.submit") %>
--- a/app/views/publish/course_wizards/steps/_visa_sponsorship.html.erb
+++ b/app/views/publish/course_wizards/steps/_visa_sponsorship.html.erb
@@ -19,7 +19,7 @@
 <% if current_step.show_accrediting_provider_inset_text? %>
   <div class="govuk-inset-text">
     <% if current_step.accrediting_provider_can_sponsor_student_visa? %>
-      <%= t("course_wizard.steps.visa_sponsorship.inset.can_sponsor", provider_name: current_step.accrediting_provider_name) %>
+      <p class="govuk-body"><%= t("course_wizard.steps.visa_sponsorship.inset.can_sponsor", provider_name: current_step.accrediting_provider_name) %></p>
     <% else %>
       <p class="govuk-body"><%= t("course_wizard.steps.visa_sponsorship.inset.cannot_sponsor", provider_name: current_step.accrediting_provider_name) %></p>
       <p class="govuk-body"><%= t("course_wizard.steps.visa_sponsorship.inset.contact_provider", provider_name: current_step.accrediting_provider_name) %></p>
--- a/app/wizards/course_wizard.rb
+++ b/app/wizards/course_wizard.rb
@@ -65,6 +65,9 @@ class CourseWizard
           # School-based providers with multiple accredited partners need
           # to choose who is accrediting the course.
           { when: :accredited_provider_selection_required?, then: :accredited_provider },
+
+          # TODO: mirror visa split here:
+          # non-fee courses -> Skilled Worker visa path, fee courses -> Student visa path.
         ],
         default: :visa_sponsorship,
       )
--- a/app/wizards/course_wizard.rb
+++ b/app/wizards/course_wizard.rb
@@ -121,16 +121,21 @@ class CourseWizard
   end
 
   def accredited_provider_selection_required?
-    !provider.accredited? && provider.accredited_partners.many?
+    !provider.accredited? && accredited_partners.many?
   end
 
   def accrediting_provider
     return if provider.accredited?
 
-    partners = provider.accredited_partners
-    return partners.first if partners.one?
+    @accrediting_provider ||= begin
+      return accredited_partners.first if accredited_partners.one?
 
-    selected_provider_code = state_store.accredited_provider_code
-    partners.find_by(provider_code: selected_provider_code) if selected_provider_code.present?
+      selected_provider_code = state_store.accredited_provider_code
+      accredited_partners.find { |partner| partner.provider_code == selected_provider_code } if selected_provider_code.present?
+    end
+  end
+
+  def accredited_partners
+    @accredited_partners ||= provider.accredited_partners.to_a
   end
 end
--- a/app/wizards/course_wizard.rb
+++ b/app/wizards/course_wizard.rb
@@ -19,7 +19,8 @@ class CourseWizard
       graph.add_node :funding_type, Steps::FundingType
       graph.add_node :study_pattern, Steps::StudyPattern
       graph.add_node :schools, Steps::Schools
-      graph.add_node :study_sites, Steps::StudySites
+      graph.add_node :study_sites, Steps::StudySites, skip_when: :skip_study_sites?
+      graph.add_node :accredited_provider, Steps::AccreditedProvider
       graph.add_node :start_date, Steps::StartDate
       graph.add_node :visa_sponsorship, Steps::VisaSponsorship
 
@@ -61,9 +62,14 @@ class CourseWizard
           { when: :further_education_level?, then: :start_date },
           # TDA also goes straight to start date.
           { when: :undergraduate_degree_with_qts?, then: :start_date },
+          # School-based providers with multiple accredited partners need
+          # to choose who is accrediting the course.
+          { when: :accredited_provider_selection_required?, then: :accredited_provider },
         ],
         default: :visa_sponsorship,
       )
+
+      graph.add_edge from: :accredited_provider, to: :visa_sponsorship
 
       graph.add_multiple_conditional_edges(
         from: :visa_sponsorship,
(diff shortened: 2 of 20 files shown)


gms-gs:
[Implement Visas sponsorship step](https://trello.com/c/6I4ZbS7O/1706-implement-visas-sponsorship-step)

app/wizards/course_wizard.rb
```diff
@@ -100,4 +115,22 @@ def provider
   def recruitment_cycle
     @recruitment_cycle ||= RecruitmentCycle.find_by!(year: recruitment_cycle_year)
   end
+
+  def skip_study_sites?
+    provider.study_sites.none?
```
tomas-stefano, line 125: what if is only one?

gms-gs: If its 1 then the user will see study_sites page. That’s intentional and matches the old controller behaviour (new only skipped when there were zero study sites). @tomas-stefano

app/wizards/course_wizard/state_stores/course_wizard_store.rb
```diff
@@ -16,6 +16,10 @@ def primary_level?
       def undergraduate_degree_with_qts?
         qualification == "undergraduate_degree_with_qts"
       end
+
+      def visa_sponsorship_required?
+        can_sponsor_student_visa == true
```
tomas-stefano: One thing to note is the `== true` comparison. ON `can_sponsor_student_visa == true` we need to make sure the repository read returns true or false not 'true' or 'false'. 

Can you double check, please? To save as true or false and then debug this method on what returns.

config/locales/en/publish/course_wizard.yml
```diff
@@ -118,13 +118,43 @@ en:
...
+          cannot_sponsor: "%{provider_name} have said they cannot sponsor Student visas so we have defaulted your answer to 'No'."
+          contact_provider: "If your organisation would like to sponsor Student visas, contact %{provider_name}."
+        options:
+          can_sponsor_student_visa:
+            label: "Yes"
+          cannot_sponsor_student_visa:
+            label: "No"
+        errors:
+          can_sponsor_student_visa:
+            blank: Select if student visas can be sponsored for this course
```
tomas-stefano, line 157: "Defaulted to No" copy with no actual default (visa_sponsorship inset). The inset says "…cannot sponsor Student visas so we have defaulted your answer to 'No'." (course_wizard.yml, inset.cannot_sponsor), but the step's  can_sponsor_student_visa attribute starts nil, so neither radio is pre-selected — and the inclusion: [true, false]  validation forces the user to pick. So the UI promises a default it doesn't apply, and the user is still required to  choose. Either pre-select "No" (set the attribute default) or reword the copy.

app/wizards/course_wizard.rb
```diff
@@ -100,4 +115,22 @@ def provider
...
+
+  def accrediting_provider
+    return if provider.accredited?
+
+    partners = provider.accredited_partners
+    return partners.first if partners.one?
+
+    selected_provider_code = state_store.accredited_provider_code
+    partners.find_by(provider_code: selected_provider_code) if selected_provider_code.present?
+  end
```
tomas-stefano, line 137: Repeated un-memoized queries: `wizard.accrediting_provider` is not memoized, and the visa view calls it three times
  (show_accrediting_provider_inset_text?, accrediting_provider_name, accrediting_provider_can_sponsor_student_visa?) —  each re-runs accredited_partners (.one? / .find_by). Also accredited_provider_selection_required? and   accrediting_provider both re-query provider.accredited_partners. Memoizing accrediting_provider (and/or the partners relation) removes several redundant queries per render.

app/views/publish/course_wizards/steps/_visa_sponsorship.html.erb
```diff
@@ -0,0 +1,35 @@
...
+<% if current_step.show_accrediting_provider_inset_text? %>
+  <div class="govuk-inset-text">
+    <% if current_step.accrediting_provider_can_sponsor_student_visa? %>
+      <%= t("course_wizard.steps.visa_sponsorship.inset.can_sponsor", provider_name: current_step.accrediting_provider_name) %>
+    <% else %>
+      <p class="govuk-body"><%= t("course_wizard.steps.visa_sponsorship.inset.cannot_sponsor", provider_name: current_step.accrediting_provider_name) %></p>
+      <p class="govuk-body"><%= t("course_wizard.steps.visa_sponsorship.inset.contact_provider", provider_name: current_step.accrediting_provider_name) %></p>
+    <% end %>
+  </div>
+<% end %>
```
tomas-stefano, line 28: Inset markup inconsistency: (_visa_sponsorship.html.erb): the can_sponsor branch emits bare text in the `govuk-inset-text div`, while cannot_sponsor wraps each line in `<p class="govuk-body">`. Just wanna flag it.

spec/system/publish/courses/add_course_wizard/accredited_provider_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def provider
+    @provider ||= @user.providers.first
+  end
+
+  def wizard_state_key
+    @wizard_state_key ||= SecureRandom.uuid
+  end
+end
```
tomas-stefano, line 131: Nice work on this — the step logic and unit coverage are thorough, and the content-based system scenarios (e.g.
  "accredited provider that cannot sponsor … shows organisation question and overseas guidance") read exactly how we'd  want feature specs to read: they assert on what the user actually sees.
  
One bit of feedback on the system specs — a few of them lean more a box than user-perspective:

  1. State is seeded through internal classes. `and_i_have_wizard_state_for_*` instantiates
  `CourseWizard::Repositories::Course` and CourseWizard::StateStores::CourseWizardStore directly. A user never touches those, and it couples the feature suite to our persistence internals — rename the store and the specs break even though nothing user-visible changed. Could we hide this behind a single shared helper (e.g.
  given_wizard_state_at(:visa_sponsorship, level: "secondary")) so the coupling lives in one place?
  2. Navigation is asserted by route, not by content. `then_i_am_taken_to_the_start_date_page` checks `have_current_path(... step: :start_date)`. That proves the URL changed, not that the user landed on the right page (a path can match while the page errors). Asserting on visible content — expect(page).to have_content("Course start 
  date") — is more robust and closer to the user's perspective. Cheap change, and IMO the highest-value one here.
  3. Optional: one true end-to-end happy path that clicks Continue from the wizard start through to the visa step, as a guard that the real flow wires up. The per-step specs can keep seeding state for speed.

Worth noting this mostly matches our existing wizard system-spec pattern (study_sites_spec/schools_spec do the same),   so I'm not blocking on it — but if we're touching this area, nudging toward content assertions would be a good direction. The path-vs-content point (#2) is the one I'd most like to see before merge.

gms-gs: I agree with all of these points. I would suggest once I have finished the remaining steps I go back and update tests and also add full E2E tests for the different branches.

app/wizards/course_wizard.rb
```diff
@@ -100,4 +118,27 @@ def provider
...
+      return accredited_partners.first if accredited_partners.one?
+
+      selected_provider_code = state_store.accredited_provider_code
+      accredited_partners.find { |partner| partner.provider_code == selected_provider_code } if selected_provider_code.present?
+    end
+  end
+
+  def accredited_partners
+    @accredited_partners ||= provider.accredited_partners.to_a
+  end
```
tomas-stefano, line 137: A bit of OO design feedback on the methods added to `CourseWizard` (`accredited_provider_selection_required?`, `accrediting_provider`, `accredited_partners`).

These three feel like a different responsibility than the wizard itself. The wizard's job is flow orchestration — "given the state, what's the next step." 

But these answer a *domain* question: "for this school-based provider and the user's choice, who is accrediting this course?" That's a cohesive cluster that's been bolted onto the orchestrator.

There's a concrete symptom of the misfit — the memoization is half-broken:

```ruby
@accrediting_provider ||= begin
  return accredited_partners.first if accredited_partners.one?   # returns from the METHOD, never assigns @accrediting_provider
  ...
end
```

The `return` inside the `begin` short-circuits the method, so the one-partner branch is never memoized, and the many-but-unselected branch memoizes `nil` (falsy, so it re-runs anyway). The control flow is tangled *because* it's living in the wrong object.

### Suggestion: extract a small collaborator

```ruby
# app/wizards/course_wizard/accreditation.rb
class CourseWizard
  class Accreditation
    def initialize(provider:, selected_provider_code: nil)
      @provider = provider
      @selected_provider_code = selected_provider_code
    end

    # School-based providers with more than one partner must choose.
    def selection_required?
      requires_partner? && partners.many?
    end

    def accrediting_provider
      return unless requires_partner?
      return partners.first if partners.one?

      selected_partner
    end

    def partners
      @partners ||= provider.accredited_partners.sort_by(&:provider_name)
    end

  private

    attr_reader :provider, :selected_provider_code

    def requires_partner?
      !provider.accredited?
    end

    def selected_partner
      return if selected_provider_code.blank?

      partners.find { |partner| partner.provider_code == selected_provider_code }
    end
  end
end
```

The wizard becomes a thin facade that builds the collaborator from the two things it owns (the provider + wizard state) and delegates:

```ruby
def accreditation
  @accreditation ||= Accreditation.new(
    provider:,
    selected_provider_code: state_store.accredited_provider_code,
  )
end

delegate :accrediting_provider, to: :accreditation

def accredited_provider_selection_required?
  accreditation.selection_required?
end
```

Memoization is now correct and trivial, each branch is a named method instead of an inline `return` in a `begin`, and it's testable as a plain object with no wizard graph/state store to stand up.

### Bonus: kills a duplication

The `AccreditedProvider` step has its own `accredited_partners` (`wizard.provider.accredited_partners.sort_by(&:provider_name)`) — the same sorted list. Both steps can route through one source of truth:

```ruby
# AccreditedProvider step
def accredited_partners = wizard.accreditation.partners

# VisaSponsorship step (private)
def accrediting_provider = wizard.accreditation.accrediting_provider
```

### Why not move these onto the state store?

I considered it, but I'd avoid it. Everything on `CourseWizardStore` today is pure state interpretation of a single persisted answer (`further_education_level?`, `visa_sponsorship_required?`), and the store has no `provider` — it's built as `CourseWizardStore.new(repository:)`. These methods depend on the Provider domain model (`provider.accredited?`, `provider.accredited_partners`), so hosting them there would force you to inject the provider into the store and mix cache-backed values with live DB queries — turning a thin cache interpreter into something with two responsibilities, and ballooning its specs into domain setup.

The one piece that *does* belong near the store is the answer itself — `accredited_provider_code` already lives there, and a predicate over it would fit the existing pattern:

```ruby
def accredited_provider_selected? = accredited_provider_code.present?
```

So: store owns the *answer*, the collaborator owns the *resolution*. Note the dependency direction — `Accreditation` takes a plain `String` code, not the store, so the store stays ignorant of providers and the collaborator stays ignorant of the cache.

app/wizards/course_wizard/state_stores/course_wizard_store.rb
```diff
@@ -16,6 +16,10 @@ def primary_level?
       def undergraduate_degree_with_qts?
         qualification == "undergraduate_degree_with_qts"
       end
+
+      def visa_sponsorship_required?
+        ActiveModel::Type::Boolean.new.cast(can_sponsor_student_visa) == true
```
inulty-dfe: You don't need the `== true`, it will return true or false by default

tomas-stefano (approved):

PR #6150, opened 2026-06-04
Add GOR `region_code` to GiasSchool

inulty-dfe:
## Context

The existing Placement Schools implementation has a region_code property that corresponds to the Government Organisation Region (GOR) in which the school is situated. This value is set by Geocoding the Site when it's created.

Since we moved from allowing providers to create arbitrary sites to only allowing GIAS schools, we have this region code information available to us directly from the GIAS service.

We need to migrate a column onto the GiasSchool model to store the region_code for each school.

- Migrate `region_code` onto the `GiasSchool` model.
- Use it to further filter the schools we import (Establishment type isn't enough to guarantee non-English schools aren't imported.

## Checklist


inulty-dfe:
[5.5 School remodel - Adding region_code to gias_school table](https://trello.com/c/gWAitP6d/1699-55-school-remodel-adding-regioncode-to-giasschool-table)

tomas-stefano (approved):

langesi (approved):
Well done

PR #6153, opened 2026-06-05
Fix NameError on unauthenticated note requests

tomas-stefano:
## What

`Find::Candidates::NotesController` includes `Find::Authentication` but never defined `reason_for_request`, so unauthenticated HTML requests raised `NameError` in `request_authentication` (Sentry, production).

## Fix

Move the `:general` default into the concern and drop the three identical per-controller overrides. Add a `NotesController` spec covering the unauthenticated redirect path.

### DIFF
--- a/app/controllers/concerns/find/authentication.rb
+++ b/app/controllers/concerns/find/authentication.rb
@@ -82,6 +82,10 @@ module Find
       end
     end
 
+    def reason_for_request
+      :general
+    end
+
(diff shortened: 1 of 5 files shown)


langesi (approved):
LGTM

inulty-dfe:
[Bug: Error in Find notes controller](https://trello.com/c/cChA2w9D/1722-bug-error-in-find-notes-controller)

PR #6154, opened 2026-06-05
Add SkilledWorkerVisa step

gms-gs:
## Context

This change introduces a dedicated Skilled Worker visa step in the add-course wizard for funding types that require it (`salary` and `apprenticeship`), instead of relying only on the existing student visa flow.

The key behaviour to note is:
- the step defaults to **No** when unset
- routing now depends on the answer:
  - **Yes** → courses index
  - **No** → start date

- Added new wizard step class: `CourseWizard::Steps::SkilledWorkerVisa`
  - boolean attribute `can_sponsor_skilled_worker_visa`
  - inclusion validation for `[true, false]` with custom error
  - default getter behaviour to `false` when nil
  - permitted params defined
- Added new view partial for the step:
  - question + yes/no radios + continue CTA
- Added i18n content for heading/question/options/errors in `course_wizard.yml`
- Updated wizard graph to include the new step and branching:
  - from `study_sites` to `skilled_worker_visa` when required
  - from `skilled_worker_visa` to `courses_index` when true, else `start_date`
- Extended `CourseWizardStore` with skilled-worker predicates:
  - `skilled_worker_visa_required?`
  - `skilled_worker_visa_sponsorship_required?`
- Added/updated test coverage:
  - system spec for skilled worker step
  - step spec
  - state store spec
  - next/previous step wizard specs
  - repository spec support updates
- Added explicit spec assertion for `.permitted_params` on `SkilledWorkerVisa`
- Updated next-step wizard tests to include `can_sponsor_skilled_worker_visa` state and assert:
  - sponsored path goes to `courses_index`
  - non-sponsored path goes to `start_date`

## Guidance to review

- Start with wizard graph/state changes:
  - `app/wizards/course_wizard.rb`
  - `app/wizards/course_wizard/state_stores/course_wizard_store.rb`
- Then review step behaviour and validation:
  - `app/wizards/course_wizard/steps/skilled_worker_visa.rb`
  - `app/views/publish/course_wizards/steps/_skilled_worker_visa.html.erb`
- Finally validate coverage:
  - `spec/system/publish/courses/add_course_wizard/skilled_worker_visa_spec.rb`
  - `spec/wizards/add_course_wizard/wizard/next_step_spec.rb`
  - `spec/wizards/add_course_wizard/steps/skilled_worker_visa_spec.rb`
  - `spec/wizards/add_course_wizard/state_stores/course_wizard_store_spec.rb`

### DIFF
--- a/app/wizards/course_wizard.rb
+++ b/app/wizards/course_wizard.rb
@@ -7,7 +7,13 @@ class CourseWizard
 
   delegate :accrediting_provider, to: :accreditation
 
-  delegate :further_education_level?, :primary_level?, :undergraduate_degree_with_qts?, :visa_sponsorship_required?, to: :state_store
+  delegate :further_education_level?,
+           :primary_level?,
+           :undergraduate_degree_with_qts?,
+           :visa_sponsorship_required?,
+           :skilled_worker_visa_required?,
+           :skilled_worker_visa_sponsorship_required?,
+           to: :state_store
 
   def steps_processor
     DfE::Wizard::StepsProcessor::Graph.draw(self) do |graph|
@@ -25,6 +31,7 @@ class CourseWizard
       graph.add_node :accredited_provider, Steps::AccreditedProvider
       graph.add_node :start_date, Steps::StartDate
       graph.add_node :visa_sponsorship, Steps::VisaSponsorship
+      graph.add_node :skilled_worker_visa, Steps::SkilledWorkerVisa
 
       graph.add_node :courses_index, DfE::Wizard::Core::Redirect
 
@@ -67,9 +74,7 @@ class CourseWizard
           # School-based providers with multiple accredited partners need
           # to choose who is accrediting the course.
           { when: :accredited_provider_selection_required?, then: :accredited_provider },
-
-          # TODO: mirror visa split here:
-          # non-fee courses -> Skilled Worker visa path, fee courses -> Student visa path.
+          { when: :skilled_worker_visa_required?, then: :skilled_worker_visa },
         ],
         default: :visa_sponsorship,
       )
@@ -86,6 +91,14 @@ class CourseWizard
       )
 
       graph.add_edge from: :start_date, to: :courses_index
+
+      graph.add_multiple_conditional_edges(
+        from: :skilled_worker_visa,
+        branches: [
+          { when: :skilled_worker_visa_sponsorship_required?, then: :courses_index },
+        ],
+        default: :start_date,
+      )
     end
   end
 
--- a/app/wizards/course_wizard.rb
+++ b/app/wizards/course_wizard.rb
@@ -84,6 +84,7 @@ class CourseWizard
       graph.add_multiple_conditional_edges(
         from: :accredited_provider,
         branches: [
+          { when: :undergraduate_degree_with_qts?, then: :start_date },
           { when: :salary_based?, then: :skilled_worker_visa },
           { when: :fee_based?, then: :visa_sponsorship },
         ],
--- a/app/wizards/course_wizard.rb
+++ b/app/wizards/course_wizard.rb
@@ -11,7 +11,8 @@ class CourseWizard
            :primary_level?,
            :undergraduate_degree_with_qts?,
            :visa_sponsorship_required?,
-           :skilled_worker_visa_required?,
+           :salary_based?,
+           :fee_based?,
            :skilled_worker_visa_sponsorship_required?,
            to: :state_store
 
@@ -74,12 +75,19 @@ class CourseWizard
           # School-based providers with multiple accredited partners need
           # to choose who is accrediting the course.
           { when: :accredited_provider_selection_required?, then: :accredited_provider },
-          { when: :skilled_worker_visa_required?, then: :skilled_worker_visa },
+          { when: :salary_based?, then: :skilled_worker_visa },
         ],
         default: :visa_sponsorship,
       )
 
-      graph.add_edge from: :accredited_provider, to: :visa_sponsorship
+      graph.add_multiple_conditional_edges(
+        from: :accredited_provider,
+        branches: [
+          { when: :salary_based?, then: :skilled_worker_visa },
+          { when: :fee_based?, then: :visa_sponsorship },
+        ],
+        default: :start_date,
+      )
 
       graph.add_multiple_conditional_edges(
         from: :visa_sponsorship,
--- a/app/wizards/course_wizard.rb
+++ b/app/wizards/course_wizard.rb
@@ -70,11 +70,12 @@ class CourseWizard
           # Further education does not require visa sponsorship in the
           # existing flow, so it goes straight to start date.
           { when: :further_education_level?, then: :start_date },
-          # TDA also goes straight to start date.
-          { when: :undergraduate_degree_with_qts?, then: :start_date },
           # School-based providers with multiple accredited partners need
           # to choose who is accrediting the course.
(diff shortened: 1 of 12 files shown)


gms-gs:
[Implement can_sponsor_skilled_worker_visa step](https://trello.com/c/PEvP3Qj3/1712-implement-cansponsorskilledworkervisa-step)

app/wizards/course_wizard.rb
```diff
@@ -62,19 +70,25 @@ def steps_processor
...
 
-      graph.add_edge from: :accredited_provider, to: :visa_sponsorship
+      graph.add_multiple_conditional_edges(
+        from: :accredited_provider,
+        branches: [
+          { when: :salary_based?, then: :skilled_worker_visa },
+          { when: :fee_based?, then: :visa_sponsorship },
+        ],
+        default: :start_date,
+      )
```
tomas-stefano, line 92: The from: accredited_provider next-step tests  (next_step_spec.rb) cover fee → visa,  salary/apprenticeship → skilled worker, but not the  default: :start_date branch. That default only triggers  when funding_type is nil — i.e. the  TDA-with-multiple-accredited-partners case, which is  the precise behaviour the precedence reordering (commit  3609a64) was written to enable. The new test only  verifies TDA reaches accredited_provider, not that it  then proceeds to start_date. Add a from accredited  provider test with qualification:  "undergraduate_degree_with_qts" (no funding_type)
  asserting start_date.

app/wizards/course_wizard.rb
```diff
@@ -62,19 +70,25 @@ def steps_processor
...
+          # TDA goes straight to start date when no accredited provider
+          # selection is required.
+          { when: :undergraduate_degree_with_qts?, then: :start_date },
+          { when: :salary_based?, then: :skilled_worker_visa },
         ],
         default: :visa_sponsorship,
       )
 
-      graph.add_edge from: :accredited_provider, to: :visa_sponsorship
+      graph.add_multiple_conditional_edges(
```
tomas-stefano, line 84: The default: :start_date silently relies on  funding_type being nil for TDA courses. It works today  (TDA skips the funding-type step), but it's fragile —
  anything that ever populated a funding type on a TDA  course would reroute it through a visa step. An  explicit { when: :undergraduate_degree_with_qts?, then:  :start_date } branch would make the intent  self-documenting and robust. And a test too.

spec/wizards/add_course_wizard/state_stores/course_wizard_store_spec.rb
```diff
@@ -51,6 +51,20 @@
...
+    end
+
+    context "when funding type is not 'fee'" do
+      let(:funding_type) { "apprenticeship" }
+
+      it "returns false" do
+        expect(store.fee_based?).to be false
+      end
+    end
+  end
```
tomas-stefano, line 82: The #fee_based? block has only one context  ("apprenticeship" → false). #salary_based? correctly  tests three. Add a funding_type: "fee" → true context  for parity.

app/wizards/course_wizard/state_stores/course_wizard_store.rb
```diff
@@ -13,13 +13,25 @@ def primary_level?
         level == "primary"
       end
 
+      def fee_based?
+        funding_type == "fee"
+      end
+
+      def salary_based?
+        funding_type.in?(%w[salary apprenticeship])
+      end
```
tomas-stefano, line 23: It is subjective but salary_based? returns true for apprenticeship too. Paired with fee_based?, the name reads as if it's only salary. Consider non_fee_based? or a one-line comment clarifying it means "employment-based (salary or apprenticeship)".

app/wizards/course_wizard/steps/skilled_worker_visa.rb
```diff
@@ -0,0 +1,23 @@
...
+      include DfE::Wizard::Step
+
+      attribute :can_sponsor_skilled_worker_visa, :boolean
+
+      validates :can_sponsor_skilled_worker_visa, inclusion: { in: [true, false], message: I18n.t("course_wizard.steps.skilled_worker_visa.errors.can_sponsor_skilled_worker_visa.blank") }
+
+      def can_sponsor_skilled_worker_visa
+        return super unless super.nil?
+
+        false
```
tomas-stefano: I've spent quite a while finding this method strange and I could not identify the why on the reasoning or effects.

So this is what I understood:

The required radio question can never fail validation, and "No" is pre-selected:

`app/wizards/course_wizard/steps/skilled_worker_visa.rb:12-16`

```ruby
def can_sponsor_skilled_worker_visa
  return super unless super.nil?

  false
end
```

Walking through what this does:

1. `super` returns the raw attribute value. When the user hasn't answered yet, that's `nil`.
2. The guard `return super unless super.nil?` means: if the value is set, return it; **if it's `nil`, fall through and return `false`**.
3. So this getter can only ever return `true` or `false` — it never returns `nil`.

Why that's a problem:

- The validation `validates :can_sponsor_skilled_worker_visa, inclusion: { in: [true, false], ... }` reads the value *through this getter*. Because the getter turns `nil` into `false`, the validation always sees `true` or `false` and **always passes**. I confirmed this empirically: a freshly built step is `valid?` with zero errors before the user touches anything.
- That makes the `inclusion` rule and its locale message `course_wizard.steps.skilled_worker_visa.errors.can_sponsor_skilled_worker_visa.blank` (`config/locales/en/publish/course_wizard.yml`) **unreachable dead code**. A user who submits without choosing is silently recorded as "No" instead of seeing "Select if candidates can get a sponsored Skilled Worker visa".
- In the view (`app/views/publish/course_wizards/steps/_skilled_worker_visa.html.erb:10`), the form reads the same getter to decide which radio is checked. Since it returns `false`, the **"No" radio renders pre-selected** — which breaks the GOV.UK pattern of leaving a required question unanswered by default.

For comparison, `VisaSponsorship#can_sponsor_student_visa` (`app/wizards/course_wizard/steps/visa_sponsorship.rb`) only defaults to `false` in one narrow case:

```ruby
def can_sponsor_student_visa
  return super unless super.nil? && default_to_no_from_accrediting_provider?

  false
end
```

Everywhere else it leaves the value `nil`, so the required validation fires and no option is pre-checked. This step looks like a copy of that one with the `&& default_to_no_from_accrediting_provider?` condition dropped.

**Recommendation:** remove the unconditional default so the validation works and no radio is pre-selected:

```ruby
# delete the getter override entirely — the plain `attribute :can_sponsor_skilled_worker_visa, :boolean` is enough
```

If defaulting to "No" really is the intended behaviour, then go the other way and remove the now-meaningless `inclusion` validation and the unused `blank` locale key, so we're not shipping dead code.

Either way, note that the step spec `spec/wizards/add_course_wizard/steps/skilled_worker_visa_spec.rb` currently *asserts* "defaults to false when nil" → it's locking in this behaviour rather than catching it, so it'll need updating too.

gms-gs: Ill not default and just allow it to be nil in that case, and add validation they have to provide a value

tomas-stefano: Sorry I don't understand still. If you remove this code/method what happens? any bug?

gms-gs: Sorry I meant  ll remove the unconditional default getter, then update the step spec so nil is invalid again (and the blank message is exercised).

tomas-stefano (approved):

PR #6157, opened 2026-06-08
Courses List changes

tomas-stefano:
## Context

Course list was hard to scan: each course's details ran together in a sentence under the name, and courses that differ only by **start date** looked identical (the date wasn't shown).

This PR delivers the Trello card — and rebuilds the page properly while doing it.

**What changed for users**
- Course details now live in their own **"Course information"** column, one per line: funding type, qualification, study type, and the **start date** (e.g. "September 2026").
- Added an "Ages X to Y" hint under each course name.
- Courses stay grouped by accredited provider (self-accredited first, then alphabetical).

(Ships immediately — no feature flag, no rollover gate.)

**What changed under the hood**

The old page used a string-keyed grouping hash and per-course status lookups (an N+1). It's now built around a Find-style query object, `Publish::Courses::Query`, which returns the courses **with their status computed in SQL** (`content_status` + `has_unpublished_changes` columns) so the list loads **no enrichment or site rows at all**. A thin `Publish::CourseList` facade groups the rows, and a small ViewComponent tree (List → Group → Table → Row → Information / StatusTag) renders them.

- **Performance:** the full list now renders in ~4 constant queries (worst real provider was ~460), with a spec guarding against the N+1 returning.
- **Source of truth:** the SQL status is a port of `Courses::ContentStatusService` / `Course#has_unpublished_changes?`; a cross-check spec asserts the read-model renders an identical status tag to the canonical `ApplicationDecorator#status_tag`, so the two can't drift.
- The training-partners course list was unified onto the same query/components.
- Course-list table styling is scoped to its own class (equal-thirds fixed columns, bottom-aligned headers, tight info lines) so it can't affect other tables.

The query object is organised for the **filters coming next** (status / education phase / funding / qualification / start date).

## Diagrams

```mermaid
flowchart TD
  %% ---------- Controllers ----------
  subgraph CTRL["Controllers"]
    CC["Publish::CoursesController#index<br/>@course_list = Publish::CourseList.new(provider:)"]
    TPC["Publish::Providers::TrainingPartners::CoursesController#index<br/>@courses = Query.call(provider: partner,<br/>params: {accredited_provider:}).map(&:decorate)"]
  end

  %% ---------- Domain / query layer ----------
  subgraph DOM["Domain / query layer"]
    CL["Publish::CourseList<br/>facade · Enumerable · #groups · #any?"]
    Q["Publish::Courses::Query<br/>.call(provider:, params:)<br/>scope-chaining (like Find's Courses::Query)"]
    GRP["Publish::CourseList::Group (Data)<br/>accredited_provider_name, courses<br/>#self_accredited? · #heading"]
  end

  %% ---------- Database (one SQL query) ----------
  subgraph DB["One SQL query"]
    CRS[("course<br/>display cols: name, course_code, funding,<br/>qualification, study_mode, start_date,<br/>age_range_in_years, application_status")]
    AP[("provider AS accredited_provider<br/>LEFT JOIN on provider_code + recruitment_cycle_id")]
    ENR[("course_enrichment<br/>LEFT JOIN LATERAL — per-course aggregate")]
    COLS["computed columns<br/>group_name · content_status · has_unpublished_changes"]
  end

  CD["CourseDecorator<br/>display helpers (qualifications_summary, age_range, …)"]
  CT["Find::CycleTimetable<br/>current_year / previous_year (no DB)"]

  %% ---------- View components (Publish::Courses::*) ----------
  subgraph VIEW["View components"]
    LC["ListComponent<br/>iterates course_list.groups"]
    GC["GroupComponent<br/>&lt;section&gt; + accredited-provider heading"]
    TC["TableComponent<br/>&lt;table&gt; + column headers"]
    RC["RowComponent<br/>&lt;tr&gt;: name cell · info cell · status cell"]
    IC["InformationComponent<br/>funding / qualification / study type / start date"]
    STC["StatusTagComponent<br/>reads content_status + has_unpublished_changes cols<br/>+ application_status + cycle branch"]
  end

  %% ===== data build flow =====
  CC --> CL
  CL -->|"Query.call(provider:)"| Q
  Q --> CRS
  Q --> AP
  Q --> ENR
  ENR --> COLS
  AP --> COLS
  Q -->|"ordered rows: course + computed columns"| CL
  CL -->|"chunk_while by group_name<br/>+ map(&:decorate)"| GRP
  GRP --> CD

  %% ===== render flow =====
  CC -->|"render ListComponent(course_list:, provider:)"| LC
  LC -->|"each Group"| GC
  GC -->|"group.courses"| TC
  TC -->|"each course"| RC
  RC --> IC
  RC --> STC
  STC --> CT
  IC --> CD
  STC --> CD

  %% ===== training-partners reuse (flat list, no grouping) =====
  TPC --> Q
  TPC -->|"render TableComponent(courses:, provider:)"| TC

  %% ===== drift guard =====
  ADC["ApplicationDecorator#status_tag<br/>canonical Ruby (course show / support pages)"]
  STC -. "cross-check spec asserts identical output" .-> ADC
```

### Reading it as layers
1. **Controller** — builds the domain object and renders the top component. The main list groups; training-partners reuses the same `Query` + `TableComponent` as a flat list.
2. **Domain / query** — `CourseList` (facade) asks `Publish::Courses::Query` for the rows, then chunks them into `Group` value objects (decorated courses). `Query` is the only place that touches the DB.
3. **Database** — a single query: `course` (all display columns) + `LEFT JOIN` the accredited provider (heading) + `LEFT JOIN LATERAL` aggregate over `course_enrichment`, producing three computed columns (`group_name`, `content_status`, `has_unpublished_changes`). No enrichment/site rows are loaded as objects.
4. **View components** — `List → Group → Table → Row`, with each row rendering `InformationComponent` (display fields via the decorator) and `StatusTagComponent` (status from the computed columns + `Find::CycleTimetable`).
5. **Drift guard** — `StatusTagComponent` re-implements the tag from SQL columns; a cross-check spec asserts it renders identically to the canonical `ApplicationDecorator#status_tag`.

## Query courses List performance

Loop through each provider and mount the query:

```
Summary
  providers:    807
  slowest:      31.8ms
  median:       4.3ms
  total > 200ms: 0
```

Comparison with main branch:

```
main vs new (per-provider, ms)
                     main           new
  total           34819.7           3738.0
  mean               43.1           4.6
  median             29.6            4.2
  max               587.4            51.8

```
**overall speedup (total main / total new): 9.32x**

Script measure the above:

```ruby
# frozen_string_literal: true

# Compares the course-list query performance of `main` against the new
# Publish::Courses::Query, across every provider in the current recruitment
# cycle. Run with:
#
#   bundle exec rails runner script/benchmark_courses_query.rb
#
# For each provider it measures, end to end:
#
#   * main  - load the provider with main's eager-loads, group via
#             Courses::Fetch.by_accrediting_provider, then touch the per-course
#             view work (status_tag -> content_status -> ContentStatusService,
#             plus summary / name_and_code).
#   * new   - Publish::Courses::Query.call(provider:).to_a (status from SQL).
#
# Both blocks reload the provider from scratch so AR identity-map caching does
# not leak between the two paths.

require "benchmark"

cycle = RecruitmentCycle.current
provider_codes = cycle.providers.order(:provider_code).pluck(:provider_code)
total = provider_codes.size
puts "Comparing main vs Publish::Courses::Query for #{total} providers in cycle #{cycle.year}..."
puts

# Reproduces the controller#index + view cost on main.
def main_path(cycle, provider_code)
  provider = cycle.providers
                  .includes(courses: %i[site_statuses enrichments provider])
                  .find_by!(provider_code:)

  grouped = ::Courses::Fetch.by_accrediting_provider(provider)
  grouped.each_value do |courses|
    courses.each do |course|
      course.status_tag
      course.summary
      course.name_and_code
    end
  end
end

# The new read-model query; status/grouping/ordering come from SQL columns.
def new_path(cycle, provider_code)
  provider = cycle.providers.find_by!(provider_code:)
  Publish::Courses::Query.call(provider:).to_a
end

results = []

provider_codes.each_with_index do |provider_code, index|
  new_ms = Benchmark.realtime { new_path(cycle, provider_code) } * 1000.0
  main_ms = Benchmark.realtime { main_path(cycle, provider_code) } * 1000.0

  results << { provider_code:, main_ms:, new_ms:, delta_ms: main_ms - new_ms }

  print "\rProcessed #{index + 1}/#{total}" if ((index + 1) % 25).zero? || index + 1 == total
end
puts
puts

# --- Per-provider buckets for the NEW query ----------------------------------
buckets = {
  "> 1s" => results.select { |r| r[:new_ms] > 1000 },
  "500ms - 1s" => results.select { |r| r[:new_ms] > 500 && r[:new_ms] <= 1000 },
  "200ms - 500ms" => results.select { |r| r[:new_ms] > 200 && r[:new_ms] <= 500 },
  "50ms - 200ms" => results.select { |r| r[:new_ms] > 50 && r[:new_ms] <= 200 },
  "0ms - 50ms" => results.select { |r| r[:new_ms] <= 50 },
}

puts "Publish::Courses::Query timing buckets"
buckets.each do |label, rows|
  puts format("  %-15s %d", label, rows.size)
end

# --- Head-to-head comparison -------------------------------------------------
def stats(values)
  sorted = values.sort
  {
    total: values.sum,
    mean: values.sum / values.size.to_f,
    median: sorted[sorted.size / 2],
    max: sorted.last,
  }
end

main_stats = stats(results.map { |r| r[:main_ms] })
new_stats = stats(results.map { |r| r[:new_ms] })
faster = results.count { |r| r[:new_ms] < r[:main_ms] }
slower = results.count { |r| r[:new_ms] > r[:main_ms] }

puts
puts "main vs new (per-provider, ms)"
puts format("  %-10s %12s %12s", "", "main", "new")
puts format("  %-10s %12.1f %12.1f", "total", main_stats[:total], new_stats[:total])
puts format("  %-10s %12.1f %12.1f", "mean", main_stats[:mean], new_stats[:mean])
puts format("  %-10s %12.1f %12.1f", "median", main_stats[:median], new_stats[:median])
puts format("  %-10s %12.1f %12.1f", "max", main_stats[:max], new_stats[:max])
puts
speedup = main_stats[:total] / new_stats[:total]
puts format("  overall speedup (total main / total new): %.2fx", speedup)
puts format("  providers faster on new: %d   slower on new: %d", faster, slower)

# --- Biggest regressions / wins ----------------------------------------------
puts
puts "Top 10 providers where new is SLOWER than main"
results.select { |r| r[:delta_ms] < 0 }.sort_by { |r| r[:delta_ms] }.first(10).each do |r|
  puts format("  %-12s main %8.1fms  new %8.1fms  (+%.1fms)", r[:provider_code], r[:main_ms], r[:new_ms], -r[:delta_ms])
end

puts
puts "Top 10 providers where new is FASTER than main"
results.sort_by { |r| -r[:delta_ms] }.first(10).each do |r|
  puts format("  %-12s main %8.1fms  new %8.1fms  (-%.1fms)", r[:provider_code], r[:main_ms], r[:new_ms], r[:delta_ms])
end

```

## Provider::Courses::Query SQL explanation

The query starts from "this provider's courses" and bolts on a few things so the page can render each row — including its status — without loading anything else.

### The base
`provider.courses` → every (non-deleted) course belonging to this provider. Everything else hangs off that.

### Join 1 — find each course's accredited provider (for the heading)
```sql
LEFT OUTER JOIN provider accredited_provider
  ON accredited_provider.provider_code = course.accredited_provider_code
  AND accredited_provider.recruitment_cycle_id = :cycle_id
```
Each course stores a *code* pointing at the university/SCITT that accredits it. This join looks that provider up (matching the code, in the same recruitment year) so we can show its name. **LEFT** join = if there's no match, don't drop the course, just leave the name blank.

### Join 2 — summarise each course's enrichments in one shot
"Enrichments" are the versioned content/state of a course (draft, published, etc.). A course has a handful. This is a **LATERAL** join — "for each course, run this little sub-query" — producing one summary row per course:
```sql
SELECT
  COUNT(*)                                AS total,                 -- how many enrichments
  COUNT(*) FILTER (WHERE status = 1)      AS published_count,       -- how many are published
  COUNT(*) FILTER (WHERE status = 0)      AS draft_count,           -- how many are draft
  COUNT(*) FILTER (WHERE status = 3)      AS withdrawn_count,       -- how many withdrawn
  COUNT(*) FILTER (WHERE status IN (0,2)) AS draft_or_rolled_count, -- draft or rolled-over
  MAX(last_published_timestamp_utc)       AS max_published_at,      -- has it ever been published?
  (ARRAY_AGG(status ORDER BY created_at DESC, id DESC))[1]                   AS latest_status,
  (ARRAY_AGG(last_published_timestamp_utc ORDER BY created_at DESC, id DESC))[1] AS latest_published_at
```
- `FILTER (WHERE ...)` just means "count only the rows that match." (status is a number: draft=0, published=1, rolled_over=2, withdrawn=3.)
- `ARRAY_AGG(... ORDER BY created_at DESC, id DESC)[1]` means "line up this course's enrichments newest-first, grab the first one's value" — so `latest_status` is the newest enrichment's status and `latest_published_at` tells us if that newest one was ever published. A compact way to get "the latest enrichment" plus the counts in one pass.

After this join, every course carries: how many enrichments it has (by type), whether it was ever published, and what its newest enrichment looks like.

### The three computed columns

**1. `group_name`** — the heading the course sits under:
```sql
CASE
  WHEN no accredited provider, OR the accredited provider is this provider itself THEN NULL
  ELSE accredited_provider.provider_name
END
```
`NULL` means "self-accredited" → rendered without a heading. Otherwise it's the partner's name.

**2. `content_status`** — the course's state, decided by the *newest* enrichment first:
```
newest is rolled_over  → 'rolled_over'
newest is published    → 'published'
newest is withdrawn    → 'withdrawn'
no enrichments at all   → 'draft'
otherwise (newest is a draft) — but it was published before, or there's more than one enrichment
                        → 'published_with_unpublished_changes'
otherwise               → 'draft'
```
In words: if the latest version is rolled over / live / withdrawn, that's the status. If the latest is just a draft but the course has history (published before, or older versions exist), it's "published, with newer unpublished edits." A brand-new single draft is just "draft." This is a one-to-one translation of `Courses::ContentStatusService`.

**3. `has_unpublished_changes`** — true/false, drives the "\*" on the tag:
```
NOT ( "everything is already published" )
AND ( there's a published version AND also a draft/rolled-over one
      OR  it was published before AND the latest isn't withdrawn )
```
"everything is already published" = either exactly one enrichment and it's published, or none are draft/withdrawn. If so there's nothing pending → false. Otherwise it's true when there are genuinely unpublished edits on top of a published course. A direct port of `Course#has_unpublished_changes?`.

### The ordering
```sql
CASE WHEN self-accredited THEN 0 ELSE 1 END,  -- self-accredited group floats to the top
LOWER(accredited_provider.provider_name),     -- then partners A→Z, case-insensitive
course.name,                                   -- then by course name
course.course_code                             -- then by code (stable tiebreak)
```
This pre-sorts the rows so Ruby just "chunks" consecutive rows with the same `group_name` into groups — no sorting in Ruby.

### Why it's shaped this way
The page needs, per course: its display fields (already plain columns on `course`), its group heading, and its status. Two joins + three columns return all of that in one query. The whole list renders in ~4 queries with **zero** per-course enrichment or site lookups — the database calculates status once, in bulk, instead of Ruby doing it course-by-course.

> Note: `content_status` and `has_unpublished_changes` are a SQL port of the canonical Ruby (`Courses::ContentStatusService` / `Course#has_unpublished_changes?`), which remains the source of truth. A cross-check spec asserts the SQL and the Ruby always agree, so the two can't drift.


app/components/publish/courses/group_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<section data-qa="courses__table-section">
```
langesi: I believe it was discussed with Dan to stop using data-qa attributes before I joined, right? You want to reconsider using it?

Will only make this one comment about data-qa, but I see it a few times used in this PR.

app/components/publish/courses/status_tag_component.rb
```diff
@@ -0,0 +1,81 @@
...
+    # Renders a course's status tag from the read-model columns produced by
+    # Publish::Courses::Query (content_status + has_unpublished_changes), the
+    # course's application_status, and the recruitment-cycle branch — so the list
+    # needs no enrichment or site rows.
+    #
+    # The text/colour mapping mirrors ApplicationDecorator#status_tags*; a
+    # cross-check spec asserts this component renders identically to
+    # ApplicationDecorator#status_tag.
+    class StatusTagComponent < ApplicationComponent
+      VACANCIES = {
```
langesi: The constant naming here is a bit confusing to me. VACANCIES reads like it refers to the old vacancies feature, which made it look like legacy code when I read this first.

From the implementation, this is actually the status tag mapping used when a course is open for applications, with `NO_VACANCIES` and `SCHEDULED` being the other mappings. Could we rename these to make the intent clearer, for example `OPEN_APPLICATION_STATUS_TAGS`, `CLOSED_APPLICATION_STATUS_TAGS`, and `SCHEDULED_STATUS_TAGS`?

app/components/publish/courses/status_tag_component.rb
```diff
@@ -0,0 +1,81 @@
...
+      ).freeze
+
+      def initialize(course:, recruitment_cycle_year:, classes: [], html_attributes: {})
+        super(classes:, html_attributes:)
+        @course = course
+        @recruitment_cycle_year = recruitment_cycle_year
+      end
+
+      def call
+        rendered = helpers.govuk_tag(text: status[:text].html_safe, colour: status[:colour])
```
langesi: Not a real risk using html_safe here, but we should try to just avoid using html_safe

langesi (commented):
Great work.

Since we’ll be adding filters and likely changing the enrichment/course status soon, I’m happy with these changes. This gives us room to insert that logic later. Nice job documenting the SQL in the Query service as well.

One suggestion though: 

Consider collapsing some of the components on the index page. It feels a little too nested for a table.

Something like:

ListComponent = owns the list structure and grouping
TableComponent = owns rendering one table
No separate GroupComponent, RowComponent, or InformationComponent unless they are reused.

That would essentially remove:
app/components/publish/courses/group_component.rb
app/components/publish/courses/group_component.html.erb
app/components/publish/courses/row_component.rb
app/components/publish/courses/row_component.html.erb
app/components/publish/courses/information_component.rb
app/components/publish/courses/information_component.html.erb

And the new components I think are still easily readable in size

langesi (approved):
Looks great, well done!

PR #6177, opened 2026-06-16
Hide age range for secondary courses in the course list

tomas-stefano:
## Context

Stop showing the age range hint on **secondary** courses in the course list. Primary (and further education) courses keep it.

## Why

Age range was only ever meant for primary courses (original ticket: "add age range to primary courses"), but it was rendering on secondary courses too.


langesi (approved):
Thanks for the fix

PR #6178, opened 2026-06-16
Course list hide uniform course information

tomas-stefano:
## Context

Make the course list's **Course information** column adaptive: hide any value (funding, qualification, study type, start date) that is identical across **all** of a provider's courses. If every value is identical, drop the column heading entirely, leaving Course | Status.

## Why

Providers told us courses in their list "look exactly the same" — repeating the same values down every row makes courses hard to tell apart. Hiding uniform values surfaces only what actually distinguishes one course from another.

## Changes

- `Publish::CourseList#visible_course_information_fields` computes which fields vary across the whole list (all accredited-provider groups).
- `TableComponent` takes that field set and renders only the varying lines; drops the column heading/cell when none vary.
- Cells with 0–1 lines get a `--sparse` modifier (20px top/bottom padding); rows already align to the top.

### DIFF
--- a/app/services/publish/course_list.rb
+++ b/app/services/publish/course_list.rb
@@ -18,10 +18,28 @@ module Publish
       end
     end
 
+    # The course-information fields shown per course, in display order. Each maps
+    # to the raw value that drives its displayed text, so uniformity is compared
+    # on that value rather than the rendered label.
+    FIELDS = {
+      funding: ->(course) { course.funding },
+      qualification: ->(course) { course.qualification },
+      study_mode: ->(course) { course.study_mode },
+      start_date: ->(course) { course.start_date.presence&.to_date },
+    }.freeze
+
     delegate :any?, to: :groups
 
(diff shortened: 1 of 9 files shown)


app/services/publish/course_list.rb
```diff
@@ -38,6 +56,10 @@ def each(&)
 
     attr_reader :provider
 
+    def all_courses
+      @all_courses ||= groups.flat_map(&:courses)
```
inulty-dfe, line 60: Did we ask how it should behave when there are multiple providers on the page? Should the uniqueness be applied to all the page or each probider?

tomas-stefano: It should be scoped for all. So for example if the first group is all September 2027 start date, but the second group has at least one course that is a different start date like July 2027, then display the start date information on all course in the list, so for both groups of courses under the two accredited providers.

inulty-dfe (approved):

PR #6179, opened 2026-06-16
Unauthenticated visits to email alerts redirect to custom signin page

inulty-dfe:
## Context

  - When the user signs in they are then redirected to the email alerts page

## Checklist

### DIFF
--- a/app/controllers/find/authentication/sessions_controller.rb
+++ b/app/controllers/find/authentication/sessions_controller.rb
@@ -43,7 +43,7 @@ module Find
             redirect_to after_auth_find_candidate_saved_courses_path
           else
             flash[:success] = t(".sign_in")
-            redirect_to(session.delete("return_to_after_authenticating") || find_root_path, allow_remote_host: false)
+            redirect_to(request.env["omniauth.origin"] || session.delete("return_to_after_authenticating") || find_root_path, allow_remote_host: false)
           end
         else
           redirect_to find_root_path, flash: { warning: t(".authentication_error") }
--- a/app/controllers/find/authentication/sessions_controller.rb
+++ b/app/controllers/find/authentication/sessions_controller.rb
@@ -43,7 +43,7 @@ module Find
             redirect_to after_auth_find_candidate_saved_courses_path
           else
             flash[:success] = t(".sign_in")
-            redirect_to(request.env["omniauth.origin"] || session.delete("return_to_after_authenticating") || find_root_path, allow_remote_host: false)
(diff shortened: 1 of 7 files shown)


inulty-dfe:
[Handle link to email alerts from email body better than 404](https://trello.com/c/ekRqsdeS/1720-handle-link-to-email-alerts-from-email-body-better-than-404)

app/views/find/candidates/email_alerts/sign_in.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1 class="govuk-heading-l">
+  Sign in to view all your email alerts
+</h1>
+
+<%= button_to("/auth/find-developer?origin=#{find_candidate_email_alerts_path}", class: %w[govuk-button], method: :post) do
```
tomas-stefano: In production One Login is enabled and the path is /auth/one-login. This button will point at a  non-existent strategy in prod.

app/controllers/find/authentication/sessions_controller.rb
```diff
@@ -43,7 +43,7 @@ def callback
             redirect_to after_auth_find_candidate_saved_courses_path
           else
             flash[:success] = t(".sign_in")
-            redirect_to(session.delete("return_to_after_authenticating") || find_root_path, allow_remote_host: false)
+            redirect_to(request.env["omniauth.origin"] || session.delete("return_to_after_authenticating") || find_root_path, allow_remote_host: false)
```
tomas-stefano: Is it not `allow_other_host`?

tomas-stefano: It would be good to have a test that crafts this into a malicious url and make sure it doesn't redirect?

tomas-stefano: Asking above because if doesn't work, `the request.env["omniauth.origin"]` into redirect_to, and that value is attacker-controllable — it comes straight from the `?origin=`

app/views/find/candidates/email_alerts/sign_in.html.erb
inulty-dfe: Ah yeah, I forgot to change that. I want the ucd team to sign off on the behaviour before I clean up the code

inulty-dfe: Fixed

app/controllers/find/authentication/sessions_controller.rb
inulty-dfe: So, `redirect_to` does not allow cross origin redirects by default so this was just an explicit option. It is the incorrect option so I've changed it to be `allow_other_hosts: false`

tomas-stefano (approved):

PR #6182, opened 2026-06-16
Exempt salaried courses publish without schools

tomas-stefano:
## What

Wires the `publish_without_schools_allowed` flag into publish validation so support-approved **salaried** and **apprenticeship** courses can publish without any schools attached. The exemption bypasses *only* the missing-schools check — every other publish validation still runs.

The exemption applies when all hold: the new `course_school` model is in use (`course_publishing_uses_new_school_model` flag), the course is salaried or apprenticeship, and `publish_without_schools_allowed` is true. It never applies to fee-paying courses or legacy sites data.

## How it decides

```mermaid
flowchart TD
    A[Publish validation: schools present?] -->|Yes| P[Pass — schools attached]
    A -->|No| B{Exemption applies?}
    B -->|"flag on<br/>AND salary? or apprenticeship?<br/>AND publish_without_schools_allowed?"| P2[Pass — exempt]
    B -->|otherwise| F[Block — :sites blank / :enter_schools]

    subgraph Exemption[SchoolPresenceExemption.applies?]
      C1[course_publishing_uses_new_school_model active] --> C2[salary? or apprenticeship?]
      C2 --> C3[publish_without_schools_allowed?]
    end
```

## Changes

- New `Courses::PublishRules::SchoolPresenceExemption` — single source of truth for the exemption decision, sibling to `SchoolPresence`
- `CoursePublishableSchoolsPresenceValidator` — skip `:sites, :blank` when exempt
- `CoursePublishableSchoolsRolloverValidator` — skip the no-schools `:enter_schools` branch when exempt (the `:check_schools` re-confirmation path is untouched)

## Validation matrix

| Funding | Has course_schools? | Exemption | Can publish? |
|---|---|---|---|
| Salaried / Apprenticeship | Yes | any | Yes |
| Salaried / Apprenticeship | No | No | No |
| Salaried / Apprenticeship | No | Yes | Yes |
| Fee | Yes | any | Yes (exemption irrelevant) |
| Fee | No | any | No (exemption ignored) |

## Tests

- New rule spec across flag/funding combinations
- Presence + rollover validator specs extended for the exemption
- Full matrix driven through the real validator + rules with the flag active

## Notes

- Gated on the new-model feature flag (consistent with `SchoolPresence`); legacy sites never qualify
- Builds on the prior data-layer change that added the `publish_without_schools_allowed` column

### DIFF
--- a/app/services/courses/publish_rules/school_presence_exemption.rb
+++ b/app/services/courses/publish_rules/school_presence_exemption.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+# Single source of truth for "is this course exempt from needing a school
+# attached at publish time?". Support can approve individual salaried or
+# apprenticeship courses to publish without schools (candidates already have
+# their placement arranged) via the `publish_without_schools_allowed` flag.
+#
+# Only applies under the new Course::School data model (the
+# :course_publishing_uses_new_school_model flag), mirroring
+# Courses::PublishRules::SchoolPresence so the flag lives in one place. Never
+# applies to fee-paying courses or to legacy Site data.
+module Courses
+  module PublishRules
+    class SchoolPresenceExemption
+      def self.applies?(course)
+        FeatureFlag.active?(:course_publishing_uses_new_school_model) &&
+          course.publish_without_schools_allowed? &&
+          (course.salary? || course.apprenticeship?)
+      end
+    end
+  end
+end
--- a/app/validators/course_publishable_schools_presence_validator.rb
+++ b/app/validators/course_publishable_schools_presence_validator.rb
@@ -7,6 +7,7 @@
 class CoursePublishableSchoolsPresenceValidator < ActiveModel::Validator
   def validate(course)
     return if Courses::PublishRules::SchoolPresence.any?(course)
+    return if Courses::PublishRules::SchoolPresenceExemption.applies?(course)
 
     course.errors.add(:sites, :blank)
   end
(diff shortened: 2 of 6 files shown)


app/services/courses/publish_rules/school_presence_exemption.rb
```diff
@@ -0,0 +1,22 @@
...
+# apprenticeship courses to publish without schools (candidates already have
+# their placement arranged) via the `publish_without_schools_allowed` flag.
+#
+# Only applies under the new Course::School data model (the
+# :course_publishing_uses_new_school_model flag), mirroring
+# Courses::PublishRules::SchoolPresence so the flag lives in one place. Never
+# applies to fee-paying courses or to legacy Site data.
+module Courses
+  module PublishRules
+    class SchoolPresenceExemption
```
inulty-dfe, line 14: Is this a Policy?

tomas-stefano: Good question. It's policy-shaped (stateless, side-effect-free predicate), but not a Policy in our Pundit sense — Publish uses more policies for Pundit related objects, those are (user, record) authorization objects in app/policies called via authorize. This one takes only a course  and answers a domain question ("does this course qualify for the no-schools exemption"), and it's invoked from the publish validators, not from authorization.

I put it under Courses::PublishRules on purpose, as a sibling to SchoolPresence — the validators use the two together and that namespace is already our single source of truth for publish decisions. Happy to discuss this! This is a good discussion for tomorrow dev weekly 🫶

inulty-dfe (approved):

PR #6183, opened 2026-06-16
Add school experience wizard to Publish salaried/apprenticeship courses

inulty-dfe:
## Context

Add a SchoolExperience DfE Wizard to the course management side in Publish.

This is a slightly non-standard multistep form.  Two steps, first to determine if the course should communicate to the candidate that this course requires or highly recommends school experience. If the course does require experience then the provider must add content to explain the details of this experience.

On the first step, if the user answers "yes", we cannot store this value in the database becuase if they click cancel before completing the second step the database is left in an inconsistent state.

We use a cached backed respository to save "yes" and commit the whole form at the end. If the user answers "no" then we just need to store the "no" in the database directly.

We show the value on the course description page too.

These changes will be followed by the implementation in Find in another PR.

- Add a DfE Wizard with two repositories
  - one model backed repository for when a step needs to be stored in the database
  - one cached backed repository for when a step needs to be stored temporarily
- This feature is only available in production only if the course in question is in the 2027 cycle or later. This allows us to demonstrate the feature in QA and review apps for now.

## Guidance to review

I did not implement the missing details component on the course edit page because none of the other properties are using it and we should implement them all together

I have not added a "find preview" to the form page. The place this content is displayed is not standard in the way that the preview component currently renders. It is a details component in a two column div so we need feedback as to whether this is the correct content to show in the preview.

## Checklist

### DIFF
--- a/app/controllers/publish/courses/school_experience/experience_required_controller.rb
+++ b/app/controllers/publish/courses/school_experience/experience_required_controller.rb
@@ -1,0 +1,38 @@
+module Publish
+  module Courses
+    module SchoolExperience
+      class ExperienceRequiredController < SchoolExperienceController
+        def state_store
+          if submitted_answer
+            # A "yes" answer can't be persisted to the course until the content
+            # arrives on the experience_details step, so it's held in the cache.
+            SchoolExperienceWizard::StateStores::SchoolExperienceWizardStore.new(
+              repository: SchoolExperienceWizard::Repositories::SchoolExperienceCacheRepository.new(
+                provider_code: params[:provider_code],
+                recruitment_cycle_year: params[:recruitment_cycle_year],
+                course_code: params[:course_code],
+              ),
+            )
+          elsif submitted_answer == false || action_name == "new"
+            # A "no" answer is written to the course; the new action reads the
+            # current value from it to pre-select the answer.
+            SchoolExperienceWizard::StateStores::SchoolExperienceWizardStore.new(
+              repository: SchoolExperienceWizard::Repositories::SchoolExperienceRepository.new(
+                record: @course,
+              ),
+            )
+          else
+            # School experience is optional. With no answer there's nothing to
+            # save, so we don't load a repository and let the wizard route back
+            # to the course.
+            SchoolExperienceWizard::StateStores::SchoolExperienceWizardStore.new
+          end
+        end
+
+        def submitted_answer
+          ActiveModel::Type::Boolean.new.cast(params.dig(:experience_required, :experience_required))
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/publish/courses/school_experience/school_experience_controller.rb
+++ b/app/controllers/publish/courses/school_experience/school_experience_controller.rb
@@ -1,0 +1,64 @@
+module Publish
+  module Courses
+    module SchoolExperience
+      class SchoolExperienceController < ApplicationController
+        before_action :assign_course
+        before_action :assign_wizard
+
+        helper_method def current_step
+          @wizard.current_step
+        end
+
+        helper_method def current_step_name
+          controller_name.to_sym
+        end
+
+        def new
+          @wizard.current_step_valid? if params[:display_errors].present?
+        end
+
+        def create
+          if @wizard.save_current_step
+            add_flash_message if @wizard.next_step == :course_edit
+            redirect_to @wizard.next_step_path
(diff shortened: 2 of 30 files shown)


inulty-dfe:
[Add school experience to salaried and apprenticeship courses](https://trello.com/c/nZROxAGa/1707-add-school-experience-to-salaried-and-apprenticeship-courses)

app/wizards/school_experience_wizard/state_stores/school_experience_wizard_store.rb
```diff
@@ -0,0 +1,20 @@
...
+      def experience_is_required?
+        string_to_boolean(experience_required)
+      end
+
+      def string_to_boolean(value)
+        case value
+        when "yes" then true
+        when "no" then false
+        end
+      end
```
inulty-dfe: I've needed to use this more than once in this wizard and another time in the A Levels wizard. How can we extract this helper?

`ActiveRecord::Type::Boolean.new.cast` doesn't help us with yes/no strings

```
[1] pry(main)> ActiveRecord::Type::Boolean.new.cast("yes")
=> true
[2] pry(main)> ActiveRecord::Type::Boolean.new.cast("no")
=> true
```

app/controllers/publish/courses/school_experience/experience_required_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+          else
+            # School experience is optional. With no answer there's nothing to
+            # save, so we don't load a repository and let the wizard route back
+            # to the course.
+            SchoolExperienceWizard::StateStores::SchoolExperienceWizardStore.new
+          end
+        end
+
+        def submitted_answer
+          params.dig(:experience_required, :experience_required)
```
langesi: You can use 

```
ActiveModel::Type::Boolean.new.cast(
	params.dig(:experience_required, :experience_required)
)
```

If you set the value for the radio buttons to :true and :false rather than :yes and :no

Not that much prettier though in my view

app/decorators/course_decorator.rb
```diff
@@ -434,6 +434,18 @@ def show_degree_requirements_row?
     !teacher_degree_apprenticeship?
   end
 
+  def show_school_experience?
+    return false unless FeatureFlag.active?(:school_experience)
```
langesi: I'd be happy to remove the feature flag and just have it based on the cycle. 

One less thing to test and go wrong.

app/wizards/school_experience_wizard/state_stores/school_experience_wizard_store.rb
langesi: Set the value :true and :false for the radio buttons, that will still be a string, then you can use ActiveRecord::Type::Boolean.new.cast

Another way is to use enums rather than booleans.

app/controllers/publish/courses/school_experience/school_experience_controller.rb
```diff
@@ -0,0 +1,66 @@
+module Publish
+  module Courses
+    module SchoolExperience
+      class SchoolExperienceController < ApplicationController
+        before_action :assign_course
+        before_action :assign_wizard
+
+        def edit; end
```
langesi: Is this one used?

langesi (commented):
I see validation here on the field, I don't see validations in the model. Can you confirm that I see that right?

If that is the case, pls add validations.

Happy with the PR, I do suggest removing the feature flag though, I don't see a need for it. The Recruitment Cycle is essentially our feature flag.

app/wizards/school_experience_wizard/state_stores/school_experience_wizard_store.rb
inulty-dfe: Good thinking :brain:

app/decorators/course_decorator.rb
inulty-dfe: I added the feature flag before i decided to run rollover in the review app for testing

inulty-dfe:
@langesi thanks for your great comments. I think I've addressed them all.

1. Pass true/false as the value options to the radio tags.
2. Added validations for the school experience
3. Added See what you wrote last cycle

langesi (approved):
Love it, well done!

inulty-dfe:
Content does not want See what you wrote last cycle on this attribute. I'll remove it.

PR #6188, opened 2026-06-19
Downcase visa sponsorship filter in email alert title

inulty-dfe:
## Context

Some issues with email alerts feature from the bug party

#### Capitalise email alert summary card title when the search has no subjects but uses location

image

#### Filter out irrelevant params when counting how many filters are being used in the search

image

#### Downcase visa sponsorship filter in email alert title

image

#### Lowercase non-proper noun subjects in create alert success notice 

image

## Checklist

### DIFF
--- a/app/components/find/courses/search_title_component.rb
+++ b/app/components/find/courses/search_title_component.rb
@@ -71,7 +71,7 @@ module Find
       end
 
       def visa_title
-        I18n.t("find.courses.search_title.visa_sponsorship")
+        I18n.t("find.courses.search_title.visa_sponsorship").downcase
       end
 
       def apprenticeship_title
--- a/app/components/find/courses/search_title_component.rb
+++ b/app/components/find/courses/search_title_component.rb
@@ -25,10 +25,13 @@ module Find
 
         named = @subjects.count.between?(1, 2)
 
-        if location?
-          key = named ? "subjects_with_location" : "no_subjects_with_location"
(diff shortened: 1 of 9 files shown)


inulty-dfe:
[Bug: Various email alerts content fixes](https://trello.com/c/Mbv9Z1RG/1761-bug-various-email-alerts-content-fixes)

inulty-dfe:
[Bug: Check page says 3 filters are applied when only two are applied](https://trello.com/c/bCAaiuSR/1760-bug-check-page-says-3-filters-are-applied-when-only-two-are-applied)

app/components/find/courses/search_title_component.rb
```diff
@@ -24,10 +32,13 @@ def title_text
         return fallback_title if @subjects.empty? && no_location?
 
         named = @subjects.count.between?(1, 2)
+        @subjects = @subjects.map { |subject| SubjectHelper.subject_name_in_sentence(subject) }
```
tomas-stefano, line 35: title_text mutates `@subjects (@subjects = @subjects.map { ... subject_name_in_sentence ... })`. It's idempotent so there's no bug, but reassigning instance state inside a method that reads like a pure accessor (and which heading/call   both invoke) is a smell. A local variable (named_subjects = @subjects.map {...}) would be cleaner and avoids any future foot-gun.

tomas-stefano (approved):
Small minor comment. I will leave to you to address it.

PR #6189, opened 2026-06-22
Add to support UI - publish without schools checkbox

tomas-stefano:
## Context

Some providers are permitted to publish courses before they have schools attached. This adds a support-only toggle so we can grant that exemption on a per-course basis, rather than blocking publication for these providers.

- Add a `publish_without_schools_allowed` checkbox to the support "Edit course" page, shown only for salaried/apprenticeship courses.
- Permit the new field in `Support::CoursesController` and the `Support::EditCourseForm`.
- Guard the flag in the form so it can only ever be set on salaried courses, even if a value is submitted for an ineligible course.
- Carry the flag over during course rollover and register it with DfE Analytics (column/migration already on main).

## Guidance to review

- Edit a salaried/apprenticeship course in support, tick the box, and confirm the value persists.
- Confirm the checkbox is hidden for non-salaried courses and that the flag cannot be set on them.

### DIFF
--- a/app/controllers/support/courses_controller.rb
+++ b/app/controllers/support/courses_controller.rb
@@ -46,7 +46,8 @@ module Support
                                    :is_send,
                                    :can_sponsor_student_visa,
                                    :can_sponsor_skilled_worker_visa,
-                                   :accredited_provider_code],
+                                   :accredited_provider_code,
+                                   :publish_without_schools_allowed],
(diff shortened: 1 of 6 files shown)


tomas-stefano:
https://trello.com/c/nCfJtWZn/1728-no-school-30-add-support-only-ui-for-publishing-salaried-courses-without-schools

gms-gs (approved):
LGTM!

PR #6191, opened 2026-06-23
Add check answers page & Save Course

gms-gs:
## Context

This PR introduces and completes the new wizard-based “Check your answers” flow for course creation, then wires the final submit action so clicking `Add course` actually creates a course (reusing legacy creation logic).  
The potentially surprising part is that wizard state keys differ from legacy `Courses::CreationService` inputs, so a mapper was added to translate and normalize data safely before persistence.

- Added the wizard `check_answers` step and routing:
  - Added `check_answers` node/step and `start_date -> check_answers -> courses_index` flow.
  - Added review callbacks for change-link return behavior.
- Implemented `CourseWizard::Review` presenter and check-answers UI:
  - Added row formatting/labels for subjects, funding, study pattern, schools/study sites, visa fields, age range, etc.
  - Mirrored legacy wording/behavior for conditional labels and display rules.
- Implemented create-on-submit from wizard check answers:
  - `Publish::CourseWizardsController` now creates a course when submitting `check_answers`.
  - Added `Courses::WizardParamsMapper` to translate wizard state keys to legacy creation params for `Courses::CreationService`.
  - Clears wizard state after successful create and redirects to courses index with success flash.
- Added/updated tests:
  - Added mapper specs covering translation permutations.
  - Updated start-date system journey to assert check-answers flow and successful creation.
  - Added/updated wizard review and navigation specs to cover the new behavior.
- Moved controller success flash strings to locale keys.

## Guidance to review

- **Flow review**
  1. Run wizard journey to `check_answers`.
  2. Click `Add course`.
  3. Confirm course is created and user is redirected to courses index.
  4. Check data is correcr
- **Mapping review**
  - Review [`app/services/courses/wizard_params_mapper.rb`](app/services/courses/wizard_params_mapper.rb) for key translation correctness:
    - subject IDs
    - funding/study mode
    - sites/study sites
    - visa deadline multipart params
- **Parity review**
  - Compare check-answers row behavior with legacy confirmation page for:
    - school labels (Employing vs Placement)
    - study site CTA behavior
    - skilled worker row visibility/default text
- **Suggested test commands**
  - `bundle exec rspec spec/services/courses/wizard_params_mapper_spec.rb`
  - `bundle exec rspec spec/system/publish/courses/add_course_wizard/start_date_spec.rb`
  - `bundle exec rspec spec/wizards/add_course_wizard/wizard/next_step_spec.rb`
  - `bundle exec rspec spec/wizards/add_course_wizard/review_spec.rb`

## Checklist


gms-gs:
[Implement Check your answers and save course data ](https://trello.com/c/f7mnPG1B/1771-implement-check-your-answers-and-save-course-data)

tomas-stefano:
This is functionally correct and well-tested (115 unit/wizard + 4 system specs green), but `CourseWizard::Review` is a
  312-line god class doing six jobs, and it duplicates state-interpretation that `WizardParamsMapper` also does. That's
  the same shape as the legacy `Courses::EditOptions` god module this wizard is meant to replace — let's not let it
  back in.

  ### Root cause — three interpreters of one state

  | "What is this course?" | Today (duplicated) | Target |
  | --- | --- | --- |
  | funding / age-range / study-mode / subjects / specialisms | `WizardParamsMapper` **and** `Review#format_value` |
  `Draft` (once) |
  | row order + visibility | `Review#course_details` and the graph | the graph (`flow_path`) |
  | record lookups (Subject / Site / Provider) | inside `Review` | `Draft` |
  | HTML + change-links | inside `Review` | the ViewComponent |

  The TDA rule alone is written **6×** (`mapper#mapped_funding`, `review.rb:44`, `:72`, `:283`, `:287`, `:21-22`). One
  rule, six edit sites = Single Choice Principle broken.

  ### Current objects

  ```mermaid
  graph TD
    V["_check_answers.html.erb"] --> R["CourseWizard::Review<br/>312 lines · 6 responsibilities"]
    R --> SS["state_store (interpret rules)"]
    R --> DB["Subject / Site / Provider (DB lookups)"]
    R --> FMT["format_value case (i18n)"]
    R --> HTML["html_safe / link_to (markup)"]
    R --> RT["route_strategy.resolve ×8 (change links)"]
    R --> ORD["course_details (hand-ordered rows)"]
    CTRL["Controller#persist_course"] --> M["WizardParamsMapper"]
    M -->|re-interprets SAME state| SS
    class R,M bad;
    classDef bad fill:#fdd,stroke:#c00,color:#000;
  ```

  ### Target objects

  Patterns: Value Object · Anti-Corruption Layer · Strategy · ViewComponent · Single Source of Truth

  ```mermaid
  classDiagram
    class Draft {
      +funding() String
      +study_modes() Array
      +age_range() String
      +subject_ids() Array
      +subjects() Array~Subject~
      +schools() Array~Site~
      +study_sites() Array~Site~
      +accrediting_provider() Provider
      +visa_deadline() VisaDeadline
      +tda?() Bool
      +employment_based?() Bool
    }
    class WizardParamsSerializer {
      +call() Params
    }
    class CheckAnswers_SummaryComponent {
      +rows() Array~Row~
      +call() Html
    }
    class Formatter {
      <<interface>>
      +call(value, draft) String
    }
    note for Formatter "FundingFormatter, AgeRangeFormatter, StudyPatternFormatter, SubjectsFormatter,
  VisaDeadlineFormatter, BooleanFormatter"
    CourseWizardStateStore --> Draft : interpreted once
    Draft --> WizardParamsSerializer : shape only
    Draft --> CheckAnswers_SummaryComponent : resolved values
    CheckAnswers_SummaryComponent --> Formatter : delegates display
    CheckAnswers_SummaryComponent --> Row : builds
  ```

  ### Responsibility split after redesign (SRP)

  | Object | Single responsibility | House precedent |
  | --- | --- | --- |
  | `CourseWizard::Draft` (PORO) | interpret state → resolved domain values + records |
  `app/lib/required_qualifications_summary.rb` |
  | `Courses::WizardParamsSerializer` (was Mapper) | translate Draft → CreationService param shape | ServicePattern
  services |
  | `Publish::CheckAnswers::SummaryComponent` | build/render `govuk_summary_list` rows |
  `app/components/courses/summary_card_component.rb` |
  | `Formatters::*` | i18n/display for one value type (Strategy) | direct `I18n.t(...label)` convention |
  | `_check_answers.html.erb` | render the component only | — |

  This deletes the `format_value` case, the duplicated predicates, `present_value?`, the inline `html_safe`, and the 8×
  `route_strategy.resolve`.

  Existing `wizard_params_mapper_spec` / `review_spec` are characterisation tests, so the refactor is safe — split them  into `draft_spec` + `summary_component_spec` + per-formatter specs.

  `dfe-wizard` agrees with this direction: it exposes `flow_path` / `saved_steps` (hydrated Step objects) precisely so  the review is derived from the flow, not a hand-maintained second list. (The `@current_step_params[:return_to_review]`  read is fine — it matches the gem's README example.)

app/wizards/course_wizard/review.rb
```diff
@@ -0,0 +1,312 @@
+# frozen_string_literal: true
+
+class CourseWizard
+  class Review
```
tomas-stefano: This class has six responsibilities (state  interpretation, DB lookups, i18n formatting,  change-link routing, HTML generation, row ordering) —  the same god-class shape as the legacy  `Courses::EditOptions` we're retiring. Split per the  design comment: `CourseWizard::Draft` (interpretation),
  `Publish::CheckAnswers::SummaryComponent`
  (rows/render).

tomas-stefano: Also `dfe-wizard` can help on check your answers: https://github.com/DFE-Digital/dfe-wizard#check-your-answers

app/services/courses/wizard_params_mapper.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+module Courses
+  class WizardParamsMapper
```
tomas-stefano: Keep this as the anti-corruption layer to  CreationService's param shape, but rename to  `WizardParamsSerializer` and strip all*interpretation*  — it should read resolved values from `Draft`, not  recompute them.

app/services/courses/wizard_params_mapper.rb
```diff
@@ -0,0 +1,124 @@
...
+
+      [
+        state_store.primary_master_subject_id,
+        state_store.secondary_master_subject_id,
+        state_store.subordinate_subject_id,
+        *selected_specialism_subject_ids,
+      ].compact_blank.uniq
+    end
+
+    def selected_specialism_subject_ids
```
tomas-stefano: `selected_specialism_subject_ids`, `mapped_age_range_in_years`,`mapped_funding`,  `mapped_study_mode` are the duplicated rules — move to  `Draft`; here just map keys/shape.

app/services/courses/wizard_params_mapper.rb
```diff
@@ -0,0 +1,124 @@
...
+      return {} if [year, month, day].any?(&:blank?)
+
+      {
+        "visa_sponsorship_application_deadline_at(1i)" => year.to_s,
+        "visa_sponsorship_application_deadline_at(2i)" => month.to_s,
+        "visa_sponsorship_application_deadline_at(3i)" => day.to_s,
+      }
+    end
+
+    def visa_deadline_parts(deadline)
```
tomas-stefano: Same Date/DateParts/Hash handling as
  `Review#format_visa_deadline`. Extract a tiny
  `VisaDeadline` value object on `Draft` that both the  serializer and the formatter consume.

app/controllers/publish/course_wizards_controller.rb
```diff
@@ -42,6 +44,23 @@ def update
...
+      @course = ::Courses::CreationService.call(course_params:, provider:, next_available_course_code: true)
+
+      if @course.save
+        flash[:success_with_body] = {
+          title: I18n.t("publish.course_wizards.flash.success_with_body.title"),
+          body: I18n.t("publish.course_wizards.flash.success_with_body.body"),
+        }
+        @wizard.clear_state
+        redirect_to publish_provider_recruitment_cycle_courses_path(provider.provider_code, provider.recruitment_cycle_year)
+      else
```
tomas-stefano: The failed-save path (copy `@course.errors`  onto `current_step`, render 422) has no test. Add a system spec asserting an invalid course
  re-renders check-answers with the error visible (wiring  is correct — `show.html.erb` renders
  `form.govuk_error_summary`).

app/controllers/publish/course_wizards_controller.rb
```diff
@@ -34,6 +34,8 @@ def create
 
     def update
       if @wizard.save_current_step
+        return persist_course if current_step_name == :check_answers
```
tomas-stefano: Branching on the magic name `:check_answers`
  might be OK for one terminal step; but I'd prefer a  step property (`wizard.final_step?`)

app/wizards/course_wizard.rb
```diff
@@ -207,4 +214,35 @@ def accreditation
...
+  def clear_specialism_answers_when_subjects_change
+    return unless current_step_name == :secondary_subjects
+
+    reset_attributes = {}
+    reset_attributes[:campaign_name] = nil unless physics_specialisms?
+    reset_attributes[:language_ids] = nil unless modern_languages_specialisms?
+    reset_attributes[:design_technology_ids] = nil unless design_technology_specialisms?
+    return if reset_attributes.empty?
+
+    state_store.write(**reset_attributes)
```
tomas-stefano, line 254: State mutation inside a `before_next_step`
  nav callback (`next_step` can run more than once per  request). Safe only because it's idempotent — I'd think we need to discuss this, a step operator (https://github.com/DFE-Digital/dfe-wizard#step-operators) OR move the clear into the  controller after `save_current_step`. Let's discuss this.

spec/wizards/add_course_wizard/review_spec.rb
```diff
@@ -0,0 +1,306 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe CourseWizard::Review do
```
tomas-stefano: `state_store` is a non-verifying  `double("CourseWizard::StateStore", …)`, so these tests  stay green even if a state-store method is renamed/removed — the bug surfaces only in production.  House style (spec/components/find/courses/sumary_compon
  ent/view_spec.rb) favours real records. After the
  `Draft` extraction, retarget: `draft_spec` exercises  rules against a real/verified store;   `summary_component_spec` renders with real Subject/Site  records.

spec/wizards/add_course_wizard/review_spec.rb
```diff
@@ -0,0 +1,306 @@
...
+        modern_languages_specialisms?: modern_languages_specialisms,
+        design_technology_specialisms?: design_technology_specialisms,
+      )
+    end
+
+    let(:modern_languages_specialisms) { false }
+    let(:design_technology_specialisms) { false }
+
+    it "does not include stale specialism subject ids when specialisms are no longer selected" do
+      expect(review.send(:selected_subject_ids)).to eq(%w[100 200])
```
tomas-stefano: These reach into private methods via `.send`
  (selected_subject_ids, school_label_with_plural,
  show_skilled_worker_visa_row?, row_for_study_sites),  coupling the suite to internals — a sign the public  surface isn't what's tested. Post-redesign these become  public `Draft` methods (#subject_ids,  #employment_based?) and `SummaryComponent` rows, tested  directly.

app/wizards/course_wizard/review.rb
inulty-dfe: I agree, this class needs attention. I feel like the formatting of each step could be dleegated to the step itself? Something to investigate

spec/wizards/add_course_wizard/wizard/next_step_spec.rb
```diff
@@ -119,6 +119,38 @@
         expect(wizard).to have_next_step(:modern_languages_specialisms)
       end
     end
+
```
tomas-stefano, line 122: `handle_back_to_review` (course_wizard.rb:237) has zero coverage. Add back-link scenarios: from the edited  origin step with `return_to_review` set → `previous_step` is `:check_answers`; without it → normal graph back-nav. Use  `have_previous_step`.

spec/wizards/add_course_wizard/wizard/next_step_spec.rb
```diff
@@ -119,6 +119,38 @@
...
+    context "when returning from check answers and subjects now require a specialism page" do
```
tomas-stefano, line 123: Good negative case, but the positive path is missing: after editing a step with `return_to_review` set and  a complete, valid flow, `next_step` must be `:check_answers`. Add it — it's the core CYA behaviour and currently
  unverified.

spec/system/publish/courses/add_course_wizard/start_date_spec.rb
```diff
@@ -8,11 +8,13 @@
     given_i_am_authenticated_as_a_provider_user(cycle_year: Date.current.year)
   end
 
-  scenario "choosing a start date and continues to courses index page" do
+  scenario "choosing a start date and adds the course from check answers" do
```
tomas-stefano, line 11: No coverage of (a) the invalid-save path (→ 422, error visible on check-answers), and (b) a "Change" link round-trip (edit a row → back on check answers with the new value). Add both.

tomas-stefano:
Some ideas we can discuss:

```mermaid

  classDiagram
    class CourseDraft {
      +funding() String
      +study_modes() Array
      +age_range() String
      +subjects() Array~Subject~
      +schools() Array~Site~
      +study_sites() Array~Site~
      +accrediting_provider() Provider
      +visa_deadline() VisaDeadline
      +tda?() Bool
      +employment_based?() Bool
    }
    class Step {
      <<dfe-wizard hydrated step>>
      +summary_rows(draft) Array~Row~
    }
    class WizardParamsSerializer {
      +call() Params
    }
    class CheckAnswers_SummaryComponent {
      +rows() Array~Row~
      +call() Html
    }
    CourseWizardStateStore --> CourseDraft : interpreted once
    CourseDraft --> WizardParamsSerializer : shape only
    CourseDraft --> Step : resolved values
    Step --> CheckAnswers_SummaryComponent : summary_rows(draft)
    CheckAnswers_SummaryComponent --> Row : assembles + renders
```

What dfe-wizard actually gives you (real, today):
  - wizard.flow_path(:check_answers) → ordered step ids
  - wizard.step(id) / wizard.saved_steps → hydrated step objects (with their typed attributes)

  What you'd add (new):
  - a #summary_rows(draft) method on each reviewable step, returning [Row] — so the component does
  wizard.step(id).summary_rows(draft) instead of the central format_value case.

Or maybe also explore https://github.com/DFE-Digital/dfe-wizard#check-your-answers

tomas-stefano:
The redesign already landed the hard part (one `Draft`, one funding/TDA rule, formatters extracted). This is about the *next* layer: 

`SummaryComponent` is still ~316 lines because it owns an 80-line `rows_for_step` dispatch `case` **and** a hand-maintained  `DEFAULT_STEP_ORDER` that duplicates the graph in `course_wizard.rb`.

The gem already hands us the spine to delete both.

  #### What `dfe-wizard` gives us for free

  `flow_steps` returns the **hydrated Step instances, in flow order, branching-aware** (it respects `skip_when:` and the  conditional edges). So:

  ```ruby
  def rows
    wizard.flow_steps.flat_map { |step| step.review_rows(draft) }.compact # review_rows doesn't exist yet.
  end
  ```

  This alone removes:
  - `DEFAULT_STEP_ORDER` (the duplicated graph) + `step_ids` / `flow_path_step_ids` / `extract_step_id`
  - `subject_step_for_level` — the graph **already** branches `level → primary_subjects | secondary_subjects`
  (`course_wizard.rb:52-59`), so `flow_path` only ever contains the right one. The component is re-deriving a decision
  the graph made.

  > Worth saying explicitly: the gem deliberately stops at **navigation** (the `CheckYourAnswers` behaviour we're
  already wired into). It will not render rows for us — presentation is ours. So this is our call to make, not a gem
  feature to wait for.

  #### The idea A: steps declare *intent*, the component renders

  A Step is a form/validation object. The risk of "just put the row on the step" is that we move the god class instead
  of deleting it — now i18n + HTML + record lookups are smeared across 18 fat steps. So the step declares **data, never 
  markup**; the `Draft` keeps resolving values/records; the formatters keep owning display; the component just lays out.

  ```ruby
  # concern, mixed into steps
  module CourseWizard::Reviewable
    extend ActiveSupport::Concern
    class_methods do
      def review(&blk) = @review_spec = blk
      def review_spec = @review_spec
    end

    # default: interpret the declaration; cross-cutting steps override this
    def review_rows(draft)
      self.class.review_spec ? RowSet.build(self.class.review_spec, self, draft) : []
    end
  end

  # a step — review intent lives next to its attributes & validations
  class CourseWizard::Steps::FundingType
    include DfE::Wizard::Step
    include CourseWizard::Reviewable

    attribute :funding_type, :string
    # ...validations...

    review do |r|
      r.row value: ->(d) { d.funding },          # Draft resolves the TDA rule, once
            label: :funding_type, formatter: :funding,
            changeable: ->(d) { !d.tda? }, show_when_blank: true
    end
  end
  ```

The RowSet:
```ruby
class CourseWizard
    module Reviewable
      # Collects the `r.row(...)` declarations from a step's `review` block and
      # resolves each into a RowSpec.
      #
      # A RowSpec is pure data: the *resolved value* + the *intent* (label key,
      # formatter key, blank-handling, changeability). Turning a RowSpec into a
      # rendered row — label lookup, formatter, change link, visibility — is the
      # SummaryComponent's job. RowSet deliberately knows nothing about i18n,
      # routes, or markup.
      class RowSet
        RowSpec = Struct.new(
          :step_id, :label_key, :label_options, :value, :formatter,
          :show_when_blank, :changeable,
          keyword_init: true,
        ) do
          def changeable? = changeable
        end

        def self.build(spec, step, draft)
          collector = new(step, draft)
          spec.call(collector)
          collector.rows
        end

        attr_reader :rows

        def initialize(step, draft)
          @step  = step
          @draft = draft
          @rows  = []
        end

        # Called as `r.row(...)` inside a step's `review` block.
        # Multiple calls = multiple rows (e.g. :level emits level + SEND).
        def row(label:, value: nil, formatter: nil, label_options: {},
                show_when_blank: false, changeable: true)
          @rows << RowSpec.new(
            step_id:         @step.step_id,
            label_key:       label,
            label_options:   resolve(label_options),
            value:           resolve(value),
            formatter:,
            show_when_blank:,
            changeable:      resolve(changeable),
          )
        end

      private

        # A declared value/changeable/label_options can be:
        #   Proc   -> called with the draft (the common case: ->(d) { d.funding })
        #   Symbol -> read the step's own attribute of that name (:funding_type)
        #   else   -> used verbatim (literals, true/false, hashes)
        def resolve(value)
          case value
          when Proc   then value.call(@draft)
          when Symbol then @step.public_send(value)
          else value
          end
        end
      end
    end
  end
```

  The component glue that consumes it

  RowSpec → Row (your existing struct) is where formatter + change-link + visibility live — note this is the same
  visibility logic that's in saved_value_row today, just moved one layer out:

```ruby
  def rows
    @rows ||= wizard.flow_steps
                    .flat_map { |step| step.review_rows(draft) }
                    .filter_map { |spec| render_row(spec) }
  end

  def render_row(spec)
    return unless wizard.saved?(spec.step_id) || spec.show_when_blank

    value = format_value(spec)
    row = Row.new(
      label:       t_label(spec.label_key, **spec.label_options),
      value:,
      change_path: spec.changeable? ? change_path_for(spec.step_id) : nil,
    )
    return row if spec.show_when_blank
    return if !present_value?(spec.value) && !present_value?(value)

    row
  end

  def format_value(spec)
    return spec.value unless spec.formatter

    formatter(spec.formatter).call(spec.value, draft)   # the registry lookup
  end
```

  How the two awkward cases land

  level → two rows falls out of multiple r.row calls, no override needed:

  review do |r|
    r.row label: :level, value: ->(d) { d.level }, formatter: :level,    changeable: false
    r.row label: :send,  value: ->(d) { d.is_send }, formatter: :send,   changeable: false
  end

  study_sites empty-state CTA is the one case that wants a link as the value — that's HTML, so it must not live in the
  step. Either (a) the step declares a normal row and a StudySitesFormatter renders the CTA when blank, or (b) that
  single step overrides review_rows and the component keeps a small special-case. I'd push it into the formatter (a) so
  the rule "steps emit data, not markup" stays absolute.

  Things to decide

  - Where present_value? lives — I kept it in the component since it also checks the formatted output. Fine as-is.
  - label_options (the count:/prefix: for pluralised labels like schools) — I made it resolvable too, since the schools
  label depends on employment_based?. If you'd rather keep labels dumb, drop that and let a formatter own pluralisation.
  - Eager vs lazy value resolution — I resolve eagerly in row(...) because draft is already in hand at build time. Keeps
  RowSpec a plain data bag with no captured closures, which makes step specs trivial to assert on
  (expect(step.review_rows(draft)).to include(have_attributes(value: "apprenticeship")))

  #### Layering this buys us

  | Layer | Responsibility |
  | --- | --- |
  | Step (`review do …`) | *what* to show — intent only |
  | `Draft` | the resolved value + record lookups |
  | `Formatters::*` | *how* it reads (i18n / html) |
  | `SummaryComponent` | iterate `flow_steps`, lay out rows |
  | gem | order + branching + navigation |

  - **Locality / Open–Closed:** adding a step = edit one file (the step), not the component's `case` + a constant.
  - **Ordering is free** and always correct (it's the graph).
  - Cross-cutting rows (`level` emits a second SEND row; `study_sites` empty-state CTA) become a one-method
  `review_rows` override on *that* step — not a branch in a shared dispatch.

tomas-stefano:
The wizard doesn't define a steps_operator at all, so every step runs the gem's default [Validate, Persist] pipeline
  (persist = write step data to the state store). The actual course creation is then hand-rolled in the controller's
  persist_course — WizardParamsSerializer → CreationService → course.save → manual error-mapping. That's exactly the
  work the gem's Step Operators are meant to own: the README lists "persistence, deletions, API calls, email" as the use  case, and an operation is the gem's unit for "do a thing when this step is saved."

  So the final step (:check_answers) should run a custom CreateCourse operation instead of leaving creation in the
  controller.

  How it looks

  An operation is just initialize(repository:, step:) + execute returning { success:, errors: }, and save_current_step
  aggregates the result and surfaces errors onto the step (that's how Validate works). The step carries wizard, so the
  operation can reach the provider and serializer:
```ruby
  # app/wizards/course_wizard/operations/create_course.rb
  class CourseWizard
    module Operations
      class CreateCourse
        def initialize(repository:, step:)
          @step   = step
          @wizard = step.wizard
        end

        def execute
          params = ::Courses::WizardParamsSerializer.call(wizard: @wizard)
          course = ::Courses::CreationService.call(
            course_params: params, provider: @wizard.provider, next_available_course_code: true
          )

          if course.save
            { success: true }
          else
            course.errors.full_messages.each { |m| @step.errors.add(:base, m) }
            { success: false, errors: @step.errors }
          end
        end

        def rollback; end
      end
    end
  end
```

  Wire it to the final step only (skip Validate/Persist — CheckAnswers has no attributes to persist):

```ruby
  # in CourseWizard
  def steps_operator
    DfE::Wizard::StepsOperator::Builder.draw(wizard: self, callable: state_store) do |b|
      b.on_step(:check_answers, use: [Operations::CreateCourse])
    end
  end

  The controller then collapses to its actual job — flow control + flash/redirect:

  def update
    if @wizard.save_current_step
      @wizard.clear_stale_specialism_answers
      return complete_course if @wizard.final_step?

      redirect_to @wizard.next_step_path
    else
      render :show, status: :unprocessable_entity
    end
  end

  def complete_course
    flash[:success_with_body] = { title: …, body: … }
    @wizard.clear_state
    redirect_to publish_provider_recruitment_cycle_courses_path(provider.provider_code, provider.recruitment_cycle_year)
  end
```

  persist_course disappears: serialization, creation, and error-mapping move into the operation; the controller no
  longer knows CreationService exists.

  Honest caveats

  - Worth doing, but it's a refactor of working code — the current flow is correct and tested. The win is layering
  (creation logic leaves the controller, lands on the gem's intended extension point), not behaviour.
  - @course isn't needed by the controller — the success path redirects to the courses index using provider, not the
  course, so nothing is lost. If you later want to redirect to the new course, expose it via the result hash or stash it
  on the wizard.
  - Don't drop validation of the whole flow. Model-level validation still runs inside CreationService/course.save. If
  you also want to guard against reaching check-answers with an invalid earlier step, that's
  valid_path_to?(:check_answers) — a separate, complementary check.
  - clear_state stays in the controller (it's tied to the success redirect). The gem also offers
  mark_completed/completed? if you want a persisted completion flag instead — optional.

tomas-stefano:
Minor

  - No spec for CreateCourse. System specs cover the happy path, but the failure branch (invalid course → errors added
  to step → save_current_step returns false → render :show) — the risky bit — is untested. Add a focused operation spec.
  - _repository = repository in CreateCourse#initialize (:7) is a no-op local to dodge a lint warning. Either use it or
  drop the line and accept the unused kwarg.
  - i18n leaking into a step: Schools' label_options does I18n.t(...schools_prefix...) inside the step. That's the
  "steps emit data, not presentation" line being crossed for the prefix. Pass employment_based? as data and resolve the
  prefix in the label/formatter layer.
  - respond_to?(:review_rows) guard (summary_component.rb:18) is now dead — every step includes Reviewable. Drop it.
  - StudySitesFormatter re-implements the route_strategy.resolve(... return_to_review ...) that change_path_for already
  owns — small duplication.

tomas-stefano:
Strong progress overall — the `Review` god class is gone, the operation seam is right, and the component is 85 lines. 

But five of the six concerns raised below are **the same root cause wearing different hats**: *one concept is interpreted/translated in several places, and "list of every field" keeps getting rebuilt centrally.* 

We deleted `DEFAULT_STEP_ORDER` and grew a formatter registry; we deleted `Review#format_value` and grew per-field formatter classes; we centralised the Draft and then re-scattered TDA rules across step overrides. Let's name the pattern and fix the pattern, not the symptoms.

```mermaid
flowchart TD
    State[(state_store)]
    State --> Draft
    State -->|interpreted again| F12[12 Formatter classes]
    Draft --> RS["#review_steps<br/>ad-hoc 'append if missing'"]
    Draft --> IQ["Qualifications#review_rows<br/>injects funding + study_pattern"]
    Draft --> IS["StudySites#review_rows<br/>injects skilled_worker"]
    RS --> C[SummaryComponent]
    IQ --> C
    IS --> C
    C --> REG["#formatter<br/>12-entry registry, built per call"]
    Draft --> SER[WizardParamsSerializer] --> P[/params hash/] --> CS[CreationService] -->|re-inflates| Course
    classDef bad fill:#fdd,stroke:#c00,color:#000;
    class RS,IQ,IS,REG,P bad;
```

Every red node is "the same answer, restated." Target:

```mermaid
flowchart TD
    State[(state_store)] --> Draft
    Draft -->|one ordering policy| RP["review_steps =<br/>flow_steps + tda defaults"]
    RP --> S["each Step.review_rows<br/>(its OWN declaration only)"]
    S --> C["SummaryComponent<br/>(render only)"]
    S -.row carries.-> F3["3 parameterised formatters<br/>+ 2 genuine one-offs"]
    Draft -->|typed input| CS[CreationService] --> Course
```

---

### 1 + 3 — Formatters bloat & the `#formatter` registry (one fix)

`formatters.rb` has 12 classes expressing **3 shapes**. `LevelFormatter`/`QualificationFormatter`/`FundingFormatter` are "scoped enum → i18n label." `SendFormatter`/`SponsorFormatter`/`YesNoFormatter` are "boolean → one of two keys." `SubjectsFormatter`/`SitesFormatter` are "list → joined." A formatter should be parameterised by **data**, not specialised by **subclass**:

```ruby
module Publish::CheckAnswers::Formatters
  # 1) scoped enum -> i18n label
  class Enum
    def initialize(scope) = @scope = scope
    def call(value, _draft, _view)
      return if value.blank?
      I18n.t("#{@scope}.#{value}.label", default: value.to_s.humanize)
    end
  end

  # 2) boolean -> one of two i18n keys
  class Bool
    def initialize(yes:, no:) = (@yes, @no = yes, no)
    def call(value, _draft, view)
      view.t(ActiveModel::Type::Boolean.new.cast(value) ? @yes : @no)
    end
  end

  # 3) list of names -> joined (optionally <br/>)
  class List
    def initialize(separator: ", ") = @separator = separator
    def call(value, _draft, view)
      return if value.blank?
      @separator == :br ? view.safe_join(value, view.tag.br) : value.join(@separator)
    end
  end
end
```

That replaces 9 of the 12. Only the genuine one-offs survive — `AgeRange` (the "other" range), `StudyPattern` (the full+part-time collapse), `VisaDeadline`, and the `StudySites` CTA — and they now stand out *as* the exceptions, which is the point.

Then **delete the registry entirely** by carrying the formatter on the row (the step already declares it), and pass the view at call time so formatters stay context-free configured values:

```ruby
# step declaration
review do |r|
  r.row :funding_type,
        value:  ->(d) { d.funding },
        format: Formatters::Enum.new("course_wizard.steps.funding_type.options"),
        changeable: ->(d) { !d.tda? }
end

# component — the 12-entry hash and #formatter(key) are gone
def format_value(spec)
  return spec.value unless spec.format
  spec.format.call(spec.value, draft, self)
end
```

No symbol→class lookup, no "instantiate 12 to use 1," no central list every new field must thread through. The registry was `DEFAULT_STEP_ORDER` reincarnated.

### 2 — `#review_steps` is the worst offender, and it's a symptom

`#review_steps` exists *only* to undo what `flow_steps` correctly did — re-appending `study_sites`/`accredited_provider` with hardcoded `unless steps.any? { … }` checks. It's fighting the gem. And it coexists with the **duplicate** injections in `Qualifications#review_rows` / `StudySites#review_rows`, where funding/study_pattern/skilled_worker rows are hand-rebuilt — the very rows those steps already declare. One rule, three edit sites, again.

Fix: **one ordering policy; every step renders its own declaration.** The TDA "show the skipped step read-only" rule lives in a single place; the steps keep their single `changeable: ->(d){ !d.tda? }`:

```ruby
# CourseWizard (or Draft) — the ONLY place that knows TDA shows skipped steps
TDA_REVIEW_DEFAULTS = %i[funding_type study_pattern skilled_worker_visa].freeze

def review_steps
  extra = draft.tda? ? TDA_REVIEW_DEFAULTS.map { |id| step(id) } : []
  (flow_steps + extra).uniq(&:step_id)
end
```

```ruby
# DELETE both overrides — the steps already declare these rows correctly:
# Qualifications#review_rows  ❌
# StudySites#review_rows      ❌
```

Now `FundingType.review_rows` is the *single* definition of the funding row; in a TDA flow it's reached via `review_steps` and renders read-only because of its own `changeable` lambda. No duplication, no ad-hoc appends, and the two skipped-step bugs (missing skilled-worker row, unreachable study-site CTA) disappear because inclusion is principled rather than coincidental. (Bonus: `study_sites`/`accredited_provider` get the same treatment — add them to the policy instead of the `unless` hack.)

### 4 — i18n scattered with no boundary

You fought hard to keep `I18n.t` *out* of steps (the whole `resolved_label_options` dance), yet the controller hardcodes `I18n.t("...flash...")`, and `FundingFormatter` mixes `context.t` (line 38) and `I18n.t` (line 40) **in the same method**. That inconsistency is the tell that no rule was agreed. State it:

| String kind | Owner | Mechanism |
|---|---|---|
| Field **value** labels | Formatters | `I18n.t` (one style) |
| Row **labels** | Component | `t_label` (already single) |
| Flash / page chrome | Controller/View | `t(…)` helper, never bare `I18n.t` |

```ruby
# controller — use the helper, and stop carrying copy in the controller body
flash[:success_with_body] = t("publish.course_wizards.flash.success_with_body").to_h
```

```ruby
# FundingFormatter — pick one style
I18n.t("course_wizard.steps.check_answers.answers.salary_apprenticeship") if value == "apprenticeship"
```

### 5 — `VisaDeadline` is necessary, but its shape is a confession

A value object for a date is fine. A value object whose `parse_date` must distinguish **three** input shapes (`to_date`-able, `DateParts`, and a string-or-symbol `Hash`) is telling you **the data isn't normalised upstream**. Worse, it reaches *up* into `CourseWizard::Steps::VisaSponsorshipApplicationDeadlineAt::DateParts` (line 47) — a `Draft` value object depending on a wizard *step's* inner class is an inverted dependency.

Fix the cause: persist **one** canonical representation, and make `DateParts` the shared value type (it already exists) with a real `#to_date`. Then `VisaDeadline` collapses to a wrapper with no `rescue`, no `is_a?` ladder:

```ruby
class VisaDeadline
  def self.wrap(value) = value.is_a?(DateParts) ? value : DateParts.parse(value)
  # ...thin accessors over a guaranteed DateParts...
end
```

If it has to keep guessing between three shapes, that's the smell — and it points at the writer, not the reader.

### 6 — `CreateCourse` → `CreationService` isn't redundant; the *double translation* is

Wrapping `CreationService` is correct — it's reusable domain logic and the operation is the right gem seam. The thing my instinct is catching is the round-trip: **`Draft` (rich) → `WizardParamsSerializer` → stringly param hash → `CreationService` re-inflates it back into domain concepts.** That's the "three interpreters of one state" theme resurfacing at the persistence boundary.

Let `CreationService` accept the typed object so there's **one** translation, not two:

```ruby
# operation stays thin; no params hop
def execute
  course = Courses::CreationService.call(draft: @wizard.draft, next_available_course_code: true)
  return { success: true } if course.persisted?

  course.errors.full_messages.each { |m| @step.errors.add(:base, m) }
  { success: false, errors: @step.errors }
end
```

`WizardParamsSerializer` then either disappears or becomes an explicit anti-corruption adapter *inside* `CreationService` for the legacy attribute names — but only one of the two layers should own that shape, not both.

---

### Bottom line
Six issues, one problem: **single source of truth keeps getting duplicated.** The cheapest high-value moves, in order: (2) delete the two `review_rows` injections + the `review_steps` hack for one ordering policy, (1/3) collapse formatters and delete the registry, (4) write down the i18n boundary. (5) and (6) are smaller but both point at the same "normalise once, interpret once" principle.

tomas-stefano (approved):
This is getting more and more complicated.

Let's merge since is behind a feature flag and discuss about the problem and the challenges on this PR.

@gms-gs @inulty-dfe @langesi

PR #6192, opened 2026-06-23
Find - no employing schools for salaried courses

tomas-stefano:
## Context 

Following recent changes in Publish, some providers can now publish salaried or apprenticeship courses without attaching a placement (employing) school.

We need to reflect this change on Find so that users are not shown misleading information when no employing school is attached.


tomas-stefano:
https://trello.com/c/itxfZDa3/1735-no-school-update-courses-on-find-when-they-have-no-placement-schools

app/components/courses/summary_card_component.html.erb
```diff
@@ -4,10 +4,14 @@
       <% summary_list.with_row do |row| %>
         <% row.with_key(text: t(".location_key.#{course.funding}")) %>
         <% row.with_value do %>
-          <%= location_value %>
+          <% if no_employing_schools? %>
+            <%= no_employing_schools_text %>
```
langesi: It's only used once, right? So how about just having the locale here rather than a method?

app/components/courses/summary_card_component.rb
```diff
@@ -54,6 +54,16 @@ def candidate_accounts_enabled?
       @candidate_accounts_enabled ||= FeatureFlag.active?(:candidate_accounts)
     end
 
+    def no_employing_schools?
+      return @no_employing_schools unless @no_employing_schools.nil?
+
+      @no_employing_schools = course.without_employing_school?
```
inulty-dfe: It's probably not a good idea to set instance variables in predicate methods

app/services/courses/query.rb
```diff
@@ -502,6 +496,45 @@ def radius_in_miles
...
   private
 
+    # A course is findable when it has at least one running, published site
+    # (the legacy proxy for "published"). Salaried/apprenticeship courses that
+    # support has approved to publish without an employing school have no such
+    # site, so — only under the new school model flag — they are also findable
+    # when they are exempt and their latest enrichment is published (mirroring
+    # Course#is_published? so a returned course never 404s on the show page).
+    def findable_courses_sql
+      sql = +<<~SQL
```
inulty-dfe: Never seen this before `var = +<<HEREDOC`

app/components/courses/summary_card_component.html.erb
```diff
@@ -4,10 +4,14 @@
...
+          <% else %>
+            <%= location_value %>
```
inulty-dfe: one is `_text`, the other is `_value`. How can we keep these variable names consistent?

langesi, line 10: Does this show now when no school is attached and the flag is turned off?

What I'm afraid of is that this feature gets enabled and disabled, what will happen then?

Easy solution, get rid of the feature flag, only support can enable this feature anyways. I think the feature flag is more of a liability in this feature than a help

app/services/courses/query.rb
```diff
@@ -502,6 +496,45 @@ def radius_in_miles
 
   private
 
+    # A course is findable when it has at least one running, published site
```
inulty-dfe: The concept of findable is going to change with the course school models. How will this need to change then?

app/components/courses/summary_card_component.html.erb
tomas-stefano: Removed the method; the ERB now calls t(".location_value.no_employing_schools") directly.

app/components/courses/summary_card_component.html.erb
tomas-stefano: Resolved as a side effect of the above — there's no longer a _text method

app/services/courses/query.rb
tomas-stefano: That was the unary + to get an unfrozen string under frozen_string_literal: true so sql << could append. Refactored away entirely — findable_courses_sql now builds an array of conditions and join(" OR ")s them, with the two SQL  fragments extracted into running_published_site_sql and exempt_school_less_course_sql

app/components/courses/summary_card_component.rb
tomas-stefano: Dropped the memoization in both SummaryCardComponent#no_employing_schools? and  TrainingLocations::View#no_employing_schools? — they now just delegate to course.without_employing_school?. :schools  is already preloaded in Courses::Query#optimisation_scope, so there's no N+1 cost to losing the memo.

app/services/courses/query.rb
tomas-stefano: When the new school model fully lands and the flag is retired, findable_courses_sql collapses to its two branches unconditionally (the if FeatureFlag.active? wrapper just goes away — that's why I extracted the second branch into its own method). The legacy running_published_site_sql (running+published course_site as a proxy for "published") stays until sites are replaced by the Course::School model; at that point it gets swapped for a direct check on published enrichment + school presence, mirroring Course#is_published?. The two-method structure makes that future swap a one-line change rather than a re-edit of one big heredoc.

app/components/courses/summary_card_component.html.erb
tomas-stefano: If the flag is turn off the record won't appear in the collection from the DB. Sorry I don't understand what you mean drop the feature flag. Where?

langesi: Sorry, I got that wrong. I thought this was displayed on the course show page. So disregard my concern there.

That said, my comment about the feature flag still stands. I don’t think we really need it, since only Support can enable courses without attached courses, and they’ve been instructed not to use this feature until the new cycle.

If you'd remove it there is one less thing to test and think of

app/models/course.rb
```diff
@@ -911,6 +911,15 @@ def offers_sponsorship?
     !no_visa_sponsorship?
   end
 
+  # A salaried/apprenticeship course that support has approved to publish
+  # without an employing school (candidates already have a placement arranged)
+  # and which has no school attached. Find shows "No employing schools listed"
+  # for these instead of the misleading "nearest employing school" prompts.
```
inulty-dfe: I suggest we remove references to exact text here, it's a libability/dependency that can get out of sync

spec/services/courses/query/base_scope_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      create(:course, :with_salary, :published, name: "Salary One").tap { |c| create(:site_status, :findable, course: c) }
+      create(:course, :with_salary, :published, name: "Salary Two").tap { |c| create(:site_status, :findable, course: c) }
+      create(:course, :fee, :published, name: "Fee One").tap { |c| create(:site_status, :findable, course: c) }
+    end
+
+    it "counts only the courses matching the filter" do
+      query = described_class.new(params:)
+      query.call
+
+      expect(query.count).to eq(2)
```
inulty-dfe: This test doesn't specify which course should be omitted from the results

spec/services/courses/query/school_less_courses_findability_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+RSpec.describe Courses::Query do # rubocop:disable RSpec/SpecFilePathFormat
+  subject(:results) { described_class.call(params: {}) }
+
+  # Salaried/apprenticeship courses that support has approved to publish without
+  # an employing school have no running, published site. They are findable on
+  # exactly the same terms as every other course: they must be published.
+  # Neither school presence nor the publish_without_schools_allowed flag affects
+  # findability — only publication does.
+  describe "school-less course findability" do
```
inulty-dfe: Could we rename this file `course_without_school_findability`  or `no_school_course`

The hyphen in "School-less course" doesn't translate well to snakecase

inulty-dfe (approved):
Couple of small comments on tests but this is great. Nice work

app/components/courses/summary_card_component.html.erb
tomas-stefano: I removed, but because of that I'd had to verify. The concept of findable doesn't exist anymore. As long the course is published (with school or no school) will appear.

spec/services/courses/query/school_less_courses_findability_spec.rb
tomas-stefano: Done

spec/services/courses/query/base_scope_spec.rb
tomas-stefano: Done

app/models/course.rb
tomas-stefano: Done

inulty-dfe:
There's nothing as far as i can see in Publish to communicate that a course is exempt from attaching schools.

tomas-stefano:
> There's nothing as far as i can see in Publish to communicate that a course is exempt from attaching schools.

From the designs there is no communication about it @kelliedesigner

PR #6199, opened 2026-06-25
Add SalaryFees json column to the course enrichment

inulty-dfe:

For the 2027 recruitment cycle, we are changing the provider course inputs for describing the fees associated with their salaried courses. Until 2026, the course attribute was titled "Salary" but we are renaming it to "Fees".

## Guidance to review

1. This change is only happening for 2027+ courses so I've run rollover in the review app
2. This is a provider in 2027 with salaried courses
3. https://publish-review-6199.test.teacherservices.cloud/publish/organisations/1CN/2027/courses

|2026|2027|
|---|---|
|image|image|

## Checklist

### DIFF
--- a/app/controllers/publish/courses/salary_fees_controller.rb
+++ b/app/controllers/publish/courses/salary_fees_controller.rb
@@ -49,7 +49,7 @@ module Publish
         @previous_cycle_enrichment ||= course.recruitment_cycle.previous&.providers&.find_by(
           provider_code: @provider.provider_code,
         )&.courses&.find_by(
-          course_code: course_enrichment.course.course_code,
+          course_code: course.course_code,
         )&.enrichments&.where(
           status: "published",
         )&.last
--- a/app/controllers/publish/courses/salary_fees_controller.rb
+++ b/app/controllers/publish/courses/salary_fees_controller.rb
@@ -14,7 +14,7 @@ module Publish
         @course_salary_fees_form = ::Publish::CourseSalaryFeesForm.new(course_enrichment, params: form_params)
 
         if @course_salary_fees_form.save!
-          course_updated_message I18n.t("publish.providers.course_salary_fees.edit.course_salary_fee")
+          course_updated_message t(".course_salary_fee")
 
           redirect_to publish_provider_recruitment_cycle_course_path(
             provider.provider_code,
--- a/app/controllers/publish/courses/salary_fees_controller.rb
+++ b/app/controllers/publish/courses/salary_fees_controller.rb
@@ -3,11 +3,15 @@
 module Publish
   module Courses
     class SalaryFeesController < ApplicationController
+      include CopyCourseContent
+
       before_action :previous_cycle_enrichment, only: :edit
 
       def edit
         @course_salary_fees_form = ::Publish::CourseSalaryFeesForm.new(course_enrichment)
         @course_salary_fees_form.valid? if show_errors_on_publish?
+        @copied_fields = copy_content_check(::Courses::Copy::SALARY_FEES_FIELDS)
+        @copied_fields_values = copied_fields_values if @copied_fields.present?
       end
 
       def update
@@ -22,6 +26,7 @@ module Publish
             course.course_code,
           )
         else
+          fetch_course_list_to_copy_from
           render :edit
         end
       end
--- a/app/controllers/publish/courses/salary_fees_controller.rb
+++ b/app/controllers/publish/courses/salary_fees_controller.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module Publish
+  module Courses
+    class SalaryFeesController < ApplicationController
+      before_action :previous_cycle_enrichment, only: :edit
+
+      def edit
+        @course_salary_fees_form = ::Publish::CourseSalaryFeesForm.new(course_enrichment)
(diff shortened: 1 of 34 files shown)


inulty-dfe:
[Update salary textbox in Publish](https://trello.com/c/C79YZJzK/1083-update-salary-textbox-in-publish)

app/controllers/publish/courses/salary_fees_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+    private
+
+      def form_params
+        params.require(:publish_course_salary_fees_form).permit(:salary_fee_details)
+      end
+
+      def course
+        @course ||= provider.courses.find_by(course_code: params[:code])
```
tomas-stefano: Can we use `find_by!`? 
Every sibling controller (`BaseFundingTypeController#course`, `Fields::BaseController#course_to_authorise`) uses `find_by!`. With a bad/stale course code this returns nil, then course.enrichments in course_enrichment blows up as a NoMethodError → HTTP `500` instead of a clean `404`.

app/views/find/courses/_salary.html.erb
```diff
@@ -1,3 +1,9 @@
-<% if course.salary_details.present? %>
-  <%= markdown(course.salary_details) %>
+<% if course.recruitment_cycle_after?(2026) %>
+  <% if course.salary_fee_details.present? %>
```
tomas-stefano, line 2: The if reads the decorator method `(enrichment_attribute → enrichments.most_recent.first)` but the render reaches through course.latest_enrichment. 

They happen to resolve to the same row today, so it's not a live crash, but it's fragile or not concise.

Also can we collapse the nested ifs into `course.recruitment_cycle_after?(2026) && course.salary_fee_details.present?`

app/controllers/publish/courses/salary_fees_controller.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Publish
+  module Courses
+    class SalaryFeesController < ApplicationController
```
tomas-stefano, line 5: `SalaryFeesController < ApplicationController` reimplements course and course_enrichment from scratch. 

The controller it mirrors, `SalaryController`, extends `BaseFundingTypeController`, which already provides a decorated, find_by!-safe
  course, an identical course_enrichment, plus SuccessMessage/GotoPreview.

Is inheriting from `BaseFundingTypeController` might be a better choice?

tomas-stefano:
`config/routes/publish.rb:~397` (recruitment_cycle_year.to_i > 2026), 
`_description_content.html.erb` (recruitment_cycle_after?(2026)), 
`_salary.html.erb:2`, plus the after? model method. 

The route uses > 2026 while views use after?(2026) — consistent today but trivially drifts. Improvement: one predicate, e.g. Course#salary_fees_enabled? / RecruitmentCycle::SALARY_FEES_FROM_YEAR?

tomas-stefano:
Also `/salary-fees` is gated to `> 2026`, but `/salary` remains routable for all years, so a 2027 course still has a live
  (unlinked) editor writing salary_details that Find will never display. Is this intentional?

app/views/find/courses/_salary.html.erb
inulty-dfe: Can you explain what you mean about the different enrichment scopes and suggest a change please?

I think the nested ifs are easier to read. the other is too long.

app/controllers/publish/courses/salary_fees_controller.rb
inulty-dfe: I can see why you might suggest this, but inheriting from `BaseFundingTypeController` creates an "is a" relationship and CourseFees is not a FundingType.

That controller implements methods that are in no way relevant to the new controller and it's a text book case of poor inheritance to reuse that class in this case

inulty-dfe: I'll review the similarities betwen the two classes and tighten up the implementation. I'll add the go to preview concern for example

inulty-dfe: Ah you've made me realise I haven't implemented the missingContentcomponent in the preview :clap:

app/views/find/courses/_salary.html.erb
tomas-stefano:
```
if course.salary_fee_details.present?
```

and 

```
<%= markdown(course.latest_enrichment.salary_fee_details) %>
```

In a short what I meant, they return the same value in most case but don't check the same thing on the if condition.

inulty-dfe: Ok got it, fixed

inulty-dfe:
> `config/routes/publish.rb:~397` (recruitment_cycle_year.to_i > 2026), `_description_content.html.erb` (recruitment_cycle_after?(2026)), `_salary.html.erb:2`, plus the after? model method.
> 
> The route uses > 2026 while views use after?(2026) — consistent today but trivially drifts. Improvement: one predicate, e.g. Course#salary_fees_enabled? / RecruitmentCycle::SALARY_FEES_FROM_YEAR?

I'll clean this up later with all the other "after_" methods.

I've left the routes as is. I've rebased a branch with updated `after?(year)` method on RecruitmentCycle and updated this branch

tomas-stefano (approved):

PR #6201, opened 2026-06-30
Update bursaries link to use stable base link

gms-gs:
## Context

The "Learn more about the bursaries and scholarships" link on the secondary subjects step pointed at a hardcoded gov.uk academic year (`2023-to-2024`). That meant providers adding courses in a different recruitment cycle could be sent to the wrong ITT funding page.

The URL path is stable; only the academic year segment changes with the cycle. This change builds that segment from the recruitment cycle year so the link stays correct as cycles roll over.

- Add `RecruitmentCycleHelper#bursary_and_scholarship_link_url` to build the gov.uk URL from the recruitment cycle year (e.g. `2026` → `2026-to-2027`)
- Move the URL template to `publish.itt_funding.bursary_and_scholarship_link_url` in `config/locales/en/publish.yml`
- Update the add course wizard secondary subjects step to use the helper with `current_step.wizard.recruitment_cycle_year`
- Update the legacy secondary subjects form to use the same helper with `course.recruitment_cycle_year` (replacing the outdated `2023-to-2024` URL)
- Remove the hardcoded URL from `course_wizard.yml`
- Add helper and system specs to verify the correct URL is rendered

## Guidance to review

- On the add course wizard secondary subjects step, confirm the bursaries link uses the academic year for the cycle you are working in
- Check the legacy secondary subjects form (`/publish/.../courses/.../subjects`) uses the same URL pattern
- Review `RecruitmentCycleHelper#bursary_and_scholarship_link_url` — the academic year is `{recruitment_cycle_year}-to-{recruitment_cycle_year + 1}`; confirm that matches gov.uk’s URL convention

## Checklist

### DIFF
--- a/app/helpers/recruitment_cycle_helper.rb
+++ b/app/helpers/recruitment_cycle_helper.rb
@@ -52,4 +52,14 @@ module RecruitmentCycleHelper
   def next_cycle_provider(provider)
     RecruitmentCycle.next.providers.find_by(provider_code: provider.provider_code)
   end
+
+  def bursary_and_scholarship_link_url(recruitment_cycle_year)
+    start_year = recruitment_cycle_year.to_i
+    academic_year_range = "#{start_year}-to-#{start_year + 1}"
+
+    t(
+      "publish.itt_funding.bursary_and_scholarship_link_url",
+      academic_year_range:,
+    )
+  end
 end
--- a/app/helpers/recruitment_cycle_helper.rb
+++ b/app/helpers/recruitment_cycle_helper.rb
@@ -52,14 +52,4 @@ module RecruitmentCycleHelper
   def next_cycle_provider(provider)
     RecruitmentCycle.next.providers.find_by(provider_code: provider.provider_code)
   end
-
-  def bursary_and_scholarship_link_url(recruitment_cycle_year)
-    start_year = recruitment_cycle_year.to_i
-    academic_year_range = "#{start_year}-to-#{start_year + 1}"
-
-    t(
-      "publish.itt_funding.bursary_and_scholarship_link_url",
-      academic_year_range:,
-    )
-  end
 end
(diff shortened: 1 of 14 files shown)


gms-gs:
[Bug: Link to bursaries is pointing to the 23/24 cycle](https://trello.com/c/E1TBrbIN/1782-bug-link-to-bursaries-is-pointing-to-the-23-24-cycle)

inulty-dfe:
I'm not sure we can guarantee that those dynamic links will work when they are activated in this code.

gms-gs:
> I'm not sure we can guarantee that those dynamic links will work when they are activated in this code.

Shall I just hardcode it for now then?

tomas-stefano (approved):

tomas-stefano:
Some options:

1.  https://www.gov.uk/government/publications/funding-initial-teacher-training-itt is a permanent landing page that always links to the latest published year's guidance (confirmed live, currently surfacing 2026-to-2027). It can never 404 and needs zero maintenance. Cost: we lose the #postgraduate-bursaries-and-scholarships deep-anchor (the hub is a listing page), so the user clicks through one more level. The dynamic helper and its specs get deleted.
2. keep the deep anchor, but clamp the year to the latest published cycle. Keep the helper but cap the funding year at Find::CycleTimetable.current_year instead of using the wizard's cycle year. A newly-opened 2027 cycle would still point at the guaranteed-live 2026-to-2027 page (with the anchor) until DfE publishes the new one. Preserves anchor UX and avoids 404s; downside is being intentionally one year behind for a few weeks each rollover, plus we keep a helper to maintain.

I'd prefer the option 1.

gms-gs:
> Some options:
> 
> 1. https://www.gov.uk/government/publications/funding-initial-teacher-training-itt is a permanent landing page that always links to the latest published year's guidance (confirmed live, currently surfacing 2026-to-2027). It can never 404 and needs zero maintenance. Cost: we lose the #postgraduate-bursaries-and-scholarships deep-anchor (the hub is a listing page), so the user clicks through one more level. The dynamic helper and its specs get deleted.
> 2. keep the deep anchor, but clamp the year to the latest published cycle. Keep the helper but cap the funding year at Find::CycleTimetable.current_year instead of using the wizard's cycle year. A newly-opened 2027 cycle would still point at the guaranteed-live 2026-to-2027 page (with the anchor) until DfE publishes the new one. Preserves anchor UX and avoids 404s; downside is being intentionally one year behind for a few weeks each rollover, plus we keep a helper to maintain.
> 
> I'd prefer the option 1.

Will go with option 1.

PR #6205, opened 2026-07-03
Add School experience to find and add dynamic preview in publish

inulty-dfe:
## Context

Add the dynamic preview to the publish step for adding school experience.
Render the value in Find for courses in recruitment cycles 2027 and beyond.
The whole row should be hidden in the course summary on results page if the school experience is not required.

Extra: guard the /salary endpoint for salary fees feature

* Rename `PublishHelper` to `MarkdownHelper` - "`PublishHelper`" is confusing
* Render the school experience information on the Find course page
* Reduce input rows from 8 to 5
* Add dynamic preview to School experience input in Publish

## Add to Find

|course summary|course show|
|---|---|
|image|image|

## Dynamic preview in Publish
image

## Checklist

### DIFF
--- a/app/components/find/courses/entry_requirements_component/view.html.erb
+++ b/app/components/find/courses/entry_requirements_component/view.html.erb
@@ -3,6 +3,15 @@
   <div data-qa="course__required_qualifications">
 
     <%= govuk_summary_list(actions: false) do |summary_list| %>
+      <% if course.show_school_experience? %>
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: school_experience_title) %>
+          <% row.with_value do %>
+            <%= govuk_details(summary_text: t(".yes_school_experience")) { markdown(course.school_experience_required_content) } %>
+          <% end %>
+        <% end %>
+      <% end %>
+
       <% summary_list.with_row do |row| %>
         <% row.with_key(text: qualification_required) %>
         <% row.with_value do %>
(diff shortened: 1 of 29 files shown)


inulty-dfe:
[Add live preview to school experience textarea page](https://trello.com/c/vhRJbbsF/1766-add-live-preview-to-school-experience-textarea-page)

app/components/find/courses/entry_requirements_component/view.html.erb
```diff
@@ -3,6 +3,17 @@
   <div data-qa="course__required_qualifications">
 
     <%= govuk_summary_list(actions: false) do |summary_list| %>
+      <% if course.recruitment_cycle_after?(2026) && course.show_school_experience? %>
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: school_experience_title) %>
+          <% row.with_value do %>
+            <% if course.school_experience_required %>
```
tomas-stefano: The row is added whenever the outer condition passes, but the value is only populated when  school_experience_required is truthy. For the common case — a post-2026 course where school
  experience isn't required (school_experience_required is false/nil) — this renders a "School experience" key with an empty `<dd>`. That's a visible blank row on the live course page. Either move course.school_experience_required into the outer if (skip the row entirely when not required). Worth confirming the intended behaviour with the ticket.

inulty-dfe: Good catch, thanks

langesi (approved):
Good work, Thanks for renaming the markdown helper!

I tested it also manually in the review app by directly going to the school-experience path in publish

inulty-dfe:
Approved by UX and content

PR #6206, opened 2026-07-03
Add migration to add UUID to provider_school table

inulty-dfe:
## Context

This PR adds a default value to the database column. For our migration, we will want to set the uuid to a specific value. For the ProviderSchools that already exist, we will update the uuid that will be generated during this schema migration.

When creating new records during the datamigration, we pass the uuid explicitly to create a new record with proscribed uuid value.

When new provider schools are created in BAU, the uuid is generated by the db by default.

<details>
<summary><b>Strong migrations recommends adding this column in separate steps to avoid locking the table.</b></summary>

=== Dangerous operation detected #strong_migrations ===

Adding a column with a volatile default blocks reads and writes while the entire table is rewritten.
Instead, add the column without a default value, then change the default.

```ruby
class AddUuidToProviderSchool < ActiveRecord::Migration[8.1]
  def up
    add_column :provider_school, :uuid, :uuid
    change_column_default :provider_school, :uuid, "uuid_generate_v4()"
  end

  def down
    remove_column :provider_school, :uuid
  end
end
```

Then backfill the existing rows in the Rails console or a separate migration with disable_ddl_transaction!.

```ruby
class BackfillAddUuidToProviderSchool < ActiveRecord::Migration[8.1]
  disable_ddl_transaction!

  def up
    ProviderSchool.unscoped.in_batches(of: 10000) do |relation|
      relation.where(uuid: nil).update_all("uuid = uuid_generate_v4()")
      sleep(0.01)
    end
  end
end
```
Then add the NOT NULL constraint in separate migrations.

</details>

1. Add UUID column to provider_school
2. Backfill uuid
3. Contrain the column to not null

## Guidance to review

The database task `*:with_data` will interleve the schema and data migrations so we can release in one go.

```
up     schema  20260703145835  Add uuid to provider school
up      data   20260706120000  Backfill provider school uuids
up     schema  20260706120100  Set provider school uuid not null
```

The next steps will be to update the *::School backfill to set the correct UUID for each provider school
## Checklist

### DIFF
--- a/db/data/20260706120000_backfill_provider_school_uuids.rb
+++ b/db/data/20260706120000_backfill_provider_school_uuids.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class BackfillProviderSchoolUuids < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  TAG = "[BackfillProviderSchoolUuids]"
+
+  def up
+    total = 0
+
+    Provider::School.unscoped.where(uuid: nil).in_batches(of: 5_000) do |batch|
+      total += batch.update_all("uuid = uuid_generate_v4()")
+    end
+
+    say "#{TAG} backfilled=#{total}"
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -25,6 +25,7 @@
   - provider_id
   - gias_school_id
   - site_code
+  - uuid
   - created_at
   - updated_at
   :data_hub_process_summary:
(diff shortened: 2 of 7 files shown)


inulty-dfe:
[5.6.1 School - Model Adding UUID to provider_school and provider_school_id to course_school table](https://trello.com/c/yDoxkeGU/1700-561-school-model-adding-uuid-to-providerschool-and-providerschoolid-to-courseschool-table)

db/data/20260706120000_backfill_provider_school_uuids.rb
```diff
@@ -0,0 +1,21 @@
...
+class BackfillProviderSchoolUuids < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  TAG = "[BackfillProviderSchoolUuids]"
+
+  def up
+    total = 0
+
+    Provider::School.unscoped.where(uuid: nil).in_batches(of: 5_000) do |batch|
+      total += batch.update_all("uuid = uuid_generate_v4()")
```
tomas-stefano, line 12: remind me again as a safe check. we won't copy the same uuids from the site table for the new cycle?

tomas-stefano (approved):

db/data/20260706120000_backfill_provider_school_uuids.rb
inulty-dfe: No, we're just using this to set UUIDs so we can have a not-null contstraint on deployment.

The real UUIDs will be set in the data migration

PR #6214, opened 2026-07-10
Add provider_school_id to the course_school table

inulty-dfe:
## Context

We want to associate a Course::School to a Provider::School so that we can join the tables when searching for courses by location.

- Run data migration to truncate the course_school table (data is not in use in production)
- Add the reference to provider school with not null and foreign key
- Update the unique index on CourseSchool by replacing `gias_school_id` + `site_code` with `provider_school_id`

## Guidance to review

The migrations will be run interleaved
```
  down     data   20260709000000  Truncate course school table
  down    schema  20260709095033  Add provider school id to course school
  down    schema  20260713102253  Change index on course school
```

## Checklist


inulty-dfe:
[5.6.1 School - Model Adding UUID to provider_school and provider_school_id to course_school table](https://trello.com/c/yDoxkeGU/1700-561-school-model-adding-uuid-to-providerschool-and-providerschoolid-to-courseschool-table)

db/data/20260709100000_backfill_course_school_provider_school_ids.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def up
+    # Link rows written before provider_school_id existed, using the same
+    # lookup CourseSchools::Creator does (course's provider + gias_school).
+    backfilled = exec_update(<<~SQL)
+      UPDATE course_school
+      SET provider_school_id = provider_school.id
+      FROM course, provider_school
+      WHERE course.id = course_school.course_id
+        AND provider_school.provider_id = course.provider_id
```
langesi: If you really want this to be accurate you also need to add

`provider_school.site_code = course_school.site_code`

As a provider can have the same gias_school twice because of the main site exception.

We will however remove all the data in this table before we runt he migration script, so I leave it up to you if you want to change it or just merge it.

app/services/course_schools/creator.rb
```diff
@@ -20,6 +20,7 @@ def call
 
       @course.schools.find_or_create_by!(gias_school_id: @gias_school_id) do |course_school|
         course_school.site_code = provider_school.site_code
+        course_school.provider_school = provider_school
```
langesi: I’ve just realised that the creator service doesn’t take site code into account. Our uniqueness constraint is course_id + gias_school_id + site_code

Main sites have an exception that allows a gias_school to be linked to a provider twice. In those cases, the creator service currently selects one of the links at random. I believe. I’ll raise a separate PR to fix this. I don't think this PR should be blocked by this.

db/schema.rb
```diff
@@ -260,10 +260,12 @@
     t.integer "course_id", null: false
     t.datetime "created_at", null: false
     t.bigint "gias_school_id", null: false
+    t.bigint "provider_school_id", null: false
     t.text "site_code", null: false
     t.datetime "updated_at", null: false
     t.index ["course_id", "gias_school_id", "site_code"], name: "index_course_school_unique", unique: true
```
langesi: As we have now the provider_school_id, we could change this index to just include :`course_id `and `:provider_school_id`.

Let me know what you think of that

langesi (commented):
Good job overall.

Can you change the unique index, please?

db/schema.rb
inulty-dfe: Yeah that makes sense. Is it necessary?

inulty-dfe: Done

db/data/20260709100000_backfill_course_school_provider_school_ids.rb
inulty-dfe: I've changed the whole PR to truncate the course_school table and create the reference not null in one go.

tomas-stefano:
I think this will break https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/services/data_hub/schools_backfill/executor.rb#L57

Can we check if my suspicious is true?

I think the tests should catch though 🤔 🤔 🤔 

Simon's comment:
Yes, it will blow up, that is not a blocker for this PR. Once this PR is in I'll amend the executor in my PR.

tomas-stefano:
Course::School now stores both gias_school_id and provider_school_id, and the two can disagree.

Maybe is good to add a validation, wdyt?

Simon's comment:
Yes, we not have a few denormalised columns. I would stick to most of them for query simplicity and performance. There would be one less join to make in most queries.

I could think of a few good validations to add to the Course::School model. We can do that though after all is done, as I don't think that any of these validations will be a make of break for this work

tomas-stefano:
> Course::School now stores both gias_school_id and provider_school_id, and the two can disagree.
> 
> Maybe is good to add a validation, wdyt?

Same applies to site_code, which is now a third  copy of data owned by Provider::School

Simon's comment:
Same applies as above

app/models/provider/school.rb
```diff
@@ -10,6 +10,8 @@ class Provider::School < ApplicationRecord
   belongs_to :provider, class_name: "::Provider", inverse_of: :schools
   belongs_to :gias_school
 
+  has_many :course_schools, class_name: "Course::School", inverse_of: :provider_school, dependent: :destroy
```
tomas-stefano, line 13: One thing is bothering me.
` dependent: :destroy` plus `on_delete: :cascade` means deletinga provider school silently detaches courses.
Two mechanisms doing the same thing, and both are destructive-by-default. A provider school that is in use by a live course should almost certainly be blocked from deletion (`dependent: :restrict_with_error`) rather than quietly cascading and leaving the course with fewer schools than  the publisher chose. 

@langesi I think we discuss that it should detach from a course but what product and design think of this? What should happens when a school from a provider is deleted when attached to courses?

spec/factories/course_schools.rb
```diff
@@ -5,6 +5,10 @@
     course
     gias_school
     site_code { "-" }
+    provider_school do
+      Provider::School.find_by(provider: course.provider, gias_school:, site_code:) ||
+        association(:provider_school, provider: course.provider, gias_school:, site_code:)
+    end
```
tomas-stefano, line 12: The block creates a Provider::School with the same site_code as the course_school, and the default  is "-". 

Provider::School validates only one "-" site per provider, so create(:course_school, course:) twice for the same course now blows up. 

That's exactly wh `spec/decorators/course_decorator_spec.rb:896-898` had to grow explicit `site_code: "A"/"B"/"C"` — thefix was applied to the symptom, not the cause. 

Give the factory a sequence for site_code (or a main_site trait for "-") so callers don't have to know about the main-site rule.

app/models/provider/school.rb
langesi: We have a deletion ticket in ready for dev where I discussed this case with UCD. 

Let's leave this one in. We can guard from wrongful deletions in the other ticket where there is context given.

spec/factories/course_schools.rb
inulty-dfe: I've changed the default site_code to be a normal site code and the exception to be main site "-"

app/models/course/school.rb
```diff
@@ -7,7 +7,30 @@ class Course::School < ApplicationRecord
...
+  validates :provider_school_id, uniqueness: { scope: :course_id }
+  validate :consistent_with_provider_school
+
+private
+
+  # gias_school_id and site_code are denormalised copies of the linked
+  # Provider::School (gias_school_id is kept so courses can be filtered by the
+  # GIAS school's lat/long without joining provider_school). Guard against the
+  # two sources of truth drifting apart.
+  def consistent_with_provider_school
```
tomas-stefano, line 22: the consistency validation checks GIAS school and site code,  but not the provider. 

  Nothing asserts provider_school.provider_id == course.provider_id. 
A course can be attached to a different provider's school and the record validates and saves. Your own new spec demonstrates it accidentally:
```ruby
  it "rejects a gias_school_id that disagrees with the provider_school" do
    provider_school = create(:provider_school)   # ← belongs to some other provider
    ...
    expect(course_school.errors[:gias_school_id]).to be_present   # passes, but only on GIAS
  end
```

  The factory's course and this provider_school have unrelated providers, and the only error raised is the GIAS one. Swap the GIAS school to a matching one and that cross-provider row is valid.
 
Add the check:
```ruby
  if provider_school.provider_id != course&.provider_id
    errors.add(:provider_school_id, "must belong to the course's provider")
  end
```

  Worth a spec that builds a course_school whose provider_school matches on GIAS school and site code but belongs to another provider — that's the case currently uncovered.

app/models/course/school.rb
```diff
@@ -7,7 +7,30 @@ class Course::School < ApplicationRecord
...
+    # Compare the records, not the *_id columns: a not-yet-saved gias_school has
+    # a nil foreign key until belongs_to autosave runs (after validation), so a
+    # column comparison would spuriously fail on create.
+    if gias_school != provider_school.gias_school
+      errors.add(:gias_school_id, "must match the provider school's GIAS school")
+    end
+
+    if site_code != provider_school.site_code
+      errors.add(:site_code, "must match the provider school's site code")
+    end
```
tomas-stefano, line 38: Remind me again why we need to denormalize the site code in course school if is in provider school?

Once a course_school exists, editing the parent provider_school's site_code instantly invalidates
  every child row: the next course.save that autosaves schools fails validation with a confusing
  "site code must match" error, far from the edit that caused it.
 Today there's no `ProviderSchools::Updater` so nothing can trigger it, but the support Site-edit path is an obvious future source of exactly this.

langesi: Originally, when planning the data remodel, I completely overlooked both the site_code field and the site UUID. As a result, neither was considered in the initial design. This was an oversight on my part.

Once Iain started work on the API ticket, we realised that these two fields were missing, and I then had to revise the plan to accommodate them.

Over time, the discussion around the UUID evolved from adding it to both tables to storing it only in the provider_school table and joining to that table whenever the UUID is needed.

The site_code, however, had already been planned for inclusion in both tables. This is because it determines whether a site is the main site, which in turn affects the name displayed for that site. Unlike the UUID, the site name is required in every request where a school is displayed. Storing site_code only in provider_school would therefore require an additional join every time school information is shown in the UI.

I see the complexity this decision has caused and I'd be happy to spend some time this afternoon to normalise the tables.

app/services/course_schools/creator.rb
```diff
@@ -18,11 +18,15 @@ def initialize(course:, gias_school_id:)
     def call
       provider_school = @course.provider.schools.find_by!(gias_school_id: @gias_school_id)
```
langesi, line 19: You can have 2 entries for the same gias school in the provider_school table, one for main site and one for a normal entry. 

This would just select any one of the 2.

Ok, so I don't want to block this PR any further though, I had a look further into the `UpdateCourseSchoolsService` and it's all based on site_id, which we want to get rid of. Which feeds this service.

So rather than you working on a frankensteins monster, I'll spend today reworking this and you can ignore the creation of schools.

I'll create a separate ticket to take care of the creator service and UpdateCourseSchoolsService

app/models/course/school.rb
langesi: I'm going to rewrok UpdateCourseSchoolsService by creating a seperate service just for our new datamodel, I can add this

langesi (approved):
Thank you @inulty-dfe , I know this ticket was a bit of a headache!

app/models/course/school.rb
inulty-dfe: done

app/services/course_schools/creator.rb
inulty-dfe: We've decided to merge this as is and focus on the creator/remover services later

PR #6216, opened 2026-07-10
Require the correct superdiff module in spec_helper

inulty-dfe:
## Context

image

  - https://splitwise.github.io/super_diff/releases/0.19.0/users/getting-started/

Require the correct module in superdiff so it can handle rails object errors properly

It avoids disasters like this 

image

## Checklist


gms-gs (approved):
Nice! 🚀

PR #6217, opened 2026-07-10
Restore link to email alerts index in weekly digest

inulty-dfe:
## Context

  - In a previous change I deleted the wrong link. We had duplicate links to the find results page, one of which should be deleted. I deleted the link to the email alerts index which should not have been deleted.

## Checklist


inulty-dfe:
[Two links in the email alert body point to the same place](https://trello.com/c/GlGpKTCw/1721-two-links-in-the-email-alert-body-point-to-the-same-place)

gms-gs (approved):

PR #6218, opened 2026-07-13
Placement schools list improvements

tomas-stefano:
## Context

User research with providers found they struggle to attach the right schools to a course, because the placement schools page lists **every** school in their account in one very long list, with each school's name shown twice (once as the checkbox label, once at the start of the address hint).

This PR makes three improvements to that page, reached via a course's **Basic details** tab → **Change** on *Placement schools* (e.g. `/publish/organisations/1Q6/2026/courses/2LCR/schools`).

### 1. Only the first 20 schools are shown, with a "Show all schools" link

A new `show-all-schools` Stimulus controller collapses the list to the first 20 on connect and renders a "Show all schools" link when the provider has more than 20. Clicking it reveals the rest.

The overflow rows stay in the DOM (they're only hidden visually), which means **"Select all schools" still selects the full list while the list is collapsed**, and every selected school is submitted — this is covered by a system spec that collapses the list, hits "Select all", submits, and asserts all 25 schools are attached.

Two implementation notes worth flagging for review:

- The rows are hidden with the `govuk-!-display-none` utility rather than the `hidden` attribute alone. GOV.UK sets `display: flex` on `.govuk-checkboxes__item`, and that author rule beats the user-agent `[hidden] { display: none }` rule — so setting `hidden` on its own does not actually hide the row.
- This is progressive enhancement: the rows are hidden and the link is revealed only by JS, so without JS a provider sees the full list and no dangling link.

### 2. A count of currently attached schools below the heading

> You currently have 23 schools attached to this course

Backed by a new `CourseDecorator#attached_schools_count`, which is feature-flag aware and mirrors the branch already used by `CourseDecorator#sorted_school_names`. Copy is pluralised (including a zero case).

### 3. The duplicated school name is removed from the address hint

The hint previously used `Site#full_address`, which prepends the location name — so the school name appeared both in the bold label and again at the start of the grey address. The hint now uses the existing `SiteDecorator#full_address`, which returns the address without the name (and applies `smart_quotes`, consistent with the rest of the app).

### Design feedback applied

- "Show all schools" link uses the standard 19px body size (it was inheriting 16px, because `.app-button-link` sets `font-size: inherit`), with extra space between it and the last checkbox.
- School address hint drops to 16px.
- The attached-schools count renders as `govuk-body` rather than `govuk-hint`.


app/views/publish/courses/schools/edit.html.erb
```diff
@@ -24,14 +24,21 @@
         provider: @provider,
       ) %>
 
+    <% sorted_sites = @provider.sites.sort_by(&:location_name) %>
+    <% collapse_after = 20 %>
```
inulty-dfe: can we avoid putting this in the template, maybe the controller or helper is better

inulty-dfe (approved):
LGTM

PR #6219, opened 2026-07-13
6.2 School remodel - Touch Provider on ProviderSchool and CourseSchool destroy action

gms-gs:
## Context
As part of the main site / schools remodel, we added logic to touch the parent Provider when school relationships change.

However, we didn’t add the destroy paths.

When a provider_school is removed, the associated Provider needs to be touched so updated_at changes.

When a course_school is removed, the provider for the associated course also needs to be touched so Apply can detect that the provider’s school data has changed and re-sync the affected provider/course/school data.

Apply uses our API to detect changed providers by checking the provider updated_at. If the provider is not touched on school removal, Apply may not pick up the change.

- Touch the associated provider when a `Provider::School` is destroyed.
- Touch the course and its provider when a `Course::School` is destroyed.
- Add specs confirming only the relevant provider is touched.
- Preserve existing create and update behaviour.

## Guidance to review

Run:

`bundle exec rspec spec/models/provider/school_spec.rb spec/models/course/school_spec.rb`

Check that destroying either school relationship updates the associated provider’s `changed_at` without affecting unrelated providers.

## Checklist

- [x] I have moved hard-coded strings to locale files (not applicable).
- [x] I have removed the usage of `data-qa` attributes in HTML files and updated the corresponding tests (not applicable).

### DIFF
--- a/spec/models/course/school_spec.rb
+++ b/spec/models/course/school_spec.rb
@@ -61,11 +61,21 @@ describe Course::School do
       end
     end
 
-    it "bumps the course and only its provider.changed_at on destroy" do
+    it "bumps course.changed_at on destroy" do
+      course_school = create(:course_school, course:)
+      course.update_columns(changed_at: 1.hour.ago)
+
+      Timecop.freeze do
+        course_school.destroy!
+
+        expect(course.reload.changed_at).to be_within(1.second).of(Time.zone.now)
+      end
+    end
+
+    it "bumps only the course's provider.changed_at on destroy" do
       course_school = create(:course_school, course:)
       provider = course.provider
       other_provider = create(:provider)
-      course.update_columns(changed_at: 1.hour.ago)
       provider.update_columns(changed_at: 1.hour.ago)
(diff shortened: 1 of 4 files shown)


gms-gs:
[6.2 School remodel - Touch Provider on ProviderSchool and CourseSchool destroy action](https://trello.com/c/xL1JJJcE/1702-62-school-remodel-touch-provider-on-providerschool-and-courseschool-destroy-action)

spec/models/course/school_spec.rb
```diff
@@ -61,13 +61,22 @@
       end
     end
 
-    it "does not bump course.changed_at on destroy" do
+    it "bumps the course and only its provider.changed_at on destroy" do
```
inulty-dfe: This test is testing that destroying a course school doesn't update the changed at of a provider that doesn't have a course with the course school attached. Is that right?

I think it would be easier to understand if there were two specs. "and" is a dead giveaway for a separate spec

inulty-dfe (approved):
LGTM

PR #6224, opened 2026-07-14
Remove DatabaseCleaner gem

inulty-dfe:
## Context

 - This gem serves no purpose that can't be managed within the application

## Checklist


inulty-dfe:
Deprioritised

PR #6227, opened 2026-07-14
Normalised site_code in course_school

langesi:
## Context

We are normalising the new school data model as part of the school remodel work.

course_school.site_code was denormalised because the site code already belongs on provider_school. This PR removes that duplication and moves course school writes towards the new provider_school / course_school model while keeping the legacy site / site_status writes in place for now.

View the changes, comment or approve

## Guidance to review

Please review this PR by topic rather than file order.

### 1. Schema and normalisation

Review:

- `db/migrate/*remove_site_code_from_course_school*`
- `db/migrate/*add_unique_index_to_provider_school_uuid*`
- `db/schema.rb`
- `app/models/course/school.rb`
- `config/analytics_blocklist.yml`

Check that `course_school.site_code` is fully removed from the new model, and that any remaining site code ownership is through `provider_school`.

### 2. Backfill logic can be skipped as this will not run now and will be disected at a later point, changes here are only to make it work for now.

files:
- `db/data/20260706120000_backfill_provider_school_uuids.rb`
- `app/services/data_hub/schools_backfill/executor.rb`

## Checklist

### DIFF
--- a/app/services/data_hub/schools_backfill/executor.rb
+++ b/app/services/data_hub/schools_backfill/executor.rb
@@ -37,11 +37,37 @@ module DataHub
 
       def insert_provider_schools
         inserted_rows = ActiveRecord::Base.connection.exec_query(<<~SQL)
-          INSERT INTO provider_school (provider_id, gias_school_id, site_code, created_at, updated_at)
+          INSERT INTO provider_school (provider_id, gias_school_id, site_code, uuid, created_at, updated_at)
           SELECT DISTINCT ON (site.provider_id, gias_school.id, site.code)
-                 site.provider_id, gias_school.id, site.code, NOW(), NOW()
+                 site.provider_id, gias_school.id, site.code, site.uuid, NOW(), NOW()
           FROM site
           JOIN gias_school ON gias_school.urn = site.urn
+          WHERE site.site_type = 0
+            AND site.discarded_at IS NULL
+            AND site.urn IS NOT NULL
+            AND site.urn <> ''
+          ON CONFLICT (provider_id, gias_school_id, site_code) DO UPDATE
+            SET uuid = EXCLUDED.uuid,
+                updated_at = NOW()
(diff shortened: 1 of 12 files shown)


app/services/data_hub/schools_backfill/executor.rb
```diff
@@ -37,29 +37,58 @@ def execute
 
       def insert_provider_schools
         inserted_rows = ActiveRecord::Base.connection.exec_query(<<~SQL)
-          INSERT INTO provider_school (provider_id, gias_school_id, site_code, created_at, updated_at)
+          INSERT INTO provider_school (provider_id, gias_school_id, site_code, uuid, created_at, updated_at)
           SELECT DISTINCT ON (site.provider_id, gias_school.id, site.code)
-                 site.provider_id, gias_school.id, site.code, NOW(), NOW()
+                 site.provider_id, gias_school.id, site.code, site.uuid, NOW(), NOW()
```
inulty-dfe, line 42: Is this critical to the site_code changes?

spec/factories/course_schools.rb
```diff
@@ -1,16 +1,27 @@
...
+      site_code { generate(:course_school_provider_site_code) }
+    end
+
     provider_school do
       Provider::School.find_by(provider: course.provider, gias_school:, site_code:) ||
         association(:provider_school, provider: course.provider, gias_school:, site_code:)
     end
 
+    after(:build) do |course_school, evaluator|
+      if course_school.has_attribute?(:site_code)
```
inulty-dfe: When will course_school have this attribute after the migration is run?

inulty-dfe (commented):
Couple questions. one on the scope of the changes related to uuid and the other about guarding against site_code being an attribute on course_school after this change is run

spec/factories/course_schools.rb
langesi: Removed from the codebase, was essentially dead code, wouldn't raise and error, but didn't do anything even though it's a callback

inulty-dfe (approved):

PR #6229, opened 2026-07-14
Provider School creation refactor

langesi:
## Context

We are moving school relationships from the legacy `Site` table to the new `Provider::School` / `Course::School` data model.

During the transition period, provider schools still need to be dual-written to both models. For the 2026 recruitment cycle and earlier, we still create legacy `Site` records for schools, but the matching `Provider::School` must use the same `site_code` and `uuid` as the legacy `Site`. This prevents identity drift between the old and new school records while both models are in use.

From the 2027 recruitment cycle onwards, legacy school `Site` records should no longer be created when adding schools to providers. At that point, adding provider schools should write only to `Provider::School`.

- Updates support and publish school-add flows so current-cycle provider school creation writes the legacy `Site` first, then creates the matching `Provider::School` using the legacy site's `site_code` and `uuid`.
- Applies the same behaviour to:
  - support single-school add
  - support multiple-school add
  - publish single-school add
  - publish multiple-school add
- Stops creating legacy school `Site` records after `Settings.schools_remodel_cycle_year`.
- Updates `ProviderSchools::Creator` to accept optional `site_code` and `uuid` values from the legacy `Site`.
- Fixes `ProviderSchools::Creator` idempotency for the post-remodel path where callers only pass `provider` and `gias_school_id`.
- Ensures the creator does not accidentally reuse a provider main-site row when creating a normal provider school for the same GIAS school.
- Keeps `RecordNotUnique` recovery working for both explicit `site_code` calls and post-remodel calls without a supplied `site_code`.
- Adds rollback coverage to prove the legacy `Site` is not left behind if `Provider::School` creation fails during the 2026 dual-write flow.
- Adds system coverage for current-cycle UUID parity and post-remodel provider-school-only creation across support and publish add-school flows.

## Guidance to review

- Review the four controller paths that add schools to providers and check that the cycle split is correct:
  - `app/controllers/support/providers/schools/checks_controller.rb`
  - `app/controllers/support/providers/schools/check_multiple_controller.rb`
  - `app/controllers/publish/providers/schools/checks_controller.rb`
  - `app/controllers/publish/providers/schools/check_multiple_controller.rb`
- For 2026 and earlier, confirm the legacy `Site` is created first and its `site_code`/`uuid` are passed into `ProviderSchools::Creator`.
- For 2027 onwards, confirm only `Provider::School` is created.
- Review `ProviderSchools::Creator` idempotency carefully, especially the no-`site_code` path used after the remodel cycle.
- Rollover and school deletion are intentionally out of scope for this PR.
- Useful specs to run:
  - `bundle exec rspec spec/services/provider_schools/creator_spec.rb`
  - `bundle exec rspec spec/requests/support/providers/schools/checks_spec.rb`
  - `bundle exec rspec spec/system/support/providers/schools/add_school_to_provider_spec.rb spec/system/support/providers/schools/creating_multiple_schools_spec.rb`
  - `bundle exec rspec spec/system/publish/providers/schools/add_school_spec.rb spec/system/publish/providers/schools/creating_multiple_schools_spec.rb`

## Checklist

### DIFF
--- a/app/controllers/publish/providers/schools/check_multiple_controller.rb
+++ b/app/controllers/publish/providers/schools/check_multiple_controller.rb
@@ -57,11 +57,19 @@ module Publish
 
         def create_provider_school(gias_school:)
           if provider_school_identity.after_schools_remodel_cycle?
-            return ::ProviderSchools::Creator.call(provider:, gias_school_id: gias_school.id)
+            create_only_provider_school(gias_school:)
+          else
+            create_provider_school_and_legacy_site(gias_school:)
           end
+        end
 
-          # rubocop:disable Style/CommentAnnotation, Lint/RedundantCopDisableDirective
-          # TODO School data remodel removal - remove this legacy Site write when publish creates Provider::School directly.
+        def create_only_provider_school(gias_school:)
+          ::ProviderSchools::Creator.call(provider:, gias_school_id: gias_school.id)
+        end
+
+        # rubocop:disable Style/CommentAnnotation, Lint/RedundantCopDisableDirective
+        # TODO School data remodel removal - remove this legacy Site write when publish creates Provider::School directly.
+        def create_provider_school_and_legacy_site(gias_school:)
           legacy_site = provider.sites.build(gias_school.school_attributes)
           ::ProviderSchools::LegacySiteCreator.call(site: legacy_site)
           # rubocop:enable Style/CommentAnnotation, Lint/RedundantCopDisableDirective
--- a/app/controllers/publish/providers/schools/check_multiple_controller.rb
+++ b/app/controllers/publish/providers/schools/check_multiple_controller.rb
@@ -10,19 +10,15 @@ module Publish
         def show; end
 
         def update
-          saved_sites = []
+          saved_schools = []
 
           gias_schools.each do |gias_school|
             ActiveRecord::Base.transaction do
-              provider_school = ::ProviderSchools::Creator.call(provider:, gias_school_id: gias_school.id)
-
-              site = provider.sites.build(gias_school.school_attributes.merge(code: provider_school.site_code))
-              ::ProviderSchools::LegacySiteCreator.call(site:)
-              saved_sites << site
+              saved_schools << create_provider_school(gias_school:)
             end
           end
 
-          schools_added_message(saved_sites)
+          schools_added_message(saved_schools)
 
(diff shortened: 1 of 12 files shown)


app/controllers/publish/providers/schools/checks_controller.rb
```diff
@@ -10,27 +10,56 @@ class ChecksController < ApplicationController
...
           render :show
         end
 
       private
 
-        def site
-          @site ||= begin
-            gias_school = GiasSchool.find(school_id)
-            @provider.sites.school.build(gias_school.school_attributes)
+        def create_provider_school
```
langesi, line 25: I’m aware that this logic could live in the creator service and that we have three other places using the same approach as this method. However, keeping it here should make it easier to remove the old data model from the controllers later.

I’ve also added a TODO to these methods so they can be found easily with grep.

app/services/provider_schools/creator.rb
```diff
@@ -9,19 +9,55 @@ module ProviderSchools
...
       end
     rescue ActiveRecord::RecordNotUnique
-      @provider.schools.find_by!(gias_school_id: @gias_school_id)
+      existing_provider_school || raise
+    end
+
+  private
+
+    def existing_provider_school
+      provider_school_by_uuid || provider_school_by_site_code || normal_provider_school
```
gms-gs, line 36: What does `normal_provider_school` actually mean?  seems ambiguous

app/controllers/publish/providers/schools/check_multiple_controller.rb
```diff
@@ -59,6 +55,34 @@ def provider
...
+
+        def create_only_provider_school(gias_school:)
+          ::ProviderSchools::Creator.call(provider:, gias_school_id: gias_school.id)
+        end
+
+        # rubocop:disable Style/CommentAnnotation, Lint/RedundantCopDisableDirective
+        # TODO School data remodel removal - remove this legacy Site write when publish creates Provider::School directly.
+        def create_provider_school_and_legacy_site(gias_school:)
+          legacy_site = provider.sites.build(gias_school.school_attributes)
+          ::ProviderSchools::LegacySiteCreator.call(site: legacy_site)
```
inulty-dfe, line 75: We could just call `legacy_site.save!`

It's the same thing

inulty-dfe (approved):

PR #6231, opened 2026-07-15
School data - Adapt rollover functionality

gms-gs:
## Context

Rollover previously copied only legacy school `Site` records into the new recruitment cycle.

This change adds support for `Provider::School` and `Course::School` while continuing to copy the legacy model. Both models are rolled over together without a feature flag.

- Isolate legacy `Site` and `SiteStatus` rollover logic.
- Add dual copiers that run the legacy and new-model paths together.
- Copy `Provider::School` relationships to the copied provider.
- Copy `Course::School` relationships to the copied course.
- Preserve `gias_school_id` and `site_code`.
- Link copied course-school records to the corresponding copied provider-school records.
- Retain existing rollover reporting based on legacy site-copy results.
- Keep study-site rollover unchanged.
- Preserve existing support course-copy behaviour.
- Add coverage for legacy, new-model, and dual-copy behaviour.

## Guidance to review

- Verify rollover creates both legacy and new school relationships.
- Confirm copied `Course::School` records reference the correct copied `Provider::School` and course.
- Check that `gias_school_id` and `site_code`, including the main-site `-` code, are preserved.
- Confirm study-site rollover remains unchanged.
- Run:

```shell
SUPPRESS_DFE_ANALYTICS_INIT=1 bundle exec rspec \
  spec/services/rollover/schools \
  spec/services/rollover_provider_service_spec.rb \
  spec/services/rollover_service_spec.rb \
  spec/services/providers/copy_to_recruitment_cycle_service_spec.rb \
  spec/services/courses/copy_to_provider_service_spec.rb
```

Once the legacy model is retired, the legacy copiers can be removed and the new-model copiers used directly instead of the dual copiers.

## Checklist

- [x] I have moved hard-coded strings to locale files. (Not applicable—no UI strings added.)
- [x] I have removed the usage of `data-qa` attributes in HTML files and updated the corresponding tests. (Not applicable—no HTML changes.)


gms-gs:
[9. School data - Adapt rollover functionality](https://trello.com/c/3maukXab/1591-9-school-data-adapt-rollover-functionality)

app/services/rollover/schools/new_provider_copier.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Rollover
+  module Schools
+    class NewProviderCopier
+      def execute(provider:, new_provider:)
+        provider.schools.each do |provider_school|
+          new_provider.schools.find_or_create_by!(
+            gias_school_id: provider_school.gias_school_id,
+            site_code: provider_school.site_code,
```
inulty-dfe, line 15: I think we need to copy the uuid here too

app/services/rollover/schools/new_course_copier.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Rollover
+  module Schools
+    class NewCourseCopier
```
inulty-dfe:
```suggestion
    class CourseCopier
```

Best not to prepend "New" because when we remove legacy "New" will be meaningless

It's ok in the Dual***Copier cause that will exist as long as the legacy route remains.

app/services/rollover/schools/new_provider_copier.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Rollover
+  module Schools
+    class NewProviderCopier
```
inulty-dfe:
```suggestion
    class ProviderCopier
```
Best not to prepend "New" because when we remove legacy "New" will be meaningless

app/services/rollover/schools/new_course_copier.rb
```diff
@@ -0,0 +1,21 @@
...
+      def call(course:, new_provider:, new_course:)
+        course.schools.includes(:provider_school).find_each do |course_school|
+          new_provider_school = new_provider.schools.find_by!(
+            gias_school_id: course_school.provider_school.gias_school_id,
+            site_code: course_school.provider_school.site_code,
```
inulty-dfe: `uuid` here too? I think there's an existing bug where `site_code` can be duplicated on a course/provider

inulty-dfe (changes requested):
So there's two routes, 
- controller (manual rollover) 
- service (scheduled bulk rollover)?

Two elements 
- course sites
- provider sites

The nesting of all these services makes it hard to understand

app/services/rollover/schools/new_provider_copier.rb
gms-gs: Maybe I am misunderstanding here or the way its setup wont allow for this. 

ProviderCopier copies schools with find_or_create_by!(uuid: …) and sets gias_school_id / site_code in the block.

CourseCopier links course schools by looking up the copied provider school via uuid (not gias_school_id + site_code), which avoids ambiguous site_code matching.

However rollover keeps both recruitment cycles in the DB at the same time, but provider_school.uuid currently has a global unique index (index_provider_school_on_uuid). That means we can’t insert the rolled-over row with the same uuid while the source row still exists — ProviderCopier and the course-copy spec both fail with PG::UniqueViolation.


@inulty-dfe

app/services/rollover/schools/course_copier.rb
```diff
@@ -0,0 +1,21 @@
...
+      def call(course:, new_provider:, new_course:)
+        course.schools.includes(:provider_school).find_each do |course_school|
+          new_provider_school = new_provider.schools.find_by!(
+            gias_school_id: course_school.provider_school.gias_school_id,
+            site_code: course_school.provider_school.site_code,
+          )
+
+          new_course.schools.find_or_create_by!(provider_school: new_provider_school) do |new_course_school|
+            new_course_school.gias_school_id = new_provider_school.gias_school_id
+            new_course_school[:site_code] = new_provider_school.site_code if new_course_school.has_attribute?(:site_code)
```
langesi: site_code is stored in the provider_school table. Since it is a delegated attribute, this line will potentially override the site_code column in provider_school.

I recommend removing this line of code.

langesi (changes requested):
Heyho, good job, have some comments.

Also, I see that your tests check that the new services call each other, however you haven't tested:

1. copies both legacy Site and new Provider::School records
2. copies both legacy SiteStatus and new Course::School records
3. links the copied Course::School to the copied Provider::School

inulty-dfe (approved):

langesi (approved):
Well done!

PR #6233, opened 2026-07-16
School remodel Find - Courses query

tomas-stefano:
## Context

Move Find's location/distance search onto the canonical `course_school → gias_school`
model, behind the existing `course_publishing_uses_new_school_model` flag. When the
flag is on, `Courses::Query` and `SavedCourses::Query` compute
`minimum_distance_to_search_location` from each school's single canonical coordinate
via `course_school`, instead of the per-provider-duplicated `course_site → site`
rows. When the flag is off, the legacy path is used unchanged.

This is stage one of two: it swaps the data source (canonical, deduplicated) while
keeping the query shape. A follow-up adds the spatial index (`geo_location` + GiST)
that turns location search into a 1–10 ms query.

### Performance

No spatial index yet, so both paths still scan and compute `ST_DistanceSphere` over
the course school / gias school set. The new path is currently **~3–13% faster** (canonical dedup +
one fewer join), e.g. London location-led 234 ms → 218 ms, Manchester 225 ms →
201 ms. The order-of-magnitude win lands with the GiST index in the follow-up.

A `location_search:bench` rake task is included in the commit history to reproduce the diff + `EXPLAIN
ANALYZE` timings on any database.

```
Fresh numbers (median of 7 EXPLAIN ANALYZE passes, Postgres time):

  ┌───────────────────────────┬────────┬────────┬──────────────┐
  │         Scenario          │ Old ms │ New ms │      Δ       │
  ├───────────────────────────┼────────┼────────┼──────────────┤
  │ London — location-led     │  234.1 │  217.9 │  −16.2 (−7%) │
  ├───────────────────────────┼────────┼────────┼──────────────┤
  │ London — filter-led       │  173.8 │  168.0 │   −5.8 (−3%) │
  ├───────────────────────────┼────────┼────────┼──────────────┤
  │ Manchester — location-led │  224.8 │  201.0 │ −23.8 (−11%) │
  ├───────────────────────────┼────────┼────────┼──────────────┤
  │ Manchester — filter-led   │  167.1 │  154.5 │  −12.6 (−8%) │
  ├───────────────────────────┼────────┼────────┼──────────────┤
  │ Penrith — location-led    │  208.0 │  181.9 │ −26.1 (−13%) │
  ├───────────────────────────┼────────┼────────┼──────────────┤
  │ Penrith — filter-led      │  164.4 │  158.3 │   −6.1 (−4%) │
  └───────────────────────────┴────────┴────────┴──────────────┘

  For now the new path is ~3–13% faster (roughly 5–25 ms). Modest — and expected. Both paths still
  sit at ~150–235 ms because there's no spatial index yet: both scan and compute ST_DistanceSphere
  over the candidate set. Stage A's gain comes only from the canonical dedup and dropping the
  course_site join.
```

## Behaviour change (measured on a prod-shaped dev DB, current cycle)

Non-location searches are unchanged. For location search the set of schools that
contribute distance moves from "running/published sites" to "all linked GIAS
schools", so results shift slightly. Measured whole-GB (all published current-cycle
courses):

- Old: 11,393 → New: 10,672 (**−721, ~6.3%**); 734 dropped, 13 added.
- **All 734 dropped courses are dropped because none of their sites map to a
  `gias_school`** (discarded sites, or URNs missing from `gias_school`) — i.e. a data
  gap, not a query-logic change. Concretely: of 380,872 findable course_sites on
  published current-cycle courses, only 1,627 (0.43%) have no `gias_school`.
- Courses taught only at non-school sites (no GIAS URN, e.g. a provider campus
  address) drop out of location search by design — they still appear in
  non-location search. (Product sign-off assumed.)

The residual narrowing shrinks as the `schools_backfill` is re-run and `gias_school`
URN coverage improves — see rollout.

## Testing

- New request specs for both queries under the flag: radius filtering, distance
  ordering, MIN across a course's schools, all-schools-outside-radius exclusion, the
  NULL-coordinate guard, duplicate edges to one school, cross-provider dedup, subject
  composition (filter-led), `#count`, and the non-school-site drop.
- Legacy specs stay green (flag off).

## Rollout / follow-ups

- **Next PR (same branch):** `geo_location` generated column + partial GiST index +  `ST_DWithin`;

### DIFF
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -350,10 +350,7 @@ module Courses
     # the legacy path's minimum_distance_to_search_location contract.
     def schools_location_scope(latitude:, longitude:, radius_in_meters:)
       @scope
-        .joins(<<~SQL)
-          INNER JOIN course_school ON course_school.course_id = course.id
-          INNER JOIN gias_school   ON gias_school.id = course_school.gias_school_id
-        SQL
+        .joins(schools: :gias_school)
         .where.not(gias_school: { latitude: nil })
         .where.not(gias_school: { longitude: nil })
         .where(
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -354,7 +354,8 @@ module Courses
           INNER JOIN course_school ON course_school.course_id = course.id
           INNER JOIN gias_school   ON gias_school.id = course_school.gias_school_id
         SQL
-        .where("gias_school.latitude IS NOT NULL AND gias_school.longitude IS NOT NULL")
+        .where.not(gias_school: { latitude: nil })
+        .where.not(gias_school: { longitude: nil })
         .where(
           <<~SQL.squish, longitude, latitude, radius_in_meters
             ST_DistanceSphere(
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -296,10 +296,16 @@ module Courses
         radius: radius_in_miles,
       }
 
-      @scope = sites_location_scope(latitude:, longitude:, radius_in_meters:)
+      @scope =
+        if FeatureFlag.active?(:course_publishing_uses_new_school_model)
+          schools_location_scope(latitude:, longitude:, radius_in_meters:)
+        else
+          sites_location_scope(latitude:, longitude:, radius_in_meters:)
+        end
     end
 
-    # Location filter over the course_site -> site model.
+    # Location filter over the legacy course_site -> site model, used while the
+    # :course_publishing_uses_new_school_model flag is off.
     def sites_location_scope(latitude:, longitude:, radius_in_meters:)
       @scope
         .joins(<<~SQL)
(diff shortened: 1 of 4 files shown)


app/services/courses/query.rb
```diff
@@ -333,6 +344,44 @@ def location_scope
...
+    # Location filter over the canonical course_school -> gias_school model, used
+    # while the :course_publishing_uses_new_school_model flag is on. Distance is
+    # measured from each school's coordinate once and exposed in miles, matching
+    # the legacy path's minimum_distance_to_search_location contract.
+    def schools_location_scope(latitude:, longitude:, radius_in_meters:)
+      @scope
+        .joins(<<~SQL)
+          INNER JOIN course_school ON course_school.course_id = course.id
+          INNER JOIN gias_school   ON gias_school.id = course_school.gias_school_id
+        SQL
```
inulty-dfe:
```suggestion
        .joins(schools: :gias_school)
```

app/services/saved_courses/query.rb
```diff
@@ -63,6 +74,35 @@ def location_scope
...
+    # Distance annotation over the canonical course_school -> gias_school model,
+    # used while the :course_publishing_uses_new_school_model flag is on. As with
+    # the legacy path, saved courses are annotated with distance but not filtered
+    # by radius.
+    def schools_location_scope(latitude:, longitude:)
+      @scope
+        .joins(<<~SQL)
+          INNER JOIN course_school ON course_school.course_id = course.id
+          INNER JOIN gias_school   ON gias_school.id = course_school.gias_school_id
+        SQL
```
inulty-dfe:
```suggestion
        .joins(schools: :gias_school)
```

inulty-dfe (approved):
LGTM

app/services/courses/query.rb
```diff
@@ -333,6 +344,44 @@ def location_scope
...
+        .where("gias_school.latitude IS NOT NULL AND gias_school.longitude IS NOT NULL")
```
inulty-dfe:
```suggestion
        .where.not(gias_school: { latitude: nil, longitude: nil })
```

app/services/saved_courses/query.rb
```diff
@@ -63,6 +74,35 @@ def location_scope
...
+        .where("gias_school.latitude IS NOT NULL AND gias_school.longitude IS NOT NULL")
```
inulty-dfe:
```suggestion
        .where.not(gias_school: { latitude: nil, longitude: nil })
```

tomas-stefano: do you mean?

```
.where.not(gias_school: { latitude: nil  }).where.not(gias_school: { longitude: nil })
```

PR #6235, opened 2026-07-17
School remodel find query spatial index

tomas-stefano:
## What

Make Find's "courses near me" search fast by adding a spatial index. This stacks on
the earlier PR that moved location search onto `course_school → gias_school`; that PR
swapped the *data source*, this one adds the index that makes it quick.

Three pieces:
1. A `geo_location geography(Point, 4326)` **generated column** on `gias_school`,
   derived from the existing `latitude`/`longitude`, with a **partial GiST index**.
2. Registration of the new field in the dfe-analytics blocklist.
3. `Courses::Query#schools_location_scope` rewritten to prune with `ST_DWithin` over
   the index and aggregate to one row per course **before** the course-side filters.

Behind the existing `course_publishing_uses_new_school_model` flag.

---

## How the GiST index works

A GiST index stores each school's point inside a tree of **bounding boxes**.
`ST_DWithin(geo_location, :point, :radius)` walks that tree and only visits schools
whose box overlaps the search circle — far-away schools are eliminated **without
computing their distance at all**. Without the index, Postgres computes the distance
to every candidate school and then throws most away.

### Worked example

A candidate searches from **Charing Cross `(51.508, −0.128)`**, radius **10 miles**:

```
        the 10-mile search envelope (what the GiST index scans)
   ┌───────────────────────────────────────────┐
   │     ● St Mary's  (2.0 mi)                   │
   │  ● Greenhill (0.6 mi)                       │            ● Elmwood, Reading (35 mi)
   │        ✕ search point                       │              ← OUTSIDE the box:
   │                     ● Kingsway (9.8 mi)     │                the index never
   └───────────────────────────────────────────┘                visits or measures it
```

| School | URN | Coordinates | Distance | GiST verdict |
|---|---|---|---:|---|
| Greenhill Academy | 100001 | 51.500, −0.120 | 0.6 mi | ✅ visited → matches |
| St Mary's Primary | 100002 | 51.530, −0.140 | 2.0 mi | ✅ visited → matches |
| Kingsway High | 100003 | 51.370, −0.100 | 9.8 mi | ✅ visited → matches (just inside) |
| Elmwood College | 100004 | 51.455, −0.970 | 35 mi | ⛔ **never visited** (box pruned) |

Real numbers: for a central-London search the index returns ~2,400 nearby schools in
~2 ms, instead of scanning all ~27,000.

The index is **partial** — `WHERE geo_location IS NOT NULL` — so schools still awaiting
geocoding never bloat it and never match a search.

---

## The query: prune, then aggregate per course

The subtlety is *when* the expensive per-course work happens. Nearby schools fan out
to many `(school × course)` edges (~42,000 for dense London), and running the
per-course "is this course published?" check on every edge is what made a naive
version slow. So the query aggregates to **one row per course first**, in a derived
table Postgres won't flatten, and only then joins `course`:

```mermaid
flowchart TD
    A["Search point + radius"] --> B["Phase 1 — GiST prune<br/>ST_DWithin(geo_location, point, radius)"]
    B -->|"far schools skipped"| C["Nearby schools only<br/>(~2,400 near central London)"]
    C --> D["Join course_school, GROUP BY course_id<br/>MIN(distance) = nearest school per course"]
    D --> E["Distinct nearby courses<br/>(~1,946)"]
    E --> F["Join course + published / base filters<br/>run ONCE per distinct course, not per edge"]
    F --> G["Ordered results with<br/>minimum_distance_to_search_location (miles)"]
```

Distances use `ST_Distance(..., false)` / `ST_DWithin(..., false)` — the trailing
`false` forces **sphere** maths, matching the legacy `ST_DistanceSphere` path so
results don't shift due to the maths.

---

## Performance

Median of 7 `EXPLAIN ANALYZE` passes, on a **restored production database** (current
cycle, ~11k published courses, 27k schools). "Old" = flag off (legacy `course_site`
path), "New" = flag on (this PR).

| Scenario | Old (ms) | New (ms) | Speed-up |
|---|--:|--:|--:|
| London — location-led (browse) | 249.8 | **62.3** | **4.0×** |
| London — filter-led (subject) | 185.5 | 43.7 | 4.2× |
| Manchester — location-led | 229.1 | **22.0** | 10.4× |
| Manchester — filter-led | 179.1 | 18.3 | 9.8× |
| Rural (Penrith) — location-led | 220.3 | **3.5** | 63× |
| Rural (Penrith) — filter-led | 170.8 | 1.6 | 107× |

The GiST prune delivers the 1–10 ms target wherever few schools are nearby (rural,
filter-led). Dense central-London "browse everything" stays ~60 ms because it
genuinely returns ~1,950 courses — the residual is the per-course published checks and
the final sort, not the distance maths. `EXPLAIN` confirms the plan uses
`index_gias_school_on_geo_location` with no sequential scan on `gias_school`.

---

## Behaviour notes

- **Result count** is unchanged from the previous (Stage A) PR — this is a pure
  performance change; the flag-on result set is identical to before, bit-for-bit.
- **Small distance shifts vs flag-off** are expected: the legacy path measures from
  `site` coordinates, the new path from **GIAS canonical** coordinates for the same
  school. For the same physical school these can differ slightly, nudging a course's
  `minimum_distance_to_search_location` by a fraction of a mile. Generally more
  accurate; never changes the radius bucket.

---

## Migration & rollout notes

- The generated column uses the **IMMUTABLE** `::geography` cast, so it is legal in a
  `STORED` column (`ST_Transform` is not IMMUTABLE and would be rejected). Note
  `ST_MakePoint(X, Y)` = `(longitude, latitude)`.
- Indexes are built `CONCURRENTLY`; the `ADD COLUMN` is wrapped in `safety_assured`
  for `strong_migrations`.
- Adding a tracked field trips dfe-analytics' boot check — hence the blocklist commit;
  the app/CI won't boot until the field is registered.
- `schema.rb` carries the generated column as
  `t.virtual "geo_location", type: :geography, … stored: true`. A **fresh**
  `db:schema:load` reproduces it correctly.

---

## Testing

- Verified on the restored production DB that flag-on results are identical to the
  previous PR (same course ids, same distances) and that the plan uses the GiST index.
- Unit specs for the new-model path (radius filtering, nearest-of-many, radius
  exclusion, NULL-coordinate guard, duplicate edges, subject composition, `#count`)
  cover the query behaviour.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_07_17_120000) do
+ActiveRecord::Schema[8.1].define(version: 2026_07_21_100824) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "btree_gist"
@@ -264,7 +264,6 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_17_120000) do
     t.datetime "updated_at", null: false
     t.index ["course_id", "provider_school_id"], name: "index_course_school_on_course_id_and_provider_school_id", unique: true
     t.index ["gias_school_id"], name: "index_course_school_fanout", include: ["course_id"]
-    t.index ["gias_school_id"], name: "index_course_school_on_gias_school_id"
     t.index ["provider_school_id"], name: "index_course_school_on_provider_school_id"
   end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_07_14_120500) do
+ActiveRecord::Schema[8.1].define(version: 2026_07_17_120000) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "btree_gin"
   enable_extension "btree_gist"
@@ -263,6 +263,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_14_120500) do
     t.bigint "provider_school_id", null: false
     t.datetime "updated_at", null: false
     t.index ["course_id", "provider_school_id"], name: "index_course_school_on_course_id_and_provider_school_id", unique: true
+    t.index ["gias_school_id"], name: "index_course_school_fanout", include: ["course_id"]
     t.index ["gias_school_id"], name: "index_course_school_on_gias_school_id"
     t.index ["provider_school_id"], name: "index_course_school_on_provider_school_id"
   end
@@ -355,6 +356,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_14_120500) do
     t.text "address3"
     t.text "county"
     t.datetime "created_at", null: false
+    t.virtual "geo_location", type: :geography, limit: {srid: 4326, type: "st_point", geographic: true}, as: "(st_setsrid(st_makepoint(longitude, latitude), 4326))::geography", stored: true
     t.text "group_code"
     t.float "latitude"
     t.float "longitude"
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -263,6 +263,7 @@
     t.bigint "provider_school_id", null: false
     t.datetime "updated_at", null: false
     t.index ["course_id", "provider_school_id"], name: "index_course_school_on_course_id_and_provider_school_id", unique: true
+    t.index ["gias_school_id"], name: "index_course_school_fanout", include: ["course_id"]
     t.index ["gias_school_id"], name: "index_course_school_on_gias_school_id"
```
inulty-dfe: Are both these indexes needed or can one be removed

tomas-stefano: Good spot. I removed in [464bba](https://github.com/DFE-Digital/publish-teacher-training/pull/6235/commits/9464bbac21f04dcde7a657947b0861d18c9f6134)

inulty-dfe (approved):
Looks good to me

PR #6238, opened 2026-07-20
Filter discarded sites from location search

tomas-stefano:
## Context

When a training provider removes a placement school, Find kept using that removed school to place the course on the map. So a course could appear in a location search because of a school that isn't there anymore.

Location search (Find results + saved courses) was positioning courses using placement schools the provider had already removed.

Both `Courses::Query#location_scope` and `SavedCourses::Query#location_scope` build their location filter in raw SQL and join `site` with no `discarded_at IS NULL` predicate. `Site` includes `Discard::Model` but discard adds no `default_scope`, and raw SQL bypasses AR scopes anyway. Discarding a site also doesn't detach its `course_site` rows — they survive with `status='R'`, `publish='Y'`, so they keep satisfying the join.

**Impact (2026 cycle, measured):** 180 discarded sites still attached to 700 live published courses; 147 of those are findable *only* via discarded schools.

## Change

Added `AND site.discarded_at IS NULL` to the `site` join in both query objects:

```sql
INNER JOIN site ON (
  site.id = site_statuses.site_id
  AND site.discarded_at IS NULL
)
```

Courses are now positioned only by schools that still exist; courses propped up solely by a discarded school drop out of location results (the correct behaviour).

### DIFF
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -303,7 +303,10 @@ module Courses
             AND site_statuses.status = 'R'
             AND site_statuses.publish = 'Y'
           )
-          INNER JOIN site ON site.id = site_statuses.site_id
+          INNER JOIN site ON (
+            site.id = site_statuses.site_id
+            AND site.discarded_at IS NULL
+          )
         SQL
         .where("(site.longitude IS NOT NULL OR site.latitude IS NOT NULL)")
         .where(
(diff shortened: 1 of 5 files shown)


tomas-stefano:
https://trello.com/c/XjpExCaW/1815-find-bug-discard-sites

langesi (approved):

langesi (changes requested):
Need to rebase, after I test the deploy I'll approve, code looks good

tomas-stefano:
> Need to rebase, after I test the deploy I'll approve, code looks good

Done. It is building now ...

langesi (approved):
LGTM, did some spot checks also in the review app

PR #6239, opened 2026-07-21
Remove N+1 lookups from rollover school and site copiers

inulty-dfe:
## Context

Rollover runs many unnecessary queries N+1 and callbacks. We can optimise the service by removing these unnecessary queries.

## Guidance to review

Rollover completed successfully and  relatively quickly

image

## Checklist


gms-gs (approved):
LGTM! Good find on the touching inefficiency!

PR #6240, opened 2026-07-21
Update schools backfill to only run on cycles 2026 and later

inulty-dfe:
## Context

Update the Site => schools backfill migrator to only target sites from providers in the 2026 recruitment cycle and later. We have no interest in backfilling the schools for 2025 and earlier as their site data is poorly matched to gias schools.

It's unlikely that this will be run on the 2027 cycle, but it's not impossible.

## Checklist

### DIFF
--- a/app/services/data_hub/schools_backfill/executor.rb
+++ b/app/services/data_hub/schools_backfill/executor.rb
@@ -3,6 +3,10 @@
 module DataHub
   module SchoolsBackfill
     class Executor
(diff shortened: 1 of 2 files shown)


inulty-dfe:
[3.2 School Data - Migrate only given cycle](https://trello.com/c/WG2gHs5c/1818-32-school-data-migrate-only-given-cycle)

gms-gs (approved):

PR #6241, opened 2026-07-21
Provider School Removal

langesi:
## Context

Provider school removal currently identifies schools using the legacy site.id route parameter. As part of the schools remodel, provider schools now also exist in the new provider_school model with their own UUID.
For the 2026 remodel cycle, migrated site.uuid and provider_school.uuid values match, so provider-school removal should continue to use the legacy site UUID. After the remodel cycle, rollover will create new rows in both models and those UUIDs will drift, so provider-school removal needs to identify and delete the new Provider::School record by its own UUID.
This PR only changes provider-level school removal and identification. It does not change adding provider schools, or assigning/removing schools from courses.

- Adds schools_remodel_cycle_year: 2026 to settings.
- Changes Publish and Support provider-school member routes to use :uuid instead of the legacy :id.
- Adds ProviderSchools::Identity to centralise cycle-aware provider-school identity logic:
	- resolves the matching legacy Site and new Provider::School records
	- filters visible provider-school rows after the remodel cycle through existing provider_school rows
- Adds ProviderSchools::Removal for deletion behaviour:
	- in the 2026 cycle, removes both the legacy Site and matching Provider::School
	- after 2026, removes only the Provider::School
	- blocks removal when the provider school is attached to course schools
- Updates Publish and Support provider-school show/delete/index links to pass the correct UUID.
- Updates the course query debug school link to use UUIDs selected by Courses::NearestSchoolQuery, avoiding an extra Site.find.

## Guidance to review

- We can do a bug party overall for rollover, check if the current functionality for this cycle breaks.
- Confirm provider-school show/delete URLs now contain UUIDs rather than site.id.
- In a 2026 recruitment cycle, remove a provider school and confirm both site and matching provider_school rows are removed.
- Confirm provider schools attached to course schools cannot be removed.
- Confirm add-school flows still behave unchanged.

## Checklist

### DIFF
--- a/app/controllers/publish/providers/schools_controller.rb
+++ b/app/controllers/publish/providers/schools_controller.rb
@@ -3,7 +3,7 @@
 module Publish
   module Providers
     class SchoolsController < ApplicationController
-      before_action :site, only: %i[show delete]
+      before_action :site, only: %i[show delete destroy]
       before_action :reset_urn_form, only: %i[index]
 
       PER_PAGE = 20
@@ -27,11 +27,13 @@ module Publish
       def delete; end
 
       def destroy
-        school_removal.call
-        flash[:success] = "School removed"
-        redirect_to publish_provider_recruitment_cycle_schools_path
-      rescue ProviderSchools::Removal::CannotRemoveSchoolError
-        render :delete, status: :unprocessable_entity
+        if school_removal.call
+          flash[:success] = "School removed"
+          redirect_to publish_provider_recruitment_cycle_schools_path
+        else
+          redirect_to delete_publish_provider_recruitment_cycle_school_path(@provider.provider_code, @site.recruitment_cycle.year, @site.uuid),
+                      flash: { warning: t(".cannot_remove_school") }
+        end
       end
 
     private
--- a/app/controllers/publish/providers/schools_controller.rb
+++ b/app/controllers/publish/providers/schools_controller.rb
@@ -9,7 +9,7 @@ module Publish
       PER_PAGE = 20
 
       def index
-        @pagy, @schools = pagy(provider.sites.order(:location_name), limit: PER_PAGE)
+        @pagy, @schools = pagy(ProviderSchools::Identity.visible_sites(provider:).order(:location_name), limit: PER_PAGE)
       end
 
       def show; end
@@ -27,15 +27,24 @@ module Publish
       def delete; end
 
       def destroy
-        site.destroy!
-        flash[:success] = "School removed"
-        redirect_to publish_provider_recruitment_cycle_schools_path
+        if school_removal.removable?
+          school_removal.call
+          flash[:success] = "School removed"
+          redirect_to publish_provider_recruitment_cycle_schools_path
+        else
+          render :delete, status: :unprocessable_entity
+        end
       end
 
     private
 
+      def school_removal
+        @school_removal ||= ProviderSchools::Removal.new(provider:, uuid: params[:uuid])
+      end
+      helper_method :school_removal
+
       def site
-        @site ||= provider.sites.find(params[:id])
+        @site ||= school_removal.site
       end
 
       def site_params(param_form_key)
--- a/app/controllers/publish/providers/schools_controller.rb
+++ b/app/controllers/publish/providers/schools_controller.rb
@@ -9,7 +9,7 @@ module Publish
       PER_PAGE = 20
 
       def index
-        @pagy, @schools = pagy(ProviderSchools::Identity.visible_sites(provider:).order(:location_name), limit: PER_PAGE)
+        @pagy, @schools = pagy(ProviderSchools::Identity.ordered_school_scope(provider:), limit: PER_PAGE)
       end
 
       def show; end
@@ -27,13 +27,11 @@ module Publish
       def delete; end
 
       def destroy
-        if school_removal.removable?
-          school_removal.call
-          flash[:success] = "School removed"
-          redirect_to publish_provider_recruitment_cycle_schools_path
-        else
-          render :delete, status: :unprocessable_entity
-        end
+        school_removal.call
+        flash[:success] = "School removed"
+        redirect_to publish_provider_recruitment_cycle_schools_path
+      rescue ProviderSchools::Removal::CannotRemoveSchoolError
+        render :delete, status: :unprocessable_entity
       end
 
     private
@@ -44,7 +42,7 @@ module Publish
       helper_method :school_removal
 
       def site
-        @site ||= school_removal.site
+        @site ||= school_removal.school
       end
 
       def site_params(param_form_key)
--- a/app/controllers/support/providers/schools_controller.rb
+++ b/app/controllers/support/providers/schools_controller.rb
@@ -11,7 +11,7 @@ module Support
       PER_PAGE = 20
 
       def index
-        @pagy, @sites = pagy(provider.sites.order(:location_name), limit: PER_PAGE)
+        @pagy, @sites = pagy(ProviderSchools::Identity.visible_sites(provider:).order(:location_name), limit: PER_PAGE)
       end
 
       def show; end
@@ -30,9 +30,12 @@ module Support
       end
 
       def destroy
-        site.destroy!
-
-        redirect_to support_recruitment_cycle_provider_schools_path(provider.recruitment_cycle_year, provider), flash: { success: t("support.flash.deleted", resource: flash_resource) }
+        if school_removal.removable?
+          school_removal.call
+          redirect_to support_recruitment_cycle_provider_schools_path(provider.recruitment_cycle_year, provider), flash: { success: t("support.flash.deleted", resource: flash_resource) }
+        else
+          render :delete, status: :unprocessable_entity
+        end
       end
 
     private
@@ -59,12 +62,17 @@ module Support
       end
 
       def site
-        @site ||= provider.sites.find(params[:id])
+        @site ||= school_removal.site
       end
 
       def reset_urn_form
         URNForm.new(provider).clear_stash
       end
+
+      def school_removal
+        @school_removal ||= ProviderSchools::Removal.new(provider:, uuid: params[:uuid])
+      end
+      helper_method :school_removal
     end
   end
 end
--- a/app/controllers/support/providers/schools_controller.rb
+++ b/app/controllers/support/providers/schools_controller.rb
@@ -6,7 +6,7 @@ module Support
       before_action :build_site, only: %i[index create]
       before_action :new_form, only: %i[index]
       before_action :reset_urn_form, only: %i[index]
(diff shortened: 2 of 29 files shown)


spec/services/provider_schools/identity_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ProviderSchools::Identity do
+  let(:remodel_cycle_year) { 2026 }
+  let(:gias_school) { create(:gias_school, urn: "123456") }
+  let(:site_uuid) { "11111111-1111-4111-8111-111111111111" }
```
langesi: I know I could use faker, but it feels like wasting time taking care of nitpicks like this

app/controllers/publish/providers/schools_controller.rb
```diff
@@ -27,15 +27,22 @@ def create
       def delete; end
 
       def destroy
-        site.destroy!
+        school_removal.call
         flash[:success] = "School removed"
         redirect_to publish_provider_recruitment_cycle_schools_path
+      rescue ProviderSchools::Removal::CannotRemoveSchoolError
+        render :delete, status: :unprocessable_entity
```
inulty-dfe: I don't think we should be using exception driven control flow. 

```ruby
if school_removal.call
  ...
else
  render :delete ...
end
```

app/controllers/support/providers/schools_controller.rb
```diff
@@ -30,9 +30,10 @@ def delete
       end
 
       def destroy
-        site.destroy!
-
+        school_removal.call
         redirect_to support_recruitment_cycle_provider_schools_path(provider.recruitment_cycle_year, provider), flash: { success: t("support.flash.deleted", resource: flash_resource) }
+      rescue ProviderSchools::Removal::CannotRemoveSchoolError
+        render :delete, status: :unprocessable_entity
```
inulty-dfe: https://github.com/DFE-Digital/publish-teacher-training/pull/6241/changes#r3630559415

app/decorators/provider/school_decorator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class Provider::SchoolDecorator < Draper::Decorator
+  include MarkdownHelper
```
inulty-dfe, line 4: I understand why we need to do this

spec/system/publish/providers/schools/delete_school_spec.rb
```diff
@@ -45,8 +45,47 @@
     and_the_school_is_deleted
   end
 
+  scenario "after the schools remodel cycle without a legacy site" do
+    recruitment_cycle = create(:recruitment_cycle, year: 2027)
```
inulty-dfe: We should use the `Settings.schools_remodel_cycle_year` in the specs too

spec/system/publish/providers/schools/delete_school_spec.rb
```diff
@@ -45,8 +45,47 @@
...
+      name: "Future School",
+      urn: "654321",
+      address1: "1 Future Road",
+      address2: "Future Building",
+      address3: "Future Quarter",
+      town: "Future Town",
+      county: "Future County",
+      postcode: "FT1 1AA",
+    )
+    provider_school = create(:provider_school, provider: future_provider, gias_school:, site_code: "-")
```
inulty-dfe: This violates the style guide we have for writing system specs

app/controllers/publish/providers/schools_controller.rb
inulty-dfe: I see why it's here now, but ideally the error wouldn't leave the service and get into the controller

app/models/provider/school.rb
```diff
@@ -23,4 +26,25 @@ class Provider::School < ApplicationRecord
               message: :only_one_main_site_per_provider,
             },
             if: -> { site_code == MAIN_SITE_CODE }
+
+  def location_name
+    return "#{gias_school.name} (Main Site)" if site_code == MAIN_SITE_CODE
```
inulty-dfe:
```suggestion
    return "#{gias_school.name} (Main site)" if site_code == MAIN_SITE_CODE
```

app/services/provider_schools/removal.rb
```diff
@@ -0,0 +1,80 @@
...
+                           else
+                             identity.provider_school_for(site:)
+                           end
+    rescue ActiveRecord::RecordNotFound
+      raise if after_schools_remodel_cycle?
+
+      nil
+    end
+
+    def uuid_for_path
```
inulty-dfe: Why is this `_for_path`, we should just leave it as a delegate to school. Having convenience accessors on a command object really confuses the purpose of it

spec/system/publish/providers/schools/delete_school_spec.rb
langesi: Was hoping I could get away with it 😆 . I'll get it sorted

spec/services/provider_schools/identity_spec.rb
langesi: Did the thing, it uses Faker now

app/controllers/publish/providers/schools_controller.rb
langesi: Thanks for the comment, it's not exception driven anymore, we show a banner and redirect

app/decorators/provider/school_decorator.rb
langesi: We use Markdown methods at the moment from the existing decorator for sites, this is just the same logic as for the school logic

app/services/provider_schools/identity.rb
```diff
@@ -0,0 +1,44 @@
...
+      new(provider:).ordered_school_scope
+    end
+
+    def initialize(provider:)
+      @provider = provider
+    end
+
+    def school_scope
+      after_schools_remodel_cycle? ? provider.schools.includes(:gias_school) : provider.sites
+    end
```
inulty-dfe: I don't see where this is used

app/services/provider_schools/identity.rb
```diff
@@ -0,0 +1,44 @@
...
+    def provider_school_for(site: nil, uuid: nil)
+      return provider.schools.find_by!(uuid:) if uuid.present?
+
+      provider
+        .schools
+        .joins(:gias_school)
+        .find_by!(gias_school: { urn: site.urn }, site_code: site.code)
+    end
+
+    def after_schools_remodel_cycle?
```
inulty-dfe, line 40:
```suggestion
    def school_locations?
```

Within the service it's not important for the cycle semantics to be present and is a bit distracting. Can we rename it to be a bit dummer "are we using sites or schools?" and ignore the why.

I can see one place where it doesn't make 100% sense to change it but overall I think it reduces cognitive load. WDYT?

app/services/provider_schools/removal.rb
```diff
@@ -0,0 +1,84 @@
...
+      true
+    end
+
+    def destroy_records!
+      if after_schools_remodel_cycle?
+        provider_school.destroy!
+      else
+        provider_school.destroy!
+        site.destroy!
+      end
```
inulty-dfe:
```suggestion
      provider_school.destroy!
      unless after_schools_remodel_cycle?
        site.destroy!
      end
```

No else needed

app/services/provider_schools/identity.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def school_scope
+      after_schools_remodel_cycle? ? provider.schools.includes(:gias_school) : provider.sites
+    end
+
+    def ordered_school_scope
+      return provider.sites.order(:location_name) unless after_schools_remodel_cycle?
+
+      provider.schools.joins(:gias_school).includes(:gias_school).order("gias_school.name")
+    end
```
inulty-dfe, line 23: There's 3 methods in this class that test `after_schools_remodel_cycle?` (although i think one doesn' get used)

Two are ternaries the other is a guard clause.

Can we make them all the same shape. In this case I'd say `if..else`. That way its easy to understand them at a  glance.

Consistency over everything IMO

app/services/provider_schools/identity.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def school_for(uuid:)
+      after_schools_remodel_cycle? ? provider.schools.find_by!(uuid:) : provider.sites.find_by!(uuid:)
+    end
+
+    def provider_school_for(site: nil, uuid: nil)
```
inulty-dfe: We don't call this method passing the uuid anywhere

inulty-dfe (commented):
I'm stopping for tonight, lets pick it up later

app/services/provider_schools/identity.rb
langesi: I didn’t think the naming was bad. This service determines whether to use the Provider::School table or the Site table, which is why it was named ProviderSchool::Identity. 

It will primarily be used to decide whether a record should be saved only in the new model or in the old one. This method specifically determines which model to use based on whether the school remodeling cycle year, currently 2026 has passed.

The decision is based on the `schools_remodel_cycle_year` setting, which I took from your API PR.

So the thought is that the Identity service tells you what School model to choose after the remodel cycle. Also looking at the if statements that use this, I don't think `school_locations?` is more clear and it's harder to read.

Would yard documentation help to make this clearer or do you still feel strongly about this? We can discuss this tomorrow

app/services/provider_schools/removal.rb
langesi: Thought of doing this before pushing the code but thought it's easy to read the way it is.

Don't feel to strong about it, but if that's stopping the PR from being approved I can change it

app/services/provider_schools/identity.rb
langesi: The choice of ternary and guard clause was purely due to the size. 

If I had written the the guard clause as ternary it would have been too long for my liking.

I rewrote the guard clause to and if else statement, however I'd like to keep the ternary, I don't think they are hard to understand and it keeps the code more concise.

app/services/provider_schools/identity.rb
langesi: It isn't. I removed the use of it but didn't remove the method

app/services/provider_schools/identity.rb
inulty-dfe: Yeah I can understand that. I think that if there are differences in how we express the same control flow within the same file we should reduce the the common denominator.

If one is too long for a ternary, none should be ternaries (unless they're like 15 chars long or something ridiculous like that.)

If one is too long and the others are kind of long then make them all the same shape

inulty-dfe (approved):

gms-gs (approved):
Goood work on this!! LGTM

PR #6244, opened 2026-07-22
Publish course list filters

tomas-stefano:
## Context

When providers manage their courses during a cycle, the course list mixes live courses in with `draft`, `rolled over` and `withdrawn` ones, which makes it hard to find what they are looking for. Providers also want to see groups of courses at a glance — all apprenticeship courses, all primary courses, and so on.

This adds a filter sidebar, removable active-filter chips, a "N selected" indicator per group, and a course count to the provider course list. It can ship immediately — no feature flag, no waiting for rollover.

## What a provider sees

- A **Filter courses** sidebar on the left with checkbox groups: Status,Education phase, Fee or salary, Qualification, Full time or part time, and Start date (every month in the cycle).
- **Active filter** chips showing what is applied, each removable, plus a "Clear all filters" link — styled the same as the Find search results.
- A **"N selected"** hint on each group that has a selection.
- A **course count** above each accredited-provider group, e.g. "3 courses", that updates as filters are applied.

Filters are applied with an **Apply filters** button, so the page works without JavaScript.

## How it works

The page already had a read model (`Publish::Courses::Query` → `Publish::CourseList` → view components) that derives each course's status in SQL from plain columns — no enrichment or school/site rows are loaded. This work adds filtering on top of that read model. **It does not depend on the new schools tables
(`provider_school` / `course_school`)** and is independent of the schools-remodel feature flag.

```mermaid
flowchart TD
    A["Provider ticks filters<br/>and clicks Apply"] --> B["GET /courses?status[]=…&level[]=…"]
    B --> C["CoursesController#index"]
    C --> D["CourseFilterParams.permit<br/>(allow only known keys)"]
    D --> E["CourseFilterForm<br/>(drop unknown values,<br/>build chips + counts)"]
    E --> F["CourseList.new(provider, params)"]
    F --> G["Publish::Courses::Query<br/>(one scope per filter,<br/>bound values only)"]
    G --> H["Grouped course rows"]

    E --> I["FilterPanelComponent<br/>sidebar + chips"]
    H --> J["ListComponent<br/>tables + per-group count"]
    I --> K["Rendered page"]
    J --> K
```

### Where each filter maps

| Filter | Source |
|---|---|
| Education phase, Fee or salary, Qualification, Full time or part time | Plain `course` columns |
| Start date | `course.start_date` matched by month range |
| Status (Open / Closed / Draft / Rolled over / Scheduled / Withdrawn) | `content_status` (from `course_enrichment`) + application status + recruitment cycle |

Status is the interesting one: the tag a provider sees is a function of the course's content status, whether it is accepting applications, and whether the page's cycle is current or in the future. That logic already lives in the status
tag component; this PR extracts the shared "which cycle" rule (`Publish::Courses::CycleBranch`) so the **filter and the tag can never disagree** — a course is returned by a status filter exactly when the list would render it with that tag. A test renders the real tag and asserts this.

### Reused, not duplicated

The active-filter chips already existed on Find. Rather than copy them, the component was **promoted out of `Find::` into a shared `Courses::ActiveFilters::View`**that takes an injected path builder, so Find and Publish render identical chips against their own URLs. The chip styles moved into the shared stylesheet folder. Find's behaviour is unchanged — its existing system specs pass untouched, and the old link URLs were pinned with tests *before* the move to prove it.

### One subtlety: the "course information" column

The list hides a course-information field (funding, qualification, etc.) when it is identical for every course, since a uniform value carries no information. That decision is now made from the **whole, unfiltered list** and held fixed while
filtering:

- If the full list mixes Fee-paying and Apprenticeship, the funding line shows — and keeps showing when you filter to just Fee-paying.
- If the full list is entirely Fee-paying, the funding line stays hidden whatever you filter by.

## Safety

- **No SQL injection.** Filter values are bound as parameters, never interpolated. The one unavoidable piece of SQL text (the status `CASE`) is a frozen constant with all values bound via Arel. Brakeman is clean, and tests push `' OR 1=1 --` and `'); DROP TABLE course; --` through every filter and assert nothing matches and the data is intact.
- **Unknown values are ignored** at a single choke point in the form, so a junk query string reaches neither the SQL nor a chip. A request test confirms a `<script>` payload is not reflected into the page.
- **No extra queries on first load.** The unfiltered-column decision reuses the rows already loaded; only a filtered page runs one extra lightweight query.

## Testing

Built test-first throughout. Coverage added across:

- **Query** — each filter (single, multiple, combined), the status matrix across cycles, month boundaries, hostile input, and a constant-query-count guard.
- **Form** — value allow-listing, counts, chip order and removal, start-date options.
- **Components** — the shared chips (Find URLs pinned + Publish URLs), the filter panel, the "N selected" hint.
- **Request** — each filter narrowing the rendered list, unknown params ignored, course counts, and the course-information column behaviour under filtering.
- **System** — end-to-end filtering, chips, clear-all, counts and the empty state, all without JavaScript.


tomas-stefano:
https://trello.com/c/Aot8AB8q/1691-course-list-add-filters-and-count-for-number-of-courses-displayed

app/assets/stylesheets/publish/components/_table.scss
```diff
@@ -34,6 +34,13 @@
     }
   }
 
+  // Hairline under each group's course count.
+  &__count {
+    padding-bottom: govuk-spacing(2);
+
+    border-bottom: 1px solid #cecece;
```
inulty-dfe: Is this hex value in the govuk frontend pallete?

app/components/courses/active_filters/view.rb
```diff
@@ -0,0 +1,33 @@
+module Courses
+  module ActiveFilters
+    # The removable chips summarising which filters are applied, plus the link
+    # that clears them all.
+    #
+    # Shared by Find and Publish, which list different things at different URLs:
+    # the caller passes a +path_builder+ that turns a params hash into a link
+    # back to its own list, so this component never names a route itself.
+    class View < ViewComponent::Base
```
inulty-dfe: Prefer not to name `view.rb` and `view.html.erb`. Especially as that pattern has no precedent in this directory.

Image

app/forms/publish/course_filter_form.rb
```diff
@@ -0,0 +1,109 @@
...
+
+module Publish
+  # The filters on the publish course list. Modelled on Courses::SearchForm (the
+  # Find search form) but deliberately separate: none of that form's subject,
+  # provider or location machinery applies here.
+  #
+  # Every group is a multi-select list of checkboxes, and each reader allows only
+  # the values that group offers. That single choke point means an unrecognised
+  # value in the query string reaches neither the SQL nor an active filter chip.
+  class CourseFilterForm < ApplicationForm
```
inulty-dfe: This feels more like a `Courses::FilterForm` than `CourseFilterForm`.

`Course` is more member and `Courses` more collection. In fact you reference a perfect example in `Courses::SearchForm`

app/forms/publish/course_filter_form.rb
```diff
@@ -0,0 +1,109 @@
...
+          remove_params: ::Courses::ActiveFilters::RemovalParams.new(
+            search_params: filter_params, attribute: group, current_value: value, all_values: values,
+          ).call,
+        )
+      end
+    end
+
+    def start_date_options
+      @start_date_options ||= ::Courses::CycleStartMonths.for(provider.recruitment_cycle_year).map do |month|
+        Option.new(value: month.strftime("%Y-%m"), label: I18n.l(month, format: :short))
```
inulty-dfe: Can we put this strftime in the DATE_FORMAT and reference it if it's not there already.

`config/initializers/date_formats.rb`

app/services/publish/courses/cycle_branch.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Publish
+  module Courses
+    # Which status vocabulary a recruitment cycle year uses. Courses in the
+    # current or previous cycle are Open/Closed depending on their application
+    # status; courses in a future cycle are Scheduled instead.
+    #
+    # Shared by StatusTagComponent (which renders the tag) and Query (which
+    # filters on it) so the two cannot disagree about what a status means.
+    module CycleBranch
```
inulty-dfe: Should this go in the cycle timetable?

inulty-dfe (commented):
Great PR, very easy to review ⭐

kelliedesigner:
Looks good to me @tomas-stefano !

app/services/publish/courses/cycle_branch.rb
tomas-stefano: Done

inulty-dfe (approved):

PR #6248, opened 2026-07-23
Add identity service for Course::School.

langesi:
## Context

This is the first, preparatory PR for creating and removing course_schools.

This PR does not wire the service into course creation or course school updates yet. It only introduces the switching contract and its coverage.

- Adds `CourseSchools::Identity`.
- Uses `Site` records for recruitment cycles before and during the schools remodel cycle.
- Uses `Provider::School` records for recruitment cycles after the schools remodel cycle.
- Adds strict validation when resolving submitted school identifiers:
  - unknown identifiers raise an error, only uuid permitted
  - identifiers from another provider raise an error
  - duplicate identifiers are deduplicated
- Adds a clear error when course-specific methods are called without a course.
- Adds specs for the before, during, and after remodel cycle boundaries.

No UI changes.

## Guidance to review

I suggest reviewing this PR in two parts:

1. `app/services/course_schools/identity.rb`
   - Check whether the cycle boundary reads correctly:
     - before remodel cycle: `Site`
     - remodel cycle: `Site`
     - after remodel cycle: `Provider::School`
   - Check whether the method naming is clear

2. `spec/services/course_schools/identity_spec.rb`
   - Check the edge-case coverage for unresolved identifiers, another provider’s schools, duplicate identifiers, mixed ID/UUID input, and missing course usage.

## Checklist

### DIFF
--- a/app/services/course_schools/identity.rb
+++ b/app/services/course_schools/identity.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+module CourseSchools
+  # This service determines which model should be used when creating, removing and in some instances viewing schools.
+  # The legacy data model uses Site, while the new data model uses Course::School.
+  # During the rollover, both models are written to and their UUIDs may diverge, so we need a way to switch to the new data model.
+  # Because Publish supports both the old and new recruitment cycles during the rollover, this switch must be based on the recruitment cycle rather than a traditional feature flag.
+  class Identity
+    delegate :after_schools_remodel_cycle?, to: :provider_identity
+    delegate :count, to: :available_schools, prefix: true
+
+    def initialize(provider:, course: nil)
+      @provider = provider
+      @course = course
+      @provider_identity = ProviderSchools::Identity.new(provider:)
+    end
+
+    def available_schools
+      provider_identity.ordered_school_scope.to_a
+    end
+
+    def current_school_uuids
+      if after_schools_remodel_cycle?
+        course.schools.includes(:provider_school).map { |course_school| course_school.provider_school.uuid.to_s }
+      else
+        course.sites.map { |site| site.uuid.to_s }
+      end
+    end
+
+    def school_records_for(school_uuids:)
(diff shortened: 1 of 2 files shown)


gms-gs (approved):

app/services/course_schools/identity.rb
```diff
@@ -0,0 +1,95 @@
...
+    delegate :count, to: :available_schools, prefix: true
+
+    def initialize(provider:, course: nil)
+      @provider = provider
+      @course = course
+      @provider_identity = ProviderSchools::Identity.new(provider:)
+    end
+
+    def available_schools
+      provider_identity.ordered_school_scope.to_a
```
inulty-dfe: Is it necessary to call `to_a` on a provider with 2,000 schools?

app/services/course_schools/identity.rb
```diff
@@ -0,0 +1,95 @@
...
+        course.sites.map { |site| site.uuid.to_s }
+      end
+    end
+
+    def school_records_for(school_uuids:)
+      school_uuids = normalize_school_uuids(school_uuids)
+      validate_uuid_values!(school_uuids)
+
+      records_by_uuid = school_records_by_uuid(school_uuids)
+      validate_all_uuids_resolved!(school_uuids, records_by_uuid)
```
inulty-dfe, line 39: Why do we need to validate that all the uuids match records?

inulty-dfe (commented):
I don't know why we need to validate in the identity class

app/services/course_schools/identity.rb
langesi: It would allow us to see if there is any unexpected behaviour such as UUIDs or IDs that shoudn't be in there, if we don't validate them we might have a bug that we do not see

app/services/course_schools/identity.rb
langesi: Good point! Might be bad for our memory! I'll change this!

PR #6250, opened 2026-07-23
REWORKED - School data - Update Schools in Basic details component

langesi:
## Context

This is the course school update part of the school data remodel. It only takes care of the school adding in the basic details component of a course in publish.

The legacy `Site` and corresponding `Provider::School` are expected to share the same UUID. This allows one submitted `Provider::School` selection to be applied consistently to both models.

Inline and queued updates are both transactional. Their handling of a missing `Provider::School` differs deliberately:

- Inline requests fail and show a form error because the submitted selection is no longer valid.
- Queued requests log and exclude `Provider::School` records removed while the job was waiting.
- The resulting effective selection is then written atomically to both models.
- A valid `Provider::School` without its expected legacy `Site` is treated as data drift and raises rather than silently leaving the models inconsistent.

- Updates the existing course schools form to display available `Provider::School` records and submit `school_uuids` instead of legacy Site IDs.
- Reads the course's current selection and attached-school count from `Course::School`.
- Uses the same `GiasSchool.available` scope for form display and validation, excluding closed schools.
- Validates that submitted UUIDs belong to an available school for the course provider.
- Reports unresolved inline selections to Sentry and displays the existing form error.
- Refactors `Publish::Schools::UpdateCourseSchoolsService` to:
  - resolve submitted `Provider::School` UUIDs once;
  - lock resolved records while the update is applied;
  - pass the same effective selection to both writers;
  - keep inline and queued dual writes within one transaction;
  - persist attributes such as `schools_validated`;
  - save the course so `provider.changed_at` is updated for downstream Apply synchronisation;
  - send school-update notifications from the resulting `Course::School` selection.
- Adds `Publish::Schools::UpdateCourseProviderSchoolsService` to synchronise `Course::School` records from already-resolved `Provider::School` records.
- Adds `Publish::Schools::UpdateCourseSiteStatusesService` to temporarily mirror the same UUID selection to legacy `SiteStatus` records.
- Makes missing legacy Site mappings explicit instead of silently skipping them.
- Updates the queued job to tolerate stale `Provider::School` UUIDs without disabling transaction protection for other failures.
- Counts unique, non-blank UUIDs consistently when deciding whether to enqueue and which success message to show.
- Adds a Site factory trait that creates a corresponding GIAS school and `Provider::School` with matching school data and UUID.
- Adds coverage for:
  - adding, removing and replacing schools in both models;
  - removing every school where this is permitted;
  - stale queued selections;
  - transaction rollback after writer or course-save failures;
  - missing legacy Site mappings;
  - inline form errors and Sentry reporting;
  - notifications;
  - provider `changed_at` updates;
  - dual writing after the remodel cycle;
  - closed schools;
  - queue boundaries, duplicate UUIDs and blank values;
  - select-all and collapsed school lists.

There are no intended visual changes beyond the form being backed by the remodelled school data.

## Guidance to review

Suggested review order:

### 1. Existing course edit form and validation

- `app/forms/publish/course_school_form.rb`
- `app/controllers/publish/courses/schools_controller.rb`
- `app/views/publish/courses/schools/edit.html.erb`
- `config/locales/en.yml`
- `app/decorators/course_decorator.rb`
- `spec/forms/publish/course_school_form_spec.rb`
- `spec/decorators/course_decorator_spec.rb`
- `spec/support/page_objects/publish/course_school_edit.rb`
- `spec/system/publish/courses/editing_course_schools_spec.rb`
- `spec/system/publish/courses/schools/school_list_display_spec.rb`

Check that the form displays available `Provider::School` records, submits UUIDs, restores its selection from `Course::School`, rejects unavailable or unrelated schools, and renders a normal form error when an inline update can no longer resolve the selection.

### 2. Update orchestration and queued behaviour

- `app/services/publish/schools/update_course_schools_service.rb`
- `app/jobs/update_course_schools_job.rb`
- `spec/services/publish/schools/update_course_schools_service_spec.rb`
- `spec/jobs/update_course_schools_job_spec.rb`

Check that submitted UUIDs are resolved once and that both writers receive the same effective selection.

Pay particular attention to the queued path: stale `Provider::School` UUIDs are logged and excluded, but the remaining dual write is still transactional. Unexpected writer, validation or course-save failures therefore roll back both models.

Also check the comment around `course.save!`. The save is required to update `provider.changed_at`, which Apply uses to decide whether the provider's courses need synchronising.

### 3. New Course::School synchronisation

- `app/services/publish/schools/update_course_provider_schools_service.rb`
- `spec/services/publish/schools/update_course_provider_schools_service_spec.rb`

Check that this service has one responsibility: synchronising `Course::School` rows from an already-resolved collection of `Provider::School` records.

UUID resolution and missing-school policy deliberately remain in the orchestrator.

### 4. Legacy SiteStatus synchronisation

- `app/services/publish/schools/update_course_site_statuses_service.rb`
- `spec/services/publish/schools/update_course_site_statuses_service_spec.rb`

Check that this temporary writer applies the same effective UUID selection to legacy `SiteStatus` records.

The important invariant is that matching `Site` and `Provider::School` records share a UUID. Missing Site mappings now raise before mutation rather than being silently skipped.

Also check that published status is applied only to active SiteStatus rows and that suspended or discontinued rows are not reactivated.

### 5. Queue boundaries and user journeys

- `spec/system/publish/courses/schools/select_all_schools_spec.rb`
- `spec/system/publish/courses/publishing_a_course_without_schools_spec.rb`
- `spec/system/publish/providers/courses/schools_rollover_validation_spec.rb`
- `spec/system/publish/providers/newly_added_tag_add_schools_to_a_course_page_spec.rb`

Check select-all and collapsed-list behaviour, the asynchronous path above 30 unique schools, rollover validation, newly added labels, and providers that are allowed to remove every school.

### 6. Supporting test and course-creation changes

- `spec/factories/sites.rb`
- `app/services/courses/creation_service.rb`
- `spec/services/courses/creation_service_spec.rb`

Check that the Site factory trait creates matching legacy and remodelled school records with the same UUID, URN and location information. The course-creation changes in this PR are limited to keeping the existing creation behavior and comments consistent with the completed backfill.


app/services/publish/schools/update_course_site_statuses_service.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      # TODO School data remodel removal - remove once submitted school UUIDs point directly at Provider::School.
+      def sites_by_uuid
+        @sites_by_uuid ||= course.provider.sites
+          .where(uuid: sites_to_attach_uuids + sites_to_remove_uuids)
+          .index_by(&:uuid)
+      end
+
+      # TODO School data remodel removal - remove with the legacy SiteStatus write path.
```
tomas-stefano, line 74: sites_by_uuid now resolves through course.provider.sites, which is scoped `where(site_type: :school).kept`. The old code used an unscoped `Site.where(id: ...)`. `course.sites` (line 55) goes through `site_statuses` and has no `.kept` scope, so a discarded — or non-school-type — site that is still attached to a course lands in `sites_to_remove_uuids`, fails to resolve, and detach_school swallows it via `return if site.blank?`.

app/services/publish/schools/update_course_provider_schools_service.rb
```diff
@@ -0,0 +1,95 @@
...
+      class UnresolvedProviderSchoolsError < StandardError; end
+
+      def initialize(course:, school_uuids:, raise_on_missing_provider_schools: false)
+        @course = course
+        @submitted_school_uuids = Array(school_uuids).compact_blank.uniq
+        @raise_on_missing_provider_schools = raise_on_missing_provider_schools
+      end
+
+      def call
+        return unless all_submitted_provider_schools_resolved?
```
tomas-stefano: A single unresolvable UUID aborts the entire `Course::School` sync — including removals and including the attaches that did resolve. The old per-site loop skipped only the unmappable site.

Concretely: a provider with one URN-less or not-yet-backfilled site (the backfill at data_hub/schools_backfill/executor.rb:44-47 requires site_type = 0 AND urn IS NOT NULL) keeps that site ticked forever, so every subsequent course-school edit for that provider is a no-op on the new model — permanent silent divergence, not "reconciles later" as the comment claims. It's also now inconsistent with `Courses::CreationService#update_schools`, which still skips per-school. Note the stated justification ("a partial resolution cannot remove existing new-model rows") only holds `if site.uuid != provider_school.uuid` — see the next comment. Scope the guard to the attach set and let removals through, or state the divergence assumption explicitly.

app/services/publish/schools/update_course_site_statuses_service.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      def site_status_attributes
+        return { publish: :published, status: :running } if course.findable?
+
+        { publish: :unpublished, status: :new_status }
+      end
+      # rubocop:enable Style/CommentAnnotation, Lint/RedundantCopDisableDirective
+
+      def send_notifications
```
tomas-stefano: The constructor issues a course.sites query to build previous_site_names on every call, but the only production caller passes send_notifications: false (update_course_schools_service.rb:48). The whole notification block here is dead in production and duplicates `UpdateCourseSchoolsService#send_notifications`, including a second `FeatureFlag.active?` check.

app/services/publish/schools/update_course_provider_schools_service.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    private
+
+      attr_reader :course, :submitted_school_uuids
+
+      def attach_provider_school(provider_school)
+        course.schools.create!(
+          provider_school:,
+          gias_school: provider_school.gias_school,
+        )
```
tomas-stefano: Course::School#consistent_with_provider_school can fail it — that surfaces as a 500 rather than a  form error. Also worth noting the handler is unreachable when the request enqueues (>30 schools): the job raises inside Sidekiq and retries 25 times against an error that will never resolve on retry.

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,134 +1,139 @@
...
-        course.sites.reload
+      def update_course_and_provider_schools
+        ActiveRecord::Base.transaction do
+          # Saving the course here is intentionally not just about persisting
+          # schools_validated. Course saves update course.changed_at and then
+          # TouchProvider updates provider.changed_at; Apply watches
+          # provider.changed_at to decide whether to sync provider data.
+          course.assign_attributes(course_attributes)
+          update_provider_schools(raise_on_missing_provider_schools: true)
+          course.save! if course.has_changes_to_save?
```
tomas-stefano: The comment says the course.save! is there so changed_at/TouchProvider fire for Apply, but the if course.has_changes_to_save? guard means it doesn't fire when only schools changed. That case happens to be covered by Course::School's TouchCourse/after_destroy callbacks, not by this save — so the comment justifies a line that isn't doing what it claims.

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,134 +1,139 @@
...
+        UpdateCourseProviderSchoolsService.call(
+          course:,
+          school_uuids: school_uuids_for_provider_school_sync,
+          raise_on_missing_provider_schools:,
+        )
       end
 
-      def sites_to_attach_ids
-        @sites_to_attach_ids ||= submitted_site_ids - current_site_ids
+      def site_status_params
```
tomas-stefano: site_status_params → params_with_school_uuids → submitted_school_uuids(default:) →
  current_provider_update_uuids → current_site_uuids is a five-method chain whose correctness  depends on argument-evaluation order: current_site_uuids (:125) must memoise before UpdateCourseSiteStatusesService reloads course.sites, which only holds because line 48 is evaluated before line 49. Swapping those two lines silently changes the fallback, and no test covers it.

app/views/publish/courses/schools/edit.html.erb
```diff
@@ -24,57 +24,57 @@
...
+            tag: "h1",
+            size: "l",
+          },
+          form_group: { data: { controller: "select-all-checkboxes show-all-schools", show_all_schools_visible_value: @course_school_form.schools_collapse_threshold } } do %>
 
-          <p class="govuk-body">
-            <%= t("publish.courses.schools.attached_count", count: course.attached_schools_count) %>
-          </p>
+            <p class="govuk-body">
+              <%= t("publish.courses.schools.attached_count", count: course.attached_schools_count) %>
```
tomas-stefano, line 37: course.attached_schools_count (app/decorators/course_decorator.rb:172) switches on `FeatureFlag :course_publishing_uses_new_school_model`, while the checkbox list, validation, and write path in this PR all switch on `Settings.schools_remodel_cycle_year`. 

For a 2027 course with the flag off, the page lists Provider::Schools but reports `object.sites.school.count`, i.e. 0. One decision, two switches — this line needs to move onto the cycle-based switch (or ProviderSchools::Identity)

app/services/publish/schools/update_course_provider_schools_service.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+
+      def provider_school_ids_by_uuid
+        @provider_school_ids_by_uuid ||= course.provider.schools
+          .where(uuid: submitted_school_uuids)
+          .pluck(:uuid, :id)
+          .to_h
```
tomas-stefano: During the remodel cycle the submitted UUIDs are Site UUIDs, and this resolves them against
  provider_school.uuid. That assumes site.uuid == provider_school.uuid. 

But ProviderSchools::Identity — whose own class comment says "During the rollover, both models are written to and their UUIDs may diverge" — deliberately maps site→provider_school by
  gias_school.urn + site_code (provider_school_for, used by ProviderSchools::Removal). Two different mapping strategies now exist for the same relationship, and when they disagree this one degrades to a logger.warn that nothing alerts on.

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,134 +1,139 @@
...
 
-      def attach_school(site)
-        ::CourseSchools::LegacySiteStatusCreator.call(course:, site:)
-
-        gias_school = gias_schools_by_urn[site.urn]
-        return unless gias_school
+      def submitted_school_uuids(default_school_uuids:)
+        # Missing school_uuids means "leave the current schools unchanged";
+        # an explicit nil or empty array means "remove all schools".
+        return default_school_uuids unless params.key?(:school_uuids)
```
tomas-stefano: Jobs already on the Sidekiq queue carry {site_ids: [...]}. After deploy, `params.key?(:school_uuids)` is false, so `submitted_school_uuids` falls back to "leave schools unchanged" and the user's update is silently dropped. 

Worse, site_ids survives into `params.except(:school_uuids) → course.assign_attributes(site_ids: [...])`, and ids_writer on a `has_many :through` calls replace, which delete_alls the join rows — bypassing the suspend-vs-destroy semantics in `Course#remove_site!`

app/controllers/publish/courses/schools_controller.rb
```diff
@@ -29,7 +29,7 @@ def update
         if @course_school_form.valid?
           Publish::Schools::UpdateCourseSchoolsService.call_or_enqueue(course: @course, params: school_params)
 
-          flash[:success] = if Array(@course_school_form.site_ids).size > Publish::Schools::UpdateCourseSchoolsService::ENQUEUE_THRESHOLD
+          flash[:success] = if selected_school_uuids_count > Publish::Schools::UpdateCourseSchoolsService::ENQUEUE_THRESHOLD
```
tomas-stefano, line 32: `app/controllers/publish/courses/schools_controller.rb:32` vs `app/services/publish/schools/update_course_schools_service.rb:15`

The two threshold checks no longer use the same operand. The controller counts compact_blanked UUIDs (:91); the service counts the raw array, which includes the GOV.UK hidden "" sentinel. At exactly 30 selected schools the service enqueues the job (31 > 30) but the flash renders `success.saved` instead of `success.enqueued_schools` — the user is told the change is applied while it's still queued. Before this PR both sides used the un-compacted array, so this is new.

tomas-stefano (commented):
The switch from `site_ids` to `school_uuids` is the right direction, but the new-model write hinges on an assumption that isn't guaranteed anywhere: during the remodel cycle the form submits **Site** UUIDs, and `UpdateCourseProviderSchoolsService` resolves them against `provider_school.uuid`. `ProviderSchools::Identity` deliberately maps site → provider_school by `gias_school.urn + site_code` precisely because, in its own words, "both models are written to and their UUIDs may diverge". 

So we now have two contradictory mappings for the same relationship, and when they disagree this one degrades to a `logger.warn` nothing alerts on. That's compounded by the all-or-nothing guard: a single unresolvable UUID skips the **entire** `Course::School` sync — attaches that did resolve, and removals too. The backfill only inserts provider schools for sites with `site_type = 0 AND urn IS NOT NULL`, so any provider holding one URN-less or un-backfilled site silently stops syncing to the new model on every subsequent course edit. That's permanent divergence, not the "reconciles later" the comment promises, and it's inconsistent with `Courses::CreationService#update_schools`, which still skips per-school.

The legacy path has its own regression: `sites_by_uuid` now resolves through `course.provider.sites` (scoped `site_type: :school` + `.kept`), while `course.sites` goes through `site_statuses` with no `kept` scope. A discarded site that is still attached to a course therefore fails to resolve, `detach_school` swallows it via `return if site.blank?`, and the `SiteStatus` is never suspended — the school stays on the course forever with no way to remove it, since it isn't rendered in the checkbox list either. The old unscoped `Site.where(id:)` handled this. Two smaller ones worth fixing before merge: the queue is not deploy-safe — jobs already enqueued carry `site_ids`, so post-deploy `params.key?(:school_uuids)` is false, the update is silently dropped, and `site_ids` still reaches `course.assign_attributes`, where `ids_writer` on a `has_many :through` calls `replace` and `delete_all`s the join rows, bypassing the suspend-vs-destroy semantics in `Course#remove_site!`. And the enqueue threshold is now checked against two different operands (controller compacts blanks, service doesn't), so at exactly 30 selected schools the job is enqueued but the flash claims the change was saved.

app/services/publish/schools/update_course_site_statuses_service.rb
```diff
@@ -0,0 +1,122 @@
...
+    # rubocop:disable Style/CommentAnnotation, Lint/RedundantCopDisableDirective
+    # TODO School data remodel removal - remove this legacy SiteStatus sync service when school associations
+    # are written only through Provider::School and Course::School.
+    class UpdateCourseSiteStatusesService
+      include ServicePattern
+
+      def initialize(course:, params:, send_notifications: true)
+        @course = course
+        @params = params.to_h.deep_symbolize_keys
+        raise ArgumentError, "school_uuids must be provided" unless @params.key?(:school_uuids)
```
gms-gs: Probably needs to be in a translation?

app/controllers/publish/courses/schools_controller.rb
```diff
@@ -43,6 +43,10 @@ def update
         else
           render :edit
         end
+      rescue Publish::Schools::UpdateCourseProviderSchoolsService::UnresolvedProviderSchoolsError => e
+        Sentry.capture_exception(e)
```
gms-gs, line 48: Nice

app/services/publish/schools/update_course_site_statuses_service.rb
langesi: Fair point. However, the work we have completed so far ensures that discarded sites are no longer used anywhere in the application. The query service was the final place where we needed to enforce that behaviour.

Given that, I do not think there is a meaningful risk if a discarded school is not removed here, as it should no longer be surfaced or used by the application.

app/services/publish/schools/update_course_provider_schools_service.rb
langesi: Thanks, I fixed that :)

Async updates log any missing schools and continue processing the additions and removals that could be resolved. Inline updates remain strict and transactional, so any unresolved school causes the whole update to fail rather than leaving it partially applied.

app/services/publish/schools/update_course_site_statuses_service.rb
langesi: Notification ownership is now entirely in the orchestrator.

app/services/publish/schools/update_course_provider_schools_service.rb
langesi: Fixed, UpdateCourseProviderSchoolsService now resolves schools through course.provider.schools and builds each Course::School using the resolved Provider::School own gias_school. Therefore both conditions checked by consistent_with_provider_school are guaranteed by construction.

app/services/publish/schools/update_course_schools_service.rb
langesi: Thanks, good spot! I changed that. `course.save! `is now unconditional and the comment accurately explains the Apply `provider.changed_at` requirement.

app/services/publish/schools/update_course_schools_service.rb
langesi: Thanks, I removed the fallback chain. `school_uuids` is resolved once during initialization and passed explicitly to both services

app/views/publish/courses/schools/edit.html.erb
langesi: We will just use the new data model from now on, as we ran the backfill. We do not have to switch anymore

PR #6252, opened 2026-07-24
Bump QA worker resources to run rollover

inulty-dfe:
## Context

  We plan to run rollover in QA often until we rollover production

There are 5 threads per worker so this change will process 10 threads concurrently, 5 per 2Gi

Increase QA worker replicas from 1 to 2
Increase memory per replica from 1.2Gi to 2Gi

## Checklist


langesi (approved):
Thanks @inulty-dfe

PR #6253, opened 2026-07-24
School Data - UI Publish - Provider school details from the new model

gms-gs:
## Context

Publish provider school pages still read from the legacy `Site` model in several places. After the schools remodel backfill, providers in the remodel cycle (2026) and beyond should be served from `Provider::School` instead.

This PR switches Publish school **show**, **delete**, and **index** pages to `ProviderSchools::Identity`, which resolves the correct model by recruitment cycle. One subtlety worth noting: from the remodel year onwards we look up `Provider::School` by UUID (matching backfilled records), while some behaviour — notably course school UUID resolution and removal — still treats 2026 as a transition year where both legacy sites and provider schools may coexist.

- Update Publish schools **show** and **delete** pages to load a decorated `Provider::School` via `ProviderSchools::Identity#school_for` instead of `@site`
- Update Publish schools **index** to list `Provider::School` records from 2026 onwards by aligning `ordered_school_scope` with `uses_provider_schools?` (same predicate as `school_for`)
- Add `uses_provider_schools?` to `Identity` (cycle year >= remodel year) and remove `provider_school_for` site-matching logic
- Refactor `ProviderSchools::Removal` to handle three phases:
  - **Before remodel:** remove legacy sites only
  - **Remodel cycle:** remove both legacy site and matching provider school when present
  - **After remodel:** remove provider school only; raise `RecordNotFound` for invalid UUIDs
- Standardise main site display name to `(Main Site)` on `Provider::School` and delegate `code`/`urn` on the decorator
- Update `NewlyAddedTagComponent` so register-import schools still show the "Newly added" tag when the index renders `Provider::School` rows — it checks `added_via` on the matching legacy site by UUID
- Add request, service, decorator, component, and system specs; update helpers to create backfilled provider schools with matching UUIDs
- Update `CourseSchools::Identity` specs to reflect transitional 2026 behaviour: available school lists use provider schools, course UUID resolution still uses legacy sites
- Pin support delete specs for legacy site-only scenarios to pre-remodel cycles so they stay stable regardless of CI current year

## Guidance to review

- **`ProviderSchools::Identity`** — confirm `uses_provider_schools?` (>= 2026) is used consistently for both `school_for` and `ordered_school_scope`, and that this matches the backfill assumption that provider school UUIDs match legacy site UUIDs during the remodel year
- **`ProviderSchools::Removal#call`** — check the three branching paths (after remodel / provider school present / site-only) and that `removable?` checks the right associations (`course_schools` vs legacy course sites) in each phase
- **Publish show/delete/index views** — verify the switch from `@site` to `school` / `@schools` reads correctly for normal and main site schools, including when the same GIAS school is linked twice with different site codes
- **`NewlyAddedTagComponent`** — confirm looking up register-import status via the legacy site is the right approach during the 2026 rollover, when index rows are provider schools but `added_via` lives on sites
- **`CourseSchools::Identity`** — note this PR does not change course UUID resolution logic; only specs were updated to match the new index behaviour
- **Support console** — support controller/views still use `@site` via `school_removal.school`; only support *specs* were updated for cycle pinning and casing

## Checklist

### DIFF
--- a/app/controllers/publish/providers/schools_controller.rb
+++ b/app/controllers/publish/providers/schools_controller.rb
@@ -44,7 +44,7 @@ module Publish
       helper_method :school_removal
 
       def school
-        @school ||= ProviderSchools::Identity.new(provider:).school_for(uuid: params[:uuid]).decorate
+        @school ||= provider.schools.find_by!(uuid: params[:uuid]).decorate
       end
       helper_method :school
 
--- a/app/controllers/publish/providers/schools_controller.rb
+++ b/app/controllers/publish/providers/schools_controller.rb
@@ -3,7 +3,7 @@
 module Publish
   module Providers
     class SchoolsController < ApplicationController
-      before_action :site, only: %i[show delete destroy]
+      before_action :school, only: %i[show delete destroy]
       before_action :reset_urn_form, only: %i[index]
 
       PER_PAGE = 20
@@ -31,7 +31,7 @@ module Publish
           flash[:success] = "School removed"
           redirect_to publish_provider_recruitment_cycle_schools_path
         else
-          redirect_to delete_publish_provider_recruitment_cycle_school_path(@provider.provider_code, @site.recruitment_cycle.year, @site.uuid),
+          redirect_to delete_publish_provider_recruitment_cycle_school_path(@provider.provider_code, school.recruitment_cycle.year, school.uuid),
                       flash: { warning: t(".cannot_remove_school") }
         end
       end
@@ -43,9 +43,10 @@ module Publish
       end
       helper_method :school_removal
 
-      def site
-        @site ||= school_removal.school
+      def school
+        @school ||= ProviderSchools::Identity.new(provider:).school_for(uuid: params[:uuid]).decorate
       end
+      helper_method :school
 
       def site_params(param_form_key)
         params.expect(param_form_key => SchoolForm::FIELDS)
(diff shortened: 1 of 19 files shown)


gms-gs:
[10.1.3 School Data - UI Publish - Provider school details from the new model](https://trello.com/c/QaRLjsNi/1820-1013-school-data-ui-publish-provider-school-details-from-the-new-model)

app/controllers/publish/providers/schools_controller.rb
```diff
@@ -43,9 +43,10 @@ def school_removal
       end
       helper_method :school_removal
 
-      def site
-        @site ||= school_removal.school
+      def school
+        @school ||= ProviderSchools::Identity.new(provider:).school_for(uuid: params[:uuid]).decorate
```
langesi: provider.schools.find_by!(params[:uuid]).decorate

To simplify

app/services/provider_schools/removal.rb
```diff
@@ -23,30 +27,22 @@ def call
...
-      raise if after_schools_remodel_cycle?
-
-      nil
+      @provider_school ||= provider.schools.find_by(uuid:)
     end
 
     def removable?
       if after_schools_remodel_cycle?
-        !provider_school.course_schools.exists?
+        !school.course_schools.exists?
```
langesi: Remove the if statement and have just

!school.course_schools.exists?

No need to check the site model after the backfill has run

langesi (approved):
LGTM, just have 2 suggestions

PR #6256, opened 2026-07-27
Add course flow uses school UUIDs instead of site_ids

inulty-dfe:
## Context

This PR changes the identifiers used by the add course flow wizard to reference the Placement/Employing schools chosen for a course from `site_id` to `school_uuid`.

`site_id` is an id of a specific model Site and transitioning to UUID makes it easy for us to swap out the underlying model from `Site` to `Provider::School`.

The school appear as a checkbox list and their values are now uuids instead of the `site_id`.

The Add course flow still creates sites and dual writes the course school

During the creation of the Add course Wizard, we retired the old add course flow. We still have lots of tests that exercise those parts. In order for us to fully transition the Add course Wizard to use UUIDs we need to remove all references to `site_ids` in the old flow too. This means deleting old tests.

- The only way to enable the old add course wizard is by changing the url on the Add course button. Remove that option.
- Delete the tests that tested the branch of execution where that button pointed to the old flow.
- Update the CourseCreation service by replacing all `site_id` with `uuid`.
- Update the Wizard to pass uuids through the repositories, state stores, templates.
- Ensure that we cannot save a course with a mismatch in Site and School
- Added a new validator to validate that the uuids that were sent actually created the schools it was supposed to

## Guidance to review

Go through the commits, they're organised fairly chronologically.

## Checklist


app/components/add_course_button.rb
```diff
@@ -37,7 +37,7 @@ def add_course_path
 private
 
   def wizard_add_course_flow?
-    FeatureFlag.active?(:wizard_add_course_flow)
+    true
```
inulty-dfe, line 40: This is so we can safely delete parts of the old course flow to allow the CreationService to drop the `site_ids`

inulty-dfe:
[10.1.3 School Data - Add course flow uses UUIDs instead of site_id](https://trello.com/c/wFig9ZTh/1859-1013-school-data-add-course-flow-uses-uuids-instead-of-siteid)

app/services/courses/creation_service.rb
```diff
@@ -115,67 +106,84 @@ def update_study_mode(course)
...
+
+      if selected_sites.empty?
+        course.errors.add(:schools, message: "Select at least one school")
+        return
+      end
 
-      course.sites = sites if site_ids.any?
+      # TODO School data remodel removal - remove the site writes # rubocop:disable Style/CommentAnnotation
+      # once add-course creation only writes Course::School.
+      course.sites = selected_sites
```
tomas-stefano, line 121: `update_schools` returns early on an unrecognised uuid, before course.sites = selected_sites (line 127). 

The error it adds is then wiped by CreateCourse's `course.save`, because save re-runs validations in the default context.

Submit two schools where one uuid has gone stale (school deleted between the schools step and check-answers) and you get a course with no schools and no sites at all, and the wizard redirects to the courses index reporting success. Commit acc6f339a ("Report unrecognised school UUIDs instead of dropping them silently") achieves the opposite of its title in the real flow. It's also worse than before: `provider.sites.find(site_ids)` used to raise RecordNotFound rather than silently proceed.

`CreateCourse#execute` needs to bail when `course.errors.any?` before calling save (or CreationService should raise).

app/services/courses/creation_service.rb
```diff
@@ -115,67 +106,84 @@ def update_study_mode(course)
...
-          # later. Same rationale as UpdateCourseSchoolsService#attach_school.
-          Rails.logger.warn(
-            "[CourseSchools] skipped course_school build — no provider_school for " \
-            "provider=#{provider.id} gias_school=#{gias_school.id}",
-          )
-          next
-        end
-
-        course.schools.build(gias_school_id: gias_school.id, provider_school:)
+        course.schools.build(gias_school: provider_school.gias_school, provider_school:)
```
tomas-stefano, line 127: The old code filtered through `GiasSchool.available` with an explicit comment that a closed school must never enter the new model; the spec asserting it was deleted. Should we continue using only the available ones?

spec/system/publish/courses/add_course_wizard/level_spec.rb
```diff
@@ -43,6 +49,16 @@ def given_i_am_authenticated_as_a_provider_user_with_a_school
     given_i_am_authenticated(user: @user)
   end
 
+  def given_i_am_authenticated_as_a_provider_user_with_a_provider_school
+    recruitment_cycle = find_or_create(:recruitment_cycle, year: Settings.schools_remodel_cycle_year + 1)
+    provider = create(:provider, :accredited_provider, recruitment_cycle:, sites: [build(:site)])
```
tomas-stefano: Maybe I understood wrong but I think the scenario doesn't test what it says.  It seeds a legacy site, so it passes regardless of whether Provider::School support works

app/wizards/course_wizard/draft.rb
```diff
@@ -174,5 +174,23 @@ def ordered_site_records(ids)
       records_by_id = Site.where(id: ids).index_by { |site| site.id.to_s }
       ids.filter_map { |id| records_by_id[id.to_s] }
     end
+
+    # Schools are stored as legacy Site UUIDs. One that no longer resolves is
+    # logged and left out, rather than blanking the whole review row.
+    def ordered_school_records(uuids)
```
tomas-stefano, line 180: `draft.rb:92` reads `state_store.school_uuids`; sessions mid-flow hold `site_ids`. On deploy, anyone past the schools step loses their selection with no warning, and school_uuids: [] sends them down the "Select at least one school" path. 

Worth a one-release read-both shim, or at least a note in the deploy plan.

tomas-stefano:
In terms of object code design.

`app/wizards/course_wizard/draft.rb`:183 vs `steps/schools.rb`:57 vs `creation_service.rb:154` — three
  separate hand-rolled "resolve uuids, warn, drop/error" implementations with three different behaviours (Draft drops silently, the step errors, the service errors). Maybe we could have one resolver. Wdyt?

app/wizards/course_wizard/draft.rb
```diff
@@ -174,5 +174,23 @@ def ordered_site_records(ids)
       records_by_id = Site.where(id: ids).index_by { |site| site.id.to_s }
       ids.filter_map { |id| records_by_id[id.to_s] }
     end
+
```
tomas-stefano, line 177: `creation_service.rb:81` dedupes with .uniq; the step and Draft don't — a duplicated uuid renders  the school twice on the check-answers row. Minor.

app/services/courses/creation_service.rb
```diff
@@ -115,67 +106,84 @@ def update_study_mode(course)
...
+      if unrecognised_school_uuids.any?
+        course.errors.add(
+          :schools,
+          message: I18n.t("course_schools.errors.unrecognised_school_uuids", support_email: Settings.support_email),
+        )
+        return
+      end
+
+      if selected_sites.empty?
+        course.errors.add(:schools, message: "Select at least one school")
```
tomas-stefano: Can we move this to a locale or add an error type for the school and move to a locale?

app/services/courses/creation_service.rb
inulty-dfe: I've added an extra `course.valid?(:new)` before the `#save` so the errors added in the creation service are recognised

app/services/courses/creation_service.rb
inulty-dfe: For now we are going to allow closed schools to be added to courses. managing this at the same time as deploying the school migration is not worth the hassle. We will remove all closed schools at rollover and devise a strategy for managing closed schools throughout the cycle later

app/wizards/course_wizard/draft.rb
inulty-dfe: I don't see how a uuid can be duplicated so I'm not worried about this. I'm removing the `uniq` from the `creation_service.rb` so it's consistent

app/wizards/course_wizard/draft.rb
inulty-dfe: We'll be sure to release after hours. The validations in the course_create will call `course.valid?(:new) and cause an error so that the user must add missing schools.

spec/system/publish/courses/add_course_wizard/level_spec.rb
inulty-dfe: Yes, this test was poorly written. I've updated it to test that the user cannot start the course wizard if they have no sites on their provider

app/services/courses/creation_service.rb
```diff
@@ -86,15 +78,15 @@ def subject_ids
...
+    def school_uuids
+      @school_uuids ||= Array(course_params["school_uuids"]).compact_blank
     end
 
     def study_mode
       @study_mode ||= if course_params["study_mode"].nil?
                         nil
                       else
-                        Array(course_params["study_mode"])&.flatten&.compact&.uniq
+                        Array(course_params["study_mode"])&.flatten&.compact
```
tomas-stefano: Is the `.uniq` was there because someone was expecting a duplicated study mode?

If so,  this will cause a 500.

  ArgumentError: 'full_time_or_full_time' is not a valid study_mode.

 Unhandled exception on study_mode: ["full_time", "full_time"]. Main returns "full_time" and saves. This has nothing to do with the UUID work — it looks like collateral from the rewrite.

app/services/courses/creation_service.rb
```diff
@@ -52,6 +47,7 @@ def build_new_course
       clean_up_visa_properties(course)
       Courses::AssignVisaSponsorshipApplicationDeadlineService.execute(course_params, course)
       course.valid?(:new) if course.errors.blank?
+      course.send(:set_applications_open_from)
```
tomas-stefano, line 51: Writing here to see if I understood this line.

`course.send(:set_applications_open_from)` is a  workaround for the new stricter save, and it's in the wrong place:

`applications_open_from` is normally filled in by a callback, not by the form:
```
  - course.rb:110 — before_save :set_applications_open_from → self.applications_open_from ||=
  recruitment_cycle.application_start_date
  - course.rb:394 — validate :validate_applications_open_from, on: %i[publish update new] → errors
  if it's blank
```

The wizard never asks for it, so at :new-validation time it is blank and that validation fails.

  Before this PR that didn't matter: CreateCourse called course.save, which validates in the default context (no applications_open_from check) and then runs before_save, which fills the value in.

The error existed but nothing acted on it.

This PR changed the operation to `course.valid?(:new) && course.save`. That previously-ignorable error becomes fatal — every wizard course would fail to save. Line 50 is the fix: manually fire the private before_save callback so the value is set. send is needed only because the method is private.

## Why I'd change it

  The ordering leaves a false error on the returned course. Line 49 validates, line 50 sets the value — in that order. So CreationService hands back a course that reports a problem it no longer has:

  The service's contract is "here's the course and why it can't be created", and that answer is now wrong. It only works because CreateCourse re-validates and clears it. Swapping lines 49 and 50 fixes this outright. (To be fair: main returns the same error — what's new is that correctness now depends on someone re-validating.)

It's load-bearing and the next failure won't look like this one. The service now has to know
 about, and hand-replay, one specific before_save callback for :new validation to pass. Add another callback-derived attribute that a :new-context validation checks and course creation breaks again — but the error will name that attribute ("Enter a course length", say), with nothing pointing at a missing callback. Whoever debugs it won't find this line.

  Suggested fixes

  - Minimum: move line 50 above line 49 and add a comment saying why it's there.
  - Alternative: drop the operation's second valid?(:new) and have CreationService be the single
  place that decides validity — it already calls valid?(:new) at line 49 and could return that verdict directly.

tomas-stefano (approved):

app/wizards/course_wizard/steps/schools.rb
```diff
@@ -7,57 +7,76 @@ class Schools
...
           value: ->(draft) { draft.schools.map(&:location_name) },
+          show_when_blank: true,
           format: Publish::CheckAnswers::Formatters::List.new(separator: :br),
         )
       end
 
-      def sites
-        @sites ||= provider_sites.sort_by(&:location_name)
+      def schools
+        @schools ||= provider.sites.order(:location_name)
```
langesi, line 26: We still depending on sites? My suggestion is just to use the new data model as we are successfully dual writing.

app/validators/course_publishable_schools_presence_validator.rb
```diff
@@ -8,7 +8,18 @@ class CoursePublishableSchoolsPresenceValidator < ActiveModel::Validator
   def validate(course)
     return if Courses::PublishRules::SchoolPresence.any?(course)
     return if Courses::PublishRules::SchoolPresenceExemption.applies?(course)
+    return if selection_made_but_unresolved?(course)
 
     course.errors.add(:sites, :blank)
```
langesi, line 13: I think we should assign the error to the new data model

    course.errors.add(:school, :blank)

app/services/schools/uuid_resolver.rb
```diff
@@ -0,0 +1,63 @@
...
+  # order they were submitted in and reporting back any UUID that does not
+  # belong to the provider.
+  #
+  # Every layer of the add-course flow needs the same lookup but reacts to an
+  # unrecognised UUID differently - the wizard step raises a form error the
+  # provider can fix, the draft degrades so the review row still renders, and
+  # course creation errors the course. Only the lookup is shared here; the
+  # reaction stays with the caller.
+  #
+  # TODO School data remodel - the scope below becomes provider.schools once the
```
langesi, line 14: We can read already from the new model, as we are dual writing and both data models are in sync

langesi (commented):
Great work, easy to read,I feel like still leans too heavy on the old Site model especially after we have now both data models in sync.

The wizard still reads and resolves schools from the legacy Site model, I don't think we need to do this anymore. It means we'll have to revisit this part of the application later again when we remove Sites all together. I think the wizard collection should now be provider.schools, with the submitted UUID resolving directly to Provider::School. I found that
	- The checkbox list is still populated from provider.sites.
	- Wizard validation considers a UUID valid only if there is a matching legacy site.
	- The check-answers page resolves selections through legacy sites.
	- Course creation starts from a legacy site and looks up its corresponding Provider::School.
	- A provider with valid Provider::School records but no legacy sites cannot create a course. 

We can either discuss this or if you change it I'd be happy to approve it.

app/wizards/course_wizard/steps/schools.rb
inulty-dfe: This PR is only to change the params that are used to identify the sites, displaying schools in the UI is another concern

langesi (approved):
All good from my end, we discussed to work on my comment in a different PR

PR #6259, opened 2026-07-27
Publish course list hide filters

tomas-stefano:
# Hide filter groups that don't vary across a provider's courses

A filter is useless when every course shares the same value — e.g. a provider whose courses are all full time gains nothing from "Full time or part time". This hides a filter group when its value is identical across **all** of the
provider's courses, so the sidebar only ever shows filters that do something.

- Each filter group (Status, Education phase, Fee or salary, Qualification, Full time or part time, Start date) shows only if its value **varies** across the provider's whole course list.
- If **nothing** varies, the whole sidebar is dropped and the course list shifts left (keeping its width, not going full-width).
- A group with an **applied** filter always stays visible, so an active filter is never hidden from its own panel.
- Education phase comes from `course.level` on the read-model rows ("use the new  data model"). No feature flag. **No JavaScript** — the "Apply filters" button isunchanged.

## How

Same "hide uniform values" rule already used for the course-information column, computed once over the **unfiltered** course set (so filtering never adds/removes a filter) and reusing the rows `CourseList` already loads (no extra query).

```mermaid
flowchart TD
    A["provider's unfiltered courses<br/>(rows already loaded by CourseList)"] --> B{"for each of the 6 facets:<br/>more than one distinct value?"}
    B -->|"yes → varies"| C["group shown"]
    B -->|"no → uniform"| D["group hidden"]
    E["group has an applied filter?"] -->|yes| C
    C --> F["visible_filter_groups"]
    D --> F
    F --> G{"any groups visible?"}
    G -->|yes| H["two columns:<br/>sidebar ⅓ + list ⅔"]
    G -->|no| I["no sidebar:<br/>list ⅔, shifted left"]
```

Facet keys per course:

| Group | Facet compared on |
|---|---|
| Status | `StatusTag.token` (open/closed/draft/rolled_over/scheduled/withdrawn) |
| Education phase | `course.level` |
| Fee or salary | `course.funding` |
| Qualification | `course.qualification` |
| Full time or part time | `course.study_mode` |
| Start date | start month (`beginning_of_month`) — the filter groups by month |

### The status token

Status needs the same value the filter selects on and the tag renders — not the
label, since "Open" and "Open \*" (unpublished changes) are the same status. New
`Publish::Courses::StatusTag.token` is that single definition, sitting between
the two existing sources:

```mermaid
flowchart LR
    T["StatusTag.token"] --- Q["Query#status_predicate<br/>(what the filter selects)"]
    T --- C["StatusTagComponent<br/>(what the tag renders)"]
    D["drift-guard spec"] -. "asserts all three agree" .-> T
```

`content_status` → `draft`/`rolled_over`/`withdrawn` map straight through;
`published`(+unpublished changes) → `open`/`closed` in the current/previous cycle
(by application status) or `scheduled` in a future cycle.

## Guide for reviewing and testing

See trello card comments for the scenarios and providers to test.

### DIFF
--- a/app/services/publish/course_list.rb
+++ b/app/services/publish/course_list.rb
@@ -31,6 +31,17 @@ module Publish
       start_date: ->(course) { course.start_date.presence&.to_date },
     }.freeze
 
+    # The value each filter group varies on, in panel order. A group whose value
+    # is identical across the whole list offers no useful choice, so it is hidden.
+    # Start date is compared by month because the filter groups by month.
+    FILTER_FACETS = {
+      level: ->(course) { course.level },
+      funding: FIELDS[:funding],
+      qualification: FIELDS[:qualification],
+      study_mode: FIELDS[:study_mode],
+      start_date: ->(course) { course.start_date.presence&.to_date&.beginning_of_month },
+    }.freeze
+
     delegate :any?, to: :groups
 
     def initialize(provider:, params: {})
@@ -46,6 +57,14 @@ module Publish
     def visible_course_information_fields
       @visible_course_information_fields ||=
         FIELDS.keys.select { |key| unfiltered_courses.map(&FIELDS.fetch(key)).uniq.size > 1 }
+    end
+
+    # The filter groups worth showing: those whose value varies across the whole
+    # (unfiltered) list. Measured on the unfiltered set so filtering never adds
+    # or removes a filter.
+    def visible_filter_groups
+      @visible_filter_groups ||=
+        FILTER_FACETS.keys.select { |group| unfiltered_courses.map(&FILTER_FACETS.fetch(group)).uniq.size > 1 }
     end
 
     def groups
--- a/app/services/publish/course_list.rb
+++ b/app/services/publish/course_list.rb
@@ -33,8 +33,10 @@ module Publish
 
     # The value each filter group varies on, in panel order. A group whose value
     # is identical across the whole list offers no useful choice, so it is hidden.
-    # Start date is compared by month because the filter groups by month.
+    # Start date is compared by month because the filter groups by month; status
+    # by the token the filter selects on rather than the rendered label.
     FILTER_FACETS = {
+      status: ->(course) { Publish::Courses::StatusTag.token(course) },
       level: ->(course) { course.level },
       funding: FIELDS[:funding],
       qualification: FIELDS[:qualification],
@@ -60,11 +62,13 @@ module Publish
     end
 
     # The filter groups worth showing: those whose value varies across the whole
-    # (unfiltered) list. Measured on the unfiltered set so filtering never adds
-    # or removes a filter.
+    # (unfiltered) list, plus any group with a filter already applied so an active
+    # filter is never hidden from its panel. Measured on the unfiltered set so
+    # filtering never removes a filter mid-use.
     def visible_filter_groups
-      @visible_filter_groups ||=
-        FILTER_FACETS.keys.select { |group| unfiltered_courses.map(&FILTER_FACETS.fetch(group)).uniq.size > 1 }
+      @visible_filter_groups ||= FILTER_FACETS.keys.select do |group|
+        params.key?(group) || unfiltered_courses.map(&FILTER_FACETS.fetch(group)).uniq.size > 1
+      end
     end
 
     def groups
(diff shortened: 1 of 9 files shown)


tomas-stefano:
https://trello.com/c/69eknLzg/1756-course-list-add-logic-for-when-to-display-filters

kelliedesigner:
Looks good to me!

app/services/publish/courses/status_tag.rb
```diff
@@ -0,0 +1,31 @@
...
+    # This is the value the status filter selects on and the status tag renders,
+    # so it is the right key for deciding whether the Status filter varies across
+    # a provider's courses. Mirrors StatusTagComponent's tag choice and inverts
+    # Query#status_predicate; a drift-guard spec ties the three together.
+    #
+    # "Open" and "Open *" (unpublished changes) share the token :open — the token
+    # is the filterable status, not the rendered label.
+    module StatusTag
+      def self.token(course)
+        case course.read_attribute(:content_status).to_s
```
langesi: What do you think of:

```
COURSE_STATUS = { draft: :draft, rolled_over: :rolled_over, withdrawn: :withdrawn }.freeze

def self.token(course)
	COURSE_STATUS.fetch(course.read_attribute(:content_status).to_s, other_states)
end

def other_states(course)
	if Find::CycleTimetable.current_or_previous_year?(course.recruitment_cycle.year)
        course.application_status_open? ? :open : :closed
      else
        :scheduled
      end
end
```

langesi (commented):
Haven't had a chance to fully review, just submitting a comment

app/services/publish/courses/status_tag.rb
tomas-stefano: Good instinct on splitting these up — token was doing two unrelated things and the second one
  deserved a name. I've taken that half in eed5c4040.

  I went with a list rather than the hash though:
```
  SELF_EVIDENT_TOKENS = %w[draft rolled_over withdrawn].freeze

  def self.token(course)
    content_status = course.read_attribute(:content_status).to_s
    return content_status.to_sym if SELF_EVIDENT_TOKENS.include?(content_status)

    published_token(course)
  end

  # published / published_with_unpublished_changes
  def self.published_token(course)
    return :scheduled unless
  Find::CycleTimetable.current_or_previous_year?(course.recruitment_cycle.year)

    course.application_status_open? ? :open : :closed
  end
```

  { draft: :draft, rolled_over: :rolled_over, withdrawn: :withdrawn } is an identity map, so it adds
  a lookup table without removing any duplication — and it has to be kept in step with its own
  keys.

app/services/publish/course_list.rb
```diff
@@ -48,6 +61,16 @@ def visible_course_information_fields
...
+    # filtering never removes a filter mid-use.
+    def visible_filter_groups
+      @visible_filter_groups ||= FILTER_FACETS.keys.select do |group|
+        params.key?(group) || unfiltered_courses.map(&FILTER_FACETS.fetch(group)).uniq.size > 1
+      end
+    end
+
     def groups
       @groups ||= courses
         .chunk_while { |a, b| a[:group_name] == b[:group_name] }
```
inulty-dfe, line 76: Never seen this used before. Just thought I'd mention it

spec/requests/publish/courses_spec.rb
```diff
@@ -35,16 +35,58 @@ def course_names
...
 
-      expect(response.parsed_body.css(".app-active-filters")).to be_empty
+      expect(filter_headings).to eq(
+        ["Status", "Education phase", "Fee or salary", "Qualification", "Full time or part time", "Start date"],
+      )
+    end
+
+    it "hides the groups the courses do not vary on" do
+      # Same status, funding, qualification, study mode and start date; only phase differs.
+      create(:course, :without_validation, :primary, provider:, funding: "fee", qualification: :qts, study_mode: :full_time, start_date: Time.zone.local(2026, 9, 1))
```
inulty-dfe, line 65: Having an explicit year here is a bit distracting. I don't think it will cause the test to fail, but it being explicit makes me wonder if this test depends on the year.

spec/services/publish/course_list_spec.rb
```diff
@@ -145,6 +145,99 @@
...
+
+      it "shows those filters in panel order" do
+        expect(course_list.visible_filter_groups).to eq(%i[funding qualification study_mode])
+      end
+    end
+
+    context "when courses start in different months" do
+      before do
+        create_course(start_date: Time.zone.local(2026, 9, 1))
+        create_course(start_date: Time.zone.local(2027, 1, 1))
```
inulty-dfe, line 191: lets not hard code dates unless it's specific to a change between those cycles

inulty-dfe (approved):
LGTM

PR #6261, opened 2026-07-29
Prevent closed GIAS schools from being added to provide or course

inulty-dfe:
#### Creating provider schools / sites
- When adding schools to a provider (via Site or ProviderSchool), we only want GIAS open schools to be available for adding.
- Provider::School currently handles this but Site does not prevent closed GIAS schools from being created on the provider.
- We need to stop closed GIAS school sites from being created to keep the dual write consistent

#### Adding provider schools to courses
- When adding schools to courses, the user is presented with the list of provider schools. We don't filter out closed GIAS schools from this list so providers can add closed gias schools when creating/updating a course.
- If a course already has a closed school attached, and the user goes to edit the attached schools, we still load the closed school, checked, in the list of schools. This is so the school isn't automatically removed.

#### Rollover of schools
When we run rollover, we copy Site/SiteStatus and Provider::School/Course::School that correspond to closed GIAS schools.

- Prevent closed GIAS schools from being added to a provider.
- If a closed GIAS school already exists on the provider, prevent the user from choosing them when adding / updating a course shools.
- Update the Course and Provider rollover schools copiers to ignore closed GIAS schools when copying fro mone recruitment cycle to another.

## Checklist

### DIFF
--- a/app/models/gias_school.rb
+++ b/app/models/gias_school.rb
@@ -26,6 +26,8 @@ class GiasSchool < ApplicationRecord
     ])
   }
 
+  scope :unavailable, -> { where.not(id: available) }
+
   enum :status_code, {
     open: "1",
     closed: "2",
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -60,11 +60,16 @@ class Site < ApplicationRecord
 
   scope :not_geocoded, -> { where(latitude: nil, longitude: nil) }
 
-  # Filter out sites we can identify as closed: those whose URN matches an
-  # unavailable GIAS school. Sites with no URN, or a URN matching no GIAS
-  # school, are kept.
+  # Sites backed by an available GIAS school. Effectively a join on urn (Site
+  # has no foreign key to gias_school) without selecting the GIAS columns, so a
+  # urn matching no GIAS row is excluded rather than passed through - from 2026
+  # every school site's urn matches a GIAS record, and an earlier cycle's
+  # unmatched urn is uncleansed data, not a school we can vouch for.
+  #
+  # Main sites are the one exemption: they are legitimately urn-less
+  # (see the urn presence validation above), so they are always kept.
   scope :with_available_gias_school, lambda {
-    where.not(urn: GiasSchool.unavailable.select(:urn)).or(where(urn: nil))
+    where(urn: nil).or(where(urn: GiasSchool.available.select(:urn)))
   }
 
   attr_accessor :skip_geocoding
--- a/app/models/site.rb
+++ b/app/models/site.rb
@@ -60,6 +60,13 @@ class Site < ApplicationRecord
 
   scope :not_geocoded, -> { where(latitude: nil, longitude: nil) }
 
+  # Filter out sites we can identify as closed: those whose URN matches an
+  # unavailable GIAS school. Sites with no URN, or a URN matching no GIAS
+  # school, are kept.
+  scope :with_available_gias_school, lambda {
+    where.not(urn: GiasSchool.unavailable.select(:urn)).or(where(urn: nil))
+  }
+
   attr_accessor :skip_geocoding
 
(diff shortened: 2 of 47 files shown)


inulty-dfe:
[School remodel - Avoid closed GIAS schools](https://trello.com/c/YtuGbS2f/1849-school-remodel-avoid-closed-gias-schools)

langesi:
Just did a QA run through the app, haven't read the code yet.

This works well, what I observed:
- Can't add a closed school to a provider, as it's not shown
- New tag is shown for closed schools on the provider school page
- Can't add a closed school in the add course wizard, as it's not shown
- Can add a closed school in the basic details tab when I add schools
	- This will save it to both, the new and old data model used University of Buckingham provider and The Cedars School for testing, as the school was backfilled and is closed
- Our backfill script has backfilled also closed schools attached to a course and provider

app/models/site.rb
```diff
@@ -60,6 +60,13 @@ class Site < ApplicationRecord
 
   scope :not_geocoded, -> { where(latitude: nil, longitude: nil) }
 
+  # Filter out sites we can identify as closed: those whose URN matches an
+  # unavailable GIAS school. Sites with no URN, or a URN matching no GIAS
+  # school, are kept.
+  scope :with_available_gias_school, lambda {
+    where.not(urn: GiasSchool.unavailable.select(:urn)).or(where(urn: nil))
```
langesi: I tink this might still show invalid URNs? Because you exclude unavailable gias schools and NULL URN, so if you have an invalid URN it would be shown

langesi: Might be safer to do

where(urn: GiasSchool.available.select(:urn)).or(where(urn: nil))

app/models/site.rb
```diff
@@ -60,6 +60,13 @@ class Site < ApplicationRecord
 
   scope :not_geocoded, -> { where(latitude: nil, longitude: nil) }
 
+  # Filter out sites we can identify as closed: those whose URN matches an
+  # unavailable GIAS school. Sites with no URN, or a URN matching no GIAS
+  # school, are kept.
+  scope :with_available_gias_school, lambda {
```
langesi, line 71: Name suggestion:

with_open_gias_school ?

app/models/gias_school.rb
```diff
@@ -26,6 +26,8 @@ class GiasSchool < ApplicationRecord
     ])
   }
 
+  scope :unavailable, -> { where.not(id: available) }
```
langesi, line 29: Name suggestion, as we use open and closed already to define the status. should we continue to use that terminology, so this scope would be `closed` or `closed_schools`

langesi (commented):
Good work, just a few small comments, also check the below out as I think you wanted to remove closed schools from being selected when adding to a course:

- Can add a closed school in the basic details tab when I add schools
	- This will save it to both, the new and old data model used University of Buckingham provider and The Cedars School for testing, as the school was backfilled and is closed

app/models/site.rb
inulty-dfe: open is a status on the model. `open`, `closed`, `proposed_to_open`, `proposed_to_close`. So i decided not to confuse the meta status of "not closed" with "open"

app/models/gias_school.rb
inulty-dfe: The point is to avoid using the names of individual statuses when describing a super type of statuses. We might add `proposed_to_close` to this scope and the name would not match the underlying statuses

app/models/site.rb
inulty-dfe: Ok, so we certainly want this scope to exclude sites that don't have a matching URN in the GIAS table :+1:

inulty-dfe:
Closing this as there is too much work to complete at this time. A new PR https://github.com/DFE-Digital/publish-teacher-training/pull/6264 covers just the rollover scenario for this cycle.

At a later time we can implement the rest of this feature

PR #6265, opened 2026-07-31
Dual write schools indefinitely

inulty-dfe:
## Context

We have now planned to continue dual writing Site and Provider::School next cycle.

Continue to dual write both Site and Provider::School models regardless of RecruitmentCycle.
I can't find anyother places that need correcting.

## Checklist


langesi:
Checked the code, looks good, deploying the PR to do some QA checks, will approve after

langesi (approved):
Great work, tested it works in this cycle in the review app and locally in the next

PR #6266, opened 2026-07-31
Fix unique course validator

inulty-dfe:
## Context

  Without returning early, the code will trigger exceptions calling
  methods on the course and allegeed associations.
  If the new course doesn't have a provider assigned,
  new_course.provider returns nil

Do not let the unique course validator raise NoMethodError when validating

## Checklist


inulty-dfe:
Well this fix has broken some tests.

image

langesi (approved):

PR #6267, opened 2026-08-03
Improve copy course content

sarahcrack:
## Context

Trello: https://trello.com/c/QXPEnJTC/1839-improve-copy-course-content

When selecting a course to copy from, the select options don’t have enough course information to identify the correct course.

**Dropdown option content**
Each course option now displays:

- Course name and course code (always)
- Age range for primary courses (always)
- Funding type (only when it differs across available courses)
- Qualification (only when it differs across available courses)
- Study mode (only when it differs across available courses)
- Start date (only when it differs across available courses)

The logic for determining whether a field should be shown reuses the existing **Publish::CourseList::FIELDS** definitions.

Only the courses available in the copy-content dropdown are considered when determining whether a field varies. The course currently being edited is excluded from this comparison, as it is not present in the dropdown.

**Drop down with course information**

Screenshot 2026-08-04 at 13 17 08

**Showing age range on primary courses**

Screenshot 2026-08-04 at 13 18 14

## Guidance to review

- Login as Colin on Publish and select a provider
- Choose on a course and click on 'Change' under 'Where you will train'
- Check the copy course content drop down displays any course information that differs and you can copy course content successully from another course

## Checklist

### DIFF
--- a/app/views/publish/courses/_course_select_option.html.erb
+++ b/app/views/publish/courses/_course_select_option.html.erb
@@ -1,1 +1,14 @@
-<option value="<%= other_course.course_code %>"><%= other_course.name %> (<%= other_course.course_code %>)</option>
+<% age_range =
+     other_course.primary_course? ?
+     "Ages #{other_course.age_range}" :
+     nil %>
+<option value="<%= other_course.course_code %>">
+  <%= [
+    other_course.name_and_code,
+    age_range,
+    I18n.t("publish.courses.course_table.funding.#{other_course.funding}"),
+    other_course.qualifications_summary,
+    other_course.study_mode_description.capitalize,
+    I18n.l(other_course.start_date.to_date, format: :short)
+  ].compact.join(", ") %>
+</option>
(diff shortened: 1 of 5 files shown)


app/views/publish/courses/_course_select_option.html.erb
sarahcrack, line 1: Do I need to move any of this code into something like a helper?

kelliedesigner:
Design review - looks good to me!

app/views/publish/courses/_course_select_option.html.erb
```diff
@@ -1 +1,13 @@
-<option value="<%= other_course.course_code %>"><%= other_course.name %> (<%= other_course.course_code %>)</option>
+<%# Course name/code and primary age range are always shown.
+    Other course information is only shown when it differs across
+    the courses available for copying. %>
+<option value="<%= other_course.course_code %>">
+  <%= [
```
langesi, line 5: This array is harder to read, however it shouldn't be a blocker.

It's a small partial, so that lessens the cognitive load

app/views/publish/courses/_course_select_option.html.erb
```diff
@@ -1 +1,13 @@
...
+    other_course.name_and_code,
+    (other_course.primary_course? ? "Ages #{other_course.age_range}" : nil),
```
langesi: The String would be good in a locale

langesi (approved):
Brain fog, removed my first comment. Great work, just add the locales I mentioned

PR #6269, opened 2026-08-03
School data - Update Schools in Basic details component

langesi:
## Context

This is the course school update part of the school data remodel. It only takes care of the school adding in the basic details component of a course in publish.

The legacy `Site` and corresponding `Provider::School` are expected to share the same UUID. This allows one submitted `Provider::School` selection to be applied consistently to both models.

Inline and queued updates are both transactional. Their handling of a missing `Provider::School` differs deliberately:

- Inline requests fail and show a form error because the submitted selection is no longer valid.
- Queued requests log and exclude `Provider::School` records removed while the job was waiting.
- The resulting effective selection is then written atomically to both models.
- A valid `Provider::School` without its expected legacy `Site` is treated as data drift and raises rather than silently leaving the models inconsistent.

- Updates the existing course schools form to display available `Provider::School` records and submit `school_uuids` instead of legacy Site IDs.
- Reads the course's current selection and attached-school count from `Course::School`.
- Uses the same `GiasSchool.available` scope for form display and validation, excluding closed schools.
- Validates that submitted UUIDs belong to an available school for the course provider.
- Reports unresolved inline selections to Sentry and displays the existing form error.
- Refactors `Publish::Schools::UpdateCourseSchoolsService` to:
  - resolve submitted `Provider::School` UUIDs once;
  - lock resolved records while the update is applied;
  - pass the same effective selection to both writers;
  - keep inline and queued dual writes within one transaction;
  - persist attributes such as `schools_validated`;
  - save the course so `provider.changed_at` is updated for downstream Apply synchronisation;
  - send school-update notifications from the resulting `Course::School` selection.
- Adds `Publish::Schools::UpdateCourseProviderSchoolsService` to synchronise `Course::School` records from already-resolved `Provider::School` records.
- Adds `Publish::Schools::UpdateCourseSiteStatusesService` to temporarily mirror the same UUID selection to legacy `SiteStatus` records.
- Makes missing legacy Site mappings explicit instead of silently skipping them.
- Updates the queued job to tolerate stale `Provider::School` UUIDs without disabling transaction protection for other failures.
- Counts unique, non-blank UUIDs consistently when deciding whether to enqueue and which success message to show.
- Adds a Site factory trait that creates a corresponding GIAS school and `Provider::School` with matching school data and UUID.
- Adds coverage for:
  - adding, removing and replacing schools in both models;
  - removing every school where this is permitted;
  - stale queued selections;
  - transaction rollback after writer or course-save failures;
  - missing legacy Site mappings;
  - inline form errors and Sentry reporting;
  - notifications;
  - provider `changed_at` updates;
  - dual writing after the remodel cycle;
  - closed schools;
  - queue boundaries, duplicate UUIDs and blank values;
  - select-all and collapsed school lists.

There are no intended visual changes beyond the form being backed by the remodelled school data.

## Guidance to review

Suggested review order:

Option 1: Check the history and review it that way

Option 2: Check by topic as stated below:

### 1. Existing course edit form and validation

- `app/forms/publish/course_school_form.rb`
- `app/controllers/publish/courses/schools_controller.rb`
- `app/views/publish/courses/schools/edit.html.erb`
- `config/locales/en.yml`
- `app/decorators/course_decorator.rb`
- `spec/forms/publish/course_school_form_spec.rb`
- `spec/decorators/course_decorator_spec.rb`
- `spec/support/page_objects/publish/course_school_edit.rb`
- `spec/system/publish/courses/editing_course_schools_spec.rb`
- `spec/system/publish/courses/schools/school_list_display_spec.rb`

Check that the form displays available `Provider::School` records, submits UUIDs, restores its selection from `Course::School`, rejects unavailable or unrelated schools, and renders a normal form error when an inline update can no longer resolve the selection.

### 2. Update orchestration and queued behaviour

- `app/services/publish/schools/update_course_schools_service.rb`
- `app/jobs/update_course_schools_job.rb`
- `spec/services/publish/schools/update_course_schools_service_spec.rb`
- `spec/jobs/update_course_schools_job_spec.rb`

Check that submitted UUIDs are resolved once and that both writers receive the same effective selection.

Pay particular attention to the queued path: stale `Provider::School` UUIDs are logged and excluded, but the remaining dual write is still transactional. Unexpected writer, validation or course-save failures therefore roll back both models.

Also check the comment around `course.save!`. The save is required to update `provider.changed_at`, which Apply uses to decide whether the provider's courses need synchronising.

### 3. New Course::School synchronisation

- `app/services/publish/schools/update_course_provider_schools_service.rb`
- `spec/services/publish/schools/update_course_provider_schools_service_spec.rb`

Check that this service has one responsibility: synchronising `Course::School` rows from an already-resolved collection of `Provider::School` records.

UUID resolution and missing-school policy deliberately remain in the orchestrator.

### 4. Legacy SiteStatus synchronisation

- `app/services/publish/schools/update_course_site_statuses_service.rb`
- `spec/services/publish/schools/update_course_site_statuses_service_spec.rb`

Check that this temporary writer applies the same effective UUID selection to legacy `SiteStatus` records.

The important invariant is that matching `Site` and `Provider::School` records share a UUID. Missing Site mappings now raise before mutation rather than being silently skipped.

Also check that published status is applied only to active SiteStatus rows and that suspended or discontinued rows are not reactivated.

### 5. Queue boundaries and user journeys

- `spec/system/publish/courses/schools/select_all_schools_spec.rb`
- `spec/system/publish/courses/publishing_a_course_without_schools_spec.rb`
- `spec/system/publish/providers/courses/schools_rollover_validation_spec.rb`
- `spec/system/publish/providers/newly_added_tag_add_schools_to_a_course_page_spec.rb`

Check select-all and collapsed-list behaviour, the asynchronous path above 30 unique schools, rollover validation, newly added labels, and providers that are allowed to remove every school.

### 6. Supporting test and course-creation changes

- `spec/factories/sites.rb`
- `app/services/courses/creation_service.rb`
- `spec/services/courses/creation_service_spec.rb`

Check that the Site factory trait creates matching legacy and remodelled school records with the same UUID, URN and location information. The course-creation changes in this PR are limited to keeping the existing creation behavior and comments consistent with the completed backfill.

### DIFF
--- a/app/controllers/publish/courses/schools_controller.rb
+++ b/app/controllers/publish/courses/schools_controller.rb
@@ -27,9 +27,12 @@ module Publish
         @course_school_form = Publish::CourseSchoolForm.new(@course, params: school_params)
 
         if @course_school_form.valid?
-          Publish::Schools::UpdateCourseSchoolsService.call_or_enqueue(course: @course, params: school_params)
+          Publish::Schools::UpdateCourseSchoolsService.call_or_enqueue(
+            course: @course,
+            school_uuids: selected_school_uuids,
+          )
 
-          flash[:success] = if selected_school_uuids_count > Publish::Schools::UpdateCourseSchoolsService::ENQUEUE_THRESHOLD
+          flash[:success] = if selected_school_uuids.size > Publish::Schools::UpdateCourseSchoolsService::ENQUEUE_THRESHOLD
                               I18n.t("success.enqueued_schools")
                             else
                               I18n.t("success.saved", value: section_key)
@@ -75,9 +78,15 @@ module Publish
       end
 
       def school_params
-        return { school_uuids: nil } if params[:publish_course_school_form][:school_uuids].all?(&:empty?)
+        @school_params ||= params
+          .expect(publish_course_school_form: [{ school_uuids: [] }])
+          .tap do |permitted_params|
+            permitted_params[:school_uuids] = nil if permitted_params[:school_uuids].all?(&:empty?)
+          end
+      end
 
-        params.expect(publish_course_school_form: [:schools_validated, { school_uuids: [] }])
+      def selected_school_uuids
+        @selected_school_uuids ||= Array(school_params[:school_uuids]).compact_blank.uniq
       end
 
       def build_course
@@ -85,11 +94,7 @@ module Publish
       end
 
       def section_key
-        "School".pluralize(selected_school_uuids_count)
-      end
-
-      def selected_school_uuids_count
-        Array(school_params[:school_uuids]).compact_blank.uniq.count
+        "School".pluralize(selected_school_uuids.size)
       end
     end
   end
--- a/app/controllers/publish/courses/schools_controller.rb
+++ b/app/controllers/publish/courses/schools_controller.rb
@@ -29,7 +29,7 @@ module Publish
         if @course_school_form.valid?
           Publish::Schools::UpdateCourseSchoolsService.call_or_enqueue(course: @course, params: school_params)
 
-          flash[:success] = if Array(@course_school_form.site_ids).size > Publish::Schools::UpdateCourseSchoolsService::ENQUEUE_THRESHOLD
+          flash[:success] = if selected_school_uuids_count > Publish::Schools::UpdateCourseSchoolsService::ENQUEUE_THRESHOLD
                               I18n.t("success.enqueued_schools")
                             else
                               I18n.t("success.saved", value: section_key)
@@ -43,6 +43,11 @@ module Publish
         else
           render :edit
         end
+      rescue Publish::Schools::UpdateCourseSchoolsService::UnresolvedProviderSchoolsError,
+             Publish::Schools::UpdateCourseSiteStatusesService::UnresolvedSitesError => e
+        Sentry.capture_exception(e)
+        @course_school_form.errors.add(:school_uuids, :school_uuids_invalid)
+        render :edit, status: :unprocessable_entity
       end
 
       def back
@@ -70,9 +75,9 @@ module Publish
       end
 
       def school_params
-        return { site_ids: nil } if params[:publish_course_school_form][:site_ids].all?(&:empty?)
+        return { school_uuids: nil } if params[:publish_course_school_form][:school_uuids].all?(&:empty?)
 
-        params.expect(publish_course_school_form: [:schools_validated, { site_ids: [] }])
+        params.expect(publish_course_school_form: [:schools_validated, { school_uuids: [] }])
       end
 
       def build_course
@@ -80,7 +85,11 @@ module Publish
       end
 
       def section_key
-        "School".pluralize(Array(school_params[:site_ids]).compact_blank.count)
+        "School".pluralize(selected_school_uuids_count)
+      end
+
+      def selected_school_uuids_count
+        Array(school_params[:school_uuids]).compact_blank.uniq.count
       end
     end
   end
--- a/app/services/publish/schools/update_course_schools_service.rb
+++ b/app/services/publish/schools/update_course_schools_service.rb
@@ -131,7 +131,7 @@ module Publish
       end
 
       def school_names_for_notification
-        course.schools.includes(:provider_school)
+        course.schools.includes(provider_school: :gias_school)
           .map { |course_school| course_school.provider_school.location_name }
           .sort
       end
--- a/app/services/publish/schools/update_course_schools_service.rb
+++ b/app/services/publish/schools/update_course_schools_service.rb
@@ -1,141 +1,139 @@
+# frozen_string_literal: true
+
 module Publish
   module Schools
+    # Coordinates course school updates from Publish by writing Course::School
+    # and legacy SiteStatus rows from the same Provider::School UUIDs.
+    # This service also updates the course and provider so that Apply syncs course changes.
     class UpdateCourseSchoolsService
+      include ServicePattern
+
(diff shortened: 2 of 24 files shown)


app/controllers/publish/courses/schools_controller.rb
```diff
@@ -43,6 +43,11 @@ def update
         else
           render :edit
         end
+      rescue Publish::Schools::UpdateCourseSchoolsService::UnresolvedProviderSchoolsError,
+             Publish::Schools::UpdateCourseSiteStatusesService::UnresolvedSitesError => e
```
inulty-dfe, line 50: We shouldn't use exception driven control flow when we own the objects.

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,140 +1,147 @@
...
+        previous_school_names = school_names_for_notification
+        previous_provider_school_ids = provider_school_ids
+
         ActiveRecord::Base.transaction do
-          assign_attributes_to_course
-          sync_schools
-          apply_publish_status_to_site_statuses
+          provider_schools = resolve_provider_schools
+
+          course.assign_attributes(course_attributes)
```
inulty-dfe: What course_attributes are we assigning here?

inulty-dfe: I think it's just `school_uuids`, if it is, can we just pass that explicitly?

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,140 +1,147 @@
...
 
     private
 
-      attr_reader :course, :params, :previous_site_names
+      attr_reader :course, :params, :submitted_school_uuids
 
-      def assign_attributes_to_course
-        course.assign_attributes(params.except(:site_ids))
+      def course_attributes
+        params.except(:school_uuids)
```
inulty-dfe: What attributes are we assigning here?

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,140 +1,147 @@
...
-          apply_publish_status_to_site_statuses
+          provider_schools = resolve_provider_schools
+
+          course.assign_attributes(course_attributes)
+          update_site_statuses(provider_schools)
+          update_provider_schools(provider_schools)
+
+          # This persists schools_validated and deliberately touches the course
+          # and provider. Apply watches provider.changed_at to decide whether it
+          # needs to sync the provider's courses.
```
inulty-dfe: We have `TouchProvider` and `TouchCourse` working to update the changed_at on each model so do we need to do this?

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,140 +1,147 @@
...
+
         ActiveRecord::Base.transaction do
-          assign_attributes_to_course
-          sync_schools
-          apply_publish_status_to_site_statuses
+          provider_schools = resolve_provider_schools
+
+          course.assign_attributes(course_attributes)
+          update_site_statuses(provider_schools)
+          update_provider_schools(provider_schools)
```
inulty-dfe:
```suggestion
          update_schools(provider_schools)
```

We're adding and removing schools on a course. I find the mention of provider school a bit confusing

app/services/publish/schools/update_course_schools_service.rb
```diff
@@ -1,140 +1,147 @@
...
-        # Reload + scope to new_or_running so we never touch site_statuses
-        # that sync_schools just suspended or destroyed. Iterating the
-        # cached collection used to flip a freshly-suspended row back to
-        # running, leaving the unticked school still attached on the
-        # rendered page.
-        course.site_statuses.reload.new_or_running.each do |site_status|
-          site_status.assign_attributes(site_status_attributes)
-        end
+      def school_names_for_notification
+        course.schools.includes(:provider_school)
```
inulty-dfe:
```suggestion
        course.schools.includes(provider_school: :gias_school)
```

We need to include gias_school for the name

inulty-dfe (commented):
Functionally looks sound

app/controllers/publish/courses/schools_controller.rb
langesi: We validate through the form first so this should only come into play as a last safeguard. Raising the error is still useful because it stops the dual write, reports the failure and its source to Sentry, and then returns to the edit flow with a clear, user-friendly error message.

I'd prefer to keep it. If you feel strongly about it we can chat about it

app/services/publish/schools/update_course_schools_service.rb
langesi: This is a left over from last years valid schools work. I simplified this part to just update the column schools_validated to true.

We'll have another ticket to rip the schools_validated logic out of the app

app/services/publish/schools/update_course_schools_service.rb
langesi: Fixed it, as mentioned above

app/services/publish/schools/update_course_schools_service.rb
langesi: TouchCourse uses update_columns, which bypasses TouchProvider. Also the schools validated feature form the last cycle is a blocker from calling save on course. I added a comment

app/services/publish/schools/update_course_schools_service.rb
langesi: Renamed it, thanks :)

inulty-dfe (approved):

PR #6270, opened 2026-08-04
Start date filter options

tomas-stefano:
## Context

Trello: [Course list — add logic for when to display start date filter options](https://trello.com/c/69eknLzg/1756-course-list-add-logic-for-when-to-display-filters)

The start date filter offers all nineteen months of the recruitment cycle window, most of which narrow a provider's list to nothing. It should offer the months their courses actually start in — so a provider whose courses run in August and September 2026 sees two checkboxes, not nineteen, and adding a course with a new start date adds its month.

No feature flag, no waiting for rollover.

Two decisions worth knowing about:

- **Start date only.** The other groups keep their fixed option lists.
- **Purely data-derived**, so a start date outside the cycle window is still offered. Otherwise a date visible in the course information column would be unfilterable.

New `Publish::Courses::AvailableStartMonths` reads the months from `Provider#courses` — the same base scope `Publish::Courses::Query` builds the list from, so an offered month always matches at least one course.

`FilterForm` is the choke point: every group reader intersects with its own options, so pointing `options_for(:start_date)` at those months narrows the checkboxes, the allowed values and the active filter chips together. Nothing else in the app changes — `CourseList#visible_filter_groups` stays correct by construction, since more than one distinct month now means more than one option.

Months are resolved in the application's zone, not UTC: a course starting `2026-09-01 00:30` is stored as `2026-08-31 23:30 UTC`, but the list displays it under September and `Query` matches it under September.

**Behaviour change:** a bookmarked `?start_date[]=2026-12` where no course starts in December is now dropped — full list, no chip — instead of kept and showing "No courses found". Same as `?level[]=bogus` already behaves.

## Guidance to review

Five commits, each green on its own; commit 2 is the only one that changes behaviour.

Sign in as a provider with courses in two different months, open the course list and expand **Start date** — only those two months. Change one course's start date and the new month appears.

```
bundle exec rspec spec/services/publish/courses/available_start_months_spec.rb \
  spec/forms/publish/courses/filter_form_spec.rb \
  spec/components/publish/courses/filter_panel_component_spec.rb \
  spec/services/publish/course_list_spec.rb \
  spec/requests/publish/courses_spec.rb \
  spec/system/publish/filtering_courses_spec.rb
```

The course list now has 25 system scenarios, up from 23.

## Checklist

### DIFF
--- a/app/services/publish/courses/available_start_months.rb
+++ b/app/services/publish/courses/available_start_months.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Publish
+  module Courses
+    # The months a provider's courses actually start in, for the start date
+    # filter on the publish course list. Offering the whole cycle window instead
+    # would fill the panel with months that narrow the list to nothing.
+    #
+    # Read from +Provider#courses+, the same base scope +Publish::Courses::Query+
+    # builds the list from, so an offered month always matches at least one
+    # course and a listed course's month is always offered.
+    module AvailableStartMonths
+      # The month is resolved in the application's zone, never in UTC: a course
+      # starting 2026-09-01 00:30 is stored as 2026-08-31 23:30 UTC, but the list
+      # displays it under September and Query matches it under September, so a
+      # UTC month would offer an August option that finds nothing.
+      #
+      # Dates rather than times, because the checkbox label comes from
+      # I18n.l(month, format: :short), which formats the two quite differently.
+      def self.for(provider)
+        provider.courses.distinct.pluck(:start_date).compact.map { |start_date|
+          start_date.in_time_zone.to_date.beginning_of_month
+        }.uniq.sort
+      end
+    end
+  end
+end
--- a/spec/services/publish/courses/available_start_months_spec.rb
+++ b/spec/services/publish/courses/available_start_months_spec.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Publish::Courses::AvailableStartMonths do
+  subject(:months) { described_class.for(provider) }
+
+  let(:provider) { create(:provider) }
+
+  it "is empty when the provider has no courses" do
+    expect(months).to be_empty
+  end
+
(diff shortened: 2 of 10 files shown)


app/services/publish/courses/available_start_months.rb
```diff
@@ -0,0 +1,27 @@
...
+    module AvailableStartMonths
+      # The month is resolved in the application's zone, never in UTC: a course
+      # starting 2026-09-01 00:30 is stored as 2026-08-31 23:30 UTC, but the list
+      # displays it under September and Query matches it under September, so a
+      # UTC month would offer an August option that finds nothing.
+      #
+      # Dates rather than times, because the checkbox label comes from
+      # I18n.l(month, format: :short), which formats the two quite differently.
+      def self.for(provider)
+        provider.courses.distinct.pluck(:start_date).compact.map { |start_date|
```
inulty-dfe, line 21: Nice, I never though to use distinct before pluck

spec/services/publish/courses/available_start_months_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  it "offers the month each course starts in" do
+    create(:course, provider:, start_date: Time.zone.local(2026, 9, 1))
+
+    expect(months).to eq([Date.new(2026, 9, 1)])
+  end
+
+  it "orders them earliest first" do
+    create(:course, provider:, start_date: Time.zone.local(2027, 1, 1))
```
inulty-dfe: I don't think we should hard code cycle years in specs cause we should expect them to fail if something changes in a new cycle

inulty-dfe (approved):

gms-gs (approved):

kelliedesigner:
Design review - approved!

kelliedesigner (approved):
Looks great!

spec/services/publish/courses/available_start_months_spec.rb
tomas-stefano: Done

PR #6274, opened 2026-08-04
Removing School remodel feature flags from publish and API

langesi:
## Context

Removed the use of the school data feature flag for everything but the Query logic.

What we'll keep temporarily:
- [app/services/courses/query.rb (line 300)](/Users/simonlangework/Documents/engineering/DFE/publish-teacher-training/app/services/courses/query.rb:300): Find course results location query.
- [app/services/saved_courses/query.rb (line 38)](/Users/simonlangework/Documents/engineering/DFE/publish-teacher-training/app/services/saved_courses/query.rb:38): saved-course distance query.
- [app/controllers/find/sitemaps_controller.rb (line 13)](/Users/simonlangework/Documents/engineering/DFE/publish-teacher-training/app/controllers/find/sitemaps_controller.rb:13): extra Find path I found; it switches between legacy and new search services, so I’d retain it for the same reason as Find results.

- Remove or simplify usage of school data flag 
- Cleanups
	- Update stale comments in SchoolPresenceExemption and CreationService that still describe this as a publish-side rollout flag.
	- Keep the feature flag in [app/lib/feature_flags.rb (line 14)](/Users/simonlangework/Documents/engineering/DFE/publish-teacher-training/app/lib/feature_flags.rb:14), but change its description so it reflects the remaining Find-only use.

## Guidance to review

You can go through the commits to review, they are organised

## Checklist


app/decorators/course_decorator.rb
```diff
@@ -153,28 +153,16 @@ def alphabetically_sorted_sites
...
-    if FeatureFlag.active?(:course_publishing_uses_new_school_model)
-      object.schools.count
-    else
-      object.sites.school.count
-    end
+    attached_schools.size
+  end
+
+  def attached_schools
+    object.schools.to_a.uniq { |school| school.provider_school_id || school.provider_school&.id || school.object_id }
```
tomas-stefano: I think might be a problem here.

In Ruby 3.4 object_id is a small monotonic Integer  the same value space as provider_school_id. A course holding a persisted school with provider_school_id == 24 plus an in-memory school built without a provider_school will silently collapse the two, dropping a school from both `sorted_school_names` and `attached_schools_count`.

tomas-stefano: The dedup also isn't needed: course_school has a unique index on (course_id, provider_school_id) (db/schema.rb:265) and has_many :schools doesn't join, so duplicates can't arise. The middle provider_school&.id term additionally fires a query per row whenever provider_school_id is nil. Suggest just object.schools.

app/decorators/course_decorator.rb
```diff
@@ -153,28 +153,16 @@ def alphabetically_sorted_sites
...
-  # the schools list keeps working from either data model.
   def sorted_school_names
-    names = if FeatureFlag.active?(:course_publishing_uses_new_school_model)
-              object.gias_schools.map(&:name)
-            else
-              object.sites.map(&:location_name)
-            end
-
-    names.sort_by(&:downcase)
+    attached_schools.map { |school| school.gias_school.name }.sort_by(&:downcase)
```
tomas-stefano: `object.gias_schools.map(&:name)` was a single `has_many :through` query. Now it's 1 query for schools + 1 per school for gias_school, and nothing preloads schools: :gias_school on the publish side (only app/services/saved_courses/query.rb:86 joins it). 

Call sites are `publish/courses/schools/edit.html.erb:37` via `AttachedSchoolsSummaryComponent`,
 ` _basic_details_tab.html.erb:104` and `confirmation.html.erb:104` — pages where a large SCITT has hundreds of attached schools, and where two of the three only want .length. 

Needs `.includes(:gias_school)`, or keep using gias_schools.

tomas-stefano:
`app/views/publish/courses/schools/edit.html.erb:37` — the count and the checkboxes below it now read different data models - `attached_schools_count` reads `course_school`, but the checkbox list immediately under it is built
  from `@course_school_form.sites / course.site_ids`, and `Publish::CourseSchoolForm#no_schools_selected (app/forms/publish/course_school_form.rb:36)` still validates against course.sites.school. For a course with school sites but no course_school rows, this page renders "You do not have any schools attached to this course" directly above a list with those schools ticked. The flag previously kept both halves on the same model.

Simons commnet: 

We have another PR that will move the checkboxed to use the new data model with uuids

spec/services/courses/creation_service_spec.rb
```diff
@@ -668,29 +630,20 @@
       end
     end
 
-    context "when no schools are selected" do
```
tomas-stefano, line 673: I think we should not remove this spec, unless I understood wrong. Why this context was deleted?

tomas-stefano:
`spec/factories/sites.rb:44`, `spec/factories/provider_schools.rb:20`, `spec/factories/course_schools.rb:25` — same 30-line block copy-pasted three times, and it silently ignores overrides

  GiasSchool.find_or_create_by!(urn: site.urn) { ... } only applies the site's name/address/postcode when it creates. If a GiasSchool with that URN already exists from another factory, the site's attributes are dropped and a spec asserting on the school name gets the pre-existing one — a confusing failure to debug. Worth extracting to one shared helper and deciding explicitly whether existing records get updated.

app/validators/course_publishable_schools_rollover_validator.rb
```diff
@@ -6,15 +6,13 @@
...
-# Courses::PublishRules::SchoolPresence so flag-on reads from the
-# Course::School model instead of the legacy Site association.
+# School presence is read from the Course::School association.
 class CoursePublishableSchoolsRolloverValidator < ActiveModel::Validator
   def validate(course)
     return if course.schools_validated?
     return unless course.latest_enrichment&.rolled_over?
 
-    if Courses::PublishRules::SchoolPresence.any?(course)
+    if course.schools.any?
```
tomas-stefano, line 15: **`course_publishable_schools_presence_validator.rb:7` — this can make a course impossible to publish**

Publish validation now reads school presence only from `course_school`. But some courses have schools attached as `site` rows with no matching `course_school` row — the backfill skips them (`no_urn`, `provider_school_missing`, and it doesn't run before cycle 2026), and both `CreationService` and `UpdateCourseSchoolsService` skip the write when there's no `Provider::School`.

Those courses now fail publish with "Select at least one school" — and the provider can't fix it. The edit page shows their schools already ticked, so saving writes nothing.

Could we add the same cycle guard used elsewhere (`provider_schools/identity.rb:25`, `api/public/v1/application_controller.rb:21`)? Or check `DataHub::SchoolsBackfillProcessSummary.last` — if `course_sites_skipped` is 0, this can't happen and we can ignore it.

app/decorators/course_decorator.rb
langesi: Thanks! You are right, that's fixed

app/decorators/course_decorator.rb
langesi: Yep, you're right, I fixed it.

app/validators/course_publishable_schools_rollover_validator.rb
langesi: Our data model are synced, also, we have PRs to also read from the new data model. I can merge the other PR before if that's a concern.

All that we are doing now is to move the validation from the site model to the new data model as we will display the new data model

tomas-stefano (approved):

PR #6276, opened 2026-08-05
Search placement schools

tomas-stefano:
Providers struggle to attach the right schools to a course: the list shows every school on their account. This adds a search box to both school pickers — the course schools edit page and the add course wizard step.

## What it does

- Search by school name, postcode or URN. Postcodes match with or without the space.
- A dfe-autocomplete dropdown suggests up to 5 schools after 3 characters. Choosing one fills the box; the list only changes when Search is pressed.
- Results are never collapsed — every match is shown, past the 20-school threshold.
- No results shows "No results found. Clear your search and try again." with a Show all schools link.
- Clear search returns to the full, collapsed list.
- Select all is hidden while a search is active: it attaches every school, including ones filtered out of view.

## How

Searching is client-side. The list holds unsaved selections, and a server round trip would either discard them or need every selected id in the URL (200+ schools is past what browsers accept). Rows are only hidden, never removed, so a school ticked and then filtered out is still submitted.

`show-all-schools` is renamed to `schools-list` and now owns row visibility. Visibility is one value derived from two rules — `matches(query) && (expanded || index < 20)` — and a search suspends the collapse, so both rules feed one `render`.

The dropdown and the checkbox filter both rank through dfe-autocomplete's exported `defaultSort`, so a search means the same thing in each. Postcode and URN travel as `data-synonyms`, matched but never displayed; the town is dropdown context only and deliberately not searchable.

The panel is rendered hidden and revealed on connect, so without JavaScript both pages are unchanged. Its select carries no name and both actions are buttons, so nothing in it is submitted with the form it sits inside.

## Worth knowing

`dfe_autocomplete_options` in the dfe-autocomplete gem wraps appended text in `<strong>`, but the library's renderer escapes it, so that markup reaches the dropdown as its own source. This component sends plain text. Anywhere else in the app passing `append:` will have the same bug.


app/components/publish/schools/search_panel_component.html.erb
```diff
@@ -0,0 +1,59 @@
+<div class="app-school-search" hidden data-qa="school-search-panel" data-schools-list-target="panel">
```
inulty-dfe: This PR has added data-qa throughout, but we decided a long time ago to avoid using this to power our elemen selection in the test suite

tomas-stefano:
https://trello.com/c/Rhc64jn9/1740-add-search-to-attach-schools-page

app/components/publish/schools/search_panel_component.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        options.unshift([nil, nil, nil])
+      end
+
+    private
+
+      attr_reader :schools, :value
+
+      def synonyms_for(school)
+        [school.urn, school.postcode, school.postcode&.delete(" ")].compact_blank.uniq
```
inulty-dfe, line 57: Why are we putting both these in? `school.postcode, school.postcode&.delete(" ")`

Is there a case when they evaluate to different values?

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,211 @@
...
+// clearing the search brings the collapsed list back. So neither rule writes to
+// the rows itself; both feed `render`, which is the only place visibility is
+// set.
+//
+// Rows are only ever hidden, never removed, so a school that is ticked and then
+// filtered or collapsed out of view is still submitted.
+//
+// The search half only wakes up on pages that render a search panel. Without
+// one this is still just the collapsing list.
+export default class extends Controller {
```
inulty-dfe, line 22: I found this controller hard to follow. It looks like its a combination of the old `show_all_schools` controller and a new autocomplete filter controller.

There's methods in here that aren't clearly delineated and it makes understanding the code quite difficult

inulty-dfe (commented):
I haven't reviewed everything, just got a bit stuck on the schools_list_controller

app/components/publish/schools/search_panel_component.rb
tomas-stefano: dfe-autocomplete matches word prefixes, not substrings, so SY2 5RJ is only found by sy2 or 5rj and someone typing sy25rj would match neither. So decided to do this too.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: I refactored. Can you check if is better. Let me know 🚀

app/javascript/publish/controllers/schools_search_controller.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  disconnect () {
+    this.instance?.destroy()
+  }
+
+  search () {
+    const query = (this.input()?.value || '').trim()
+
+    if (query === '') return this.clear()
```
langesi, line 68: Just used the review app and there was an error here:

`TypeError: this.clear is not a function`

So if you remove everything from the search, it'll error as the function clear doesn't exist. Did you mean this.clearSearch ?

tomas-stefano: Done

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,104 @@
...
+
+    this.expanded = true
+    this.render()
+
+    // The button we just hid held focus, which would otherwise fall back to the
+    // document root. Move focus to the first school we revealed instead.
+    firstCollapsed?.focus()
+  }
+
+  render () {
```
inulty-dfe, line 43: Could we order the definition of the methods based on their primacy / order they appear. So I think render should be the first method under `connect` cause `connect` calls `render`. Then `matchingSchools` cause that's the first method that `render` calls.

It makes reading the code much easier

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,104 @@
...
+    const matching = this.matchingSchools()
```
inulty-dfe: I don't know what `matching` means here. Matching what?

app/javascript/publish/controllers/schools_search_controller.js
```diff
@@ -0,0 +1,124 @@
...
+
+    // An empty box is no search at all, which is the state clearing it leaves
+    // behind. Matching on the empty query instead would allow every school while
+    // still counting as a search, suspending the collapse and hiding select all.
+    if (query === '') return this.clearSearch()
+
+    const allowed = matchingValues(query, this.options)
+
+    this.noResultsTarget.hidden = allowed.length > 0
+    this.dispatch('filter', { detail: { allowed } })
```
inulty-dfe: As far as I can tell, this is the first time in the app we use Stimulus dispatch.

Could we add this to the PR description

https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events

app/assets/stylesheets/publish/components/_school-search.scss
```diff
@@ -0,0 +1,49 @@
+// A plain grey panel - deliberately not .app-callout, which is inset text and
+// so comes with a grey rule down its left edge.
```
inulty-dfe: How much of `.app-callout` is this duplicating? I think this should be a `.app-callout` with a modifier `--no-edge`. I think you can come up with a better modifier name than that though.

app/components/publish/schools/search_panel_component.rb
```diff
@@ -0,0 +1,71 @@
...
+  module Schools
+    # The search box above a list of school checkboxes, shared by the course
+    # schools edit page and the add course wizard's schools step.
+    #
+    # Everything here is inert until the schools-list Stimulus controller
+    # connects: the panel is hidden, the select carries no name and both actions
+    # are plain buttons, so a provider without JavaScript sees - and submits -
+    # exactly what they did before.
+    class SearchPanelComponent < ViewComponent::Base
+      SELECT_ID = "school-search".freeze
```
inulty-dfe: Is there a reason this is a constant? It's declared in a component called `SearchPanelComponent`.

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,104 @@
...
+// GOV.UK sets `display: flex` on `.govuk-checkboxes__item`, which overrides the
+// `[hidden]` user-agent rule, so we hide rows with the `!important` display
+// utility class as well as the `hidden` attribute.
+const HIDDEN_CLASS = 'govuk-!-display-none'
+
+// Owns which schools in a checkbox list are on screen.
+//
+// Two rules decide that. A search says which schools are allowed at all - it
+// arrives as `allowed`, a set of uuids, or null when nothing is being searched
+// for. The collapse threshold then decides how many of those to show. They
```
inulty-dfe: I don't think `allowed` is the right term here. Allowed what? To be displayed?
Also it's double function of meaning "here are the search results" and "is a search happening" should be separated somehow.

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,104 @@
...
+// neither writes to the rows itself: both feed `render`, the only place
+// visibility is set.
+//
+// Rows are only ever hidden, never removed, so a school that is ticked and then
+// filtered or collapsed out of view is still submitted.
+//
+// Pages without a search panel simply never send `allowed`, and this stays what
+// it has always been: the collapsing list.
+export default class extends Controller {
+  static targets = ['school', 'showAll', 'bulkSelect']
```
inulty-dfe: I wonder if anyone else would be in favour of suffixing the targets with the type of element they are. `schoolCheckBox`

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,104 @@
...
+  static values = { visible: { type: Number, default: 20 } }
```
inulty-dfe: Could this be `maxVisibleSchools` or something a bit more specifically related to the meaning of the number than `visible`

inulty-dfe (commented):
I think the separation of the controllers makes a big difference, I'm concerned many of the variable names aren't particularly indicative. A lot of my difficulty is around the controllers again, but for different reasons.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Done — reordered depth-first from connect: connect, clear, render, schoolsMatchingSearch,  collapsing, setRowHidden, setHidden, row, then the externally-triggered filter, restore, showAll.  connect now just calls clear(), since the initial state and the cleared state are the same thing —  that also dropped a duplicated three-line body.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Renamed the method to schoolsMatchingSearch() and the local to matchingSearch, so both sites now say what's being matched. Added a line noting that with no search running every school matches, which is why it still returns all of them.

app/javascript/publish/controllers/schools_search_controller.js
tomas-stefano: Good spot. The description now links the Stimulus docs and says why events rather than outlets: outlets couple the two controllers directly, and the list has to keep working on pages with no search panel.

app/assets/stylesheets/publish/components/_school-search.scss
tomas-stefano: The overlap is six declarations — padding at both breakpoints, the grey background, the responsive top/bottom margins, the last-child reset. .app-callout extends .govuk-inset-text, so adopting it also brings the 5px left rule, govuk-font(19), the text colour, clear: both, a > :first-child reset and six colour modifiers; a --no-edge callout is a callout with its defining feature switched off, and _callout.scss is shared with Find. This panel was deliberately moved off .app-callout in 2a9a90a7b for exactly that reason, so I've kept it as its own block and expanded the file's comment to name the six shared declarations — happy to revisit if you still think it should be a modifier.

app/components/publish/schools/search_panel_component.rb
tomas-stefano: No good reason — replaced with private select_id / hint_id methods. That also removed the verbose Publish::Schools::SearchPanelComponent::SELECT_ID in its own template and fixed a latent drift: "school-search-hint" was a separate string literal, so the hint id could diverge from the select id. It now derives.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Agreed, gone. The two meanings are now separate fields — searching (boolean) and searchMatches (Set) — set together in filter/clear, and searching() the method became this.searching the boolean. The null sentinel is gone from the wire too: schools-search now dispatches three events (filter carrying detail.matches, clear, restore) instead of overloading filter with a null payload.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Taken for this one: school → schoolCheckbox, since the controller already calls the variable checkbox internally. Worth knowing for the wider question though — 28 of the 35 targets in the codebase are bare role nouns, and the compound style (searchInput, optionsList) only appears in filter_search_controller.js, the oldest one. So this is against the current grain; happy either way if the team wants to settle on a convention.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Agreed, renamed — though to collapseAfter rather than maxVisibleSchools: the Ruby side already calls this SchoolsList::COLLAPSE_AFTER and schools_collapse_threshold, so this ties the two vocabularies together instead of introducing a third name for the same number. The attribute is now data-schools-list-collapse-after-value.

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,113 @@
...
+  clear () {
+    this.searching = false
+    this.searchMatches = new Set()
+    this.expanded = false
+
+    this.render()
+  }
+
+  render () {
+    const matchingSearch = this.schoolsMatchingSearch()
```
inulty-dfe: Does `searchResults` make better sense here?

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,113 @@
...
+
+    if (this.hasShowAllTarget) {
+      this.showAllTarget.hidden = !(this.collapsing() && matchingSearch.length > this.collapseAfterValue)
+    }
+
+    // Select all takes in every school, including the ones a search has filtered
+    // out of view, so hide it while searching rather than leave it offering to
+    // do something wider than what is on screen. It stays in the DOM, so it goes
+    // on tracking the boxes ticked while filtered and is right when it returns.
+    this.bulkSelectTargets.forEach(element => this.setHidden(element, this.searching))
```
inulty-dfe, line 57: Is this an implementation cost? If we removed the rows from the DOM, this wouldn't be an issue. The full list of schools could be kept in the auto complete. Even if they were disabled, they wouldn't submit.

I'm just mentioning this rather than suggesting it.

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,113 @@
...
+
+  // A search is never collapsed: the point of searching a long list is to see
+  // every match.
+  collapsing () {
+    return !this.searching && !this.expanded
+  }
+
+  setRowHidden (checkbox, hidden) {
+    this.setHidden(this.row(checkbox), hidden)
+  }
```
inulty-dfe: I'm not sure there's any value in wrapping `setHidden` with `setRowHidden`

app/javascript/publish/controllers/schools_search_controller.js
```diff
@@ -0,0 +1,125 @@
...
+
+    // An empty box is no search at all, which is the state clearing it leaves
+    // behind. Matching on the empty query instead would allow every school while
+    // still counting as a search, suspending the collapse and hiding select all.
+    if (query === '') return this.clearSearch()
+
+    const matches = matchingValues(query, this.options)
+
+    this.noResultsTarget.hidden = matches.length > 0
+    this.dispatch('filter', { detail: { matches } })
```
inulty-dfe: Why `detail`? This is just an arbitrary label to contain the payload?
Would `data` be more conventional?
Or is it necessary at all to nest the matches (or results) in an object with the key `data` at all?

app/javascript/publish/controllers/schools_search_controller.js
```diff
@@ -0,0 +1,125 @@
...
+
+  search () {
+    const query = (this.input()?.value || '').trim()
+
+    // An empty box is no search at all, which is the state clearing it leaves
+    // behind. Matching on the empty query instead would allow every school while
+    // still counting as a search, suspending the collapse and hiding select all.
+    if (query === '') return this.clearSearch()
+
+    const matches = matchingValues(query, this.options)
```
inulty-dfe: I think results is better than `matchingValues` or `matches`

app/javascript/publish/controllers/schools_search_controller.js
```diff
@@ -0,0 +1,125 @@
...
+    this.dispatch('clear')
+    this.input()?.focus()
+  }
+
+  // "Show all schools" from the no results message. The same as clearing the
+  // search, except the button is about to hide itself, so focus has to move -
+  // and only the list knows where its first row is.
+  showAllSchools () {
+    this.reset()
+    this.dispatch('restore')
```
inulty-dfe, line 96: How do we know that another controller later that listens for a restore event doesn't receive this one by accident? This applies to all the events in this PR

app/components/publish/schools/search_panel_component.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    private
+
+      attr_reader :schools, :value
+
+      # accessible-autocomplete gives the input it renders the select's id, so
+      # this is also what the schools-search controller looks the box up by.
+      def select_id
+        "school-search"
+      end
```
inulty-dfe, line 48: When I asked if this needed to be a constant, I was really asking if it can just be hard coded in the template. I see it's used in `hint_id` so i guess i hardens the case for it being referencable

app/javascript/publish/controllers/schools_list_controller.js
```diff
@@ -0,0 +1,113 @@
...
+  // ruled out.
+  schoolsMatchingSearch () {
+    if (!this.searching) return this.schoolCheckboxTargets
+
+    return this.schoolCheckboxTargets.filter(checkbox => this.searchMatches.has(checkbox.value))
+  }
+
+  // A search is never collapsed: the point of searching a long list is to see
+  // every match.
+  collapsing () {
```
inulty-dfe: This present participle form doesn't fit with the rest of the function and variable names. And it doesn't describe an action that is in process. Is `collapsed` more appropriate here?

`searching` makes sense but I don't think it extends to `collapsing`

inulty-dfe (commented):
This is a big improvement from the single controller.
I've added some more comments, I appreciate this is a hard piece of work to get right.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Yes. I made the changes - the  field this.searchResults holds the school values the query returned, while the local holds the  checkbox elements still in play, and giving two different types the same word seemed worse than  either name alone. So it's searchResults end to end for the payload (searchResults() in  schools_search.js, detail.results on the wire, this.searchResults on the list) and schools for the rows.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Agreed, gone — it was a one-line wrapper with a single caller. render() now does this.setHidden(this.row(checkbox), !shown.has(checkbox)) directly. setHidden stays because  bulkSelectTargets uses it too, and row stays for the focus lookup in restore.

app/javascript/publish/controllers/schools_search_controller.js
tomas-stefano: detail isn't Stimulus's label — it's the native CustomEvent property. dispatch is:

`const event = new CustomEvent(type, { detail, bubbles, cancelable })` (stimulus.js:2548), 

so detail goes straight into the DOM CustomEventInit, whose only recognised keys are detail, bubbles, cancelable and composed; a data key would be silently dropped by the constructor. The listener reads the
  read-only native event.detail. Nesting isn't optional either — dispatch exposes no other channel, so the only way to un-nest would be hand-rolling the event and mutating the instance before dispatchEvent.

app/javascript/publish/controllers/schools_search_controller.js
tomas-stefano: Done — the exported function is now searchResults(query, options), the local is results, and the
  event carries detail.results. spec/javascript/schools_search.spec.js renamed with it.

app/javascript/publish/controllers/schools_search_controller.js
tomas-stefano: No — dispatch namespaces by controller identifier: 

`const type = prefix ? \${prefix}:${eventName}`  : eventNamewithprefix = this.identifier`(stimulus.js:2547). 

So these fire  as `schools-search:filter/:clear/:restore`, never bare, and that's the exact string the `data-action` parser consumes. Nothing else in the repo dispatches at all, and there's no  barefilter/clear/restorelistener anywhere. 

A future controller calling `this.dispatch('restore')` would emit under its own identifier — it could only collide by explicitly passing a matching or falsy prefix.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: Agreed and renamed. searching is a real ongoing state; collapsing was just borrowing its shape while reporting a fact about the list.

app/javascript/publish/controllers/schools_list_controller.js
tomas-stefano: The catch is that disabled and removed controls both don't submit, which is the one thing we can't have here — a school ticked before a search has to stay attached while it's filtered out of view, and there's a spec pinning that journey. 

**Keeping rows in the DOM also lets the browser own the checked state for free**; remove them and we're stashing ticks, re-creating rows and restoring them on clear. 

That said, you're right that hiding "Select all" is a compromise — it's just not caused by keeping the rows. It's caused by select-all-checkboxes operating on every checkbox regardless of visibility; scope that to visible rows and "Select all" could stay and mean "select all results", with no rows removed and no state to own.

app/components/publish/schools/search_panel_component.rb
tomas-stefano: about the referencing of the  string: label[for], the hint's id/aria-describedby, and the JS, which reads the select's id to tell the enhanced input apart from the hint input accessible-autocomplete renders beside it — so one method is the minimum that keeps a single  source, and hardcoding is what let "school-search-hint" drift into a separate literal. 

So I think we should keep as is, what do you think?

gms-gs (approved):
LGTM, I agree with Iain's points. So if all have been actioned lets get it merged

PR #6292, opened 2026-08-05
Fix invalid applications open date created during course rollover

langesi:
Context

We ran the rollover a few times in review apps. During that some rolled-over courses had an invalid applications_open_from date in September 2027. This was due to the wrong application start date set during the rollover task.

The recruitment cycle application window is from 7 October 2026 to 23 September 2027. The generated date falls outside the permitted range, causing course validation to fail when the course is updated.

An affected example is course F220 for provider 1CP.

Changes proposed in this pull request

* Correct the rollover logic so rolled-over courses receive a valid applications_open_from date for the new recruitment cycle
* Add test coverage for the date assigned during course rollover
* Prevent rolled-over courses from failing validation because of an automatically generated application start date

## Guidance to review

Check the code. I'll start the roll over, you guys can do QA as well.

## Checklist

### DIFF
--- a/app/forms/support/recruitment_cycle_form.rb
+++ b/app/forms/support/recruitment_cycle_form.rb
@@ -52,7 +52,11 @@ module Support
       if expected_application_start_date.blank?
         errors.add(:application_start_date, :missing_cycle_timetable)
       elsif application_start_date != expected_application_start_date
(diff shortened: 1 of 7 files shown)


gms-gs (approved):
LGTM!

PR #6296, opened 2026-08-07
Touch courses with no schools when their providers schools change

inulty-dfe:
## Context

Since we have allowed some courses to be published without attaching schools, we have decided to inject all the providers schools into the API response for those schools locations.

This creates a dependency between those courses and the providers schools.

When the providers schools change we need to touch the no-schools courses becuase their API payload changes in tandem.

We need to touch the no-schools courses when a providers schools change. So that means when one or more is added, or when one is deleted.

When one or more is added, we suppress the model callback which would fire N times, for multiple school additions. We then fire the touch once when the schools are added.

### Avoiding over reach

If  Provider::School is saved without any real changes (call `#save`, the updated_at is changed) then we want to avoid the cascade of touches. So scope the `update` callback to only cases where some properties of value have changed.

## Guidance to review

- Add `TouchNoSchoolCourses` to add callbacks on `Provider::School`
- ADd `TouchParents` to touch the provider and all the providers no-school courses in one go

## Checklist

### DIFF
--- a/app/models/concerns/touch_no_school_courses.rb
+++ b/app/models/concerns/touch_no_school_courses.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+# A course that is allowed to publish without schools is served by the API with
+# all of its provider's schools as locations, so writing a provider school
+# changes that course's payload. Bumping `changed_at` is what tells API
+# consumers to re-fetch it.
+#
+# Shaped like TouchProvider: the write itself is the trigger, so every writer
+# gets this for free. Bulk writers that would otherwise repeat the sweep once
+# per row wrap their loop in TouchSuppression.suppress and call
+# `.touch_no_school_courses_for(provider)` once at the end instead.
+#
+# after_commit rather than after_save/after_destroy: ProviderSchools::Removal
+# destroys inside `with_lock`, and sweeping there would hold that row lock for
+# the length of the sweep and take course row locks under it — the opposite
+# order to attaching a school, which takes the provider school's key-share lock
+# first and then touches the course. Running after the transaction commits
+# keeps the two paths from deadlocking, and only touches once the write is
+# durable.
+module TouchNoSchoolCourses
+  extend ActiveSupport::Concern
+
+  # The columns a fallback location is serialized from: `code` reads
+  # `site_code`, and everything else is read through `gias_school`. An update
+  # that leaves both alone cannot change what the API serves, and
+  # `after_commit on: :update` fires even for a save that issued no UPDATE at
+  # all, so gate the update case on them. Create and destroy always change the
+  # set of schools served.
+  PAYLOAD_COLUMNS = %w[gias_school_id site_code].freeze
+
+  # Two callbacks, two method names: ActiveSupport dedupes callbacks by filter,
+  # so registering the same symbol twice replaces the first registration rather
+  # than adding to it.
+  included do
+    after_commit :touch_no_school_courses, on: %i[create destroy]
+    after_commit :touch_no_school_courses_on_update, on: :update
+  end
+
+  class_methods do
+    def touch_no_school_courses_for(provider)
+      # `changed_at` is UNIQUE, so every row needs its own timestamp.
+      # update_columns skips the per-course provider touch — TouchProvider has
+      # already bumped the provider once.
+      Courses::PublishRules::SchoolPresenceExemption
+        .falling_back_to_provider_schools(provider)
+        .find_each { |course| course.update_columns(changed_at: Time.zone.now) }
+    end
+  end
+
+private
+
+  def touch_no_school_courses
+    return if TouchSuppression.suppressed?
+
+    self.class.touch_no_school_courses_for(provider)
+  end
+
+  def touch_no_school_courses_on_update
+    return unless saved_changes.keys.intersect?(PAYLOAD_COLUMNS)
+
+    touch_no_school_courses
+  end
+end
--- a/app/controllers/publish/providers/schools/check_multiple_controller.rb
+++ b/app/controllers/publish/providers/schools/check_multiple_controller.rb
@@ -12,11 +12,17 @@ module Publish
         def update
           saved_schools = []
 
-          gias_schools.each do |gias_school|
-            ActiveRecord::Base.transaction do
-              saved_schools << create_provider_school_and_legacy_site(gias_school:)
+          # Each school would otherwise touch the provider and sweep its
+          # no-school courses on save. Suppress that and stamp them once.
+          TouchSuppression.suppress do
+            gias_schools.each do |gias_school|
+              ActiveRecord::Base.transaction do
+                saved_schools << create_provider_school_and_legacy_site(gias_school:)
+              end
             end
           end
+
+          ::ProviderSchools::TouchParents.call(provider:) if saved_schools.any?
 
           schools_added_message(saved_schools)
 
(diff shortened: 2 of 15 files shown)


inulty-dfe:
[No Schools - Touch a course, if new school added](https://trello.com/c/f5Cf7IgI/1856-no-schools-touch-a-course-if-new-school-added)

app/controllers/concerns/touch_no_school_courses.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# A course that is allowed to publish without schools is served by the API with
+# all of its provider's schools as locations, so changing that provider's
+# schools changes the course's payload. Bumping `changed_at` is what tells API
+# consumers to re-fetch it.
+module TouchNoSchoolCourses
+  extend ActiveSupport::Concern
```
tomas-stefano: I think we can drop `extend ActiveSupport::Concern`.  there's no included do and no class_methods, so it's just `module X; private; def y; end; end`

app/controllers/concerns/touch_no_school_courses.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def touch_all_no_school_courses
+    # `changed_at` has a UNIQUE index, so every row needs its own timestamp.
+    # update_columns skips the provider touch, which the provider school write
+    # has already done.
+    provider.courses.publishable_without_schools.find_each do |course|
+      course.update_columns(changed_at: Time.zone.now)
+    end
+  end
+end
```
tomas-stefano: I think we could move this on the model, not on six controllers**

Six `include`s and six call sites is six chances to forget one — `Rollover::Schools::ProviderCopier:14` already creates `Provider::School` rows and is missed. `Provider::School` is the choke point and already carries `include TouchProvider` for exactly this job; mirror it.

First, keep the predicate and its scope in the class that claims to own it, so they can't drift:

```ruby
# app/services/courses/publish_rules/school_presence_exemption.rb
EXEMPT_FUNDINGS = %w[salary apprenticeship].freeze

def self.applies?(course)
  course.publish_without_schools_allowed? && EXEMPT_FUNDINGS.include?(course.funding)
end

# Courses whose API locations fall back to the provider's schools
# (LocationsController#remodelled_locations) — so a provider school
# write changes their payload.
def self.falling_back_to_provider_schools(provider)
  provider.courses
          .where(publish_without_schools_allowed: true, funding: EXEMPT_FUNDINGS)
          .where.missing(:schools)
end
```

Then the concern becomes a model concern shaped like `TouchProvider`:

```ruby
# app/models/concerns/touch_no_school_courses.rb
module TouchNoSchoolCourses
  extend ActiveSupport::Concern

  included do
    after_save :touch_no_school_courses
    after_destroy :touch_no_school_courses
  end

private

  def touch_no_school_courses
    return if TouchSuppression.suppressed?

    # `changed_at` is UNIQUE, so every row needs its own timestamp.
    # update_columns skips the per-course provider touch — TouchProvider
    # has already bumped the provider once.
    Courses::PublishRules::SchoolPresenceExemption
      .falling_back_to_provider_schools(provider)
      .find_each { |course| course.update_columns(changed_at: Time.zone.now) }
  end
end
```

```ruby
# app/models/provider/school.rb
class Provider::School < ApplicationRecord
  include TouchProvider
  include TouchNoSchoolCourses   # after TouchProvider, so the provider is touched first
```

I know I am not a fan of callbacks but this might the only few scenarios where it could make sense.

That drops all six controller `include`s and all six call sites, kills the inside-the-`with_lock` call in `publish/providers/schools/checks_controller.rb:16`, picks up rollover and any future writer for free, and honours `TouchSuppression` — which matters now that it runs from a callback, since `Providers::CopyToRecruitmentCycleService:21` suppresses touches during bulk rollover.

app/controllers/publish/providers/schools/checks_controller.rb
```diff
@@ -4,14 +4,18 @@ module Publish
...
+
         helper_method :school_id
         before_action :site
 
         def show; end
 
         def update
           provider_school = ActiveRecord::Base.transaction do
-            create_provider_school_and_legacy_site
+            created = create_provider_school_and_legacy_site
```
tomas-stefano: We might not have this problem during the cycle. Only in the beginning of the cycle if any of those schools are being updated at same time by different users.

`ProviderSchools::Creator` takes `@provider.with_lock` (`creator.rb:20`). Nested inside this outer `ActiveRecord::Base.transaction` that's just a savepoint, so the `SELECT … FOR UPDATE` on the provider row isn't released when the inner block exits — it's held until the *outer* transaction commits. As written, that means every per-course `UPDATE` from `touch_all_no_school_courses` runs with the provider exclusively locked, blocking any concurrent school add/remove for that provider for the duration.

The touch isn't part of the write's atomicity — it's cache invalidation — so it can sit outside, which is also what the other five call sites do:

```ruby
def update
  provider_school = ActiveRecord::Base.transaction do
    create_provider_school_and_legacy_site
  end

  touch_all_no_school_courses

  redirect_to publish_provider_recruitment_cycle_schools_path,
              flash: { success_with_body: { title: t(".added"), body: provider_school.gias_school.name } }
rescue ActiveRecord::RecordInvalid
  render :show
end
```

`Creator` raises on failure, so the `rescue` still keeps the touch from firing when the school isn't created.

Worth noting for the model-callback suggestion on the other thread: `after_save` would reintroduce this exact problem, since it fires inside `with_lock`. `after_commit on: %i[create destroy]` avoids it and is the more natural hook for an invalidation touch anyway.

app/controllers/concerns/touch_no_school_courses.rb
inulty-dfe: TIL
```ruby
.where.missing(:schools)
```

inulty-dfe: Yeah I think this is the right call :rocket:

app/models/concerns/touch_no_school_courses.rb
```diff
@@ -0,0 +1,38 @@
...
+#
+# Shaped like TouchProvider: the write itself is the trigger, so every writer
+# gets this for free. Bulk writers that would otherwise repeat the sweep once
+# per row wrap their loop in TouchSuppression.suppress and call
+# `.touch_no_school_courses_for(provider)` once at the end instead.
+module TouchNoSchoolCourses
+  extend ActiveSupport::Concern
+
+  included do
+    after_save :touch_no_school_courses
```
tomas-stefano, line 16: `after_save` reintroduces the lock problem.

ProviderSchools::Creator#call (creator.rb:20-28) does the save inside `@provider.with_lock`, so the sweep runs while the provider row is held `FOR UPDATE`. 

Removing the call from  publish/.../checks_controller.rb didn't remove the problem — it pushed it one level down an widened it from one controller to all four add paths. Destroy is the same shape: Removal#call (removal.rb:14-26) wraps provider_school.destroy! in a transaction plus `with_lock`, and TouchProvider's `after_destroy` takes an exclusive lock on the provider row in that same  transaction.

`after_commit on: %i[create update destroy]` is the fix — it fires outside the lock, and suppression still works because the commit happens inside the TouchSuppression.suppress block in the CheckMultiple controllers.

inulty-dfe: I came to the same conclusion

tomas-stefano (approved):

PR #6297, opened 2026-08-07
Upgraded Yarn v1.x to v4.x

dcyoung-dev:
One Login Service banner needed special treatment to get the v4.x from GitHub as it's not available on NPM. This involved manually naming the package, which is not allowed to have `/` characters. `govuk-one-login/service-header` changed to `govuk-one-login-service-header`

## What to do
```bash
  asdf reshim
  corepack enable                  # creates the yarn shim in node 24.13.0's bin dir
  rm -rf node_modules
  yarn install                     # corepack fetches yarn 4.18.0 per packageManager
```

## Checklist


package.json
```diff
@@ -57,7 +58,7 @@
     "stylelint-config-standard": "^39.0.0",
     "stylelint-config-standard-scss": "^15.0.1",
     "stylelint-selector-bem-pattern": "^4.0.1",
-    "playwright": "^1.52.0",
     "vitest": "^4.1.3"
```
dcyoung-dev, line 61: `yarn install` switched the order of these dependancies 🤷‍♂️

inulty-dfe:
Thanks David, could you add these instructions to make it easy for the devs to migrate? Correct me if this is not accurate:


## What to do

```
  corepack enable                  # creates the yarn shim in node 24.13.0's bin dir
  rm -rf node_modules
  yarn install                     # corepack fetches yarn 4.18.0 per packageManager
```

dcyoung-dev:
> Thanks David, could you add these instructions to make it easy for the devs to migrate? Correct me if this is not accurate:
> 
> ## What to do
> ```
>   corepack enable                  # creates the yarn shim in node 24.13.0's bin dir
>   rm -rf node_modules
>   yarn install                     # corepack fetches yarn 4.18.0 per packageManager
> ```

@inulty-dfe great suggestion - where would be best to put this information?

inulty-dfe:
> > Thanks David, could you add these instructions to make it easy for the devs to migrate? Correct me if this is not accurate:
> > ## What to do
> > ```
> >   corepack enable                  # creates the yarn shim in node 24.13.0's bin dir
> >   rm -rf node_modules
> >   yarn install                     # corepack fetches yarn 4.18.0 per packageManager
> > ```
> 
> @inulty-dfe great suggestion - where would be best to put this information?

In the PR description. But we can share it in the team chat when it's merged too

.yarnrc.yml
```diff
@@ -0,0 +1,11 @@
...
+  - "https://github.com/govuk-one-login/service-header"
+  - "https://github.com/govuk-one-login/service-header.git"
+  - "https://github.com/DFE-Digital/dfe-autocomplete"
+  - "https://github.com/DFE-Digital/dfe-autocomplete.git"
+
+enableScripts: true
+
+nodeLinker: node-modules
+
+npmMinimalAgeGate: 0
```
tomas-stefano: Yarn 4.18's default is 1440 (24h). Should it be 0?

package.json
```diff
@@ -57,7 +58,7 @@
...
-  }
+  },
+  "packageManager": "yarn@4.18.0"
```
tomas-stefano, line 64: packageManager is honoured only by Corepack. Maybe is fine.

Should we add a engine guard?

`  "engines": { "yarn": ">=4" }`

.github/workflows/build-and-deploy.yml
```diff
@@ -189,8 +195,11 @@ jobs:
       - uses: actions/setup-node@v7
         with:
           node-version: '24.x'
-          cache: yarn
-      - run: yarn install --frozen-lockfile
+      - name: Enable Corepack (Yarn 4)
+        run: |
+          corepack enable
+          corepack prepare yarn@4.18.0 --activate
+      - run: yarn install --immutable
```
tomas-stefano, line 214: Should we have a cache? cache: yarn was removed from all three setup-node.

Three of the five lint matrix entries (rubocop, erb_lint, brakeman) don't touch node at all and still pay for the install.

.yarnrc.yml
samjohnson-dev: This was more to be dev friendly but happy to chnage to 1440 as default 🤝

.github/workflows/build-and-deploy.yml
samjohnson-dev: Good point, i had to remove at first it gave a more reliable build but can add back in

package.json
samjohnson-dev: Added this back in 🫡

tomas-stefano (approved):

PR #6298, opened 2026-08-07
Remove provider and course schools identity services

inulty-dfe:
## Context

Both identity services were migration scaffolding for the move from Site to
the Provider::School model, and have outlived it.

A bundled bug fix:

We decided to maintain consistent school writing behaviour across this cycle and next. We updated the place where schools are added, but missed the place where schools are deleted.

Remove the now useless `ProviderSchoool::Identity` and `CourseSchoool::Identity`.
Fix dual writing when provider schools are removed. 

## Checklist


inulty-dfe:
[School data - Remove Provider School Identity Service](https://trello.com/c/ehf6JlMe/1877-school-data-remove-provider-school-identity-service)

app/services/provider_schools/removal.rb
```diff
@@ -53,13 +42,8 @@ def destroy_records_if_removable!
       true
     end
 
-    def destroy_site_if_removable!
-      site.destroy!
-      true
```
tomas-stefano, line 58: Unless I understood wrong but dropping destroy_site_if_removable! means a Site
  with no matching `Provider::School` row is now permanently undeletable through the UI (find_by! → RecordNotFound → 404). 

By the way ignore this if all provider school matches a site.

tomas-stefano (approved):
Two questions. One on the PR and another on teams. Looks good to me

app/services/provider_schools/removal.rb
inulty-dfe: I think at this point we need to trust that we have the backfill and dual writing working.

PR #6303, opened 2026-08-10
Course list keep accredited provider headings

tomas-stefano:
# Keep accredited provider headings when filters empty a group

Trello: [1691 — add filters and count](https://trello.com/c/Aot8AB8q/1691-course-list-add-filters-and-count-for-number-of-courses-displayed) · [1756 — logic for when to display filters](https://trello.com/c/69eknLzg/1756-course-list-add-logic-for-when-to-display-filters)

## Why

`Publish::CourseList#groups` chunked the **filtered** query rows, so a group whose courses were all filtered out simply disappeared. For a provider with several accredited providers this is disorienting: the heading vanishes and the list jumps to the next AP, so the provider can't tell whether that AP has no matching courses or has stopped existing.

The group now keeps its place and reads **0 courses**.

## Try it

```
git checkout td/course-list-keep-accredited-provider-headings
bin/dev            # serves on http://localhost:3001
```

All URLs below are verified against the sanitised 2026 data in the dev database. **3F4** is the provider from the card — 100 courses across four accredited providers, listed 13 / 20 / 17 / 50 with no filters applied. Counts below follow that same order.

| # | Scenario | URL | What you should see |
| --- | --- | --- | --- |
| 1 | Baseline, no filters | `/publish/organisations/3F4/2026/courses` | Four AP groups — 13 / 20 / 17 / 50 |
| 2 | **The bug on the card** — filter by QTS only | `/publish/organisations/3F4/2026/courses?qualification[]=qts` | **0** / 15 / **0** / 29 — the first and third headings stay in place at 0 courses |
| 3 | The designer's mock, exactly | `/publish/organisations/3F4/2026/courses?status[]=rolled_over` | **0** / 20 / **0** / 3 |
| 4 | Emptied group between two populated ones | `/publish/organisations/3F4/2026/courses?status[]=withdrawn` | 3 / **0** / **0** / 1 — order unchanged |
| 5 | Two filters combined, one group left | `/publish/organisations/3F4/2026/courses?qualification[]=qts&study_mode[]=part_time` | **0** / **0** / **0** / 1 |
| 6 | Nothing matches anywhere | `/publish/organisations/3F4/2026/courses?level[]=further_education` | A single "No courses found" — no wall of zeros |
| 7 | The provider's own (headingless) group empties | `/publish/organisations/P83/2026/courses?status[]=closed` | The provider's own group holds its place at **0 courses**, its one AP group shows 32 |

Scenarios 6 and 7 are the two judgement calls worth a look:

- **6** — when *every* group is emptied we keep the existing single "No courses found" rather than listing every AP at zero.
- **7** — a provider's own courses render without a heading, and that group still holds its place at 0 so the list doesn't shift under you. Only one provider in the 2026 data is in that state, hence the separate example.

### DIFF
--- a/app/components/publish/courses/list_component.html.erb
+++ b/app/components/publish/courses/list_component.html.erb
@@ -12,11 +12,14 @@
         <%= t("publish.courses.index.course_count", count: group.courses.size) %>
       </p>
 
-      <%= render Publish::Courses::TableComponent.new(
-            courses: group.courses,
-            provider:,
-            course_information_fields: course_list.visible_course_information_fields,
-          ) %>
+      <%# A group the filters emptied keeps its heading and count, but has no table to show. %>
+      <% if group.courses.any? %>
+        <%= render Publish::Courses::TableComponent.new(
+              courses: group.courses,
+              provider:,
+              course_information_fields: course_list.visible_course_information_fields,
+            ) %>
+      <% end %>
     </section>
   <% end %>
 <% else %>
--- a/app/services/publish/course_list.rb
+++ b/app/services/publish/course_list.rb
@@ -44,7 +44,11 @@ module Publish
       start_date: ->(course) { course.start_date.presence&.to_date&.beginning_of_month },
     }.freeze
 
-    delegate :any?, to: :groups
+    # Asks whether any course survived the filters rather than whether any group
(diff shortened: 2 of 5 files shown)


langesi (approved):
LGTM, well done.

Also checked the review app

PR #6304, opened 2026-08-10
Stop rollover duplicating provider sites

tomas-stefano:
## Problem

Study sites (and school sites) appeared three times in the new recruitment cycle.

Rolling over a single course runs the whole provider copy pipeline, so a provider with three draft courses gets rolled over three times. The legacy `Site` copiers had no "already copied?" check, so every site was duplicated on each run. Nothing caught it: `Sites::CopyToProviderService` saves with `validate: false`, bypassing `Site`'s uniqueness validations, and `site` has no unique index. `Provider::School` was unaffected — it has one.

## Solution

New `Sites::ExistingSiteIndex` answers "does the target provider already have this site?", using `Site`'s own uniqueness rules: school sites by URN (site code for URN-less main sites), study sites by location name and URN. The legacy provider copier and the study site copy skip what it matches — study sites before `StudySiteCodeOrchestrator` sees them, so it doesn't burn codes on sites we skip. `Sites::CopyToCourseService` now uses `find_or_create_by`.

Rollover reporting gains `sites_already_present` / `study_sites_already_present`, separate from `*_skipped`, which means "failed".

### DIFF
--- a/app/services/sites/existing_site_index.rb
+++ b/app/services/sites/existing_site_index.rb
@@ -7,12 +7,16 @@ module Sites
   #
   # Identity follows those same rules:
   #
-  # * school sites are identified by URN, which Site validates as unique per
-  #   provider, falling back to the site code for the URN-less sites the URN
-  #   presence validation exempts — in practice, main sites
+  # * school sites are identified by URN and code together. URN alone is not
+  #   identity: a provider can hold one URN twice, on a main site and on an
+  #   ordinary site, since main sites are matched to GIAS by postcode. A copy
+  #   keeps its source's code, so the pair is stable across runs, and the URN-less
+  #   sites the presence validation exempts simply key on their code alone
   # * study sites are identified by location name, which Site validates as unique
   #   per provider and site type, and by URN when they have one — the two checks
-  #   Publish::StudySiteForm makes before letting anyone add a study site
+  #   Publish::StudySiteForm makes before letting anyone add a study site. Code is
+  #   deliberately absent here: StudySiteCodeOrchestrator reassigns study site
+  #   codes as it copies them, so a code is not stable across runs
   #
   # The index is a snapshot of what the target provider had before the copy
   # started, and is deliberately never updated with the sites being copied.
@@ -35,10 +39,8 @@ module Sites
           [:study_site, :location_name, normalize(site.location_name)],
           ([:study_site, :urn, normalize(site.urn)] if site.urn.present?),
         ].compact
-      elsif site.urn.present?
-        [[:school, :urn, normalize(site.urn)]]
       else
-        [[:school, :code, site.code]]
+        [[:school, normalize(site.urn), site.code]]
       end
     end
 
--- a/app/services/sites/existing_site_index.rb
+++ b/app/services/sites/existing_site_index.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module Sites
+  # Answers "does the target provider already have this site?" for the rollover
+  # copiers, which save with `validate: false` and so bypass the uniqueness rules
+  # Site would otherwise enforce.
+  #
+  # Identity follows those same rules:
+  #
+  # * school sites are identified by URN, which Site validates as unique per
+  #   provider, falling back to the site code for the URN-less sites the URN
+  #   presence validation exempts — in practice, main sites
+  # * study sites are identified by location name, which Site validates as unique
+  #   per provider and site type, and by URN when they have one — the two checks
+  #   Publish::StudySiteForm makes before letting anyone add a study site
+  #
+  # The index is a snapshot of what the target provider had before the copy
+  # started, and is deliberately never updated with the sites being copied.
+  # Rollover::Schools::LegacyCourseCopier and Courses::CopyToProviderService both
+  # resolve a source site to its copy by code, so folding fresh copies in would
+  # collapse two same-named source sites onto one copy and silently drop the
+  # second one's course placement.
+  class ExistingSiteIndex
+    def self.for(provider:, site_type:)
+      sites = Site.kept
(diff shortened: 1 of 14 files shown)


spec/system/publish/courses/rolling_over_draft_courses_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+require "rails_helper"
+
+# Rolling a course over runs the whole provider pipeline, sites included, so a
+# provider with several draft courses gets rolled over once per course. The
+# reported bug was a provider whose study sites appeared three times in the new
+# cycle after its three courses were rolled over one at a time.
+RSpec.describe "Rolling over draft courses one at a time" do
+  around do |example|
+    Timecop.travel(Find::CycleTimetable.apply_deadline(2025) - 1.hour) do
```
inulty-dfe: Why is this locked to 2025 apply deadline?

inulty-dfe (commented):
Did you find out what change caused this bug to appear? I wonder if reverting the change and adding regression tests might be a simpler than this implementation?

app/services/sites/existing_site_index.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def self.keys_for(site)
+      if site.study_site?
+        [
+          [:study_site, :location_name, normalize(site.location_name)],
+          ([:study_site, :urn, normalize(site.urn)] if site.urn.present?),
+        ].compact
+      elsif site.urn.present?
+        [[:school, :urn, normalize(site.urn)]]
```
langesi: You can't just use 

:urn, as a main site might also have the same urn as a normal site entry.

So you'll need to combine urn + site code to find the right entry

app/services/sites/existing_site_index.rb
```diff
@@ -0,0 +1,61 @@
...
+      else
+        [[:school, :code, site.code]]
```
langesi: site codes are not unique, have a look at best practice network, however we should not to this part of the if statement anyways, as all active sites have URN entries in our db

langesi (commented):
Good work, need you to revisit the identification of already existing sites though

spec/system/publish/courses/rolling_over_draft_courses_spec.rb
tomas-stefano: Fair question — nothing in this spec turns on the apply deadline. I copied the block from viewing_a_course_spec.rb, which is one of only two files left  using a hand-rolled Timecop `around`.

  Changed in 553ba9aa7 to the repo's `travel:` metadata at a neutral moment:

```ruby
RSpec.describe "Rolling over draft courses one at a time", travel: mid_cycle(2025) do
```

tomas-stefano:
> Did you find out what change caused this bug to appear? I wonder if reverting the change and adding regression tests might be a simpler than this implementation?

Yes — three commits on 2025-08-21:

    08599fbd7  Added StudySiteCodeOrchestrator, and narrowed the guard in Sites::CopyToProviderService from `sites + study_sites` to `new_provider.sites.exists?(code:)`
    cb14694a9  Made the orchestrator return original codes when the source provider's own study site codes are internally unique
    9c7c374cb  Deleted the guard entirely, along with its regression spec (18 lines from spec/services/sites/copy_to_provider_service_spec.rb)

  The guard itself dated back to 2019 (137313bf4). I did look at reverting first, but it isn't available as a shortcut:

  1. Reverting 9c7c374cb alone is a no-op. The orchestrator from 08599fbd7  hands out a code that is guaranteed not to be in use, so  `exists?(code: assigned_code)` is false by construction.

  2. Provider#sites is scoped `where(site_type: :school).kept`, so that guard  never saw study sites at all — it could not have prevented the study site  duplication in the ticket.

  3. Reverting the whole sequence restores the pre-2019 guard that read school  and study site codes as one namespace. That reintroduces the data loss it was replaced for: a study site whose code collided with a school site's   code was silently dropped and never rolled over. 08599fbd7's commit  message is exactly that bug ("study site codes are not uniq from each  provider making rollover doesn't copy all study sites").

  I've added that case as a test in rollover_provider_service_idempotency_spec.rb  — a study site sharing a school site's code must be copied, and copied only  once. Both halves have to hold, and only one of them can be expressed with a
  code-based guard.

  Underneath all this: `course_codes` narrows only `copy_courses`.  `copy_schools` and `copy_study_sites` always walk the provider's whole site  list, so a per-course rollover re-copies everything each time. Gating site  copying on "the provider didn't already exist in the target cycle" would be  simpler, but it's wrong — a batch rollover retry still has to pick up sites  added since the last run.

app/services/sites/existing_site_index.rb
tomas-stefano: Agreed, changed in 553ba9aa7 — school sites now key on a single composite [:school, normalize(urn), code]

app/services/sites/existing_site_index.rb
tomas-stefano: Gone — it folds into the composite key rather than being a branch of its own. A URN-less site keys on ["", code], which is what the fallback was doing, but it can no longer be confused with a site that has a URN.

  It also fixes an inconsistency you may not have seen: `normalize` collapses nil, "" and " " alike, so the empty-string main-site URNs in the data now take the same path as the nil ones. Previously `site.urn.present?` sent them down the code-only branch.

  Study sites I've left on location name + URN. Their codes are reassigned by StudySiteCodeOrchestrator as they're copied, so a code isn't stable across runs and can't be part of their key — and that pair is exactly what
  Publish::StudySiteForm validates before anyone can add one. I've written that reasoning into the class comment so the asymmetry with school sites doesn't read as an oversight

langesi (approved):

PR #6306, opened 2026-08-10
Stop the length controller spec depending on a random length

tomas-stefano:
`spec/controllers/publish/courses/length_controller_spec.rb` failed on CI without anyone touching it. 

https://github.com/DFE-Digital/publish-teacher-training/actions/runs/31395512610/job/93477907109?pr=6276

The course enrichment factory picks `course_length` at random from a list that includes `"TwoYears"`, which is the value these examples submit. On the runs where it came up, the published course already had the length being submitted, so `expect(...course_length).not_to eq("TwoYears")` failed even though the controller had behaved correctly and persisted nothing. The other two assertions in that example, that the interstitial rendered, passed.

The starting length is now pinned to `"OneYear"` and the example asserts the draft still holds it, rather than only that it isn't the submitted value. The draft context is pinned for the same reason, where the coincidence had the opposite effect: the example asserting the change was saved would have passed without anything being saved. The factory is untouched, since the randomness is used across the suite and it is the spec that should pin what it depends on. Verified by forcing the seed to reproduce CI's failure exactly, then running the file 30 times with random seeds, zero failures.

### DIFF
--- a/spec/controllers/publish/courses/length_controller_spec.rb
+++ b/spec/controllers/publish/courses/length_controller_spec.rb
@@ -6,7 +6,21 @@ RSpec.describe Publish::Courses::LengthController, type: :controller do
   render_views
   let(:user) { create(:user, :with_provider) }
   let(:provider) { user.providers.first }
-  let(:course) { create(:course, :published, provider:) }
+  # The course starts on a length that is pinned, and deliberately not the one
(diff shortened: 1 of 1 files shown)


gms-gs (approved):
LGTM!

inulty-dfe (approved):

PR #6307, opened 2026-08-10
Only copy placement schools the target provider has

inulty-dfe:
## Context

Ticking "Copy placement schools?" in the support copy-courses flow has
done nothing since e47dbfba9 (Dec 2024). That commit bound the form to
`CopyCoursesForm`, which namespaced every builder field under
support_copy_courses_form. It moved the provider code field back to the
top level but missed the checkbox, so `params[:schools]` has been nil ever
since and the copy was always skipped.

Read the checkbox under the form's param key so it works again.

That on its own would have been unsafe. The copying behind the dead gate
was rewritten in 33f1713b2, months after it had stopped running, to
create any placement school the target provider was missing, which
contradicts the form's own hint that unlinked schools are not copied.
Restoring the gate alone would have made that live on the first ticked
copy.

CourseSchools::CopyToCourse replaces it, narrowing a course's placements
to schools the target provider already has and pairing each with the
target's own legacy Site by their shared uuid. An unlinked school is
dropped and nothing is created for the target.

The request specs cannot catch the param mismatch, since they choose the
params themselves. The system spec now ticks the real checkbox and
follows a copied course to its Basic details tab to assert the shared
school is listed.

1. Fix the form param so it's read in the controller
2. Change the class that copies the course so it doesn't create provider_schools

## Checklist

### DIFF
--- a/app/controllers/support/providers/copy_courses_controller.rb
+++ b/app/controllers/support/providers/copy_courses_controller.rb
@@ -15,8 +15,6 @@ module Support
         @copy_courses_form = CopyCoursesForm.new(@target_provider, recruitment_cycle.providers.find_by(provider_code: params[:course][:autocompleted_provider_code]))
 
         if @copy_courses_form.valid?
-          copy_provider_schools_to_target if copy_schools?
-
           copier = ::Courses::CopyToProviderService.new(
             schools_copy_to_course: schools_copy_to_course_service,
             sites_copy_to_course: Sites::CopyToCourseService,
@@ -44,23 +42,15 @@ module Support
     private
 
       def copy_schools?
-        params[:schools].present?
+        params.dig(CopyCoursesForm.model_name.param_key, :schools).present?
       end
 
-      def copy_provider_schools_to_target
-        Rollover::Schools::ProviderCopier.new.execute(
-          provider: @copy_courses_form.provider,
-          new_provider: @copy_courses_form.target_provider,
-        )
-      end
-
+      # Placement schools are only linked where the target provider already has
+      # the school; this flow never creates schools for the target provider.
       def schools_copy_to_course_service
         return ->(**) {} unless copy_schools?
 
-        Rollover::Schools::DualCourseCopier.new(
-          legacy_copier: Rollover::Schools::LegacyCourseCopier.new(site_copier: Sites::CopyToCourseService),
-          new_copier: Rollover::Schools::CourseCopier.new,
-        )
+        CourseSchools::CopyToCourse.new
       end
     end
   end
--- a/app/services/course_schools/copy_to_course.rb
+++ b/app/services/course_schools/copy_to_course.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# Copies a course's placement schools onto a copy of that course under another
+# provider.
+#
+# Only copy course schools to the new provider if the new provider has the
+# provider schools
+module CourseSchools
+  class CopyToCourse
+    def call(course:, new_provider:, new_course:)
+      schools = schools_to_copy(course, new_provider)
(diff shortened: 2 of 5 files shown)


inulty-dfe:
[Bug: Support Copy courses to provider is broken](https://trello.com/c/XZ9EAhaF/1888-bug-support-copy-courses-to-provider-is-broken)

app/controllers/support/providers/copy_courses_controller.rb
```diff
@@ -44,23 +42,15 @@ def create
...
       def copy_schools?
-        params[:schools].present?
-      end
-
-      def copy_provider_schools_to_target
-        Rollover::Schools::ProviderCopier.new.execute(
-          provider: @copy_courses_form.provider,
-          new_provider: @copy_courses_form.target_provider,
-        )
+        params.dig(CopyCoursesForm.model_name.param_key, :schools).present?
```
gms-gs, line 45: I may be interpreting this wrong but if govuk_check_box submits the unchecked value as "false". Since "false".present? is true, schools will still be copied when the checkbox is unticked.

Could this use ActiveModel::Type::Boolean.new.cast(...)? The unticked test should submit "schools" => "false" or exercise the real form.

app/services/course_schools/copy_to_course.rb
```diff
@@ -0,0 +1,35 @@
...
+      sites = paired_sites(schools, new_provider)
+
+      schools.each do |provider_school|
+        new_course.schools.create!(
+          provider_school:,
+          gias_school_id: provider_school.gias_school_id,
+        )
+
+        # TODO: School data remodel removal - delete when courses no longer dual-write to SiteStatus.
+        Sites::CopyToCourseService.call(new_site: sites.fetch(provider_school.uuid), new_course:)
```
gms-gs, line 21: Provider::School can exist without a paired legacy Site, particularly after the school remodel. 

sites.fetch(provider_school.uuid) could raise a KeyError (I think)  and roll back the copy. Maybe conditionally create the legacy SiteStatus, or explicitly handle and test this case?

app/controllers/support/providers/copy_courses_controller.rb
inulty-dfe: Fair shout. The value comes from a checkbox though and rails doesn't return the param as `false`, it doesn't appear in the params object

app/services/course_schools/copy_to_course.rb
inulty-dfe: We will not be running this feature in the short time between rollover and backfilling so its not something I'm worried about. What does anyone else feel?

gms-gs (approved):

app/controllers/support/providers/copy_courses_controller.rb
inulty-dfe: When the checkbox is ticked

```ruby
{"authenticity_token" => "e91_KfNtsSwoontdUm02OfmO2dKVTuGB78nDJdT-S_eLIVWKjI-4g98kM2_yA9vemmf9RG3RpZ0PQ60gcBra3Q",
 "support_copy_courses_form" => {"provider" => "ARK Teacher Training (1CS)", "schools" => "true"},
...
```

When it's not ticked

```ruby
{"authenticity_token" => "e91_KfNtsSwoontdUm02OfmO2dKVTuGB78nDJdT-S_eLIVWKjI-4g98kM2_yA9vemmf9RG3RpZ0PQ60gcBra3Q",
 "support_copy_courses_form" => {"provider" => "ARK Teacher Training (1CS)"},
 "course" => {"autocompleted_provider_code" => "1CS"},
...
```

langesi (approved):
Great work!

PR #6308, opened 2026-08-11
Fix sites invalid rolled over cycle

tomas-stefano:
## Make study site optional when adding a course

Adding a course forced providers to pick a study site. It's optional everywhere else — no publish validation, and the edit form lets you clear every one off a live course.

- Drops the "Select at least one study site" guard in `Courses::CreationService` (covers both add-course flows)
- Removes the plumbing that only existed to report it: the legacy step's `error_keys`, error wrappers on both study-site views, two dead locale messages, and an orphan method with no callers
- Details page, legacy confirmation page and wizard check-answers now present it like any other optional field: labelled "(optional)", "Not entered" when empty, ordinary Change/Add action instead of the `app-inset-text--important` call to action reserved for things that block publishing

Two deliberate exceptions: a provider with no study sites on their organisation still gets an "Add a study site" prompt (the step is skipped for them, so Change has nowhere to go), and the row stays read-only on a withdrawn course.

### DIFF
--- a/spec/system/publish/courses/editing_study_sites_spec.rb
+++ b/spec/system/publish/courses/editing_study_sites_spec.rb
@@ -32,10 +32,6 @@ RSpec.describe "updating study sites on a course" do
     expect(page).to have_link("Add a study site")
   end
 
-  def then_i_see_the_error_message_add_one_study_site
-    expect(page).to have_link("Add at least one study site")
-  end
-
   def and_i_am_authenticated_as_a_provider_user
     given_i_am_authenticated(
       user: create(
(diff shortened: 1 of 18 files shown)


inulty-dfe:
[Bug: Study site is mandatory when it should be optional](https://trello.com/c/1Jb8CWb2/1883-bug-study-site-is-mandatory-when-it-should-be-optional)

inulty-dfe:
We render the word "None" when there are no placement schools on a course

## Check Answers page

image

## Basic details page

image

inulty-dfe (commented):
I think we're jus missing some text when there are no study sites on the check answers page

spec/system/publish/courses/editing_study_sites_spec.rb
```diff
@@ -34,7 +34,7 @@ def then_i_see_the_success_message
   def then_the_study_site_row_reads_as_optional_and_unanswered
     row = page.find(".govuk-summary-list__row", text: "Study site (optional)")
 
-    expect(row).to have_text("Not entered")
```
inulty-dfe: I think you misunderstood. I was just saying that the edit page uses "Not entered", not to change it from that.

I was talking about the check answers page.

Image

inulty-dfe (changes requested):
I think you misunderstood what my last comment was

spec/system/publish/courses/editing_study_sites_spec.rb
tomas-stefano: Ok. I pushed the fix!

inulty-dfe (approved):

PR #6309, opened 2026-08-11
Remove findable scope in API for new data model

gms-gs:
## Context

The 2027 school data model uses `Course::School` rather than legacy `site_status` records.

`CourseSearchServiceSchools` still applied the legacy `findable` scope when requested by API consumers. This could exclude valid 2027 courses from location-based API results because the scope depends on `site_status`.

- Remove the legacy `findable` scope from `CourseSearchServiceSchools`.
- Remove the now-unused `findable?` helper.
- Confirm `findable: true` no longer applies the legacy scope.
- Add API coverage confirming 2027 location searches return courses using the new school data model.

There are no UI changes.

## Guidance to review

Run:

```shell
bundle exec rspec spec/services/course_search_service_schools_spec.rb spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```

Review that:
- The legacy search service remains unchanged.
- The 2027 API still accepts `findable: true`.
- Location filtering uses attached `Course::School` records.

## Checklist

### DIFF
--- a/spec/controllers/api/public/v1/providers/courses_controller_spec.rb
+++ b/spec/controllers/api/public/v1/providers/courses_controller_spec.rb
@@ -253,6 +253,28 @@ RSpec.describe API::Public::V1::Providers::CoursesController do
           expect(ids).to include(near_course.id.to_s)
           expect(ids).not_to include(far_course.id.to_s)
         end
+
+        context "when the findable filter is requested for 2027" do
+          before do
+            get :index, params: {
+              recruitment_cycle_year: provider.recruitment_cycle.year,
+              provider_code: provider.provider_code,
+              filter: {
+                findable: true,
+                latitude: 51.5074,
+                longitude: -0.1278,
+                radius: 10,
+              },
+            }
+          end
+
+          it "returns courses using schools from the new data model" do
+            ids = json_response["data"].map { |course| course["id"] }
+
+            expect(ids).to include(near_course.id.to_s)
+            expect(ids).not_to include(far_course.id.to_s)
+          end
+        end
       end
     end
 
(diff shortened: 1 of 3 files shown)


spec/controllers/api/public/v1/providers/courses_controller_spec.rb
```diff
@@ -253,6 +253,28 @@
           expect(ids).to include(near_course.id.to_s)
           expect(ids).not_to include(far_course.id.to_s)
         end
+
+        context "when the findable filter is requested for 2027" do
```
inulty-dfe: What makes this test any thing to do with 2027 specifically?

inulty-dfe:
[School Data - Remove findable scope in API for new data model](https://trello.com/c/HpgbY4Tn/1885-school-data-remove-findable-scope-in-api-for-new-data-model)

spec/controllers/api/public/v1/providers/courses_controller_spec.rb
gms-gs: `let(:provider) { create(:provider, recruitment_cycle: find_or_create(:recruitment_cycle, year: Settings.schools_remodel_cycle_year)) }`

But now I read it. Its not 2027. Its 2026 cycle.

inulty-dfe: Ah I see, its set on the provider factory creation.

inulty-dfe: It's "after 2026", so 2027. But I'd avoid putting specific dates in. Let the setting decide what the year is

inulty-dfe (approved):

PR #6312, opened 2026-08-11
School remodel: Sync uuid after rollover automagically

langesi:
## Context

We have to run the backfill after every rollover at the moment and manual rollover is also broken due to the divergence of the uuid field in both models. This will fix it by updating the new provider_schools and changing the uuid

- Create a service that updated the uuid of the provider_school to match the site table
## Checklist

### DIFF
--- a/app/services/provider_schools/sync_legacy_site_uuids.rb
+++ b/app/services/provider_schools/sync_legacy_site_uuids.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module ProviderSchools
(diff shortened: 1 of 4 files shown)


inulty-dfe (commented):
Is it too awkward to pass the uuid from the school creation to the creation of the legacy site?

Also, we should consider suppressing the touch on the Provider and "no-schools" when it comes in.

tomas-stefano:
We need to see how this PR will play out together with #6304

inulty-dfe (approved):
LGTM

PR #6315, opened 2026-08-12
Bug: Apply for this course button in a weird location on preview

tomas-stefano:
## Problem

On the course preview in Publish, the "Apply for this course" button renders below the Course summary, and the "Sign in to save this course" prompt is missing entirely. On the live Find course page Apply sits above the summary with the save prompt beside it.

The preview is meant to show a provider how their course will appear on Find. It duplicates `find/courses/show.html.erb` instead of sharing it, and the two have drifted.

Two things stopped Find's markup being reused as-is:

- `authenticated?` is a `helper_method` on Find controllers only. Publish defines its own, but it answers for the signed-in provider rather than a candidate, so the partial would have rendered a working "Save this course" form posting to Find's candidate routes.
- `.save-course-button__*` and `.unstyled-button` existed only in the Find CSS bundle, so the prompt would have drawn as unstyled body text under the Publish layout.

## Solution

- Move the top Apply render above `SummaryComponent`, into the same grid row Find uses, with the save prompt beside it.
- Set `@apply_action_column_class` in `Publish::CoursesController#preview`, mirroring Find's rule, so the two columns size the same way.
- Give `_save_button` a defaulted `preview` local and branch on it first, so `authenticated?` is never reached from Publish. In preview the prompt renders inert — same icon and label, no link or button — because the provider is not a candidate.
- Extract `_save_button_content` so the icon and label exist once rather than once per branch.
- Move `_save-course-button.scss` into the `stylesheets/components/` folder shared by both apps.


tomas-stefano:
https://trello.com/c/cjoi8Si6/1881-bug-apply-for-this-course-button-in-a-weird-location-on-preview

app/views/find/courses/_save_button.html.erb
```diff
@@ -1,33 +1,22 @@
-<% if authenticated? %>
+<% preview = local_assigns.fetch(:preview, false) %>
```
gms-gs: Can this be handled in the controller instead of doing variable assignment in the view?

tomas-stefano: Done! I used `preview?` helper

gms-gs (approved):

PR #6316, opened 2026-08-12
Update SchoolPresenceExemption to allow fee paying courses

inulty-dfe:
## Context

We permit salary and apprenticeship courses to publish without course schools attached, if the course has been exempted by Support. 

We are now going to change it so that fee paying courses are allowed to behave the same way. This means that all funding types can be exempt from attaching course schools in future.

- Allow Support users to exempt any course from requiring schools attached
- Relax the validation for exempt courses to permit any funding type
- Fix a flakey test regarding GCSE Grade

## Checklist

### DIFF
--- a/app/views/support/courses/edit.html.erb
+++ b/app/views/support/courses/edit.html.erb
@@ -52,13 +52,11 @@
 
       <% end %>
 
-      <% if @course.decorate.salaried? %>
-        <%= f.govuk_check_boxes_fieldset :publish_without_schools_allowed,
-              legend: { text: "Publishing this course without schools" },
-              hint: { text: "Only tick this box if you have confirmed that this provider is allowed to publish courses without schools attached." },
-              multiple: false do %>
-          <% f.govuk_check_box :publish_without_schools_allowed, "true", "false", label: { text: "Allow this course to be published without schools attached" }, multiple: false %>
-        <% end %>
+      <%= f.govuk_check_boxes_fieldset :publish_without_schools_allowed,
+            legend: { text: "Publishing this course without schools" },
+            hint: { text: "Only tick this box if you have confirmed that this provider is allowed to publish courses without schools attached." },
+            multiple: false do %>
+        <% f.govuk_check_box :publish_without_schools_allowed, "true", "false", label: { text: "Allow this course to be published without schools attached" }, multiple: false %>
       <% end %>
 
       <%= f.govuk_submit t("support.update_record") %>
--- a/spec/factories/providers.rb
+++ b/spec/factories/providers.rb
@@ -5,7 +5,9 @@ FactoryBot.define do
     provider_name { "ACME SCITT#{rand(1_000_000)}" }
 
     sequence(:provider_code) do |n|
-      ProviderCodeGenerator.new(n).call
+      # Avoid generating "I30" code by chance This code has specific
+      # and unique GCSE grade requirements that can cause flaky tests
+      ProviderCodeGenerator.new(n).call(avoid: %w[I30])
     end
 
     trait :with_anonymised_data do
@@ -81,6 +83,10 @@ FactoryBot.define do
 
     trait :previous_recruitment_cycle do
       recruitment_cycle { find_or_create :recruitment_cycle, :previous }
+    end
+
+    trait :gcse_grade_5 do
+      provider_code { "I30" }
     end
 
     trait :published_scitt do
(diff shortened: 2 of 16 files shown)


inulty-dfe:
[Expand not attaching schools to courses out to fee funded courses](https://trello.com/c/u9h5WUgE/1813-expand-not-attaching-schools-to-courses-out-to-fee-funded-courses)

spec/factories/providers.rb
```diff
@@ -5,7 +5,9 @@
     provider_name { "ACME SCITT#{rand(1_000_000)}" }
 
     sequence(:provider_code) do |n|
-      ProviderCodeGenerator.new(n).call
+      # Avoid generating "I30" code by chance This code has specific
+      # and unique GCSE grade requirements that can cause flaky tests
+      ProviderCodeGenerator.new(n).call(avoid: %w[I30])
```
tomas-stefano, line 10: There is a `avoid:` vs `excluded_codes` param name.

spec/support/provider_code_generator.rb
```diff
@@ -4,16 +4,18 @@ class ProviderCodeGenerator
   def initialize(sequence_number)
     @sequence_number = sequence_number
     @existing_codes = Provider.pluck(:provider_code).to_set
+    @unwanted_codes = @existing_codes
   end
 
-  def call
+  def call(excluded_codes = [])
```
tomas-stefano: The method takes a positional arg, but `spec/factories/providers.rb:10` calls
  .call(avoid: %w[I30]). 

The sequence can still emit I30, so gcse_equivalency_spec stays as flaky as before and the new :gcse_grade_5 trait can now collide with a generated code. 

I'd make it def call(avoid: []).

app/views/support/courses/edit.html.erb
```diff
@@ -52,13 +52,11 @@
...
-              legend: { text: "Publishing this course without schools" },
-              hint: { text: "Only tick this box if you have confirmed that this provider is allowed to publish courses without schools attached." },
-              multiple: false do %>
-          <% f.govuk_check_box :publish_without_schools_allowed, "true", "false", label: { text: "Allow this course to be published without schools attached" }, multiple: false %>
-        <% end %>
+      <%= f.govuk_check_boxes_fieldset :publish_without_schools_allowed,
+            legend: { text: "Publishing this course without schools" },
+            hint: { text: "Only tick this box if you have confirmed that this provider is allowed to publish courses without schools attached." },
+            multiple: false do %>
+        <% f.govuk_check_box :publish_without_schools_allowed, true, false, label: { text: "Allow this course to be published without schools attached" }, multiple: false %>
```
tomas-stefano: On update failure the controller re-renders :edit with the form populated from params, so
  publish_without_schools_allowed is the String "true". 

ActionView::Helpers::Tags::CheckBox#checked?
  takes the when String branch and compares "true" == true → false, so the box renders unticked after any validation error (e.g. a bad start date) and the support user silently loses the change on resubmit. 

The three sibling checkboxes at lines 38/44/50 all still pass "true", "false". No spec covers the error re-render path. 

Can you add a spec to cover this?

tomas-stefano:
app/services/courses/publish_rules/school_presence_exemption.rb:3-9 — comment now contradicts the code

  The header still reads "Support can approve individual salaried or apprenticeship courses…" and "Never applies to fee-paying courses" on the file that's declared the single source of truth for this rule. course.rb got its comment updated; this one didn't.

spec/factories/providers.rb
inulty-dfe: Well that's embarrassing

spec/system/support/providers/courses/editing_publish_without_schools_spec.rb
```diff
@@ -29,7 +29,7 @@
   context "with a fee course" do
     scenario "the publish without schools checkbox is not available" do
       when_i_visit_the_edit_page_for(fee_course)
-      then_i_do_not_see_the_publish_without_schools_checkbox
+      then_i_see_the_publish_without_schools_checkbox
```
tomas-stefano, line 43: The scenario name is inverted

Still scenario "the publish without schools checkbox is not available" while the body now asserts `then_i_see_the_publish_without_schools_checkbox`

gms-gs (approved):

spec/factories/providers.rb
gms-gs: Good spot

PR #6319, opened 2026-08-13
Validate at least one provider school exists before deleting one

inulty-dfe:
## Context

It is possible for the system to reach an invalid state WRT no-school courses' providers schools. We need to make sure that a provider always has at least one provider school once at least one has been added.

#### Vulnerable scenario

A course that has been exempt from attaching schools must have at least one provider school so it can be injected into the API response for course locations. Otherwise dependent systems in the Apply service will fail.

1. New provider is added
2. User adds a provider school
3. User creates a course and attaches the provider school
4. User requests that the course be marked eligible for no schools attached
5. User removes the course school
6. User publishes the course
7. User removes the provider school
8. Apply requests the course and locations via the api and receives no schools

Validate a providers schools on removal to ensure that at least one provider school exists.
If the user has one provider school and wants to remove it, they must add another before removing the other.
Update the `destroy` action flash message to match the reason the school couldn't be removed.

## Guidance to review

- The error messages are baked into the partial
- The flash message corresponds to the reason the school couldn't be removed

## Checklist

### DIFF
--- a/app/controllers/publish/providers/schools_controller.rb
+++ b/app/controllers/publish/providers/schools_controller.rb
@@ -32,11 +32,17 @@ module Publish
           redirect_to publish_provider_recruitment_cycle_schools_path
         else
           redirect_to delete_publish_provider_recruitment_cycle_school_path(@provider.provider_code, school.recruitment_cycle.year, school.uuid),
-                      flash: { warning: t(".cannot_remove_school") }
+                      flash: { warning: cannot_remove_school_message }
         end
       end
 
     private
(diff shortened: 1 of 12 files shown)


inulty-dfe:
[Prevent a no-school course from being published if the Provider has no schools](https://trello.com/c/uWqpj3LE/1896-prevent-a-no-school-course-from-being-published-if-the-provider-has-no-schools)

tomas-stefano (approved):

PR #6324, opened 2026-08-14
Add CSV downloads to the course list page

kmcd:
## Context

Research with providers found they need CSVs of course information and of the
schools attached to their courses, for reporting and auditing.

https://trello.com/c/Q8J3k77z/1841-add-csv-downloads-to-the-course-list-page

Add "Download course information" callout to the Publish course list page with two CSV downloads: 
* basic course information
* schools attached to each course

## Guidance to review

The card descopes hiding columns whose value is the same on every course. After discussion the team preferred a stable column set. This means some providers will see redundant columns.

https://publish-review-6324.test.teacherservices.cloud

**N.B.**

- CSV download ignores the page filters. They export whole list similar to the training partners export.
- CSV headers use the UI wording, not the mockup in the card; e.g. "Fee-paying" not "Fee".
- Age range is included for secondary courses, which the list page hides but is in the mockup.  
-  Discarded schools are exported

https://github.com/user-attachments/assets/81248381-3b5e-4f85-b7ae-29ac543e899a

## Checklist

### DIFF
--- a/app/controllers/publish/courses/exports_controller.rb
+++ b/app/controllers/publish/courses/exports_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Publish
+  module Courses
+    class ExportsController < ApplicationController
+      def course_information
+        send_csv(::Exports::CourseInformationList.new(provider:))
+      end
+
+      def schools
+        send_csv(::Exports::CourseSchoolsList.new(provider:))
+      end
+
+    private
+
+      def send_csv(export)
+        authorize :provider, :index?
+
+        respond_to do |format|
+          format.csv { send_data(export.data, filename: export.filename, disposition: :attachment) }
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/publish/courses/exports_controller.rb
+++ b/app/controllers/publish/courses/exports_controller.rb
@@ -14,8 +14,6 @@ module Publish
     private
 
       def send_csv(export)
-        authorize :provider, :index?
-
         respond_to do |format|
           format.csv { send_data(export.data, filename: export.filename, disposition: :attachment) }
         end
--- a/app/services/exports/course_information_list.rb
+++ b/app/services/exports/course_information_list.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module Exports
+  class CourseInformationList
+    include ActionView::Helpers::NumberHelper
+    include CourseColumns
+
+    CSV_HEADERS = [
+      "Course name",
+      "Course code",
+      "Accredited provider",
+      "Status",
+      "Age range",
+      "Fee or salary",
+      "Qualification",
+      "Study mode",
+      "Start date",
+      "Course length",
+      "UK fee",
+      "Non-UK fee",
+    ].freeze
+
+    def initialize(provider:)
+      @provider = provider
+    end
+
+    def data
+      CSV.generate(headers: CSV_HEADERS, write_headers: true) do |csv|
+        courses.each do |course|
+          enrichment = course.latest_enrichment
+
+          csv << [
+            course.name,
+            course.course_code,
+            accredited_provider(course),
+            status(course),
+            age_range(course),
+            I18n.t("publish.courses.course_table.funding.#{course.funding}"),
+            course.qualifications_summary,
+            course.study_mode_description.capitalize,
+            start_date(course),
+            course_length(enrichment&.course_length),
+            number_to_currency(enrichment&.fee_uk_eu),
+            number_to_currency(enrichment&.fee_international),
+          ]
+        end
+      end
+    end
+
+    def filename
+      "course-information-#{provider.provider_code}-#{Time.zone.today}.csv"
+    end
+
+  private
+
+    attr_reader :provider
+
+    def courses
+      Publish::Courses::Query.call(provider:).preload(:latest_enrichment)
+    end
+
+    def accredited_provider(course)
+      course[:group_name] || provider.provider_name
+    end
+
+    def start_date(course)
+      return if course.start_date.blank?
+
+      I18n.l(course.start_date.to_date, format: :short)
+    end
+
+    def course_length(value)
+      return if value.blank?
+
+      I18n.t("courses.summary_card_component.length_value.#{value}", default: value)
+    end
+  end
+end
--- a/app/services/exports/course_information_list.rb
+++ b/app/services/exports/course_information_list.rb
@@ -29,7 +29,7 @@ module Exports
     def data
       BYTE_ORDER_MARK + CSV.generate(headers: CSV_HEADERS, write_headers: true) do |csv|
         courses.each do |course|
-          enrichment = course.latest_enrichment
+          enrichment = course.latest_non_draft_enrichment
 
           csv << [
             course.name,
@@ -58,7 +58,7 @@ module Exports
     attr_reader :provider
 
     def courses
-      Publish::Courses::Query.call(provider:).preload(:latest_enrichment)
+      Publish::Courses::Query.call(provider:).preload(:latest_non_draft_enrichment)
     end
 
     def accredited_provider(course)
--- a/app/services/exports/course_information_list.rb
+++ b/app/services/exports/course_information_list.rb
@@ -29,7 +29,7 @@ module Exports
     def data
       BYTE_ORDER_MARK + CSV.generate(headers: CSV_HEADERS, write_headers: true) do |csv|
         courses.each do |course|
-          enrichment = course.latest_non_draft_enrichment
+          enrichment = reported_enrichment(course)
 
           csv << [
             course.name,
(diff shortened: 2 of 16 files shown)


app/services/exports/course_schools_list.rb
```diff
@@ -0,0 +1,60 @@
...
+    attr_reader :provider
+
+    def rows
+      courses
+        .flat_map { |course| placement_sites(course).map { |site| [site, course] } }
+        .sort_by { |site, course| [site.location_name.to_s.downcase, course.name.to_s.downcase, course.course_code.to_s] }
+    end
+
+    def placement_sites(course)
+      course.site_statuses
```
tomas-stefano: the export reads the deprecated school model, so names can disagree with the schools page.

I think is best to use `course.schools` instead. 

But happy for @langesi to jump in to confirm it.

app/controllers/publish/courses/exports_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      def schools
+        send_csv(::Exports::CourseSchoolsList.new(provider:))
+      end
+
+    private
+
+      def send_csv(export)
+        authorize :provider, :index?
```
tomas-stefano: authorize :provider, :index? is a no-op

  ProviderPolicy aliases index? to new?, which is just `user.present?`, and the symbol form doesn't pass the provider record in. Every signed-in user passes this check. 

The real protection is the inherited `before_action :authorize_provider` in `Publish::ApplicationController:13` (`authorize  provider, :show?`) — which is what makes the 403 spec pass.

app/services/exports/course_information_list.rb
```diff
@@ -0,0 +1,80 @@
...
+  private
+
+    attr_reader :provider
+
+    def courses
+      Publish::Courses::Query.call(provider:).preload(:latest_enrichment)
+    end
+
+    def accredited_provider(course)
+      course[:group_name] || provider.provider_name
```
tomas-stefano: I think this can name the wrong provider in the "Accredited provider" column.

`group_name` comes from the CASE in `Publish::Courses::Query#status_columns_select`, over a LEFT OUTER JOIN onto `provider` matched on `provider_code` **and** `recruitment_cycle_id`. So it comes back NULL in two different situations:

1. the course really is self-accredited (`accredited_provider_code` is NULL or the provider's own code); and
2. `accredited_provider_code` is set, but the join found no provider row for that code in this cycle, so `accredited_provider.provider_name` is NULL.

`||` can't distinguish them, and in case 2 we fall back to the downloading provider's name.

Case 2 is reachable during rollover: providers are stored per recruitment cycle, so a course rolled into 2026 with `accredited_provider_code = "B01"` has no matching 2026 row until Brighton is rolled over too. The CSV then reports:

```csv
Chemistry,2KGZ,Some SCITT,...
```

when the course is actually ratified by University of Brighton — the opposite of the truth, in an export providers are using for auditing.

The list page shares the NULL but only decides whether to render a group heading, so there it degrades to a missing heading rather than a wrong claim.

Could we decide from the course's own column, which is always populated, rather than from the join result?

```ruby
def accredited_provider(course)
  code = course.accredited_provider_code
  return provider.provider_name if code.blank? || code == provider.provider_code

  course[:group_name] || code
end
```

That falls back to the code (or a blank cell, if you prefer) instead of asserting the wrong provider. Might be worth a spec next to the existing ones in `course_information_list_spec.rb`: a course whose `accredited_provider_code` has no provider row in the cycle, asserting the cell isn't the downloading provider's name.

app/services/exports/course_schools_list.rb
```diff
@@ -0,0 +1,60 @@
...
+      "schools-attached-to-courses-#{provider.provider_code}-#{Time.zone.today}.csv"
+    end
+
+  private
+
+    attr_reader :provider
+
+    def rows
+      courses
+        .flat_map { |course| placement_sites(course).map { |site| [site, course] } }
```
tomas-stefano: courses with no schools disappear from the  export

We have now a very rare cases with courses with no schools.

Let's add a spec to reproduce the issue?

app/services/exports/course_schools_list.rb
kmcd: Good catch! Done ✅

app/controllers/publish/courses/exports_controller.rb
kmcd: Good spot. I went with the existing patterns in the repo. 

This no-op happens in four other places:
- [`courses_controller.rb:10`](https://github.com/DFE-Digital/publish-teacher-training/blob/6dfaa4c668839482e9cac927cd914d396e1a2351/app/controllers/publish/courses_controller.rb#L10)
- [`schools/added_schools_controller.rb:10`](https://github.com/DFE-Digital/publish-teacher-training/blob/6dfaa4c668839482e9cac927cd914d396e1a2351/app/controllers/publish/providers/schools/added_schools_controller.rb#L10)
- [`schools/removed_schools_controller.rb:8`](https://github.com/DFE-Digital/publish-teacher-training/blob/6dfaa4c668839482e9cac927cd914d396e1a2351/app/controllers/publish/providers/schools/removed_schools_controller.rb#L8)
- [`training_partners/courses_controller.rb:8`](https://github.com/DFE-Digital/publish-teacher-training/blob/6dfaa4c668839482e9cac927cd914d396e1a2351/app/controllers/publish/providers/training_partners/courses_controller.rb#L8)

Should I tidy these up here or open a new PR for them?

app/services/exports/course_information_list.rb
kmcd: Thanks for looking into this one. Bit of a mind bender.

Yes, reproduced the defect: export shows downloading provider, e.g. "AA Teamworks West Yorkshire SCITT", when `group_name` is `nil`.

I went with the code fallback rather than blank cell. A blank cell looks like missing data, so they can't tell "self-accredited" from "we couldn't resolve it". 

Every course in the current cycle with an `accredited_provider_code` has a matching provider row. So think this is future proofing subsequent rollovers.

Fixed and added a spec 👍

app/services/exports/course_schools_list.rb
kmcd: > Let's add a spec to reproduce the issue?

Think there's already a spec for this one: [`course_schools_list_spec.rb:104-109`](https://github.com/DFE-Digital/publish-teacher-training/blob/d92d0cb4f2ba77586961967e173ed384b5ea04b5/spec/services/exports/course_schools_list_spec.rb#L104-L109)

The school model change you suggested fixed one of the missing courses I noticed. The only remaining missing course is where there are no schools.

The CSV file is one row per school per course. 

Should a course with no placement schools show as a row with a blank school cell instead? 

I don't think so but could be wrong ... 🤔

app/services/exports/course_schools_list.rb
kmcd: I'm assuming it's OK to have placement schools that no longer exist @tomas-stefano / @langesi?

Closed GIAS schools are in the CSV: 20,989 rows from 3,792 courses (about 5%). 

A school could have closed during the cycle and you'd still want to see it.

app/controllers/publish/courses/exports_controller.rb
tomas-stefano: I think best open a new PR, no worries :)

app/services/exports/course_schools_list.rb
tomas-stefano: I think is best to ask this scenario to @kelliedesigner 

> Should a course with no placement schools show as a row with a blank school cell instead?


I think so but that's my opinion as a dev. Worth confirming it with @kelliedesigner and @langesi

tomas-stefano (approved):
Besides the no school comment, everything looks good to me. After confirming feel free to add/merge

kmcd:
> Besides the no school comment, everything looks good to me. After confirming feel free to add/merge

Great - thanks for the thorough review @tomas-stefano . Very relieved to have another safety net 😄 

I've asked @kelliedesigner to take a look so hopefully we can resolve the no schools question.

app/services/exports/course_information_list.rb
```diff
@@ -0,0 +1,83 @@
...
+    ].freeze
+
+    def initialize(provider:)
+      @provider = provider
+    end
+
+    def data
+      CSV.generate(headers: CSV_HEADERS, write_headers: true) do |csv|
+        courses.each do |course|
+          enrichment = course.latest_enrichment
```
tomas-stefano: small question: if the publish user edit a course and doesn't publish? On the CSV should we show the latest draft? If so this line is correct :).

If only the latest published edit - then it should be `latest_published_enrichment`.

kmcd: Good question @tomas-stefano ... 🤔 

Think you're correct, for example, 196/J534 Art and design:

```
┌────────────┬───────────┬────────┐
│ Enrichment │  Status   │ UK fee │                                                                                                                                                            
├────────────┼───────────┼────────┤
│ older      │ published │ £9,535 │
├────────────┼───────────┼────────┤
│ newer      │ draft     │ £9,790 │
└────────────┴───────────┴────────┘
```

In this case the CSV shows the draft which is incorrect.

However, using `latest_published_enrichment` means we also need to handle withdrawn  courses. Looks like withdrawn courses are not marked as published, so `latest_published_enrichment` would be `nil` for those.

Instead of `latest_published_enrichment`, perhaps "the latest enrichment that isn't a draft" would work? 

Added a commit to fix this here: https://github.com/DFE-Digital/publish-teacher-training/pull/6324/changes/6152dc4c84b6109737de566a268ac6ffd6f6c822

☝️ Does this make sense?

PR #6326, opened 2026-08-14
Bug: Saved course issue on candidate accounts

gms-gs:
## Context

Saved courses from a previous recruitment cycle remain visible in candidate accounts. If the course was not rolled over, its existing link leads to a “Page not found” page.

- Remove the hyperlink from previous-year saved course titles.
- Add a red “Course from a previous year” tag.
- Keep current-cycle course links unchanged.
- Add component and system test coverage.

## Guidance to review

1. Save a course to a candidate account.
2. Move the course’s provider to the previous recruitment cycle.
3. View the candidate’s saved courses.
4. Confirm the course title is not linked and the previous-year tag is displayed.
5. Confirm current-cycle saved courses remain linked.

## Checklist

### DIFF
--- a/app/components/saved_courses/summary_card_component.rb
+++ b/app/components/saved_courses/summary_card_component.rb
@@ -72,7 +72,7 @@ module SavedCourses
     end
 
     def course_from_previous_year?
-      course.recruitment_cycle.year.to_i < Find::CycleTimetable.current_year
+      @course_from_previous_year ||= course.recruitment_cycle.year.to_i < Find::CycleTimetable.current_year
     end
   end
 end
--- a/app/components/saved_courses/summary_card_component.rb
+++ b/app/components/saved_courses/summary_card_component.rb
@@ -17,7 +17,11 @@ module SavedCourses
     def title
       provider_span = content_tag(:span, course.provider_name)
       course_span = content_tag(:span, course.name_and_code)
-      status_tag = course.decorate.saved_status_tag
+      status_tag = if course_from_previous_year?
+                     helpers.govuk_tag(text: t(".course_from_previous_year"), colour: "red")
+                   else
+                     course.decorate.saved_status_tag
+                   end
       status_block = (content_tag(:div, status_tag, class: "app-saved-course__status-tag") if status_tag.present?)
 
       title_inner = safe_join(
@@ -29,11 +33,14 @@ module SavedCourses
         ].compact,
       )
 
-      course_info =
-        govuk_link_to(
-          find_course_path(provider_code: course.provider_code, course_code: course.course_code),
-          class: "govuk-link",
-        ) { title_inner }
+      course_info = if course_from_previous_year?
+                      title_inner
+                    else
+                      govuk_link_to(
+                        find_course_path(provider_code: course.provider_code, course_code: course.course_code),
+                        class: "govuk-link",
+                      ) { title_inner }
+                    end
 
       content_tag(:div, class: "app-saved-course__card-title") do
         safe_join(
@@ -62,6 +69,10 @@ module SavedCourses
       @location.present? &&
         saved_course.respond_to?(:minimum_distance_to_search_location) &&
         saved_course.minimum_distance_to_search_location.present?
+    end
+
+    def course_from_previous_year?
+      course.recruitment_cycle.year.to_i < Find::CycleTimetable.current_year
     end
   end
 end
--- a/app/services/saved_courses/query.rb
+++ b/app/services/saved_courses/query.rb
@@ -44,21 +44,24 @@ module SavedCourses
 
     # Distance annotation over the legacy course_site -> site model, used while
     # the :course_publishing_uses_new_school_model flag is off. Saved courses are
-    # annotated with distance but not filtered by radius.
+    # annotated with distance but not filtered by radius. Previous-cycle courses
+    # are retained when their old placement sites are no longer publishable.
     def sites_location_scope(latitude:, longitude:)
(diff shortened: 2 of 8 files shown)


gms-gs:
[Bug: Saved course issue on candidate accounts](https://trello.com/c/TQToA0UB/1148-bug-saved-course-issue-on-candidate-accounts)

kmcd:
Looking at this one now @gms-gs

app/services/saved_courses/query.rb
```diff
@@ -44,21 +44,24 @@ def location_scope
...
-          INNER JOIN site ON (
+          LEFT JOIN site ON (
             site.id = course_site.site_id
             AND site.discarded_at IS NULL
+            AND site.longitude IS NOT NULL
+            AND site.latitude IS NOT NULL
           )
         SQL
-        .where("(site.longitude IS NOT NULL OR site.latitude IS NOT NULL)")
+        .where("site.id IS NOT NULL OR provider.recruitment_cycle_id <> ?", RecruitmentCycle.current.id)
```
kmcd: Can `RecruitmentCycle.current` ever be `nil` ?

app/components/saved_courses/summary_card_component.rb
```diff
@@ -63,5 +70,9 @@ def search_by_location?
         saved_course.respond_to?(:minimum_distance_to_search_location) &&
         saved_course.minimum_distance_to_search_location.present?
     end
+
+    def course_from_previous_year?
+      course.recruitment_cycle.year.to_i < Find::CycleTimetable.current_year
```
kmcd: Looks like there's an N+1 from recruitment_cycle. One query for each card maybe?

app/services/saved_courses/query.rb
```diff
@@ -140,7 +150,8 @@ def fee_uk_ascending_order_scope
 
       @scope
         .select("saved_course.*, #{funding_sorting}, provider.provider_name, MAX((course_enrichment.json_data->>'FeeUkEu')::integer) as uk_fee")
-        .joins(latest_published_enrichment_join_sql)
+        .joins(latest_published_enrichment_left_join_sql)
```
kmcd: I couldn't find published enrichments in previous cycles. 

Will this work with published enrichments?

app/services/saved_courses/query.rb
```diff
@@ -44,21 +44,24 @@ def location_scope
...
             course_site.course_id = course.id
             AND course_site.status = 'R'
             AND course_site.publish = 'Y'
           )
-          INNER JOIN site ON (
+          LEFT JOIN site ON (
             site.id = course_site.site_id
             AND site.discarded_at IS NULL
+            AND site.longitude IS NOT NULL
+            AND site.latitude IS NOT NULL
```
kmcd, line 62: This was an OR query: `.where("(site.longitude IS NOT NULL OR site.latitude IS NOT NULL)")`

Is this deliberate change from OR to AND or a side effect?

app/services/saved_courses/query.rb
```diff
@@ -178,5 +190,9 @@ def preload_scope
         course: [:provider, :latest_published_enrichment, { subjects: [:financial_incentive] }],
       )
     end
+
+    def latest_published_enrichment_left_join_sql
+      latest_published_enrichment_join_sql.sub("INNER JOIN LATERAL", "LEFT JOIN LATERAL")
```
kmcd: This looks odd - it's re-writing the query?

Does this affect lines 154 and 175 ? `.where("course_enrichment.id IS NOT NULL OR provider.recruitment_cycle_id <> ?", …)
`

kmcd (changes requested):

app/services/saved_courses/query.rb
gms-gs: Yes, those wheres are the other half of it. LEFT JOIN means enrichment can be missing; the where keeps current-cycle rows only if they have one, and keeps previous-cycle rows either way. The .sub was a bit opaque, so the parent join now takes join_type: "LEFT" instead.

app/services/saved_courses/query.rb
gms-gs: Deliberate. ST_MakePoint needs both coordinates, so a site with only lat or only lng isn’t usable for distance. Those sites are treated as having no location; previous-cycle courses are still kept by the recruitment_cycle_id condition.

app/services/saved_courses/query.rb
gms-gs: Yes. Previous-cycle courses often have no published enrichment, which is why this is a LEFT JOIN — otherwise they disappear from fee sort. If they do have one, the lateral join still uses it. Current-cycle courses still need a published enrichment. I’ve added a spec that a previous-cycle course with a published fee sorts in the right place.

app/components/saved_courses/summary_card_component.rb
gms-gs: Good catch. The card reads course.recruitment_cycle.year, which goes through the provider, and we weren’t preloading that. SavedCourses::Query now preloads { provider: :recruitment_cycle }.

app/services/saved_courses/query.rb
gms-gs: Yes — current is find_by, so it can be nil and .id would raise. It shouldn’t happen in normal use, but I’ve switched this to RecruitmentCycle.current_recruitment_cycle! and memoised the id so we fail clearly if the current cycle row is missing.

kmcd (approved):

PR #6332, opened 2026-08-18
School changes summary

tomas-stefano:
## Context

Providers updating long placement school lists across many courses couldn't see what they'd actually changed before saving. The attach schools page now names those schools under the list, live.

Same block on the add course wizard step — adds only, since nothing is attached yet.

Wording follows the end state, not which control was used: every school ticked reads "adding all schools in your list", none ticked reads "removing all schools in your list".

JS-only and hidden by default. What gets submitted is unchanged.

### How the JS works

```mermaid
flowchart TD
    A[Tick a school] -->|change bubbles| C
    B[Tick Select all] -->|change bubbles| C
    C[schools-changes controller] --> D["Read every checkbox<br/>(including ones the search hid)"]
    D --> E{Compare against<br/>attached uuids from server}
    E -->|no difference| F[Hide summary]
    E -->|difference| G[Name added and removed schools]
```

Select all sets the other boxes programmatically, firing no event on them — but its own `change` bubbles, so one delegated listener covers both paths.

Attached uuids come from the server, not from the boxes: after a validation error the form re-renders from the submitted params, so the boxes no longer remember what's on the course.

## Testing

2 pages to acess and test:

1. Attach schools (edit) — adds and removes

```
  /publish/organisations/:provider_code/:year/courses/:code/schools
```

  2. Add course wizard, schools step — adds only

```
  /publish/organisations/:provider_code/:year/course_wizard/:state_key/:step
```

  with :step = schools. 

Start the wizard at …/course_wizard/new and it'll walk you to that step with a generated :state_key — you can't hand-craft that URL. The wizard is behind the wizard_add_course_flow feature flag, so enable that first if it's off.


tomas-stefano:
https://trello.com/c/x3OLfcMY/1876-display-schools-added-and-removed-on-attach-page

app/javascript/publish/controllers/schools_changes_controller.js
```diff
@@ -0,0 +1,84 @@
+import { Controller } from '@hotwired/stimulus'
+
+// The play-back under a list of school checkboxes: which schools the provider
```
inulty-dfe: play-back sounds jargony here, it's not explained

app/javascript/publish/controllers/schools_changes_controller.js
```diff
@@ -0,0 +1,104 @@
...
+    const list = document.createElement('ul')
+
+    list.className = 'govuk-list govuk-list--bullet'
+
+    schools.forEach(school => {
+      const item = document.createElement('li')
+
+      item.textContent = school.dataset.schoolName
+      list.append(item)
+    })
```
inulty-dfe: It's not ideal generating html in stimulus controllers with govuk classes int javascript classes but not sure what else is an improvement

app/forms/publish/course_school_form.rb
```diff
@@ -10,7 +10,14 @@ class CourseSchoolForm < BaseCourseForm
...
     def compute_fields
-      { school_uuids: current_school_uuids }.merge(new_attributes)
+      { school_uuids: attached_school_uuids }.merge(new_attributes)
+    end
+
+    # The schools on the course as it stands. It seeds the ticked boxes, and the
+    # page also hands it to the browser as what the provider's changes are
+    # measured against.
+    def attached_school_uuids
+      course.schools.joins(:provider_school).pluck("provider_school.uuid")
```
inulty-dfe: Does this need memoizing?

spec/system/publish/courses/schools/school_changes_summary_spec.rb
```diff
@@ -0,0 +1,225 @@
...
+      (() => {
+        const summary = document.querySelector('.app-schools-changes')
+        const button = document.querySelector('button.govuk-button[type="submit"]')
+
+        return button.getBoundingClientRect().top - summary.getBoundingClientRect().bottom
+      })()
+    JS
+
+    expect(gap).to eq(pixels)
+  end
```
inulty-dfe: I'm not keen on testing the pixel gap between two elements. Anyway, this breaks on smaller viewports

app/components/publish/schools/changes_summary_component.html.erb
```diff
@@ -0,0 +1,23 @@
+<div data-controller="schools-changes"
+     data-schools-changes-attached-value="<%= attached_value %>"
+     data-action="change->schools-changes#update">
+  <%= content %>
+
+  <div class="app-schools-changes govuk-!-margin-top-6"
```
inulty-dfe: This goes from hidden to visible and rewrites its contents on every tick with no `role="status"` / `aria-live`, so a screen reader user gets no announcement that the play-back appeared or changed, they'd only find it by navigating past the whole checkbox list. `SearchPanelComponent` already does this with a visually hidden `role="status" aria-live="polite"`, so I assume this is an oversight rather than a decision.

app/views/publish/course_wizards/steps/_schools.html.erb
```diff
@@ -25,37 +25,45 @@
...
-            data-schools-list-target="showAll"
-            data-action="schools-list#showAll">
-      <%= t("publish.courses.schools.show_all") %>
-    </button>
-  <% end %>
+<%# The summary plays back what the ticks add up to; it takes in the whole list,
+    so it wraps the search and the collapse rather than sitting inside either of
+    them. A course that does not exist yet has no schools attached, so it is only
+    ever told about schools being added. %>
+<%= render ::Publish::Schools::ChangesSummaryComponent.new do %>
```
inulty-dfe: The comment is right for first entry, but the step persists `school_uuids` in the state store and `Reviewable` puts a Change link for schools on check-answers, so Back or Change re-renders with the boxes already
  ticked. `connect()` measures those against a baseline of `[]`, and the provider is told they are adding N schools before they have touched anything.

  Passing the step's persisted `school_uuids` as the baseline plays back nothing until something actually changes.

app/javascript/publish/controllers/schools_changes_controller.js
```diff
@@ -0,0 +1,104 @@
...
+  // measured against. The server has to say: after a validation error the form
+  // comes back holding what was submitted, so the boxes no longer remember it.
+  static values = { attached: Array }
+
+  connect () {
+    this.update()
+  }
+
+  update () {
+    const schools = this.schools()
```
inulty-dfe: `update()` reaches straight for `summaryTarget`, so it throws `Missing target element "summary"` if Stimulus connects before the browser has parsed the children.
  ```suggestion
    update () {
      if (!this.hasSummaryTarget) return

      const schools = this.schools()
  ```

app/javascript/publish/controllers/schools_changes_controller.js
```diff
@@ -0,0 +1,104 @@
...
+    this.summaryTarget.hidden = added.length === 0 && removed.length === 0
+
+    // Every school ticked, or none left ticked, is better said than listed - and
+    // it is the end state that decides, not which control got them there.
+    this.fill(this.addedTarget, added, checked.length === schools.length)
+    this.fill(this.removedTarget, removed, checked.length === 0)
+  }
+
+  schools () {
+    return Array.from(this.element.querySelectorAll('input[type=checkbox][data-school-name]'))
```
inulty-dfe: This finds the school checkboxes by looking for `data-school-name`, so any checkbox that happens to carry a name counts as one. It also means you can't tell from the ERB that this controller touches these inputs. The other two controllers declare a target in the data hash right there on the checkbox, this one doesn't.

Adding `schools_changes_target: "school"` to that hash is one line and makes all three consistent.

Worth keeping `data-school-name` itself, though. The label has the newly added tag joined onto the name, so you can't read the name back off it.

inulty-dfe (commented):
We are using events for one behaviour and bubbling for another in the same page. It would be better if there was consistency.

This schools list is getting **super** complicated

tomas-stefano:
> We are using events for one behaviour and bubbling for another in the same page. It would be better if there was consistency.
> 
> This schools list is getting **super** complicated

Agreed, done. Rather than deferred. schools-search folded into schools-list — the three custom events between them were the only ones in the codebase and existed solely to join the two. Four controllers on the page → three, and this.dispatch now appears nowhere. 2f4560abc

app/javascript/publish/controllers/schools_changes_controller.js
tomas-stefano: Gone — it's "the summary" in comments, spec names and describes. 22e8666c2

app/javascript/publish/controllers/schools_changes_controller.js
tomas-stefano: the `<ul>`, `<li>`, `<h3>` and `<p>` now live in `<template>` elements in the component, cloned and filled with textContent. No govuk-* in JS, and & in a school name still  survives. 7245e0397

app/forms/publish/course_school_form.rb
tomas-stefano: Yes — the page calls it twice (ticking the boxes, then the baseline), so it was two queries. Memoised, matching schools below it. 2155c7b41

spec/system/publish/courses/schools/school_changes_summary_spec.rb
tomas-stefano: Deleted. You're right it breaks on narrow viewports, and it can no longer fail anyway c77c2c76a

app/components/publish/schools/changes_summary_component.html.erb
tomas-stefano: Oversight, fixed. 

Visually hidden role="status" following SearchPanelComponent. 

Counts only — "Adding 1 school. Removing 2 schools." — since the names are already on screen under their own heading. b81eacca8, c37cce5fa

app/views/publish/course_wizards/steps/_schools.html.erb
tomas-stefano: Real bug, fixed. 

1. the attribute is nil on first entry (serialises to "null", breaking the Stimulus Array value) and carries a leading "" from the  fieldset's hidden input — hence Array(...).compact_blank. 
2. it reads the store, not the step: after a failed submit the step is re-hydrated from params, so it holds what was just submitted bc5f8bc0e

app/javascript/publish/controllers/schools_changes_controller.js
tomas-stefano: Done

app/javascript/publish/controllers/schools_changes_controller.js
tomas-stefano: Done. 

`schools_changes_target: "school"` on the **checkbox**, 
`querySelectorAll` replaced by `this.schoolTargets`, and `data-school-name` kept for the reason you gave. 

I also put `schools-changes#update` in the same hash and dropped the wrapper's change-> action, so the row now declares both the target and the action — that's what fixes the input/change mismatch in your other comment. Cost is that the wizard checkbox gains a data-action where it previously had none. Which is fine fff290e54

app/javascript/publish/controllers/schools_changes_controller.js
```diff
@@ -0,0 +1,135 @@
...
+  static values = { attached: Array }
+
+  connect () {
+    this.update()
+  }
+
+  update () {
+    // Stimulus connects an element as soon as the parser reaches it, which can be
+    // before its children exist - and the summary is the last of them.
+    if (!this.hasSummaryTarget) return
```
inulty-dfe: If this ever returns early, nothing runs it again, so the summary sits stale until the first tick. `summaryTargetConnected()` is the callback for this case.

  Worth checking the other side of it too: `schools_list_controller` does the same work in `connect` with no guard at all. One of the two is wrong and I do not mind which way we settle it, I would just rather they matched.

app/javascript/publish/controllers/schools_changes_controller.js
```diff
@@ -0,0 +1,135 @@
...
+
+    const schools = this.schoolTargets
+    const attached = new Set(this.attachedValue)
+
+    const checked = schools.filter(school => school.checked)
+
+    // Every school ticked, or none left ticked, is better said than listed - and
+    // it is the end state that decides, not which control got them there.
+    const adding = {
```
inulty-dfe: Could we pull the add and remove decision out into a plain module, the way you did with `schools_search.js`?

  Everything interesting is these six lines plus the branching in `count` and `fill`, and the only thing testing it is the `:js` system specs. A `summarise(schools, attached)` returning `{ adding, removing }` would be testable in `spec/javascript` with no DOM, same as `schools_search.spec.js` already is, and the controller drops to DOM plumbing.

app/views/publish/courses/schools/edit.html.erb
```diff
@@ -29,58 +29,67 @@
...
-                  data: {
-                    select_all_checkboxes_target: "checkbox",
-                    schools_list_target: "schoolCheckbox",
-                    action: "select-all-checkboxes#updateSelectAll",
-                  } %>
-           <% end %>
+              <% @course_school_form.schools.each_with_index do |school, index| %>
+                <%# The name travels as an attribute of its own: the label may also
+                    carry the newly added tag, which the summary must not repeat. %>
+                <%= f.govuk_check_box :school_uuids,
```
inulty-dfe: This block and `_schools.html.erb:42-56` are near enough identical now, same comment and all, and each checkbox is carrying four data attributes plus a two action string.

It is the one place all three controllers meet, so it is the bit I would most want owned by a single component. Not asking you to do it here. Can we get a ticket for it?

app/components/publish/schools/changes_summary_component.html.erb
```diff
@@ -0,0 +1,54 @@
...
+       aria-live="polite"
+       data-schools-changes-target="status"
+       data-adding-one="<%= t(".status_adding_one") %>"
+       data-adding-other="<%= t(".status_adding_other") %>"
+       data-adding-all="<%= t(".status_adding_all") %>"
+       data-removing-one="<%= t(".status_removing_one") %>"
+       data-removing-other="<%= t(".status_removing_other") %>"
+       data-removing-all="<%= t(".status_removing_all") %>"></div>
+
+  <div class="app-schools-changes govuk-!-margin-top-6"
```
inulty-dfe: `.app-schools-changes`, `__added` and `__removed` have no CSS anywhere in the app. They only exist so the specs and the page object can find things.

Compare `.app-school-search`, which earns its class in `_school-search.scss`. The page object already reaches for `[data-schools-changes-target='status']` for the announcement, so it could do the same for the rest and we would lose three classes that look like styling but are not.

app/views/publish/courses/schools/edit.html.erb
```diff
@@ -29,58 +29,67 @@
...
-                     data-select-all-checkboxes-target="selectAll"
-                     data-action="select-all-checkboxes#toggleAll">
+              <div class="govuk-checkboxes__item" data-schools-list-target="bulkSelect">
+                <input type="checkbox"
+                       class="govuk-checkboxes__input"
+                       id="select-all-schools"
+                       data-select-all-checkboxes-target="selectAll"
+                       <%# Setting the boxes fires nothing on them, so the summary has
+                           to hear about it from this one. %>
+                       data-action="select-all-checkboxes#toggleAll schools-changes#update">
```
inulty-dfe: `data-action="select-all-checkboxes#toggleAll schools-changes#update"` works because Stimulus runs the two in the order the attribute lists them. Swap them and the summary reads the ticks from before the toggle.

Your Select all spec catches it, so this is not a risk, it just is not visible from the code.

  Could `toggleAll` dispatch instead of the two racing?
```js
      toggleAll (event) {
        const checked = event.target.checked
        this.checkboxTargets.forEach(box => { box.checked = checked })
        this.selectAllTarget.indeterminate = false
        this.dispatch('toggled', { target: this.selectAllTarget })
      }
```
  and on the row:

```html
      data-action="select-all-checkboxes#toggleAll
                   select-all-checkboxes:toggled->schools-changes#update"
```

  Both declarations stay on the row, nothing goes back on the wrapper, and the order stops mattering because the second entry listens for something that cannot exist until the first has finished. Support feedbacks uses the same controller and just dispatches into nothing.

I know you took three custom events out when you folded search into the list, so say if you would rather not put one back. I think this one is a different kind: those three were one controller telling another what to do, this is a controller saying it has finished and anything is free to ignore it. If you would rather leave it alone, extending the comment above to say the order is load bearing would do me.

inulty-dfe (commented):
Looks much easier to understand since merging the controller and dropping much of the intercontroller events.

app/components/publish/schools/changes_summary_component.rb
```diff
@@ -0,0 +1,34 @@
...
+    # summary ships hidden and empty, and the schools-changes Stimulus
+    # controller is the only thing that ever fills it in, so a provider without
+    # JavaScript sees - and submits - exactly what they did before.
+    #
+    # The wording travels as data attributes rather than being built here
+    # because the counts are only known in the browser. `{count}` is substituted
+    # by the controller, not by I18n.
+    class ChangesSummaryComponent < ViewComponent::Base
+      # The schools attached to the course when the page was served: the state
+      # the provider's changes are measured against. The wizard attaches schools
```
samjohnson-dev: I think this comment could be updated slightly as it would start empty but the choosen schools would be attached right? so basically the last sentence

app/javascript/publish/controllers/schools_changes_controller.js
```diff
@@ -0,0 +1,135 @@
...
+export default class extends Controller {
+  static targets = [
+    'summary', 'added', 'removed', 'status', 'school',
+    'countTemplate', 'itemTemplate', 'messageTemplate'
+  ]
+
+  // The schools attached when the page was served, which is what the ticks are
+  // measured against. The server has to say: after a validation error the form
+  // comes back holding what was submitted, so the boxes no longer remember it.
+  static values = { attached: Array }
```
samjohnson-dev, line 27: `attached` is here so that after a validation error the boxes no longer mean “what the course had”.

I dont think we havbe a system spec for: uncheck every school -> submit -> error page still shows “removing all schools in your list” with no extra click. 
Is that worth doing?

samjohnson-dev (commented):
Couplke extra bits but agree with Iains comments I dont have further comments.

app/javascript/publish/controllers/schools_changes_controller.js
tomas-stefano: Good — added! :)

Passes as written: read the baseline off the boxes instead and both halves come out empty, the summary hides itself and the spec fails. Added on d33b6bda9

app/components/publish/schools/changes_summary_component.rb
tomas-stefano: Good spot. That went stale when the wizard started reading the state store. Reworded: empty on the way in, the schools already chosen when Back or Change returns. ae63023a7

app/javascript/publish/controllers/schools_changes_controller.js
tomas-stefano: Done. `schools_changes.js` next to `schools_search.js`, ten cases in spec/javascript, no DOM and noconfig. 

It also fixed something I'd missed: adding.schools was carrying live inputs all the way down to where the list items get written, so the decision and the page were never really separate. Names travel now. e2dbc4c31

app/views/publish/courses/schools/edit.html.erb
tomas-stefano: Took the dispatch. Worth saying it isn't a race though.

Stimulus sorts bindings by attribute index deliberately (`EventListener` keeps them in a field called `unorderedBindings` and sorts on every dispatch), so it's deterministic today. 

The trap is that the guarantee only survives while both tokens stay bare on the same element; put change-> on either and they become two listeners and the order stops meaning anything. Checked it both ways: reversing the two actions fails the new spec before the change, passes after. 9ff511ab4

app/components/publish/schools/changes_summary_component.html.erb
tomas-stefano: Right, no CSS anywhere. Swapped to the targets. One thing though: app-table--courses__* is five unstyled classes I added in June specifically to replace data-qa on the course list, so the two are now inconsistent. Worth deciding what happens to that one. 9f085824e

app/views/publish/courses/schools/edit.html.erb
tomas-stefano: Done — partial with locals rather than a component. The builder goes in as a local the way Rails expects, and everything in publish/schools takes plain data and gets render_inline'd, which a builder would spoil. Two knobs: select_all (wizard has no bulk tick) and include_hidden, which is a real difference — the edit page needs the hidden field so unticking everything submits an empty list. 7506a4584

app/javascript/publish/controllers/schools_changes_controller.js
tomas-stefano: My fault, I added that last round and the commit message claiming a race was wrong. 

Stimulus awaits `domReady` before starting its router, so `connect()` can't run mid-parse.

`select-all-checkboxes` proves it on this same page — dereferences `selectAllTarget` unguarded in connect and has never thrown. Deleted rather than reaching for summaryTargetConnected, since there's no case to handle. That's also how the two end up matching. 3feb38444

samjohnson-dev (approved):
LGTM

PR #6339, opened 2026-08-19
Freeze time in the BigQuery event specs, not the Time class

kmcd:
## Context

`Time.freeze do` in these two specs is `Object#freeze` on the `Time` class, not ActiveSupport's `freeze_time`. So the block never runs.

Use `freeze_time`.

## Guidance to review

Passing CI should be enough.

## Checklist


inulty-dfe (approved):
LGTM

kmcd:
> LGTM

Thanks @inulty-dfe ! :shipit:

PR #6340, opened 2026-08-19
Remove the no-op provider index authorization

kmcd:
## Context

Follow-up to @tomas-stefano's [review comment](https://github.com/DFE-Digital/publish-teacher-training/pull/6324#discussion_r3802802329) on #6324. 

Remove no-op auth guards.

## Guidance to review

Sanity check the auth guards. 

## Checklist

### DIFF
--- a/app/controllers/publish/courses_controller.rb
+++ b/app/controllers/publish/courses_controller.rb
@@ -7,8 +7,6 @@ module Publish
     decorates_assigned :course
 
     def index
-      authorize :provider, :index?
-
       @filter_form = ::Publish::Courses::FilterForm.new(provider:, **course_filter_params)
       @course_list = ::Publish::CourseList.new(provider:, params: @filter_form.filter_params)
     end
--- a/app/controllers/publish/providers/schools/added_schools_controller.rb
+++ b/app/controllers/publish/providers/schools/added_schools_controller.rb
@@ -7,8 +7,6 @@ module Publish
         before_action :render_not_found, unless: :rollover_period_2026?
 
         def index
-          authorize(provider, :index?)
-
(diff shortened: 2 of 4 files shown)


samjohnson-dev (approved):
LGTM

kmcd:
> LGTM

Thanks @samjohnson-dev ! :shipit:

tomas-stefano (approved):
Approving. All four call sites sit under `resources :providers, param: :code`, so `authorize_provider` still runs — no behaviour change, and nothing here skips `verify_authorized`, so a broken chain would raise rather than serve the page. 

Good call leaving `ProvidersController#index` alone; a short comment there would stop the next person applying the same reasoning to it. 

Non-blocking follow-up: extend `authorization_spec.rb` to cover added-schools, removed-schools and training-partners courses, now that nothing local documents the guarantee.

kmcd:
> Approving. All four call sites sit under `resources :providers, param: :code`, so `authorize_provider` still runs — no behaviour change, and nothing here skips `verify_authorized`, so a broken chain would raise rather than serve the page.
> 
> Good call leaving `ProvidersController#index` alone; a short comment there would stop the next person applying the same reasoning to it.

Think the existing specs are enough to catch it if anyone changes this.

> Non-blocking follow-up: extend `authorization_spec.rb` to cover added-schools, removed-schools and training-partners courses, now that nothing local documents the guarantee.

Do you think these specs already cover it? 

```
spec/system/publish/viewing_training_partners_and_their_courses_spec.rb
spec/system/publish/providers/providers_added_schools_spec.rb
spec/system/publish/providers/providers_removed_schools_spec.rb
```

PR #6243, opened 2026-07-22
Banner centre

dcyoung-dev:
## Guidance to review

- Log in as a Support User
- Go to "Settings" then "Banners"
- Create a new banner and assign it to any interface.
- Log out and visit the interfaces the Banner has been created for.
- See banner(s)
- If the interface includes a flash message, the banner will not appear.

## Checklist

- [ ] I have moved hard-coded strings to locale files.
- [ ] I have removed the usage of `data-qa` attributes in HTML files and updated the corresponding tests.

PR #6341, opened 2026-08-19
Finalise banner centre

kmcd:
## Context

Banners are hard-coded in views today, so putting one up needs a PR and a deploy.

This lets support users write and schedule their own for Find, Publish and
Support.

Picks up @dcyoung-dev's #6243 prototype.

On top of David's prototype:
- A banner can't be saved empty, or with nowhere to show.
- Moved into view components for consistency

Delete is only offered before a banner is published. After that it can only be expired,
so there's an audit trail of what was shown.

Screenshots to follow.

## Guidance to review

- Settings → Banners. Add one, tick a service, publish it.
- Check it shows on the services, and doesn't obscure flash messages
- Saving an empty banner should trigger validation errors.

**Questions**:

- Does this need a feature flag? It does nothing until someone publishes a banner.
- Is markdown right for the body? Support text will show on public Find pages.
- Should support choose green styling and heading levels per banner?
- Do we need a review on the public facing text?

## Checklist


db/schema.rb
```diff
@@ -25,6 +25,11 @@
     t.string "ProductVersion", limit: 32, null: false
   end
 
+  create_table "airbyte_heartbeat", id: false, force: :cascade do |t|
+    t.integer "id"
+    t.timestamptz "last_heartbeat"
+  end
```
tomas-stefano: `airbyte_heartbeat` isn't on main and no migration here creates it. Probably local DB leaked into the dump.

docs/database-diagram.mmd
```diff
@@ -442,64 +453,67 @@ erDiagram
...
+	RecruitmentCycle o|..}o "Provider::School" : ""
+	RecruitmentCycle o|..}o "Course::School" : ""
+	RecruitmentCycle o|..}o GiasSchool : ""
+	Site ||--}o SiteStatus : ""
+	Site ||--}o StudySitePlacement : ""
+	SubjectGroup o|--}o Subject : ""
+	SubjectArea ||--}o Subject : ""
+	User ||--}o UserNotification : ""
+	User ||--}o UserPermission : ""
+	User o|--}o CourseEnrichment : ""
```
tomas-stefano, line 527: same stale DB above.

db/migrate/20260716112933_create_banner.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.datetime :published_at
+      t.datetime :expired_at
+      t.boolean :display_on_find
+      t.boolean :display_on_publish
+      t.boolean :display_on_support
+
+      t.timestamps
+    end
+  end
+end
```
tomas-stefano, line 16: `create_banner` adds `title`, `title_heading_level` and `heading`, then two later migrations drop all three and a fourth re-adds heading. None of it has shipped.

db/migrate/20260716112933_create_banner.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string :name, null: false
+      t.string :title
+      t.integer :title_heading_level
+      t.string :heading
+      t.text :body
+      t.datetime :published_at
+      t.datetime :expired_at
+      t.boolean :display_on_find
+      t.boolean :display_on_publish
+      t.boolean :display_on_support
```
tomas-stefano: body, published_at and the three display_on_* are nullable with no defaults, but the model requires them. can we add a `null: false` on the first two, `default: false` on the booleans?

app/views/layouts/support.html.erb
```diff
@@ -54,6 +54,10 @@
       <%= yield :before_content %>
 
       <main class="govuk-main-wrapper" id="main-content" role="main">
+        <div class="govuk-width-container">
```
tomas-stefano: The wrapping govuk-width-container sits inside main, which is already inside one (line 52). Double
  padding, banner doesn't line up with the rest of the page. Not sure if that's the intention. Can you confirm it?

app/views/support/banners/index.html.erb
```diff
@@ -0,0 +1,60 @@
...
+
+  <%= govuk_button_link_to t(".add_banner"), new_support_banner_path %>
+
+  <%= tag.div do %>
+    <% tabs = [
+      { status: :scheduled, path_to: scheduled_support_banners_path },
+      { status: :active, path_to: active_support_banners_path },
+      { status: :expired, path_to: expired_support_banners_path },
+    ] %>
+    <ul class="govuk-tabs__list">
```
tomas-stefano: Hand-rolled govuk-tabs__list with no `govuk-tabs` parent, so it gets none of the component's  styling, and the current tab is signalled by a CSS class only. 

Is it intentional?

app/components/active_banners_component.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def initialize(interface:, flash:)
+    super()
+
+    @interface = interface
+    @flash = flash
+  end
+
+  def render?
+    flash.empty? && banners.any?
```
tomas-stefano: Not sure `flash.empty?` is the gate we want. `application.html.erb` is the layout for all of Publish, and 30 controllers there set a success flash on save — so a provider going through course creation has a flash on nearly every step and never sees the banner. On Find flash is rare, which is where the `DeadlineBannerComponent` precedent comes from, so we've taken the cheap version of the trade and applied it where it costs most.

It also doesn't get us to one banner at a time. Here's what's actually on the page, top to bottom:

```mermaid
flowchart LR
  subgraph P["Publish — application.html.erb"]
    direction TB
    P1["ActiveBanners"]:::gated --> P2["FlashBanner"] --> P3["MaintenanceBanner"] --> P4["rollover banner"]
  end

  subgraph F["Find — find.html.erb"]
    direction TB
    F1["MaintenanceBanner"] --> F2["DeadlineBanner"]:::gated --> F3["FinancialIncentivesBanner"] --> F4["ActiveBanners"]:::gated --> F5["FlashBanner"]
  end

  classDef gated stroke-dasharray: 5 4;
```

Dashed = gated on flash. Two of five, the rest stack freely, and `with_collection` means two active records are already two notification banners. Worth a think about who owns the "only one banner" rule — right now nobody does, and it's getting crowded.

One thing that is just a bug: `flash.empty?` isn't the same as "a flash banner will render". `FlashBanner#render?` is `flash.any?` but its template only handles `success`, `warning`, `info`, `error`, `success_with_body`, `sign_in`. `flash[:alert]` (`support/feedbacks_controller.rb:18`) matches none of them, so we hide our banner for a flash that draws nothing.

Suggest dropping the check and capping the collection if stacking is the concern. Also the two layouts disagree on placement (above `FlashBanner` on Publish, below three banners on Find) — probably want those consistent.

What do you think?

db/schema.rb
kmcd: This was WIP cruft. Gone now 👍 

Added `airbyte_heartbeat` to `SchemaDumper.ignore_tables` to prevent this leaking again.

docs/database-diagram.mmd
kmcd: This was WIP cruft. Gone now 👍

db/migrate/20260716112933_create_banner.rb
kmcd: Good call - done 👍 

In previous projects I've worked on, this was enforced with a linter or similar.

db/migrate/20260716112933_create_banner.rb
kmcd: Yep, these are WIP migrations, I'll wrap them up.

> None of it has shipped.

They were needed for the review app, where they did ship.

app/views/layouts/support.html.erb
kmcd: Good catch - it's a typo. Gone now 👍

app/views/support/banners/index.html.erb
kmcd: Leftover from David's prototype. I thought it was OK ... 

However, after looking at the rest of the app, swapped for `TabNavigation`, similar to what Support uses.

app/components/active_banners_component.rb
kmcd: OK quite a bit to unpack here ...

> Not sure flash.empty?

Good spot - dropped the flash checking.

Also made the banners consistent as suggested:
* Publish is Maintenance → rollover → ours → flash.
* Find is Maintenance → deadline → incentives → ours → flash.

Not concerned about stacking and don't think we should cap. If we did overlapping banners would be automatically hidden at runtime which would be confusing for Support users. Think it's better they both render and it's obvious what's going on.

Agree nobody owns the "only one banner" rule, but think it's out of scope for this PR.

app/lib/potential_date_time.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def to_time
+    return unless [year, month, day].all? { |part| part.to_s.match?(/\A\d+\z/) }
+    return unless [hour, minute].all? { |part| part.blank? || part.to_s.match?(/\A\d+\z/) }
+    return unless Date.valid_date?(year.to_i, month.to_i, day.to_i)
+    return unless hour.to_i.between?(0, 23) && minute.to_i.between?(0, 59)
+
+    Time.zone.local(year.to_i, month.to_i, day.to_i, hour.to_i, minute.to_i)
+  end
```
tomas-stefano, line 31: There are two gaps in the date parsing.

The year isn't range-checked, so `"26"` sails through as year 0026: the banner goes live immediately, never expires, and shows as `0026-01-01 09:00 LMT` in the list. `"0"` and `"12345"` do the same. Given the commit is about rejecting mistypes, a two-digit year is the one to catch.

Blank hour and minute become `00:00` through `to_i`. Fine for publish, wrong for expiry — "expires 5 September" with the time left blank pulls the banner at the start of the 5th, so it isn't shown that day at all. Either require the time or take a blank expiry time as end of day.

Both drop into `spec/forms/support/banner_form_spec.rb` and fail as it stands:

```ruby
it "rejects a two-digit year" do
  subject = form(published_at: { "1i" => "26" })

  expect(subject).not_to be_valid
  expect(subject.errors[:published_at]).to contain_exactly("Enter a valid publish date and time")
end

it "expires at the end of the day when no expiry time is given" do
  subject = form(expired_at: { "1i" => "2026", "2i" => "9", "3i" => "5", "4i" => "", "5i" => "" })

  expect(subject.expired_at).to eq(Time.zone.local(2026, 9, 5).end_of_day)
end
```

Today the first gives `published_at` of `0026-01-01 09:00` and passes validation; the second gives `2026-09-05 00:00`.

docs/database-diagram.mmd
```diff
@@ -152,6 +163,7 @@ erDiagram
 		integer gias_school_id FK
 		integer id PK
 		integer provider_school_id FK
+		text site_code
```
tomas-stefano: is this correct? I think still outdated 🤔

app/controllers/support/banners_controller.rb
```diff
@@ -0,0 +1,104 @@
...
+      @banner = Banner.find(params[:id])
+    end
+
+    def merge_date_errors(form)
+      @banner.validate
+
+      form.errors.each do |error|
+        @banner.errors.delete(error.attribute)
+        @banner.errors.import(error)
+      end
```
tomas-stefano, line 70: The error summary comes out in the wrong order when a date error and a model error land together.

`merge_date_errors` runs `@banner.validate` first, which fills the errors in the model's declaration order and ends on `displayed_on_an_interface`. The form's date errors are then deleted and re-imported, so they get appended after everything else.

To see it:

1. Support → Banners → Add banner
2. Name and Body: anything valid
3. Publish date: year `2O26` (capital O), or day `32`
4. Leave all three "Displayed on" boxes unticked
5. Continue

Summary reads:

```
Select at least one interface to display the banner on
Enter a valid publish date and time
```

The form asks for the dates above the checkboxes, so the summary should be the other way round — GOV.UK expects summary order to match the order of the fields on the page.

Same thing as a spec, alongside the existing ones in `spec/requests/support/banners_spec.rb`:

```ruby
it "lists errors in the order the questions are asked" do
  post support_banners_path, params: { banner: {
    name: "Test banner",
    body: "text",
    "published_at(1i)" => "2O26",
    "published_at(2i)" => "1",
    "published_at(3i)" => "1",
    "published_at(4i)" => "9",
    "published_at(5i)" => "0",
    display_on_find: "false",
    display_on_publish: "false",
    display_on_support: "false",
  } }

  summary = Nokogiri::HTML(response.body).css(".govuk-error-summary__list li").map { |item| item.text.strip }

  expect(summary).to eq([
    "Enter a valid publish date and time",
    "Select at least one interface to display the banner on",
  ])
end
```

Easiest fix is to run the form's validations first and merge the model's errors into it, rather than the other way round.

tomas-stefano (approved):
Minor things. I think the potential outdated doc is the one to have a closer eye. The rest are minor, I will approve it, good work 🚀

docs/database-diagram.mmd
kmcd: Yep, forgot to remove the auto-generated field. Gone now.

app/lib/potential_date_time.rb
kmcd: Good spot - the date parsing is a bit more complicated than I'd like.

Fixed now.

app/controllers/support/banners_controller.rb
kmcd: Actually went with the UK Gov gem's built in way to order the errors: https://github.com/DFE-Digital/publish-teacher-training/pull/6341/changes/1506bfaaa6293f7a3f71074d7f8ac675466d8748#diff-3c669fde083dfdbcee8f94815175939c1de0e0356949cc620b0d4c94a4aca183R3

Didn't spot this before ...

kmcd:
[Finalising the banner center](https://trello.com/c/0CZeykt8/1901-finalising-the-banner-center)

inulty-dfe (approved):
Looks great nice work

PR #6344, opened 2026-08-21
Move error reporting rate limiter onto Rails.cache

samjohnson-dev:
## Context

`ErrorReporting::RateLimiter` was the only code path still talking to Redis via `RedisClient.cache`. Everything else already uses `Rails.cache`. This moves the limiter onto `Rails.cache` so we can drop that dedicated Redis client.

No infra or cache-store changes — production still uses Redis via `redis_cache_store`.

- Refactor `ErrorReporting::RateLimiter` to store event timestamps in `Rails.cache` (sliding window kept)
- Update the rate limiter specs
- Remove `RedisClient.cache` from `config/initializers/redis.rb` and `spec/support/mock_redis.rb`

### DIFF
--- a/app/lib/error_reporting/rate_limiter.rb
+++ b/app/lib/error_reporting/rate_limiter.rb
@@ -1,30 +1,32 @@
 module ErrorReporting
-  # Use Redis to keep track of errors reported.
-  # If more than threshold of any error exists then report
-  # Always delete entries older than the window
+  # Count errors in Rails.cache using per-minute counters.
+  # Report when the total across the window reaches the threshold.
   #
-  # Return true if more than threshold entries exist withiing the window
+  # Returns true when the threshold has been reached within the window.
   class RateLimiter
+    BUCKET = 1.minute
+
     def self.report?(key:, threshold:, window: 1.hour)
-      redis = RedisClient.cache
-      now = Time.zone.now.utc.to_f
-      events_key = "error_reporting:events:#{key}"
+      now = Time.zone.now
+      bucket = now.to_i / BUCKET.to_i
+      expires_in = window.to_i + BUCKET.to_i
 
-      count = redis.multi { |r|
-        # Add one or more members to a sorted set
-        r.zadd(events_key, now, "#{now}-#{SecureRandom.hex(4)}")
-        # Remove all members in a sorted set within the given scores.
-        r.zremrangebyscore(events_key, "-inf", now - window.to_i)
-        # Set a key's time to live in seconds.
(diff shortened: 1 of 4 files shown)


inulty-dfe (commented):
Behaviour looks the same when one request runs at a time, I checked that pretty
  carefully. The problem is what happens when a few come in at once.

  Before, Redis did the whole count-and-update in one go, so nothing could get lost.
  Now we read the list, add to it in Ruby, then write it back. If two requests do that
  at the same time, they both read the same list and the second one overwrites the
  first. I tried it with 20 requests at once and only 1 got saved, so it never hit the
  threshold at all.

  That's a problem here because these auth errors tend to come in a flood, and across
  several servers, which is exactly when they'd trample each other. So we'd end up
  undercounting and the Sentry alert either fires late or not at all, and we'd have no
  idea.

  Other thing: we're now sending the entire list back and forth on every single error,
  and it just keeps growing. During a big spike that's a lot of traffic to Redis at the
  worst possible moment.

  Easiest fix is probably to keep a simple counter per minute and add them up, rather
  than storing every timestamp. Redis can bump a counter safely even when things happen
  at once, and it stays small.

samjohnson-dev:
Thanks @inulty-dfe, made that change and tested locally best i could with the caching on and threshold workjing, ie firing on the 10th. also the old count drops off after the window period.

Havent tested a real multi-service race case but think that should be an issue as we are caching using increment compared to first approach.

app/lib/error_reporting/rate_limiter.rb
```diff
@@ -1,30 +1,32 @@
...
-      }.last
+      minutes = window.to_i / BUCKET.to_i
+      previous_keys = (1...minutes).map { |offset| cache_key(key, bucket - offset) }
+      previous_total = Rails.cache.read_multi(*previous_keys, raw: true).values.sum(&:to_i)
 
-      count >= threshold
-    rescue Redis::BaseError => e
+      current + previous_total >= threshold
+    rescue StandardError => e
       Rails.logger.tagged("ErrorReporting::RateLimiter") { |l| l.error(e.message) }
```
inulty-dfe: `e.message` logs only the error text, so `e.full_message` is used instead to include the exception class and backtrace, which tells you which line and which cache call actually failed rather than leaving a bare string like "Connection refused".

https://docs.ruby-lang.org/en/3.4/Exception.html#method-i-full_message

inulty-dfe (approved):
Last comment, I'll approve now so but recommend changing the exception method

PR #6346, opened 2026-08-21
Exclude closed courses from email alerts

tomas-stefano:
## Context

Candidates were getting email alerts that listed courses which are closed for applications. Clicking one of those courses in the email took them to a page saying it is not accepting applications. This happens because the alert reuses the Find search query, and that query only hides closed courses when the candidate ticked the "Applications open" box. Most saved alerts never ticked it, so closed courses were counted in the total and listed in the email.

The alert now always leaves closed courses out, whatever the candidate saved. If every new course that week is closed, no email is sent at all. Find search itself is not changed: it still shows closed courses with a "Closed" tag, which is right for browsing but wrong for an alert. Five specs cover this, split across five commits. Two unit specs fail without the fix, and three end to end specs check a mixed set of open and closed courses, a set where all are closed, and an alert that had already ticked the box.

### DIFF
--- a/app/services/find/process_weekly_email_alerts_service.rb
+++ b/app/services/find/process_weekly_email_alerts_service.rb
@@ -32,7 +32,9 @@ module Find
   private
 
     def find_matching_courses(alert, recently_published_ids)
-      ::Courses::Query.call(params: alert.search_params.dup).where(id: recently_published_ids)
+      params = alert.search_params.dup.merge(applications_open: true)
+
+      ::Courses::Query.call(params:).where(id: recently_published_ids)
     end
   end
 end
--- a/app/services/find/process_weekly_email_alerts_service.rb
+++ b/app/services/find/process_weekly_email_alerts_service.rb
@@ -32,9 +32,7 @@ module Find
   private
 
     def find_matching_courses(alert, recently_published_ids)
-      params = alert.search_params.dup.merge(applications_open: true)
-
-      ::Courses::Query.call(params:).where(id: recently_published_ids)
+      ::Courses::Query.call(params: alert.search_params_for_matching).where(id: recently_published_ids)
     end
   end
 end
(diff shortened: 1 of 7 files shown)


tomas-stefano:
https://trello.com/c/vVcQgtDx/1923-investigate-closed-courses-vs-email-alerts

app/services/find/process_weekly_email_alerts_service.rb
```diff
@@ -32,7 +32,9 @@ def call
   private
 
     def find_matching_courses(alert, recently_published_ids)
```
inulty-dfe, line 34: Should we rename this to `find_matching_open_courses`?

tomas-stefano: I think it would be better not. 

Because if more and more criterias are added the method might become inconcise or long. 

Find matching courses whathever the matching will be. Using business language, t's part of what an email alert is. An alert means "new courses you can apply to".

inulty-dfe (approved):

samjohnson-dev:
Looks good although i think in the email that gets sent there is a view more button and that will still show the closed courses as its built from the orginal query? Is that something we want to enforce if the user only wants applications open? Just a thought? Otherwise good and happy to approve

samjohnson-dev (approved):

tomas-stefano:
> Looks good although i think in the email that gets sent there is a view more button and that will still show the closed courses as its built from the orginal query? Is that something we want to enforce if the user only wants applications open? Just a thought? Otherwise good and happy to approve

Good catch, you were right. The link is built from the raw saved criteria in `EmailAlertMailer`, so it never picked up the filter.

Worth noting it was already a superset: it drops the recently-published constraint too, so it's really "your search on Find, newest first" rather than the remaining N. But I agree it shouldn't be offering closed courses when the email just excluded them.

Fixed in 54614ac  — I moved the forced filter onto `Candidate::EmailAlert#search_params_for_matching` so the matching and the link read one definition, instead of merging it in two places that can drift. The link now carries `applications_open=true`.

On whether to enforce it: I think yes. `applications_open` is a visible checkbox on the results page with a "selected" count, so the candidate can see it was applied and untick it if they want the wider list. The criteria section of the email still renders from `search_attributes`, so it keeps showing only what they actually chose.

PR #6347, opened 2026-08-21
Add a filter box to the provider schools list

inulty-dfe:
## Context

  The provider schools list paginates twenty at a time with no way to reach a particular school, so a provider with sixty pages through them by hand.

  Two things worth knowing. The suggestions endpoint answers to `?query=`, not `?filter=`, because the shared `remote-autocomplete` Stimulus controller hardcodes that parameter, the same one  `api/school_suggestions` already answers to. The list's own narrowing stays `filter`. And `Provider::School.filtered_by` matches through a subquery rather than a `merge`, because pg_search builds its own SELECT  and ORDER BY for ranking, which fights `ordered_by_name` and pagy's count.

  I considered giving `Publish::Schools::SearchPanelComponent` a remote mode instead of adding a component, but it filters checkboxes in the DOM where this submits a form, so the flag would have changed nearly  everything while the course schools page and wizard step still depend on the old behaviour. I also considered swapping the table client-side; a full page GET keeps pagination, the back button and shareable URLs  for free, and works with JavaScript off.

  - A filter box above the list. It GETs the index, which narrows server-side and paginates as before.
  - A provider-scoped suggestions endpoint feeds the autocomplete, so every suggestion is a school already in the list.
  - Filtering to nothing shows "No results found" and a link back, rather than a bare page.

  Screenshots to follow.

  ## Guidance to review

  Run the three new spec files under `spec/system/publish/providers/schools/` and `spec/requests/publish/providers/schools/`, then filter a provider's schools page by name, postcode and URN.

  The subquery in `filtered_by` is the part I am least sure of. Separately, the suggestions system spec is the repo's first JS-level coverage of `remote-autocomplete`, and it is what caught the `?query=` bug;  worth a view on whether it is the right shape to copy for the other autocompletes.

  ## Checklist

  - [ ] I have removed the usage of `data-qa` attributes in HTML files and updated the corresponding tests. Nothing to remove: the branch adds no `data-qa` attributes and touches no file containing one.

### DIFF
--- a/app/components/publish/providers/schools/filter_component.html.erb
+++ b/app/components/publish/providers/schools/filter_component.html.erb
@@ -1,0 +1,38 @@
+<%= form_with(url: index_path, method: :get, local: true) do |f| %>
+  <div class="app-school-search">
+    <div class="govuk-form-group govuk-!-margin-bottom-3">
+      <label class="govuk-label govuk-label--m" for="filter">
+        <%= t(".label") %>
+      </label>
+
+      <div class="govuk-hint" id="filter-hint"><%= t(".hint") %></div>
+
+      <div class="app-school-search__row">
+        <%# The controller wraps the field alone: it enhances the first
+            non-hidden input it finds, which must be the box and not the
+            Search button. %>
+        <div <%= tag.attributes(data: remote_autocomplete_data(path: suggestions_path, format: "school")) %>>
+          <%= text_field_tag :filter,
+                filter,
+                class: "govuk-input",
+                autocomplete: "off",
+                aria: { describedby: "filter-hint" } %>
+        </div>
+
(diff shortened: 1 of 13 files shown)


langesi:
First observation from doing a visuall QA of the review app: I found that the search worked as expected except in one case. When a main site is attached, it does not appear in the search results when searching for “Main Site.”

inulty-dfe:
> First observation from doing a visuall QA of the review app: I found that the search worked as expected except in one case. When a main site is attached, it does not appear in the search results when searching for “Main Site.”

Nice QA work :mage: 

I'll ask UCD about it

I think the implementation of including Main site is significant enough that it would be better to tackle in a new ticket

langesi (approved):
Great work @inulty-dfe! LGTM!

PR #6348, opened 2026-08-21
Fix flaky site code updates

kmcd:
## Context

`spec/models/provider/school_spec.rb` is flaky.

The provider school factory creates site codes with `String#next`:

However, when there are already 26 schools, this is a no-op:

```ruby
provider_school.update!(site_code: "Z")
```

Which means `changed_at` isn't updated.

Update with two character site code instead.

## Guidance to review

Run this on main and then in this branch to reproduce:

```
bundle exec parallel_rspec --pattern "spec/models/.*_spec.rb" -n 4 --only-group 2 --group-by filesize
```

## Checklist

### DIFF
--- a/spec/models/provider/school_spec.rb
+++ b/spec/models/provider/school_spec.rb
@@ -102,7 +102,7 @@ describe Provider::School do
       provider.update_columns(changed_at: 1.hour.ago)
 
       Timecop.freeze do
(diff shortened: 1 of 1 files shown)


inulty-dfe (approved):

kmcd:
Thanks @inulty-dfe ! :shipit:

PR #6349, opened 2026-08-21
Require authentication for Find search results

langesi:
## Context

The Find `/results` endpoint might be vulnerable to sudden high load from aggressive bots and scrapers because each request can perform geolocation and course-search queries.

This change introduces an emergency feature flag that requires candidates to authenticate before viewing search results. The flag is disabled by default and takes effect immediately when changed, without an application restart.

Candidates should return to their original search—including filters, ordering and pagination—after signing in.

- Add the `require_authentication_for_find_results` feature flag.
- Stop unauthenticated requests before starting geolocation or course-search work when the flag is active.
- Store the requested results path for up to 30 minutes and return the candidate to it after authentication.
- Limit stored return paths to 2 KB and validate that they are same-site before redirecting.
- Prefer the stored destination over OmniAuth’s origin, which is normally the homepage containing the sign-in form.
- Keep One Login routes and the One Login header available when the emergency flag is active.
- Keep saved courses and other candidate-account pages controlled separately by the existing `candidate_accounts` flag.
- Add messaging explaining that the candidate will return to their search after signing in.
- Add controller, request and component coverage for the feature flag and redirect behaviour.

There are no substantial layout changes requiring Before and After screenshots.

## Guidance to review

1. Leave `require_authentication_for_find_results` inactive and confirm `/results` remains publicly accessible.
2. Activate the flag through Support → Settings → Feature Flags.
3. Visit a filtered results URL while signed out and confirm:
   - the request redirects before executing the search;
   - the sign-in message explains that the search will be restored;
   - signing in returns to the exact filtered results URL.
4. Confirm an already authenticated candidate can access `/results` normally.
5. Confirm activating this flag alone does not enable saved courses or other candidate-account pages.
6. Deactivate the flag and confirm anonymous results become accessible immediately without restarting the application.

Review feedback would be particularly useful on the incident-control flag naming, the 30-minute return-path expiry, and the 2 KB path-size limit.

### DIFF
--- a/app/controllers/concerns/find/authentication.rb
+++ b/app/controllers/concerns/find/authentication.rb
@@ -3,6 +3,8 @@ module Find
     extend ActiveSupport::Concern
 
     RETURN_TO_AFTER_AUTHENTICATING_SESSION_KEY = "return_to_after_authenticating".freeze
+    # This intent is stored in the anonymous Rails cookie session before a
+    # database-backed candidate Session exists, so it needs its own expiry.
     RETURN_TO_AFTER_AUTHENTICATING_MAX_AGE = 30.minutes
     RETURN_TO_AFTER_AUTHENTICATING_MAX_BYTES = 2.kilobytes
 
--- a/app/controllers/concerns/find/authentication.rb
+++ b/app/controllers/concerns/find/authentication.rb
@@ -2,6 +2,10 @@ module Find
   module Authentication
     extend ActiveSupport::Concern
 
+    RETURN_TO_AFTER_AUTHENTICATING_SESSION_KEY = "return_to_after_authenticating".freeze
+    RETURN_TO_AFTER_AUTHENTICATING_MAX_AGE = 30.minutes
+    RETURN_TO_AFTER_AUTHENTICATING_MAX_BYTES = 2.kilobytes
+
     included do
       before_action :load_session
       helper_method :authenticated?
@@ -62,7 +66,7 @@ module Find
     end
 
     def request_authentication
-      session["return_to_after_authenticating"] = request.fullpath
+      store_return_to_after_authenticating
 
       respond_to do |format|
         format.html do
@@ -82,6 +86,13 @@ module Find
       end
     end
 
+    def return_to_after_authenticating
+      stored_value = session.delete(RETURN_TO_AFTER_AUTHENTICATING_SESSION_KEY)
+      path = return_to_path_from(stored_value)
+
+      url_from(path)
+    end
+
     def reason_for_request
       :general
     end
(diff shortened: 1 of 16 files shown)


app/views/layouts/find.html.erb
```diff
@@ -35,15 +35,17 @@
   <body class="govuk-template__body">
     <%= govuk_skip_link %>
 
-    <% if FeatureFlag.active?(:candidate_accounts) %>
+    <% if FeatureFlag.candidate_authentication_active? %>
```
kmcd, line 38: When I activate the flag the "This is a new service ..." strapline disappears:

Image

&nbsp;

Here's how I'm activating the flag locally:

```ruby
# activate
bin/rails runner 'FeatureFlag.deactivate(:candidate_accounts); FeatureFlag.activate(:require_authentication_for_find_results)'

# restore
bin/rails runner 'RedisClient.current.del("feature_flags_require_authentication_for_find_results"); FeatureFlag.activate(:candidate_accounts)'
```

app/controllers/concerns/find/authentication.rb
```diff
@@ -62,7 +66,7 @@ def start_new_session_for(user, oauth)
     end
 
     def request_authentication
-      session["return_to_after_authenticating"] = request.fullpath
+      store_return_to_after_authenticating
```
kmcd: Will this break the `record_recent_search` stuff ?

kmcd (changes requested):

app/views/layouts/find.html.erb
kmcd: Otherwise looks good 👍

app/controllers/concerns/find/authentication.rb
```diff
@@ -2,6 +2,10 @@ module Find
   module Authentication
     extend ActiveSupport::Concern
 
+    RETURN_TO_AFTER_AUTHENTICATING_SESSION_KEY = "return_to_after_authenticating".freeze
+    RETURN_TO_AFTER_AUTHENTICATING_MAX_AGE = 30.minutes
```
kmcd, line 8: This is the same as `Session::INACTIVITY_TIMEOUT` so it's defunct.

app/lib/feature_flags.rb
```diff
@@ -8,6 +8,7 @@ def self.all
       [:publish_rollover_banner, "Displays the rollover banner in Publish", "Find and Publish team"],
       [:bursaries_and_scholarships_announced, "Display scholarship and bursary information", "Find and Publish team"],
       [:candidate_accounts, "Enable candidate accounts feature", "Find and Publish team"],
+      [:require_authentication_for_find_results, "Require candidates to sign in before viewing Find search results", "Find and Publish team"],
```
kmcd: This could be a bit more intention revealing, something like:

"[EMERGENCY MODE] Require candidates to sign in before viewing Find search results. Off by default. Turn off after DDOS"

app/views/layouts/find.html.erb
langesi: Thanks, fixed

app/lib/feature_flags.rb
langesi: Good point, changed that, thanks!

app/controllers/concerns/find/authentication.rb
langesi: It's just incidental that they have the same value. `Session::INACTIVITY_TIMEOUT `applies to the database-backed session of an authenticated candidates.  The return path is stored in the Rails cookie session while the visitor is still anonymous. Usually before a candidate Session exists

kmcd (approved):
LGTM! 👍

PR #6366, opened 2026-08-24
Serve robots.txt from a route instead of a static file

inulty-dfe:
## Context

  `robots.txt` has served `Disallow: /` since January 2023, added in a commit about enabling Find routes in staging. There were no restrictions before that.

  Two effects. Find's sitemap, regenerated daily with over a thousand course URLs, has never been processed, because `Disallow: /` covers `/sitemap.xml` itself. And Find's course pages are absent from search
  results. An `Allow:` line added later to exempt the sitemap used an absolute URL where the standard requires a path, so parsers discard it.

  Editing the file in place was rejected. `ActionDispatch::Static` matches on path and ignores the Host header, so Find, Publish, the API and every non-production environment receive identical bytes, and all of
  those hosts serve content publicly. A route is the only way to vary the response.

  A blocklist of named AI crawlers was rejected in favour of an allowlist, which covers new crawlers without further edits.

  `RobotsController` does not inherit `ApplicationController`, because `enforce_basic_auth` returns 401 wherever basic auth is enabled and Google treats any 4xx on `robots.txt` as no restrictions. `indexable?`
  tests `Settings.environment.name` rather than `Rails.env`, since non-production environments run `RAILS_ENV=production`.

  - `robots.txt` varies by host. Only the canonical production Find domain admits crawlers; Publish, the API, `find-temp` and all non-production hosts keep the blanket block.
  - Approved search engines and link preview bots may crawl Find. Everything else, including AI training crawlers, falls to `User-agent: * / Disallow: /`. Applebot and `facebookexternalhit` are approved;
  `Applebot-Extended` and `meta-externalagent` are not.
  - Candidate accounts, authentication, unsubscribe links, tracking redirects and JSON endpoints are disallowed.
  - The response is cacheable for a day, matching `Find::SitemapsController`.
  - Add `noindex` meta tags to support and publish layouts

  No UI changes, so no screenshots.

  ## Guidance to review

  `bundle exec rspec spec/requests/robots_spec.rb` covers every host and environment.

  Check the disallow list hardest. It was derived from `config/routes/find.rb`, and a missing entry means a page gets indexed that should not be. `/candidate/`, `/auth/` and `/email-alerts/` matter most, the last
  because those URLs carry personal access tokens.

  The approved crawler list is a product decision as much as a technical one, particularly `facebookexternalhit`. Dropping it is a one-line change.

  Not addressed here: allowing `/results` with query parameters lets Google crawl unbounded filter combinations. That needs `rel=canonical` and selective `noindex`, which is a separate ticket.
## Checklist

- [ ] I have moved hard-coded strings to locale files. — n/a, no user-facing strings; `robots.txt` is a machine-read file with a fixed English format.
- [x] I have removed the usage of `data-qa` attributes in HTML files and updated the corresponding tests. — n/a, no HTML changed.

### DIFF
--- a/app/controllers/robots_controller.rb
+++ b/app/controllers/robots_controller.rb
@@ -10,6 +10,8 @@ class RobotsController < ActionController::Base
   # rubocop:enable Rails/ApplicationController
 
   def show
+    expires_in 1.day, public: true
+
     render template: robots_template, formats: :text, layout: false, content_type: "text/plain"
   end
 
--- a/app/controllers/robots_controller.rb
+++ b/app/controllers/robots_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# Served from a route rather than public/, because a static file is host-blind:
+# Find, Publish, the API and the test environments share one public directory.
+#
+# Not ApplicationController: enforce_basic_auth returns 401 wherever basic auth
+# is enabled, and Google reads any 4xx on robots.txt as "no restrictions".
+# rubocop:disable Rails/ApplicationController
+class RobotsController < ActionController::Base
+  # rubocop:enable Rails/ApplicationController
+
+  def show
+    render template: "robots/disallow", formats: :text, layout: false, content_type: "text/plain"
+  end
+end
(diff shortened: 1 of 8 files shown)


inulty-dfe:
[Review robots.txt](https://trello.com/c/BDhDIBgS/1913-review-robotstxt)

langesi (approved):
LGTM! Good work @inulty-dfe

PR #6369, opened 2026-08-25
Update Schools list content and styling

gms-gs:
## Context

Providers need a clearer schools list: school code and URN are less useful on this page than where the school is, how many courses use it, and a way to remove it without opening the school first.

Without a return path, the remove journey always sent people back to the school details page. From the list, Back and Cancel should return to the list, including any filter or page they were on.

- Drop school code and URN from the Publish schools table.
- Show the address under the school name, using `govuk-body govuk-hint`.
- Add a **Number of courses attached** column (`1 course` / `12 courses`), counting kept courses only.
- Add a **Remove school** column with a destructive link into the existing delete journey.
- Send Back and Cancel to the schools list when removal starts from the index (preserving filter and page), and to the school page when it starts from school details.

School details still show code and URN. Support’s schools list is unchanged.

## Guidance to review

Check the Publish schools list:

- Address sits under the name; code and URN are gone.
- Course counts exclude discarded courses.
- **Remove school** from the list: Back and Cancel return to the list (try with a filter and another page).
- **Remove school** from school details: Back and Cancel still return to that school.
- Successful removal still lands on the list.

Worth a look at `school_delete_return_path` in `Publish::Providers::SchoolsController` and whether `from=index` is the right way to carry the return path.

Covered by request, model, and system specs around the list, course counts, and delete Back/Cancel.

## Checklist

### DIFF
--- a/app/controllers/publish/providers/schools_controller.rb
+++ b/app/controllers/publish/providers/schools_controller.rb
@@ -10,7 +10,10 @@ module Publish
 
       def index
         @filter = params[:filter]
-        @pagy, @schools = pagy(provider.schools.filtered_by(@filter).ordered_by_name, limit: PER_PAGE)
+        @pagy, @schools = pagy(
+          provider.schools.filtered_by(@filter).ordered_by_name.includes(:kept_courses),
+          limit: PER_PAGE,
+        )
       end
 
       def show; end
@@ -32,7 +35,7 @@ module Publish
           flash[:success] = "School removed"
           redirect_to publish_provider_recruitment_cycle_schools_path
         else
-          redirect_to delete_publish_provider_recruitment_cycle_school_path(@provider.provider_code, school.recruitment_cycle.year, school.uuid),
+          redirect_to school_delete_path_with_return,
                       flash: { warning: cannot_remove_school_message }
         end
       end
@@ -54,6 +57,45 @@ module Publish
         @school ||= provider.schools.find_by!(uuid: params[:uuid]).decorate
       end
       helper_method :school
+
+      def school_delete_return_path
+        if returning_to_schools_index?
+          publish_provider_recruitment_cycle_schools_path(
+            @provider.provider_code,
+            school.recruitment_cycle.year,
+            **schools_index_return_params,
+          )
+        else
+          publish_provider_recruitment_cycle_school_path(
+            @provider.provider_code,
+            school.recruitment_cycle.year,
+            school.uuid,
+          )
+        end
+      end
+      helper_method :school_delete_return_path
+
+      def school_delete_path_with_return
+        delete_publish_provider_recruitment_cycle_school_path(
+          @provider.provider_code,
+          school.recruitment_cycle.year,
+          school.uuid,
+          **school_delete_return_params,
+        )
+      end
+      helper_method :school_delete_path_with_return
+
+      def returning_to_schools_index?
+        params[:from] == "index"
+      end
+
+      def school_delete_return_params
+        params.permit(:from, :filter, :page).to_h.compact_blank.symbolize_keys
+      end
+
+      def schools_index_return_params
+        school_delete_return_params.except(:from)
+      end
 
       def site_params(param_form_key)
         params.expect(param_form_key => SchoolForm::FIELDS)
--- a/app/models/provider/school.rb
+++ b/app/models/provider/school.rb
@@ -15,6 +15,7 @@ class Provider::School < ApplicationRecord
   belongs_to :gias_school
 
   has_many :course_schools, class_name: "Course::School", inverse_of: :provider_school, dependent: :destroy
+  has_many :kept_courses, -> { kept }, through: :course_schools, source: :course
 
   # Provider schools whose GIAS record is still available (i.e. not closed).
   # Used to keep closed schools out of the rollover.
@@ -70,5 +71,9 @@ class Provider::School < ApplicationRecord
       gias_school.county,
       gias_school.postcode,
     ].compact_blank.join(join_on_separator)
+  end
+
+  def courses_count
+    kept_courses.size
   end
 end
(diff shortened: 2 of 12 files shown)


gms-gs:
[Update Schools list content and styling](https://trello.com/c/BK2j87qv/1844-update-schools-list-content-and-styling)

app/views/publish/providers/schools/_school.html.erb
```diff
@@ -3,14 +3,22 @@
     <%= govuk_link_to school.location_name, publish_provider_recruitment_cycle_school_path(@provider.provider_code, school.recruitment_cycle.year, school.uuid) %>
 
     <%= render Publish::Schools::NewlyAddedTagComponent.new(school:) %>
+    <p class="govuk-body govuk-hint address"><%= school.full_address %></p>
```
kmcd, line 6: On this school, there's zero courses: http://publish.localhost:3001/publish/organisations/2B5/2026/schools

When I remove it, it returns an error saying you can't remove the school.

Image

&nbsp;

32 providers in 2026 have one school, 4 of them have no courses.

app/views/publish/providers/schools/_school.html.erb
```diff
@@ -3,14 +3,22 @@
...
   </th>
 
-  <td class="govuk-table__cell code">
-    <%= school.code %> <%= school.code == "-" ? "(dash)" : "" %>
+  <td class="govuk-table__cell courses-count">
+    <%= t("publish.providers.schools.index.courses_count", count: school.courses_count) %>
+  </td>
+  <td class="govuk-table__cell remove">
+    <%= govuk_link_to(
+      t("publish.providers.schools.index.remove"),
```
kmcd: Think this will fail an accessibility test.

Screen readers will need a unique accessible name like "Remove school XYZ"

app/views/publish/providers/schools/index.html.erb
```diff
@@ -25,14 +25,8 @@
...
-            <th class="govuk-table__header" scope="col"><%= t(".code") %></th>
-            <% if urn_required?(@recruitment_cycle.year.to_i) %>
-              <th class="govuk-table__header" scope="col">
-                <abbr class="app-!-text-decoration-underline-dotted" title="<%= t(".urn_abbr") %>">
-                  <%= t(".urn") %>
-                </abbr>
-              </th>
-            <% end %>
+            <th class="govuk-table__header" scope="col"><%= t(".courses_attached") %></th>
+            <th class="govuk-table__header" scope="col"><%= t(".remove") %></th>
```
kmcd, line 29: On http://publish.localhost:3001/publish/organisations/24X/2026/schools

Think the columns should be wider as "Remove School" link wraps over the line:

Image

app/views/publish/providers/schools/index.html.erb
```diff
@@ -25,14 +25,8 @@
...
             <th class="govuk-table__header" scope="col"><%= t(".name") %></th>
-            <th class="govuk-table__header" scope="col"><%= t(".code") %></th>
-            <% if urn_required?(@recruitment_cycle.year.to_i) %>
-              <th class="govuk-table__header" scope="col">
-                <abbr class="app-!-text-decoration-underline-dotted" title="<%= t(".urn_abbr") %>">
-                  <%= t(".urn") %>
-                </abbr>
-              </th>
-            <% end %>
+            <th class="govuk-table__header" scope="col"><%= t(".courses_attached") %></th>
```
kmcd, line 28: Are the column header and row contents OK?

e.g. "Number of courses attached" and "36 courses". Should it just be 36?

Screenshot 2026-08-26 at 14 43 12

app/models/provider/school.rb
```diff
@@ -15,6 +15,7 @@ class Provider::School < ApplicationRecord
   belongs_to :gias_school
 
   has_many :course_schools, class_name: "Course::School", inverse_of: :provider_school, dependent: :destroy
+  has_many :kept_courses, -> { kept }, through: :course_schools, source: :course
```
kmcd, line 18: On http://publish.localhost:3001/publish/organisations/3J3/2026/schools

There are 19 schools with 0 courses. 

Is the `.kept` scope hiding discarded courses?

app/controllers/publish/providers/schools_controller.rb
```diff
@@ -32,7 +35,7 @@ def destroy
           flash[:success] = "School removed"
           redirect_to publish_provider_recruitment_cycle_schools_path
```
kmcd, line 36: Does this need to handle `params[:from]` ?

kmcd (changes requested):

app/views/publish/providers/schools/index.html.erb
gms-gs: This is based off the designs, Kellie approved the designs so will keep as is I think.

app/views/publish/providers/schools/_school.html.erb
gms-gs: This is expected because it is the provider’s only school. The existing removal rules prevent a provider from having no schools, regardless of whether the school has courses attached.

app/views/publish/providers/schools/_school.html.erb
gms-gs: Good catch. I’ll add the school name as visually hidden text so each link has a unique accessible name

app/models/provider/school.rb
gms-gs: Yes, deliberately. The number represents active courses currently attached to the school. This also matches the removal logic, which only treats kept courses as preventing removal.

app/controllers/publish/providers/schools_controller.rb
gms-gs: On successful removal it should always return to the schools list because the removed school’s details page no longer exists. On failure, school_delete_path_with_return preserves from, filter, and page so Back and Cancel still return to the correct place.

app/views/publish/providers/schools/index.html.erb
gms-gs: Similar to the other one, this is based off the design and uses `govuk-grid-column-two-thirds` which aligns how we show other tables in the application. So will leave as is

kmcd (approved):
LGTM!

PR #6372, opened 2026-08-25
Add Solid Cache plumbing while keeping Redis as the live cache store

samjohnson-dev:
## Context

We want to move Find/Publish cache off Redis onto Solid Cache, following Apply: a `solid_cache_entries` table on the **primary** Postgres, not a second Azure database.

Prod Redis cache is a C1 (1GB) instance at ~140MB with no evictions, so we do not need `extra_databases` or a terraform-modules promotion. Redis stays the live `cache_store`. This PR only lays the table and dump hygiene down so a later PR can switch the store.

Cache blobs can include postcodes and addresses. Redis is not in Postgres dumps today; a table on the primary is. Sanitised dumps (review/local) must wipe the table.

- Add `solid_cache` and create `solid_cache_entries` on the primary database (Apply-shaped columns/indexes).
- Keep `config.cache_store = :redis_cache_store`. `config/cache.yml` is unused until cutover; `max_size` is 1GB to match the C1 SKU.
- Wipe `solid_cache_entries` in `db/scripts/sanitise.sql` (no-op if the table is missing). Block the table from analytics.
- Regenerated `docs/database-diagram.mmd` on migrate (adds `SolidCache::Entry`; also picks up stale `main` schema such as `GiasSchool.geo_location` and dropped `Course::School.site_code`).

No UI changes.

## Guidance to review

- Confirm Redis is still the live store and there is no `CACHE_DATABASE_URL` / `extra_databases` / `database: cache`.
- Migration and `schema.rb` should match Apply’s `solid_cache_entries` shape.
- Sanitise should delete cache rows and tolerate dumps from before the migration.
- The mermaid diff is mostly rails-erd reorder; the new entity is the cache table.

### DIFF
--- a/spec/db/scripts_spec.rb
+++ b/spec/db/scripts_spec.rb
@@ -17,6 +17,16 @@ RSpec.describe "db/scripts" do
     expect { run_script("sanitise") }.not_to raise_error
   end
 
+  it "wipes solid_cache_entries" do
+    store = ActiveSupport::Cache.lookup_store(:solid_cache_store)
+    store.write("sanitise-cache-probe", "secret")
+    expect(SolidCache::Entry.count).to be_positive
+
+    run_script("sanitise")
+
+    expect(SolidCache::Entry.count).to eq(0)
+  end
+
   it "integration_setup.sql runs cleanly against the current schema" do
     # The provider insert requires a recruitment cycle (recruitment_cycle_id is NOT NULL).
     create(:recruitment_cycle)
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
 ARG ZLIB_VERSION=1.3.2-r0
+ARG LIBEXPAT_VERSION=2.8.4-r0
 
 FROM ruby:3.4.10-alpine3.23 AS middleman
 ARG ZLIB_VERSION
+ARG LIBEXPAT_VERSION
 # TODO: remove zlib pin when ruby:3.4.8-alpine3.23 base image is refreshed.
-RUN apk add --no-cache zlib=${ZLIB_VERSION} libxml2
+RUN apk add --no-cache zlib=${ZLIB_VERSION} libexpat=${LIBEXPAT_VERSION} libxml2
 RUN apk add --update --no-cache npm git build-base
 
 COPY docs/Gemfile docs/Gemfile.lock /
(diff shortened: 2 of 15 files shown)


samjohnson-dev:
⚠️ Please don’t merge until a review app has been checked.

**What this does:** sets up a spare Postgres DB for Solid Cache later. **Live cache is still Redis** — nothing should feel different.

**Plan to check on a review app:**
1. Deploy succeeds (use `USE_DB_SETUP_COMMAND=true` on first deploy)
2. Pod stays up
3. Cache DB exists and has `solid_cache_entries` (empty is fine)

samjohnson-dev:
I have checked the tables exist on review app and are empty then a record is created in the right place 🥳

.rubocop.yml
```diff
@@ -22,5 +22,10 @@ AllCops:
   ParserEngine: parser_prism
   TargetRubyVersion: 3.4
 
+# Historical migrations may have obsolete disable comments; do not edit them to satisfy cops.
+Lint/RedundantCopDisableDirective:
+  Exclude:
+    - "db/migrate/**/*"
```
tomas-stefano: I think this is not needed. Any cop failed? Removed this block and ran bundle exec rubocop: 2285 files, no offenses

spec/db/scripts_spec.rb
```diff
@@ -17,6 +17,16 @@ def run_script(name)
...
   end
 
+  it "wipes solid_cache_entries" do
+    store = ActiveSupport::Cache.lookup_store(:solid_cache_store)
+    store.write("sanitise-cache-probe", "secret")
+    expect(SolidCache::Entry.count).to be_positive
+
+    run_script("sanitise")
+
+    expect(SolidCache::Entry.count).to eq(0)
```
tomas-stefano, line 27: This tests the gem, which the gem already tests. 

The thing that can regress here is the wiring the PR asks reviewers to confirm, and neither half is asserted. 

Worth pinning:

```ruby
  expect(SolidCache.configuration.connects_to).to be_nil
  expect(SolidCache::Entry.connection_db_config.name).to eq("primary")
```

That's what fails the day someone adds database: cache to cache.yml and expects an Azure database that isn't provisioned.

tomas-stefano (approved):
2 minor comments. After is fixed feel free to merge it.

.rubocop.yml
samjohnson-dev: I had errors when i removed this but will try again 🫡

PR #6376, opened 2026-08-27
Allow removing a school that still has courses, unless it is the last one

samjohnson-dev:
## Context

Providers could not remove a school if any courses used it, and the page did not say why or what to do next. They can now remove it when every attached course has another school. Removal is still blocked if this is the provider’s last school, or the only placement school on a course.

- Restyle the Publish school removal page: heading is “Remove [school name]”, with address and URN underneath
- If the school has no courses: “This school is not attached to any courses”
- If it has courses that also have other schools: warn how many will be detached and list them (name/code, funding, qualification, study mode, start date)
- If it is the only school on any of those courses: “You cannot remove…” page listing only those courses, with links to the course description tab (new tab)
- Success banner on the schools list is now “[School name] has been removed from your account”
- Support uses the same removal rule (no UI restyle)

## Guidance to review

- `bundle exec rspec spec/services/provider_schools/removal_spec.rb spec/system/publish/providers/schools/delete_school_spec.rb spec/requests/publish/providers/schools_spec.rb spec/requests/support/providers/schools_spec.rb spec/system/support/providers/schools/delete_school_spec.rb spec/components/publish/providers/schools/removal_courses_component_spec.rb`
- On Publish, try removing: a school with 0 courses; a school that is an extra school on a course; a school that is the only school on a course; the provider’s last school
- For the “only school on a course” case, add another placement school on that course, refresh, and check the red remove button appears


app/components/publish/providers/schools/removal_courses_component.html.erb
```diff
@@ -0,0 +1,12 @@
+<ul class="govuk-list govuk-list--bullet">
+  <% courses.each do |course| %>
+    <li>
+      <% if link_to_course? %>
+        <%= govuk_link_to course.name_and_code, course_path(course), new_tab: t(".opens_in_new_tab") %>
+      <% else %>
+        <h2 class="govuk-heading-s govuk-!-margin-bottom-1"><%= course.name_and_code %></h2>
```
kelliedesigner: Replace `<h2>` with `<p>`, as these list items shouldn't be classed as headings.

Suggest:
```
<li>
  <p class="govuk-!-margin-0 govuk-!-font-weight-bold">Biology (2LCR)</p>
  <span class="govuk-hint govuk-!-font-size-16">
    Fee‑paying, QTS with PGCE, Full time, September 2026
  </span>
</li>
```

app/components/publish/providers/schools/removal_courses_component.html.erb
```diff
@@ -0,0 +1,12 @@
+<ul class="govuk-list govuk-list--bullet">
+  <% courses.each do |course| %>
+    <li>
+      <% if link_to_course? %>
+        <%= govuk_link_to course.name_and_code, course_path(course), new_tab: t(".opens_in_new_tab") %>
```
kelliedesigner: Suggest wrapping the `<a>` in a `<p>`, and changing the hint to `<span>` to improve spacing.

Also add `govuk-!-font-weight-bold` to make the link bold and consist with the other unordered list.

Suggest:

```
<li>
  <p class="govuk-!-margin-0">
    <a class="govuk-link govuk-!-font-weight-bold" ... >
      Primary (SEND) (Z501) (opens in a new tab)
    </a>
  </p>
  <span class="govuk-hint govuk-!-font-size-16">Fee‑paying, QTS, Full time, September 2026
  </span>
</li>
```

kelliedesigner (changes requested):

app/components/publish/providers/schools/removal_courses_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<ul class="govuk-list govuk-list--bullet">
+  <% courses.each do |course| %>
+    <li>
+      <% if link_to_course? %>
+        <p class="govuk-link govuk-!-font-weight-bold">
```
kelliedesigner: Class on the `<p>` should be `<govuk-body>`.

kelliedesigner: Add class `govuk-!-margin-0` to the `<p>` to reduce spacing between the course name and hint text.

kelliedesigner (changes requested):

app/views/publish/providers/schools/delete.html.erb
```diff
@@ -1,40 +1,59 @@
...
+      <p class="govuk-body">
+        <%= t(".attach_another_school", school_name:) %>
+      </p>
+      <%= render Publish::Providers::Schools::RemovalCoursesComponent.new(
+        courses: school_removal.sole_school_courses,
+        provider: @provider,
+        link_to_course: true,
+      ) %>
+    <% else %>
+      <h2 class="govuk-heading-m"><%= t(".are_you_sure") %></h2>
```
kmcd: On publish, there's a warning when you're about to delete a course:

Image

http://publish.localhost:3001/publish/organisations/1GT/2026/schools/cf58d81b-ca9b-4b83-8aed-b6deb4069968/delete

The same school in support doesn't have a warning: http://publish.localhost:3001/support/2026/providers/21683/schools/cf58d81b-ca9b-4b83-8aed-b6deb4069968/delete

Image

Is this an oversight or intentional?

app/services/provider_schools/removal.rb
```diff
@@ -54,5 +72,15 @@ def destroy_records!
       school.destroy!
       site&.destroy!
     end
+
+    def sole_school_course_ids
```
kmcd, line 93: OK and edge case within the same transition lock. Unlikely but not impossible ...

This course: "Geography (C411) at provider 4A3:" has two placement schools: Avanti House School and Avanti Fields School.

Within the same request transaction, two **different** requests could delete Avanti House School and Avanti Fields School. 

This would leave Geography (C411) with no school.

app/services/provider_schools/removal.rb
```diff
@@ -33,10 +33,28 @@ def only_school?
       provider.schools.one?
     end
 
+    # Kept courses this school is attached to. Discarded courses are omitted so
```
kmcd, line 37: An email is fired off when courses are updated. Couldn't find any email notifications in this PR ...

For example, a provider could remove a school from an accredited course and the accrediting body won't know about it.

kmcd, line 37: Do we need to fire off some audit logs and analytics jobs for deleting a school as well?

kmcd (changes requested):

app/views/publish/providers/schools/delete.html.erb
samjohnson-dev: Good point, i would have to ask Sarah. i think whether this is ok to have.

sarahcrack:
Amazing work @samjohnson-dev 🚀

Have tested the removal of:
1. a school with 0 courses attached
2. a school with multiple courses attached
3. a school that is the only school attached to a course
4. a school that is the last school on a provider's account

From a design perspective, all looks great to me! 👍🏻

app/views/publish/providers/schools/delete.html.erb
sarahcrack: really good spot @kmcd! Will have a look into this 😊

app/services/provider_schools/removal.rb
langesi: Yes and Yes. 

We still haven't switched fully to airbyte, so we depend on the analytics job to update bigquery. Once Airbyte data replication is the source for bigquery we can get rid of the analytics jobs for database record changes

langesi: I’m fairly certain that we do want to continue sending emails as before. However, I think this may have been overlooked, and there’s a potential risk of users receiving multiple emails that look very similar or several emails being sent at once.

I’d suggest raising this with the UCD team as a potential spam/volume risk, and I’ll flag it with them as well.

In my view, sending multiple emails shouldn’t be a blocker and could potentially be treated as a future improvement. For now, I think we should continue sending emails as we did previously, but the final decision sits with the UCD team.

app/services/provider_schools/removal.rb
```diff
@@ -54,5 +73,33 @@ def destroy_records!
...
+    # Lock kept courses this school is on, in id order, before the sole-school
+    # check. Two removals can otherwise each see two schools on the same course
+    # and both proceed, leaving it with none.
+    def lock_attached_courses!
+      course_ids = school.course_schools
+        .joins(:course)
+        .merge(Course.kept)
+        .distinct
+        .order(:course_id)
+        .pluck(:course_id)
```
inulty-dfe, line 84: Do we need to use `pluck`. Can't we pass the ActiveRecord Relation directly to the next as a subquery? then `order` and `distinct` are not needed

app/services/provider_schools/removal.rb
```diff
@@ -54,5 +73,33 @@ def destroy_records!
...
+
+      Course.where(id: course_ids).order(:id).lock.load if course_ids.any?
```
inulty-dfe, line 86: "Lock and load" 😂

app/views/publish/providers/schools/delete.html.erb
```diff
@@ -1,40 +1,59 @@
-<%= render PageTitle.new(title: "publish.providers.schools.delete") %>
+<% school_name = school.location_name %>
+<% cannot_remove = !school_removal.removable? %>
```
inulty-dfe: prefer not setting local variables in the template. I'd prefer them set as instance vars in the controller just to keep the views tidy

app/controllers/publish/providers/schools_controller.rb
```diff
@@ -33,8 +33,9 @@ def create
       def delete; end
 
       def destroy
+        school_name = school.location_name
         if school_removal.call
-          flash[:success] = "School removed"
+          flash[:success] = t(".removed", school_name:)
```
inulty-dfe: Is it worth adding a new line so the `t` picks up `school_name:` implicitly?

app/services/provider_schools/removal.rb
```diff
@@ -54,5 +73,33 @@ def destroy_records!
...
+    end
+
+    def sole_school_course_ids
+      @sole_school_course_ids ||= Course::School
+        .joins(:course)
+        .merge(Course.kept)
+        .where(course_id: school.course_schools.select(:course_id))
+        .group(:course_id)
+        .having("COUNT(*) = 1")
+        .pluck(:course_id)
```
inulty-dfe, line 100: Do we need to pluck the ids or can we pass around the scope?

inulty-dfe (commented):
I'm finding this Pr difficult to review. I think the main problem is `ProviderSchools::Removal`. It's a command and a view model at the same time.

app/services/provider_schools/removal.rb
samjohnson-dev: I think its useful to have the pluck here this would egt re-used as is used twice.

app/services/provider_schools/removal.rb
samjohnson-dev: agree distinct and order could go but pluck is more important for the locking in the two step query, one option is to use `.kept_courses` but then the lock would lock both rows rather than just course?

app/views/publish/providers/schools/delete.html.erb
samjohnson-dev: I dont mind this either way and realise more of a style query so happy either way? school and school_removal already set as helper methods so would probs jsut be setting on the empty delete action.

app/controllers/publish/providers/schools_controller.rb
samjohnson-dev: Yes it could be condensed down if preferred 🫡

config/analytics.yml
```diff
@@ -82,6 +82,13 @@ shared:
     - school_experience_required
     - school_experience_required_content
     - publish_without_schools_allowed
+  course_school:
+    - id
+    - course_id
+    - gias_school_id
+    - provider_school_id
+    - created_at
+    - updated_at
```
inulty-dfe: We need to tell data insights that they'll start seeing these tables before merging. Is it necessary to do this in this PR?

inulty-dfe (approved):

kelliedesigner (approved):
LGMT!

kmcd (approved):

PR #6377, opened 2026-08-28
Update onboarding docs with local setup prerequisites

kmcd:
## Context

`./bin/setup` didn't work out of the box for me. Updating the onboarding docs with my findings for future new starters.

A Prerequisites section with `asdf install` covering Ruby, Node and Caddy from `.tool-versions`. PostGIS needs brew.

Added `Caddyfile.dev`, run directly from `Procfile.dev`, because `Caddyfile` is gitignored.

## Guidance to review

Worth checking the Caddy section as I use `bin/dev`.

## Checklist

### DIFF
--- a/Caddyfile.example
+++ b/Caddyfile.example
@@ -1,0 +1,15 @@
+# Copy to `Caddyfile` (gitignored) before running `bin/dev`, which starts Caddy
+# from the Procfile. Routing is by host, so all three services share one Rails
+# process on port 3001.
+
+publish.localhost {
+	reverse_proxy localhost:3001
+}
+
+find.localhost {
+	reverse_proxy localhost:3001
+}
+
+api.localhost {
+	reverse_proxy localhost:3001
+}
--- a/guides/development-workflow.md
+++ b/guides/development-workflow.md
@@ -42,6 +42,17 @@
   - If a feature is not ready for usage, it should be hidden behind a `Settings.features.x` configuration.
   - If a feature needs to be turned on or off, it should be hidden behind a feature flag.
 
+### Review apps
+
+- Add the `deploy` label to a PR to deploy it as a review app. Without the label the deployment job is skipped.
+
+- The deployment runs after the lint, analytics, JavaScript and Rails test jobs pass, so a review app only
+  appears once CI is green. It redeploys on each new push to the branch.
+
+- Once deployment is successful, a comment is posted to the PR with the Publish, Find and API review app URLs.
+
+- Prototype branches use a different label and bypass the code quality checks — see [prototype branches](prototype-branches.md).
+
 ### Things to check before marking a PR as ready for review
 
 - Renaming and removing database columns requires multi-step PRs:
--- a/guides/setup-development.md
+++ b/guides/setup-development.md
@@ -4,6 +4,58 @@ Clone the repo:
 
     git clone git@github.com:DFE-Digital/publish-teacher-training.git
 
+## Prerequisites
+
+`./bin/setup` installs the application's own dependencies, but it assumes three
+things are already on your machine. Each stops the setup at a different point.
+
+If you would rather not install any of them, skip to [Using Docker](#using-docker)
+— the compose file brings its own Postgres, Ruby and Node.
+
+### Ruby 3.4.10 and Node 24.13.0
+
+- **Ruby** is pinned in `.ruby-version`, which the `Gemfile` reads. `bundle check`
+  fails on any other version.
+- **Node** is pinned by `package.json` (`engines: { node: "24.x" }`), so `yarn
+  install` refuses to run on anything else. That is the first thing `./bin/setup`
+  does, so the wrong Node stops it immediately, and it stops all four asset
+  watchers in `./bin/dev` as well.
+
+Install them however you normally manage versions. `.tool-versions` records both,
+so a version manager that reads it will pick them up; otherwise take the two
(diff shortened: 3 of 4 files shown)


guides/development-workflow.md
```diff
@@ -42,6 +42,17 @@
...
+### Review apps
+
+- Add the `deploy` label to a PR to deploy it as a review app. Without the label the deployment job is skipped.
+
+- The deployment runs after the lint, analytics, JavaScript and Rails test jobs pass, so a review app only
+  appears once CI is green. It redeploys on each new push to the branch.
+
+- Once deployment is successful, a comment is posted to the PR with the Publish, Find and API review app URLs.
+
+- Prototype branches use a different label and bypass the code quality checks — see [prototype branches](prototype-branches.md).
```
samjohnson-dev, line 54: 👏

guides/setup-development.md
```diff
@@ -4,27 +4,65 @@ Clone the repo:
...
 
-If `yarn -v` still shows Yarn 1, run:
+Only needed if you run `./bin/dev`, which starts Caddy from the `Procfile.dev` —
+foreman takes the whole stack down if the binary is missing.
 
 ```bash
-corepack enable
-corepack prepare yarn@4.18.0 --activate
-yarn -v
-yarn install --immutable
```
samjohnson-dev, line 24: I didnt try this and used the above with corepack, this was added in my most recent PR?

guides/setup-development.md
```diff
@@ -89,6 +127,11 @@ sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keyc
 
 The commands from the previous section will seed the database with some test data, but you must seed the database with a sanitised production dump to run the application locally using the personas.
 
+The personas page at `/personas` renders whether or not you have the dump, but
+the users it signs you in as (Anne, Susy, Mary) only exist in it. Until you have
```
samjohnson-dev, line 131: Are we missing one here? Colin is it?

guides/setup-development.md
```diff
@@ -4,27 +4,65 @@ Clone the repo:
...
-If `yarn -v` still shows Yarn 1, run:
+Only needed if you run `./bin/dev`, which starts Caddy from the `Procfile.dev` —
+foreman takes the whole stack down if the binary is missing.
 
 ```bash
-corepack enable
-corepack prepare yarn@4.18.0 --activate
-yarn -v
-yarn install --immutable
+brew install caddy
```
dcyoung-dev, line 45: This is also available through the `.tool-versions`. We usually recommend `asdf` for managing dependancies, which would handle Caddy.

Caddyfile.example
dcyoung-dev, line 1: How do you feel about renaming this to `Caddyfile.dev` and running it directly, rather than copying and renaming?

dcyoung-dev (commented):
I've added some comments. In an ideal world I'd like to see us able to clone a repo, run `asdf install` and then `bin/setup` which would in turn run install all app dependancies and run `bin/dev` which would spin up the server appropriately. 
This ideal mimics what a new Rails app would let us do 🤞 The main reason for this will be to help the Run State team switch between projects without having to think about setup.

inulty-dfe (approved):

guides/setup-development.md
kmcd: Done. Prerequisite docs specifies `asdf install` with `.tool-versions`. Should install Ruby, Node and Caddy in one step.

Caddyfile.example
kmcd: Good call. Renamed to `Caddyfile.dev` and `Procfile.dev`.

guides/setup-development.md
kmcd: Good spot - reverted.

guides/setup-development.md
kmcd: Good catch! Added Colin.

kmcd:
Thanks both. Pushed a commit addressing the comments.

@dcyoung-dev on the ideal flow — `asdf install` then `bin/setup` is now what the guide documents, and `bin/setup` already `exec`s `bin/dev` at the end, so a clean run does leave the server up. Two gaps remain between that and a stock Rails app:

- **PostGIS.** `config/database.yml` uses `adapter: postgis`, and asdf cannot install a Postgres extension, so `db:prepare` fails without `brew install postgis`. Documented rather than solved.
- **`bin/setup:25` is `system("yarn install")`, with no `!`** — the one step whose exit status is not checked. On the wrong Node, yarn refuses, setup carries on to bundler and `db:prepare`, and reports success with `node_modules` missing. That was what actually bit me. I have documented it here rather than widen a docs PR, but making it `system!` looks like the real fix if you agree — happy to raise it separately.

kmcd:
> I've added some comments. In an ideal world I'd like to see us able to clone a repo, run `asdf install` and then `bin/setup` which would in turn run install all app dependancies and run `bin/dev` which would spin up the server appropriately. This ideal mimics what a new Rails app would let us do 🤞 The main reason for this will be to help the Run State team switch between projects without having to think about setup.

Good suggestions - all done now.

`asdf install` then `bin/setup` brings up the server OK. Setup now checks and installs dependencies (Postgres, PostGIS, Caddy etc.)

`Caddyfile.dev` runs from `Procfile.dev` now, so no need to copy the example file.

The `.tool-versions` stuff is a bit messy with `asdf` and `make`. Left a note in the docs showing the command install the plugins (ruby, nodejs, caddy) needed for dev.

PR #6382, opened 2026-09-01
Fix location search grouping error on new school data model feature flag

tomas-stefano:
# Group the course_school location scope

Sorting a location search by fee or by newest course 500s the results page, which is what Sentry is
reporting as `PG::GroupingError: column "nearby_courses.distance_m" must appear in the GROUP BY clause`.
`schools_location_scope` — the location path behind `:course_publishing_uses_new_school_model` — annotates
each course with a distance read straight out of a derived table, as a raw column. Three orderings
(`newest_course`, `fee_uk_ascending`, `fee_intl_ascending`) add a `GROUP BY` of their own, and grouping by
`course.id` covers `course.*` by functional dependency but does not reach into the joined derived table.
So the two disagree and Postgres rejects the query. The legacy `sites_location_scope` never hit this: it
selects `MIN(ST_DistanceSphere(...))` under its own `GROUP BY`, so a second one merges harmlessly.

The same missing `GROUP BY` causes a second, quieter bug: `subjects_scope` joins a `has_many`, so a course
teaching two of the searched subjects comes back twice. The count beside the results is computed separately
with `.distinct.count(:id)`, so it keeps saying one — the page shows "1 course" above two identical cards.
Both bugs arrived together when the location filter moved from `course_site` to `course_school`, and both
are fixed by the same line. Worth noting for whoever picks up the Sentry issue: `EmailAlertMailer` builds
its "see all results" link with `order=newest_course`, and alert subscribers search by location, so those
emails send people straight onto the broken combination.

The fix restores the grouping contract the site-based scope always had:
`GROUP BY course.id, provider.id, provider.provider_name, nearby_courses.distance_m`. It costs nothing in
rows — the derived table already emits exactly one row per course — and it leaves the index work from the
`ST_DWithin` pruning where it was, inside the derived table and ahead of the join. Tests were written
first and each was watched fail with the reported error before the fix landed: service specs for the three
orderings, the duplicate listing, and count-versus-listing agreement, plus a browser spec that searches a
location, picks each sort and reads back the rendered cards. I also checked all seven sort options against
a location, with and without a subject filter, so the orderings that were previously fine stay fine under
the new grouping.

```
                  ┌─ nearby_courses (derived table) ──────┐
                  │ GROUP BY course_school.course_id      │  already one row
                  │ MIN(ST_Distance(...)) AS distance_m   │  per course
                  └───────────────────┬───────────────────┘
                                      │ INNER JOIN
                                      ▼
     subjects ──── JOIN ──────────► course ◄──── JOIN ──── provider
     one row per matching
     subject → duplicate cards ✗ (2)

     SELECT course.*, provider.provider_name,
            nearby_courses.distance_m / 1609.344      ← raw column, not aggregated
     GROUP BY course.id, provider.id, provider.provider_name
              └── added by the newest_course / fee_uk / fee_intl orderings;
                  distance_m is absent → PG::GroupingError ✗ (1)
```

### DIFF
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -379,6 +379,7 @@ module Courses
       @scope
         .joins(nearby_courses)
         .select("course.*, provider.provider_name, (nearby_courses.distance_m / 1609.344) AS minimum_distance_to_search_location")
+        .group("course.id, provider.id, provider.provider_name, nearby_courses.distance_m")
     end
 
     def default_ordering_scope
--- a/app/services/courses/query.rb
+++ b/app/services/courses/query.rb
@@ -359,6 +359,14 @@ module Courses
     # ST_Distance/ST_DWithin take the trailing `false` to force sphere maths, so
     # results match the legacy ST_DistanceSphere path. Distance is returned in
     # miles, preserving the minimum_distance_to_search_location contract.
+    #
+    # The GROUP BY carries the same contract sites_location_scope has always had,
+    # and both of its jobs matter. It collapses the subjects_scope join, which
+    # otherwise repeats a course once per matching subject, and it keeps the
+    # distance column groupable for the orderings that add a GROUP BY of their own
+    # (newest_course, fee_uk_ascending, fee_intl_ascending) - without it those
+    # combinations raise PG::GroupingError. Grouping costs nothing here: the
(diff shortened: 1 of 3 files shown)


langesi (approved):
Thanks @tomas-stefano

PR #6386, opened 2026-09-01
Canonical schools distance queries

tomas-stefano:
## Context

Find had two queries left on the legacy `Site` / `SiteStatus` model: `NearestSchoolQuery`, which drives the distance on the course page, and `SchoolDistancesQuery`, which drives the `?debug` panel. Find's search had already moved to the canonical `Course::School -> GiasSchool` chain behind `:course_publishing_uses_new_school_model`, so with the flag on the results card and the course page for the same course were computed from two different data sources — and two different divisors, 1609.344 against 1609.34.

The visible consequence was a 500. Opening a course with a `?location=` param raised `NoMethodError` for any course with `Course::School` rows but no legacy `Site`: the query inner-joined `course_site -> site`, came back empty, and `.first.distance_to_search_location` was called on nil. Both queries now have a canonical path gated on the same flag, mirroring the `sites_`/`schools_` split in `Courses::Query`, and the shared SELECT lives in one module so the two can't drift and quietly disagree about the same school's distance.

The debug tooling moved with them: the school link now uses `Provider::School#uuid` (the old `site_uuid` fallback built a link that 404s in Publish, since that route resolves `provider.schools.find_by!(uuid:)`), and "Nearest Site" became "Nearest School". The course page no longer renders "0 miles from London" when there's no school to measure from. Not addressed: `Find::PlacementsController` is still on `site_statuses`, so a canonical-only course links to an empty placements page — pre-existing, worth a follow-up.

## URLs to test

Flag on (`FeatureFlag.activate(:course_publishing_uses_new_school_model)`), against a 2026 course with `Course::School` rows and no `SiteStatus`:

- `/results?location=London&latitude=51.5074&longitude=-0.1278&radius=10` — card distance
- `/course/:provider_code/:course_code?location=London&latitude=51.5074&longitude=-0.1278` — should render the distance, and match the card above (previously 500)
- `/results?location=London&latitude=51.5074&longitude=-0.1278&radius=10&debug=true` — "Nearest School" column, school link resolves in Publish, per-course distances panel
- Null a `GiasSchool`'s `latitude`, then reload the course page URL — loads with the funding hint instead of "0 miles"

### DIFF
--- a/app/services/courses/canonical_school_distance.rb
+++ b/app/services/courses/canonical_school_distance.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Courses
+  # Shared SQL for the two canonical course_school -> gias_school distance queries.
+  #
+  # NearestSchoolQuery and SchoolDistancesQuery differ only in how many rows they
+  # keep per course, so the distance maths and the school columns live here. If the
+  # two drifted apart, the course page and the ?debug panel would quietly disagree
+  # about how far away the same school is.
+  module CanonicalSchoolDistance
+    # geo_location is a stored generated column, so it is NULL whenever either
+    # coordinate is missing - one guard covers both, and the partial GiST index
+    # index_gias_school_on_geo_location covers the guard.
+    GEOCODED_SCHOOL = "gias_school.geo_location IS NOT NULL"
+
+  private
+
+    # ST_Distance takes the trailing `false` to force sphere maths, matching the
+    # legacy ST_DistanceSphere path and Courses::Query#schools_location_scope;
+    # metres are converted with the same 1609.344 the results page uses, so a
+    # course's distance reads the same on the results card and on its own page.
+    #
+    # The CASE reproduces Provider::School#location_name in SQL.
+    def school_columns_sql
+      <<~SQL.squish
+        course.id AS course_id,
+        course.*,
+        provider_school.uuid AS provider_school_uuid,
+        CASE
+          WHEN provider_school.site_code = #{Course.connection.quote(Provider::School::MAIN_SITE_CODE)}
+          THEN gias_school.name || ' (Main Site)'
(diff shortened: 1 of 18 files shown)


tomas-stefano:
https://trello.com/c/69FqyNtr/1941-school-data-migrate-find-nearest-school-distance-queries

app/services/courses/canonical_school_distance.rb
```diff
@@ -0,0 +1,46 @@
...
+        course.*,
+        provider_school.uuid AS provider_school_uuid,
+        CASE
+          WHEN provider_school.site_code = #{Course.connection.quote(Provider::School::MAIN_SITE_CODE)}
+          THEN gias_school.name || ' (Main Site)'
+          ELSE gias_school.name
+        END AS location_name,
+        gias_school.latitude,
+        gias_school.longitude,
+        ST_Distance(gias_school.geo_location, #{search_point}, false) / 1609.344 AS distance_to_search_location
```
samjohnson-dev: could this not be a constant and also it should be shared in the `app/services/courses/query.rb` file

samjohnson-dev: actually this number is shared all over the app at least 6 places?

samjohnson-dev (dismissed):

app/services/courses/canonical_school_distance.rb
tomas-stefano: Done! [a84898c](https://github.com/DFE-Digital/publish-teacher-training/pull/6386/commits/a84898c18e82de796d0a652429a7cc462de1bc5c)

kmcd: Confirmed - looks good now 👍

kmcd (approved):
LGTM!

PR #6387, opened 2026-09-01
Canonical placement schools

tomas-stefano:
## Context

Find's placement-schools list was the last page reading schools through the legacy `course_site -> site` chain. One decorator method fed it — `CourseDecorator#preview_site_statuses` — rendered by one partial, `shared/courses/_placements.html.erb`, which is shared by both `Find::PlacementsController` and `Publish::Courses::SchoolPlacementsController`. So a course with `Course::School` rows and no `SiteStatus` listed nothing on either page. That was the gap left by the distance-queries branch this stacks on: the course page shows a distance and a "View list of school placements" link, and the linked page was empty.

The canonical reader is flag-gated on `:course_publishing_uses_new_school_model`, like the rest of the remodel. It renders the GIAS name rather than `Provider::School#location_name` (which appends " (Main Site)", a Publish-internal affordance) and the address via `Provider::SchoolDecorator`, which gained a smart-quoted `full_address` so "St Mary's" keeps its curly apostrophe as it does today. Schools are de-duplicated on `gias_school_id`, since `course_school` is unique on `(course_id, provider_school_id)` and a course can reach one GIAS school through two site codes.

**One behaviour change to agree to, not just review:** `course_school` has no `status`/`publish`/`vac_status` column, so the canonical list cannot reproduce the legacy `new_or_running` filter. Under the flag, a school whose legacy `SiteStatus` was suspended now appears. A test asserts both paths side by side so it stays visible. Also of note: the existing preload had never worked — `preview_site_statuses` re-queried through a scope and discarded it — so the new reader walks the loaded association, the preload is lifted into a shared concern, and Publish's controller (which preloaded nothing at all) gets it too. Study sites are untouched and stay on `Site`.

## URLs to test

Flag on (`FeatureFlag.activate(:course_publishing_uses_new_school_model)`), on a 2026 course with `Course::School` rows and no `SiteStatus`:

- `/course/:provider_code/:course_code/placements` — lists every attached school once, name and address, where it previously listed nothing
- `/publish/organisations/:provider_code/2026/courses/:course_code/placements` — the same list, same order
- Attach two `Provider::School`s sharing one `GiasSchool` to the course, reload either — it appears once
- Set `provider.selectable_school = false` — Find 404s and the course-page link disappears; Publish preview still serves the page directly, unchanged
- Check the course page's study-sites block still renders from `Site`


tomas-stefano:
https://trello.com/c/EpTE3clU/1942-school-data-migrate-the-find-placement-school-list

app/views/shared/courses/_placements.html.erb
```diff
@@ -2,18 +2,28 @@
...
-  <% course.preview_site_statuses.each do |site_status| %>
-    <li>
-      <strong><%= site_status.site.location_name %></strong>
-      <br>
-      <%= site_status.site.decorate.full_address %>
-    </li>
+  <% if FeatureFlag.active?(:course_publishing_uses_new_school_model) %>
+    <% course.preview_placement_schools.each do |school| %>
+      <li>
+        <strong><%= school.gias_school.name %></strong>
```
kmcd, line 15: Looks like this introduces an N+1 query ...

This course: `6B1/T998` has 3,012 schools which produces 3,016 SQL queries.

Perhaps `preview_placement_schools` doesn't preload the `gias_school` association?

If so, adding this to `school_preloads` should fix it: `{ schools: [{ provider_school: :gias_school }] }`

app/controllers/publish/courses/school_placements_controller.rb
```diff
@@ -11,8 +11,18 @@ def index
...
     private
 
+      # The shared placements partial walks a course's schools, so preload them:
+      # this controller had no includes at all and issued a query per school.
       def course_to_authorise
-        @course_to_authorise ||= provider.courses.find_by!(course_code: params[:code])
+        @course_to_authorise ||= provider.courses.includes(**school_preloads).find_by!(course_code: params[:code])
+      end
+
+      def school_preloads
```
kmcd: Looks like this is duplicated with: [app/controllers/concerns/find/course_school_preloads.rb](https://github.com/DFE-Digital/publish-teacher-training/blob/292c42337559a3ed951b5f3bf2a6e48ce4ce9d06/app/controllers/concerns/find/course_school_preloads.rb)

kmcd (changes requested):

app/controllers/publish/courses/school_placements_controller.rb
tomas-stefano: Done

app/views/shared/courses/_placements.html.erb
tomas-stefano: Done

kmcd (approved):
LGTM! 👍

PR #6388, opened 2026-09-01
Training partner find links

tomas-stefano:
## Context

Research with accredited providers found they need more information about their training partners' courses in order to oversee that those courses are accurate. Today the accredited provider's view of a partner's courses is a dead end: the course names are not links on that page, so there is no way to see a course as a candidate would.

This adds a "View course" column to that list. Each row links to the course on Find when there is a live page to point at, and is left empty otherwise; the column disappears entirely when nothing in the list is live. The same page also now applies the existing "hide a value that is identical across every course" rule to the Course information column, which it had been skipping and always showing all four values. Most partners have a single ratified course, so that list will often show no course information at all — that is the rule working as specified.

The live-vs-preview condition was spread across the course show page's button panel partials, so it is extracted into Courses::PublishRules::LiveOnFind and shared by both call sites; the show page's behaviour is unchanged. The uniformity rule was already written out twice, so it is lifted to Publish::CourseList.visible_course_information_fields and all three callers point at it. Deciding whether a row is live reads site statuses, so those are preloaded for this page only, and a request spec pins the page at a constant number of queries — Bullet does not raise in test and would let a per-row query through unnoticed.

## URLs to test

- `/publish/organisations/{ACCREDITED_CODE}/2026/training-partners/{PARTNER_CODE}/courses` — the new column and the course information rule
- `/publish/organisations/{PROVIDER_CODE}/2026/courses` — the provider's own list, which must be unchanged
- `/publish/organisations/{PROVIDER_CODE}/2026/courses/{COURSE_CODE}` — the course page's View live course / Preview course link, also unchanged

### DIFF
--- a/app/assets/stylesheets/publish/components/_table.scss
+++ b/app/assets/stylesheets/publish/components/_table.scss
@@ -12,9 +12,9 @@
 
 .app-table--courses {
   width: 100%;
-  // Fixed layout keeps the Course, Course information and Status columns at an
-  // equal third each, so long course names wrap inside their column instead of
-  // pushing the others out. Adding or dropping a column switches to auto below.
+  // Fixed layout with no widths declared divides the table equally between
+  // whatever columns are present — thirds for three, quarters for four — so long
+  // course names wrap inside their column instead of pushing the others out.
   table-layout: fixed;
 
   // Bottom-align the headers so a wrapped "Course information" header lines up
@@ -51,12 +51,12 @@
   }
 }
 
-// Once the column count is no longer three, equal thirds stop making sense:
-// let the Course column take the remaining width and shrink the narrow columns
-// to their content. This covers the course information column being dropped and
-// the View course column being added, together or separately.
-.app-table--courses--no-information,
-.app-table--courses--view-course {
+// Down to two columns, an equal half is far more room than a status tag needs,
+// so this one case drops out of fixed layout: the Course column takes the width
+// and Status shrinks to its content. Every other combination keeps equal
+// columns, which is what the design asks for and what stops a link wrapping
+// inside a column squeezed to its narrowest word.
+.app-table--courses--no-information:not(.app-table--courses--view-course) {
   table-layout: auto;
 
   .app-table--courses__status {
@@ -64,10 +64,4 @@
 
     white-space: nowrap;
   }
-}
-
-// No nowrap here: "View live course" is long enough that it should be allowed
-// to wrap on a narrow viewport rather than force the table wider.
-.app-table--courses--view-course .app-table--courses__view-course {
-  width: 1%;
 }
--- a/app/assets/stylesheets/publish/components/_table.scss
+++ b/app/assets/stylesheets/publish/components/_table.scss
@@ -12,8 +12,9 @@
 
 .app-table--courses {
   width: 100%;
-  // Fixed layout keeps the three columns at an equal third each, so long course
-  // names wrap inside their column instead of pushing the others out.
+  // Fixed layout keeps the Course, Course information and Status columns at an
+  // equal third each, so long course names wrap inside their column instead of
+  // pushing the others out. Adding or dropping a column switches to auto below.
   table-layout: fixed;
 
   // Bottom-align the headers so a wrapped "Course information" header lines up
@@ -50,9 +51,12 @@
   }
 }
 
-// With no course information at all, the column is dropped: let the Course
-// column take the remaining width and shrink Status to its content on the right.
-.app-table--courses--no-information {
+// Once the column count is no longer three, equal thirds stop making sense:
+// let the Course column take the remaining width and shrink the narrow columns
+// to their content. This covers the course information column being dropped and
+// the View course column being added, together or separately.
+.app-table--courses--no-information,
+.app-table--courses--view-course {
   table-layout: auto;
 
   .app-table--courses__status {
@@ -60,4 +64,10 @@
 
     white-space: nowrap;
   }
+}
+
+// No nowrap here: "View live course" is long enough that it should be allowed
+// to wrap on a narrow viewport rather than force the table wider.
+.app-table--courses--view-course .app-table--courses__view-course {
+  width: 1%;
 }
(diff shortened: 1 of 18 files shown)


tomas-stefano:
https://trello.com/c/yWxQIlIA/1838-update-training-partners-course-list-page

app/assets/stylesheets/publish/components/_table.scss
```diff
@@ -50,9 +51,12 @@
...
 
-// With no course information at all, the column is dropped: let the Course
-// column take the remaining width and shrink Status to its content on the right.
-.app-table--courses--no-information {
+// Once the column count is no longer three, equal thirds stop making sense:
+// let the Course column take the remaining width and shrink the narrow columns
+// to their content. This covers the course information column being dropped and
+// the View course column being added, together or separately.
+.app-table--courses--no-information,
+.app-table--courses--view-course {
```
kmcd: Image

The columns don't look correctly aligned here: http://publish.localhost:3001/publish/organisations/S14/2026/training-partners/1GT/courses

Same here: http://publish.localhost:3001/publish/organisations/S14/2026/training-partners/1MH/courses

app/assets/stylesheets/publish/components/_table.scss
```diff
@@ -61,3 +65,9 @@
     white-space: nowrap;
   }
 }
+
+// No nowrap here: "View live course" is long enough that it should be allowed
+// to wrap on a narrow viewport rather than force the table wider.
+.app-table--courses--view-course .app-table--courses__view-course {
+  width: 1%;
```
kmcd: Think this is causing the links to wrap onto new lines. 

See the image above.

app/controllers/publish/providers/training_partners/courses_controller.rb
```diff
@@ -14,11 +15,21 @@ def training_partner
...
+        # schools for the few courses support has exempted. The provider's own
+        # course list never asks, so the preload belongs here rather than in the
+        # query. preload rather than includes: the relation carries a custom
+        # SELECT and a lateral join that eager_load would have to fold into.
         def fetch_courses
           Publish::Courses::Query.call(
             provider: training_partner,
             params: { accredited_provider: provider.provider_code },
-          ).map(&:decorate)
+          ).preload(:site_statuses, :schools).map(&:decorate)
```
kmcd: Not sure if we need `:schools` here. 

In app/services/courses/publish_rules/live_on_find.rb 👇 

```ruby
content_status.to_s == "published" && 
course.recruitment_cycle_year.to_i <= Find::CycleTimetable.current_year 
&& (course.is_running? || course.without_employing_school?)
```

Does this mean `:schools` is used if there are support exempted courses?

I'm not seeing any exempted courses locally ...

app/services/courses/publish_rules/live_on_find.rb
```diff
@@ -0,0 +1,30 @@
...
+# The year comes off the course's already-loaded cycle and is compared against
+# the timetable rather than through Course#next_recruitment_cycle?, which looks
+# the current cycle up from the database on every call — once per row on a
+# course list.
+module Courses
+  module PublishRules
+    class LiveOnFind
+      def self.applies?(course, content_status:)
+        content_status.to_s == "published" &&
+          course.recruitment_cycle_year.to_i <= Find::CycleTimetable.current_year &&
```
kmcd: Looking at: http://publish.localhost:3001/publish/organisations/S14/2025/training-partners/146/courses

There's a link to http://find.localhost:3001/course/146/36VP

Seems that a 2025 row links to the 2026 course with the same code.

I haven't imported the production DB since before the rollover. Is this a local data issue?

Not sure because Find looks up using only RecruitmentCycle.current course code ...

app/components/publish/courses/table_component.rb
```diff
@@ -24,6 +26,31 @@ def course_information_column?
...
+      # not a link and Find is the only way to see the course. Dropped when
+      # nothing in the list is live, rather than heading a column of empty cells.
+      def view_course_column?
+        @view_course_column && courses.any? { |course| live_on_find?(course) }
+      end
+
+      # Reads the content status from the row's computed column rather than
+      # Course#content_status, which would run the enrichment service per row.
+      def live_on_find?(course)
+        @live_on_find.fetch(course.id) do
```
kmcd, line 39: I can see this course on Find: http://find.localhost:3001/course/2KM/Z061

But this link isn't available on Publish:  http://publish.localhost:3001/publish/organisations/28E/2026/training-partners/2KM/courses

Image

kmcd (changes requested):

app/assets/stylesheets/publish/components/_table.scss
tomas-stefano: Good catch — the 4-column table was matching the table-layout: auto escape, where width: 1% means "shrink to the narrowest word". 

Fixed in 5dcefb9

app/assets/stylesheets/publish/components/_table.scss
tomas-stefano: Fixed in https://github.com/DFE-Digital/publish-teacher-training/commit/5dcefb97249742cf48b06e341cd35d5908078d99

app/controllers/publish/providers/training_partners/courses_controller.rb
tomas-stefano: You're right — narrowing the rule (below) means liveness reads only the content_status column, so both preloads are now dead; dropped in fe0824f.

 `publish_without_schools_allowed` is set on zero courses in the DB anyway, so :schools was buying nothing regardless.

app/services/courses/publish_rules/live_on_find.rb
tomas-stefano: The cycle comparison is now ==, so that page drops the column entirely. fixed in 8fa18bf

app/components/publish/courses/table_component.rb
tomas-stefano: My rule required is_running?, which Find never checks — `Find::CoursesController#show` renders any published course. Fixed in 8fa18bf; that page goes 89 → 90 links

kmcd (approved):
LGTM!

PR #6402, opened 2026-09-03
Show certain courses from previous cycle in find

gms-gs:
## Context

Apply has links to Find courses that start after September. Once we move into a new recruitment cycle, `/course/:provider_code/:course_code` looks up the current-cycle course. That can show the rolled-over version by mistake, or 404 if the course was not rolled over and published again.

This change lets candidates and support users view those previous-cycle courses on Find through a cycle-specific URL, until the course start date. After that, the page is no longer available.

This only applies to previous-cycle courses from **2026 onwards**. Enabling it for 2025 would break the course page because of enrichment format changes.

- Add `/course/:provider_code/:course_code/cycle/:cycle_year` for viewing a course in a specific cycle
- Show the page only when all of these are true:
  - the course is from the immediately previous cycle
  - the cycle year is 2026 or later
  - the course starts after 30 September of that cycle
  - the start date has not yet passed
  - the course is published
- Show a warning banner: the course is from a previous cycle, it is for information only, and it cannot be applied to
- Hide apply and save actions on that page
- Point `CourseDecorator#find_url` at the cycle URL for eligible previous-cycle courses, so Apply links do not hit the current-cycle page
- Leave existing current-cycle course pages unchanged

## Guidance to review

- `bundle exec rspec spec/services/find/previous_cycle_course_spec.rb spec/requests/find/previous_cycle_courses_spec.rb spec/decorators/course_decorator_spec.rb spec/controllers/find/courses_controller_spec.rb`
- On Find, with 2027 as the current cycle, try:
  - a previous-cycle course that starts after 30 September and has not started yet, for example `/course/1EP/R554/cycle/2026`
  - the same course after its start date (should 404)
  - a previous-cycle September-start course (should 404)
  - a 2025 previous-cycle course (should 404)
  - a normal current-cycle course page (should still work, with apply)

Without Link
Screenshot 2026-09-08 at 12 31 40

With Link
Screenshot 2026-09-08 at 12 38 10

## Checklist


gms-gs:
[Show certain courses from previous cycle in find](https://trello.com/c/d8JBvZfZ/1669-show-certain-courses-from-previous-cycle-in-find)

spec/requests/find/previous_cycle_courses_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    it "leaves the current-cycle course page unchanged" do
+      get find_course_path(current_course.provider.provider_code, current_course.course_code)
+
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to include("Geography")
+      expect(response.body).not_to include("This is a course from a previous year")
+      expect(response.body).to include("Apply for this course")
+    end
+
+    it "does not serve a current-cycle course through the cycle path" do
```
langesi, line 79: Could you maybe add the other cases as well?

Specifically I’d like us to lock down the behaviour when:

1. the course exists only in the previous cycle, the normal /course/:provider/:course URL should 404, while /cycle/2026 should return it if it's supposed to
2. the course exists only in the current cycle, the normal URL should return it and the previous-cycle URL should 404.

langesi (approved):
Thanks @gms-gs ! Got only one comment on specs

spec/requests/find/previous_cycle_courses_spec.rb
gms-gs: This is done! good catch

PR #6404, opened 2026-09-04
Bulk update schools - part 1 (options)

tomas-stefano:
## Context -  Bulk update placement schools (1/3) — the options page

Large providers edit the same school course by course. This is the first of three
stacked PRs adding a bulk path.

**The attach page stops writing.** It records what was ticked into a draft and asks
which courses the change is for. `Only this course` is today's behaviour; anything
wider is queued.

```mermaid
flowchart LR
    A["/schools<br/>tick schools"] -->|unchanged| W["write this course"]
    A -->|changed| B["/schools/bulk-update/:state_key<br/>which courses?"]
    B -->|only this course| W
    B -->|wider scope| Q[["BulkUpdateCourseSchoolsJob"]]
    W --> D["Basic details"]
    Q --> D
```

The draft lives in `Rails.cache` for 24h behind a `state_key` in the URL — the
session is a cookie store and cannot hold hundreds of UUIDs. It carries the
baseline as well as the selection, so the diff applied is the diff shown. A key
that no longer resolves redirects back to `/schools`.

`BulkUpdate::Scope` is the single source of the radio vocabulary — label *and*
matching courses — so what was offered, shown and written cannot drift.

### Test

`…/courses/{code}/schools` on any provider with several courses.

- [ ] Button reads **Continue to choose which courses to apply this change to**
- [ ] Tick a school → options page; nothing written yet (check Basic details in another tab)
- [ ] Options match the course: funding always; secondary only if secondary; subject (Primary/Further education name their level); All courses
- [ ] Nothing pre-selected; submitting blank shows an error
- [ ] **Only this course** → Basic details, that course changed, no others
- [ ] A wider scope → Basic details, banner counts the courses, others changed too
- [ ] Back from the options page keeps the ticks
- [ ] Reuse a spent `state_key` → back to `/schools` with an expiry message

### Notes

- No feature flag: the three branches merge in order.
- **Card 1's "remove the unpublished changes screen" needed no code** — the schools
  section never used that interstitial. `confirm_live_changes_spec.rb` already
  covered it; extended to the new path.
- Two known issues are fixed in PR 3, not here: `Apply` swallows per-course
  failures, and the exclusion query is unbounded.


tomas-stefano:
https://trello.com/c/4Yc9e86i/1710-add-bulk-update-options-page-to-placement-schools

app/services/publish/schools/bulk_update/apply.rb
```diff
@@ -0,0 +1,76 @@
...
+      private
+
+        attr_reader :courses, :added_uuids, :removed_uuids
+
+        def apply_to(course)
+          UpdateCourseSchoolsService.call(
+            course:,
+            school_uuids: school_uuids_for(course),
+            raise_on_missing_provider_schools: false,
+            notify: false,
```
kmcd, line 60: Currently providers receive an email when updating a single course.

Do we need a digest email for bulk updates?

Otherwise providers won't know about school changes applied to more than one course.

app/services/publish/schools/bulk_update/repository.rb
```diff
@@ -0,0 +1,35 @@
...
+    module BulkUpdate
+      # Where a bulk update's draft lives between requests: Rails.cache, keyed by
+      # a state key the URL carries, exactly as the add-course wizard does
+      # (CourseWizard::Repositories::Course).
+      #
+      # The provider, cycle and course are part of the key rather than the
+      # payload, so a state key issued for one course cannot resolve against
+      # another - which is what makes Draft.find safe to call with whatever the
+      # URL happens to hold.
+      class Repository < DfE::Wizard::Repository::Cache
```
inulty-dfe: Interesting that we're using a feature from the DfE::Wizard outside of a full wizard

app/services/publish/schools/bulk_update/repository.rb
```diff
@@ -0,0 +1,35 @@
...
+        def initialize(course:, state_key:, expires_in:, cache: Rails.cache)
+          super(
+            cache:,
+            key: self.class.cache_key(course:, state_key:),
+            expires_in:,
+          )
+        end
+
+        def self.cache_key(course:, state_key:)
```
kmcd: **N.B.** This could be irrelevant as I think we should be using ActiveRecord instead of the cache here.

This creates a bug in development:
* `rm tmp/caching-dev.txt`
*  restart dev server
* go to http://publish.localhost:3001/publish/organisations/12L/2026/courses/W280/schools 
* untick a school, press Continue - you're now stuck in a loop.

Caused by development cache_store set to `:null_store`

bin/rails dev:cache fixes it. 

Not sure what the remedy is: change dev caching or add some docs?

app/views/publish/courses/schools/bulk_updates/edit.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        <%= render CaptionText.new(text: course.name_and_code) %>
+        <%= page_title %>
+      </h1>
+
+      <%= f.govuk_radio_buttons_fieldset :scope,
+            legend: { text: t("publish.courses.schools.bulk_update.courses_legend"), size: "m", tag: "h2" } do %>
+        <% @form.scopes.each_with_index do |scope, index| %>
+          <%= f.govuk_radio_button :scope, scope.token,
+                label: { text: scope.label },
+                hint: scope.hint.present? ? { text: scope.hint } : nil,
```
kmcd: Low priority but ... why is this `nil` ?

Should it be `govuk-!-font-size-16` same as: https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/views/publish/schools/_checkbox_list.html.erb#L28

config/locales/en/publish/courses/schools/bulk_update.yml
```diff
@@ -0,0 +1,29 @@
...
+              blank: Select what courses you want to apply this change to
+              inclusion: Select what courses you want to apply this change to
+  publish:
+    courses:
+      schools:
+        bulk_update:
+          expired: Your school selection has expired. Select the schools you want again.
+          heading: Updating %{schools}
+          courses_legend: What courses do you want to apply this change to?
+          continue: Continue to view the courses that will be updated
```
kmcd, line 17: The wording here is a bit off: 

http://publish.localhost:3001/publish/organisations/12L/2026/courses/W280/schools

Image

When you select only one course, this button is a bit misleading.

config/locales/en/publish.yml
```diff
@@ -40,6 +40,7 @@ en:
         unsalaried: Placement
         select_all: Select all schools
         show_all: Show all schools
+        continue_to_bulk_update: Continue to choose which courses to apply this change to
```
kmcd, line 43: The wording here is a bit off: http://publish.localhost:3001/publish/organisations/12L/2026/courses/W280/schools

Seems a bit clunky ... Think @RoseMackworth / @kelliedesigner / @sarahcrack  might have suggestions here.

Image

app/services/publish/schools/bulk_update/draft.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+module Publish
+  module Schools
+    module BulkUpdate
```
kmcd: Don't think we should use the cache here instead of ActiveRecord.

Big change I know ...

kmcd (changes requested):

app/services/publish/schools/bulk_update/apply.rb
kelliedesigner: Can you speak to Simon and Sam about this, as I've been having a separate conversation with them about disabling notifications when changes are made to schools.

config/locales/en/publish/courses/schools/bulk_update.yml
kelliedesigner: It's a fair point, but I think this is fine to proceed with for the first iteration.

config/locales/en/publish.yml
kelliedesigner: It's a fair point, but I think this is fine to proceed with for the first iteration.

app/services/publish/schools/bulk_update/draft.rb
tomas-stefano: Agreed and done — `Course::SchoolBulkUpdateDraft`, keyed by course + uuid, 24h, swept daily. The repository class and the wizard-gem borrow are gone.

app/services/publish/schools/bulk_update/repository.rb
tomas-stefano: The cache was the whole cause — Draft.create wrote to a store that keeps nothing. Gone with the move to a table; no dev config or docs needed.

app/views/publish/courses/schools/bulk_updates/edit.html.erb
tomas-stefano: Yes — the card asked for it and I missed it. Sized to match; the radio is now built without a hint rather than handed nil.

app/services/publish/schools/bulk_update/apply.rb
tomas-stefano: I think this should be a separate work IMO.

app/views/publish/courses/schools/bulk_updates/edit.html.erb
kmcd: 👍

app/services/publish/schools/bulk_update/repository.rb
kmcd: 👍

config/locales/en/publish/courses/schools/bulk_update.yml
kmcd: 👍

app/services/publish/schools/bulk_update/draft.rb
kmcd: 👍

kmcd (approved):
Nice work!

PR #6405, opened 2026-09-04
Bulk update schools - part 2 (review)

tomas-stefano:
## Context - Bulk update placement schools (2/3) — the review page

Stacked on **1/3**. Between choosing a scope and writing, show which courses will
change — and which cannot.

```mermaid
flowchart LR
    B["bulk-update/:state_key<br/>which courses?"] -->|only this course| W["write this course"]
    B -->|wider scope| R["…/:state_key/review<br/>the courses, and the<br/>ones that cannot change"]
    R -->|"Update … on N courses"| Q[["job"]]
    R -->|Cancel| X["nothing written"]
```

**A course cannot be left with no schools.** One losing its last school is set
aside under *These courses will not be updated*, unless support exempted it. A
course gains whatever is added, so this is only possible on a pure removal —
the query is skipped otherwise. A course that already had none is not set aside;
it is not losing its last one.

Reuses `Publish::Courses::TableComponent`; `Publish::Courses::Query` gains an
`ids` filter so the table gets its status tags without a second read model.

Nothing is written until the button, so **Cancel** truly cancels.

### Test

Needs two courses sharing a funding type, one with a single school.

- [ ] Wider scope → review page; **Only this course** still skips it
- [ ] Table lists exactly the matched courses, with statuses (Closed, Rolled over)
- [ ] Courses from different accredited providers appear in one list
- [ ] Remove a school that is another matched course's only one → it appears under the warning, with the reason, and is untouched after confirming
- [ ] Button count = banner count = courses actually changed
- [ ] **Cancel** → nothing changed anywhere
- [ ] Back keeps the chosen radio; confirming twice does not apply twice

### Notes

The banner says *Schools updated on N courses* while the write is queued. There is
no applied count at that moment — honest copy is an open product question.


tomas-stefano:
https://trello.com/c/YyPZA4QJ/1718-add-bulk-update-review-page-to-placement-schools

app/controllers/concerns/publish/schools/bulk_update_draft.rb
```diff
@@ -0,0 +1,55 @@
...
+          @course.recruitment_cycle_year,
+          @course.course_code,
+        )
+      end
+
+      def scope
+        @scope ||= Publish::Schools::BulkUpdate::Scope.find(course: @course, token: @draft.scope)
+      end
+
+      def matched_courses
```
kmcd, line 46: Getting a 500 error on this one:

1. sign in at http://publish.localhost:3001/personas as Colin (Admin), 
2. Go to http://publish.localhost:3001/publish/organisations/17Y/2026/courses/M303/schools
3. Untick one school then Continue
4. choose All courses then Continue
5. 💥 

```
NoMethodError at /publish/organisations/17Y/2026/courses/M303/schools/bulk-update/e7ce55b8-f6e7-4048-9261-d02a841e057d/review
undefined method 'age_range' for an instance of Course
```

app/services/courses/publish_rules/school_presence_exemption.rb
```diff
@@ -14,6 +14,12 @@ def self.applies?(course)
         course.publish_without_schools_allowed?
       end
 
+      # The same rule as a relation, for callers deciding about many courses at
+      # once rather than loading each one to ask.
+      def self.not_exempt
+        ::Course.where(publish_without_schools_allowed: false)
```
kmcd: This currently matches all 133,440 courses in the DB.

Think we need test coverage and handling for when there are courses which are allowed to have no schools.

kmcd (changes requested):

app/controllers/concerns/publish/schools/bulk_update_draft.rb
tomas-stefano: A primary course in the system spec (the factory gives it an age range) fails exactly so. Fixed by decorating the rows in `MatchedCourses`, where `Publish::CourseList` does it for the course list.

app/services/courses/publish_rules/school_presence_exemption.rb
tomas-stefano: It was a relation meant to be merged into one already bounded, and nothing said so. It now takes the relation and narrows it (requiring_a_school(courses)), so there's nothing to run or misread. Added the review-page scenario that was missing.

app/controllers/concerns/publish/schools/bulk_update_draft.rb
kmcd: 👍

app/services/courses/publish_rules/school_presence_exemption.rb
kmcd: 👍

kmcd (approved):
LGTM! 👍

PR #6406, opened 2026-09-04
Bulk update schools - part 3 (changes)

tomas-stefano:
## Context - Bulk update placement schools (3/3) — play back the change, and fixes

Stacked on **2/3**. Two parts: the playback both bulk pages were missing, and
four fixes to code from the earlier PRs.

### Playback

Both pages ask the provider to commit to a change made elsewhere, so both now
show it. Server-rendered — there are no ticks to watch here.

```mermaid
flowchart TD
    C["SchoolChanges<br/>submitted vs baseline"] --> A["You are adding N schools"]
    C --> R["You are removing N schools"]
    C --> N["all schools in your list<br/>/ no schools added or removed"]
    A --> D{"either side<br/>9 or more?"}
    R --> D
    D -->|yes| E["both lists into details"]
    D -->|no| F["both lists open"]
```

It is the twin of the attach page's JS summary: same wording
(`publish.schools.changes.*`), and `SchoolChanges` mirrors `schools_changes.js`
case for case, including its asymmetric "all" — every school ticked, or none left.

### Fixes to 1/3 and 2/3

Landed here rather than rewriting pushed branches. Reviewers of the earlier PRs
will see the pre-fix code.

| | |
|---|---|
| Exclusion subqueries ran unbounded over `course_school`; Postgres materialised the whole table and rescanned it per matched course | **1,672ms → 12ms** (150-course provider) |
| Confirm ran the full list query for a list of integers | 77ms → 3ms |
| Draft deleted *before* the job was queued — a queue outage lost the provider's whole selection | queue first, then delete; the diff is idempotent so a double submit is harmless |
| Per-course failures swallowed; job then succeeded, so Sidekiq never retried | returns failed ids, comes back for those alone, reports once when it gives up |

Also `docs/bulk-update-placement-schools.md`, and the statements variant of
`count_queries` moved beside the shared helper.

### Test

- [ ] Add and remove schools → both named on the options page and again on review
- [ ] Tick every school → *adding all schools in your list*; untick every one (needs an exempt course) → *removing all*
- [ ] 8 added stays open; 9 added collapses **both** halves into details; 9 removed likewise
- [ ] A wider scope on a provider with a few hundred courses: review page returns promptly
- [ ] Confirm still applies; excluded courses still untouched

### Notes

*Removing all schools in your list* is only reachable on a course support has
allowed to publish without schools — anything else is stopped by validation first.
Worth telling the designer.


tomas-stefano:
https://trello.com/c/KEMyocRR/1717-display-schools-added-and-removed-on-bulk-update-pages

kmcd (approved):
LGTM!

PR #6407, opened 2026-09-04
Link and label ids in Blazer results

inulty-dfe:
## Context

  Blazer results are full of ids that people copy out and paste into the address bar, and several queries join to `provider` or `course` purely so their own output is readable. Blazer does both jobs from config,
  so this turns them on.

  The catch is that `linked_columns` interpolates only `{value}`, and url-escapes it. A column can be linked only when its own value identifies the resource, which rules out `provider_id` entirely: the Publish
  route is keyed by provider code, and the id-keyed Support route also needs a recruitment cycle year. `provider_id` is handled as a smart column instead, which is the better answer anyway, since it removes the
  join rather than just shortening the copy and paste.

  Two things were rejected. A `recruitment_cycle_id` smart column, because no query actually selects that column. And computing the cycle default instead of hardcoding it, because `Blazer.settings` loads at boot
  and `Dockerfile:69` precompiles assets with no database; a spec pins the literal to `Find::CycleTimetable.current_year` so it fails at rollover rather than drifting.

  `user_name` is a bug fix riding along. Blazer defaults it to `:name`, which `User` does not respond to, so the creator column read blank for everyone except the person looking at it. That was invisible until the
  attribution fix started recording creators.

  - Provider codes, accredited provider codes and URNs are links, to Publish and to GIAS.
  - Provider ids, accredited provider codes and course ids show a name under the raw value. The value stays put, and CSV exports are unchanged.
  - Queries can take a cycle from a dropdown with `{recruitment_cycle_year}`, defaulting to the current cycle so a parameterised query still runs on open.
  - Query creators render their names.
  - Specs run every lookup and recognise every internal link, so a mistake in the YAML fails the build instead of surfacing as an empty dropdown months later.

  ## Guidance to review

  `bundle exec rspec spec/config/blazer_spec.rb`. To see it, visit `/blazer` and open query 408, which selects `provider_id` and `provider_code`: the code should link to the provider in Publish and the id should
  carry the provider name beneath it. The Docs link on any query page lists everything configured.

  Two things I would like an opinion on. The `provider_code` dropdown holds around 2,854 entries, which select2 searches fine but is large. And whether `urn` earns its linked column: only three of the 45 queries
  marked keep output one, though they are the site duplication queries that get re-run most.
## Checklist

### DIFF
--- a/config/blazer.yml
+++ b/config/blazer.yml
@@ -1,4 +1,10 @@
 override_csp: true
+
+# Blazer renders a query's creator as `creator.try(Blazer.user_name)`. Its
+# default is :name, which User does not respond to, so the whole column read
+# blank for everyone except the person looking at it.
+user_name: full_name
+
 data_sources:
   main:
     url: <%= Rails.configuration.x.read_only_database_url %>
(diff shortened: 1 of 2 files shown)


inulty-dfe:
[Use userful blazer features](https://trello.com/c/mvsBjw44/1958-use-userful-blazer-features)

tomas-stefano (approved):

PR #6409, opened 2026-09-04
Offer rollover on any course that has not been rolled over

inulty-dfe:
### Providers cannot roll over a course they publish after their rollover has run

Manual rollover was for courses a provider had not published: draft, empty and already rolled over courses could be rolled over one at a time, and anything published was left to the bulk rollover. A provider who creates and publishes a course after their organisation has been rolled over falls outside that, so the course has no route into the next cycle and support has to move it by hand.

The rule now turns on the destination rather than the course's content status. A course can be rolled over if it is not in the next cycle already, is in the current cycle, and there is an upcoming cycle open to publish. Every content status passes that, so the status lists in the model, the controller and the form no longer exclude anything and all three now call a single `Course#manually_rollable?`.

The controller and the form guard different requests. The redirect keeps the confirmation page from rendering for a course that cannot be rolled over, and the validation refuses the submission however the request reached it.

Nothing below the controller stops a second rollover. The endpoint calls `RolloverProviderService` with `force: true`, which bypasses `Provider#rollable?`, so `manually_rollable?` is the only check standing between a stale page and a duplicate course in the next cycle.

- The "Roll over course" button appears on published and withdrawn courses that have not been rolled over, alongside the draft and empty courses that already had it.
- `Publish::Courses::DraftRolloverController` redirects to the course list unless the course is manually rollable, rather than only when it is published.
- `Publish::CourseRolloverForm` validates the same condition, so a POST for a course already present in the next cycle is rejected.
- The validation message changes from "Course must have draft, empty or rolled over status." to "This course cannot be rolled over into the next recruitment cycle."
- The withdrawn panel's preview link sits in the button group rather than in a `<p>` inside it, matching the unpublished panel.

## Guidance to review

```
bundle exec rspec spec/models/course_spec.rb spec/forms/publish/course_rollover_form_spec.rb spec/system/publish/viewing_a_course_spec.rb
```

To check it by hand, sign in as a provider persona, make sure the next recruitment cycle exists with `available_in_publish_from` in the past, then open a published course and a withdrawn course from the courses list. Both offer "Roll over course", and the confirmation page ends on "Course rolled over". Open the same course again and the button is gone.

The withdrawn case is the part I am least sure of. A withdrawn course rolls into the next cycle as a copy the provider then has to publish or withdraw again, and I do not know whether that is wanted or whether the ask covered published courses only. The shared `manually_rollable?` is worth a second opinion too: a change to the button's condition now changes what the endpoint accepts.

### DIFF
--- a/app/controllers/publish/courses/draft_rollover_controller.rb
+++ b/app/controllers/publish/courses/draft_rollover_controller.rb
@@ -3,7 +3,7 @@
 module Publish
   module Courses
     class DraftRolloverController < ApplicationController
-      before_action :redirect_to_courses, if: -> { course.is_published? }
+      before_action :redirect_to_courses, unless: -> { course.manually_rollable? }
 
       def edit
         @course_rollover_form = CourseRolloverForm.new(course)
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -446,9 +446,7 @@ class Course < ApplicationRecord
   end
 
   def manually_rollable?
-    rollover_conditions = !rolled_over? && RecruitmentCycle.upcoming_cycles_open_to_publish? && recruitment_cycle == RecruitmentCycle.current
-
-    rollover_conditions && %i[empty draft rolled_over].include?(content_status)
+    !rolled_over? && RecruitmentCycle.upcoming_cycles_open_to_publish? && recruitment_cycle == RecruitmentCycle.current
   end
 
   def rolled_over?
--- a/app/forms/publish/course_rollover_form.rb
+++ b/app/forms/publish/course_rollover_form.rb
@@ -15,7 +15,7 @@ module Publish
   private
 
     def course_is_rollable?
-      return false if %i[draft empty rolled_over].include? course.content_status
+      return if course.manually_rollable?
 
       errors.add(:course_is_rollable)
     end
--- a/app/views/publish/courses/course_button_panel/_published.html.erb
+++ b/app/views/publish/courses/course_button_panel/_published.html.erb
@@ -5,12 +5,19 @@
 <div class="govuk-button-group">
   <%= govuk_link_to course.application_status_closed? ? "Open course" : "Close course", application_status_publish_provider_recruitment_cycle_course_path, class: "govuk-button govuk-!-margin-right-2" %>
 
+  <% if course.manually_rollable? %>
+    <%= govuk_button_to("Roll over course", rollover_publish_provider_recruitment_cycle_course_path(@provider.provider_code, course.recruitment_cycle_year, course.course_code),
+                      class: " govuk-button--secondary govuk-!-margin-right-2",
+                      data: { qa: "course__rollover" },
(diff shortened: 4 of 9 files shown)


app/controllers/publish/courses/draft_rollover_controller.rb
```diff
@@ -3,7 +3,7 @@
 module Publish
   module Courses
     class DraftRolloverController < ApplicationController
-      before_action :redirect_to_courses, if: -> { course.is_published? }
+      before_action :redirect_to_courses, unless: -> { course.manually_rollable? }
```
tomas-stefano, line 6: One thing to note (not on this line but on the controller) - we flash "Course rolled over" without checking the service result. copy_courses rescues failures into courses_failed inside the transaction, and CopyToProviderService#execute silently returns if the code already exists — either way the provider is told it worked, and can end up with the provider and sites copied but no course. We could check courses/courses_failed before claiming success.

app/models/course.rb
```diff
@@ -446,9 +446,7 @@ def rollable?
   end
 
   def manually_rollable?
-    rollover_conditions = !rolled_over? && RecruitmentCycle.upcoming_cycles_open_to_publish? && recruitment_cycle == RecruitmentCycle.current
-
-    rollover_conditions && %i[empty draft rolled_over].include?(content_status)
+    !rolled_over? && RecruitmentCycle.upcoming_cycles_open_to_publish? && recruitment_cycle == RecruitmentCycle.current
```
tomas-stefano, line 449: Every manual rollover runs the full provider pipeline — sites, study sites, partnerships. rolling_over_draft_courses_spec.rb exists because that duplicated study sites.

If provider rollover published courses successfully one by one will they duplicated the sites/schools etc?

app/controllers/publish/courses/draft_rollover_controller.rb
inulty-dfe: How do we determine if this service ran successfully?

https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/services/providers/copy_to_recruitment_cycle_service.rb#L13

Compare the return value of the `execute` method to expectations based on the arguments passed?

app/models/course.rb
inulty-dfe: Yes, it's all idempotent.

https://github.com/DFE-Digital/publish-teacher-training/blob/main/spec/services/rollover_provider_service_idempotency_spec.rb

app/controllers/publish/courses/draft_rollover_controller.rb
tomas-stefano: I think so yes

tomas-stefano (approved):

inulty-dfe:
[Investigate why we can't manually roll over live courses](https://trello.com/c/GnqRLqOQ/1954-investigate-why-we-cant-manually-roll-over-live-courses)

PR #6415, opened 2026-09-07
Blocking bots on our feedback form

kmcd:
## Context

Bots are polluting feedback data which needs a monthly manual clean-up. 

https://trello.com/c/NSBD9dAD/1931-blocking-bots-on-our-feedback-form

Add honeypot input field which is invisible to users but blocks bots if submitted.

Detected bots get the ordinary success page, so they can't tell they were blocked. 

## Guidance to review

Check http://find.localhost:3001/feedback/new and confirm the honeypot is invisible and the console clean.

Screenshot 2026-09-07 at 14 41 58

## Checklist

### DIFF
--- a/app/views/find/feedbacks/new.html.erb
+++ b/app/views/find/feedbacks/new.html.erb
@@ -22,6 +22,7 @@
             hint: { text: t(".experience_hint") },
             rows: 5,
             max_chars: Feedback::MAX_EXPERIENCE_LENGTH %>
+      <%= invisible_captcha :subject, :feedback, nonce: true %>
       <%= f.govuk_submit t(".submit_feedback_button") %>
     <% end %>
   </div>
--- a/Gemfile
+++ b/Gemfile
@@ -145,6 +145,9 @@ gem "http"
 # For configuring domains and assets
 gem "rack-cors"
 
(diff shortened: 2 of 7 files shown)


inulty-dfe (approved):

app/views/find/feedbacks/new.html.erb
```diff
@@ -22,6 +22,7 @@
             hint: { text: t(".experience_hint") },
             rows: 5,
             max_chars: Feedback::MAX_EXPERIENCE_LENGTH %>
+      <%= invisible_captcha :subject, :feedback, nonce: true %>
```
inulty-dfe, line 25: What is the `nonce` attr for here?

kmcd: Need this one otherwise the honeypot field won't show.

The Content Security Policy [config/initializers/content_security_policy.rb](https://github.com/DFE-Digital/publish-teacher-training/pull/6415/files#diff-c1c619ffb7b249550067cb696b8e7d6c29d1efe2ed4cf5b7a8bb6bed47b409d1R28) sets style-src :self with no unsafe inline.

Inline styles don't have an origin, so the honeypot field wouldn't show without setting `nonce : true`.

PR #6418, opened 2026-09-07
Upgrade sentry-rails and sentry-sidekiq to 7.0.0

inulty-dfe:
## Context

This change updates the sentry-rails and sentry-sidekiq libraries from version 6.7.0 to version 7.0.0.

It replaces the Dependabot pull requests #6412 and #6413, because they do not contain the necessary configuration change. 

Version 7.0.0 starts structured logging automatically. Structured logging sends one message to Sentry for each page request, and one message for each database query. Each message has a cost. The filter that removes private data does not operate on these messages, it operates only on error reports (`lib/sentry/client.rb:239`).

I set `config.rails.structured_logging.enabled` to false at `config/initializers/sentry.rb:41`. The application now operates as it did before the update. I examined two other options: keep structured logging on, or keep only the messages for page requests. Both options change the quantity of data that we send. This is a separate decision, and it is on [a Trello card](https://trello.com/c/27BuMV8n/1965-tech-debt-improve-sentry-data). The other changes in the version 7.0.0 release notes do not apply to this application.

## Guidance to review

  - Start the application with a `SENTRY_DSN` value. The setting `structured_logging.enabled` shows false.
  - Run `bundle exec rspec spec/requests/find/errors_spec.rb spec/requests/publish/errors_spec.rb spec/controllers/concerns/error_handlers/base_spec.rb`. All tests pass.
  - CI runs the full test suite.

## Checklist


langesi (approved):
Thanks for creating a trello ticket as follow up!

gms-gs (approved):

PR #6419, opened 2026-09-07
Report rollover failures instead of claiming success

inulty-dfe:
### The page said "Course rolled over" when the course did not roll over

A provider rolls over one course. `Providers::CopyToRecruitmentCycleService` copies the provider, its schools and its study sites first. Then it copies the course.

If the course copy fails, the service writes the failure into `courses_failed`. It does not stop, and it does not raise. The database saves the rest of the work. So the provider is in the next cycle, and the course is not.

`Publish::Courses::DraftRolloverController` threw the result away. It always showed "Course rolled over". The provider saw success, and support moved the course by hand later.

The page now shows an error instead. The provider can try again, because a course that did not copy still passes `Course#manually_rollable?`.

`Support::Providers::ManualRolloverController` reads the same way, and this PR leaves it alone. `copy_courses` selects no courses when `force` is true and the caller gives no course codes. That is how the support controller calls the service. No course copy runs there, so no course copy can fail. That controller copies no courses at all and still reports success. That is a different problem, and it needs its own card.

I rejected one other option: roll the whole transaction back when a course fails. The bulk rollover must continue through the other courses. So the failure stays with the course, and the caller reports it.

image

- The rollover page shows an error when the course does not copy. The success message does not appear.
- The provider can start the rollover again after the error.

## Guidance to review

```
bundle exec rspec spec/system/publish/courses/rolling_over_a_course_that_fails_spec.rb
```

The test makes the copy fail for real. It clears `start_date` on the course row, because `Courses::CopyToProviderService` adds a year to that date. The test mocks nothing. Tell me if you know a better real trigger.

I checked the test fails without the new guard. Without it, the page says "Course rolled over" for a course that is not in the next cycle.

## Checklist

### DIFF
--- a/app/controllers/publish/courses/draft_rollover_controller.rb
+++ b/app/controllers/publish/courses/draft_rollover_controller.rb
@@ -11,16 +11,23 @@ module Publish
 
       def update
         @course_rollover_form = CourseRolloverForm.new(course)
-        if @course_rollover_form.valid?
-          RolloverProviderService.call(provider_code: params[:provider_code], course_codes: params[:code]&.split, force: true)
+        return render :edit unless @course_rollover_form.valid?
+
+        result = RolloverProviderService.call(provider_code: params[:provider_code], course_codes: params[:code]&.split, force: true)
+
+        # The service collects a failed course copy into `courses_failed` and
+        # commits the rest of the transaction, so the provider and its sites can
+        # be in the next cycle without the course.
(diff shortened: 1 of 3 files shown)


inulty-dfe:
[Report provider course rollover failures](https://trello.com/c/nXP55KLJ/1966-report-provider-course-rollover-failures)

langesi (approved):
Thanks for fixing this

PR #6433, opened 2026-09-09
Track click visa and radius links

tomas-stefano:
## Context

Adds `track_click` tracking to three untracked candidate-facing links, so the upcoming analytics work has full coverage of service track links.

## What's tracked now

| Link | `utm_content` |
|---|---|
| Visa sponsorship → types of visa (fee course, sponsors) | `student_visa_available_types_of_visa` |
| Visa sponsorship → types of visa (salaried, sponsors) | `skilled_worker_visa_available_types_of_visa` |
| Visa sponsorship → types of visa (salaried, cannot sponsor) | `skilled_worker_visa_not_available_types_of_visa` |
| Visa sponsorship → contact the training provider | `<visa_type>_available_contact_the_training_provider` |
| No results → wider radius suggestions | `no_results_radius_quick_link_{10,20,50,100}_miles` |

The ticket only asks for courses that sponsor visas, but the types of visa link also renders on salaried courses that cannot sponsor a Skilled Worker visa, so that variant is covered too.

## Notes for review

- The two visa links were hand-written `<a>` tags inside a block of HTML in `find.yml`, one with an **unquoted** `href=%{provider_url}`. The first commit moves the anchors into `govuk_link_to` calls, matching `entry_requirements_component`, and changes nothing on the page — the unquoted attribute would have broken once an `&`-bearing track URL went into it.
- The radius links are built from the visitor's own query string, and that value is now what a redirect acts on. There's a spec proving `host`, `protocol` and `script_name` in the query can't push it off site: they arrive as string keys, so `url_for` treats them as ordinary query params. `safe_redirect_url` is a second barrier behind that.
- Publish preview needs no special handling — `/track_click` on the publish host resolves to `Publish::TrackController`, which no-ops the event.
- **Separate commit, worth flagging to whoever owns the dashboards:** four advice callout links passed `utm:` rather than `utm_content:`, and the specs mirrored the typo, so they've been recording clicks with a null `utm_content`. That value stops being null for `how_to_choose_a_training_provider`, `interview_preparation`, `school_placement_activities` and `teacher_training_expectations`.

### DIFF
--- a/app/components/shared/courses/how_to_choose_a_training_provider_component.html.erb
+++ b/app/components/shared/courses/how_to_choose_a_training_provider_component.html.erb
@@ -4,7 +4,7 @@
     <% if is_preview %>
       <%= govuk_link_to t(".content.find_out_how_to_choose_a_training_provider"), t("find.get_into_teaching.url_type_of_course_provider") %>
     <% else %>
-      <%= govuk_link_to t(".content.find_out_how_to_choose_a_training_provider"), find_track_click_path(utm: "how_to_choose_a_training_provider", url: t("find.get_into_teaching.url_type_of_course_provider")) %>
+      <%= govuk_link_to t(".content.find_out_how_to_choose_a_training_provider"), find_track_click_path(utm_content: "how_to_choose_a_training_provider", url: t("find.get_into_teaching.url_type_of_course_provider")) %>
     <% end %>
   </p>
 <% end %>
--- a/app/services/courses/radius_quick_link_suggestions.rb
+++ b/app/services/courses/radius_quick_link_suggestions.rb
@@ -31,7 +31,7 @@ module Courses
     def build_link(values)
       {
         text: translate("radius_in_miles", count: values[:radius], course_count: format_course_count(values[:count])),
-        url: find_results_path(values[:radius]),
+        url: tracked_results_path(values[:radius]),
       }
     end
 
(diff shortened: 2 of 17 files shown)


inulty-dfe:
[Add tracking for missing track_click events](https://trello.com/c/nJGBTMX0/1421-add-tracking-for-missing-trackclick-events)

app/services/courses/radius_quick_link_suggestions.rb
```diff
@@ -49,10 +49,21 @@ def format_course_count(count)
       translate(key, count:)
     end
 
+    def tracked_results_path(radius)
```
samjohnson-dev, line 52: The destination URL includes the visitor’s search query (latitude / longitude / location), and ClickEvent stores that whole url in BigQuery. Same data already rides in search_results — can you confirm warehouse/analytics are happy with it also appearing on track_click?

app/components/shared/courses/how_to_choose_a_training_provider_component.html.erb
```diff
@@ -4,7 +4,7 @@
     <% if is_preview %>
       <%= govuk_link_to t(".content.find_out_how_to_choose_a_training_provider"), t("find.get_into_teaching.url_type_of_course_provider") %>
     <% else %>
-      <%= govuk_link_to t(".content.find_out_how_to_choose_a_training_provider"), find_track_click_path(utm: "how_to_choose_a_training_provider", url: t("find.get_into_teaching.url_type_of_course_provider")) %>
+      <%= govuk_link_to t(".content.find_out_how_to_choose_a_training_provider"), find_track_click_path(utm_content: "how_to_choose_a_training_provider", url: t("find.get_into_teaching.url_type_of_course_provider")) %>
```
samjohnson-dev, line 7: So these 4: how_to_choose_a_training_provider, interview_preparation, school_placement_activities, teacher_training_expectations will start showing up in big query now did we need to warn anyone who is charting them?

samjohnson-dev (commented):
Looks good just few comments

app/components/shared/courses/how_to_choose_a_training_provider_component.html.erb
tomas-stefano: No, don't need. Lewis is aware of this card so it is fine.

app/services/courses/radius_quick_link_suggestions.rb
tomas-stefano: Yes, is on the card to action those. This is only for the no results from location searches.

inulty-dfe (approved):

samjohnson-dev (approved):
🚀

PR #6435, opened 2026-09-09
Make solid_cache_entries UNLOGGED to reduce WAL pressure

samjohnson-dev:
## Context

Solid Cache’s `solid_cache_entries` table is already deployed as a normal logged table. Before we switch Rails.cache to Solid Cache, we should make it `UNLOGGED` so disposable cache writes do not add unnecessary WAL / replication pressure (Airbyte is enabled on production).

This must land and deploy before the cache cutover PR.

- Add a migration to `ALTER TABLE solid_cache_entries SET UNLOGGED` (with a `SET LOGGED` rollback).

## Guidance to review

- Confirm the migration matches the Apply `SET UNLOGGED` pattern (`disable_ddl_transaction!` + `safety_assured`).
- Safe while the table is still empty / unused in production.

### DIFF
--- a/db/migrate/20260909161043_set_solid_cache_entries_unlogged.rb
+++ b/db/migrate/20260909161043_set_solid_cache_entries_unlogged.rb
@@ -1,0 +1,11 @@
+class SetSolidCacheEntriesUnlogged < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  def up
+    safety_assured { execute "ALTER TABLE solid_cache_entries SET UNLOGGED" }
+  end
+
+  def down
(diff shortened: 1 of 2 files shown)


tomas-stefano (approved):
Approving. Two things to be aware of:

  - `UNLOGGED` isn't dumped to schema.rb, so any environment built via `db:setup/db:prepare` gets a `LOGGED` table and won't re-run this migration. Worth re-asserting it at deploy, or documenting it as a production-only property.
- Any failover or unclean restart will erase the cache. Just sstating to be aware of.

langesi (approved):
Well done!

PR #6442, opened 2026-09-11
Make sure test suite doesn't fail in diferent cycle periods

inulty-dfe:
## Context

Find and Publish change how they behave at points in the recruitment year. The code finds that point from the clock.

A spec that does not control the clock can assert a different thing on a different date. Some specs pass for most of the year. They then fail in the last weeks of the cycle, after the Apply deadline. The failure has no link to the change under test.

This branch gives every example a fixed instant. An example with `travel:` keeps its own instant. Every other example runs at `mid_cycle`. Find is open, Apply is open, and the deadline is not past.

The default fixes the period, but not the year. `mid_cycle` reads `current_year`, and `current_year` reads the real clock. On the day of a rollover, every example without `travel:` moves to the new year. A spec that needs one year must name it with `travel:`.

We looked at an override that sets the instant for a whole run from the command line. We chose the fixed default instead. It gives every example a known period, and nobody must remember a flag. An override is still useful. It must add to the default, and not replace it. The guide holds the rules for it.

## Changes

- Add documentation around testing cycle periods
- pin all tests to mid_cycle of the current cycle year that don't explicitly use Timecop

## Guidance for review

I changed my laptop clock to 6am Find closes date and ran the test suite and it passes 100%

## Checklist


inulty-dfe:
[Fix flakey specs during find closing](https://trello.com/c/LwflhHtj/1986-fix-flakey-specs-during-find-closing)

inulty-dfe:
[Fix flakey specs during find closing](https://trello.com/c/LwflhHtj/1986-fix-flakey-specs-during-find-closing)

gms-gs (approved):
LGTM! Good work on this

PR #6451, opened 2026-09-14
Add link to same course in current cycle on Saved courses

gms-gs:
## Context

Saved courses from a previous recruitment cycle are no longer linked because their original Find URL may return a 404 after the new cycle starts.

When the same course has been rolled over and published in the current cycle, candidates should be able to access the new version without searching for it again.

- Add a **“View this year’s course”** link to previous-year saved course cards.
- Match courses using their provider code and course code.
- Only show the link when the matching current-cycle course is published.
- Do not show additional content when no matching course exists.
- Position the new link alongside **Delete**, aligned with the previous-year label.
- Move the new link text into the locale file.
- Add component and system coverage for courses with and without a current-cycle equivalent.

The controller loads all matching current-cycle courses in one query and indexes them by provider and course code. This avoids querying `in_next_cycle` separately for every saved course card, which would introduce an N+1 query. The lookup is request-local rather than persistently cached, avoiding stale links when courses are published, withdrawn or discarded.

### Before

Previous-year saved courses had no way to reach the current version.

### After

Screenshot 2026-09-14 at 17 37 30

Previous-year saved courses show **“View this year’s course”** when a published equivalent exists.

## Guidance to review

1. Save a course from the previous recruitment cycle which has a published equivalent in the current cycle.
2. Visit **Saved courses**.
3. Confirm **“View this year’s course”** appears and opens the correct current-cycle course.
4. Confirm a previous-year course without an equivalent has no additional link.
5. Confirm the link and **Delete** remain inline and align with the previous-year label.
6. Review the bulk lookup in `SavedCoursesController`, particularly the provider code and course code matching.

## Checklist


gms-gs:
[Add link to same course in current cycle on Saved courses](https://trello.com/c/PjZozVOi/1974-add-link-to-same-course-in-current-cycle-on-saved-courses)

config/settings.yml
```diff
@@ -48,7 +48,7 @@ authentication:
   # mode: persona     # none critical systems, ie localhost
 
 one_login:
-  enabled: true
+  enabled: false
```
tomas-stefano: This flips one_login.enabled to false in the base settings. It this on purpose?

app/components/saved_courses/summary_card_component.html.erb
```diff
@@ -1,5 +1,15 @@
-<%= govuk_summary_card(classes: %w[course-summary-card], title:) do |card| %>
+<%= govuk_summary_card(classes: %w[course-summary-card app-saved-course-summary-card], title:) do |card| %>
   <% card.with_action do %>
+    <% if current_cycle_course.present? %>
```
tomas-stefano: The hash is keyed by [provider_code, course_code] and the view looks up every saved course, but
  codes are the same across cycles. So if a candidate saved X/ABC in both 2025 and 2026:

  - 2025 card → finds 2026 course → link ✅
  - 2026 card → finds 2026 course → itself → "View this year's course" links to its own page ❌

  Could you guard the link with course_from_previous_year? (or only pass current_cycle_course: for
  previous-cycle rows) and add a spec for this case?

config/settings.yml
gms-gs: Whoops! need to revert this! Was just testing locally. Good spot

tomas-stefano (approved):

PR #6453, opened 2026-09-15
Deploy Solid Queue beside Sidekiq with Mission Control

samjohnson-dev:
Stand up a coexisting Solid Queue worker, queue config, and system-admin Mission Control so we can monitor delayed jobs without changing the global Sidekiq adapter.

## Context
Solid Queue tables are already on the primary DB. This stands up a coexisting Solid Queue worker and Mission Control beside Sidekiq so we can canary delayed jobs before moving any product work. Global adapter stays :sidekiq; Redis is untouched.

- Add queue.yml / non_production_queue.yml with explicit dispatchers and all PTT queues (mailers kept separate from bulk).
- Mount Mission Control at /jobs for system admins; keep /sidekiq.
- Redact Solid Queue args in logs; add TestJob::DispatcherCanaryJob for a post-deploy canary.
- Add a dedicated solid-queue-worker Terraform deployment with a smaller DB pool so it does not inherit web’s RAILS_MAX_THREADS: 50.

## Guidance to review
- Queue coverage + dispatcher config.
- Terraform name/probe/command; Sidekiq worker unchanged.
- After deploy: TestJob::DispatcherCanaryJob.set(wait_until: 1.minute.from_now).perform_later and confirm scheduled → ready → completed in Mission Control.


samjohnson-dev:
- `/jobs` loads for system admin
- Canary: `TestJob::DispatcherCanaryJob.set(wait_until: …).perform_later`
- Result: job id 1 finished (`scheduled_at` → `finished_at` ~4s later on `default`)
- Global adapter still Sidekiq; Solid Queue worker coexists

samjohnson-dev:
Screenshot 2026-09-16 at 15 38 29

terraform/aks/application.tf
```diff
@@ -65,3 +65,38 @@ module "worker_application" {
...
+    "-c",
+    # Include solid-queue-worker in the command line so the probe can find this process.
+    # $${...} so Terraform leaves the shell default for DATABASE_CONNECTION_POOL_SIZE alone.
+    "SOLID_QUEUE_PROCESS_NAME=solid-queue-worker DATABASE_CONNECTION_POOL_SIZE=$${DATABASE_CONNECTION_POOL_SIZE:-5} bundle exec rake solid_queue:start",
+  ]
+  max_memory      = var.solid_queue_worker_memory_max
+  replicas        = var.solid_queue_worker_replicas
+  enable_logit    = var.enable_logit
+  run_as_non_root = var.run_as_non_root
+  probe_command   = ["pgrep", "-f", "solid-queue-worker"]
```
tomas-stefano: `pgrep -f solid-queue-worker`

Will only matches after a worker has forked and set its proctitle. 

During Rails boot the cmdline is `bundle exec rake solid_queue:start`, so the probe fails until the first fork. 

The `SOLID_QUEUE_PROCESS_NAME=` trick on line 93 doesn't help — ash execs the single command, so the wrapper string is gone.

Maybe we can do `pgrep -f "solid.queue"` — matches both the rake task and the forks — and drop the fake env var. Worth checking the probe's initial delay on the vendor module so a slow boot doesn't restart-loop the pod.

terraform/aks/application.tf
```diff
@@ -65,3 +65,38 @@ module "worker_application" {
...
+
+  docker_image = var.docker_image
+  # Shared ConfigMap sets RAILS_MAX_THREADS=50 for web. Override the DB pool only
+  # for this process so each Solid Queue fork does not open a 50-connection pool.
+  command = [
+    "/bin/sh",
+    "-c",
+    # Include solid-queue-worker in the command line so the probe can find this process.
+    # $${...} so Terraform leaves the shell default for DATABASE_CONNECTION_POOL_SIZE alone.
+    "SOLID_QUEUE_PROCESS_NAME=solid-queue-worker DATABASE_CONNECTION_POOL_SIZE=$${DATABASE_CONNECTION_POOL_SIZE:-5} bundle exec rake solid_queue:start",
```
tomas-stefano: Pool is pinned to 5 but the threads in queue.yml come from env vars. 

If someone bumps `DEFAULT_QUEUE_THREADS` in the ConfigMap without bumping this, Solid Queue just warns and then queues on the pool. 

A comment here or in `queue.yml` would save someone a bad afternoon. :)

config/application.rb
```diff
@@ -73,12 +73,29 @@ class Application < Rails::Application
...
     config.log_level = Settings.log_level
 
+    # QA/staging/sandbox/review use dedicated RAILS_ENV values (see app_config.yml).
+    config.before_configuration do
+      ENV["SOLID_QUEUE_CONFIG"] =
+        if Rails.env.production? || Rails.env.development?
+          "config/queue.yml"
+        else
+          "config/non_production_queue.yml"
+        end
```
tomas-stefano: Two things here:

1. `before_configuration` has already run by this point, so the block executes straight away. It's just an ENV assignment wrapped in a hook that does nothing.
2. It overwrites `SOLID_QUEUE_CONFIG` if someone sets it in the ConfigMap.


One thing we could dos: `queue.yml` is already keyed by env, so add the qa:, staging: etc. sections there, delete non_production_queue.yml and this block. 

Solid Queue picks the right section on its own.

config/application.rb
```diff
@@ -73,12 +73,29 @@ class Application < Rails::Application
...
     config.exceptions_app = routes
     config.active_job.queue_adapter = :sidekiq
-    # Retention only — Solid Queue is not yet the live adapter or a running worker.
+    # Retention only — Solid Queue is not yet the live adapter. Sidekiq remains global.
     config.solid_queue.clear_finished_jobs_after = 1.hour
 
+    # Mission Control observes Solid Queue only; Sidekiq Web stays at /sidekiq.
+    config.mission_control.jobs.adapters = [:solid_queue]
+    config.mission_control.jobs.http_basic_auth_enabled = false
+    config.mission_control.jobs.filter_arguments = %w[
```
tomas-stefano: Heads up: `filter_arguments` only matches top-level hash keys. Positional args and nested hashes ({ user: { email: … } }) show in full. Same as Sidekiq Web today so not a blocker, just don't lean on it for PII.

We already have config.filter_parameters (partial match, recursive) — can we reuse that list?

Gemfile
```diff
@@ -17,6 +17,9 @@ gem "solid_cache", "~> 1.0"
 # Database-backed Active Job backend (schema first; Sidekiq remains the live adapter)
 gem "solid_queue", "~> 1.7"
 
+# Operational UI for Solid Queue (system-admin auth; Sidekiq Web remains)
+gem "mission_control-jobs", "~> 1.3"
```
tomas-stefano, line 21: `mission_control-jobs` pulls in `turbo-rails`, which draws three public GET routes into the app (/recede_historical_location etc.). 

Maybe is fine, but we didn't ask for them.
`config.turbo.draw_routes = false`

 if we'd rather keep the route table clean.

tomas-stefano (commented):
Awesome work - just minor comments. ❤️

terraform/aks/application.tf
samjohnson-dev: good catch, switched to pgrep -f solid.queue and dropped the fake env var

terraform/aks/application.tf
samjohnson-dev: Added a comment tying pool size to queue.yml thread/process totals

config/application.rb
samjohnson-dev: Collapsed into one queue.yml keyed by env and  removed the override.

Gemfile
samjohnson-dev: Good plan!

config/application.rb
samjohnson-dev: Now derived from filter_parameters + our job keys and noted mission control still only redacts top-level.

config/initializers/mission_control.rb
```diff
@@ -0,0 +1,15 @@
...
+Rails.application.config.after_initialize do
+  from_filter_parameters = Rails.application.config.filter_parameters.filter_map do |param|
+    case param
+    when String, Symbol then param.to_s
+    end
+  end
+
+  keys = (from_filter_parameters + %w[email_address code data body hidden_data headers]).uniq
+  MissionControl::Jobs.filter_arguments = keys
+  Rails.application.config.mission_control.jobs.filter_arguments = keys
```
tomas-stefano, line 14: worth a line to verfy: filter_parameters is partial-match (passw catches password) but Mission Control is exact-match.

tomas-stefano (approved):
One minor thing to verify but all look good to me 🚀

config/initializers/mission_control.rb
samjohnson-dev: yeah il tighten the exact list before we do real jobs 🫡

PR #6457, opened 2026-09-16
Convert remaining Sidekiq jobs to Active Job with explicit retries

samjohnson-dev:
## Context

Convert the remaining native `Sidekiq::Job` classes to Active Job so they can later run on Solid Queue, and lock in intentional retry behaviour. Global adapter stays `:sidekiq`; Redis/Sidekiq infra unchanged. Stacks on `sj/solid-queue-runtime`.

Solid Queue has no Sidekiq-style default retries. Intended policies:

| Job / group | Policy |
|-------------|--------|
| `UpdateCourseSchoolsJob` | `retry_on StandardError, attempts: 0` (sends change notifications) |
| `RolloverProviderJob` | `retry_on StandardError, attempts: 0` (preserves previous `sidekiq_options retry: false`) |
| `RolloverMonitoringJob` | `retry_on StandardError, attempts: 0` (self-reschedules; mirrors blank-coordinates monitoring) |
| Already no-auto-retry | `RolloverProvidersBatchJob`, blank-coordinates jobs, `SendWeeklyEmailAlertsJob` |
| Other PTT `ApplicationJob`s | ApplicationJob defaults only (`Deadlocked` retry, discard deserialization). No blanket retries added — Slack/mail side effects stay manual-retry via Sidekiq UI now / Mission Control later |
| `ActionMailer::MailDeliveryJob` | Rails defaults; unchanged |
| `DfE::Analytics::*` | Gem `AnalyticsJob` already `retry_on StandardError, attempts: 5`; unchanged |

Failed jobs remain visible under Sidekiq while it is still the global adapter.

- Convert `UpdateCourseSchoolsJob`, `RolloverProviderJob`, and `RolloverMonitoringJob` from `Sidekiq::Job` to `ApplicationJob` with explicit no-auto-retry
- Replace `perform_async` / `perform_in` with `perform_later` / `set(wait:).perform_later` at all call sites
- Update specs for the Active Job enqueue API

## Guidance to review

- Confirm retry choices for the three converted jobs (especially no auto-retry on school updates because of notifications)
- Confirm enqueue API changes match blank-coordinates monitoring (`set(wait:)` + `perform_later`)
- Adapter must still be `:sidekiq` in `config/application.rb`
- Review-app checks: large course-school save (>30 schools) and Sidekiq Web job visibility


app/jobs/test_job/dispatcher_canary_job.rb
samjohnson-dev, line 8: Probs dont need this as tested on review app and locally but will get removed quickly and useful to have just incase

app/jobs/application_job.rb
```diff
@@ -4,4 +4,14 @@ class ApplicationJob < ActiveJob::Base
   retry_on ActiveRecord::Deadlocked
 
   discard_on ActiveJob::DeserializationError
+
+  # No Active Job retries for StandardError. The terminal block reports and does
+  # not re-raise: a bare `retry_on ..., attempts: 0` re-raises after AJ gives up,
+  # and Sidekiq's JobWrapper (retry: true) would then apply its default retries
+  # while the global adapter is still :sidekiq.
+  def self.without_auto_retry
```
kmcd, line 17: This looks like all errors are now silently dropped.

The `register_error_subscriber` for `Rails.error` is `false`.

Previously they notified Sentry and were viewable in Sidekiq UI in the retries or dead tabs.

This is potentially a blocker.

app/jobs/application_job.rb
```diff
@@ -4,4 +4,14 @@ class ApplicationJob < ActiveJob::Base
...
+    retry_on StandardError, attempts: 0 do |_job, error|
```
kmcd: `without_auto_retry` is registered **after** `retry_on ActiveRecord::Deadlocked`

`rescue_from` uses last registered first, so the deadlock retry will never be called.

Don't think this is theoretical as `UpdateCourseSchoolsJob` uses two row locks inside a transaction.

Another potential blocker.

spec/jobs/rollover_provider_job_spec.rb
```diff
@@ -9,7 +9,15 @@
...
+      described_class.perform_now("ABC", 123, process_summary.id)
     end
   end
+
+  it "does not automatically retry StandardError" do
+    expect(DataHub::Rollover::ProviderProcessor).to receive(:process).once.and_raise(StandardError, "boom")
+
+    expect {
+      described_class.perform_now("ABC", 123, process_summary.id)
+    }.not_to raise_error
```
kmcd, line 22: Think the coverage might be a bit sparse for eight classes.

Maybe ensure that the jobs have or haven't been enqueued; e.g.

```ruby
expect { described_class.perform_now("ABC", 123) }
  .not_to have_enqueued_job(described_class)
```

Also might be worth testing attempts 0/5 etc.

spec/services/data_hub/rollover/job_orchestrator_spec.rb
```diff
@@ -37,12 +30,13 @@
...
-        kind_of(Numeric),
-        anything,
-        1,
-      )
+      allow(RolloverMonitoringJob).to receive(:set).and_return(RolloverMonitoringJob)
+      allow(RolloverMonitoringJob).to receive(:perform_later)
+
       subject
+
+      expect(RolloverMonitoringJob).to have_received(:set).with(wait: DataHub::Rollover::JobOrchestrator::MONITORING_START_TIME)
```
kmcd: I'd prefer to avoid mocking/stubbing if possible.

e.g. This just asserts a constant so any value passes.

Can we use `have_enqueued_job(RolloverMonitoringJob)` for these tests instead?

spec/services/data_hub/rollover/monitoring_manager_spec.rb
```diff
@@ -33,16 +33,20 @@
       end
 
       it "schedules next check when under attempt limit" do
-        expect(RolloverMonitoringJob).to receive(:perform_in).with(5.minutes, process_summary.id, 2)
+        allow(RolloverMonitoringJob).to receive(:set).and_return(RolloverMonitoringJob)
```
kmcd: Same as the other suggestion - can we use `have_enqueued_job(RolloverMonitoringJob)` instead?

app/jobs/blank_coordinates_backfill/backfill_job.rb
```diff
@@ -1,7 +1,7 @@
 module BlankCoordinatesBackfill
   class BackfillJob < ApplicationJob
     queue_as :default
-    retry_on StandardError, attempts: 0
+    without_auto_retry
```
kmcd, line 4: Can we use Rails built in `discard_on StandardError, report: true` instead?

kmcd (changes requested):

spec/lib/custom_log_formatter_spec.rb
inulty-dfe: Can we redact inidividual job arguments instead of all of them?

By that I mean, leave unsensitive args unredacted?

inulty-dfe: Can we have a test for each of the arguments that should be redacted?

terraform/aks/workspace_variables/dv_review.tfvars.json
inulty-dfe: Should all envs have the same memory allocated as prod?

inulty-dfe (commented):
Couple questions

app/jobs/application_job.rb
samjohnson-dev: good catch have addressed the silent errors

app/jobs/application_job.rb
samjohnson-dev: Good spot, re-delcared after the discard

spec/jobs/rollover_provider_job_spec.rb
samjohnson-dev: ive added a bit more spec for discard/deadlocks

spec/lib/custom_log_formatter_spec.rb
samjohnson-dev: have done this now good idea

terraform/aks/workspace_variables/dv_review.tfvars.json
samjohnson-dev: have changed for non-prod envs

spec/services/data_hub/rollover/job_orchestrator_spec.rb
samjohnson-dev: switched these to have_enqueued_job(RolloverMonitoringJob) so we assert the real enqueue (including wait), not a stubbed constant.

app/lib/custom_log_formatter.rb
```diff
@@ -68,7 +68,25 @@ def redact_solid_queue_arguments
...
+
+  def filter_job_arguments(value)
+    case value
+    when Array
+      value.map { |item| filter_job_arguments(item) }
+    when Hash
+      job_argument_filter.filter(value)
+    else
+      value
+    end
```
tomas-stefano, line 82: `rails_semantic_logger's` `ActiveJob` subscriber puts `payload[:arguments]` in as a JSON string
  (`JSON.pretty_generate(job.arguments…)`), not an Array. 

`filter_job_arguments` hits the else branch and returns it  untouched. 

The spec is green only because it hands the formatter Ruby arrays.

tomas-stefano (approved):

kmcd (approved):

PR #6459, opened 2026-09-17
Update Schools list in the Support console

gms-gs:
## Context

Support users need more useful information on the provider’s Schools tab and a quicker way to remove schools.

Previously, the list only showed the school name, code and URN. Support users had to open the school details page before starting the removal journey.

- Show each school’s address beneath its name.
- Add a **Number of courses attached** column with pluralised counts.
- Add a **Remove school** column with a destructive link to the existing removal journey.
- Preload attached courses to avoid an additional query for every school.
- When removal starts from the index, Back and Cancel return to the index and preserve the pagination page.
- When removal starts from school details, Back and Cancel continue returning to the details page.
- Preserve the original return destination if removal fails.
- Add request and system coverage for the new content and navigation.

### Before

The list showed the school name, code and URN.

### After

The list also shows the address, attached-course count and a direct removal link.

## Guidance to review

1. Open a provider’s **Schools** tab in Support.
2. Confirm the address appears beneath each school name.
3. Confirm the attached-course count is correct and pluralised.
4. Select **Remove school** from the list.
5. Confirm Back and Cancel return to the Schools index.
6. Start removal from school details and confirm Back and Cancel return to that school.
7. Check the behaviour from page 2 to confirm pagination is preserved.

Worth reviewing: the attached-course preload and the `from=index` return-path handling.

## Checklist


gms-gs:
[Update Schools list in the Support console](https://trello.com/c/yRyYaJHH/1964-update-schools-list-in-the-support-console)

tomas-stefano (approved):

PR #6460, opened 2026-09-17
Retry yarn install when it crashes in CI

kmcd:
## Context

Main has gone red three times since 16 September on `yarn install --immutable`.

Retry yarn install three times.

## Guidance to review

I tested this locally by replacing yarn with a bash script that blows up.

FYI: node_modules is cached but not Yarn's own cache, so every job refetches everything.

## Checklist


samjohnson-dev (approved):
Yeah i think this looks good and happy to merge! 🥇

PR #6464, opened 2026-09-18
Fix n plus one queries

tomas-stefano:
## Context

13 N+1 queries from the request and system specs, and a fresh full-suite run turns up a few more in code that branch predates. This fixes them, one commit per query, so each is easy to read and revert on its own.

 1. Course => [:provider] — support courses list + copy courses
  2. Provider => [:recruitment_cycle] @ course.rb — those two + Courses::Query (Find results); spec
     added
  3. Provider => [:recruitment_cycle] @ provider.rb — support candidate saved courses/notes
  4. Course => [:latest_enrichment] @ course.rb — support list + copy courses
  5. Course => [:latest_enrichment] @ exports — fallback picked from preloaded rows; count spec
     added
  6. Course => [:site_statuses] — support list
  7. Course => [:course_subjects] — copy courses
  8. Course => [:accrediting_provider] @ course.rb — support list; request spec added holding the
     whole list to a constant query count
  9. Course => [:accrediting_provider] @ fetch.rb — copy-content sidebar; behaviour + count specs
     added
  10. CourseSubject => [:subject] @ course.rb — validations go by subject_id;
      validate_modern_languages now has specs
  11. CourseSubject => [:subject] @ course_subjects_form.rb
  12. CourseSubject => [:subject] @ course_decorator.rb
  13. SecondarySubject => [:financial_incentive] — request spec added

Most are a missing `preload`/`includes` where a list is loaded; the rest are reads that didn't need the association at all (comparing `subject_id` instead of loading each subject, picking the export's fallback enrichment from rows already loaded, comparing years in `RecruitmentCycle#current?` rather than querying the current cycle).

One thing worth knowing: `Course#accrediting_provider` has an instance-dependent scope, so wherever it's preloaded the provider and recruitment cycle have to be preloaded *ahead of it* — the other way round, Rails loads them per course. Each site has a comment, and the support courses list spec fails if the order flips.

Specs added where the page matters (Find results, the copy-content sidebar, the support courses list, the course exports, Find's secondary subjects page, the public API courses endpoint) — each holds the page to a constant number of queries as rows are added, so a preload that stops matching what the row reads fails the build. `validate_modern_languages` had no spec and now does.

After this, the full suite's Bullet log is empty.


tomas-stefano:
https://trello.com/c/4L52BeR2/2000-review-and-tighten-bullet-gem-configuration

app/controllers/api/public/v1/providers/courses/locations_controller.rb
```diff
@@ -64,7 +64,7 @@ def course
               @course ||= if schools_remodelled
                             provider.courses.includes(schools: %i[gias_school provider_school]).find_by(course_code: params[:course_code])
                           else
-                            provider.courses.includes(site_statuses: [:site]).find_by(course_code: params[:course_code])
+                            provider.courses.includes(site_statuses: { site: :provider }).find_by(course_code: params[:course_code])
```
kmcd: I don't think this fixes the N+1 🤔 

This preloads `site_statuses.site.provider` but locations renders `course.sites`

In fact I think this introduces an N+1 for `Course#sites` ...

app/models/course.rb
```diff
@@ -905,8 +905,11 @@ def master_subject_nil?
     master_subject_id.nil?
   end
 
+  # Checked on every save with subjects. The course subjects may only just
+  # have been built, so this goes by subject_id rather than loading each
+  # subject in turn.
   def has_any_modern_language_subject_type?
-    course_subjects.any? { |cs| cs.subject.type == "ModernLanguagesSubject" }
+    ModernLanguagesSubject.exists?(id: course_subjects.map(&:subject_id))
```
kmcd: `.exists?` hits the DB on every call so think this might introduce an N+1.

Maybe keep the in-memory check and preload course_subjects: :subject instead?

app/models/course.rb
```diff
@@ -1197,20 +1204,20 @@ def validate_subject_count
     end
   end
 
+  # Goes by the course subjects' subject_ids: they may only just have been
+  # built, and reading subjects then loads each one's subject in turn.
   def validate_subject_consistency
-    subjects_excluding_discontinued = subjects.reject do |subject|
-      DiscontinuedSubject.exists?(id: subject.id)
-    end
+    subject_ids = Subject.where(id: course_subjects.map(&:subject_id)).where.not(type: "DiscontinuedSubject").ids
```
kmcd: Could use the existing scope for this instead: https://github.com/DFE-Digital/publish-teacher-training/pull/6464/changes#diff-b42bbf291accc96e1e25a8fd9975a816bba944d51a4b5f440b9533b7118927f8R15

kmcd (changes requested):

PR #6465, opened 2026-09-18
Update load test docs and scripts

inulty-dfe:
## Context

We must run load tests on Find before the new cycle starts. I read the load test docs first. The docs did not agree with the code, and some of the differences stopped the tests.

Three things to note:

- The GitHub Actions workflow never worked. Every step called `find/find-load-test.js`, but the file is `find/load-test.js`.
- The README listed `find:baseline`, `find:peak` and `find:stress` under a heading called "Running on Grafana". Those scripts call `k6 run`, not `k6 cloud`. The load comes from your own machine.
- `find:dev:quick` did not run the quick scenario. The script passed `--vus 5 --duration 10s`, and k6 says `"cli" level configuration overrode scenarios configuration entirely`. The run lost the scenario and its tags.

The scenario sizes in the README were about 10 times too large. Commit `fb020117a` lowered the targets, but no one changed the docs.

For local runs, the `local` environment used `https://find.localhost`. That address needs Caddy, and Caddy needs port 443. A normal user account cannot use port 443.

k6 can map the host name itself. The `local` environment now uses `http://find.localhost`, and k6 sends the requests to `127.0.0.1:3001`. Find only answers on the host `find.localhost`, so the host name must stay.

I did not use `sudo caddy start` or `setcap`. Each developer then needs a privileged step. Also, mise removes the capability when it installs a new Caddy.

For the Grafana project ID, I connected the `cloudOptions` object to `options.cloud`. I did not delete it as dead code. `GRAFANA_PROJECT_ID` is in `.env.example`, so the value has a purpose.

- The workflow can start k6. Before this change, all four scenarios failed immediately.
- `npm run find:dev:quick` runs the quick scenario: 10 users for 25 seconds, with the scenario tags.
- Local load tests need no proxy and no TLS certificate. Start the application with `./bin/dev`, then run the test.
- k6 cloud runs go to the project in `GRAFANA_PROJECT_ID`. Before, the code read the value but never used it, so results went to the default project.
- The token in `.env.example` is now `K6_CLOUD_API_TOKEN`. This is the name that the `grafana:login` script and the workflow already use.
- The README gives the correct user counts, the correct script names, and the list of environments. It also says which commands make load on your machine, and which make load in Grafana Cloud.
- The scenario tags report the correct user counts to Grafana: 80, 200 and 400, not 250, 3000 and 4000.

## Guidance to review

To run the test:

```
./bin/dev
cd load_testing
npm run find:dev:quick
```

To check the options for each environment:

```
cd load_testing
k6 inspect --env SCENARIO=quick --env ENVIRONMENT=local find/load-test.js
k6 inspect --env SCENARIO=quick --env ENVIRONMENT=staging find/load-test.js
```

The `local` output has a `hosts` map. The `staging` output has no `hosts` key.

Add `--env GRAFANA_PROJECT_ID=123456`, and the `cloud` block shows `"projectID": 123456`. The command `yarn standard load_testing/find/config/environment.js load_testing/find/load-test.js` passes.

A local run of `npm run find:dev:quick` made 74 requests, with no 4xx and no 5xx. The response times in a local run are slow, because the application runs in development mode. Do not use them to measure performance.

I am least sure about the workflow. Only a real run can show that the new path is correct, and that run needs the `K6_CLOUD_API_TOKEN` secret.

The workflow also looks for a dashboard URL like `https://app.k6.io/runs/<id>`. Grafana Cloud prints a different form now. A real cloud run must confirm the new form.

## Checklist

- [ ] I have moved hard-coded strings to locale files. Not applicable. This pull request adds no strings to views.
- [ ] I have removed the usage of `data-qa` attributes in HTML files and updated the corresponding tests. Not applicable. This pull request changes no HTML.


inulty-dfe:
[Fix and update load test docs and scripts](https://trello.com/c/vWRVjRSG/2021-fix-and-update-load-test-docs-and-scripts)

tomas-stefano (approved):

load_testing/README.md
```diff
@@ -28,58 +28,89 @@ Comprehensive and scalable load testing suite for **Find** & (future) Publish Te
...
+Every npm script runs `k6 run`, which generates the load from the machine you
+run it on. Only the GitHub Actions workflow runs `k6 cloud`.
 
-**Super quick local test (5 users, 10s):**
-```
-npm run find:dev
-```
+#### Local runs against a development machine
+
+Start the application with `./bin/dev` first. You do not need Caddy. Find only
```
kmcd: `./bin/dev` starts Caddy so this is incorrect.

inulty-dfe: Yes, you're right. I use Procfile.local and I don't run caddy, that must be how this got interpreted

kmcd (approved):
LGTM!

